PR #15364, opened 2016-10-05
[SPARK-17792][ML] L-BFGS solver for linear regression does not accept general numeric label column types

sethah:

Before, we computed `instances` in LinearRegression in two spots, even though they did the same thing. One of them did not cast the label column to `DoubleType`. This patch consolidates the computation and always casts the label column to `DoubleType`.
## How was this patch tested?

Added a unit test to check all solvers. This test failed before this patch.


sethah:
cc @jkbradley @yanboliang

sethah:
also ping @BenFradet

yanboliang (approved):
LGTM

BenFradet:
Lgtm :+1:

yanboliang:
Merged into master, thanks!

PR #14818, opened 2016-08-25
[SPARK-17157][SPARKR][WIP]: Add multiclass logistic regression SparkR Wrapper

wangmiao1981:
## What changes were proposed in this pull request?

(Please fill in changes proposed in this fix)
Add SparkR wrapper for multicalss logistic regression.

Not ready for review now. Only R part code is added.
## How was this patch tested?

(Please explain how this patch was tested. E.g. unit tests, integration tests, manual tests)

(If this patch involves UI changes, please attach a screenshot; otherwise, remove this)

PR #15365, opened 2016-10-05
[SPARK-17157][SPARKR]: Add multiclass logistic regression SparkR Wrapper

wangmiao1981:

As we discussed in #14818, I added a separate R wrapper spark.logit for logistic regression. 

This single interface supports both binary and multinomial logistic regression. It also has "predict" and "summary" for binary logistic regression. 
## How was this patch tested?

New unit tests are added.

### DIFF
--- a/R/pkg/R/mllib.R
+++ b/R/pkg/R/mllib.R
@@ -754,8 +754,9 @@ setMethod("spark.logit", signature(data = "SparkDataFrame", formula = "formula")
             }
 
             jobj <- callJStatic("org.apache.spark.ml.r.LogisticRegressionWrapper", "fit",
-                                data@sdf, formula, as.numeric(regParam), as.numeric(elasticNetParam),
-                                as.integer(maxIter), as.numeric(tol), as.logical(fitIntercept),
+                                data@sdf, formula, as.numeric(regParam),
+                                as.numeric(elasticNetParam), as.integer(maxIter),
+                                as.numeric(tol), as.logical(fitIntercept),
                                 as.character(family), as.logical(standardization),
                                 as.numeric(threshold), thresholds,
                                 as.character(weightCol), as.integer(aggregationDepth))
--- a/R/pkg/R/mllib.R
+++ b/R/pkg/R/mllib.R
@@ -95,6 +95,13 @@ setClass("ALSModel", representation(jobj = "jobj"))
 #' @note KSTest since 2.1.0
 setClass("KSTest", representation(jobj = "jobj"))
 
+#' S4 class that represents an LogisticRegressionModel
+#'
+#' @param jobj a Java object reference to the backing Scala LogisticRegressionModel
+#' @export
+#' @note LogisticRegressionModel since 2.1.0
+setClass("LogisticRegressionModel", representation(jobj = "jobj"))
+
 #' Saves the MLlib model to the input path
 #'
 #' Saves the MLlib model to the input path. For more information, see the specific
@@ -105,7 +112,7 @@ setClass("KSTest", representation(jobj = "jobj"))
 #' @seealso \link{spark.glm}, \link{glm},
 #' @seealso \link{spark.als}, \link{spark.gaussianMixture}, \link{spark.isoreg}, \link{spark.kmeans},
 #' @seealso \link{spark.lda}, \link{spark.mlp}, \link{spark.naiveBayes}, \link{spark.survreg}
-#' @seealso \link{read.ml}
+#' @seealso \link{spark.logit}, \link{read.ml}
 NULL
 
 #' Makes predictions from a MLlib model
@@ -117,7 +124,7 @@ NULL
 #' @export
 #' @seealso \link{spark.glm}, \link{glm},
 #' @seealso \link{spark.als}, \link{spark.gaussianMixture}, \link{spark.isoreg}, \link{spark.kmeans},
-#' @seealso \link{spark.mlp}, \link{spark.naiveBayes}, \link{spark.survreg}
+#' @seealso \link{spark.mlp}, \link{spark.naiveBayes}, \link{spark.survreg}, \link{spark.logit}
 NULL
 
 write_internal <- function(object, path, overwrite = FALSE) {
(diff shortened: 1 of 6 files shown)


wangmiao1981:
@felixcheung  When run check-cran, there are errors:
 Error : requireNamespace("e1071", quietly = TRUE) is not TRUE
Error : requireNamespace("e1071", quietly = TRUE) is not TRUE
Error : requireNamespace("e1071", quietly = TRUE) is not TRUE
Error : requireNamespace("e1071", quietly = TRUE) is not TRUE
Error : requireNamespace("e1071", quietly = TRUE) is not TRUE

I am trying to figure out what is the problem. Any hints?

Thanks!

wangmiao1981:
From Jekins, I saw the error:
WARNING: There was 1 warning.
NOTE: There were 3 notes.
See
  '/home/jenkins/workspace/SparkPullRequestBuilder/R/SparkR.Rcheck/00check.log'
for details.

Had CRAN check errors; see logs.

How can I access the above file?

wangmiao1981:
I see the error message on local test:

LaTeX errors when creating PDF version.
This typically indicates Rd problems.
* checking PDF version of manual without hyperrefs or index ... ERROR
Re-running with no redirection of stdout/stderr.
Hmm ... looks like a package
Error in texi2dvi(file = file, pdf = TRUE, clean = clean, quiet = quiet,  : 
  pdflatex is not available
Error in texi2dvi(file = file, pdf = TRUE, clean = clean, quiet = quiet,  : 
  pdflatex is not available
Error in running tools::texi2pdf()
You may want to clean up by 'rm -rf /var/folders/s_/83b0sgvj2kl2kwq4stvft_pm0000gn/T//RtmpXHJrOk/Rd2pdfac8961d3ab54'
* DONE

vectorijk:
@wangmiao1981 I saw the similar error on Jekin. Same with question with you.
Regarding to `e1071`, I think we only need to install that package locally.

wangmiao1981:
@vectorijk  Thanks for your information! I installed e1071 and installed tex package. I just want to find what causes the error.

wangmiao1981:
My local tests passed.

Jenkins, retest this please.

wangmiao1981:
re-test please

wangmiao1981:
@felixcheung I fixed the cran errors. It is ready to review now. Thanks!

wangmiao1981:
cc @sethah @yanboliang

felixcheung:
It would be great to get some feedback on the name `spark.logit`
What do folks think about it?

R/pkg/R/mllib.R
```diff
@@ -647,6 +654,195 @@ setMethod("predict", signature(object = "KMeansModel"),
...
+#' @name spark.logit
+#' @export
+#' @examples
+#' \dontrun{
+#' sparkR.session()
+#' # binary logistic regression
+#' label <- c(1.0, 1.0, 1.0, 0.0, 0.0)
+#' feature <- c(1.1419053, 0.9194079, -0.9498666, -1.1069903, 0.2809776)
+#' binary_data <- as.data.frame(cbind(label, feature))
+#' binary_df <- suppressWarnings(createDataFrame(binary_data))
```
wangmiao1981: I followed other test examples. If it is not necessary, I can remove all of them as a followup JIRA.

wangmiao1981:
For the name, as we previously discussed, we can't use glm as interface changes and glm only support binominal logistic regression. We don't use glmnet because current spark.logit only provides logistic regressions which are subset of glmnet.

R/pkg/R/mllib.R
felixcheung: Not for all data frame - it is only necessary for data frame with column names containing `.`, eg. `iris`. So not for this case.

felixcheung:
Sure-  I recall that discussion.

wangmiao1981:
I just pick the name for simplicity. Hope to receive feedback from the community and I can make changes accordingly. Thanks!

R/pkg/R/mllib.R
wangmiao1981: I see. Thanks for your explanation!

R/pkg/R/mllib.R
```diff
@@ -105,7 +112,7 @@ setClass("KSTest", representation(jobj = "jobj"))
 #' @seealso \link{spark.glm}, \link{glm},
 #' @seealso \link{spark.als}, \link{spark.gaussianMixture}, \link{spark.isoreg}, \link{spark.kmeans},
 #' @seealso \link{spark.lda}, \link{spark.mlp}, \link{spark.naiveBayes}, \link{spark.survreg}
-#' @seealso \link{read.ml}
+#' @seealso \link{spark.logit}, \link{read.ml}
```
vectorijk: this group of links could be sorted

R/pkg/R/mllib.R
```diff
@@ -117,7 +124,7 @@ NULL
 #' @export
 #' @seealso \link{spark.glm}, \link{glm},
 #' @seealso \link{spark.als}, \link{spark.gaussianMixture}, \link{spark.isoreg}, \link{spark.kmeans},
-#' @seealso \link{spark.mlp}, \link{spark.naiveBayes}, \link{spark.survreg}
+#' @seealso \link{spark.mlp}, \link{spark.naiveBayes}, \link{spark.survreg}, \link{spark.logit}
```
vectorijk: same here.

R/pkg/R/mllib.R
```diff
@@ -647,6 +654,195 @@ setMethod("predict", signature(object = "KMeansModel"),
...
+#'                operators are supported, including '~', '.', ':', '+', and '-'.
+#' @param regParam the regularization parameter. Default is 0.0.
+#' @param elasticNetParam the ElasticNet mixing parameter. For alpha = 0, the penalty is an L2 penalty.
+#'                        For alpha = 1, it is an L1 penalty. For 0 < alpha < 1, the penalty is a combination
+#'                        of L1 and L2. Default is 0.0 which is an L2 penalty.
+#' @param maxIter maximum iteration number.
+#' @param tol convergence tolerance of iterations.
+#' @param fitIntercept whether to fit an intercept term. Default is TRUE.
+#' @param family the name of family which is a description of the label distribution to be used in the model.
+#'               Supported options:
```
felixcheung: roxygen2 is going to trim all the whitespaces, you will need to add formatting with \item and so on

R/pkg/R/mllib.R
```diff
@@ -647,6 +654,195 @@ setMethod("predict", signature(object = "KMeansModel"),
...
+#'                        of models will be always returned on the original scale, so it will be transparent for
+#'                        users. Note that with/without standardization, the models should be always converged
+#'                        to the same solution when no regularization is applied. Default is TRUE, same as glmnet.
+#' @param threshold in binary classification, in range [0, 1]. If the estimated probability of class label 1
+#'                  is > threshold, then predict 1, else 0. A high threshold encourages the model to predict 0
+#'                  more often; a low threshold encourages the model to predict 1 more often. Note: Setting this with
+#'                  threshold p is equivalent to setting thresholds (Array(1-p, p)). When threshold is set, any user-set
+#'                  value for thresholds will be cleared. If both threshold and thresholds are set, then they must be
+#'                  equivalent. Default is 0.5.
+#' @param thresholds in multiclass (or binary) classification to adjust the probability of predicting each class.
```
felixcheung: this might be confusing, in particular with parameter matching in R, should this be `thresholds` when length == 1 for binary and > 1 for multiclass?

R/pkg/R/mllib.R
```diff
@@ -647,6 +654,195 @@ setMethod("predict", signature(object = "KMeansModel"),
...
+#'                 - "multinomial": Multinomial logistic (softmax) regression without pivoting.
+#'                 Default is "auto".
+#' @param standardization whether to standardize the training features before fitting the model. The coefficients
+#'                        of models will be always returned on the original scale, so it will be transparent for
+#'                        users. Note that with/without standardization, the models should be always converged
+#'                        to the same solution when no regularization is applied. Default is TRUE, same as glmnet.
+#' @param threshold in binary classification, in range [0, 1]. If the estimated probability of class label 1
+#'                  is > threshold, then predict 1, else 0. A high threshold encourages the model to predict 0
+#'                  more often; a low threshold encourages the model to predict 1 more often. Note: Setting this with
+#'                  threshold p is equivalent to setting thresholds (Array(1-p, p)). When threshold is set, any user-set
```
felixcheung: do you have a more R-like representation for `(Array(1-p, p))`?

R/pkg/R/mllib.R
```diff
@@ -647,6 +654,195 @@ setMethod("predict", signature(object = "KMeansModel"),
...
+#'                  value for thresholds will be cleared. If both threshold and thresholds are set, then they must be
+#'                  equivalent. Default is 0.5.
+#' @param thresholds in multiclass (or binary) classification to adjust the probability of predicting each class.
+#'                   Array must have length equal to the number of classes, with values > 0, excepting that at most one
```
felixcheung: Array is not often used in R, could we handle vector?

R/pkg/R/mllib.R
felixcheung: so let's remove this suppressWarnings here.

R/pkg/R/mllib.R
```diff
@@ -647,6 +654,195 @@ setMethod("predict", signature(object = "KMeansModel"),
...
+#'
+#' # multinomial logistic regression
+#'
+#' label <- c(0.0, 1.0, 2.0, 0.0, 0.0)
+#' feature1 <- c(4.845940, 5.64480, 7.430381, 6.464263, 5.555667)
+#' feature2 <- c(2.941319, 2.614812, 2.162451, 3.339474, 2.970987)
+#' feature3 <- c(1.322733, 1.348044, 3.861237, 9.686976, 3.447130)
+#' feature4 <- c(1.3246388, 0.5510444, 0.9225810, 1.2147881, 1.6020842)
+#' data <- as.data.frame(cbind(label, feature1, feature2, feature3, feature4))
+#' df <- suppressWarnings(createDataFrame(data))
```
felixcheung: same here for `suppressWarnings`

R/pkg/R/mllib.R
```diff
@@ -647,6 +654,195 @@ setMethod("predict", signature(object = "KMeansModel"),
...
+                                as.numeric(threshold), thresholds,
+                                as.character(weightCol), as.integer(aggregationDepth))
+            new("LogisticRegressionModel", jobj = jobj)
+          })
+
+#  Predicted values based on an LogisticRegressionModel model
+
+#' @param newData a SparkDataFrame for testing.
+#' @return \code{predict} returns the predicted values based on an LogisticRegressionModel.
+#' @rdname spark.logit
```
felixcheung: add `@aliases`

R/pkg/R/mllib.R
```diff
@@ -647,6 +654,195 @@ setMethod("predict", signature(object = "KMeansModel"),
...
+#' @aliases spark.logit,SparkDataFrame,formula-method
+#' @export
+#' @note summary(LogisticRegressionModel) since 2.1.0
+setMethod("summary", signature(object = "LogisticRegressionModel"),
+          function(object) {
+            jobj <- object@jobj
+            is.loaded <- callJMethod(jobj, "isLoaded")
+
+            roc <- if (is.loaded) {
+              NULL
```
felixcheung: instead of return NULL, perhaps we should `stop()` with a message?

felixcheung (changes requested):
How is this related to `spark.glm(family = "binomial")`?

R/pkg/R/mllib.R
```diff
@@ -647,6 +654,195 @@ setMethod("predict", signature(object = "KMeansModel"),
...
+#' df <- suppressWarnings(createDataFrame(data))
+#'
+#' model <- spark.logit(df, label ~ ., family = "multinomial", thresholds=c(0, 1, 1))
+#' predict1 <- collect(select(predict(model, df), "prediction"))
+#' }
+#' @note spark.logit since 2.1.0
+setMethod("spark.logit", signature(data = "SparkDataFrame", formula = "formula"),
+          function(data, formula, regParam = 0.0, elasticNetParam = 0.0, maxIter = 100,
+                   tol = 1E-6, fitIntercept = TRUE, family = "auto", standardization = TRUE,
+                   threshold = 0.5, thresholds = NULL, weightCol = NULL, aggregationDepth = 2) {
```
felixcheung: should it support `probabilityCol`

R/pkg/R/mllib.R
wangmiao1981: I will make changes when we agree on the name. Thanks!

R/pkg/R/mllib.R
wangmiao1981: Refactored the code. Stop() if it is loaded and remove `if` statements in other places.

R/pkg/R/mllib.R
wangmiao1981: Done.

R/pkg/R/mllib.R
wangmiao1981: Done.

R/pkg/R/mllib.R
wangmiao1981: I don't quite understand this comment. There is a JIRA SPARK-11543 discussing the expected relationship, but it is not implemented yet. So, we keep both threshold and thresholds on Scala side.

R/pkg/R/mllib.R
wangmiao1981: Done.

R/pkg/R/mllib.R
felixcheung: perhaps turning that into a list/array on the Scala side so it becomes list on the R side?

wangmiao1981: Let me check whether it is feasible to do that, because the trait is shared by other algorithms.

R/pkg/R/mllib.R
wangmiao1981: Done.

R/pkg/R/mllib.R
wangmiao1981: @felixcheung  I don't find a native method to convert the `Vector` to `Array` when using select on R side.
Optionally, I can add a method `getProbability` in scala wrapper, which converts the Rows of Vectors to a dataframe. Then, on R side, we can call that method in `summary` method. At the same time, we remove the `probabilityCol` from the function definition.
What do you think?

wangmiao1981:
ping @felixcheung

felixcheung:
re: https://github.com/apache/spark/pull/15365#discussion_r84387432
This is definitely an issue as Vector/SparseVector/DenseVector is mapped to environment in R, and within the DataFrame, environment is hard to operate with.

It might likely be how serde/type mapping is handling this but from my digging so far I haven't pinpointed where and how we fix this. It could be separated into another JIRA though.

I'll look at the rest of this PR today.

wangmiao1981:
Thanks for your response! I will do some research on how to bring Vectors back at spare time.

R/pkg/R/mllib.R
```diff
@@ -647,6 +654,173 @@ setMethod("predict", signature(object = "KMeansModel"),
...
+#' with pivoting; "multinomial": Multinomial logistic (softmax) regression without pivoting, similar to glmnet.
+#' Users can print, make predictions on the produced model and save the model to the input path.
+#'
+#' @param data SparkDataFrame for training
+#' @param formula A symbolic description of the model to be fitted. Currently only a few formula
+#'                operators are supported, including '~', '.', ':', '+', and '-'.
+#' @param regParam the regularization parameter. Default is 0.0.
+#' @param elasticNetParam the ElasticNet mixing parameter. For alpha = 0, the penalty is an L2 penalty.
+#'                        For alpha = 1, it is an L1 penalty. For 0 < alpha < 1, the penalty is a combination
+#'                        of L1 and L2. Default is 0.0 which is an L2 penalty.
```
felixcheung: if this is supposed to be numeric, could you change it to 0.0 or 1.0 consistently throughout this comment?

R/pkg/R/mllib.R
felixcheung: suppressWarnings still here?

R/pkg/R/mllib.R
```diff
@@ -647,6 +654,173 @@ setMethod("predict", signature(object = "KMeansModel"),
...
+            predict_internal(object, newData)
+          })
+
+#  Get the summary of an LogisticRegressionModel
+
+#' @param object an LogisticRegressionModel fitted by \code{spark.logit}
+#' @return \code{summary} returns the Binary Logistic regression results of a given model as lists. Note that
+#'                        Multinomial logistic regression summary is not available now.
+#' @rdname spark.logit
+#' @aliases spark.logit,SparkDataFrame,formula-method
```
felixcheung: This is not the right alias..

R/pkg/R/mllib.R
```diff
@@ -647,6 +654,173 @@ setMethod("predict", signature(object = "KMeansModel"),
...
+            }
+
+            roc <- dataFrame(callJMethod(jobj, "roc"))
+
+            areaUnderROC <- callJMethod(jobj, "areaUnderROC")
+
+            pr <- dataFrame(callJMethod(jobj, "pr"))
+
+            fMeasureByThreshold <-
+              dataFrame(callJMethod(jobj, "fMeasureByThreshold"))
```
felixcheung: are there reasons these are all in 2 lines?

R/pkg/inst/tests/testthat/test_mllib.R
```diff
@@ -587,6 +587,59 @@ test_that("spark.isotonicRegression", {
...
 })
 
+test_that("spark.logit", {
+  # test binary logistic regression
+  label <- c(1.0, 1.0, 1.0, 0.0, 0.0)
+  feature <- c(1.1419053, 0.9194079, -0.9498666, -1.1069903, 0.2809776)
+  binary_data <- as.data.frame(cbind(label, feature))
+  binary_df <- createDataFrame(binary_data)
+
+  blr_model <- spark.logit(binary_df, label ~ feature, threshold = 1.0)
```
felixcheung: I would update the example to use `thresholds` instead

R/pkg/inst/tests/testthat/test_mllib.R
```diff
@@ -587,6 +587,59 @@ test_that("spark.isotonicRegression", {
...
+  blr_predict <- collect(select(predict(blr_model, binary_df), "prediction"))
+  expect_equal(blr_predict$prediction, c(0, 0, 0, 0, 0))
+  blr_model1 <- spark.logit(binary_df, label ~ feature, threshold = 0.0)
```
felixcheung: ditto thresholds

R/pkg/R/mllib.R
```diff
@@ -647,6 +654,173 @@ setMethod("predict", signature(object = "KMeansModel"),
...
+#'                        For alpha = 1, it is an L1 penalty. For 0 < alpha < 1, the penalty is a combination
+#'                        of L1 and L2. Default is 0.0 which is an L2 penalty.
+#' @param maxIter maximum iteration number.
+#' @param tol convergence tolerance of iterations.
+#' @param fitIntercept whether to fit an intercept term. Default is TRUE.
+#' @param family the name of family which is a description of the label distribution to be used in the model.
+#'               Supported options:
+#'                 \itemize{
+#'                   \item{"auto": Automatically select the family based on the number of classes:
+#'                           If numClasses == 1 || numClasses == 2, set to "binomial".
```
felixcheung: is there supposed to be a `numClasses` parameter?

mllib/src/main/scala/org/apache/spark/ml/r/LogisticRegressionWrapper.scala
```diff
@@ -0,0 +1,163 @@
...
+
+  private val logisticRegressionModel: LogisticRegressionModel =
+    pipeline.stages(1).asInstanceOf[LogisticRegressionModel]
+
+  lazy val totalIterations: Int = logisticRegressionModel.summary.totalIterations
+
+  lazy val objectiveHistory: Array[Double] = logisticRegressionModel.summary.objectiveHistory
+
+  lazy val roc: DataFrame =
+    logisticRegressionModel.summary.asInstanceOf[BinaryLogisticRegressionSummary].roc
```
felixcheung: consider another `val` for `logisticRegressionModel.summary.asInstanceOf[BinaryLogisticRegressionSummary]` to avoid duplication

felixcheung (changes requested):

R/pkg/R/mllib.R
wangmiao1981: Changed to 0.0 and 1.0 in the comments.

R/pkg/R/mllib.R
wangmiao1981: There is no `numClasses` parameter. The algorithm infers the number of classes. I changed to `number of classes` to avoid confusion.

R/pkg/R/mllib.R
wangmiao1981: `suppressWarnings` is removed. For the aliases, I followed `spark.kmeans`. Are there any specific rules?

R/pkg/R/mllib.R
wangmiao1981: I thought it could exceed the the line limit. Now, I changed it back to 1 line.

R/pkg/inst/tests/testthat/test_mllib.R
wangmiao1981: Updated tests and example.

R/pkg/inst/tests/testthat/test_mllib.R
wangmiao1981: Done.

mllib/src/main/scala/org/apache/spark/ml/r/LogisticRegressionWrapper.scala
wangmiao1981: Add a new `val blrSummary`

R/pkg/R/mllib.R
felixcheung: this is the `summary` method, not `spark.logit` method?

wangmiao1981: Oh. I see. I looked the wrong line number.

wangmiao1981: fixed. Thanks!

felixcheung:
LGTM.

Let's see if anyone has any other comments.

Could you open a JIRA on Vector/SparseVector/DenseVector?

wangmiao1981:
Sure. I will do it. Thanks!

felixcheung:
merged to master.

PR #15366, opened 2016-10-05
[SPARK-17793] [Web UI] Sorting on the description on the Job or Stage page doesn’t always work

ajbozarth:

Added secondary sorting on stage name for the description column. This provide a clearer behavior in the common case where the Description column only comprises of Stage names instead of the option description value.
## How was this patch tested?

manual testing and dev/run-tests

Screenshots of sorting on both description and stage name as well as an example of both:
screen shot 2016-10-04 at 1 09 39 pm
screen shot 2016-10-04 at 1 09 51 pm
screen shot 2016-10-05 at 1 14 45 pm
screen shot 2016-10-05 at 1 14 51 pm
screen shot 2016-10-05 at 4 42 52 pm

### DIFF
--- a/core/src/main/scala/org/apache/spark/ui/jobs/AllJobsPage.scala
+++ b/core/src/main/scala/org/apache/spark/ui/jobs/AllJobsPage.scala
@@ -455,26 +455,11 @@ private[ui] class JobDataSource(
    * Return Ordering according to sortColumn and desc
    */
   private def ordering(sortColumn: String, desc: Boolean): Ordering[JobTableRowData] = {
-    val ordering = sortColumn match {
-      case "Job Id" | "Job Id (Job Group)" => new Ordering[JobTableRowData] {
-        override def compare(x: JobTableRowData, y: JobTableRowData): Int =
-          Ordering.Int.compare(x.jobData.jobId, y.jobData.jobId)
-      }
-      case "Description" => new Ordering[JobTableRowData] {
-        override def compare(x: JobTableRowData, y: JobTableRowData): Int =
-          (x.lastStageDescription, y.lastStageDescription) match {
-            case ("", "") => Ordering.String.compare(x.lastStageName, y.lastStageName)
-            case (xDesc, yDesc) => Ordering.String.compare(xDesc, yDesc)
-          }
-      }
-      case "Submitted" => new Ordering[JobTableRowData] {
-        override def compare(x: JobTableRowData, y: JobTableRowData): Int =
-          Ordering.Long.compare(x.submissionTime, y.submissionTime)
-      }
-      case "Duration" => new Ordering[JobTableRowData] {
-        override def compare(x: JobTableRowData, y: JobTableRowData): Int =
-          Ordering.Long.compare(x.duration, y.duration)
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/apache/spark/ui/jobs/AllJobsPage.scala
```diff
@@ -462,7 +462,10 @@ private[ui] class JobDataSource(
       }
       case "Description" => new Ordering[JobTableRowData] {
         override def compare(x: JobTableRowData, y: JobTableRowData): Int =
-          Ordering.String.compare(x.lastStageDescription, y.lastStageDescription)
+          (x.lastStageDescription, y.lastStageDescription) match {
```
srowen: You know, this and all the other Orderings here could be simpler. This one can be

`case "Description" => Ordering.by(x => (x.lastStageName, x.lastStageDescription))`

Others like the one above should be `Ordering.by(_.jobData.jobId)`. If that's right I think it may be useful to simplify this pattern everywhere while here.

core/src/main/scala/org/apache/spark/ui/jobs/AllJobsPage.scala
```diff
@@ -498,8 +501,7 @@ private[ui] class JobPagedTable(
     sortColumn: String,
     desc: Boolean
   ) extends PagedTable[JobTableRowData] {
-  val parameterPath = UIUtils.prependBaseUri(basePath) + s"/$subPath/?" +
-    parameterOtherTable.mkString("&")
+  val parameterPath = basePath + s"/$subPath/?" + parameterOtherTable.mkString("&")
```
srowen: Is this related?

srowen (changes requested):

core/src/main/scala/org/apache/spark/ui/jobs/AllJobsPage.scala
ajbozarth: Peripherally, `UIUtils.prependBaseUri()` is called on `basePath` before it's passed here, so this is redundant and breaks all the sorting links that rely on `parameterPath` when using yarn. I fixed it here since it was breaking my fix on yarn.

core/src/main/scala/org/apache/spark/ui/jobs/AllJobsPage.scala
ajbozarth: I knew this function seemed bulky, I'll go through and update it as you suggested.

ajbozarth:
Jenkins, retest this please

srowen (approved):

ajbozarth:
Jenkins, retest this please

srowen (approved):
Looks good. That is so much simpler.

ajbozarth:
Jenkin, retest this please

ajbozarth:
It's failed twice now on the same issue but it's completely unrelated to my change. Trying one more time, but is this failure a known issue?

ajbozarth:
Jenkins, retest this please

srowen:
Merged to master

PR #15367, opened 2016-10-06
[SPARK-17346][SQL][test-maven]Add Kafka source for Structured Streaming (branch 2.0)

zsxwing:

Backport https://github.com/apache/spark/commit/9293734d35eb3d6e4fd4ebb86f54dd5d3a35e6db and https://github.com/apache/spark/commit/b678e465afa417780b54db0fbbaa311621311f15 into branch 2.0.

The only difference is the Spark version in pom file.
## How was this patch tested?

Jenkins.

### DIFF
--- a/core/src/main/scala/org/apache/spark/util/UninterruptibleThread.scala
+++ b/core/src/main/scala/org/apache/spark/util/UninterruptibleThread.scala
@@ -90,13 +90,6 @@ private[spark] class UninterruptibleThread(name: String) extends Thread(name) {
   }
 
   /**
-   * Tests whether `interrupt()` has been called.
-   */
-  override def isInterrupted: Boolean = {
-    super.isInterrupted || uninterruptibleLock.synchronized { shouldInterruptThread }
-  }
-
-  /**
    * Interrupt `this` thread if possible. If `this` is in the uninterruptible status, it won't be
    * interrupted until it enters into the interruptible status.
    */
--- a/dev/run-tests.py
+++ b/dev/run-tests.py
@@ -110,7 +110,7 @@ def determine_modules_to_test(changed_modules):
     ['graphx', 'examples']
     >>> x = [x.name for x in determine_modules_to_test([modules.sql])]
     >>> x # doctest: +NORMALIZE_WHITESPACE
-    ['sql', 'hive', 'mllib', 'examples', 'hive-thriftserver',
+    ['sql', 'hive', 'mllib', 'sql-kafka-0-10', 'examples', 'hive-thriftserver',
      'pyspark-sql', 'sparkr', 'pyspark-mllib', 'pyspark-ml']
     """
     modules_to_test = set()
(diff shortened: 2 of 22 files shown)


zsxwing:
cc @tdas

zsxwing:
retest this please

zsxwing:
retest this please

holdenk:
This is rather a lot of functionality to backport, our backport guide doesn't seem to quite line up with this (based on  https://www.mail-archive.com/dev@spark.apache.org/msg10284.html / https://cwiki.apache.org/confluence/display/SPARK/Committers#Committers-PolicyonBackportingBugFixes ) but @pwendell will know better.

zsxwing:
@holdenk this is basically adding an external project. Without the Kafka source, it's hard for people to test Structured Streaming with their real data. Considering 2.0.2 will be out before 2.1.0, it's better to add this one into branch 2.0 in order to test it early. Since Structured Streaming is experimental, it won't break any production usage and should be safe.

/cc @rxin @marmbrus

marmbrus:
We should definitly vet this PR carefully to make sure its safe.  One thing that is missing from that guide, that I do believe is accepted practice, is more leeway when the feature is marked experimental (i.e. during Spark 1.0-1.3 I used to backport pretty much everything that went into SQL).  Since I think many people are simply not using structured streaming due to lack of kafka support, I'd error on the side of more users sooner.  If you think any part of this PR is unsafe with respect stable components though, please do raise those concerns.

tdas:
@zsxwing Have you test the maven build?

zsxwing:
> @zsxwing Have you tested the maven build?



> Test build #66433 has finished for PR 15367 at commit 0915826.
> 
> This patch passes all tests.
> This patch merges cleanly.
> This patch adds no public classes.

This one is maven build.

holdenk:
@marmbrus @zsxwing I agree its experimental and we should have more flexibility here with backports. I also very much agree that structured streaming in its current state on 2.0 isn't usable - but I'm not super sure that backporting fixes is the best way to do this? Honestly I spend most of my time focused on Python & ML (and I've only really been looking at structured streaming with those two hats on).

I'm really cautious about the idea 2k+ line backport which hasn't even been released otherwise but I don't have any specific objections to the changes its just making me nervous. The fact the whats being backported seems to still be under development is also concerning since doing this backport now puts us in a position of backporting more (not yet merged into mainline) fixes.

Of course - If the people with the most experience in this area all agree (and most of y'all [ @marmbrus @zsxwing @tdas but maybe missing @koeninger  ] seem to already be on this PR so I'll leave you to it) that this backport reasonable that is great - it would probably be good to follow up to the original backport mailing list thread and update the wiki as well.

koeninger:
Does backporting reduce the likelihood of change if user feedback indicates we got it wrong?

My technical concerns were largely addressed, that's my big remaining organizational concern.

marmbrus:
No, if we backport this I would plan to continue to backport changes (that are safe) until the next release.  Either way this should not affect what goes into master.

zsxwing:
Thanks! I'm going to merge this one since the concern from @koeninger is addressed.

PR #15368, opened 2016-10-06
[SPARK-17346][SQL][test-maven]Generate the sql test jar to fix the maven build

zsxwing:

Generate the sql test jar to fix the maven build
## How was this patch tested?

Jenkins


tdas:
Since this is a maven change, I am just going to merge it after compilation passes, to unbreak the master branch.

external/kafka-0-10-sql/src/test/scala/org/apache/spark/sql/kafka010/KafkaSourceSuite.scala
```diff
@@ -151,6 +151,7 @@ class KafkaSourceSuite extends KafkaSourceTest {
     val mapped = kafka.map(kv => kv._2.toInt + 1)
 
     testStream(mapped)(
+      makeSureGetOffsetCalled,
```
zsxwing: @tdas I noticed this test will fail with InterruptedException if we stop the stream before the stream thread starts. I just added this line to make sure we will stop the stream after the thread starts.

PR #15369, opened 2016-10-06
[SPARK-17795] [Web UI] Sorting on stage or job tables doesn’t reload page on that table

ajbozarth:

Added anchor on table header id to sorting links on job and stage tables. This make the page reload after a sort load the page at the sorted table. 

This only changes page load behavior so no UI changes
## How was this patch tested?

manually tested and dev/run-tests

### DIFF
--- a/core/src/main/scala/org/apache/spark/ui/jobs/StageTable.scala
+++ b/core/src/main/scala/org/apache/spark/ui/jobs/StageTable.scala
@@ -34,6 +34,7 @@ import org.apache.spark.util.Utils
 private[ui] class StageTableBase(
     request: HttpServletRequest,
     stages: Seq[StageInfo],
+    tableHeaderID: String,
     stageTag: String,
     basePath: String,
     subPath: String,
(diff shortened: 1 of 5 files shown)


srowen (approved):
Looks good though I'm missing where the table ID gets added to the output, or was it already there? if it were there I guess I'd have expected it was already a parameter passed around to these places, but I haven't looked into it.

core/src/main/scala/org/apache/spark/ui/jobs/StageTable.scala
```diff
@@ -241,7 +246,8 @@ private[ui] class StagePagedTable(
             val headerLink = Unparsed(
               parameterPath +
                 s"&$stageTag.sort=${URLEncoder.encode(header, "UTF-8")}" +
-                s"&$stageTag.pageSize=$pageSize")
+                s"&$stageTag.pageSize=$pageSize") +
+                s"#$tableHeaderId"
```
ajbozarth: lines like this are where `tableHeaderId` is used

ajbozarth: yeah those was already there for the quick links at the top of those pages

srowen:
merged to master. Seems like a clean improvement.

PR #15371, opened 2016-10-06
[SPARK-17816] [Core] Fix ConcurrentModificationException issue in BlockStatusesAccumulator

seyfe:

Change the BlockStatusesAccumulator to return immutable object when value method is called.
## How was this patch tested?

Existing tests plus I verified this change by running a pipeline which consistently repro this issue. 

This is the stack trace for this exception:
`
java.util.ConcurrentModificationException
        at java.util.ArrayList$Itr.checkForComodification(ArrayList.java:901)
        at java.util.ArrayList$Itr.next(ArrayList.java:851)
        at scala.collection.convert.Wrappers$JIteratorWrapper.next(Wrappers.scala:43)
        at scala.collection.generic.Growable$class.$plus$plus$eq(Growable.scala:59)
        at scala.collection.mutable.ListBuffer.$plus$plus$eq(ListBuffer.scala:183)
        at scala.collection.mutable.ListBuffer.$plus$plus$eq(ListBuffer.scala:45)
        at scala.collection.TraversableLike$class.to(TraversableLike.scala:590)
        at scala.collection.AbstractTraversable.to(Traversable.scala:104)
        at scala.collection.TraversableOnce$class.toList(TraversableOnce.scala:294)
        at scala.collection.AbstractTraversable.toList(Traversable.scala:104)
        at org.apache.spark.util.JsonProtocol$.accumValueToJson(JsonProtocol.scala:314)
        at org.apache.spark.util.JsonProtocol$$anonfun$accumulableInfoToJson$5.apply(JsonProtocol.scala:291)
        at org.apache.spark.util.JsonProtocol$$anonfun$accumulableInfoToJson$5.apply(JsonProtocol.scala:291)
        at scala.Option.map(Option.scala:146)
        at org.apache.spark.util.JsonProtocol$.accumulableInfoToJson(JsonProtocol.scala:291)
        at org.apache.spark.util.JsonProtocol$$anonfun$taskInfoToJson$12.apply(JsonProtocol.scala:283)
        at org.apache.spark.util.JsonProtocol$$anonfun$taskInfoToJson$12.apply(JsonProtocol.scala:283)
        at scala.collection.generic.TraversableForwarder$class.foreach(TraversableForwarder.scala:35)
        at scala.collection.mutable.ListBuffer.foreach(ListBuffer.scala:45)
        at scala.collection.AbstractTraversable.map(Traversable.scala:104)
        at org.apache.spark.util.JsonProtocol$.taskInfoToJson(JsonProtocol.scala:283)
        at org.apache.spark.util.JsonProtocol$.taskEndToJson(JsonProtocol.scala:145)
        at org.apache.spark.util.JsonProtocol$.sparkEventToJson(JsonProtocol.scala:76)
`

### DIFF
--- a/core/src/main/scala/org/apache/spark/executor/TaskMetrics.scala
+++ b/core/src/main/scala/org/apache/spark/executor/TaskMetrics.scala
@@ -330,7 +330,7 @@ private[spark] class BlockStatusesAccumulator
 
   override def copy(): BlockStatusesAccumulator = {
     val newAcc = new BlockStatusesAccumulator
-    newAcc.merge(this)
+    newAcc.setValue(this.value)
     newAcc
   }
 }
--- a/core/src/main/scala/org/apache/spark/executor/TaskMetrics.scala
+++ b/core/src/main/scala/org/apache/spark/executor/TaskMetrics.scala
@@ -54,7 +54,7 @@ class TaskMetrics private[spark] () extends Serializable {
   private val _memoryBytesSpilled = new LongAccumulator
   private val _diskBytesSpilled = new LongAccumulator
   private val _peakExecutionMemory = new LongAccumulator
-  private val _updatedBlockStatuses = new BlockStatusesAccumulator
+  private val _updatedBlockStatuses = new CollectionAccumulator[(BlockId, BlockStatus)]
 
   /**
    * Time taken on the executor to deserialize this task.
@@ -319,18 +319,5 @@ private[spark] object TaskMetrics extends Logging {
 
     tm.externalAccums ++= externalAccums
     tm
-  }
-}
-
-
-private[spark] class BlockStatusesAccumulator
-  extends CollectionAccumulator[(BlockId, BlockStatus)] {
-
-  override def copyAndReset(): BlockStatusesAccumulator = new BlockStatusesAccumulator
-
-  override def copy(): BlockStatusesAccumulator = {
-    val newAcc = new BlockStatusesAccumulator
-    newAcc.setValue(this.value)
-    newAcc
   }
 }
--- a/core/src/main/scala/org/apache/spark/executor/TaskMetrics.scala
+++ b/core/src/main/scala/org/apache/spark/executor/TaskMetrics.scala
@@ -115,11 +115,7 @@ class TaskMetrics private[spark] () extends Serializable {
   /**
    * Storage statuses of any blocks that have been updated as a result of this task.
    */
-  def updatedBlockStatuses: Seq[(BlockId, BlockStatus)] = {
-    // This is called on driver. All accumulator updates have a fixed value. So it's safe to use
-    // `asScala` which accesses the internal values using `java.util.Iterator`.
-    _updatedBlockStatuses.value.asScala
-  }
+  def updatedBlockStatuses: List[(BlockId, BlockStatus)] = _updatedBlockStatuses.value
 
   // Setters and increment-ers
   private[spark] def setExecutorDeserializeTime(v: Long): Unit =
@@ -137,9 +133,7 @@ class TaskMetrics private[spark] () extends Serializable {
   private[spark] def incPeakExecutionMemory(v: Long): Unit = _peakExecutionMemory.add(v)
   private[spark] def incUpdatedBlockStatuses(v: (BlockId, BlockStatus)): Unit =
     _updatedBlockStatuses.add(v)
-  private[spark] def setUpdatedBlockStatuses(v: java.util.List[(BlockId, BlockStatus)]): Unit =
-    _updatedBlockStatuses.setValue(v)
-  private[spark] def setUpdatedBlockStatuses(v: Seq[(BlockId, BlockStatus)]): Unit =
+  private[spark] def setUpdatedBlockStatuses(v: List[(BlockId, BlockStatus)]): Unit =
     _updatedBlockStatuses.setValue(v.asJava)
 
   /**
@@ -295,7 +289,7 @@ private[spark] object TaskMetrics extends Logging {
       val name = info.name.get
       val value = info.update.get
       if (name == UPDATED_BLOCK_STATUSES) {
-        tm.setUpdatedBlockStatuses(value.asInstanceOf[java.util.List[(BlockId, BlockStatus)]])
+        tm.setUpdatedBlockStatuses(value.asInstanceOf[List[(BlockId, BlockStatus)]])
       } else {
         tm.nameToAccums.get(name).foreach(
           _.asInstanceOf[LongAccumulator].setValue(value.asInstanceOf[Long])
@@ -326,7 +320,7 @@ private[spark] object TaskMetrics extends Logging {
 
 
 private[spark] class BlockStatusesAccumulator
-  extends AccumulatorV2[(BlockId, BlockStatus), java.util.List[(BlockId, BlockStatus)]] {
+  extends AccumulatorV2[(BlockId, BlockStatus), List[(BlockId, BlockStatus)]] {
   private val _seq = Collections.synchronizedList(new ArrayList[(BlockId, BlockStatus)]())
 
   override def isZero(): Boolean = _seq.isEmpty
@@ -344,15 +338,20 @@ private[spark] class BlockStatusesAccumulator
   override def add(v: (BlockId, BlockStatus)): Unit = _seq.add(v)
 
   override def merge(
-    other: AccumulatorV2[(BlockId, BlockStatus), java.util.List[(BlockId, BlockStatus)]]): Unit = {
+    other: AccumulatorV2[(BlockId, BlockStatus), List[(BlockId, BlockStatus)]]): Unit = {
     other match {
-      case o: BlockStatusesAccumulator => _seq.addAll(o.value)
+      case o: BlockStatusesAccumulator => o.value.foreach(_seq.add)
       case _ => throw new UnsupportedOperationException(
         s"Cannot merge ${this.getClass.getName} with ${other.getClass.getName}")
     }
   }
 
-  override def value: java.util.List[(BlockId, BlockStatus)] = _seq
+  // `asScala` accesses the internal values using `java.util.Iterator` so needs to be synchronized
+  override def value: List[(BlockId, BlockStatus)] = {
+    _seq.synchronized {
+      _seq.asScala.toList
+    }
+  }
 
   def setValue(newValue: java.util.List[(BlockId, BlockStatus)]): Unit = {
     _seq.clear()
--- a/core/src/main/scala/org/apache/spark/executor/TaskMetrics.scala
+++ b/core/src/main/scala/org/apache/spark/executor/TaskMetrics.scala
@@ -17,8 +17,6 @@
 
 package org.apache.spark.executor
 
-import java.util.{ArrayList, Collections}
-
 import scala.collection.JavaConverters._
 import scala.collection.mutable.{ArrayBuffer, LinkedHashMap}
 
@@ -27,7 +25,7 @@ import org.apache.spark.annotation.DeveloperApi
 import org.apache.spark.internal.Logging
 import org.apache.spark.scheduler.AccumulableInfo
 import org.apache.spark.storage.{BlockId, BlockStatus}
-import org.apache.spark.util.{AccumulatorContext, AccumulatorMetadata, AccumulatorV2, LongAccumulator}
+import org.apache.spark.util._
 
 
 /**
@@ -115,7 +113,11 @@ class TaskMetrics private[spark] () extends Serializable {
   /**
    * Storage statuses of any blocks that have been updated as a result of this task.
    */
-  def updatedBlockStatuses: List[(BlockId, BlockStatus)] = _updatedBlockStatuses.value
+  def updatedBlockStatuses: Seq[(BlockId, BlockStatus)] = {
+    // This is called on driver. All accumulator updates have a fixed value. So it's safe to use
+    // `asScala` which accesses the internal values using `java.util.Iterator`.
+    _updatedBlockStatuses.value.asScala
+  }
 
   // Setters and increment-ers
   private[spark] def setExecutorDeserializeTime(v: Long): Unit =
@@ -133,7 +135,9 @@ class TaskMetrics private[spark] () extends Serializable {
   private[spark] def incPeakExecutionMemory(v: Long): Unit = _peakExecutionMemory.add(v)
   private[spark] def incUpdatedBlockStatuses(v: (BlockId, BlockStatus)): Unit =
     _updatedBlockStatuses.add(v)
-  private[spark] def setUpdatedBlockStatuses(v: List[(BlockId, BlockStatus)]): Unit =
+  private[spark] def setUpdatedBlockStatuses(v: java.util.List[(BlockId, BlockStatus)]): Unit =
+    _updatedBlockStatuses.setValue(v)
+  private[spark] def setUpdatedBlockStatuses(v: Seq[(BlockId, BlockStatus)]): Unit =
     _updatedBlockStatuses.setValue(v.asJava)
 
   /**
@@ -289,7 +293,7 @@ private[spark] object TaskMetrics extends Logging {
       val name = info.name.get
       val value = info.update.get
       if (name == UPDATED_BLOCK_STATUSES) {
-        tm.setUpdatedBlockStatuses(value.asInstanceOf[List[(BlockId, BlockStatus)]])
+        tm.setUpdatedBlockStatuses(value.asInstanceOf[java.util.List[(BlockId, BlockStatus)]])
       } else {
         tm.nameToAccums.get(name).foreach(
           _.asInstanceOf[LongAccumulator].setValue(value.asInstanceOf[Long])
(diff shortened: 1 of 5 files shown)


hvanhovell:
ok to test

seyfe:
Unfortunately, this PR doesn't fix the java.util.ConcurrentModificationException. I can still repro it. I will spend more time on it tomorrow morning.

core/src/main/scala/org/apache/spark/util/JsonProtocol.scala
```diff
@@ -312,11 +312,15 @@ private[spark] object JsonProtocol {
         // We only have 3 kind of internal accumulator types, so if it's not int or long, it must be
         // the blocks accumulator, whose type is `java.util.List[(BlockId, BlockStatus)]`
         case v =>
-          JArray(v.asInstanceOf[java.util.List[(BlockId, BlockStatus)]].asScala.toList.map {
-            case (id, status) =>
-              ("Block ID" -> id.toString) ~
-              ("Status" -> blockStatusToJson(status))
-          })
+          val blockAccumulator =
```
srowen: I don't see why this would help. You add a wrapper, but, synchronizing your local access to it doesn't do anything because nothing else is synchronizing on it.

PS can the `List[(BlockId,BlockStatus)]` type just be part of the match predicate?

core/src/main/scala/org/apache/spark/util/JsonProtocol.scala
```diff
@@ -281,7 +281,7 @@ private[spark] object JsonProtocol {
     ("Finish Time" -> taskInfo.finishTime) ~
     ("Failed" -> taskInfo.failed) ~
     ("Killed" -> taskInfo.killed) ~
-    ("Accumulables" -> JArray(taskInfo.accumulables.map(accumulableInfoToJson).toList))
+    ("Accumulables" -> JArray(taskInfo.accumulables.toList.map(accumulableInfoToJson)))
```
srowen: What does this do -- just puts a copy before the work of mapping? I could see how that would tend to help.

srowen (changes requested):
Yeah I'm not sure this is the issue. It's being modified while it's being serialized.

rxin:
One comment: once you figured out the proper fix, please add some comments inline so they don't get accidentally removed in the future.

zsxwing:
@seyfe Thanks for reporting this one. Actually, it's different from SPARK-17463. Could you create a new ticket for this issue, please? The cause is we send a mutable TaskInfo to listeners but we may still update TaskInfo's fields (e.g., accumulables) in another thread... Ideally, all events sent to the listeners should be immutable.

seyfe:
Hi @zsxwing. I have a fix ready and testing it now. I will create a new ticket and send an updated PR today.

core/src/main/scala/org/apache/spark/util/JsonProtocol.scala
seyfe: This wasn't the root cause but it's something nice to have. If you prefer, I can revert this line.

zsxwing:
@seyfe The issue is `TaskInfo.accumulables` is accessed in multiple threads without synchronization. `TaskMetrics.updatedBlockStatuses` is fine.

seyfe:
@zsxwing. `BlockStatusesAccumulator` is using synchronizedList which is `_seq` for concurrency. But when the `value` method is executed and the output is used by `asScala.toList` it caused the `ConcurrentModificationException`. 

`Collections.synchronizedList` suggests using synchronized keyword when using its iterator which is done by asScala method. Does that make sense?

zsxwing:
> @zsxwing. BlockStatusesAccumulator is using synchronizedList which is _seq for concurrency. But when the value method is executed and the output is used by asScala.toList it caused the ConcurrentModificationException.
> 
> Collections.synchronizedList suggests using synchronized keyword when using its iterator which is done by asScala method. Does that make sense?

@seyfe The comment you are deleting explains why it's safe: the driver doesn't modify `BlockStatusesAccumulator`.

seyfe:
@zsxwing . `JArray(v.asInstanceOf[java.util.List[(BlockId, BlockStatus)]].asScala.toList.map` this is the line of code which is causing ConcurrentModificationException. And you can see that from call stack. This is coming from BlockStatusesAccumulator. Am I missing something here?

seyfe:
I also want to point out that below is the core part of fix. Rest of the code changes are side-effects of it.

```
+  // `asScala` accesses the internal values using `java.util.Iterator` so needs to be synchronized
 +  override def value: List[(BlockId, BlockStatus)] = {
 +    _seq.synchronized {
 +      _seq.asScala.toList
 +    }
 +  }
```

core/src/main/scala/org/apache/spark/util/JsonProtocol.scala
srowen: Seems OK if it's related cleanup, and potentially helps a closely related manifestation

srowen (commented):
Dumb question but how big is this? the solution here is to copy the data structure, which is a good defensive move, as long as it's not big and nothing is actually relying on observing changes to the underlying data. Is that valid?

seyfe:
Hi @srowen, 

This PR doesn't introduce any extra data copy operations. It moves the data copy code from `JsonProtocol:accumValueToJson` method to `BlockStatusesAccumulator:value` method so it can be called inside the synchronized block. But let me try to answer your question. 

I checked the data size using 3 different pipelines. 99% of the time ArrayList has less than 4 items. There was one of case where it maxed at 4000 items but that was less than 1% of the time. I ran my test with 4000 executors, I think that is why this 4000 number came up.

I debated other options as well. Moving Json serialization into `BlockStatusesAccumulator` would help but I didn't want to mix match these classes. I considered introducing Akka Actors for thread safety but I think that would be an overkill in this scenario since data size is small. That is why I proposed this solution but I would love to hear your proposals.

I don't know the answer for the second part of your question (below), but existing behavior is not changed. Only change is that we can convert ArrayList to a Scala List inside a synchronized block so we won't get `ConcurrentModificationException` . If we do need to rely on observing changes, I think we do need to have a bigger change. @zsxwing, would you please comment on that part? 
>  nothing is actually relying on observing changes to the underlying data

srowen:
I see the new copy (of course or else this wouldn't help) but where is a copy removed? I'm probably overlooking it. A `toList` call was moved, which we discussed above briefly. I still think it's all fine, just worth double-checking both of our logics.

seyfe:
Hi @srowen , this is the part that I removed extra copy operation. I changed this line because this conversion is already done by BlockStatusesAccumulator.

`JArray(v.asInstanceOf[java.util.List[(BlockId, BlockStatus)]].asScala.toList.map {`
with
`JArray(v.map {`

zsxwing:
@seyfe I'm taking my words back. Yea, `BlockStatusesAccumulator.merge` will be called in driver.

core/src/main/scala/org/apache/spark/executor/TaskMetrics.scala
```diff
@@ -344,15 +338,20 @@ private[spark] class BlockStatusesAccumulator
...
+      case o: BlockStatusesAccumulator => o.value.foreach(_seq.add)
       case _ => throw new UnsupportedOperationException(
         s"Cannot merge ${this.getClass.getName} with ${other.getClass.getName}")
     }
   }
 
-  override def value: java.util.List[(BlockId, BlockStatus)] = _seq
+  // `asScala` accesses the internal values using `java.util.Iterator` so needs to be synchronized
+  override def value: List[(BlockId, BlockStatus)] = {
+    _seq.synchronized {
```
zsxwing: `_seq.synchronized` is wrong. `Collections.synchronizedList` uses its internal `mutex` to lock instead of `this`.

Why changes them to Scala List? Just change this one to `java.util.Collections.unmodifiableList(new ArrayList[(BlockId, BlockStatus)](_seq))` should be enough.

seyfe: Thanks for the review @zsxwing . I checked the java doc and it says that getting iterator is not thread safe and suggests below usage. That's why I did `_seq.synchronized`

https://docs.oracle.com/javase/7/docs/api/java/util/Collections.html

```
  List list = Collections.synchronizedList(new ArrayList());
      ...
  synchronized (list) {
      Iterator i = list.iterator(); // Must be in synchronized block
      while (i.hasNext())
          foo(i.next());
  }
 
```

Regarding your second questions, `JsonProtocal` is using it as Scala collection that is why I converted it to a Scala collection so we won't need to convert again.

seyfe:
Hi @zsxwing ,

The test failed with below error. I don't think that it's related with my change. Should we just re-run the test or do you have any suggestion?

```
java.lang.RuntimeException: spark-core: Binary compatibility check failed!
[error] (core/*:mimaReportBinaryIssues) spark-core: Binary compatibility check failed!
```

zsxwing:
retest this please

seyfe:
I don't know if it's related but I found a bug in the last iteration. We also need to override copy and copyAndReset methods. Otherwise it throws java.lang.ClassCastException error.

seyfe:
@zsxwing , I built it and the test pipelines works fine. So fix is good. But I don't know how to fix the MiMa tests. Would you mind helping me on this?

zsxwing:
You can find the following errors on Jenkins:
```
[error]  * abstract method merge(org.apache.spark.util.AccumulatorV2)Unit in class org.apache.spark.util.AccumulatorV2 does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectAbstractMethodProblem]("org.apache.spark.util.AccumulatorV2.merge")
[error]  * abstract method isZero()Boolean in class org.apache.spark.util.AccumulatorV2 does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectAbstractMethodProblem]("org.apache.spark.util.AccumulatorV2.isZero")
[error]  * abstract method reset()Unit in class org.apache.spark.util.AccumulatorV2 does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectAbstractMethodProblem]("org.apache.spark.util.AccumulatorV2.reset")
```

Just add these exclude rules to `project/MimaExcludes.scala`.

project/MimaExcludes.scala
```diff
@@ -833,6 +833,11 @@ object MimaExcludes {
       // [SPARK-12221] Add CPU time to metrics
       ProblemFilters.exclude[DirectMissingMethodProblem]("org.apache.spark.status.api.v1.TaskMetrics.this"),
       ProblemFilters.exclude[DirectMissingMethodProblem]("org.apache.spark.status.api.v1.TaskMetricDistributions.this")
+    ) ++ Seq(
```
zsxwing: They should be in `lazy val v21excludes = v20excludes`. Master is supposed to be released as 2.1.

seyfe: Sure, I update it.

core/src/main/scala/org/apache/spark/executor/TaskMetrics.scala
```diff
@@ -326,36 +324,13 @@ private[spark] object TaskMetrics extends Logging {
...
-
-  override def isZero(): Boolean = _seq.isEmpty
+  extends CollectionAccumulator[(BlockId, BlockStatus)] {
 
   override def copyAndReset(): BlockStatusesAccumulator = new BlockStatusesAccumulator
 
   override def copy(): BlockStatusesAccumulator = {
     val newAcc = new BlockStatusesAccumulator
-    newAcc._seq.addAll(_seq)
+    newAcc._list.addAll(_list)
```
srowen: Don't you just want to call merge() here? it's better not to expose the internal representation even to subclasses if possible, because we're exposing it permanently to everyone.

seyfe: Good point. I will do the proposed change. I am also going to change the base class (`CollectionAccumulator`) because behavior of `addAll` is undefined if the specified collection is modified and `merge()` is working on immutable copy when calling `addAll`.

project/MimaExcludes.scala
```diff
@@ -55,7 +55,11 @@ object MimaExcludes {
       ProblemFilters.exclude[ReversedMissingMethodProblem]("org.apache.spark.sql.catalog.Catalog.getFunction"),
       ProblemFilters.exclude[ReversedMissingMethodProblem]("org.apache.spark.sql.catalog.Catalog.databaseExists"),
       ProblemFilters.exclude[ReversedMissingMethodProblem]("org.apache.spark.sql.catalog.Catalog.tableExists"),
-      ProblemFilters.exclude[ReversedMissingMethodProblem]("org.apache.spark.sql.catalog.Catalog.functionExists")
+      ProblemFilters.exclude[ReversedMissingMethodProblem]("org.apache.spark.sql.catalog.Catalog.functionExists"),
+      // [SPARK-17816] Fix ConcurrentModificationException issue in BlockStatusesAccumulator
+      ProblemFilters.exclude[DirectAbstractMethodProblem]("org.apache.spark.util.AccumulatorV2.merge"),
```
srowen: I don't think this change should be necessary?

core/src/main/scala/org/apache/spark/util/AccumulatorV2.scala
```diff
@@ -444,7 +444,7 @@ class CollectionAccumulator[T] extends AccumulatorV2[T, java.util.List[T]] {
 
   override def copy(): CollectionAccumulator[T] = {
     val newAcc = new CollectionAccumulator[T]
-    newAcc._list.addAll(_list)
+    newAcc.merge(this)
```
srowen: Yeah I see why this is probably worth fixing in the base class too. The only concern I have is that `merge` causes a copy of the argument's list, when here we could get away with `_list.synchronized { newAcc._list.addAll(_list) }`? It would also avoid a tiny behavior change, that subclasses would now find copy calls merge (though I don't know of a reason that would be a problem now.) We could extend the same logic to `merge` to avoid calling `value` but it would require holding locks on two lists at once and there's an opening for a deadlock there, I think, that I don't want to mess with.

srowen (changes requested):

project/MimaExcludes.scala
seyfe: But MiMa tests are consistently failing if I don't add these exclusions.
i.e. https://amplab.cs.berkeley.edu/jenkins/job/SparkPullRequestBuilder/66541/consoleFull

core/src/main/scala/org/apache/spark/util/AccumulatorV2.scala
seyfe: I don't have a strong opinion so let me change `merge()` with `synchronized` block as you suggested. 

I also propose changing `BlockStatusesAccumulator` to use `setValue()` rather than `merge()` to avoid extra type matching.

core/src/main/scala/org/apache/spark/util/AccumulatorV2.scala
```diff
@@ -444,7 +444,9 @@ class CollectionAccumulator[T] extends AccumulatorV2[T, java.util.List[T]] {
 
   override def copy(): CollectionAccumulator[T] = {
     val newAcc = new CollectionAccumulator[T]
-    newAcc._list.addAll(_list)
+    _list.synchronized {
```
zsxwing: Good catch

zsxwing:
@seyfe I think we can remove `BlockStatusesAccumulator` and just use `private val _updatedBlockStatuses = new CollectionAccumulator[(BlockId, BlockStatus)]` instead. `BlockStatusesAccumulator` doesn't provide more functions than `CollectionAccumulator`. Sorry that I didn't find that early.

seyfe:
@zsxwing , I think that is a good idea. I search it and that is the only place we use `BlockStatusesAccumulator`. Let me remove it.

zsxwing:
retest this please

project/MimaExcludes.scala
srowen: Hm, but this class didn't have any change in its API, not at this point. I can't see why it would still be needed. In fact there's a merge conflict here anyway, can you just try removing it and see if it passes?

Looking strong to me otherwise, great cleanup.

core/src/main/scala/org/apache/spark/util/AccumulatorV2.scala
mridulm: I dont think this helps since all modifications to _list is happening without it being synchronized. What am I missing here ?

mridulm: (context) synchronizedList is not protected by 'this', but by a private variable - so the locks are different.

seyfe: @mridulm , I don't think that your statement is correct. `SynchronizedCollection` is protected by
`         final Object mutex;     // Object on which to synchronize
`

If you look at the constructor you will see that
`            mutex = this;
`

Does that make sense now?

mridulm: You are right, in this codepath, it is indeed protected by mutex = this.

zsxwing:

zsxwing:
There are some conflicts with 2.0. @seyfe could you submit a PR for branch-2.0, please? Thanks!

seyfe:
Thanks @zsxwing.

Here is the PR for branch-2.0
https://github.com/apache/spark/pull/15425

PR #15373, opened 2016-10-06
[SPARK-17798][SQL] Remove redundant Experimental annotations in sql.streaming

rxin:

I was looking through API annotations to catch mislabeled APIs, and realized DataStreamReader and DataStreamWriter classes are already annotated as Experimental, and as a result there is no need to annotate each method within them.
## How was this patch tested?

N/A


rxin:
cc @zsxwing

zsxwing:
LGTM

zsxwing:

PR #15374, opened 2016-10-06
[SPARK-17800] Introduce InterfaceStability annotation

rxin:

This patch introduces three new annotations under InterfaceStability:
- Stable
- Evolving
- Unstable

This is inspired by Hadoop's InterfaceStability, and the first step towards switching over to a new API stability annotation framework.
## How was this patch tested?

N/A

### DIFF
--- a/common/tags/src/main/java/org/apache/spark/annotation/InterfaceStability.java
+++ b/common/tags/src/main/java/org/apache/spark/annotation/InterfaceStability.java
@@ -1,0 +1,49 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.spark.annotation;
+
+import java.lang.annotation.Documented;
+
+/**
+ * Annotation to inform users of how much to rely on a particular package,
+ * class or method not changing over time.
+ */
+public class InterfaceStability {
+
+  /**
+   * Stable APIs that retain source and binary compatibility within a major release.
+   * These interfaces can change from one major release to another major release
+   * (e.g. from 1.0 to 2.0).
+   */
+  @Documented
+  public @interface Stable {};
+
+  /**
+   * APIs that are meant to evolve towards becoming stable APIs, but are not stable APIs yet.
+   * Evolving interfaces can change from one feature release to another release (i.e. 2.1 to 2.2).
+   */
+  @Documented
+  public @interface Evolving {};
+
+  /**
+   * Unstable APIs, with no guarantee on stability.
+   * Classes that are unannotated are considered Unstable.
+   */
+  @Documented
+  public @interface Unstable {};
+}


mridulm:
👍 LGTM

common/tags/src/main/java/org/apache/spark/annotation/InterfaceStability.java
```diff
@@ -0,0 +1,49 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.spark.annotation;
+
+import java.lang.annotation.Documented;
+
+/**
+ * Annotation to inform users of how much to rely on a particular package,
+ * class or method not changing over time.
```
tgravescs: Are we only going to be applying these to Public apis?  I guess that question might go with the InterfaceAudience change.  Just wondering if we wanted to state the default is Unstable for private classes but we can make that change later

tgravescs:
sorry Reynold, looks like I somehow missed your response on email about this. 

Comment question, otherwise looks good.

common/tags/src/main/java/org/apache/spark/annotation/InterfaceStability.java
rxin: I am not including the InterfaceAudience change for now since it is less clear what the "audience" should be. In addition, the biggest bang for the buck does come from this stability annotation, and less from audience.

For private APIs -- what are the reasons you think we should annotate private APIs with stability levels?

vanzin:
+1. I also agree that audience is not as useful as these; the three annotations here should cover the vast majority of use cases without the need to differentiate who's using the API.

common/tags/src/main/java/org/apache/spark/annotation/InterfaceStability.java
tgravescs: I probably should have said do we want to say something like classes not annotated are considered Unstable.  This would just cover if we missed something or clarify to the user if they found a class that wasn't annotated.  

It really only applies once we have the Audience tag and what those audience tags are.  You could have a tag that is for internal Spark developers/components (not applications using spark) which would apply to other components or components really closely tied to Spark (perhaps Bahir). In the Hadoop world some things are considered limited private (which I don't think we should use that tag) but are allowed to use some interfaces end user is not.

Not a big deal at this point.

rxin: That's a good point. I did say in the doc for unstable anything unlabeled are considered unstable. I can promote it up here to make that message more visible; perhaps when we finish annotating most classes (otherwise all classes are considered unstable immediately).

mridulm: I was assuming that classes/interfaces not annotated with InterfaceStability or other org.apache.spark.annotation should be considered private and not exposed (which is default irrespective of private[spark] exists or not).

Only those which are explicitly tagged are public.
A public api can still be Unstable, but the assumption is that there might be an intention to open it up based on feedback/design evolution.

Is this a correct understanding ?

rxin: Yes -- but private apis are by definition "unstable" ...

common/tags/src/main/java/org/apache/spark/annotation/InterfaceStability.java
```diff
@@ -0,0 +1,49 @@
...
+   */
+  @Documented
+  public @interface Stable {};
+
+  /**
+   * APIs that are meant to evolve towards becoming stable APIs, but are not stable APIs yet.
+   * Evolving interfaces can change from one feature release to another release (i.e. 2.1 to 2.2).
+   */
+  @Documented
+  public @interface Evolving {};
```
viirya: I can't see significant difference between `Evolving` and `Unstable` in the view of changing possibility. Does it guarantee that an `Evolving` API becomes `Stable` one in next major release?

rxin: No but for evolving API we at least say it won't change within a feature release, whereas unstable API has no guarantee at all.

rxin:
Alright I'm going to merge this in master. I will go through some files to do annotation, and new lessons might come up from annotating those.

common/tags/src/main/java/org/apache/spark/annotation/InterfaceStability.java
vanzin: In my view "Evolving" would mean "this is sort of what the API will look like, there might be some adjustments in the next release, so you have to be willing to rev your code when the next release comes out".

"Unstable" means "user at your own risk, API may even not really work, and can be deleted without notice".

steveloughran:
Interesting to compare this with Hadoop's annotation, where I have mixed opinions.

A key advantage Apache spark has is that Scala language lets you really scope out things as `private[spark]`; in contrast all Hadoop has is package private, which doesn't scale beyond a single package. By using the private[spark] option you can implicitly lock out external code *provided that code doesn't just use the `org.apache.spark` package name. Which, while it's easy to condemn, is similar to what spark SQL has to do to get at Hive internals under `org.apache.hive`.

Where Hadoop really falls down is that so many classes are essential to day to day use of Private/limitedPrivate tags that its devalued, take `UserGroupInformation` as an example. Want to work in Kerberos: you have to use that, even though you aren't on the "approved" list. I note that @tgravescs owns the [homework there|https://issues.apache.org/jira/browse/HADOOP-10506] to take a sharp knife to that; IMO going to public/unstable is the way to go —at least for any API tagged as limited private YARN/MAPREDUCE.

The big issue with that LimitedPrivate tag is that it viewed by some on the Hadoop dev team as giving a green light to any change, without feeling guilty about breaking things. If I find something downstream breaks against branch-2, I could walk up the colleague and say "you broke my build" and get the response "you shouldn't have used it"....I think everyone needs to do some downstream coding to appreciate the Joy there.

Anyway: LimitedPrivate is essentially worthless, as least as far as I'm concerned. Probably better to have a documented policy about injected code into the `org.apache.spark.*` packages with something like "bahir can ask nicely" and for anyone else "if you have to do this, don't complain when things break". 

For public stuff, Evolving, Stable and perhaps something about retirement. Scala's `deprecated` tag does that & seems pretty well used in Spark. I think Hadoop should go back and look at it's markup there, maybe add something of its own which does text+JIRA+since; or at least have a policy of mandating the javadoc text to go with the method annotation.

PR #15375, opened 2016-10-06
[SPARK-17790][SPARKR] Support for parallelizing R data.frame larger than 2GB

falaki:

If the R data structure that is being parallelized is larger than `INT_MAX` we use files to transfer data to JVM. The serialization protocol mimics Python pickling. This allows us to simply call `PythonRDD.readRDDFromFile` to create the RDD.

I tested this on my MacBook. Following code works with this patch:

``` R
intMax <- .Machine$integer.max
largeVec <- 1:intMax
rdd <- SparkR:::parallelize(sc, largeVec, 2)
```
## How was this patch tested?
- [x] Unit tests


felixcheung:
I think we need to delete the temp file?

falaki:
@felixcheung added clean up for the temp file and unit test. PTAL.

R/pkg/R/context.R
```diff
@@ -123,19 +126,46 @@ parallelize <- function(sc, coll, numSlices = 1) {
   if (numSlices > length(coll))
     numSlices <- length(coll)
 
+  sizeLimit <- as.numeric(
+    sparkR.conf("spark.r.maxAllocationLimit", toString(.Machine$integer.max - 10240)))
+  objectSize <- object.size(coll)
```
davies: Since the guess of size could easily be wrong, and writing them into disk is not that bad anyway, should we have a much smaller default value (for example, 100M)?

falaki: Done!

R/pkg/R/context.R
```diff
@@ -126,13 +126,13 @@ parallelize <- function(sc, coll, numSlices = 1) {
   if (numSlices > length(coll))
     numSlices <- length(coll)
 
-  sizeLimit <- .Machine$integer.max - 10240 # Safe margin bellow maximum allocation limit
+  sizeLimit <- as.numeric(
```
felixcheung: shouldn't this be `as.integer(`?

felixcheung:
Odd, this is the error from appveyor:
```
ontext: Fail to set Spark caller context
java.lang.ClassNotFoundException: org.apache.hadoop.ipc.CallerContext
	at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:331)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	at java.lang.Class.forName0(Native Method)
	at java.lang.Class.forName(Class.java:264)
	at org.apache.spark.util.CallerContext.setCurrentContext(Utils.scala:2485)
	at org.apache.spark.scheduler.Task.run(Task.scala:96)
16/10/07 03:45:58 INFO Executor: Finished task 1.
```

R/pkg/R/context.R
```diff
@@ -123,19 +126,48 @@ parallelize <- function(sc, coll, numSlices = 1) {
...
+  # The PRC backend cannot handle arguments larger than 2GB (INT_MAX)
+  # If serialized data is safely less than that threshold we send it over the PRC channel.
+  # Otherwise, we write it to a file and send the file name
+  if (objectSize < sizeLimit) {
+    jrdd <- callJStatic("org.apache.spark.api.r.RRDD", "createRDDFromArray", sc, serializedSlices)
+  } else {
+    fileName <- writeToTempFile(serializedSlices)
+    jrdd <- callJStatic(
+      "org.apache.spark.api.r.RRDD", "createRDDFromFile", sc, fileName, as.integer(numSlices))
+    file.remove(fileName)
```
felixcheung: if the JVM call throws an exception, I don't think this line will execute, perhaps wrap this in tryCatch?

R/pkg/R/context.R
falaki: This number is not serialized anywhere. I think as.numeric is fine.

felixcheung: agreed, probably not a big deal, an user could set `spark.r.maxAllocationLimit` to `0.01` though.

R/pkg/R/context.R
```diff
@@ -87,6 +87,9 @@ objectFile <- function(sc, path, minPartitions = NULL) {
 #' in the list are split into \code{numSlices} slices and distributed to nodes
 #' in the cluster.
 #'
+#' If size of serialized slices is larger than 2GB (or INT_MAX bytes), the function
```
felixcheung: shouldn't this be 200MB now, same for L141

falaki: Done.

falaki:
@felixcheung does it look OK now?

felixcheung:
This LGTM. Spark unit tests are failing?

felixcheung:
It's interesting AppVeyor is not running for this PR even though there are R changes.

falaki:
Seems like a flaky test in `DirectKafkaStreamSuite`:
```
DirectKafkaStreamSuite:
- pattern based subscription *** FAILED *** (1 minute, 41 seconds)
```

If jenkins listens to your commands, maybe we can have it retest this?

shivaram:
Jenkins, retest this please

tdas:
@falaki @felixcheung  The DirectKafkaStreamSuite is a known flaky test. Nothing in this patch should affect Kafka.

felixcheung:
everything passed, I'm merging to this master and branch-2.0

PR #15376, opened 2016-10-06
[SPARK-17796][SQL] Support wildcard character in filename for LOAD DATA LOCAL INPATH

dongjoon-hyun:

Currently, Spark 2.0 raises an `input path does not exist` AnalysisException if the file name contains '*'. It is misleading since it occurs when there exist some matched files. Also, it was a supported feature in Spark 1.6.2. This PR aims to support wildcard characters in filename for `LOAD DATA LOCAL INPATH` SQL command like Spark 1.6.2.

**Reported Error Scenario**

``` scala
scala> sql("CREATE TABLE t(a string)")

scala> sql("LOAD DATA LOCAL INPATH '/tmp/x*' INTO TABLE t")
org.apache.spark.sql.AnalysisException: LOAD DATA input path does not exist: /tmp/x*;
```
## How was this patch tested?

Pass the Jenkins test with a new test case.

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/command/tables.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/command/tables.scala
@@ -252,7 +252,7 @@ case class LoadDataCommand(
           val fileSystem = FileSystems.getDefault
           val pathPattern = fileSystem.getPath(filePath)
           val dir = pathPattern.getParent.toString
-          val filePattern = pathPattern.getName(pathPattern.getNameCount - 1).toString
+          val filePattern = pathPattern.getFileName.toString
           if (dir.contains("*")) {
             throw new AnalysisException(
               s"LOAD DATA input path allows only filename wildcard: $path")
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/command/tables.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/command/tables.scala
@@ -252,7 +252,6 @@ case class LoadDataCommand(
           val fileSystem = FileSystems.getDefault
           val pathPattern = fileSystem.getPath(filePath)
           val dir = pathPattern.getParent.toString
-          val filePattern = pathPattern.getFileName.toString
           if (dir.contains("*")) {
             throw new AnalysisException(
               s"LOAD DATA input path allows only filename wildcard: $path")
@@ -262,8 +261,8 @@ case class LoadDataCommand(
           if (files == null) {
             false
           } else {
-            val matcher = fileSystem.getPathMatcher("glob:" + filePattern)
-            files.exists(f => matcher.matches(fileSystem.getPath(f.getName)))
+            val matcher = fileSystem.getPathMatcher("glob:" + pathPattern.toAbsolutePath)
+            files.exists(f => matcher.matches(fileSystem.getPath(f.getAbsolutePath)))
           }
         } else {
           new File(filePath).exists()
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/command/tables.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/command/tables.scala
@@ -19,6 +19,7 @@ package org.apache.spark.sql.execution.command
 
 import java.io.File
 import java.net.URI
+import java.nio.file.FileSystems
 import java.util.Date
 
 import scala.collection.mutable.ArrayBuffer
@@ -246,7 +247,27 @@ case class LoadDataCommand(
     val loadPath =
       if (isLocal) {
         val uri = Utils.resolveURI(path)
-        if (!new File(uri.getPath()).exists()) {
+        val filePath = uri.getPath()
+        val exists = if (filePath.contains("*")) {
+          val splitPath = filePath.split(File.separator)
+          val filePattern = splitPath.last
+          val dir = splitPath.dropRight(1).mkString(File.separator)
+          if (dir.contains("*")) {
+            throw new AnalysisException(
+              s"LOAD DATA input path allows only filename wildcard: $path")
+          }
+
+          val files = new File(dir).listFiles()
+          if (files == null) {
+            false
+          } else {
+            val matcher = FileSystems.getDefault.getPathMatcher("glob:" + filePattern)
+            files.exists(f => matcher.matches(FileSystems.getDefault.getPath(f.getName)))
+          }
+        } else {
+          new File(filePath).exists()
+        }
+        if (!exists) {
           throw new AnalysisException(s"LOAD DATA input path does not exist: $path")
         }
         uri
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/command/tables.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/command/tables.scala
@@ -249,9 +249,10 @@ case class LoadDataCommand(
         val uri = Utils.resolveURI(path)
         val filePath = uri.getPath()
         val exists = if (filePath.contains("*")) {
-          val splitPath = filePath.split(File.separator)
-          val filePattern = splitPath.last
-          val dir = splitPath.dropRight(1).mkString(File.separator)
+          val fileSystem = FileSystems.getDefault
+          val pathPattern = fileSystem.getPath(filePath)
+          val dir = pathPattern.getParent.toString
+          val filePattern = pathPattern.getName(pathPattern.getNameCount - 1).toString
           if (dir.contains("*")) {
             throw new AnalysisException(
               s"LOAD DATA input path allows only filename wildcard: $path")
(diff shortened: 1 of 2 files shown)


dongjoon-hyun:
The only test failure is irrelevant. Locally, it passed.
```
[info] - MAP append/extract *** FAILED *** (2 milliseconds)
[info]   java.lang.IllegalArgumentException:
```

dongjoon-hyun:
Retest this please.

srowen:
Hm, so this did used to work? I always thought inpath was a single file or directory. The Hive docs suggest this is the case, and we've tried to follow Hive I guess.
https://cwiki.apache.org/confluence/display/Hive/LanguageManual+DML

dongjoon-hyun:
Oh. Indeed. I thought it's a supported way since it works on Hive 1.2.
```sql
hive> load data local inpath '/data/t/*.txt' INTO TABLE x;
Loading data to table default.x
Table default.x stats: [numFiles=12, totalSize=613224000]
OK
Time taken: 3.712 seconds
```
According to your advice and the URL, it seems not a normal or recommended way.
Thank you for @srowen . You prefer close this PR and the issue, do you?

sql/core/src/main/scala/org/apache/spark/sql/execution/command/tables.scala
```diff
@@ -246,7 +247,27 @@ case class LoadDataCommand(
     val loadPath =
       if (isLocal) {
         val uri = Utils.resolveURI(path)
-        if (!new File(uri.getPath()).exists()) {
+        val filePath = uri.getPath()
+        val exists = if (filePath.contains("*")) {
+          val splitPath = filePath.split(File.separator)
+          val filePattern = splitPath.last
+          val dir = splitPath.dropRight(1).mkString(File.separator)
```
srowen: I usually write `.init` to mean "all but the last element"
Hm, if you're using the local file system here, can you use the JDK `java.nio.file.Path` API to do the parsing of elements?

sql/core/src/main/scala/org/apache/spark/sql/execution/command/tables.scala
```diff
@@ -246,7 +247,27 @@ case class LoadDataCommand(
...
+          if (dir.contains("*")) {
+            throw new AnalysisException(
+              s"LOAD DATA input path allows only filename wildcard: $path")
+          }
+
+          val files = new File(dir).listFiles()
+          if (files == null) {
+            false
+          } else {
+            val matcher = FileSystems.getDefault.getPathMatcher("glob:" + filePattern)
```
srowen: Nit, you might get one reference to `FileSystems.getDefault` and reuse it rather than keep retrieving it.

srowen (commented):
Hm, looks like you're right, empirically. OK, I could see an argument for this. Let me leave some comments here.

dongjoon-hyun:
Thank you for review, @srowen . I'll update the PR.
Also, I'll investigate more if there is a reason not to recommend this way.

sql/core/src/main/scala/org/apache/spark/sql/execution/command/tables.scala
dongjoon-hyun: Now, it uses `java.nio.file.Path` APIs.

dongjoon-hyun:
`KafkaSourceSuite` failure seems to be irrelevant.
```
[info] KafkaSourceSuite:
[info] - cannot stop Kafka stream (1 minute, 1 second)
[info] - subscribing topic by name from latest offsets *** FAILED *** (10 seconds, 511 milliseconds)
[info]   The code passed to eventually never returned normally. Attempted 669 times over 10.012014778000001 seconds. Last failure message: assertion failed: Partition [topic-2, 0] metadata not propagated after timeout. (KafkaTestUtils.scala:312)
[info]   org.scalatest.exceptions.TestFailedDueToTimeoutException:
```

dongjoon-hyun:
Retest this please.

sql/core/src/main/scala/org/apache/spark/sql/execution/command/tables.scala
```diff
@@ -246,7 +247,28 @@ case class LoadDataCommand(
     val loadPath =
       if (isLocal) {
         val uri = Utils.resolveURI(path)
-        if (!new File(uri.getPath()).exists()) {
+        val filePath = uri.getPath()
+        val exists = if (filePath.contains("*")) {
+          val fileSystem = FileSystems.getDefault
+          val pathPattern = fileSystem.getPath(filePath)
+          val dir = pathPattern.getParent.toString
+          val filePattern = pathPattern.getName(pathPattern.getNameCount - 1).toString
```
srowen: I think `getFileName` returns the last element in the path?

sql/core/src/main/scala/org/apache/spark/sql/execution/command/tables.scala
```diff
@@ -246,7 +247,28 @@ case class LoadDataCommand(
...
+          if (dir.contains("*")) {
+            throw new AnalysisException(
+              s"LOAD DATA input path allows only filename wildcard: $path")
+          }
+
+          val files = new File(dir).listFiles()
+          if (files == null) {
+            false
+          } else {
+            val matcher = fileSystem.getPathMatcher("glob:" + filePattern)
```
srowen: I was looking up how this works, and found http://stackoverflow.com/a/14164134/64174 which suggests that this might not work unless the glob starts with "**". However, I wonder if you can just pass this method `"glob:" + pathPattern` in this case anyway to have it match the whole absolute path?

sql/hive/src/test/scala/org/apache/spark/sql/hive/execution/SQLQuerySuite.scala
```diff
@@ -1886,6 +1887,37 @@ class SQLQuerySuite extends QueryTest with SQLTestUtils with TestHiveSingleton {
     }
   }
 
+  test("SPARK-17796 Support wildcard character in filename for LOAD DATA LOCAL INPATH") {
+    withTempDir { dir =>
+      for (i <- 1 to 3) {
+        val writer = new PrintWriter(new File(s"$dir/part-r-0000$i"))
```
srowen: PS I think you can use Files.write from Guava to do this a little more easily

srowen (changes requested):
Yeah, looking pretty good. It's another interesting question about how much to follow Hive's behavior, but given it's a simple change and aligns more with Hive I think it's good.

sql/core/src/main/scala/org/apache/spark/sql/execution/command/tables.scala
dongjoon-hyun: Thanks. I'll use that.

sql/hive/src/test/scala/org/apache/spark/sql/hive/execution/SQLQuerySuite.scala
dongjoon-hyun: Sure, I'll use Guava one here, too.

sql/core/src/main/scala/org/apache/spark/sql/execution/command/tables.scala
dongjoon-hyun: Yes. It matches the whole absolute path.
```scala
scala> val fs = java.nio.file.FileSystems.getDefault
fs: java.nio.file.FileSystem = sun.nio.fs.MacOSXFileSystem@782dc5

scala> fs.getPathMatcher("glob:/x/1.dat").matches(fs.getPath("/x/1.dat"))
res0: Boolean = true

scala> fs.getPathMatcher("glob:/x/*.dat").matches(fs.getPath("/x/1.dat"))
res1: Boolean = true
```

dongjoon-hyun:
Retest this please.

srowen (approved):

dongjoon-hyun:
Thank you for your review and approval, @srowen !

srowen:
Merged to master

dongjoon-hyun:
Thank you, @srowen !

sql/core/src/main/scala/org/apache/spark/sql/execution/command/tables.scala
```diff
@@ -246,7 +247,27 @@ case class LoadDataCommand(
     val loadPath =
       if (isLocal) {
         val uri = Utils.resolveURI(path)
-        if (!new File(uri.getPath()).exists()) {
+        val filePath = uri.getPath()
+        val exists = if (filePath.contains("*")) {
+          val fileSystem = FileSystems.getDefault
+          val pathPattern = fileSystem.getPath(filePath)
+          val dir = pathPattern.getParent.toString
+          if (dir.contains("*")) {
```
cloud-fan: what if the `*` appears in the grandparent dir? e.g. `dir*/subdir/fileName`

srowen: It will result in an `AnalysisException` and that seems like the intended behavior. Only a `*` in the file itself is supported.

PR #14659, opened 2016-08-16
[SPARK-16757] Set up Spark caller context to HDFS and YARN

weiqingy:
## What changes were proposed in this pull request?
1. Pass `jobId` to Task.
2. Invoke Hadoop APIs. 
   - A new function `setCallerContext` is added in `Utils`. `setCallerContext` function invokes APIs of   `org.apache.hadoop.ipc.CallerContext` to set up spark caller contexts, which will be written into `hdfs-audit.log` and Yarn RM audit log.
   - For HDFS: Spark sets up its caller context by invoking`org.apache.hadoop.ipc.CallerContext` in `Task` and Yarn `Client` and `ApplicationMaster`.
   - For Yarn: Spark sets up its caller context by invoking `org.apache.hadoop.ipc.CallerContext` in Yarn `Client`.
## How was this patch tested?

Manual Tests against some Spark applications in Yarn client mode and Yarn cluster mode. Need to check if spark caller contexts are written into HDFS hdfs-audit.log and Yarn RM audit log successfully.

For example, run SparkKmeans in Yarn client mode: 

```
./bin/spark-submit --verbose --executor-cores 3 --num-executors 1 --master yarn --deploy-mode client --class org.apache.spark.examples.SparkKMeans examples/target/original-spark-examples_2.11-2.1.0-SNAPSHOT.jar hdfs://localhost:9000/lr_big.txt 2 5
```

**Before**:
There will be no Spark caller context in records of `hdfs-audit.log` and Yarn RM audit log.

**After**:
Spark caller contexts will be written in records of `hdfs-audit.log` and Yarn RM audit log.

These are records in `hdfs-audit.log`:

```
2016-09-20 11:54:24,116 INFO FSNamesystem.audit: allowed=true   ugi=wyang (auth:SIMPLE) ip=/127.0.0.1   cmd=open    src=/lr_big.txt dst=null    perm=null   proto=rpc   callerContext=SPARK_CLIENT_AppId_application_1474394339641_0005
2016-09-20 11:54:28,164 INFO FSNamesystem.audit: allowed=true   ugi=wyang (auth:SIMPLE) ip=/127.0.0.1   cmd=open    src=/lr_big.txt dst=null    perm=null   proto=rpc   callerContext=SPARK_TASK_AppId_application_1474394339641_0005_JobId_0_StageId_0_AttemptId_0_TaskId_2_AttemptNum_0
[...]

PR #15377, opened 2016-10-06
[SPARK-17802] Improved caller context logging.

lins05:

[SPARK-16757](https://issues.apache.org/jira/browse/SPARK-16757) sets the hadoop `CallerContext` when calling hadoop/hdfs apis to make spark applications more diagnosable in hadoop/hdfs logs. However, the `org.apache.hadoop.ipc.CallerContext` class is only added since [hadoop 2.8](https://issues.apache.org/jira/browse/HDFS-9184), which is not officially releaed yet. So each time `utils.CallerContext.setCurrentContext()` is called (e.g [when a task is created](https://github.com/apache/spark/blob/b678e46/core/src/main/scala/org/apache/spark/scheduler/Task.scala#L95-L96)), a "java.lang.ClassNotFoundException: org.apache.hadoop.ipc.CallerContext"
error is logged, which pollutes the spark logs when there are lots of tasks.

This patch improves this behaviour by only logging the `ClassNotFoundException` once.
## How was this patch tested?

Existing tests.

### DIFF
--- a/core/src/main/scala/org/apache/spark/util/Utils.scala
+++ b/core/src/main/scala/org/apache/spark/util/Utils.scala
@@ -2510,8 +2510,7 @@ private[spark] object Utils extends Logging {
 
 private[util] object CallerContext extends Logging {
   val callerContextSupported: Boolean = {
-    val conf = SparkHadoopUtil.get.conf
-    conf.getBoolean("hadoop.caller.context.enabled", false) && {
+    SparkHadoopUtil.get.conf.getBoolean("hadoop.caller.context.enabled", false) && {
       try {
         // scalastyle:off classforname
         Class.forName("org.apache.hadoop.ipc.CallerContext")
--- a/core/src/main/scala/org/apache/spark/util/Utils.scala
+++ b/core/src/main/scala/org/apache/spark/util/Utils.scala
@@ -2508,6 +2508,27 @@ private[spark] object Utils extends Logging {
   }
 }
 
+private[util] object CallerContext extends Logging {
+  val callerContextSupported: Boolean = {
+    val conf = SparkHadoopUtil.get.conf
+    conf.getBoolean("hadoop.caller.context.enabled", false) && {
+      try {
+        // scalastyle:off classforname
+        Class.forName("org.apache.hadoop.ipc.CallerContext")
+        Class.forName("org.apache.hadoop.ipc.CallerContext$Builder")
+        // scalastyle:on classforname
+        true
+      } catch {
+        case _: ClassNotFoundException =>
+          false
+        case NonFatal(e) =>
+          logWarning("Fail to load the CallerContext class", e)
+          false
+      }
+    }
+  }
+}
+
 /**
  * An utility class used to set up Spark caller contexts to HDFS and Yarn. The `context` will be
  * constructed by parameters passed in.
@@ -2554,21 +2575,21 @@ private[spark] class CallerContext(
    * Set up the caller context [[context]] by invoking Hadoop CallerContext API of
    * [[org.apache.hadoop.ipc.CallerContext]], which was added in hadoop 2.8.
    */
-  def setCurrentContext(): Boolean = {
-    var succeed = false
-    try {
-      // scalastyle:off classforname
-      val callerContext = Class.forName("org.apache.hadoop.ipc.CallerContext")
-      val Builder = Class.forName("org.apache.hadoop.ipc.CallerContext$Builder")
-      // scalastyle:on classforname
-      val builderInst = Builder.getConstructor(classOf[String]).newInstance(context)
-      val hdfsContext = Builder.getMethod("build").invoke(builderInst)
-      callerContext.getMethod("setCurrent", callerContext).invoke(null, hdfsContext)
-      succeed = true
-    } catch {
-      case NonFatal(e) => logInfo("Fail to set Spark caller context", e)
+  def setCurrentContext(): Unit = {
+    if (CallerContext.callerContextSupported) {
+      try {
+        // scalastyle:off classforname
+        val callerContext = Class.forName("org.apache.hadoop.ipc.CallerContext")
+        val builder = Class.forName("org.apache.hadoop.ipc.CallerContext$Builder")
+        // scalastyle:on classforname
+        val builderInst = builder.getConstructor(classOf[String]).newInstance(context)
+        val hdfsContext = builder.getMethod("build").invoke(builderInst)
+        callerContext.getMethod("setCurrent", callerContext).invoke(null, hdfsContext)
+      } catch {
+        case NonFatal(e) =>
+          logWarning("Fail to set Spark caller context", e)
+      }
     }
-    succeed
   }
 }
 
--- a/core/src/test/scala/org/apache/spark/util/UtilsSuite.scala
+++ b/core/src/test/scala/org/apache/spark/util/UtilsSuite.scala
@@ -843,14 +843,11 @@ class UtilsSuite extends SparkFunSuite with ResetSystemProperties with Logging {
 
   test("Set Spark CallerContext") {
     val context = "test"
-    try {
+    new CallerContext(context).setCurrentContext()
+    if (CallerContext.callerContextSupported) {
       val callerContext = Utils.classForName("org.apache.hadoop.ipc.CallerContext")
-      assert(new CallerContext(context).setCurrentContext())
       assert(s"SPARK_$context" ===
         callerContext.getMethod("getCurrent").invoke(null).toString)
-    } catch {
-      case e: ClassNotFoundException =>
-        assert(!new CallerContext(context).setCurrentContext())
     }
   }
 


lins05:
cc @weiqingy (who worked on [SPARK-16757](https://issues.apache.org/jira/browse/SPARK-16757).)

core/src/main/scala/org/apache/spark/util/Utils.scala
```diff
@@ -2474,25 +2474,36 @@ private[spark] class CallerContext(
    val context = "SPARK_" + from + appIdStr + appAttemptIdStr +
      jobIdStr + stageIdStr + stageAttemptIdStr + taskIdStr + taskAttemptNumberStr
 
+   var callerContextSupported : Option[Boolean] = _
```
srowen: Why not just Boolean?

core/src/main/scala/org/apache/spark/util/Utils.scala
```diff
@@ -2474,25 +2474,36 @@ private[spark] class CallerContext(
...
-      val hdfsContext = Builder.getMethod("build").invoke(builderInst)
-      callerContext.getMethod("setCurrent", callerContext).invoke(null, hdfsContext)
-      succeed = true
-    } catch {
-      case NonFatal(e) => logInfo("Fail to set Spark caller context", e)
+    callerContextSupported match {
+      case Some(false) =>
+        false
+      case _ =>
+        var succeed = false
```
srowen: This was also superfluous, just end the method with the right return value

srowen (changes requested):

core/src/main/scala/org/apache/spark/util/Utils.scala
lins05: Makes sense. I have simplied the logic.

core/src/main/scala/org/apache/spark/util/Utils.scala
lins05: done

core/src/main/scala/org/apache/spark/util/Utils.scala
```diff
@@ -2474,25 +2474,35 @@ private[spark] class CallerContext(
...
+        // scalastyle:on classforname
+        val builderInst = Builder.getConstructor(classOf[String]).newInstance(context)
+        val hdfsContext = Builder.getMethod("build").invoke(builderInst)
+        callerContext.getMethod("setCurrent", callerContext).invoke(null, hdfsContext)
+        true
+      } catch {
+        case e : ClassNotFoundException =>
+          logInfo("Fail to set Spark caller context: requires hadoop 2.8 or later.", e)
+          callerContextSupported = false
+          false
```
srowen: Just put `false` at the end of the method like before. Do you really want to log the exception at info level? it looks like a problem when it's not. Nits: Builder -> builder, hadoop -> Hadoop. Don't put a space before the colons, and the new callerContextSupported flag should be private.

srowen (changes requested):

core/src/main/scala/org/apache/spark/util/Utils.scala
lins05: done

core/src/main/scala/org/apache/spark/util/Utils.scala
```diff
@@ -2474,25 +2474,34 @@ private[spark] class CallerContext(
...
+        val callerContext = Class.forName("org.apache.hadoop.ipc.CallerContext")
+        val builder = Class.forName("org.apache.hadoop.ipc.CallerContext$Builder")
+        // scalastyle:on classforname
+        val builderInst = builder.getConstructor(classOf[String]).newInstance(context)
+        val hdfsContext = builder.getMethod("build").invoke(builderInst)
+        callerContext.getMethod("setCurrent", callerContext).invoke(null, hdfsContext)
+        true
+      } catch {
+        case e: ClassNotFoundException =>
+          logInfo("Fail to set Spark caller context: requires Hadoop 2.8 or later.", e)
```
srowen: I still think you perhaps should not log the exception but just log the message (so that the caller sees what class was not found, like `logInfo(s"... or later: ${e.getMessage}")`

srowen (changes requested):
Also, in the case of another non-fatal error, would you not also consider the context unsupported? it seems more like a warning than error because you can complete without it.

core/src/main/scala/org/apache/spark/util/Utils.scala
lins05: done

core/src/main/scala/org/apache/spark/util/Utils.scala
```diff
@@ -2474,25 +2474,36 @@ private[spark] class CallerContext(
...
+        callerContext.getMethod("setCurrent", callerContext).invoke(null, hdfsContext)
+        true
+      } catch {
+        case e: ClassNotFoundException =>
+          logInfo(
+            s"Fail to set Spark caller context: requires Hadoop 2.8 or later: ${e.getMessage}")
+          callerContextSupported = false
+          false
+        case NonFatal(e) =>
+          logWarning("Fail to set Spark caller context", e)
```
lins05: Emm, I prefer not to suppress the non-ClassNotFound errors, because they are real errors, compared to the ClassNotFound error, which is rather a *conditional feature* based on the hadoop environment.

srowen (approved):
OK, it's a corner case anyway.

weiqingy:
Thanks for proposing this PR. LGTM.

core/src/main/scala/org/apache/spark/util/Utils.scala
```diff
@@ -2432,6 +2432,10 @@ private[spark] object Utils extends Logging {
   }
 }
 
+private[spark] object CallerContext {
+  var callerContextSupported: Boolean = true
```
lins05: @srowen I updated the code to make `callerContextSupported` a static variable because the usage pattern of `CallerContext` is `new CallerContext(...).setCurrentContext()`

srowen: OK, could it even be `private[util]`?

core/src/main/scala/org/apache/spark/util/Utils.scala
lins05: Thanks for the tip about `hadoop.caller.context.enabled` and the suggestion, i'll do that.

core/src/main/scala/org/apache/spark/util/Utils.scala
```diff
@@ -2432,6 +2432,10 @@ private[spark] object Utils extends Logging {
   }
 }
 
+private[util] object CallerContext {
+  var callerContextSupported: Boolean = true
```
jerryshao: What is the usage of this flag? I don't see any other place use it, all just setters.

core/src/main/scala/org/apache/spark/util/Utils.scala
```diff
@@ -2474,25 +2478,42 @@ private[spark] class CallerContext(
...
-      val callerContext = Class.forName("org.apache.hadoop.ipc.CallerContext")
-      val Builder = Class.forName("org.apache.hadoop.ipc.CallerContext$Builder")
-      // scalastyle:on classforname
-      val builderInst = Builder.getConstructor(classOf[String]).newInstance(context)
-      val hdfsContext = Builder.getMethod("build").invoke(builderInst)
-      callerContext.getMethod("setCurrent", callerContext).invoke(null, hdfsContext)
-      succeed = true
-    } catch {
-      case NonFatal(e) => logInfo("Fail to set Spark caller context", e)
+    if (!CallerContext.callerContextSupported) {
```
jerryshao: IIUC, code will never trap into this branch, since you intialize it as `true`.

core/src/main/scala/org/apache/spark/util/Utils.scala
```diff
@@ -2474,25 +2478,42 @@ private[spark] class CallerContext(
    val context = "SPARK_" + from + appIdStr + appAttemptIdStr +
      jobIdStr + stageIdStr + stageAttemptIdStr + taskIdStr + taskAttemptNumberStr
 
+  lazy val conf = new Configuration
```
jerryshao: Please use `SparkHadoopUtils#conf`.

core/src/main/scala/org/apache/spark/util/Utils.scala
```diff
@@ -2474,25 +2478,42 @@ private[spark] class CallerContext(
...
+      false
+    } else {
+      if (!conf.getBoolean("hadoop.caller.context.enabled", false)) {
+        logInfo("Hadoop caller context is not enabled")
+        CallerContext.callerContextSupported = false
+        false
+      } else {
+        try {
+            // scalastyle:off classforname
+            val callerContext = Class.forName("org.apache.hadoop.ipc.CallerContext")
```
jerryshao: Please use `Utils#classForName`.

core/src/main/scala/org/apache/spark/util/Utils.scala
```diff
@@ -2474,25 +2478,42 @@ private[spark] class CallerContext(
...
+    if (!CallerContext.callerContextSupported) {
+      false
+    } else {
+      if (!conf.getBoolean("hadoop.caller.context.enabled", false)) {
+        logInfo("Hadoop caller context is not enabled")
+        CallerContext.callerContextSupported = false
+        false
+      } else {
+        try {
+            // scalastyle:off classforname
```
jerryshao: Nit: indent is not correct, use 2 ws.

core/src/main/scala/org/apache/spark/util/Utils.scala
lins05: Nope, when there is a ClassNotFound exception, it would be set to false.

core/src/main/scala/org/apache/spark/util/Utils.scala
lins05: done

core/src/main/scala/org/apache/spark/util/Utils.scala
lins05: done

core/src/main/scala/org/apache/spark/util/Utils.scala
lins05: done

core/src/main/scala/org/apache/spark/util/Utils.scala
jerryshao: The usage of `CallerContext.callerContextSupported` is quite weird to me, looks like you're implementing a singleton factory like patten to avoid re-calling this method if class not found.

Will this introduce threading issue for  the state of `CallerContext.callerContextSupported`?

Also do we have side affect to call `setCurrentContext` multiple times in one JVM? 

I would be in favor of avoiding such usage of global flag.

core/src/main/scala/org/apache/spark/util/Utils.scala
```diff
@@ -2479,20 +2483,35 @@ private[spark] class CallerContext(
...
+    if (!CallerContext.callerContextSupported) {
+      false
+    } else {
+      if (!SparkHadoopUtil.get.conf.getBoolean("hadoop.caller.context.enabled", false)) {
+        logInfo("Hadoop caller context is not enabled")
+        CallerContext.callerContextSupported = false
+        false
+      } else {
+        try {
+          // scalastyle:off classforname
```
jerryshao: Remove this comment.

jerryshao:
Another thing, do you verify it locally? Since there's no unit test to cover it.

lins05:
> Another thing, do you verify it locally? Since there's no unit test to cover it.
@jerryshao Yeah, I did test it locally to ensure the error is only logged once.

core/src/main/scala/org/apache/spark/util/Utils.scala
lins05: Good catch, done.

weiqingy:
Utils#classForName will make maven build fail, please do not use it before figure out : https://issues.apache.org/jira/browse/SPARK-17714. The failure is here: https://issues.apache.org/jira/browse/SPARK-17710 and .

srowen:
@weiqingy all the Maven/SBT builds pass now, are you sure this is still an issue? I know the PR builder is just using SBT, yes, but thought we had fixed this.

weiqingy:
Before 28/Sep/16, using Utils#classForName could pass the PR builder, but would fail the Spark QA tests in master Maven builds. Unit now (from 28/Sep/16 - now), if there is no change in maven build of Spark QA test, I think using Utils#classForName here will still fail the Spark QA tests in master Maven builds. 

There is the related discussion about this (the last part of conversations): https://github.com/apache/spark/pull/14659#issuecomment-248402339)

core/src/main/scala/org/apache/spark/util/Utils.scala
srowen: Yes, but re-executing the code an extra doesn't do anything bad except log an additional message. We're trying to prevent it from executing a bunch of times. The flag can only go from true to false. It keeps executing the same deterministic path until something stops it. Unless you mean this really never executes more than once anyway, I don't think there's a problem right here.

jerryshao: If you cannot fully preventing re-executing this code, and to the worst all threads will executing the same logics again, so is it necessary enough to add such flag? For me I think it is some kind of undeterministic that will confuse the user (some tasks printed the log while others not).

srowen: The point is to avoid a bunch of exception stack traces in the log over and over. I don't think thread-safety is an issue here. If it prints twice instead of once, no big deal. Otherwise this error/warning prints on every task execution. That's not great.

lins05:
@weiqingy @srowen I see. So do you suggest to avoid using `Utils.classForName` to get this one merged, or rather wait for SPARK-17714?

srowen:
I think you can avoid it for now, and if that other issue is addressed, it can go back and use the `Utils` method again everywhere.

lins05:
@srowen done.

srowen:
@jerryshao can I take your temperature about how against this change you are? maybe @lins05 can elaborate again on what this is preventing, and what happens in case of a race between two threads. That is, what do we expect in the average case before and after?

jerryshao:
@srowen I'm not against this change, personally because the usage of flag is wired to me and frankly saying I haven't seen such pattern in the Spark code.

Since we want to avoid re-executing the code if CallContext is not enabled or not existed. We could get this information immutably like:

```scala
private[util] object CallerContext {
  val callerContextSupported: Boolean = if (callContext is not enabled) {
     false
  } else if (callContext is not existed) {
    false 
  } else {
    true
  }
}

```

And in `setCurrentContext` we could get rid of this mutable things, use the flag to decide whether to set the caller context without reset flag. That will simplify the codes and make others easy to understand.

What do you think?

srowen:
Yeah, you're saying all the checks can be done upfront? I agree. It will need a tiny bit of code duplication but is more robust. I think we could also potentially get rid of the return value from `setCallerContext`?

jerryshao:
Yup, that's what I mean.

lins05:
Thanks @jerryshao @srowen . I've updated the code like what you suggested.

weiqingy:
The test case wanted to test “if it’s running on a hadoop version that supports caller context, our code does set it”. But in this change , if somehow the flag was set incorrectly (being false instead of true), it would’t be able to find out.

lins05:
@weiqingy I agree that's a problem. But i don't see how to unit test the `callerContextSupported` method without repeating the same logic in the test code. Do you have any suggestion?

weiqingy:
@lins05 Maybe something like this: 

```
 test("Set Spark CallerContext") {
     val context = "test"
     try {
         new CallerContext(context).setCurrentContext()
         val callerContext = Utils.classForName("org.apache.hadoop.ipc.CallerContext")
         assert(CallerContext.callerContextSupported)
         assert(s"SPARK_$context" ===
             callerContext.getMethod("getCurrent").invoke(null).toString)
        } catch {
          case e: ClassNotFoundException =>
             assert(!CallerContext.callerContextSupported)
      }
    }
```

core/src/main/scala/org/apache/spark/util/Utils.scala
```diff
@@ -2432,6 +2432,26 @@ private[spark] object Utils extends Logging {
   }
 }
 
+private[util] object CallerContext {
+  private def callerContextClassExists(): Boolean = {
+    try {
+      // scalastyle:off classforname
+      val callerContext = Class.forName("org.apache.hadoop.ipc.CallerContext")
+      val builder = Class.forName("org.apache.hadoop.ipc.CallerContext$Builder")
```
jerryshao: These two values are not used at all, I think you could remove them.

jerryshao:
LGTM just some minor things.

lins05:
@weiqingy Emm, then we would also add the logic of checking "hadoop.caller.context.enabled" in the test code, which makes the test code simply duplicates the code path of `CallerContext.callerContextSupported`, and IMHO this doesn't make much sense...

On the other hand I do agree we should test the code path, but it seems not easy to do it.

@srowen @jerryshao What do you think?

core/src/main/scala/org/apache/spark/util/Utils.scala
```diff
@@ -2432,6 +2432,26 @@ private[spark] object Utils extends Logging {
...
+    } catch {
+      case _ : ClassNotFoundException => false
+    }
+  }
+
+  private def callerContextEnabled(): Boolean = {
+    SparkHadoopUtil.get.conf.getBoolean("hadoop.caller.context.enabled", false)
+  }
+
+  lazy val callerContextSupported: Boolean = callerContextEnabled && callerContextClassExists
```
srowen: I don't think this needs to be lazy (?) and doesn't need to be broken out into methods. It can just be one `val` initialized directly upfront in one expression.

weiqingy:
Not sure why you need to check hadoop.caller.context.enabled" since just need to make callerContextClassExists be access by test case.

srowen (commented):
I see the issue with the tests, yeah, and it's a fair point. My personal opinion is that the test here is sufficient because one would have to reimplement the code in question to test it, and, we'll soon drop support for versions that don't support this anyway, and, the failure that this wouldn't catch is not bad

core/src/main/scala/org/apache/spark/util/Utils.scala
```diff
@@ -2432,6 +2432,18 @@ private[spark] object Utils extends Logging {
   }
 }
 
+private[util] object CallerContext {
+  val callerContextSupported: Boolean = {
+    SparkHadoopUtil
```
srowen: Let's find a different way to wrap this; it's non-standard and hard to read the condition here. Maybe break out `conf` as a local `val` to make it simpler.

I see why you're using Try, though I wonder if we are suppressing potentially important errors this way. For example, ClassNotFoundException obviously means 'not supported' and that's normal. A fatal error should probably just propagate. But a NonFatal error could be logged with a warning? because that would be unexpected. It would still mean the caller context isn't supported rather than cause a failure.

srowen (changes requested):
The rest seems OK to me.

core/src/main/scala/org/apache/spark/util/Utils.scala
```diff
@@ -2432,6 +2432,30 @@ private[spark] object Utils extends Logging {
...
+    } catch {
+      case _: ClassNotFoundException =>
+        false
+      case NonFatal(e) =>
+        logWarning("Fail to load the CallerContext class", e)
+        false
+    }
+  }
+
+  private val conf = SparkHadoopUtil.get.conf
```
srowen: This doesn't need to be a separate member; it can be a local val.
Still, why have more than one member here? Initializing `callerContextSupported` in one block of code is sufficient and simpler.

srowen (changes requested):
The rest is looking good. Let's keep this simple.

core/src/main/scala/org/apache/spark/util/Utils.scala
```diff
@@ -2508,6 +2508,27 @@ private[spark] object Utils extends Logging {
   }
 }
 
+private[util] object CallerContext extends Logging {
+  val callerContextSupported: Boolean = {
+    val conf = SparkHadoopUtil.get.conf
+    conf.getBoolean("hadoop.caller.context.enabled", false) && {
```
srowen: Although this construct looks funny to me, a more conventional structure would just be a little longer. I'm OK with it.

srowen (approved):

core/src/main/scala/org/apache/spark/util/Utils.scala
lins05: Right, I removed the local conf variable.

lins05:
@srowen Could we get this merged since the tests are now green? Not sure why it failed previously, it just turned green without me doing anything.

srowen:
Merged to master

PR #13868, opened 2016-06-23
[SPARK-15899] [SQL] Fix the construction of the file path with hadoop Path

avulanov:
## What changes were proposed in this pull request?

Fix the construction of the file path. Previous way of construction caused the creation of incorrect path on Windows.
## How was this patch tested?

Run SQL unit tests on Windows

PR #15382, opened 2016-10-06
[SPARK-17810] [SQL] Default spark.sql.warehouse.dir is relative to local FS but can resolve as HDFS path

srowen:

Always resolve spark.sql.warehouse.dir as a local path, and as relative to working dir not home dir
## How was this patch tested?

Existing tests.

### DIFF
--- a/docs/sql-programming-guide.md
+++ b/docs/sql-programming-guide.md
@@ -904,50 +904,27 @@ access data stored in Hive.
 Configuration of Hive is done by placing your `hive-site.xml`, `core-site.xml` (for security configuration),
 and `hdfs-site.xml` (for HDFS configuration) file in `conf/`.
 
-<div class="codetabs">
-
-<div data-lang="scala"  markdown="1">
-
 When working with Hive, one must instantiate `SparkSession` with Hive support, including
 connectivity to a persistent Hive metastore, support for Hive serdes, and Hive user-defined functions.
 Users who do not have an existing Hive deployment can still enable Hive support. When not configured
 by the `hive-site.xml`, the context automatically creates `metastore_db` in the current directory and
 creates a directory configured by `spark.sql.warehouse.dir`, which defaults to the directory
-`spark-warehouse` in the current directory that the spark application is started. Note that
+`spark-warehouse` in the current directory that the Spark application is started. Note that
 the `hive.metastore.warehouse.dir` property in `hive-site.xml` is deprecated since Spark 2.0.0.
 Instead, use `spark.sql.warehouse.dir` to specify the default location of database in warehouse.
-You may need to grant write privilege to the user who starts the spark application.
+You may need to grant write privilege to the user who starts the Spark application.
 
+<div class="codetabs">
+
+<div data-lang="scala"  markdown="1">
 {% include_example spark_hive scala/org/apache/spark/examples/sql/hive/SparkHiveExample.scala %}
 </div>
 
 <div data-lang="java"  markdown="1">
-
-When working with Hive, one must instantiate `SparkSession` with Hive support, including
-connectivity to a persistent Hive metastore, support for Hive serdes, and Hive user-defined functions.
-Users who do not have an existing Hive deployment can still enable Hive support. When not configured
-by the `hive-site.xml`, the context automatically creates `metastore_db` in the current directory and
-creates a directory configured by `spark.sql.warehouse.dir`, which defaults to the directory
-`spark-warehouse` in the current directory that the spark application is started. Note that
-the `hive.metastore.warehouse.dir` property in `hive-site.xml` is deprecated since Spark 2.0.0.
-Instead, use `spark.sql.warehouse.dir` to specify the default location of database in warehouse.
-You may need to grant write privilege to the user who starts the spark application.
-
 {% include_example spark_hive java/org/apache/spark/examples/sql/hive/JavaSparkHiveExample.java %}
 </div>
 
 <div data-lang="python"  markdown="1">
-
-When working with Hive, one must instantiate `SparkSession` with Hive support, including
-connectivity to a persistent Hive metastore, support for Hive serdes, and Hive user-defined functions.
-Users who do not have an existing Hive deployment can still enable Hive support. When not configured
-by the `hive-site.xml`, the context automatically creates `metastore_db` in the current directory and
-creates a directory configured by `spark.sql.warehouse.dir`, which defaults to the directory
-`spark-warehouse` in the current directory that the spark application is started. Note that
-the `hive.metastore.warehouse.dir` property in `hive-site.xml` is deprecated since Spark 2.0.0.
-Instead, use `spark.sql.warehouse.dir` to specify the default location of database in warehouse.
-You may need to grant write privilege to the user who starts the spark application.
-
 {% include_example spark_hive python/sql/hive.py %}
 </div>
 
--- a/examples/src/main/java/org/apache/spark/examples/sql/hive/JavaSparkHiveExample.java
+++ b/examples/src/main/java/org/apache/spark/examples/sql/hive/JavaSparkHiveExample.java
@@ -56,7 +56,7 @@ public class JavaSparkHiveExample {
   public static void main(String[] args) {
     // $example on:spark_hive$
     // warehouseLocation points to the default location for managed databases and tables
-    String warehouseLocation = "file:" + System.getProperty("user.dir") + "spark-warehouse";
+    String warehouseLocation = "spark-warehouse";
     SparkSession spark = SparkSession
       .builder()
       .appName("Java Spark Hive Example")
--- a/sql/core/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala
@@ -23,6 +23,7 @@ import java.util.concurrent.TimeUnit
 import scala.collection.JavaConverters._
 import scala.collection.immutable
 
+import org.apache.hadoop.fs.Path
 import org.apache.parquet.hadoop.ParquetOutputCommitter
 
 import org.apache.spark.internal.Logging
@@ -56,7 +57,7 @@ object SQLConf {
   val WAREHOUSE_PATH = SQLConfigBuilder("spark.sql.warehouse.dir")
     .doc("The default location for managed databases and tables.")
     .stringConf
-    .createWithDefault("spark-warehouse")
+    .createWithDefault(Utils.resolveURI("spark-warehouse").toString)
 
   val OPTIMIZER_MAX_ITERATIONS = SQLConfigBuilder("spark.sql.optimizer.maxIterations")
     .internal()
@@ -755,7 +756,7 @@ private[sql] class SQLConf extends Serializable with CatalystConf with Logging {
 
   def variableSubstituteDepth: Int = getConf(VARIABLE_SUBSTITUTE_DEPTH)
 
-  def warehousePath: String = Utils.resolveURI(getConf(WAREHOUSE_PATH)).toString
+  def warehousePath: String = new Path(getConf(WAREHOUSE_PATH)).toString
 
   def ignoreCorruptFiles: Boolean = getConf(IGNORE_CORRUPT_FILES)
 
--- a/sql/core/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala
@@ -23,13 +23,13 @@ import java.util.concurrent.TimeUnit
 import scala.collection.JavaConverters._
 import scala.collection.immutable
 
-import org.apache.hadoop.fs.Path
 import org.apache.parquet.hadoop.ParquetOutputCommitter
 
 import org.apache.spark.internal.Logging
 import org.apache.spark.internal.config._
 import org.apache.spark.network.util.ByteUnit
 import org.apache.spark.sql.catalyst.CatalystConf
+import org.apache.spark.util.Utils
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 // This file defines the configuration options for Spark SQL.
@@ -56,7 +56,7 @@ object SQLConf {
   val WAREHOUSE_PATH = SQLConfigBuilder("spark.sql.warehouse.dir")
     .doc("The default location for managed databases and tables.")
     .stringConf
-    .createWithDefault("${system:user.dir}/spark-warehouse")
+    .createWithDefault("spark-warehouse")
 
   val OPTIMIZER_MAX_ITERATIONS = SQLConfigBuilder("spark.sql.optimizer.maxIterations")
     .internal()
@@ -755,7 +755,7 @@ private[sql] class SQLConf extends Serializable with CatalystConf with Logging {
 
   def variableSubstituteDepth: Int = getConf(VARIABLE_SUBSTITUTE_DEPTH)
 
-  def warehousePath: String = new Path(getConf(WAREHOUSE_PATH)).toString
+  def warehousePath: String = Utils.resolveURI(getConf(WAREHOUSE_PATH)).toString
 
   def ignoreCorruptFiles: Boolean = getConf(IGNORE_CORRUPT_FILES)
 
--- a/sql/core/src/test/scala/org/apache/spark/sql/internal/SQLConfSuite.scala
+++ b/sql/core/src/test/scala/org/apache/spark/sql/internal/SQLConfSuite.scala
@@ -17,7 +17,7 @@
 
 package org.apache.spark.sql.internal
 
-import org.apache.hadoop.fs.Path
+import java.io.File
 
 import org.apache.spark.SparkContext
 import org.apache.spark.sql._
@@ -219,8 +219,7 @@ class SQLConfSuite extends QueryTest with SharedSQLContext {
     try {
       // to get the default value, always unset it
       spark.conf.unset(SQLConf.WAREHOUSE_PATH.key)
-      assert(spark.sessionState.conf.warehousePath
-        === new Path(s"${System.getProperty("user.dir")}/spark-warehouse").toString)
+      assert(new File("spark-warehouse").toURI.toString === spark.sessionState.conf.warehousePath)
     } finally {
       sql(s"set ${SQLConf.WAREHOUSE_PATH}=$original")
     }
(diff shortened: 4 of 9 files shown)


srowen:
CC @vanzin -- I think your solution at https://github.com/apache/spark/pull/13868#discussion_r82252926 is better, will probably switch to it.

CC @avulanov and @koertkuipers

docs/sql-programming-guide.md
```diff
@@ -869,50 +869,27 @@ access data stored in Hive.
...
+You may need to grant write privilege to the user who starts the Spark application.
 
+<div class="codetabs">
+
+<div data-lang="scala"  markdown="1">
 {% include_example spark_hive scala/org/apache/spark/examples/sql/hive/SparkHiveExample.scala %}
 </div>
 
 <div data-lang="java"  markdown="1">
-
```
srowen: Here I am just fixing some duplication of docs that relate to this change. There were 3 identical stanzas

koertkuipers:
i think working dir makes more sense than home dir. but could this catch people by surprise because we now expect write permission in the working dir?

srowen:
Good point. The docs all say it should be working dir, so I was following the apparent intent along the way here. I don't know if one is more surprising than the other.

srowen:
Update on my thinking:

Docs say the default should be "spark-warehouse" in the local working directory. The original code did establish a default on the local file system... though in the home dir, not working dir. So I still favor fixing that, but that's a small, orthogonal question.

Although the default appears to be intended to be on the local FS, I don't think it must be. @avulanov mentions it might be on S3; the Hive warehouse dir is on HDFS too by default, in comparison. I don't have evidence that this has been prohibited so I guess we shouldn't assume this is a local path.

It seems simple to default to "spark-warehouse" in the local working dir of whatever file system is being used, be it local or HDFS. That means SQLConf shouldn't specify a default with any URI scheme or path. SQLConf should then resolve it, though, at the moment there's no good way to plumb through a Hadoop Configuration. (I just made a new default one in here for the moment.)

Then we're back to an interesting detail, that Hadoop Path's toURI won't add a trailing slash in the case of a local file. Maybe it doesn't in general since it can't always tell whether a path is a directory or not. Local URIs however will have a trailing slash. It's probably best to leave the behavior and adjust the tests to not expect a trailing slash, as the original code seemed to (?)

Does this still work on Windows? As far as I can tell it would.

Is it OK to instantiate `FileSystem.get(new Configuration())` in `SQLConf`? there's no easy access to the actual conf, but, it is only trying to read the default file system which shouldn't depend on Spark-specific Hadoop conf.

I think this change actually _restores_ the ability to specify a warehouse path with a URI scheme. This second, it will be interpreted as a file path only if you write "file:/foo". However there's no way to write a relative file URI, so no way to guarantee that a default of "spark-warehouse" is a relative path on the local FS. After this change, HDFS users would see that this correctly defaults to "spark-warehouse" in an HDFS working directory.

examples/src/main/java/org/apache/spark/examples/sql/hive/JavaSparkHiveExample.java
```diff
@@ -56,7 +56,7 @@ public void setValue(String value) {
   public static void main(String[] args) {
     // $example on:spark_hive$
     // warehouseLocation points to the default location for managed databases and tables
-    String warehouseLocation = "file:" + System.getProperty("user.dir") + "spark-warehouse";
+    String warehouseLocation = "spark-warehouse";
```
srowen: Here I've just updated the example's location to match the new default.

koertkuipers:
i don't think there is such a thing as a HDFS working directory, but that probably means it just uses the home dir on hdfs (/user/<user.name>) for any relative paths

sql/core/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala
```diff
@@ -757,7 +758,10 @@ private[sql] class SQLConf extends Serializable with CatalystConf with Logging {
 
   def variableSubstituteDepth: Int = getConf(VARIABLE_SUBSTITUTE_DEPTH)
 
-  def warehousePath: String = new Path(getConf(WAREHOUSE_PATH)).toString
+  def warehousePath: String = {
+    val path = new Path(getConf(WAREHOUSE_PATH))
+    FileSystem.get(path.toUri, new Configuration()).makeQualified(path).toString
```
avulanov: `Utils.resolveURI` creates an `java.net.URI` from a String path, and returns the URI if it has schema. Apparently, it treats Windows paths as having a schema and breaks down with back slashes. `Utils.resolveURI` returns a different results comparing to hadoop `Path.toUri`:
````
scala> (new URI("/my")).getScheme()
res22: String = null

scala> (new URI("c:/my")).getScheme()
res23: String = c

scala> (new URI("c:\\my")).getScheme()
java.net.URISyntaxException: Illegal character
````
hadoop Path adds a slash to the path:
````
scala> new Path("c:/my").toUri
res25: java.net.URI = /c:/my
````

Comment (srowen) [inline sql/core/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala]: @avulanov just so I'm testing the right things, what would a user set for the warehouse dir, on Windows, if they wanted to specify a local path? It would have to be like `file:///C:/...` right? That seems to be correctly resolved. That is, we can't expect "C:/..." or "C:\..." to work, because we need this to be parsed as a URI or relative path -- is that right? If so then I suspect @vanzin is right, let me try that approach.

**Comment by srowen** [comment]
OK, now using `Utils.resolveURI`. This means all inputs without a scheme resolve as `file:` URIs, including the default "spark-warehouse". This will continue to resolve to the local working dir by default, as per docs.

Inputs with a scheme can be specified, still, to point to HDFS or S3.

I adjusted the tests to, I think, further standardize the handling of URIs.

File: sql/core/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala (hunk 3 above)
Comment (avulanov) [inline sql/core/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala]: There are two concerns here. Linux paths will work without adding `file://`, and Windows will not. That might lead to confusion. Also, I doubt that both Windows and Linux users typically add `file://` if they specify a local path. As a rule of thumb, if no schema is given, then by default it is a local path. The second concern is that `URI` requires percent encoded string, so it does not work with white spaces or special symbols in the path. As far as I understand, URI is for URIs and not for paths. If we want to use URI then we need to change the variable name to `warehouseURI` and add a proper documentation about this configuration parameter.

Comment (vanzin) [inline sql/core/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala]: Agree, if the Utils method doesn't work well with Windows paths it should be avoided... I seem to remember at some point it used to handle those (using the `windowsDrive` method in Utils) but that seems to be gone.

Comment (srowen) [inline sql/core/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala]: I should say I'm assuming the input must be a URI, because we have to allow users to disambiguate by specifying a scheme. And if we do that, then Windows paths can't be allowed. Of course, Windows file URIs should work, like `file:///C:/...` So yes I am certainly assuming this must be a URI, not a path, because I think we have to? If so, sure, documenting it sounds good.

Comment (avulanov) [inline sql/core/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala]: URI standard might be too restrictive for something that a user has to manually specify as a string. According to [RFC 3986](https://tools.ietf.org/html/rfc3986#section-2.3), it allows only numbers, digits, dash, underscore, dot and tilde. Anything else is required to be [percent-encoded](https://tools.ietf.org/html/rfc3986#section-2.1). This means that we have to require user to always do percent-encoding of her path if we stick with URI. There are other paths in Spark that user can specify and they are not required to be encoded. This might add unnecessary confusion. I would suggest to keep it as a path, document that one can optionally add a schema if the path is not local, and try to handle different options within Spark.

Comment (srowen) [inline sql/core/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala]: Hm, yes I can see how that would be tricky, mostly because of spaces in a path.
OK, how about this: if it parses as a URI with `java.net.URI`, parse it as a URI. If not, try to parse it as a URI with `java.io.File`. Does that present any problems I'm missing?

I think we have to support URIs because they always have been (?), and that also enables non-file URIs, which is implicitly supported right now. Although I'm reluctant to expand the possibilities here I tend to be persuaded that it's worth trying here specially.

Comment (srowen) [inline sql/core/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala]: There I'm looking at, for example, https://github.com/avulanov/spark/blob/ea24b59fe83c37dbab27579141b5c63cccee138d/sql/core/src/test/scala/org/apache/spark/sql/execution/command/DDLSuite.scala#L141 in the test code.  In non-test code I think it's the same source you copied, in SessionCatalog, line 154. 

Although these code locations can also deal with a scheme vs no scheme, it seemed to be easier to deal with it upfront, where it's returned to the rest of the code from the conf object. I think it'll be the same code, same complexity either place.

The fact that `resolveURI` doesn't quite do what we want here suggests, I suppose, that lots of things in Spark aren't going to play well with a Windows path with spaces. See:

```
scala> resolveURI("file:///C:/My Programs/path")
res14: java.net.URI = file:/Users/srowen/file:/C:/My%20Programs/path

scala> resolveURI("/C:/My Programs/path")
res15: java.net.URI = file:/C:/My%20Programs/path

scala> resolveURI("C:/My Programs/path")
res16: java.net.URI = file:/Users/srowen/C:/My%20Programs/path

scala> resolveURI("/My Programs/path")
res17: java.net.URI = file:/My%20Programs/path
```

The second (possibly alternative absolute Windows path with space) and fourth examples (Linux path with space) happen to come out right. If we're willing to also accept here that it's just what's going to work and not work, then, at least there is a working syntax for all scenarios when using this method.

I believe there's an argument for further changing resolveURI to work with more variants here, but I'd have to even figure out first what is supposed to work and not work!

Comment (srowen) [inline sql/core/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala]: Yeah I agree that's why this behaves this way, and it's how similar paths/URIs would be treated in a lot of places in Spark. Therefore I'm wondering if that's the right thing to do, locally, to accept the same behavior and quirks in parsing this path (and then consider changing that globally later if needed). If I'm right then there is at least a syntax available for hdfs: and file: URIs, and one that works for Linux and Windows, including paths with spaces. Not all possible strings work as expected, maybe, but the more common ones seem to.

The net change at this point is just to use `Utils.resolveURI` to resolve this thing like other places in Spark, and then default to "spark-warehouse" which will now go back to defaulting to a local working dir path as intended. The rest is just test/doc/example updates.

What's your opinion on this current state, compared to the problem being solved here?

Comment (avulanov) [inline sql/core/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala]: `resolveURI` function does not process correctly some of the inputs (even some Linux paths). It contains a bug and should be fixed. The name itself is a bit misleading, because it processes `String` rather than `URI`, and is capable of resolving Paths. I propose to rename it to `resolvePath`, and rewrite it according to its documentation (below). Then we can use it to fix the initial issue. Does it sound reasonable to you?
```
  /**
   * Return a well-formed URI for the file described by a user input string.
   *
   * If the supplied path does not contain a scheme, or is a relative path, it will be
   * converted into an absolute path with a file:// scheme.
   */
```

Comment (srowen) [inline sql/core/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala]: Do you mean just updating the documentation or changing the behavior? Updating the docs are fine. Updating the behavior (and name I guess) would affect a bunch of other places in the code, and that seems like a bigger issue to tackle separately.

Comment (vanzin) [inline sql/core/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala]: Can we split the difference here and treat those two as different issues?

- for this bug, let's hardcode `.transform(new File(_).toURI().toString())` into the config constant definition; that's on par with the previous version of this code modified in 11a6844beb, which hardcoded the `file:` scheme into the default value.

- separately, if it's a real issue, let's make changes to make this support non-local directories.

That way we fix the immediate issue and can have more discussions about how to properly solve the other and maybe even fix `resolveURI`.

Comment (srowen) [inline sql/core/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala]: I could do that. I could make a different PR with a subset/variant of this PR that would assume the path must be local and then rebase this one on top.

The thing is, I'm not clear it was ever actually meant to only be a local path, even though its default is a local path. It was always parsed as a Hadoop Path. Right now in the 2.0.1 release it's definitely going to support non-local paths. And this is supposed to be some analog of the Hive warehouse path, right? which can be on HDFS.

Therefore I kind of convinced myself that forcing this to a local file isn't going to be a valid intermediate state, and so we shouldn't stop there. Then we're just back to the change here.

How strongly do you feel about it, or rather, what's the problem with just allowing it to be non-local as is implemented here?

File: sql/core/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala [hunk 4]
```diff
@@ -741,7 +741,7 @@ private[sql] class SQLConf extends Serializable with CatalystConf with Logging {
 
   def variableSubstituteDepth: Int = getConf(VARIABLE_SUBSTITUTE_DEPTH)
 
-  def warehousePath: String = new Path(getConf(WAREHOUSE_PATH)).toString
+  def warehousePath: String = Utils.resolveURI(getConf(WAREHOUSE_PATH)).toString
```
vanzin: So, I know I suggested this, but looking at it now, it seems wrong.

Before, if a user explicitly set this value in the configuration to `"/user/blah/warehouse"`, that would be interpreted as an HDFS path (or rather a path under `fs.defaultFS`). With this change it would be a local path. Right? So `Utils.resolveURI` should really only apply to the default value.

srowen: Well, before SPARK-15899 it was always interpreted as a local file path. After that change (i.e. right now) it would be interpreted as an HDFS path if that's what `fs.defaultFS` says. My proposition is that:

- It should default to `spark-warehouse` in the local filesystem working dir, because that's what the docs say (therefore, we have a bug at the moment after SPARK-15899)
- It should be interpreted as a local path if no scheme is given, because other stuff in Spark works that way via `Utils.resolveURI`
- It should be possible to set this to a non-local path, because you can do this with the Hive warehouse dir, which this option sort of parallels

I think the current change matches those requirements. I think we do not want "/user/blah/warehouse" to be interpreted as an HDFS path by default because that's not how it worked historically, and that it works this way now is I believe an error.

Is that convincing?

vanzin: > interpreted as an HDFS path by default because that's not how it worked historically

Except that is how it's worked historically for this piece of code at least. Taking a small detour, the reason I really dislike the use of `resolveURI` everywhere (and I commented on this when that code was added) is that it makes it awkward to use HDFS. Instead of relying on `fs.defaultFS` now you have to know the HDFS scheme (which can be "hdfs", "maprfs", "s3", "webhdfs", "viewfs", or others depending on your particular configuration) or things will break in mysterious ways.

Anyway, I don't really care that much for this particular setting, so if you think changing the behavior is better, it's fine with me.

srowen: You are right, the difference was that the _default_ before was always a local path. Right now ends up defaulting to an HDFS path (when using with HDFS for example). But explicitly setting "/foo/bar" has been interpreted as a HDFS path in the past and that would change. Hm, OK I see your point about only applying this to the default value.

Looking at `hive.metastore.warehouse.dir`, that seems to accept a path without scheme and that's supposed to be read as an HDFS path, I suppose.

The catch remains Windows paths. It should always be possible to write these as URIs, but, then becomes mandatory to write them as a URI, if I make this new change. That's a downside. That's only if you need to specify a local, non-default path on Windows, which might be a fairly corner case.

I can make the change to have it available for consideration here which I assume @vanzin  would favor, and I can see the logic of. How about @avulanov ? I know this has dragged on but I want to get it right. It's complex.

koertkuipers: i agree with @vanzin about dislike for resolveURI. i expect paths without schemes to on my default filesystem.

mridulm: @srowen Just for clarity, what is the proposed change in behavior expected (other than default being under cwd instead of $HOME) ? Thx

srowen:
@vanzin this may be more like what you're suggesting. I think the Windows issue makes this a little bit more complex than what you're describing but I think this ends up being workable.

@avulanov FYI

@mridulm I think the change in behavior from what's in `master` now is:

- Default to a local filesystem path (again) regardless of Hadoop config
- Default to `spark-warehouse` in working dir not user dir (as in the docs)

The net change is actually pretty minor after all.

vanzin:
LGTM. This should restore the behavior pre- 11a6844 and be compatible with Windows.

dongjoon-hyun:
Great. Is it going to 2.0.x, too?

gatorsmile:
cc @yhuai

sql/core/src/test/scala/org/apache/spark/sql/internal/SQLConfSuite.scala
```diff
@@ -219,8 +219,8 @@ class SQLConfSuite extends QueryTest with SharedSQLContext {
     try {
       // to get the default value, always unset it
       spark.conf.unset(SQLConf.WAREHOUSE_PATH.key)
-      assert(spark.sessionState.conf.warehousePath
-        === new Path(s"${System.getProperty("user.dir")}/spark-warehouse").toString)
+      assert(new Path(Utils.resolveURI("spark-warehouse")).toString ===
+        spark.sessionState.conf.warehousePath + "/")
```
felixcheung: could "/" change on Windows?

srowen: I think it will be a forward-slash because this will have to be a URI in any event in order to work on Windows. (The reason it's there is because the URI of a local directory will resolve with a trailing slash if the JDK knows the directory exists, otherwise it will be left as-is. The default URI won't have a trailing slash because it doesn't find the directory; it will exist at this point in the test though. Both are equally fine for Spark's purposes but that's why this exists in the test, for the record.)

srowen:
I think this would have to go into 2.0.x as I think it's a regression from 2.0.0, in that the default location moved from a local file to HDFS for many users, and, it's an HDFS path that can't be created. 

I'll leave it open for a few more days.

srowen:
Merged to master/2.0

PR #15387, opened 2016-10-07
[SPARK-17782][STREAMING][KAFKA] eliminate race condition of poll twice

koeninger:

Kafka consumers can't subscribe or maintain heartbeat without polling, but polling ordinarily consumes messages and adjusts position.  We don't want this on the driver, so we poll with a timeout of 0 and pause all topicpartitions.

Some consumer strategies that seek to particular positions have to poll first, but they weren't pausing immediately thereafter.  Thus, there was a race condition where the second poll() in the DStream start method might actually adjust consumer position.

Eliminated (or at least drastically reduced the chance of) the race condition via pausing in the relevant consumer strategies, and assert on startup that no messages were consumed.
## How was this patch tested?

I reliably reproduced the intermittent test failure by inserting a thread.sleep directly before returning from SubscribePattern.  The suggested fix eliminated the failure.

### DIFF
--- a/external/kafka-0-10/src/main/scala/org/apache/spark/streaming/kafka010/ConsumerStrategy.scala
+++ b/external/kafka-0-10/src/main/scala/org/apache/spark/streaming/kafka010/ConsumerStrategy.scala
@@ -159,6 +159,7 @@ private case class SubscribePattern[K, V](
       // we've called poll, we must pause or next poll may consume messages and set position
       consumer.pause(consumer.assignment())
     }
+
     consumer
   }
 }
--- a/external/kafka-0-10/src/main/scala/org/apache/spark/streaming/kafka010/ConsumerStrategy.scala
+++ b/external/kafka-0-10/src/main/scala/org/apache/spark/streaming/kafka010/ConsumerStrategy.scala
@@ -104,6 +104,8 @@ private case class Subscribe[K, V](
       toSeek.asScala.foreach { case (topicPartition, offset) =>
           consumer.seek(topicPartition, offset)
       }
+      // we've called poll, we must pause or next poll may consume messages and set position
+      consumer.pause(consumer.assignment())
     }
 
     consumer
@@ -154,8 +156,9 @@ private case class SubscribePattern[K, V](
       toSeek.asScala.foreach { case (topicPartition, offset) =>
           consumer.seek(topicPartition, offset)
       }
+      // we've called poll, we must pause or next poll may consume messages and set position
+      consumer.pause(consumer.assignment())
     }
-
     consumer
   }
 }
--- a/external/kafka-0-10/src/main/scala/org/apache/spark/streaming/kafka010/DirectKafkaInputDStream.scala
+++ b/external/kafka-0-10/src/main/scala/org/apache/spark/streaming/kafka010/DirectKafkaInputDStream.scala
@@ -223,7 +223,7 @@ private[spark] class DirectKafkaInputDStream[K, V](
 
   override def start(): Unit = {
     val c = consumer
-    c.poll(0)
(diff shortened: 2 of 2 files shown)


koeninger:
I'm not going to say anything is impossible, which is the point of the
assert. If it does somehow happen, it will be at start, so should be
obvious.

The whole poll 0 / pause thing is a gross hack, but it's what was suggested
by the kafka project dev list.

zsxwing:
> but polling ordinarily consumes messages and adjusts position.

Even if `enable.auto.commit` is `false`? In the doc, it says `automatically set as the last committed offset`, so I guess setting `enable.auto.commit` to `false` will prevent positions from updating.
```
     * On each poll, consumer will try to use the last consumed offset as the starting offset and fetch sequentially. The last
     * consumed offset can be manually set through {@link #seek(TopicPartition, long)} or automatically set as the last committed
     * offset for the subscribed list of partitions
```

koeninger:
I set auto commit to false, and still recreated the test failure.

That makes sense to me, consumer position should still be getting updated
in memory even if it isn't saved to storage anywhere.

At any rate, there are valid (albeit ill advised in my opinion) reasons to
turn on autocommit, so I'm not sure it matters.

zsxwing:
@koeninger If `poll(0)` returns a non empty results, how about just subtract the offsets back using the result size? Then we don't need to count on `poll(0)` not updating the offsets.

zsxwing:
#15397 is the fix for structured streaming.

koeninger:
You dont want poll consuming messages, its not just about offset
correctness, the driver shouldnt be spending time or bandwidth doing that.
What is the substantive concern with this solution, beyond the general
ugliness inherent in the consumer?  The chance that anyone out there has a
custom consumerstrategy that will be impacted by this, and wont have the
sophistication to deal with it, is pretty much nil.

koeninger:
Poll also isn't going to return you just messages for a single
topicpartition, so to do what you're suggesting you'd have to go through
all the messages and do additional processing, not just use result size.

koeninger:
If the concern is TD's comment,
"Future calls to {@link #poll(long)} will not return any records from these partitions until they have been resumed using {@link #resume(Collection)}."
https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/consumer/KafkaConsumer.java#L1353

During the original implementation I had verified that calling pause kills the internal message buffer, which is one of the complications leading to a cached consumer per partition.

I really don't think it's going to happen, but the assert is in there for paranoia, and to be explicit about the conditions.

koeninger:
Let me know if you guys like that alternative PR better

zsxwing:
> During the original implementation I had verified that calling pause kills the internal message buffer, which is one of the complications leading to a cached consumer per partition.

I observed the same behavior during my debug. I found that the first `poll(0)` will always send a request to prefetch the data. Pausing partitions just prevents the second `poll(0)` from returning anything at here: https://github.com/apache/kafka/blob/0.10.0.1/clients/src/main/java/org/apache/kafka/clients/consumer/internals/Fetcher.java#L527

> You dont want poll consuming messages, its not just about offset
correctness, the driver shouldnt be spending time or bandwidth doing that.

I think you have agreed that this is impossible via current KafkaConsumer APIs as well.

However, the unknown thing to me is that if the first `poll(0)` will return something. I saw the first `poll(0)` will always send the fetching request, but I'm not sure that if it's possible that the response will be processed in the first `poll(0)`. If this could happen, pausing partitions will not help in such case since it's called after the first `poll(0)`. In addition, since it's unclear in javadoc, it could be changed in the future. That's why I decided to manually seek to the beginning in #15397.

koeninger:
If you're worried about it then accept the alternative PR I linked.

PR #15388, opened 2016-10-07
[SPARK-17821][SQL] Support And and Or in Expression Canonicalize

viirya:

Currently `Canonicalize` object doesn't support `And` and `Or`. So we can compare canonicalized form of predicates consistently. We should add the support.
## How was this patch tested?

Jenkins tests.

### DIFF
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/Canonicalize.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/Canonicalize.scala
@@ -62,8 +62,12 @@ object Canonicalize extends {
     case a: Add => orderCommutative(a, { case Add(l, r) => Seq(l, r) }).reduce(Add)
     case m: Multiply => orderCommutative(m, { case Multiply(l, r) => Seq(l, r) }).reduce(Multiply)
 
-    case o: Or => orderCommutative(o, { case Or(l, r) => Seq(l, r) }).reduce(Or)
-    case a: And => orderCommutative(a, { case And(l, r) => Seq(l, r)}).reduce(And)
+    case o: Or =>
+      orderCommutative(o, { case Or(l, r) if l.deterministic && r.deterministic => Seq(l, r) })
+        .reduce(Or)
+    case a: And =>
+      orderCommutative(a, { case And(l, r) if l.deterministic && r.deterministic => Seq(l, r)})
+        .reduce(And)
 
     case EqualTo(l, r) if l.hashCode() > r.hashCode() => EqualTo(r, l)
     case EqualNullSafe(l, r) if l.hashCode() > r.hashCode() => EqualNullSafe(r, l)
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/Canonicalize.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/Canonicalize.scala
@@ -62,6 +62,9 @@ object Canonicalize extends {
     case a: Add => orderCommutative(a, { case Add(l, r) => Seq(l, r) }).reduce(Add)
     case m: Multiply => orderCommutative(m, { case Multiply(l, r) => Seq(l, r) }).reduce(Multiply)
 
+    case o: Or => orderCommutative(o, { case Or(l, r) => Seq(l, r) }).reduce(Or)
+    case a: And => orderCommutative(a, { case And(l, r) => Seq(l, r)}).reduce(And)
+
     case EqualTo(l, r) if l.hashCode() > r.hashCode() => EqualTo(r, l)
     case EqualNullSafe(l, r) if l.hashCode() > r.hashCode() => EqualNullSafe(r, l)
 
--- a/sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/ExpressionSetSuite.scala
+++ b/sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/ExpressionSetSuite.scala
@@ -80,6 +80,7 @@ class ExpressionSetSuite extends SparkFunSuite {
   setTest(1, Not(aUpper >= 1), aUpper < 1, Not(Literal(1) <= aUpper), Literal(1) > aUpper)
   setTest(1, Not(aUpper <= 1), aUpper > 1, Not(Literal(1) >= aUpper), Literal(1) < aUpper)
 
+  // Reordering AND/OR expressions
   setTest(1, aUpper > bUpper && aUpper <= 10, aUpper <= 10 && aUpper > bUpper)
   setTest(1,
     aUpper > bUpper && bUpper > 100 && aUpper <= 10,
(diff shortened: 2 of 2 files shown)


viirya:
cc @hvanhovell @cloud-fan

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/Canonicalize.scala
```diff
@@ -62,6 +62,9 @@ object Canonicalize extends {
     case a: Add => orderCommutative(a, { case Add(l, r) => Seq(l, r) }).reduce(Add)
     case m: Multiply => orderCommutative(m, { case Multiply(l, r) => Seq(l, r) }).reduce(Multiply)
 
+    case o: Or => orderCommutative(o, { case Or(l, r) => Seq(l, r) }).reduce(Or)
```
rxin: technically these are not communicative if there are side effects -- not sure if this matters in Spark SQL.

hvanhovell: Both`OR` and `AND` short circuit evaluation. This means that if you use a stateful (non-deterministic) expression in an `OR` or an `AND` you will change the result by reordering, for example: `RAND() > 0.5 OR x = 1`. The optimizer should not really change the output of a logical plan (and we have enforced this in the past).

I think we should only reorder expressions if they are deterministic.

viirya: oh. it makes sense. updated.

viirya:
ping @hvanhovell @rxin Can you take a look again? Thanks!

gatorsmile:
Why the following case can pass?
```Scala
  setTest(1,
    bUpper > 100 || 
      aUpper <= 10 &&
      aUpper > bUpper,
    bUpper > 100 || 
      (aUpper <= 10 && aUpper > bUpper))
```

gatorsmile:
BTW, we need to add more mixed cases in the test cases when supporting `And` and `Or`

viirya:
@gatorsmile Because `bUpper > 100 || aUpper <= 10 && aUpper > bUpper` is actually `bUpper > 100 || (aUpper <= 10 && aUpper > bUpper)` due to operator precedence.

cloud-fan:
LGTM, +1 on the suggestion of @gatorsmile to improve test coverage.

gatorsmile:
LGTM

Forgot the precedence of `AND` is higher than `OR`. 

BTW, also checked SQL. It follows the same precedence order. That is, `AND` is higher than `OR`.
```
From highest to lowest:
• ( ) (parenthetical expressions)
• +, -, ~ (unary operators)
• *, /, % (mathematical operators)
• +, - (arithmetic operators)
• =, >, <, >=, <=, <>, !=, !>, !< (comparison operators)
• ^ (bitwise exclusive OR), & (bitwise AND), | (bitwise OR)
• NOT
• AND
• ALL, ANY, BETWEEN, IN, LIKE, OR, SOME
• = (variable assignment)
```

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/ExpressionSetSuite.scala
```diff
@@ -80,6 +80,59 @@ class ExpressionSetSuite extends SparkFunSuite {
...
+  setTest(1,
+    aUpper >= 10 || bUpper <= 10 && aUpper === bUpper && aUpper < 100 || bUpper >= 100,
+    aUpper === bUpper && aUpper < 100 && bUpper <= 10 || bUpper >= 100 || aUpper >= 10,
+    aUpper < 100 && bUpper <= 10 && aUpper === bUpper || aUpper >= 10 || bUpper >= 100,
+    ((bUpper <= 10 && aUpper === bUpper) && aUpper < 100) || (aUpper >= 10 || bUpper >= 100))
+
+  // Don't reorder non-deterministic expression in AND/OR.
+  setTest(2, Rand(1L) > aUpper && aUpper <= 10, aUpper <= 10 && Rand(1L) > aUpper)
+  setTest(2,
+    aUpper > bUpper &&
```
rxin: a nit: why not just put these on one line?

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/ExpressionSetSuite.scala
```diff
@@ -80,6 +80,59 @@ class ExpressionSetSuite extends SparkFunSuite {
   setTest(1, Not(aUpper >= 1), aUpper < 1, Not(Literal(1) <= aUpper), Literal(1) > aUpper)
   setTest(1, Not(aUpper <= 1), aUpper > 1, Not(Literal(1) >= aUpper), Literal(1) < aUpper)
 
+  setTest(1, aUpper > bUpper && aUpper <= 10, aUpper <= 10 && aUpper > bUpper)
```
rxin: I'd also add more obvious inline comment to indicate these tests are for and/or reordering.

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/ExpressionSetSuite.scala
```diff
@@ -80,6 +80,59 @@ class ExpressionSetSuite extends SparkFunSuite {
...
+    aUpper > bUpper && bUpper > 100 && aUpper <= 10,
+    bUpper > 100 && aUpper <= 10 && aUpper > bUpper)
+
+  setTest(1, aUpper > bUpper || aUpper <= 10, aUpper <= 10 || aUpper > bUpper)
+  setTest(1,
+    aUpper > bUpper || bUpper > 100 || aUpper <= 10,
+    bUpper > 100 || aUpper <= 10 || aUpper > bUpper)
+
+  setTest(1,
+    bUpper > 100 || aUpper <= 10 && aUpper > bUpper,
```
rxin: I would add explicit parentheses to make the test case more clear; otherwise it is really confusing to think about precedence. You can argue precedence are important too, but that is not what these test cases are for. Even if you want to explicitly test precedence, you are merely testing scala operator precedence here, not SQL's.

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/ExpressionSetSuite.scala
```diff
@@ -80,6 +80,59 @@ class ExpressionSetSuite extends SparkFunSuite {
...
+    bUpper > 100 || (aUpper <= 10 && aUpper > bUpper))
+
+  setTest(1,
+    aUpper > 10 && bUpper < 10 || aUpper >= bUpper,
+    (bUpper < 10 && aUpper > 10) || aUpper >= bUpper)
+
+  setTest(1,
```
rxin: For these few test cases, they are getting too complicated that a human won't be able to tell immediately what the case is testing for. We should add some comment explaining what is being tested.

rxin:
@cloud-fan / @gatorsmile 

I left some comments on improving clarity. It's pretty important to the maintenance of the project.

gatorsmile:
@rxin Agree. Sorry for that. Will be more careful in the future.

viirya:
ping @rxin @hvanhovell @cloud-fan @gatorsmile any else need to address?

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/ExpressionSetSuite.scala
```diff
@@ -80,6 +80,65 @@ class ExpressionSetSuite extends SparkFunSuite {
...
+    (aUpper < 100 && bUpper <= 10 && aUpper === bUpper) || (aUpper >= 10) || (bUpper >= 100),
+    ((bUpper <= 10 && aUpper === bUpper) && aUpper < 100) || ((aUpper >= 10) || (bUpper >= 100)))
+
+  // Don't reorder non-deterministic expression in AND/OR.
+  setTest(2, Rand(1L) > aUpper && aUpper <= 10, aUpper <= 10 && Rand(1L) > aUpper)
+  setTest(2,
+    aUpper > bUpper && bUpper > 100 && Rand(1L) > aUpper,
+    bUpper > 100 && Rand(1L) > aUpper && aUpper > bUpper)
+
+  setTest(2, Rand(1L) > aUpper || aUpper <= 10, aUpper <= 10 || Rand(1L) > aUpper)
```
viirya: Good to have. Added.

cloud-fan:
thanks, merging to master!

viirya:
Thanks! @rxin @cloud-fan @hvanhovell @gatorsmile

PR #15389, opened 2016-10-07
[SPARK-17817][PySpark] PySpark RDD Repartitioning Results in Highly Skewed Partition Sizes

viirya:

Quoted from JIRA description:

Calling repartition on a PySpark RDD to increase the number of partitions results in highly skewed partition sizes, with most having 0 rows. The repartition method should evenly spread out the rows across the partitions, and this behavior is correctly seen on the Scala side.

Please reference the following code for a reproducible example of this issue:

```
num_partitions = 20000
a = sc.parallelize(range(int(1e6)), 2)  # start with 2 even partitions
l = a.repartition(num_partitions).glom().map(len).collect()  # get length of each partition
min(l), max(l), sum(l)/len(l), len(l)  # skewed!
```

In Scala's `repartition` code, we will distribute elements evenly across output partitions. However, the RDD from Python is serialized as a single binary data, so the distribution fails. We need to convert the RDD in Python to java object before repartitioning.
## How was this patch tested?

Jenkins tests.

### DIFF
--- a/python/pyspark/rdd.py
+++ b/python/pyspark/rdd.py
@@ -2017,9 +2017,7 @@ class RDD(object):
          >>> len(rdd.repartition(10).glom().collect())
          10
         """
-        data_java_rdd = self._to_java_object_rdd().repartition(numPartitions)
-        jrdd = self.ctx._jvm.SerDeUtil.javaToPython(data_java_rdd)
-        return RDD(jrdd, self.ctx, self._jrdd_deserializer)
+        return self.coalesce(numPartitions, shuffle=True)
 
     def coalesce(self, numPartitions, shuffle=False):
         """
--- a/python/pyspark/rdd.py
+++ b/python/pyspark/rdd.py
@@ -2017,7 +2017,8 @@ class RDD(object):
          >>> len(rdd.repartition(10).glom().collect())
          10
         """
-        jrdd = self._jrdd.repartition(numPartitions)
+        data_java_rdd = self._to_java_object_rdd().repartition(numPartitions)
+        jrdd = self.ctx._jvm.SerDeUtil.javaToPython(data_java_rdd)
         return RDD(jrdd, self.ctx, self._jrdd_deserializer)
 
     def coalesce(self, numPartitions, shuffle=False):
@@ -2029,7 +2030,11 @@ class RDD(object):
         >>> sc.parallelize([1, 2, 3, 4, 5], 3).coalesce(1).glom().collect()
         [[1, 2, 3, 4, 5]]
         """
-        jrdd = self._jrdd.coalesce(numPartitions, shuffle)
+        if shuffle:
+          data_java_rdd = self._to_java_object_rdd().coalesce(numPartitions, shuffle)
+          jrdd = self.ctx._jvm.SerDeUtil.javaToPython(data_java_rdd)
+        else:
+          jrdd = self._jrdd.coalesce(numPartitions, shuffle)
         return RDD(jrdd, self.ctx, self._jrdd_deserializer)
 
     def zip(self, other):
--- a/python/pyspark/rdd.py
+++ b/python/pyspark/rdd.py
@@ -2029,6 +2029,10 @@ class RDD(object):
         [[1, 2, 3, 4, 5]]
         """
         if shuffle:
+            # In Scala's repartition code, we will distribute elements evenly across output
+            # partitions. However, the RDD from Python is serialized as a single binary data,
+            # so the distribution fails and produces highly skewed partitions. We need to
+            # convert it to a RDD of java object before repartitioning.
             data_java_rdd = self._to_java_object_rdd().coalesce(numPartitions, shuffle)
             jrdd = self.ctx._jvm.SerDeUtil.javaToPython(data_java_rdd)
         else:
--- a/python/pyspark/rdd.py
+++ b/python/pyspark/rdd.py
@@ -2031,10 +2031,10 @@ class RDD(object):
         [[1, 2, 3, 4, 5]]
         """
         if shuffle:
-          data_java_rdd = self._to_java_object_rdd().coalesce(numPartitions, shuffle)
-          jrdd = self.ctx._jvm.SerDeUtil.javaToPython(data_java_rdd)
+            data_java_rdd = self._to_java_object_rdd().coalesce(numPartitions, shuffle)
+            jrdd = self.ctx._jvm.SerDeUtil.javaToPython(data_java_rdd)
         else:
-          jrdd = self._jrdd.coalesce(numPartitions, shuffle)
+            jrdd = self._jrdd.coalesce(numPartitions, shuffle)
         return RDD(jrdd, self.ctx, self._jrdd_deserializer)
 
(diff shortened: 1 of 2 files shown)


viirya:
cc @davies

python/pyspark/rdd.py
```diff
@@ -2029,7 +2030,11 @@ def coalesce(self, numPartitions, shuffle=False):
         >>> sc.parallelize([1, 2, 3, 4, 5], 3).coalesce(1).glom().collect()
         [[1, 2, 3, 4, 5]]
         """
-        jrdd = self._jrdd.coalesce(numPartitions, shuffle)
+        if shuffle:
```
holdenk: Seems you could just call `repartition` here to avoid the code duplication or swap `repartition` to call `coalesce`.

python/pyspark/rdd.py
```diff
@@ -2029,7 +2030,11 @@ def coalesce(self, numPartitions, shuffle=False):
...
+            data_java_rdd = self._to_java_object_rdd().coalesce(numPartitions, shuffle)
+            jrdd = self.ctx._jvm.SerDeUtil.javaToPython(data_java_rdd)
```
holdenk: I'm not as familiar with this part as I should be, but do we have a good idea of how expensive this is? It might be good to do some quick benchmarking just to make sure that this change doesn't have any unintended side effects?

holdenk (commented):
Thanks for working on this - yay better partitioning for Python :) Some minor comments and it might make sense to do a quick benchmark to make sure we don't have any unintentional regression here?

python/pyspark/rdd.py
dusenberrymw: Yeah to be consistent with the Scala side, we should rearrange this to have `repartition(...)` call `coalesce(..., shuffle=True)`.

dusenberrymw: Yeah to be consistent with the Scala side, it would be nice to rearrange this to have Python's `repartition(...)` call Python's `coalesce(..., shuffle=True)`.

dusenberrymw (commented):
Thanks @viirya for jumping on this so quickly!  I agree with a few thoughts by @holdenk, and overall looking forward to making use of this fix!

python/pyspark/rdd.py
viirya: Make sense. Updated.

python/pyspark/rdd.py
viirya: I do a simple benchmark:

    import time
    num_partitions = 20000
    a = sc.parallelize(range(int(1e6)), 2)
    start = time.time()
    l = a.repartition(num_partitions).glom().map(len).collect()
    end = time.time()
    print(end - start)

Before: 419.447577953
After: 421.916361094

I think there is no significant difference.

holdenk: Yah that seems close enough we don't need to worry (and for the big cases presumably the impact of having better balanced partitions is well worth the slight overhead).

holdenk:
This looks good to me, one alternative is that we could try and fix it by doing better shuffling of the batched chunks but this wouldn't work well for increasing the number of partitions.
Maybe @davies can take a look and see if there is anything else that needs to be done?

holdenk:
Maybe @hyukjinkwon could also do a review pass while we wait for @davies or someone with commit privileges to come by and do a final review.

HyukjinKwon:
@holdenk Thanks you for cc'ing me. It looks okay to me as targeted but I feel we need a sign-off.

viirya:
@holdenk @dusenberrymw  @HyukjinKwon Thanks for review!

felixcheung:
great! LGTM and thank you for the thorough review/test/feedback from everyone.
let's hold for another day or 2 :)

python/pyspark/rdd.py
```diff
@@ -2029,7 +2028,11 @@ def coalesce(self, numPartitions, shuffle=False):
         >>> sc.parallelize([1, 2, 3, 4, 5], 3).coalesce(1).glom().collect()
         [[1, 2, 3, 4, 5]]
         """
-        jrdd = self._jrdd.coalesce(numPartitions, shuffle)
+        if shuffle:
+            data_java_rdd = self._to_java_object_rdd().coalesce(numPartitions, shuffle)
```
rxin: would be great to add some inline comment explaining why this is necessary. otherwise somebody can just come in 6 month from now and change this back to `jrdd = self._jrdd.coalesce(numPartitions, shuffle)`

viirya:
retest this please.

viirya:
Seems Jenkins are not in working status?

viirya:
retest this please.

python/pyspark/rdd.py
```diff
@@ -2029,7 +2028,15 @@ def coalesce(self, numPartitions, shuffle=False):
         >>> sc.parallelize([1, 2, 3, 4, 5], 3).coalesce(1).glom().collect()
         [[1, 2, 3, 4, 5]]
         """
-        jrdd = self._jrdd.coalesce(numPartitions, shuffle)
+        if shuffle:
+            # In Scala's repartition code, we will distribute elements evenly across output
+            # partitions. However, the RDD from Python is serialized as a single binary data,
+            # so the distribution fails and produces highly skewed partitions. We need to
+            # convert it to a RDD of java object before repartitioning.
+            data_java_rdd = self._to_java_object_rdd().coalesce(numPartitions, shuffle)
```
davies: The reason that cause the skew should be large batch size, I think we could decrease the batch size to 10, then call repartition in JVM.

My worry is that _to_java_object_rdd() could be expensive, maybe we should have some benchmark for that.

davies:
Did not release that we already merged this, should we left a message here or in the JIRA so we can know who merge this?

rxin:
@felixcheung merged it I believe.

@felixcheung please make sure you leave a msg saying it's merged (along with the branch) when you merge prs.

python/pyspark/rdd.py
HyukjinKwon: Hi @davies, actually it seems a simple benchmark was done in https://github.com/apache/spark/pull/15389#discussion_r82444378

If you worry, then, I'd like to proceed another benchmark with larger data and then will share when I have some time.

viirya: @davies Thank you! I do a simple benchmark as above with decreasing the batch size, I don't see an improvement in running time. I.e.,

    import time
    num_partitions = 20000
    a = sc.parallelize(range(int(1e6)), 2)
    start = time.time()
    l = a.repartition(num_partitions).glom().map(len).collect()
    end = time.time()
    print(end - start)

Before: 419.447577953
_to_java_object_rdd(): 421.916361094
decreasing the batch size: 423.712255955

Maybe it depends how is expensive actually converting to java object case by case. Is it generally faster than _to_java_object_rdd()? I would open a followup for this change.

felixcheung:
@rxin sorry, it's me and I thought commit history and JIRA would show who did it - I'll be sure to add a note the next time.
@davies sorry I'll make sure to ask to wait a bit longer for you the next time 😄 

This was merged to master.

PR #15390, opened 2016-10-07
[SPARK-17806] [SQL] fix bug in join key rewritten in HashJoin

davies:

In HashJoin, we try to rewrite the join key as Long to improve the performance of finding a match. The rewriting part is not well tested, has a bug that could cause wrong result when there are at least three integral columns in the joining key also the total length of the key exceed 8 bytes. 
## How was this patch tested?

Added unit test to covering the rewriting with different number of columns and different data types. Manually test the reported case and confirmed that this PR fix the bug.

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/joins/HashJoin.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/joins/HashJoin.scala
@@ -63,45 +63,16 @@ trait HashJoin {
   protected lazy val (buildKeys, streamedKeys) = {
     require(leftKeys.map(_.dataType) == rightKeys.map(_.dataType),
       "Join keys from two sides should have same types")
-    val lkeys = rewriteKeyExpr(leftKeys).map(BindReferences.bindReference(_, left.output))
-    val rkeys = rewriteKeyExpr(rightKeys).map(BindReferences.bindReference(_, right.output))
+    val lkeys = HashJoin.rewriteKeyExpr(leftKeys).map(BindReferences.bindReference(_, left.output))
+    val rkeys = HashJoin.rewriteKeyExpr(rightKeys)
+      .map(BindReferences.bindReference(_, right.output))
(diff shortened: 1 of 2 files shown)


davies:
cc @rxin

sql/core/src/main/scala/org/apache/spark/sql/execution/joins/HashJoin.scala
```diff
@@ -247,3 +218,37 @@ trait HashJoin {
...
+              keyExpr = Cast(e, LongType)
+            } else {
+              keyExpr = e
+            }
+            width = dt.defaultSize
+          } else {
+            val bits = dt.defaultSize * 8
+            keyExpr = BitwiseOr(ShiftLeft(keyExpr, Literal(bits)),
+              BitwiseAnd(Cast(e, LongType), Literal((1L << bits) - 1)))
+            width += dt.defaultSize
```
hvanhovell: This is the only change (for people also reviewing this).

hvanhovell:
LGTM pending Jenkins.

sql/core/src/main/scala/org/apache/spark/sql/execution/joins/HashJoin.scala
```diff
@@ -247,3 +218,37 @@ trait HashJoin {
...
   }
 }
+
+object HashJoin {
+  /**
+   * Try to rewrite the key as LongType so we can use getLong(), if they key can fit with a long.
+   *
+   * If not, returns the original expressions.
+   */
+  private[joins] def rewriteKeyExpr(keys: Seq[Expression]): Seq[Expression] = {
```
rxin: This will be a lot clear and less error prone if we can write the check first and go through the data twice, e.g.

```
if (keys.map(_.dataType.defaultSize).sum <= 8) {
  return keys
}

// do the rewrite here
```

rxin:
LGTM, pending the new test run.

PR #15394, opened 2016-10-07
[SPARK-17748][ML] One pass solver for Weighted Least Squares with ElasticNet

sethah:
1. Make a pluggable solver interface for `WeightedLeastSquares`
2. Add a `QuasiNewton` solver to handle elastic net regularization for `WeightedLeastSquares`
3. Add method `BLAS.dspmv` used by QN solver
4. Add mechanism for WLS to handle singular covariance matrices by falling back to QN solver when Cholesky fails.
## How was this patch tested?

Unit tests - see below.
## Design choices

**Pluggable Normal Solver**

Before, the `WeightedLeastSquares` package always used the Cholesky decomposition solver to compute the solution to the normal equations. Now, we specify the solver as a constructor argument to the `WeightedLeastSquares`. We introduce a new trait:

``` scala
private[ml] sealed trait NormalEquationSolver {

  def solve(
      bBar: Double,
      bbBar: Double,
      abBar: DenseVector,
      aaBar: DenseVector,
      aBar: DenseVector): NormalEquationSolution
}
```

We extend this trait for different variants of normal equation solvers. In the future, we can easily add others (like QR) using this interface.

**Always train in the standardized space**

The normal solver did not previously standardize the data, but this patch introduces a change such that we always solve the normal equations in the standardized space. We convert back to the original space in the same way that is done for distributed L-BFGS/OWL-QN. We add test cases for zero variance features/labels.

**Use L-BFGS locally to solve normal equations for singular matrix**

When linear regression with the normal solver is called for a singular matrix, we initially try to solve with Cholesky. We use the output of `lapack.dppsv` to determine if the matrix is singular. If it is, we fall back to using L-BFGS locally to solve the normal equations. We add test cases for this as well.
## Test cases

I found it helpful to enumerate some of the test cases and hopefully it makes review easier.

**WeightedLeastSquares**
1. Constant columns - Cholesky solver fails with no regularization, Auto solver falls back to QN, and QN trains successfully.
2. Collinear features - Cholesky solver fails with no regularization, Auto solver falls back to QN, and QN trains successfully.
3. Label is constant zero - no training is performed regardless of intercept. Coefficients are zero and intercept is zero.
4. Label is constant - if fitIntercept, then no training is performed and intercept equals label mean. If not fitIntercept, then we train and return an answer that matches R's lm package.
5. Test with L1 - go through various combinations of L1/L2, standardization, fitIntercept and verify that output matches glmnet.
6. Initial intercept - verify that setting the initial intercept to label mean is correct by training model with strong L1 regularization so that all coefficients are zero and intercept converges to label mean.
7. Test diagInvAtWA - since we are standardizing features now during training, we should test that the inverse is computed to match R.

**LinearRegression**
1. For all existing L1 test cases, test the "normal" solver too.
2. Check that using the normal solver now handles singular matrices.
3. Check that using the normal solver with L1 produces an objective history in the model summary, but does not produce the inverse of AtA.

**BLAS**
1. Test new method `dspmv`.
## Performance Testing

This patch will speed up linear regression with L1/elasticnet penalties when the feature size is < 4096. I have not conducted performance tests at scale, only observed by testing locally that there is a speed improvement. 

We should decide if this PR needs to be blocked before performance testing is conducted.

### DIFF
--- a/mllib-local/src/main/scala/org/apache/spark/ml/linalg/BLAS.scala
+++ b/mllib-local/src/main/scala/org/apache/spark/ml/linalg/BLAS.scala
@@ -244,12 +244,20 @@ private[spark] object BLAS extends Serializable {
   }
 
   /**
-   * y += alpha * A * x
+   * y := alpha*A*x + beta*y
    *
-   * @param A The upper triangular part of A in a [[DenseVector]] (column major)
+   * @param A The upper triangular part of A in a [[DenseVector]] (column major).
+   * @param x The [[DenseVector]] transformed by A.
+   * @param y The [[DenseVector]] to be modified in place.
    */
-  def dspmv(n: Int, alpha: Double, A: DenseVector, x: DenseVector, y: DenseVector): Unit = {
-    f2jBLAS.dspmv("U", n, alpha, A.values, x.values, 1, 1.0, y.values, 1)
+  def dspmv(
+      n: Int,
+      alpha: Double,
+      A: DenseVector,
+      x: DenseVector,
+      beta: Double,
+      y: DenseVector): Unit = {
+    f2jBLAS.dspmv("U", n, alpha, A.values, x.values, 1, beta, y.values, 1)
   }
 
   /**
--- a/mllib-local/src/main/scala/org/apache/spark/ml/linalg/BLAS.scala
+++ b/mllib-local/src/main/scala/org/apache/spark/ml/linalg/BLAS.scala
@@ -244,6 +244,15 @@ private[spark] object BLAS extends Serializable {
   }
 
   /**
+   * y += alpha * A * x
+   *
+   * @param A The upper triangular part of A in a [[DenseVector]] (column major)
+   */
+  def dspmv(n: Int, alpha: Double, A: DenseVector, x: DenseVector, y: DenseVector): Unit = {
+    f2jBLAS.dspmv("U", n, alpha, A.values, x.values, 1, 1.0, y.values, 1)
+  }
+
+  /**
    * Adds alpha * x * x.t to a matrix in-place. This is the same as BLAS's ?SPR.
    *
    * @param U the upper triangular part of the matrix packed in an array (column major)
--- a/mllib-local/src/main/scala/org/apache/spark/ml/linalg/BLAS.scala
+++ b/mllib-local/src/main/scala/org/apache/spark/ml/linalg/BLAS.scala
@@ -246,6 +246,7 @@ private[spark] object BLAS extends Serializable {
   /**
    * y := alpha*A*x + beta*y
    *
+   * @param n The order of the n by n matrix A.
    * @param A The upper triangular part of A in a [[DenseVector]] (column major).
    * @param x The [[DenseVector]] transformed by A.
    * @param y The [[DenseVector]] to be modified in place.
--- a/mllib-local/src/test/scala/org/apache/spark/ml/linalg/BLASSuite.scala
+++ b/mllib-local/src/test/scala/org/apache/spark/ml/linalg/BLASSuite.scala
@@ -422,4 +422,37 @@ class BLASSuite extends SparkMLFunSuite {
     assert(dATT.multiply(sx) ~== expected absTol 1e-15)
     assert(sATT.multiply(sx) ~== expected absTol 1e-15)
   }
+
+  test("spmv") {
+    /*
+      A = [[3.0, -2.0, 2.0, -4.0],
+           [-2.0, -8.0, 4.0, 7.0],
+           [2.0, 4.0, -3.0, -3.0],
+           [-4.0, 7.0, -3.0, 0.0]]
+      x =  [5.0, 2.0, 1.0, 9.0]
+      Ax = [ 45., -93.,  48.,  -3.]
+     */
+    val A = new DenseVector(Array(3.0, -2.0, -8.0, 2.0, 4.0, -3.0, -4.0, 7.0, -3.0, 0.0))
+    val x = new DenseVector(Array(5.0, 2.0, -1.0, -9.0))
+    val n = 4
+
+    val y1 = new DenseVector(Array(-3.0, 6.0, -8.0, -3.0))
+    val y2 = y1.copy
+    val y3 = y1.copy
+    val y4 = y1.copy
+
+    val expected1 = new DenseVector(Array(42.0, -87.0, 40.0, -6.0))
+    val expected2 = new DenseVector(Array(19.5, -40.5, 16.0, -4.5))
+    val expected3 = new DenseVector(Array(-25.5, 52.5, -32.0, -1.5))
+    val expected4 = new DenseVector(Array(-3.0, 6.0, -8.0, -3.0))
+
+    dspmv(n, 1.0, A, x, y1)
+    dspmv(n, 0.5, A, x, y2)
+    dspmv(n, -0.5, A, x, y3)
+    dspmv(n, 0.0, A, x, y4)
+    assert(y1 ~== expected1 absTol 1e-8)
+    assert(y2 ~== expected2 absTol 1e-8)
+    assert(y3 ~== expected3 absTol 1e-8)
+    assert(y4 ~== expected4 absTol 1e-8)
+  }
 }
--- a/mllib-local/src/test/scala/org/apache/spark/ml/linalg/BLASSuite.scala
+++ b/mllib-local/src/test/scala/org/apache/spark/ml/linalg/BLASSuite.scala
@@ -429,7 +429,7 @@ class BLASSuite extends SparkMLFunSuite {
            [-2.0, -8.0, 4.0, 7.0],
            [2.0, 4.0, -3.0, -3.0],
            [-4.0, 7.0, -3.0, 0.0]]
-      x =  [5.0, 2.0, 1.0, 9.0]
+      x =  [5.0, 2.0, -1.0, -9.0]
       Ax = [ 45., -93.,  48.,  -3.]
      */
     val A = new DenseVector(Array(3.0, -2.0, -8.0, 2.0, 4.0, -3.0, -4.0, 7.0, -3.0, 0.0))
@@ -440,19 +440,31 @@ class BLASSuite extends SparkMLFunSuite {
     val y2 = y1.copy
     val y3 = y1.copy
     val y4 = y1.copy
+    val y5 = y1.copy
+    val y6 = y1.copy
+    val y7 = y1.copy
 
     val expected1 = new DenseVector(Array(42.0, -87.0, 40.0, -6.0))
     val expected2 = new DenseVector(Array(19.5, -40.5, 16.0, -4.5))
     val expected3 = new DenseVector(Array(-25.5, 52.5, -32.0, -1.5))
     val expected4 = new DenseVector(Array(-3.0, 6.0, -8.0, -3.0))
+    val expected5 = new DenseVector(Array(43.5, -90.0, 44.0, -4.5))
+    val expected6 = new DenseVector(Array(46.5, -96.0, 52.0, -1.5))
+    val expected7 = new DenseVector(Array(45.0, -93.0, 48.0, -3.0))
 
-    dspmv(n, 1.0, A, x, y1)
-    dspmv(n, 0.5, A, x, y2)
-    dspmv(n, -0.5, A, x, y3)
-    dspmv(n, 0.0, A, x, y4)
+    dspmv(n, 1.0, A, x, 1.0, y1)
+    dspmv(n, 0.5, A, x, 1.0, y2)
+    dspmv(n, -0.5, A, x, 1.0, y3)
+    dspmv(n, 0.0, A, x, 1.0, y4)
+    dspmv(n, 1.0, A, x, 0.5, y5)
+    dspmv(n, 1.0, A, x, -0.5, y6)
+    dspmv(n, 1.0, A, x, 0.0, y7)
     assert(y1 ~== expected1 absTol 1e-8)
     assert(y2 ~== expected2 absTol 1e-8)
     assert(y3 ~== expected3 absTol 1e-8)
     assert(y4 ~== expected4 absTol 1e-8)
+    assert(y5 ~== expected5 absTol 1e-8)
+    assert(y6 ~== expected6 absTol 1e-8)
+    assert(y7 ~== expected7 absTol 1e-8)
   }
 }
--- a/mllib/src/main/scala/org/apache/spark/ml/optim/NormalEquationSolver.scala
+++ b/mllib/src/main/scala/org/apache/spark/ml/optim/NormalEquationSolver.scala
@@ -26,18 +26,7 @@ import org.apache.spark.mllib.linalg.CholeskyDecomposition
 private[ml] class NormalEquationSolution(
     val coefficients: Array[Double],
     val aaInv: Option[Array[Double]],
-    val objectiveHistory: Option[Array[Double]]) {
-
-//  def coefficients: Array[Double] = {
-//    if (fitIntercept) {
-//      _coefficients.slice(0, _coefficients.length - 1)
-//    } else {
-//      _coefficients
-//    }
-//  }
-//
-//  def intercept: Double = if (fitIntercept) _coefficients.last else 0.0
-}
+    val objectiveHistory: Option[Array[Double]])
 
 /**
  * Interface for classes that solve the normal equations locally.
--- a/mllib/src/main/scala/org/apache/spark/ml/optim/NormalEquationSolver.scala
+++ b/mllib/src/main/scala/org/apache/spark/ml/optim/NormalEquationSolver.scala
@@ -1,0 +1,165 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.spark.ml.optim
+
+import breeze.linalg.{DenseVector => BDV}
+import breeze.optimize.{CachedDiffFunction, DiffFunction, LBFGS => BreezeLBFGS, OWLQN => BreezeOWLQN}
+import scala.collection.mutable
+
+import org.apache.spark.ml.linalg.{BLAS, DenseVector, Vectors}
+import org.apache.spark.mllib.linalg.CholeskyDecomposition
+
+private[ml] class NormalEquationSolution(
+    val fitIntercept: Boolean,
+    private val _coefficients: Array[Double],
+    val aaInv: Option[DenseVector],
+    val objectiveHistory: Option[Array[Double]]) {
+
+  def coefficients: DenseVector = {
+    if (fitIntercept) {
+      new DenseVector(_coefficients.slice(0, _coefficients.length - 1))
+    } else {
+      new DenseVector(_coefficients)
+    }
+  }
+
+  def intercept: Double = if (fitIntercept) _coefficients.last else 0.0
+}
+
+/**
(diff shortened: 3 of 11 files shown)


mllib/src/main/scala/org/apache/spark/ml/optim/WeightedLeastSquares.scala
```diff
@@ -85,73 +101,193 @@ private[ml] class WeightedLeastSquares(
...
+      new QuasiNewtonSolver(fitIntercept, maxIter, tol, effectiveL1RegFun)
     } else {
-      abBar.values
+      new CholeskySolver(fitIntercept)
     }
 
-    val x = CholeskyDecomposition.solve(aa, ab)
+    val solution = solver match {
+      case cholesky: CholeskySolver =>
+        try {
```
sethah: I don't really like this solution. I'd appreciate feedback on alternatives here.

sethah:
cc @yanboliang @dbtsai

sethah:
jenkins retest this please

mllib-local/src/main/scala/org/apache/spark/ml/linalg/BLAS.scala
```diff
@@ -244,6 +244,15 @@ private[spark] object BLAS extends Serializable {
   }
 
   /**
+   * y += alpha * A * x
+   *
+   * @param A The upper triangular part of A in a [[DenseVector]] (column major)
+   */
+  def dspmv(n: Int, alpha: Double, A: DenseVector, x: DenseVector, y: DenseVector): Unit = {
```
yanboliang: Actually BLAS DSPMV performs matrix-vector operation ```y := alpha*A*x + beta*y```, should we provide the same function as original BLAS? Since it would be used by other MLlib function in the future.

mllib/src/main/scala/org/apache/spark/ml/optim/NormalEquationSolver.scala
```diff
@@ -0,0 +1,165 @@
...
+    val aaInv = CholeskyDecomposition.inverse(aaBar.values, k)
+
+    new NormalEquationSolution(fitIntercept, x, Some(new DenseVector(aaInv)), None)
+  }
+}
+
+/**
+ * A class for solving the normal equations using Quasi-Newton optimization methods.
+ */
+private[ml] class QuasiNewtonSolver(
```
yanboliang: should ```LocalQuasiNewtonSolver``` be better? It's easy to confuse with distributed L-BFGS solver after we add optimizer interface at [SPARK-17136](https://issues.apache.org/jira/browse/SPARK-17136). Or other suggestion?

yanboliang (commented):
@sethah Looks like a great change. I made a quick pass and left some comments. Thanks.

yanboliang:
Will make another indepth pass in a few days. Thanks.

mllib/src/main/scala/org/apache/spark/ml/optim/NormalEquationSolver.scala
sethah: I think the fact that it extends `NormalEquationSolver` implies it is local. We can easily add another `QuasiNewtonSolver` that extends from some distributed implementation (or even change the name of this later if needed).

mllib/src/main/scala/org/apache/spark/ml/regression/LinearRegression.scala
```diff
@@ -217,7 +215,7 @@ class LinearRegression @Since("1.3.0") (@Since("1.3.0") override val uid: String
         $(featuresCol),
         summaryModel,
         model.diagInvAtWA.toArray,
-        Array(0D))
+        model.objectiveHistory)
```
sethah: How is it different? Mathematically it should be the same.

mllib/src/main/scala/org/apache/spark/ml/optim/NormalEquationSolver.scala
```diff
@@ -0,0 +1,165 @@
...
+import breeze.optimize.{CachedDiffFunction, DiffFunction, LBFGS => BreezeLBFGS, OWLQN => BreezeOWLQN}
+import scala.collection.mutable
+
+import org.apache.spark.ml.linalg.{BLAS, DenseVector, Vectors}
+import org.apache.spark.mllib.linalg.CholeskyDecomposition
+
+private[ml] class NormalEquationSolution(
+    val fitIntercept: Boolean,
+    private val _coefficients: Array[Double],
+    val aaInv: Option[DenseVector],
```
yanboliang: I'd use ```Option[Array[Double]]``` to represent ```aaInv```, since we will scale it to original space at ```WeightedLeastSquares``` L263.
BTW, should we use also expose ```coefficients``` as ```Array[Double]``` rather than ```DenseVector```?
I found we did in-place update to scale back at ```WeightedLeastSquares``` L255. I'm more prefer to get coefficients in array by solver, then scale back and wrap it with ```DenseVector``` at upper layer. You can refer the ```LinearRegression``` solved by ```L-BFGS```. Only suggestion, not big deal.

mllib/src/main/scala/org/apache/spark/ml/regression/LinearRegression.scala
yanboliang: Sorry for mistake, I agree it should be equivalent.

yanboliang (commented):
@sethah I left some comments, thanks for working on this great feature.

sethah:
Thanks a lot @yanboliang for your review. I will get to these today.

mllib/src/test/scala/org/apache/spark/ml/optim/WeightedLeastSquaresSuite.scala
```diff
@@ -132,24 +197,234 @@ class WeightedLeastSquaresSuite extends SparkFunSuite with MLlibTestSparkContext
...
+      // for the case of no intercept, this would not have failed before but since we train
+      // in the standardized space now, it will fail
+      intercept[SingularMatrixException] {
+        wls.fit(constantFeatures)
+      }
+    }
+
+    // should not fail when regularization is added
+    new WeightedLeastSquares(true, 0.5, 0.0, standardizeFeatures = true,
+      standardizeLabel = true, solverType = WeightedLeastSquares.Cholesky).fit(constantFeatures)
```
sethah: ok I added some checks. Actually, Cholesky now fails if regularization is added, but standardization is false. This is because in this case we have to divide by the variance of the features along the diagonal, and that would cause divide by zero, so we make that element of the diagonal zero. I think this is ok, and not a regression, since it will fail but fall back to QN solver. I added test cases to make this clearer.

mllib/src/test/scala/org/apache/spark/ml/optim/WeightedLeastSquaresSuite.scala
```diff
@@ -201,11 +476,13 @@ class WeightedLeastSquaresSuite extends SparkFunSuite with MLlibTestSparkContext
...
+      for (solver <- WeightedLeastSquares.supportedSolvers) {
+        val wls = new WeightedLeastSquares(fitIntercept, regParam, elasticNetParam = 0.0,
+          standardizeFeatures, standardizeLabel = true, solverType = solver)
+          .fit(instances)
+        val actual = Vectors.dense(wls.intercept, wls.coefficients(0), wls.coefficients(1))
+        assert(actual ~== expected(idx) absTol 1e-4)
+      }
       idx += 1
     }
   }
```
yanboliang: Yeah, may be we can leave TODO here to consider remove this arguments for WLS. Let's address it in follow up work when we are confidence.

yanboliang (commented):
@sethah I made another pass, the main feedback is to move `fitIntercept` out of NE solver. Thanks for the great effort.

mllib-local/src/test/scala/org/apache/spark/ml/linalg/BLASSuite.scala
```diff
@@ -422,4 +422,49 @@ class BLASSuite extends SparkMLFunSuite {
     assert(dATT.multiply(sx) ~== expected absTol 1e-15)
     assert(sATT.multiply(sx) ~== expected absTol 1e-15)
   }
+
+  test("spmv") {
```
sethah: the naming conventions for the other BLAS tests is not to use the "d" or "s" prefix - for example the test for "gemm" and "syr" (not "dgemm" or "dsyr"). I think this is correct given the other test names.

mllib/src/main/scala/org/apache/spark/ml/optim/NormalEquationSolver.scala
```diff
@@ -0,0 +1,165 @@
...
+      bbBar: Double,
+      abBar: DenseVector,
+      aaBar: DenseVector,
+      aBar: DenseVector): NormalEquationSolution
+}
+
+/**
+ * A class that solves the normal equations directly, using Cholesky decomposition.
+ */
+private[ml] class CholeskySolver(val fitIntercept: Boolean) extends NormalEquationSolver {
```
sethah: In the `NormalEquationCostFun` we set the last coefficient (the intercept) to the analytically correct value in every iteration. I'm not sure how we can do this without having knowledge of whether or not we are fitting an intercept term. This modification does in fact have an effect on the convergence of the algorithm, so I think it's important to keep. 

We can still move it out of `CholeskySolver` and `NormalEquationSolution` if you prefer. Not sure about `QuasiNewtonSolver`. Thoughts?

mllib/src/main/scala/org/apache/spark/ml/optim/WeightedLeastSquares.scala
```diff
@@ -85,73 +104,191 @@ private[ml] class WeightedLeastSquares(
...
-    val aa = if (fitIntercept) {
-      Array.concat(aaBar.values, aBar.values, Array(1.0))
-    } else {
-      aaBar.values
-    }
-    val ab = if (fitIntercept) {
-      Array.concat(abBar.values, Array(bBar))
+    val solver = if ((solverType == WeightedLeastSquares.Auto && elasticNetParam != 0.0 &&
+      regParam != 0.0) || (solverType == WeightedLeastSquares.QuasiNewton)) {
+      val effectiveL1RegFun: Option[(Int) => Double] = if (effectiveL1RegParam != 0.0) {
```
sethah: QuasiNewton solver needs some way to determine if it should use LBFGS or OWLQN. That logic is essentially - is the L1 param non-zero? Using an option for the L1 function here gives us an easy way of implementing that decision, without passing in the reg params themselves. 

Do you have some other suggestion?

mllib/src/main/scala/org/apache/spark/ml/optim/NormalEquationSolver.scala
yanboliang: That's a good point! I agree to keep ```fitIntercept``` as an argument of ```QuasiNewtonSolver``` constructor, only used for ```NormalEquationCostFun``` to accelerate convergence. To ```CholeskySolver``` and ```NormalEquationSolution```, we can move it out.

yanboliang:
@sethah Please see my inline comments, thanks.

mllib/src/main/scala/org/apache/spark/ml/optim/WeightedLeastSquares.scala
```diff
@@ -85,73 +104,191 @@ private[ml] class WeightedLeastSquares(
...
+        return new WeightedLeastSquaresModel(coefficients, intercept, diagInvAtWA, Array(0D))
+      } else {
+        require(!(regParam > 0.0 && standardizeLabel), "The standard deviation of the label is " +
+          "zero. Model cannot be regularized with standardization=true")
+        logWarning(s"The standard deviation of the label is zero. Consider setting " +
+          s"fitIntercept=true.")
+      }
+    }
+
+    val aBarValues = aBar.values
```
sethah: Done

mllib/src/main/scala/org/apache/spark/ml/optim/WeightedLeastSquares.scala
```diff
@@ -85,73 +104,191 @@ private[ml] class WeightedLeastSquares(
...
+      if (!standardizeLabel) {
+        lambda *= bStd
       }
-      aaValues(i) += lambda
+      aaBarValues(i) += lambda
       i += j
       j += 1
     }
+    val aa = getAtA(aaBar.values, aBar.values)
+    val ab = getAtB(abBar.values, bBarStd)
```
sethah: Not sure I understand this comment. Could you elaborate?

mllib-local/src/main/scala/org/apache/spark/ml/linalg/BLAS.scala
```diff
@@ -244,6 +244,23 @@ private[spark] object BLAS extends Serializable {
   }
 
   /**
+   * y := alpha*A*x + beta*y
+   *
+   * @param A The upper triangular part of A in a [[DenseVector]] (column major).
+   * @param x The [[DenseVector]] transformed by A.
+   * @param y The [[DenseVector]] to be modified in place.
```
sethah: Done

sethah:
@yanboliang Thanks for reviewing, I have addressed most of your comments. Let me know if you see anything else :)

yanboliang (commented):
@sethah Only left some minor comments, otherwise LGTM. Thanks.

mllib/src/main/scala/org/apache/spark/ml/optim/NormalEquationSolver.scala
```diff
@@ -0,0 +1,153 @@
...
+package org.apache.spark.ml.optim
+
+import breeze.linalg.{DenseVector => BDV}
+import breeze.optimize.{CachedDiffFunction, DiffFunction, LBFGS => BreezeLBFGS, OWLQN => BreezeOWLQN}
+import scala.collection.mutable
+
+import org.apache.spark.ml.linalg.{BLAS, DenseVector, Vectors}
+import org.apache.spark.mllib.linalg.CholeskyDecomposition
+
+private[ml] class NormalEquationSolution(
```
sethah: Done.

mllib/src/main/scala/org/apache/spark/ml/optim/NormalEquationSolver.scala
```diff
@@ -0,0 +1,153 @@
...
+
+    new NormalEquationSolution(x, Some(aaInv), None)
+  }
+}
+
+/**
+ * A class for solving the normal equations using Quasi-Newton optimization methods.
+ */
+private[ml] class QuasiNewtonSolver(
+    val fitIntercept: Boolean,
```
sethah: Done

mllib/src/main/scala/org/apache/spark/ml/optim/NormalEquationSolver.scala
```diff
@@ -0,0 +1,153 @@
...
+          dotProd += coefValues(j) * aBarValues(j)
+          j += 1
+        }
+        coefValues(numFeatures) = bBar - dotProd
+      }
+      val aax = new DenseVector(new Array[Double](numFeaturesPlusIntercept))
+      BLAS.dspmv(numFeaturesPlusIntercept, 1.0, aa, coef, 1.0, aax)
+      // loss = 1/2 (b^T W b - 2 x^T A^T W b + x^T A^T W A x)
+      val loss = 0.5 * bbBar - BLAS.dot(ab, coef) + 0.5 * BLAS.dot(coef, aax)
+      // -gradient = A^T W A x - A^T W b
```
sethah: Yep, thanks

mllib/src/main/scala/org/apache/spark/ml/optim/WeightedLeastSquares.scala
```diff
@@ -44,35 +46,52 @@ private[ml] class WeightedLeastSquaresModel(
...
  * match R's `lm`.
  * Turn on [[standardizeLabel]] to match R's `glmnet`.
  *
+ * @note The coefficients and intercept are always trained in the scaled space, but are returned
+ *       on the original scale. [[standardizeFeatures]] and [[standardizeLabel]] can be used to
+ *       control whether regularization is applied in the original space or the scaled space.
  * @param fitIntercept whether to fit intercept. If false, z is 0.0.
- * @param regParam L2 regularization parameter (lambda)
+ * @param regParam Regularization parameter (lambda).
+ * @param elasticNetParam the ElasticNet mixing parameter.
```
sethah: Done

mllib/src/main/scala/org/apache/spark/ml/optim/WeightedLeastSquares.scala
```diff
@@ -44,35 +46,52 @@ private[ml] class WeightedLeastSquaresModel(
...
  * @param standardizeFeatures whether to standardize features. If true, sigma_,,j,, is the
```
sethah: Done

mllib/src/main/scala/org/apache/spark/ml/optim/WeightedLeastSquares.scala
```diff
@@ -85,73 +104,198 @@ private[ml] class WeightedLeastSquares(
...
+    if (fitIntercept) {
+      new DenseVector(Array.concat(aaBar, aBar, Array(1.0)))
     } else {
-      (new DenseVector(x), 0.0)
+      new DenseVector(aaBar.clone())
     }
+  }
 
-    new WeightedLeastSquaresModel(coefficients, intercept, diagInvAtWA)
+  /** Construct A^T^ B from summary statistics. */
```
sethah: Done

sethah:
cc @dbtsai Would be great if you get a chance to look at this.

sethah:
jenkins retest this please

yanboliang:
Jenkins test this please.

mllib/src/test/scala/org/apache/spark/ml/optim/WeightedLeastSquaresSuite.scala
```diff
@@ -132,28 +232,256 @@ class WeightedLeastSquaresSuite extends SparkFunSuite with MLlibTestSparkContext
...
+      Vectors.dense(8.218905, 0.000000, 1.517413),
+      Vectors.dense(9.798771, 0.000000, 1.365503),
+      Vectors.dense(9.919095, 0.000000, 1.353933),
+      Vectors.dense(10.052804, 0.000000, 1.341077))
+
+    idx = 0
+    for (fitIntercept <- Seq(false, true);
+         standardization <- Seq(false, true);
+         (lambda, alpha) <- Seq((0.0, 0.0), (0.5, 0.0), (0.5, 0.5), (0.5, 1.0))) {
+      for (solver <- Seq(WeightedLeastSquares.Auto, WeightedLeastSquares.Cholesky)) {
```
yanboliang: Should the enumeration be removed? Since it's not used. This is a very minor issue, we can address it in follow up work.

yanboliang:
LGTM. I'll get this in. 
@dbtsai @mengxr We still appropriate any comments and feedback. If any, we can address them in follow up work.
@sethah Thanks for working on this.

yhuai:
This breaks the scala 2.10 build. Can you fix the problem?

```
[error] /home/jenkins/workspace/spark-master-compile-sbt-scala-2.10/mllib/src/test/scala/org/apache/spark/ml/optim/WeightedLeastSquaresSuite.scala:475: not enough arguments for constructor WeightedLeastSquares: (fitIntercept: Boolean, regParam: Double, elasticNetParam: Double, standardizeFeatures: Boolean, standardizeLabel: Boolean, solverType: org.apache.spark.ml.optim.WeightedLeastSquares.Solver, maxIter: Int, tol: Double)org.apache.spark.ml.optim.WeightedLeastSquares.
[error] Unspecified value parameter standardizeFeatures.
[error]       val wls = new WeightedLeastSquares(fitIntercept, regParam, elasticNetParam = elasticNetParam,
[error]                 ^
[error] /home/jenkins/workspace/spark-master-compile-sbt-scala-2.10/mllib/src/test/scala/org/apache/spark/ml/optim/WeightedLeastSquaresSuite.scala:533: not enough arguments for constructor WeightedLeastSquares: (fitIntercept: Boolean, regParam: Double, elasticNetParam: Double, standardizeFeatures: Boolean, standardizeLabel: Boolean, solverType: org.apache.spark.ml.optim.WeightedLeastSquares.Solver, maxIter: Int, tol: Double)org.apache.spark.ml.optim.WeightedLeastSquares.
[error] Unspecified value parameter standardizeFeatures.
[error]         val wls = new WeightedLeastSquares(fitIntercept, regParam, elasticNetParam = 0.0,
[error]                
```
https://amplab.cs.berkeley.edu/jenkins/view/Spark%20QA%20Compile/job/spark-master-compile-sbt-scala-2.10/2918/console

yanboliang:
@yhuai Sure, I will send a follow-up PR soon. Thanks for remind.

yanboliang:
Open #15625 to fix build error of Scala 2.10. Thanks.

PR #15396, opened 2016-10-07
[SPARK-14804][Spark][Graphx] Fix checkpointing of VertexRDD/EdgeRDD

tdas:

EdgeRDD/VertexRDD overrides checkpoint() and isCheckpointed() to forward these to the internal partitionRDD. So when checkpoint() is called on them, its the partitionRDD that actually gets checkpointed. However since isCheckpointed() also overridden to call partitionRDD.isCheckpointed, EdgeRDD/VertexRDD.isCheckpointed returns true even though this RDD is actually not checkpointed.

This would have been fine except the RDD's internal logic for computing the RDD depends on isCheckpointed(). So for VertexRDD/EdgeRDD, since isCheckpointed is true, when computing Spark tries to read checkpoint data of VertexRDD/EdgeRDD even though they are not actually checkpointed. Through a crazy sequence of call forwarding, it reads checkpoint data of partitionsRDD and tries to cast it to types in Vertex/EdgeRDD. This leads to ClassCastException.

The minimal fix that does not change any public behavior is to modify RDD internal to not use public override-able API for internal logic. 
## How was this patch tested?

New unit tests.

### DIFF
--- a/core/src/main/scala/org/apache/spark/rdd/RDD.scala
+++ b/core/src/main/scala/org/apache/spark/rdd/RDD.scala
@@ -1589,7 +1589,8 @@ abstract class RDD[T: ClassTag](
    * This is introduced as an alias for `isCheckpointed` to clarify the semantics of the
    * return value. Exposed for testing.
    */
-  private[spark] def isCheckpointedAndMaterialized: Boolean = isCheckpointed
+  private[spark] def isCheckpointedAndMaterialized: Boolean =
+    checkpointData.exists(_.isCheckpointed)
 
   /**
    * Return whether this RDD is marked for local checkpointing.
--- a/core/src/main/scala/org/apache/spark/rdd/RDD.scala
+++ b/core/src/main/scala/org/apache/spark/rdd/RDD.scala
@@ -1582,7 +1582,7 @@ abstract class RDD[T: ClassTag](
   /**
    * Return whether this RDD is checkpointed and materialized, either reliably or locally.
    */
-  def isCheckpointed: Boolean = checkpointData.exists(_.isCheckpointed)
+  def isCheckpointed: Boolean = isCheckpointedAndMaterialized
 
   /**
    * Return whether this RDD is checkpointed and materialized, either reliably or locally.
--- a/graphx/src/test/scala/org/apache/spark/graphx/EdgeRDDSuite.scala
+++ b/graphx/src/test/scala/org/apache/spark/graphx/EdgeRDDSuite.scala
@@ -19,6 +19,7 @@ package org.apache.spark.graphx
 
 import org.apache.spark.SparkFunSuite
 import org.apache.spark.storage.StorageLevel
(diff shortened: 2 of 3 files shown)


jkbradley:
I'm no expert on checkpointing, but the tests look fine to me.  You could eliminated duplicated code in the tests by putting the shared code in a helper method.

Tested locally to confirm unit tests fail before the fix.

rxin:
Can you reference the earlier pull requests here too?

core/src/main/scala/org/apache/spark/rdd/RDD.scala
```diff
@@ -1589,7 +1589,8 @@ abstract class RDD[T: ClassTag](
    * This is introduced as an alias for `isCheckpointed` to clarify the semantics of the
```
andrewor14: need to fix the comment here

andrewor14: oh wait I just realized this is exactly the same as `isCheckpointed`. It's kind of confusing why we have two methods that do the same thing. Is it because VertexRDD or something overrides one but not the other?

andrewor14:
Looks good. I left a suggestion that I think will make the code cleaner.

core/src/main/scala/org/apache/spark/rdd/RDD.scala
apivovarov: This method left as is in this PR - https://github.com/apache/spark/pull/15447

core/src/main/scala/org/apache/spark/rdd/RDD.scala
```diff
@@ -1589,7 +1589,8 @@ abstract class RDD[T: ClassTag](
...
    * return value. Exposed for testing.
    */
-  private[spark] def isCheckpointedAndMaterialized: Boolean = isCheckpointed
+  private[spark] def isCheckpointedAndMaterialized: Boolean =
+    checkpointData.exists(_.isCheckpointed)
```
andrewor14: No, I meant:
```
private[spark] final def isCheckpointedAndMaterialized: Boolean = {
  checkpointData.exists(_.isCheckpointed)
}

def isCheckpointed: Boolean = isCheckpointedAndMaterialized
```

neggert:
Any update on this? Would love to see it in an upcoming release.

core/src/main/scala/org/apache/spark/rdd/RDD.scala
tdas: Addressed it. Can you check again?

tdas:
I am merging this to master and 2.1

core/src/main/scala/org/apache/spark/rdd/RDD.scala
andrewor14: looks good, but you forgot to make this final

PR #15397, opened 2016-10-07
[SPARK-17834][SQL]Fetch the earliest offsets manually in KafkaSource instead of counting on KafkaConsumer

zsxwing:

Because `KafkaConsumer.poll(0)` may update the partition offsets, this PR just calls `seekToBeginning` to manually set the earliest offsets for the KafkaSource initial offsets.
## How was this patch tested?

Existing tests.

### DIFF
--- a/external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/KafkaSource.scala
+++ b/external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/KafkaSource.scala
@@ -235,18 +235,15 @@ private[kafka010] case class KafkaSource(
    * Fetch the earliest offsets of partitions.
    */
   private def fetchEarliestOffsets(): Map[TopicPartition, Long] = withRetriesWithoutInterrupt {
-    // Make sure `KafkaConsumer.poll` won't be interrupted (KAFKA-1894)
-    assert(Thread.currentThread().isInstanceOf[StreamExecutionThread])
     // Poll to get the latest assigned partitions
     consumer.poll(0)
     val partitions = consumer.assignment()
     consumer.pause(partitions)
-    logDebug(s"Partitioned assigned to consumer: $partitions")
+    logDebug(s"Partitions assigned to consumer: $partitions. Seeking to the beginning")
 
-    logDebug("Seeked to the beginning")
     consumer.seekToBeginning(partitions)
     val partitionOffsets = partitions.asScala.map(p => p -> consumer.position(p)).toMap
-    logDebug(s"Got offsets for partition : $partitionOffsets")
+    logDebug(s"Got earliest offsets for partition : $partitionOffsets")
     partitionOffsets
   }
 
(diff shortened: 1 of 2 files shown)


koeninger:
How is this going to work with assign?  It seems like it's just avoiding the problem, not fixing it.

zsxwing:
> How is this going to work with assign? It seems like it's just avoiding the problem, not fixing it.

We can seek to the offsets provided by the user.

koeninger:
Look at the poll/seek implementation in the DStream's subscribe and
subscribe pattern when user offsets are provided, i.e. the problem that
triggered this ticket to begin with. You're going to have to solve the same
problem there with the structured stream, unless the structured stream
somehow wants to limit assigning specific partitions only to the assign
strategy, which eliminates lots of valid use cases.

zsxwing:
@koeninger sorry for the delay. Right now KafkaSource doesn't support external group id, so we don't need to concern about how to fetching committed offsets. Any other cases that I'm missing?

koeninger:
My main point is that whoever implements SPARK-17812 is going to have to deal with the issue shown in SPARK-17782, which means much of this patch is going to need to be changed anyway.

But It's not just about external group id.  Committed offsets would actually make the issue in SPARK-17782 less of a problem, because they would take precedence over auto.offset.reset

zsxwing:
> My main point is that whoever implements SPARK-17812 is going to have to deal with the issue shown in SPARK-17782, which means much of this patch is going to need to be changed anyway.

@koeninger I agreed that this patch will be changed. However, this PR does fix a known issue for the current supported features and there is not user facing changes. Considering 2.0.2 may come out soon and I don't think SPARK-17812 will be done soon, I would like to merge this to fix issues for 2.0.2. What do you think?

external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/KafkaSource.scala
```diff
@@ -256,8 +269,6 @@ private[kafka010] case class KafkaSource(
    */
   private def fetchNewPartitionEarliestOffsets(
       newPartitions: Seq[TopicPartition]): Map[TopicPartition, Long] = withRetriesWithoutInterrupt {
-    // Make sure `KafkaConsumer.poll` won't be interrupted (KAFKA-1894)
-    assert(Thread.currentThread().isInstanceOf[StreamExecutionThread])
     // Poll to get the latest assigned partitions
     consumer.poll(0)
     val partitions = consumer.assignment()
```
koeninger: Is there a reason not to pause all partitions here as well?

external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/KafkaSource.scala
```diff
@@ -270,7 +281,7 @@ private[kafka010] case class KafkaSource(
       // So we need to ignore them
       partitions.contains(p)
     }.map(p => p -> consumer.position(p)).toMap
-    logDebug(s"Got offsets for new partitions: $partitionToOffsets")
+    logDebug(s"Got earliest offsets for new partitions: $partitionToOffsets")
     partitionToOffsets
```
koeninger: nit: different variable name partitionToOffsets vs partitionOffsets for what is essentially the same thing

koeninger (commented):
Did another once over, couple more minor things.

If the plan is to wait for SPARK-17812 to fix up the other stuff I was concerned about, that's ok with me, but I really hope it doesn't slip past another release.  To reiterate, I'm fine with doing that work.

koeninger:
LGTM, thanks for talking it through

zsxwing:

PR #15398, opened 2016-10-08
[SPARK-17647][SQL] Fix backslash escaping in 'LIKE' patterns.

jodersky:

This patch fixes a bug in the way LIKE patterns are translated to Java regexes. The bug causes any character following an escaped backslash to be escaped, i.e. there is double-escaping.
A concrete example is the following pattern:`'%\\%'`. The expected Java regex that this pattern should correspond to (according to the behavior described below) is `'.*\\.*'`, however the current situation leads to `'.*\\%'` instead.

---

Update: in light of the discussion that ensued, we should explicitly define the expected behaviour of LIKE expressions, especially in certain edge cases. With the help of @gatorsmile, we put together a list of different RDBMS and their variations wrt to certain standard features.

| RDBMS\Features | Wildcards | Default escape [1] | Case sensitivity |
| --- | --- | --- | --- |
| [MS SQL Server](https://msdn.microsoft.com/en-us/library/ms179859.aspx) | _, %, [], [^] | none | no |
| [Oracle](https://docs.oracle.com/cd/B12037_01/server.101/b10759/conditions016.htm) | _, % | none | yes |
| [DB2 z/OS](http://www.ibm.com/support/knowledgecenter/SSEPEK_11.0.0/sqlref/src/tpc/db2z_likepredicate.html) | _, % | none | yes |
| [MySQL](http://dev.mysql.com/doc/refman/5.7/en/string-comparison-functions.html) | _, % | none | no |
| [PostreSQL](https://www.postgresql.org/docs/9.0/static/functions-matching.html) | _, % | \ | yes |
| [Hive](https://cwiki.apache.org/confluence/display/Hive/LanguageManual+UDF) | _, % | none | yes |
| Current Spark | _, % | \ | yes |

[1] Default escape character: most systems do not have a default escape character, instead the user can specify one by calling a like expression with an escape argument [A] LIKE [B] ESCAPE [C]. This syntax is currently not supported by Spark, however I would volunteer to implement this feature in a separate ticket.

The specifications are often quite terse and certain scenarios are undocumented, so here is a list of scenarios that I am uncertain about and would appreciate any input. Specifically I am looking for feedback on whether or not Spark's current behavior should be changed.
1. [x] Ending a pattern with the escape sequence, e.g. `like 'a\'`.
   PostreSQL gives an error: 'LIKE pattern must not end with escape character', which I personally find logical. Currently, Spark allows "non-terminated" escapes and simply ignores them as part of the pattern.
   According to [DB2's documentation](http://www.ibm.com/support/knowledgecenter/SSEPGG_9.7.0/com.ibm.db2.luw.messages.sql.doc/doc/msql00130n.html), ending a pattern in an escape character is invalid.
   _Proposed new behaviour in Spark: throw AnalysisException_
2. [x] Empty input, e.g. `'' like ''`
   Postgres and DB2 will match empty input only if the pattern is empty as well, any other combination of empty input will not match. Spark currently follows this rule.
3. [x] Escape before a non-special character, e.g. `'a' like '\a'`.
   Escaping a non-wildcard character is not really documented but PostgreSQL just treats it verbatim, which I also find the least surprising behavior. Spark does the same.
   According to [DB2's documentation](http://www.ibm.com/support/knowledgecenter/SSEPGG_9.7.0/com.ibm.db2.luw.messages.sql.doc/doc/msql00130n.html), it is invalid to follow an escape character with anything other than an escape character, an underscore or a percent sign.
   _Proposed new behaviour in Spark: throw AnalysisException_

The current specification is also described in the operator's source code in this patch.
## How was this patch tested?

Extra case in regex unit tests.

### DIFF
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/regexpExpressions.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/regexpExpressions.scala
@@ -70,18 +70,28 @@ trait StringRegexExpression extends ImplicitCastInputTypes {
 @ExpressionDescription(
   usage = "str _FUNC_ pattern - Returns true if str matches pattern, " +
     "null if any arguments are null, false otherwise.",
-  extended =
-    "The pattern is a string which is matched literally, with exception to the following " +
-    "special symbols:\n\n" +
-    "    _ matches any one character in the input (similar to . in posix " +
-    "regular expressions)\n\n" +
-    "    % matches zero ore more characters in the input (similar to .* in " +
-    "posix regular expressions)\n\n" +
-    "The escape character is '\\'. If an escape character precedes a special symbol or " +
-    "another escape character, the following character is matched literally, For example, " +
-    "the expression `<path> like \\%SystemDrive\\%\\\\Users%` will match any `<path>` that " +
-    "starts with '%SystemDrive%\\Users'. It is invalid to escape any other character.\n\n" +
-    "Use RLIKE to match with standard regular expressions.")
+  extended = """
+    Arguments:
+      str - a string expression
+      pattern - a string expression. The pattern is a string which is matched literally, with
+        exception to the following special symbols:
+
+          _ matches any one character in the input (similar to . in posix regular expressions)
+
+          % matches zero ore more characters in the input (similar to .* in posix regular
+          expressions)
+
+        The escape character is '\'. If an escape character precedes a special symbol or another
+        escape character, the following character is matched literally. It is invalid to escape
+        any other character.
+
+    Examples:
+      > SELECT '%SystemDrive%\Users\John' _FUNC_ '\%SystemDrive\%\\Users%'
+      true
+
+    See also:
+      Use RLIKE to match with standard regular expressions.
+""")
 case class Like(left: Expression, right: Expression)
   extends BinaryExpression with StringRegexExpression {
 
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/regexpExpressions.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/regexpExpressions.scala
@@ -70,20 +70,18 @@ trait StringRegexExpression extends ImplicitCastInputTypes {
 @ExpressionDescription(
   usage = "str _FUNC_ pattern - Returns true if str matches pattern, " +
     "null if any arguments are null, false otherwise.",
-  extended = """
-    |The pattern is a string which is matched literally, with exception to the
-    |following symbols:
-    |
-    |    _ matches any one character in the input (similar to . in posix
-    |    regular expressions)
-    |
-    |    % matches zero ore more characters in the input (similar to .* in
-    |    posix regular expressions)
-    |
-    |The default escape character is '\\'. Any character after the escape
-    |character will be matched against literally.
-    |
-    |Use RLIKE to match with standard regular expressions.""")
+  extended =
+    "The pattern is a string which is matched literally, with exception to the " +
+    "following special symbols:\n\n" +
+    "    _ matches any one character in the input (similar to . in posix " +
+    "regular expressions)\n\n" +
+    "    % matches zero ore more characters in the input (similar to .* in " +
+    "posix regular expressions\n\n" +
+    "The default escape character is '\\'. If an escape character precedes a special symbol or " +
+    "another escape character, the following character is matched literally, otherwise the " +
+     "escape character is treated literally. I.e. '\\%' would match '%', whereas '\\a' matches " +
+     "'\\a'.\n\n" +
+    "Use RLIKE to match with standard regular expressions.")
 case class Like(left: Expression, right: Expression)
   extends BinaryExpression with StringRegexExpression {
 
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/regexpExpressions.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/regexpExpressions.scala
@@ -71,16 +71,16 @@ trait StringRegexExpression extends ImplicitCastInputTypes {
   usage = "str _FUNC_ pattern - Returns true if str matches pattern, " +
     "null if any arguments are null, false otherwise.",
   extended =
-    "The pattern is a string which is matched literally, with exception to the " +
-    "following special symbols:\n\n" +
+    "The pattern is a string which is matched literally, with exception to the following " +
+    "special symbols:\n\n" +
     "    _ matches any one character in the input (similar to . in posix " +
     "regular expressions)\n\n" +
     "    % matches zero ore more characters in the input (similar to .* in " +
-    "posix regular expressions\n\n" +
-    "The default escape character is '\\'. If an escape character precedes a special symbol or " +
-    "another escape character, the following character is matched literally, otherwise the " +
-     "escape character is treated literally. I.e. '\\%' would match '%', whereas '\\a' matches " +
-     "'\\a'.\n\n" +
+    "posix regular expressions)\n\n" +
+    "The escape character is '\\'. If an escape character precedes a special symbol or " +
+    "another escape character, the following character is matched literally, For example, " +
+    "the expression `<path> like \\%SystemDrive\\%\\\\Users%` will match any `<path>` that " +
+    "starts with '%SystemDrive%\\Users'. It is invalid to escape any other character.\n\n" +
     "Use RLIKE to match with standard regular expressions.")
 case class Like(left: Expression, right: Expression)
   extends BinaryExpression with StringRegexExpression {
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/regexpExpressions.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/regexpExpressions.scala
@@ -68,7 +68,22 @@ trait StringRegexExpression extends ImplicitCastInputTypes {
  * Simple RegEx pattern matching function
  */
 @ExpressionDescription(
-  usage = "str _FUNC_ pattern - Returns true if `str` matches `pattern`, or false otherwise.")
+  usage = "str _FUNC_ pattern - Returns true if str matches pattern, " +
+    "null if any arguments are null, false otherwise.",
+  extended = """
+    |The pattern is a string which is matched literally, with exception to the
+    |following symbols:
+    |
+    |    _ matches any one character in the input (similar to . in posix
+    |    regular expressions)
+    |
+    |    % matches zero ore more characters in the input (similar to .* in
+    |    posix regular expressions)
+    |
+    |The default escape character is '\\'. Any character after the escape
+    |character will be matched against literally.
+    |
+    |Use RLIKE to match with standard regular expressions.""")
 case class Like(left: Expression, right: Expression)
   extends BinaryExpression with StringRegexExpression {
 
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/StringUtils.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/StringUtils.scala
@@ -27,19 +27,20 @@ object StringUtils {
   // replace the % with .*, match 0 or more times with any character
   def escapeLikeRegex(str: String): String = {
     val builder = new StringBuilder()
-    str.foldLeft(false) { case (escaping, next) =>
+    var escaping = false
+    for (next <- str) {
       if (escaping) {
         builder ++= Pattern.quote(Character.toString(next))
-        false
+        escaping = false
       } else if (next == '\\') {
-        true
+        escaping = true
       } else {
         builder ++= (next match {
           case '_' => "."
           case '%' => ".*"
           case _ => Pattern.quote(Character.toString(next))
         })
-        false
+        escaping = false
       }
     }
     "(?s)" + builder.result() // (?s) enables dotall mode, causing "." to match new lines
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/StringUtils.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/StringUtils.scala
@@ -19,6 +19,7 @@ package org.apache.spark.sql.catalyst.util
 
 import java.util.regex.{Pattern, PatternSyntaxException}
 
+import org.apache.spark.sql.AnalysisException
 import org.apache.spark.unsafe.types.UTF8String
 
 object StringUtils {
@@ -28,16 +29,18 @@ object StringUtils {
     val in = str.toIterator
     val out = new StringBuilder()
 
+    def fail(message: String) = throw new AnalysisException(
+      s"the pattern '$str' is invalid, $message")
+
     while (in.hasNext) {
       in.next match {
         case '\\' if in.hasNext =>
-          in.next match {
-            case '\\' => out ++= Pattern.quote("\\")
-            case '_' => out ++= Pattern.quote("_")
-            case '%' => out ++= Pattern.quote("%")
-            // escape before non-escapable character treated literally
-            case c => out ++= Pattern.quote("\\" + c)
+          val c = in.next
+          c match {
+            case '_' | '%' | '\\' => out ++= Pattern.quote(Character.toString(c))
+            case _ => fail(s"the escape character is not allowed to precede '$c'")
           }
+        case '\\' => fail("it is not allowed to end with the escape character")
         case '_' => out ++= "."
         case '%' => out ++= ".*"
         case c => out ++= Pattern.quote(Character.toString(c))
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/StringUtils.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/StringUtils.scala
@@ -25,26 +25,24 @@ object StringUtils {
 
   // replace the _ with .{1} exactly match 1 time of any character
   // replace the % with .*, match 0 or more times with any character
-  def escapeLikeRegex(v: String): String = {
-    if (!v.isEmpty) {
-      "(?s)" + (' ' +: v.init).zip(v).flatMap {
-        case (prev, '\\') => ""
-        case ('\\', c) =>
-          c match {
-            case '_' => "_"
-            case '%' => "%"
-            case _ => Pattern.quote("\\" + c)
-          }
-        case (prev, c) =>
-          c match {
-            case '_' => "."
-            case '%' => ".*"
-            case _ => Pattern.quote(Character.toString(c))
-          }
-      }.mkString
-    } else {
-      v
+  def escapeLikeRegex(str: String): String = {
+    val builder = new StringBuilder()
+    str.foldLeft(false) { case (escaping, next) =>
+      if (escaping) {
+        builder ++= Pattern.quote(Character.toString(next))
+        false
+      } else if (next == '\\') {
+        true
+      } else {
+        builder ++= (next match {
+          case '_' => "."
+          case '%' => ".*"
+          case _ => Pattern.quote(Character.toString(next))
+        })
+        false
+      }
     }
+    "(?s)" + builder.result() // (?s) enables dotall mode, causing "." to match new lines
   }
 
   private[this] val trueStrings = Set("t", "true", "y", "yes", "1").map(UTF8String.fromString)
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/StringUtils.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/StringUtils.scala
@@ -24,13 +24,23 @@ import org.apache.spark.unsafe.types.UTF8String
 
 object StringUtils {
 
-  /** Convert 'like' pattern to Java regex. */
-  def escapeLikeRegex(str: String): String = {
-    val in = str.toIterator
+  /**
+   * Validate and convert SQL 'like' pattern to a Java regular expression.
+   *
+   * Underscores (_) are converted to '.' and percent signs (%) are converted to '.*', other
+   * characters are quoted literally. Escaping is done according to the rules specified in
+   * [[org.apache.spark.sql.catalyst.expressions.Like]] usage documentation. An invalid pattern will
+   * throw an [[AnalysisException]].
+   *
+   * @param pattern the SQL pattern to convert
+   * @return the equivalent Java regular expression of the pattern
+   */
+  def escapeLikeRegex(pattern: String): String = {
+    val in = pattern.toIterator
     val out = new StringBuilder()
 
     def fail(message: String) = throw new AnalysisException(
-      s"the pattern '$str' is invalid, $message")
+      s"the pattern '$pattern' is invalid, $message")
 
     while (in.hasNext) {
       in.next match {
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/StringUtils.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/StringUtils.scala
@@ -23,27 +23,27 @@ import org.apache.spark.unsafe.types.UTF8String
 
 object StringUtils {
 
-  // replace the _ with .{1} exactly match 1 time of any character
-  // replace the % with .*, match 0 or more times with any character
+  /** Convert 'like' pattern to Java regex. */
   def escapeLikeRegex(str: String): String = {
-    val builder = new StringBuilder()
-    var escaping = false
-    for (next <- str) {
-      if (escaping) {
-        builder ++= Pattern.quote(Character.toString(next))
-        escaping = false
-      } else if (next == '\\') {
-        escaping = true
-      } else {
-        builder ++= (next match {
-          case '_' => "."
-          case '%' => ".*"
-          case _ => Pattern.quote(Character.toString(next))
-        })
-        escaping = false
+    val in = str.toIterator
+    val out = new StringBuilder()
+
+    while (in.hasNext) {
+      in.next match {
+        case '\\' if in.hasNext =>
+          in.next match {
+            case '\\' => out ++= Pattern.quote("\\")
+            case '_' => out ++= Pattern.quote("_")
+            case '%' => out ++= Pattern.quote("%")
+            // escape before non-escapable character treated literally
+            case c => out ++= Pattern.quote("\\" + c)
+          }
+        case '_' => out ++= "."
+        case '%' => out ++= ".*"
+        case c => out ++= Pattern.quote(Character.toString(c))
       }
     }
-    "(?s)" + builder.result() // (?s) enables dotall mode, causing "." to match new lines
+    "(?s)" + out.result() // (?s) enables dotall mode, causing "." to match new lines
   }
 
   private[this] val trueStrings = Set("t", "true", "y", "yes", "1").map(UTF8String.fromString)
--- a/sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/RegexpExpressionsSuite.scala
+++ b/sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/RegexpExpressionsSuite.scala
@@ -62,7 +62,6 @@ class RegexpExpressionsSuite extends SparkFunSuite with ExpressionEvalHelper {
     checkEvaluation("a" like "", false)
     checkEvaluation("" like "a", false)
 
-
     // SI-17647 double-escaping backslash
     checkEvaluation("""\\\\""" like """%\\%""", true) // triple quotes to avoid java string escaping
     checkEvaluation("""%%""" like """%%""", true)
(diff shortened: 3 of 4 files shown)


jodersky:
I'll add some more tests

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/StringUtils.scala
```diff
@@ -25,26 +25,24 @@ object StringUtils {
...
-            case '_' => "."
-            case '%' => ".*"
-            case _ => Pattern.quote(Character.toString(c))
-          }
-      }.mkString
-    } else {
-      v
+  def escapeLikeRegex(str: String): String = {
+    val builder = new StringBuilder()
+    str.foldLeft(false) { case (escaping, next) =>
```
rxin: can we get rid of the fold here? The fold makes it more difficult to parse (e.g. what's the return type, what's escaping, what's next).

jodersky: updated in latest commit

jodersky:
Every non-java-pattern character is quoted now, updated the StringUtils test suite

jodersky:
cc @viirya, I changed your original code

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/StringUtils.scala
```diff
@@ -25,26 +25,25 @@ object StringUtils {
...
-          }
-      }.mkString
-    } else {
-      v
+  def escapeLikeRegex(str: String): String = {
+    val builder = new StringBuilder()
+    var escaping = false
+    for (next <- str) {
+      if (escaping) {
+        builder ++= Pattern.quote(Character.toString(next))
```
viirya: How about `"\\a"`? Previously it is `\Q\a\E`, now it seems becoming `\Qa\E`.

gatorsmile:
Also cc @yhuai and @JoshRosen @mengxr Please check whether the changes here can satisfy what you want. Thanks!

yhuai:
@jodersky Thank you for the patch! How about we add a summary related to the behavior of escape in pr description and jira?

yhuai:
I think we should consider other databases behavior and see if our behavior makes sense.

gatorsmile:
I see. Agree. Will do an investigation with @jodersky Thanks!

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/StringUtils.scala
jodersky: Every character after a backslash is quoted, so `\\a` becomes `\Q\\E\Qa\E`. Is this not the intended behaviour?

viirya: `\Q\\E\Qa\E` is correct. But doesn't it become `\Qa\E` in this change?

For `\\a`, the prefixing `\\` will go the next branch and enable `escaping`. Then the next char `a` will be quoted here. So it becomes `\Qa\E`. BTW, before this change, it will be `\Q\a\E`.

jodersky:
thanks @yhuai for the feedback! I enumerated some issues in the pull request description, do you have any input on them? My personal opinion is that we should change the regex matching to conform to point 1.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/StringUtils.scala
jodersky: I think we're misunderstanding each other, there are so many different contexts in which a backslash is used :)

The prefix `'\\'` that matches the escape char here is the *scala source* escape for the literal character `'\`' (If scala had triple quotes for chars as it does for strings, I could have written `'''\'''`). So, in the if statement `\\` will consume one character, a `\`.
The subsequent `\` will be quoted in the `escaping` case, and the final `a` will also be quoted in the last case, resulting in a `\Q\\E\Qa\E`.

jodersky:
The style issues are weird. Running scalastyle in sbt worked fine...

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/RegexpExpressionsSuite.scala
```diff
@@ -54,20 +57,32 @@ class RegexpExpressionsSuite extends SparkFunSuite with ExpressionEvalHelper {
...
     checkEvaluation("""\__""" like """\\\__""", true)
-
     checkEvaluation("""\\\__""" like """%\\%\%""", false)
-    checkEvaluation("""\_""" like """\_\_""", false)
     checkEvaluation("""_\\\%""" like """%\\""", false)
-    checkEvaluation("""_\\__""" like """_\___""", false)
+
+    // unicode
+    checkEvaluation("a\u20ACa" like "_\u20AC_", true)
+    checkEvaluation("a€a" like "_€_", true)
```
gatorsmile: // scalastyle:off
// non ascii characters are not allowed in the source code, so we disable the scalastyle.
checkEvaluation("a€a" like "_€_", true)
...
// scalastyle:on

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/RegexpExpressionsSuite.scala
```diff
@@ -90,6 +105,28 @@ class RegexpExpressionsSuite extends SparkFunSuite with ExpressionEvalHelper {
...
+    checkEvaluation("""\__""" like regEx, true, create_row("""\\\__"""))
+    checkEvaluation("""\\\__""" like regEx, false, create_row("""%\\%\%"""))
+    checkEvaluation("""_\\\%""" like regEx, false, create_row("""%\\"""))
+
+    checkEvaluation("a\u20ACa" like regEx, true, create_row("_\u20AC_"))
+    checkEvaluation("a€a" like regEx, true, create_row("_€_"))
+    checkEvaluation("a€a" like regEx, true, create_row("_\u20AC_"))
+    checkEvaluation("a\u20ACa" like regEx, true, create_row("_€_"))
+
+    checkEvaluation("""a\""" like regEx, false, create_row("""a\"""))  // TODO: should throw an exception?
```
gatorsmile: more than 100 characters.

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/RegexpExpressionsSuite.scala
```diff
@@ -90,6 +105,28 @@ class RegexpExpressionsSuite extends SparkFunSuite with ExpressionEvalHelper {
...
+
+    checkEvaluation("A" like regEx, false, create_row("a%"))
+    checkEvaluation("a" like regEx, false, create_row("A%"))
+    checkEvaluation("AaA" like regEx, true, create_row("_a_"))
+
```
gatorsmile: remove this empty line

jodersky: yeah, I just saw those.  I turns out that `scalastyle` is a scoped task, so you have to run `test:scalastyle` in sbt to make sure everything gets checked.

gatorsmile:
As @jodersky said, SQL2003 syntax is like
```
WHERE expression [NOT] LIKE string_pattern [ESCAPE escape_sequence]
```
```
- ESCAPE escape_sequence
Allows you to search for the presence of characters that would normally be interpreted as wildcards.
```

@yhuai @rxin Should we allow users to specify the ESCAPE characters? Thanks!

rxin:
Sounds like a good idea. It doesn't need to be in this PR though ...

gatorsmile:
@jodersky Please follow @rxin 's suggestion and submit a separate PR to support it. Thanks!

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/RegexpExpressionsSuite.scala
```diff
@@ -74,6 +107,31 @@ class RegexpExpressionsSuite extends SparkFunSuite with ExpressionEvalHelper {
...
+    checkEvaluation("""_\\\%""" like regEx, false, create_row("""%\\"""))
+
+    // scalastyle:off nonascii
+    checkEvaluation("a\u20ACa" like regEx, true, create_row("_\u20AC_"))
+    checkEvaluation("a€a" like regEx, true, create_row("_€_"))
+    checkEvaluation("a€a" like regEx, true, create_row("_\u20AC_"))
+    checkEvaluation("a\u20ACa" like regEx, true, create_row("_€_"))
+    // scalastyle:on nonascii
+
+    // TODO: should throw an exception?
```
gatorsmile: See the description of SQL -130

```
SQL0130N
The ESCAPE clause is not a single character, or the pattern string contains an invalid occurrence of the escape character.
Explanation
The escape character must be a single character no more than two bytes in length. It can only appear in the pattern string if it is followed by itself, a percent sign, or an underscore. For more information about the ESCAPE clause on the LIKE predicate, refer to the SQL Reference.

User response
Correct the pattern string or the escape character accordingly.
```

jodersky:
Thanks for the feedback. It was my plan to add the escape syntax in a separate PR. The main questions that I'm concerned with here are the points I enumerate in this PRs description, namely:

- should spark allow ending patterns in the escape character?
- how to handle empty input and/or patterns?
- escaping "regular" characters?

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/RegexpExpressionsSuite.scala
jodersky: > The escape character must be a single character no more than two bytes in length. It can only appear in the pattern string if it is followed by itself, a percent sign, or an underscore.

does this mean that PostgreSQL's behaviour is wrong? Should spark throw an exception in point 2?

gatorsmile:
Also cc @yhuai , @JoshRosen and @mengxr

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/RegexpExpressionsSuite.scala
gatorsmile: To answer your point 3, I did a try in DB2. 
```
db2 => select actkwd from act where actkwd like '%A%\a' escape '\'
SQL0130N  The ESCAPE clause is not a single character, or the pattern string 
contains an invalid occurrence of the escape character.  SQLSTATE=22025
```

In DB2, normally, our design is very conservative. If we think this could be a user error, we will stop it with an error. We do not want to give users any surprise.

viirya:
For escaping before a non-special character, I don't know if DB2 is special. Because as I try, MySQL behaving like PostgreSQL.

gatorsmile:
That is a design decision we need to make. Personally, MySQL and PostgreSQL are not good examples we should follow. Let us ask ourselves a simple question: 
> When users input `'\a'`, what is their expected results?

gatorsmile:
Also checked SQL Server. It behaves differently from both `Postgre` and `DB2`. It simply ignores it. 

> If there is no character after an escape character in the LIKE pattern, the pattern is not valid and the LIKE returns FALSE. If the character after an escape character is not a wildcard character, the escape character is discarded and the character following the escape is treated as a regular character in the pattern. This includes the percent sign (%), underscore (_), and left bracket ([) wildcard characters when they are enclosed in double brackets ([ ]). Also, within the double bracket characters ([ ]), escape characters can be used and the caret (^), hyphen (-), and right bracket (]) can be escaped.

Let me check Oracle now

viirya:
> If the character after an escape character is not a wildcard character, the escape character is discarded and the character following the escape is treated as a regular character in the pattern.

@gatorsmile Looks like it is the same as PostgreSQL?

gatorsmile:
This is different. 
> If there is no character after an escape character in the LIKE pattern, the pattern is not valid and the LIKE returns FALSE.

gatorsmile:
Checked Oracle. Oracle has the same behavior like DB2:
```
SQL> SELECT USERNAME FROM ALL_USERS WHERE USERNAME LIKE '%P%\' ESCAPE '\';
SELECT USERNAME FROM ALL_USERS WHERE USERNAME LIKE '%P%\' ESCAPE '\'
                                                   *
ERROR at line 1:
ORA-01424: missing or illegal character following the escape character


SQL> SELECT USERNAME FROM ALL_USERS WHERE USERNAME LIKE '%P%\a' ESCAPE '\';
SELECT USERNAME FROM ALL_USERS WHERE USERNAME LIKE '%P%\a' ESCAPE '\'
                                                   *
ERROR at line 1:
ORA-01424: missing or illegal character following the escape character
```

gatorsmile:
```
ORA-01424: missing or illegal character following the escape character
Cause:	The character following the escape character in LIKE pattern is missing or not one of the escape character, '%', or '_'.
Action:	Remove the escape character or specify the missing character.
```

viirya:
@gatorsmile That is for ending a pattern with the escape sequence. I mean escaping before a non-special character.

gatorsmile:
We need to answer all the three questions listed by @jodersky

viirya:
Maybe more important is, how Hive performs `like`. For escaping before a non-special character, loos like it is different to above examples. If you gives pattern like `\a`, it matches exactly `\a`, not `a`:

https://github.com/apache/hive/blob/ff67cdda1c538dc65087878eeba3e165cf3230f4/ql/src/java/org/apache/hadoop/hive/ql/udf/UDFLike.java#L64

For ending a pattern with the escape character, looks like it allows this usage too. It simply see the escape character as normal character.

gatorsmile:
If you check what we did in the other PRs, we do not need to strictly follow Hive. I do not have a strong opinion in all the above options. Let @rxin @yhuai make a decision.

rxin:
In the case that there is no clear standard (e.g. all databases diverge), I'd say it's easier to just follow Hive's, unless it really doesn't make sense or is extremely difficult to support.

jodersky:
Agreed, I'll update the regex parsing to copy Hive's behaviour

mengxr (commented):
I only checked the doc and did some quick tests. Is "Escape before a non-special character" a HIVE documented behavior or an implementation glitch? This leads to inconsistent behavior among SQL and Scala/Java/Python. For example,

``` sql
SELECT '\a\t' LIKE '\a\t'
```

returns true, while this is illegal string in Scala/Python. Note that SQL doesn't treat strings as verbatim (as triple quoted in Scala). `\t` still means a tab. I like @gatorsmile 's question:

> When users input '\a' in SQL LIKE statements, what is their expected results?

My guess is that `a` is very unlikely what they expected.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/regexpExpressions.scala
```diff
@@ -68,7 +68,20 @@ trait StringRegexExpression extends ImplicitCastInputTypes {
  * Simple RegEx pattern matching function
  */
 @ExpressionDescription(
-  usage = "str _FUNC_ pattern - Returns true if str matches pattern and false otherwise.")
+  usage = "str _FUNC_ pattern - Returns true if str matches pattern, " +
+    "null if any arguments are null, false otherwise.",
+  extended =
+    "The pattern is a string which is matched literally, with exception to the " +
```
jodersky: Triple quotes would help in the source, however if want correct indentation I would need to use pipes and call `.stripMargin` on the string. That is unfortunately not possible since the string would then not be considered a constant anymore (constants are required for annotation parameters)

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/regexpExpressions.scala
```diff
@@ -68,7 +68,20 @@ trait StringRegexExpression extends ImplicitCastInputTypes {
...
+    "following special symbols:\n\n" +
+    "    _ matches any one character in the input (similar to . in posix " +
+    "regular expressions)\n\n" +
+    "    % matches zero ore more characters in the input (similar to .* in " +
+    "posix regular expressions\n\n" +
+    "The default escape character is '\\'. If an escape character precedes a special symbol or " +
```
jodersky: ack

jodersky:
I understand your comment about the weird escaping behaviour, @mengxr. Putting myself into the shoes of a new user, I would be least surprised if Spark were to treat the String verbatim (as in "anything after a backslash is treated literally") or give me an error and tell me that my escape sequence doesn't make sense. Changing the parsing of the escape character based on what follows also makes the code more complicated.

However, I also understand the high priority of making migration from Hive to Spark SQL most straightforward. Therefore, I concluded that existing queries should behave the same, and reimplemented Hive's 'like' pattern matching. Maybe this wasn't the right trade-off for non Hive users?

I was unable to find any docs that detail the escaping behaviour in the scenarios mentioned above, however I basically followed the implementation thanks to the [link Simon provided](https://github.com/apache/hive/blob/ff67cdda1c538dc65087878eeba3e165cf3230f4/ql/src/java/org/apache/hadoop/hive/ql/udf/UDFLike.java#L64).

@rxin, do you think this case warrants diverging from Hive?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/regexpExpressions.scala
```diff
@@ -68,7 +68,20 @@ trait StringRegexExpression extends ImplicitCastInputTypes {
...
+    "another escape character, the following character is matched literally, otherwise the " +
+     "escape character is treated literally. I.e. '\\%' would match '%', whereas '\\a' matches " +
+     "'\\a'.\n\n" +
```
jodersky: good idea

jodersky:
FWIW, I just read this about the standard
> In an ANSI-compliant database, you can only use the LIKE escape character to escape a percent sign ( % ), an underscore ( _ ), or the escape character itself.

[source](https://www.ibm.com/support/knowledgecenter/en/SSGU8G_12.1.0/com.ibm.sqls.doc/ids_sqs_1388.htm). I think we should reconsider the need to follow Hive here

mengxr:
Thanks for checking the standard! I think that behavior is what most people expect:)

jodersky:
I went ahead and changed the parsing function to throw an AnalysisException in case the input escaper sequence is invalid. Feel free to revert (or just tell me and I'll do it) if we should stick to Hive's behaviour

Also, I hope an `AnalysisException` is the correct way to report an error here, let me know if it should be something else.

viirya:
@jodersky I think the current approach is fine if we have documented it clear.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/StringUtils.scala
```diff
@@ -19,32 +19,34 @@ package org.apache.spark.sql.catalyst.util
...
-      "(?s)" + (' ' +: v.init).zip(v).flatMap {
-        case (prev, '\\') => ""
-        case ('\\', c) =>
-          c match {
-            case '_' => "_"
-            case '%' => "%"
-            case _ => Pattern.quote("\\" + c)
-          }
-        case (prev, c) =>
+  /** Convert 'like' pattern to Java regex. */
```
gatorsmile: Please improve this to explain the behavior of this function.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/regexpExpressions.scala
```diff
@@ -68,7 +68,20 @@ trait StringRegexExpression extends ImplicitCastInputTypes {
...
+    "special symbols:\n\n" +
+    "    _ matches any one character in the input (similar to . in posix " +
+    "regular expressions)\n\n" +
+    "    % matches zero ore more characters in the input (similar to .* in " +
+    "posix regular expressions)\n\n" +
+    "The escape character is '\\'. If an escape character precedes a special symbol or " +
+    "another escape character, the following character is matched literally, For example, " +
+    "the expression `<path> like \\%SystemDrive\\%\\\\Users%` will match any `<path>` that " +
+    "starts with '%SystemDrive%\\Users'. It is invalid to escape any other character.\n\n" +
+    "Use RLIKE to match with standard regular expressions.")
```
gatorsmile: https://github.com/apache/spark/pull/15513 is adding examples and changing the function description format and contents for each function. Could you check it and try to make it consistent with the others?

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/RegexpExpressionsSuite.scala
```diff
@@ -53,6 +57,40 @@ class RegexpExpressionsSuite extends SparkFunSuite with ExpressionEvalHelper {
...
+    checkEvaluation("a\u20ACa" like "_€_", true)
+    // scalastyle:on nonascii
+
+    // invalid escaping
+    intercept[AnalysisException] {
+      evaluate("""a""" like """\a""")
+    }
+    intercept[AnalysisException] {
+      evaluate("""a""" like """a\""")
+    }
```
gatorsmile: For these two invalid cases, you also need to check the error messages. For example,
```Scala
    val e = intercept[AnalysisException] {
      evaluate("""a""" like """\a""")
    }.getMessage
    assert(e.contains("xyz"))
```

gatorsmile:
@jodersky Sorry. I misunderstood your point in the offline discussion. After reading the codes, I think your original proposal is right. This PR needs to clean up the test suite `LIKE literal Regular Expression` and `LIKE Non-literal Regular Expression` to avoid the duplicate codes.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/StringUtils.scala
```diff
@@ -19,32 +19,34 @@ package org.apache.spark.sql.catalyst.util
...
+  def escapeLikeRegex(str: String): String = {
+    val in = str.toIterator
+    val out = new StringBuilder()
+
+    def fail(message: String) = throw new AnalysisException(
+      s"the pattern '$str' is invalid, $message")
```
jodersky: >  org.apache.spark.sql.AnalysisException: the pattern '\a' is invalid, the escape character is not allowed to precede 'a';

or 

> org.apache.spark.sql.AnalysisException: The pattern '\a' is invalid. Reason: the escape character is not allowed to precede 'a';
 
Isn't this just a matter of preference? I preferred the first one because it avoids using sentences that are expected to end in a period, when a semi colon is always appended after an analysis exception. It's a really really superficial thing though, so I'm happy to change it either way

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/StringUtils.scala
jodersky: ack

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/regexpExpressions.scala
jodersky: ack

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/RegexpExpressionsSuite.scala
jodersky: ack

jodersky:
@gatorsmile thx for the review. I'll update the test suite as we discussed initially :)

jodersky:
@gatorsmile, I updated the PR according to your comments.

Whilst adding the example test
```scala
 checkEvaluation("""%SystemDrive%\Users\John""" like """\%SystemDrive\%\\Users%""", true)
```
I noticed that the equivalent SQL string:
```scala
spark.sql("""'%SystemDrive%\Users\John like '\%SystemDrive\%\\Users%'""")
```
will throw an exception:
`org.apache.spark.sql.AnalysisException: the pattern '\%SystemDrive\%\Users%' is invalid, the escape character is not allowed to precede 'U';` (Note that a backslash is missing before 'Users')

I traced down the issue to the ANTLR4 parser, that will call [unescapeSQLString](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/parser/ParserUtils.scala#L100) on the string passed to `spark.sql`, replacing any escaped character except for `\_` and `\%`, and thereby removing one backslash before "Users". To solve the issue there are two potential solutions, both flawed in my opinion:

1. Escape the backlash twice, requiring 4 backslashes in total. `spark.sql("""'%SystemDrive%\Users\John like '\%SystemDrive\%\\Users%'""")`
This is inconsistent as escaping the percent requires only one backslash

2. Add a special rule to the parser that ignores double slashes. This won't place nice with custom escape characters in the future.

My question is more fundamental, why are SQL strings escaped in the first place? Should it not be up to the frontend language to escape such strings? I.e. scala/java/python should handle replacing things such as `\n` with the newline or replacing unicode sequences with their corresponding symbols. Getting rid of the extra unescaping would be the cleanest solution to avoid double-escaping in my opinion

gatorsmile:
> My question is more fundamental, why are SQL strings escaped in the first place? Should it not be up to the frontend language to escape such strings? I.e. scala/java/python should handle replacing things such as \n with the newline or replacing unicode sequences with their corresponding symbols.

Based on my understanding, the escape behavior/definition in SQL is different from the frontend language. That is why we need to call `unescapeSQLString` on the string literals. Let us discuss f2f how to handle the case tomorrow.

gatorsmile:
Also cc @hvanhovell who is the major reviewer about parser-related PRs.

jodersky:
@hvanhovell, I just wanted to add that option 2 from my last comment won't work: if we replace `\\` with `\\\\` in the unescape function (as is done for `\%` and `\_`), it will be impossible for a user to specify a single backslash.

Considering that, another potential solution would be change the ANTLR parser to handle escaping differently in LIKE expressions. This would however introduce a special case which complicates the overall rules.
Maybe leaving the current behaviour but documenting that backslashes are escaped on the parser-level is the best option?

JoshRosen:
@mengxr, @rxin, @yhuai, @hvanhovell, what's the status of this PR? How much works remains before this is good to go? What decisions are we blocking on? I ask because this conflicts with a small perf. optimization PR in `escapeLikeRegex` (#16893) which I'd like to see get in because it offers a big perf. boost for one of my workloads.

dbtsai:
Ping @jodersky, @rxin, @mengxr, @yhuai, @hvanhovell we also run into this bug. Will be great we have it by 2.2 release. Thanks.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/regexpExpressions.scala
```diff
@@ -68,7 +68,30 @@ trait StringRegexExpression extends ImplicitCastInputTypes {
...
+    "null if any arguments are null, false otherwise.",
+  extended = """
+    Arguments:
+      str - a string expression
+      pattern - a string expression. The pattern is a string which is matched literally, with
+        exception to the following special symbols:
+
+          _ matches any one character in the input (similar to . in posix regular expressions)
+
+          % matches zero ore more characters in the input (similar to .* in posix regular
```
viirya: ore -> or?

viirya:
Re-checked the current change, I think it is in a good shape. Do we have unsolved issues or decisions on this?

ping @jodersky Would you like to update this with master? Thanks.

cloud-fan:
LGTM

rxin:
I've resolved the conflict and merged this in master/branch-2.1. Thanks.

ericl:
This seems to have broken the build in branch-2.1, e.g. https://amplab.cs.berkeley.edu/jenkins/view/Spark%20QA%20Compile/job/spark-branch-2.1-compile-maven-hadoop-2.6/591/consoleFull

```
[info] Compiling 196 Scala sources and 26 Java sources to /home/jenkins/workspace/spark-branch-2.1-compile-maven-hadoop-2.6/sql/catalyst/target/scala-2.11/classes...
[error] /home/jenkins/workspace/spark-branch-2.1-compile-maven-hadoop-2.6/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/SessionCatalog.scala:1107: not found: value parser
[error]         Try(parser.parseFunctionIdentifier(f)) match {
[error]             ^
[error] /home/jenkins/workspace/spark-branch-2.1-compile-maven-hadoop-2.6/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/SessionCatalog.scala:1118: value funcName is not a member of Any
[error]       case f if FunctionRegistry.functionSet.contains(f.funcName) => (f, "SYSTEM")
[error]                                                         ^
[error] /home/jenkins/workspace/spark-branch-2.1-compile-maven-hadoop-2.6/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/SessionCatalog.scala:1120: type mismatch;
[error]  found   : Seq[(Any, String)]
[error]  required: Seq[(org.apache.spark.sql.catalyst.FunctionIdentifier, String)]
[error]     }.distinct
[error]       ^
[error] /home/jenkins/workspace/spark-branch-2.1-compile-maven-hadoop-2.6/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/regexpExpressions.scala:95: illegal inheritance;
[error]  self-type org.apache.spark.sql.catalyst.expressions.Like does not conform to org.apache.spark.sql.catalyst.expressions.StringRegexExpression's selftype org.apache.spark.sql.catalyst.expressions.StringRegexExpression with org.apache.spark.sql.catalyst.expressions.BinaryExpression
[error] case class Like(left: Expression, right: Expression) extends StringRegexExpression {
[error]                                                              ^
[error] /home/jenkins/workspace/spark-branch-2.1-compile-maven-hadoop-2.6/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/regexpExpressions.scala:134: not found: value nullSafeCodeGen
[error]       nullSafeCodeGen(ctx, ev, (eval1, eval2) => {
[error]       ^
[error] 5 errors found
[error] Compile failed at Apr 17, 2017 12:06:12 PM [13.382s]
```

rxin:
I pushed a commit. Hopefully that fixes it.

jodersky:
Glad to see this being merged. Let me know if there is anything more I can do to help with any issues.

gatorsmile:
Submit a PR to fix the issues in my PR. Thanks!

PR #9238, opened 2015-10-23
[SPARK-11272][Core][UI] Support importing and exporting event logs from HistoryServer web portal

jerryshao:
This patch helps user to easily download the event logs from HistoryServer, or upload other's log to debug and replay. Quite useful for user to seek help or identify problem for others. Here is the screenshot:

screen shot 2015-10-23 at 11 47 40 am

Please review. Thanks a lot.

PR #15400, opened 2016-10-08
[SPARK-11272] [Web UI] Add support for downloading event logs from HistoryServer UI

ajbozarth:

This is a reworked PR based on feedback in #9238 after it was closed and not reopened. As suggested in that PR I've only added the download feature. This functionality already exists in the api and this allows easier access to download event logs to share with others.

I've attached a screenshot of the committed version, but I will also include alternate options with screen shots in the comments below. I'm personally not sure which option is best.
## How was this patch tested?

Manual testing 

screen shot 2016-10-07 at 6 11 12 pm

### DIFF
--- a/core/src/main/resources/org/apache/spark/ui/static/historypage-template.html
+++ b/core/src/main/resources/org/apache/spark/ui/static/historypage-template.html
@@ -59,7 +59,11 @@
           Last Updated
         </span>
       </th>
-    </tr>
+      <th>
+        <span data-toggle="tooltip" data-placement="above" title="Download the event log for this application">
+          Event Log
+        </span>
+      </th>
   </thead>
   <tbody>
   {{#applications}}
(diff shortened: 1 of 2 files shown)


ajbozarth:
The above screen shot shows a version with an Event Log column that allows download of each application attempt individually. Below is the same without a button:
screen shot 2016-10-07 at 5 59 45 pm


The next two versions don't have a separate column and are a download link for each application, meaning if there's more than one attempt there's only one link to download them all zipped together:
screen shot 2016-10-07 at 4 39 58 pm
screen shot 2016-10-07 at 5 05 29 pm

The last one is tied as my preference with the one I committed, it just a matter of if we want the download to be per application or attempt

ajbozarth:
@steveloughran @andrewor14 Would you be willing to take a look at this and give some feedback since you reviewed the original pr for this?

andrewor14:
retest this please

ajbozarth:
@andrewor14 Is there a reason you retested this? Just wondering since it had already passed tests.

andrewor14:
Usually we retest the PR if it's been a few days since it last ran tests. We have had build breaks before where we merged a PR that passed tests a long time ago.

andrewor14:
This one LGTM I'm merging it into master. Thanks for working on this.

PR #15401, opened 2016-10-08
[SPARK-17782][STREAMING][KAFKA] alternative eliminate race condition of poll twice

koeninger:

Alternative approach to https://github.com/apache/spark/pull/15387


zsxwing:
@koeninger sorry for the delay. Actually, my concern is `poll(0)` in `ConsumerStrategy.onStart` may update the offsets.

koeninger:
@zsxwing so poll is only called in consumer strategy in situations in which starting offsets have been provided, and seek is called immediately thereafter for those offsets.  What is the specific concern at that point?

If you want me to make paranoidPoll a static method and call it from consumer offsets as well I can, but....

zsxwing:
@koeninger Sorry. I didn't notice the `seek`. LGTM. Merging to master and 2.0.

PR #15402, opened 2016-10-08
[SPARK-17835][ML][MLlib] Optimize NaiveBayes mllib wrapper to eliminate extra pass on data

yanboliang:

[SPARK-14077](https://issues.apache.org/jira/browse/SPARK-14077) copied the `NaiveBayes` implementation from mllib to ml and left mllib as a wrapper. However, there are some difference between mllib and ml to handle labels:
- mllib allow input labels as {-1, +1}, however, ml assumes the input labels in range [0, numClasses).
- mllib `NaiveBayesModel` expose `labels` but ml did not due to the assumption mention above.

During the copy in [SPARK-14077](https://issues.apache.org/jira/browse/SPARK-14077), we use
`val labels = data.map(_.label).distinct().collect().sorted`
to get the distinct labels firstly, and then encode the labels for training. It involves extra Spark job compared with the original implementation. Since `NaiveBayes` only do one pass aggregation during training, adding another one seems less efficient. We can get the labels in a single pass along with `NaiveBayes` training and send them to MLlib side.
## How was this patch tested?

Existing tests.

### DIFF
--- a/mllib/src/main/scala/org/apache/spark/ml/classification/NaiveBayes.scala
+++ b/mllib/src/main/scala/org/apache/spark/ml/classification/NaiveBayes.scala
@@ -110,16 +110,28 @@ class NaiveBayes @Since("1.5.0") (
   @Since("2.1.0")
   def setWeightCol(value: String): this.type = set(weightCol, value)
 
+  /**
+   * ml assumes input labels in range [0, numClasses). But this implementation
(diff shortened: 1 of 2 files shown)


yanboliang:
cc @zhengruifeng @sethah @srowen

zhengruifeng:
LGTM

yanboliang:
Merged into master. Thanks for review. @zhengruifeng

PR #15403, opened 2016-10-08
[SPARK-17832][SQL] TableIdentifier.quotedString creates un-parseable names when name contains a backtick

jiangxb1987:

The `quotedString` method in `TableIdentifier` and `FunctionIdentifier` produce an illegal (un-parseable) name when the name contains a backtick. For example:

```
import org.apache.spark.sql.catalyst.parser.CatalystSqlParser._
import org.apache.spark.sql.catalyst.TableIdentifier
import org.apache.spark.sql.catalyst.analysis.UnresolvedAttribute
val complexName = TableIdentifier("`weird`table`name", Some("`d`b`1"))
parseTableIdentifier(complexName.unquotedString) // Does not work
parseTableIdentifier(complexName.quotedString) // Does not work
parseExpression(complexName.unquotedString) // Does not work
parseExpression(complexName.quotedString) // Does not work
```

We should handle the backtick properly to make `quotedString` parseable.
## How was this patch tested?

Add new testcases in `TableIdentifierParserSuite` and `ExpressionParserSuite`.

### DIFF
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/identifiers.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/identifiers.scala
@@ -29,8 +29,16 @@ sealed trait IdentifierWithDatabase {
 
   def database: Option[String]
 
+  private def replaceBackticks(origin: String): String = origin.replace("`", "``")
+
   def quotedString: String = {
-    if (database.isDefined) s"`${database.get}`.`$identifier`" else s"`$identifier`"
+    val replacedIdentifier = replaceBackticks(identifier)
+    if (database.isDefined) {
+      val replacedDatabase = replaceBackticks(database.get)
+      s"`$replacedDatabase`.`$replacedIdentifier`"
+    } else {
+      s"`$replacedIdentifier`"
+    }
   }
 
   def unquotedString: String = {
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/identifiers.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/identifiers.scala
@@ -41,10 +41,7 @@ sealed trait IdentifierWithDatabase {
   }
 
   def unquotedString: String = {
-    val replacedId = quoteIdentifier(identifier)
-    val replacedDb = database.map(quoteIdentifier(_))
-
-    if (replacedDb.isDefined) s"${replacedDb.get}.$replacedId" else replacedId
+    if (database.isDefined) s"${database.get}.$identifier" else identifier
   }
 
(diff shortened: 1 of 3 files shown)


jiangxb1987:
Besides, we have similar problem in `lookupFunction` as well as `lookupTable` if one of `database`、`table`、`function` contains backtick, should we resolve the problem in this PR too?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/identifiers.scala
```diff
@@ -29,12 +28,23 @@ sealed trait IdentifierWithDatabase {
...
   def quotedString: String = {
-    if (database.isDefined) s"`${database.get}`.`$identifier`" else s"`$identifier`"
+    val replacedId = quoteIdentifier(identifier)
+    val replacedDb = database.map(quoteIdentifier(_))
+
+    if (replacedDb.isDefined) s"`${replacedDb.get}`.`$replacedId`" else s"`$replacedId`"
   }
 
   def unquotedString: String = {
-    if (database.isDefined) s"${database.get}.$identifier" else identifier
```
hvanhovell: I think we should keep `unquotedString ` this as it is. That gives some a more readable (less parseable) version of the identifier. I just added an example with `unquotedIdentifier` to show that there was no way to get a parseable string from the table identifier.

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/parser/TableIdentifierParserSuite.scala
```diff
@@ -104,4 +104,15 @@ class TableIdentifierParserSuite extends SparkFunSuite {
     // ".123BD" should not be treated as token of type BIGDECIMAL_LITERAL
     assert(parseTableIdentifier("a.123BD_LIST") == TableIdentifier("123BD_LIST", Some("a")))
   }
+
+  test("SPARK-17832 table identifier - contains backtick") {
+    val complexName = TableIdentifier("`weird`table`name", Some("`d`b`1"))
+    assert(TableIdentifier("`weird`table`name", Some("`d`b`1")) ===
```
hvanhovell: Use complexName? Why create the identifier twice?

hvanhovell:
@jiangxb1987 This looks pretty good.

What is the problem `getTable` and `getFunction`? Could you give an example? assume you are talking about the `org.apache.spark.sql.catalog.Catalog` here?

jiangxb1987:
@hvanhovell nvm about the `catalog.getTable` issue, it turns out to be my mistake. Sorry about that...

hvanhovell:
LGTM - merging to master/2.0. Thanks!

PR #15406, opened 2016-10-09
[Spark-17745][ml][PySpark] update NB python api - add weight col parameter

WeichenXu123:

update python api for NaiveBayes: add weight col parameter.
## How was this patch tested?

doctests added.

### DIFF
--- a/python/pyspark/ml/classification.py
+++ b/python/pyspark/ml/classification.py
@@ -981,7 +981,7 @@ class GBTClassificationModel(TreeEnsembleModel, JavaPredictionModel, JavaMLWrita
 
 @inherit_doc
 class NaiveBayes(JavaEstimator, HasFeaturesCol, HasLabelCol, HasPredictionCol, HasProbabilityCol,
-                 HasRawPredictionCol, HasThresholds, JavaMLWritable, JavaMLReadable):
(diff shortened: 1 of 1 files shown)


sethah:
Would you mind changing the title to reflect the actual change? Something to indicate that you are adding weight column param to Python? Thanks!

WeichenXu123:
@sethah OK. and I'm checking whether there is something else need to update...

sethah:
We should add weights to the doctests to demonstrate them and make sure they're working.

python/pyspark/ml/classification.py
```diff
@@ -1032,6 +1032,16 @@ class NaiveBayes(JavaEstimator, HasFeaturesCol, HasLabelCol, HasPredictionCol, H
     >>> result = model3.transform(test0).head()
     >>> result.prediction
     0.0
+    >>> dfWithWeight = spark.createDataFrame([
```
sethah: Sorry for the confusion, I was thinking we could just add weights to the existing example. I don't think an entirely separate example is necessary. What do you think?

yanboliang: +1 @sethah

sethah:
LGTM

yanboliang:
LGTM2, merged into master. Thanks!

PR #15408, opened 2016-10-09
[SPARK-17839][CORE] Use Nio's directbuffer instead of BufferedInputStream in order to avoid additional copy from os buffer cache to user buffer

sitalkedia:

Currently we use BufferedInputStream to read the shuffle file which copies the file content from os buffer cache to the user buffer. This adds additional latency in reading the spill files. We made a change to use java nio's direct buffer to read the spill files and for certain pipelines spilling significant amount of data, we see up to 7% speedup for the entire pipeline. 
## How was this patch tested?

Tested by running the job in the cluster and observed up to 7% speedup.

### DIFF
--- a/core/src/main/java/org/apache/spark/io/NioBasedBufferedFileInputStream.java
+++ b/core/src/main/java/org/apache/spark/io/NioBasedBufferedFileInputStream.java
@@ -92,7 +92,7 @@ public final class NioBasedBufferedFileInputStream extends InputStream {
     if (n < 0L) {
       return 0L;
     }
-    if (byteBuffer.remaining() > n) {
+    if (byteBuffer.remaining() >= n) {
       // The buffered content is enough to skip
       byteBuffer.position(byteBuffer.position() + (int) n);
       return n;
--- a/core/src/main/java/org/apache/spark/io/NioBasedBufferedFileInputStream.java
+++ b/core/src/main/java/org/apache/spark/io/NioBasedBufferedFileInputStream.java
@@ -14,7 +14,6 @@
 package org.apache.spark.io;
 
 import java.io.File;
-import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.ByteBuffer;
@@ -34,15 +33,14 @@ public final class NioBasedBufferedFileInputStream extends InputStream {
 
   private static int DEFAULT_BUFFER_SIZE = 8192;
 
-  private final ByteBuffer bb;
+  private final ByteBuffer byteBuffer;
 
-  private final FileChannel ch;
+  private final FileChannel fileChannel;
 
   public NioBasedBufferedFileInputStream(File file, int bufferSize) throws IOException {
-    bb = ByteBuffer.allocateDirect(bufferSize);
-    ch = FileChannel.open(file.toPath(), StandardOpenOption.READ);
-    ch.read(bb);
-    bb.flip();
+    byteBuffer = ByteBuffer.allocateDirect(bufferSize);
+    fileChannel = FileChannel.open(file.toPath(), StandardOpenOption.READ);
+    byteBuffer.flip();
   }
 
   public NioBasedBufferedFileInputStream(File file) throws IOException {
@@ -50,13 +48,13 @@ public final class NioBasedBufferedFileInputStream extends InputStream {
   }
 
   private boolean refill() throws IOException {
-    if (!bb.hasRemaining()) {
-      bb.clear();
-      int nRead = ch.read(bb);
+    if (!byteBuffer.hasRemaining()) {
+      byteBuffer.clear();
+      int nRead = fileChannel.read(byteBuffer);
       if (nRead == -1) {
         return false;
       }
-      bb.flip();
+      byteBuffer.flip();
     }
     return true;
   }
@@ -66,7 +64,7 @@ public final class NioBasedBufferedFileInputStream extends InputStream {
     if (!refill()) {
       return -1;
     }
-    return bb.get() & 0xFF;
+    return byteBuffer.get() & 0xFF;
   }
 
   @Override
@@ -74,18 +72,49 @@ public final class NioBasedBufferedFileInputStream extends InputStream {
     if (!refill()) {
       return -1;
     }
-    len = Math.min(len, bb.remaining());
-    bb.get(b, off, len);
+    len = Math.min(len, byteBuffer.remaining());
+    byteBuffer.get(b, off, len);
     return len;
   }
 
   @Override
   public int available() throws IOException {
-    return bb.remaining();
+    return byteBuffer.remaining();
+  }
+
+
+  @Override
+  public long skip(long n) throws IOException {
+    if(n < 0L) {
+      return 0L;
+    }
+    if (byteBuffer.remaining() > n) {
+      // The buffered content is enough to skip
+      byteBuffer.position(byteBuffer.position() + (int) n);
+      return n;
+    }
+    long toSkip = n - byteBuffer.remaining();
+    // Discard everything we have read in the buffer.
+    byteBuffer.position(0);
+    byteBuffer.flip();
+    return skipFromFileChannel(toSkip);
+  }
+
+  private long skipFromFileChannel(long n) throws IOException {
+    long currentFilePosition = fileChannel.position();
+    long size = fileChannel.size();
+    if (currentFilePosition + n > size) {
+      fileChannel.position(size);
+      return size - currentFilePosition;
+    }
+    else {
+      fileChannel.position(currentFilePosition + n);
+      return n;
+    }
   }
 
   @Override
   public void close() throws IOException {
-    ch.close();
+    fileChannel.close();
   }
 }
--- a/core/src/main/java/org/apache/spark/io/NioBasedBufferedFileInputStream.java
+++ b/core/src/main/java/org/apache/spark/io/NioBasedBufferedFileInputStream.java
@@ -13,6 +13,8 @@
  */
 package org.apache.spark.io;
 
+import org.apache.spark.storage.StorageUtils;
+
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
@@ -120,5 +122,6 @@ public final class NioBasedBufferedFileInputStream extends InputStream {
   @Override
   public void close() throws IOException {
     fileChannel.close();
+    StorageUtils.dispose(byteBuffer);
   }
 }
--- a/core/src/main/java/org/apache/spark/io/NioBasedBufferedFileInputStream.java
+++ b/core/src/main/java/org/apache/spark/io/NioBasedBufferedFileInputStream.java
@@ -19,28 +19,37 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.nio.ByteBuffer;
 import java.nio.channels.FileChannel;
+import java.nio.file.StandardOpenOption;
 
 /**
  * {@link InputStream} implementation which uses direct buffer
  * to read a file to avoid extra copy of data between Java and
- * native memory which happens when using {@link java.io.BufferedInputStream}
+ * native memory which happens when using {@link java.io.BufferedInputStream}.
+ * Unfortunately, this is not something already available in JDK,
+ * {@link sun.nio.ch.ChannelInputStream} supports reading a file using nio,
+ * but does not support buffering.
  *
  */
 public final class NioBasedBufferedFileInputStream extends InputStream {
 
-  ByteBuffer bb;
+  private static int DEFAULT_BUFFER_SIZE = 8192;
 
-  FileChannel ch;
+  private final ByteBuffer bb;
+
+  private final FileChannel ch;
 
   public NioBasedBufferedFileInputStream(File file, int bufferSize) throws IOException {
     bb = ByteBuffer.allocateDirect(bufferSize);
-    FileInputStream f = new FileInputStream(file);
-    ch = f.getChannel();
+    ch = FileChannel.open(file.toPath(), StandardOpenOption.READ);
     ch.read(bb);
     bb.flip();
   }
 
-  public boolean refill() throws IOException {
+  public NioBasedBufferedFileInputStream(File file) throws IOException {
+    this(file, DEFAULT_BUFFER_SIZE);
+  }
+
+  private boolean refill() throws IOException {
     if (!bb.hasRemaining()) {
       bb.clear();
       int nRead = ch.read(bb);
@@ -57,7 +66,7 @@ public final class NioBasedBufferedFileInputStream extends InputStream {
     if (!refill()) {
       return -1;
     }
-    return bb.get();
+    return bb.get() & 0xFF;
   }
 
   @Override
@@ -68,6 +77,11 @@ public final class NioBasedBufferedFileInputStream extends InputStream {
     len = Math.min(len, bb.remaining());
     bb.get(b, off, len);
     return len;
+  }
+
+  @Override
+  public int available() throws IOException {
+    return bb.remaining();
   }
 
   @Override
--- a/core/src/main/java/org/apache/spark/io/NioBasedBufferedFileInputStream.java
+++ b/core/src/main/java/org/apache/spark/io/NioBasedBufferedFileInputStream.java
@@ -1,0 +1,77 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.spark.io;
+
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.ByteBuffer;
+import java.nio.channels.FileChannel;
+
+/**
+ * {@link InputStream} implementation which uses direct buffer
+ * to read a file to avoid extra copy of data between Java and
+ * native memory which happens when using {@link java.io.BufferedInputStream}
+ *
+ */
+public final class NioBasedBufferedFileInputStream extends InputStream {
+
+  ByteBuffer bb;
+
+  FileChannel ch;
+
+  public NioBasedBufferedFileInputStream(File file, int bufferSize) throws IOException {
+    bb = ByteBuffer.allocateDirect(bufferSize);
+    FileInputStream f = new FileInputStream(file);
+    ch = f.getChannel();
+    ch.read(bb);
+    bb.flip();
+  }
+
+  public boolean refill() throws IOException {
+    if (!bb.hasRemaining()) {
+      bb.clear();
+      int nRead = ch.read(bb);
+      if (nRead == -1) {
+        return false;
+      }
+      bb.flip();
+    }
+    return true;
+  }
+
+  @Override
+  public int read() throws IOException {
+    if (!refill()) {
+      return -1;
+    }
+    return bb.get();
+  }
+
+  @Override
+  public int read(byte[] b, int off, int len) throws IOException {
+    if (!refill()) {
+      return -1;
+    }
+    len = Math.min(len, bb.remaining());
+    bb.get(b, off, len);
+    return len;
+  }
+
+  @Override
+  public void close() throws IOException {
+    ch.close();
+  }
+}
--- a/core/src/main/java/org/apache/spark/io/NioBasedBufferedFileInputStream.java
+++ b/core/src/main/java/org/apache/spark/io/NioBasedBufferedFileInputStream.java
@@ -31,22 +31,27 @@ import java.nio.file.StandardOpenOption;
  */
 public final class NioBasedBufferedFileInputStream extends InputStream {
 
-  private static int DEFAULT_BUFFER_SIZE = 8192;
+  private static int DEFAULT_BUFFER_SIZE_BYTES = 8192;
 
   private final ByteBuffer byteBuffer;
 
   private final FileChannel fileChannel;
 
-  public NioBasedBufferedFileInputStream(File file, int bufferSize) throws IOException {
-    byteBuffer = ByteBuffer.allocateDirect(bufferSize);
+  public NioBasedBufferedFileInputStream(File file, int bufferSizeInBytes) throws IOException {
+    byteBuffer = ByteBuffer.allocateDirect(bufferSizeInBytes);
     fileChannel = FileChannel.open(file.toPath(), StandardOpenOption.READ);
     byteBuffer.flip();
   }
 
   public NioBasedBufferedFileInputStream(File file) throws IOException {
-    this(file, DEFAULT_BUFFER_SIZE);
+    this(file, DEFAULT_BUFFER_SIZE_BYTES);
   }
 
+  /**
+   * Checks weather data is left to be read from the input stream.
+   * @return true if data is left, false otherwise
+   * @throws IOException
+   */
   private boolean refill() throws IOException {
     if (!byteBuffer.hasRemaining()) {
       byteBuffer.clear();
@@ -68,12 +73,12 @@ public final class NioBasedBufferedFileInputStream extends InputStream {
   }
 
   @Override
-  public int read(byte[] b, int off, int len) throws IOException {
+  public int read(byte[] b, int offset, int len) throws IOException {
     if (!refill()) {
       return -1;
     }
     len = Math.min(len, byteBuffer.remaining());
-    byteBuffer.get(b, off, len);
+    byteBuffer.get(b, offset, len);
     return len;
   }
 
@@ -82,10 +87,9 @@ public final class NioBasedBufferedFileInputStream extends InputStream {
     return byteBuffer.remaining();
   }
 
-
   @Override
   public long skip(long n) throws IOException {
-    if(n < 0L) {
+    if (n < 0L) {
       return 0L;
     }
     if (byteBuffer.remaining() > n) {
@@ -93,11 +97,12 @@ public final class NioBasedBufferedFileInputStream extends InputStream {
       byteBuffer.position(byteBuffer.position() + (int) n);
       return n;
     }
-    long toSkip = n - byteBuffer.remaining();
+    long skippedFromBuffer = byteBuffer.remaining();
+    long toSkipFromFileChannel = n - skippedFromBuffer;
     // Discard everything we have read in the buffer.
     byteBuffer.position(0);
     byteBuffer.flip();
-    return skipFromFileChannel(toSkip);
+    return skippedFromBuffer + skipFromFileChannel(toSkipFromFileChannel);
   }
 
   private long skipFromFileChannel(long n) throws IOException {
@@ -106,8 +111,7 @@ public final class NioBasedBufferedFileInputStream extends InputStream {
     if (currentFilePosition + n > size) {
       fileChannel.position(size);
       return size - currentFilePosition;
-    }
-    else {
+    } else {
       fileChannel.position(currentFilePosition + n);
       return n;
     }
--- a/core/src/main/java/org/apache/spark/io/NioBufferedFileInputStream.java
+++ b/core/src/main/java/org/apache/spark/io/NioBufferedFileInputStream.java
@@ -83,13 +83,11 @@ public final class NioBufferedFileInputStream extends InputStream {
 
   @Override
   public synchronized int read(byte[] b, int offset, int len) throws IOException {
-    if (!refill()) {
-      return -1;
-    }
     if ((offset | len | (offset + len) | (b.length - (offset + len))) < 0) {
       throw new IndexOutOfBoundsException();
-    } else if (len == 0) {
-      return 0;
+    }
+    if (!refill()) {
+      return -1;
     }
     len = Math.min(len, byteBuffer.remaining());
     byteBuffer.get(b, offset, len);
@@ -132,8 +130,13 @@ public final class NioBufferedFileInputStream extends InputStream {
   }
 
   @Override
-  public void close() throws IOException {
+  public synchronized void close() throws IOException {
     fileChannel.close();
     StorageUtils.dispose(byteBuffer);
+  }
+
+  @Override
+  protected void finalize() throws IOException {
+      close();
   }
 }
--- a/core/src/main/java/org/apache/spark/io/NioBufferedFileInputStream.java
+++ b/core/src/main/java/org/apache/spark/io/NioBufferedFileInputStream.java
@@ -15,6 +15,7 @@ package org.apache.spark.io;
 
 import org.apache.spark.storage.StorageUtils;
 
+import javax.annotation.concurrent.ThreadSafe;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
@@ -33,6 +34,7 @@ import java.nio.file.StandardOpenOption;
  * TODO: support {@link #mark(int)}/{@link #reset()}
  *
  */
+@ThreadSafe
 public final class NioBufferedFileInputStream extends InputStream {
 
   private static int DEFAULT_BUFFER_SIZE_BYTES = 8192;
@@ -72,7 +74,7 @@ public final class NioBufferedFileInputStream extends InputStream {
   }
 
   @Override
-  public int read() throws IOException {
+  public synchronized int read() throws IOException {
     if (!refill()) {
       return -1;
     }
@@ -80,9 +82,14 @@ public final class NioBufferedFileInputStream extends InputStream {
   }
 
   @Override
-  public int read(byte[] b, int offset, int len) throws IOException {
+  public synchronized int read(byte[] b, int offset, int len) throws IOException {
     if (!refill()) {
       return -1;
+    }
+    if ((offset | len | (offset + len) | (b.length - (offset + len))) < 0) {
+      throw new IndexOutOfBoundsException();
+    } else if (len == 0) {
+      return 0;
     }
     len = Math.min(len, byteBuffer.remaining());
     byteBuffer.get(b, offset, len);
@@ -90,12 +97,12 @@ public final class NioBufferedFileInputStream extends InputStream {
   }
 
   @Override
-  public int available() throws IOException {
+  public synchronized int available() throws IOException {
     return byteBuffer.remaining();
   }
 
   @Override
-  public long skip(long n) throws IOException {
+  public synchronized long skip(long n) throws IOException {
     if (n <= 0L) {
       return 0L;
     }
@@ -115,7 +122,7 @@ public final class NioBufferedFileInputStream extends InputStream {
   private long skipFromFileChannel(long n) throws IOException {
     long currentFilePosition = fileChannel.position();
     long size = fileChannel.size();
-    if (currentFilePosition + n > size) {
+    if (n > size - currentFilePosition) {
       fileChannel.position(size);
       return size - currentFilePosition;
     } else {
--- a/core/src/main/java/org/apache/spark/io/NioBufferedFileInputStream.java
+++ b/core/src/main/java/org/apache/spark/io/NioBufferedFileInputStream.java
@@ -78,7 +78,7 @@ public final class NioBufferedFileInputStream extends InputStream {
 
   @Override
   public synchronized int read(byte[] b, int offset, int len) throws IOException {
-    if (offset < 0 || len < 0 || offset + len > b.length) {
+    if (offset < 0 || len < 0 || offset + len < 0 || offset + len > b.length) {
       throw new IndexOutOfBoundsException();
     }
     if (!refill()) {
--- a/core/src/main/java/org/apache/spark/io/NioBufferedFileInputStream.java
+++ b/core/src/main/java/org/apache/spark/io/NioBufferedFileInputStream.java
@@ -15,7 +15,6 @@ package org.apache.spark.io;
 
 import org.apache.spark.storage.StorageUtils;
 
-import javax.annotation.concurrent.ThreadSafe;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
@@ -79,7 +78,7 @@ public final class NioBufferedFileInputStream extends InputStream {
 
   @Override
   public synchronized int read(byte[] b, int offset, int len) throws IOException {
-    if (offset < 0 || len < 0 || (b.length - (offset + len)) < 0) {
+    if (offset < 0 || len < 0 || offset + len > b.length) {
       throw new IndexOutOfBoundsException();
     }
     if (!refill()) {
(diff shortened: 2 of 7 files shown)


sitalkedia:
cc - @rxin

core/src/main/java/org/apache/spark/io/NioBasedBufferedFileInputStream.java
```diff
@@ -0,0 +1,77 @@
...
+ */
+public final class NioBasedBufferedFileInputStream extends InputStream {
+
+  ByteBuffer bb;
+
+  FileChannel ch;
+
+  public NioBasedBufferedFileInputStream(File file, int bufferSize) throws IOException {
+    bb = ByteBuffer.allocateDirect(bufferSize);
+    FileInputStream f = new FileInputStream(file);
```
srowen: I'm not entirely sure about this, but seems so from reading the source: closing the channel won't free the file descriptor? can you just hang on to this ref as well to close it in `close()` for completeness?

Better yet, can't we just use `FileChannel.open(...)` and not bother with the stream?

core/src/main/java/org/apache/spark/io/NioBasedBufferedFileInputStream.java
```diff
@@ -0,0 +1,77 @@
...
+    ch = f.getChannel();
+    ch.read(bb);
+    bb.flip();
+  }
+
+  public boolean refill() throws IOException {
```
srowen: `private`?

core/src/main/java/org/apache/spark/io/NioBasedBufferedFileInputStream.java
```diff
@@ -0,0 +1,77 @@
...
+    len = Math.min(len, bb.remaining());
+    bb.get(b, off, len);
+    return len;
+  }
+
+  @Override
+  public void close() throws IOException {
+    ch.close();
+  }
+}
```
srowen: I think you'll want to implement `skip()` and `available()` too.

core/src/main/java/org/apache/spark/util/collection/unsafe/sort/UnsafeSorterSpillReader.java
```diff
@@ -69,8 +70,8 @@ public UnsafeSorterSpillReader(
       bufferSizeBytes = DEFAULT_BUFFER_SIZE_BYTES;
     }
 
-    final BufferedInputStream bs =
-        new BufferedInputStream(new FileInputStream(file), (int) bufferSizeBytes);
+    final NioBasedBufferedFileInputStream bs =
```
srowen: BTW I think IndexShuffleBlockResolver, RowQueue have instances that could use a similar treatment.

Can this reference be `InputStream`?

srowen (changes requested):
Seems conceptually sound and the benchmark shows good results. I wonder, is this not something already available in the JDK? I could only find this from the JDK, which doesn't do internal buffering, but does some more stuff that might be worth comparing notes with: http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/8-b132/sun/nio/ch/ChannelInputStream.java/

core/src/main/java/org/apache/spark/io/NioBasedBufferedFileInputStream.java
```diff
@@ -0,0 +1,77 @@
...
+
+/**
+ * {@link InputStream} implementation which uses direct buffer
+ * to read a file to avoid extra copy of data between Java and
+ * native memory which happens when using {@link java.io.BufferedInputStream}
+ *
+ */
+public final class NioBasedBufferedFileInputStream extends InputStream {
+
+  ByteBuffer bb;
```
sitalkedia: done.

core/src/main/java/org/apache/spark/io/NioBasedBufferedFileInputStream.java
```diff
@@ -0,0 +1,77 @@
...
+
+  FileChannel ch;
+
+  public NioBasedBufferedFileInputStream(File file, int bufferSize) throws IOException {
```
sitalkedia: that makes sense, changed accordingly.

core/src/main/java/org/apache/spark/io/NioBasedBufferedFileInputStream.java
sitalkedia: changed accordingly.

core/src/main/java/org/apache/spark/io/NioBasedBufferedFileInputStream.java
sitalkedia: `skip()` is already implemented by underlying `InputStream` so that is not needed. Implemented `available()` .

core/src/main/java/org/apache/spark/util/collection/unsafe/sort/UnsafeSorterSpillReader.java
sitalkedia: Yes, there are other instance in the code where we are using `BufferedInputStream`. I will file raise another PR to replace those with `NioBasedBufferedFileInputStream`.

sitalkedia:
>> Seems conceptually sound and the benchmark shows good results. I wonder, is this not something already available in the JDK? 

Yes, I looked around a bit and unfortunately, I could not find any implementation in JDK which support buffering based nio read.

core/src/main/java/org/apache/spark/io/NioBasedBufferedFileInputStream.java
```diff
@@ -0,0 +1,91 @@
...
+/**
+ * {@link InputStream} implementation which uses direct buffer
+ * to read a file to avoid extra copy of data between Java and
+ * native memory which happens when using {@link java.io.BufferedInputStream}.
+ * Unfortunately, this is not something already available in JDK,
+ * {@link sun.nio.ch.ChannelInputStream} supports reading a file using nio,
+ * but does not support buffering.
+ *
+ */
+public final class NioBasedBufferedFileInputStream extends InputStream {
```
rxin: You should add a test suite specifically for this.

rxin:
Can you also expand on what the 7% means? Is it some workload end-to-end that's been improved by 7%, or the sorting itself improves by 7%?

core/src/main/java/org/apache/spark/io/NioBasedBufferedFileInputStream.java
witgo: `skip()` in InputStream will call ` read()`,  this is not the optimal solution

core/src/main/java/org/apache/spark/io/NioBasedBufferedFileInputStream.java
```diff
@@ -0,0 +1,91 @@
...
+
+  private final ByteBuffer bb;
+
+  private final FileChannel ch;
+
+  public NioBasedBufferedFileInputStream(File file, int bufferSize) throws IOException {
+    bb = ByteBuffer.allocateDirect(bufferSize);
+    ch = FileChannel.open(file.toPath(), StandardOpenOption.READ);
+    ch.read(bb);
+    bb.flip();
```
witgo: ' ch.read(bb);' can be removed

sitalkedia: removed, thanks!

core/src/main/java/org/apache/spark/io/NioBasedBufferedFileInputStream.java
sitalkedia: Added a test suite for this.

core/src/main/java/org/apache/spark/io/NioBasedBufferedFileInputStream.java
sitalkedia: Although, we are not using `skip()` anywhere, but I implemented an optimal version of it anyways.

sitalkedia:
>> Can you also expand on what the 7% means? Is it some workload end-to-end that's been improved by 7%, or the sorting itself improves by 7%?

The perf improvement was end-to-end which means the sorting improvement is definitely more than that.

core/src/main/java/org/apache/spark/io/NioBasedBufferedFileInputStream.java
```diff
@@ -0,0 +1,120 @@
...
+ */
+public final class NioBasedBufferedFileInputStream extends InputStream {
+
+  private static int DEFAULT_BUFFER_SIZE = 8192;
+
+  private final ByteBuffer byteBuffer;
+
+  private final FileChannel fileChannel;
+
+  public NioBasedBufferedFileInputStream(File file, int bufferSize) throws IOException {
```
rxin: here too, bufferSizeInBytes

core/src/main/java/org/apache/spark/io/NioBasedBufferedFileInputStream.java
```diff
@@ -0,0 +1,120 @@
...
+    byteBuffer = ByteBuffer.allocateDirect(bufferSize);
+    fileChannel = FileChannel.open(file.toPath(), StandardOpenOption.READ);
+    byteBuffer.flip();
+  }
+
+  public NioBasedBufferedFileInputStream(File file) throws IOException {
+    this(file, DEFAULT_BUFFER_SIZE);
+  }
+
+  private boolean refill() throws IOException {
```
rxin: document this function and explain what the ret value means?

core/src/main/java/org/apache/spark/io/NioBasedBufferedFileInputStream.java
```diff
@@ -0,0 +1,120 @@
...
+  @Override
+  public int read() throws IOException {
+    if (!refill()) {
+      return -1;
+    }
+    return byteBuffer.get() & 0xFF;
+  }
+
+  @Override
+  public int read(byte[] b, int off, int len) throws IOException {
```
rxin: off -> offset

core/src/main/java/org/apache/spark/io/NioBasedBufferedFileInputStream.java
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  private long skipFromFileChannel(long n) throws IOException {
+    long currentFilePosition = fileChannel.position();
+    long size = fileChannel.size();
+    if (currentFilePosition + n > size) {
+      fileChannel.position(size);
+      return size - currentFilePosition;
+    }
+    else {
```
rxin: move else to previous line

core/src/main/java/org/apache/spark/io/NioBasedBufferedFileInputStream.java
```diff
@@ -0,0 +1,120 @@
...
+      return n;
+    }
+    long toSkip = n - byteBuffer.remaining();
+    // Discard everything we have read in the buffer.
+    byteBuffer.position(0);
+    byteBuffer.flip();
+    return skipFromFileChannel(toSkip);
+  }
+
+  private long skipFromFileChannel(long n) throws IOException {
```
rxin: maybe just inline this in the previous method, and add an else case so both cases are indented the same level?

rxin: actually never mind this comment

core/src/main/java/org/apache/spark/io/NioBasedBufferedFileInputStream.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+    else {
+      fileChannel.position(currentFilePosition + n);
+      return n;
+    }
+  }
+
+  @Override
+  public void close() throws IOException {
+    fileChannel.close();
```
rxin: do we need to free the buffer or anything?

core/src/test/java/org/apache/spark/io/NioBasedBufferedFileInputStreamSuite.java
```diff
@@ -0,0 +1,102 @@
...
+      int read = inputStream.read(readBytes, 0, 8 * 1024);
+      for (int j = 0; j < read; j++) {
+        assertEquals(randomBytes[i], readBytes[j]);
+        i++;
+      }
+    }
+  }
+
+  @Test
+  public void testBytesSkipped() throws IOException {
```
rxin: can we add a test case for negative skips?

core/src/test/java/org/apache/spark/io/NioBasedBufferedFileInputStreamSuite.java
```diff
@@ -0,0 +1,102 @@
...
+  @Test
+  public void testReadOneByte() throws IOException {
+    InputStream inputStream = new NioBasedBufferedFileInputStream(inputFile);
+    for (int i = 0; i < randomBytes.length; i++) {
+      assertEquals(randomBytes[i], (byte) inputStream.read());
+    }
+  }
+
+  @Test
+  public void testReadMultipleBytes() throws IOException {
```
rxin: we should construct a test case that triggers the code path in skipFromFileChannel very explicitly

rxin:
Looks pretty good overall. I left some comments mostly on styling, documentation, and test cases.

core/src/main/java/org/apache/spark/io/NioBasedBufferedFileInputStream.java
```diff
@@ -0,0 +1,120 @@
...
+    if (!refill()) {
+      return -1;
+    }
+    len = Math.min(len, byteBuffer.remaining());
+    byteBuffer.get(b, off, len);
+    return len;
+  }
+
+  @Override
+  public int available() throws IOException {
```
srowen: It's a standard API method for `InputStream`, so if something does call it, it's better to return a useful answer because it's pretty easy to do so.

core/src/main/java/org/apache/spark/io/NioBasedBufferedFileInputStream.java
```diff
@@ -0,0 +1,120 @@
...
+    return byteBuffer.remaining();
+  }
+
+
+  @Override
+  public long skip(long n) throws IOException {
+    if(n < 0L) {
+      return 0L;
+    }
+    if (byteBuffer.remaining() > n) {
```
srowen: >= n ? or did I miss something obvious?

core/src/test/java/org/apache/spark/io/NioBasedBufferedFileInputStreamSuite.java
```diff
@@ -0,0 +1,102 @@
...
+import java.io.InputStream;
+
+import static org.junit.Assert.assertEquals;
+
+/**
+ * Tests functionality of {@link NioBasedBufferedFileInputStream}
+ */
+public class NioBasedBufferedFileInputStreamSuite {
+
+  byte[] randomBytes;
```
srowen: Nit: `private` should always be the default

srowen (changes requested):
Looking good; I don't know that it makes sense to change other instances separately? Changing all instances of buffering a file input stream seems like one coherent logical change (but the titles need to be updates)

core/src/main/java/org/apache/spark/io/NioBasedBufferedFileInputStream.java
sitalkedia: If we have enough data in the bytebuffer to skip then we can skip from the buffer and no need to skip from the file. Looks good to me.

sitalkedia:
Thanks @srowen, @rxin  for the review comments. I have addressed them.

core/src/test/java/org/apache/spark/io/NioBasedBufferedFileInputStreamSuite.java
sitalkedia: Good idea. Actually I discovered a bug with my previous implementation after adding this test case, which I have fixed.

core/src/test/java/org/apache/spark/io/NioBasedBufferedFileInputStreamSuite.java
sitalkedia: done.

core/src/main/java/org/apache/spark/io/NioBasedBufferedFileInputStream.java
sitalkedia: done.

core/src/main/java/org/apache/spark/io/NioBasedBufferedFileInputStream.java
```diff
@@ -0,0 +1,120 @@
...
+
+  @Override
+  public int available() throws IOException {
+    return byteBuffer.remaining();
+  }
+
+
+  @Override
+  public long skip(long n) throws IOException {
+    if(n < 0L) {
```
sitalkedia: done.

core/src/main/java/org/apache/spark/io/NioBasedBufferedFileInputStream.java
```diff
@@ -0,0 +1,120 @@
...
+    len = Math.min(len, byteBuffer.remaining());
+    byteBuffer.get(b, off, len);
+    return len;
+  }
+
+  @Override
+  public int available() throws IOException {
+    return byteBuffer.remaining();
+  }
+
```
sitalkedia: done

core/src/main/java/org/apache/spark/io/NioBasedBufferedFileInputStream.java
sitalkedia: done.

core/src/main/java/org/apache/spark/io/NioBasedBufferedFileInputStream.java
sitalkedia: done.

core/src/main/java/org/apache/spark/io/NioBasedBufferedFileInputStream.java
sitalkedia: Yes, we can change the condition to be `>=n`. Although the behavior remains the same because skipping 0 bytes from the underlying file channel is essentially a no-op. But changed it avoid confusion.

core/src/test/java/org/apache/spark/io/NioBasedBufferedFileInputStreamSuite.java
rxin: This reminds me --- we should add test coverage reporting to Spark so it becomes obvious what branches are not tested sufficiently.

core/src/main/java/org/apache/spark/io/NioBasedBufferedFileInputStream.java
sitalkedia: I see...that makes sense. Added an explicit call to cleanup the byte buffer.

core/src/main/java/org/apache/spark/io/NioBasedBufferedFileInputStream.java
mridulm: I agree, InputStream simply has a stub impl (just like read(byte[]) looks through read). Providing .remaining() can be useful

core/src/main/java/org/apache/spark/io/NioBasedBufferedFileInputStream.java
```diff
@@ -0,0 +1,127 @@
...
+  @Override
+  public long skip(long n) throws IOException {
+    if (n < 0L) {
```
mridulm: <= instead ?

rxin:
This is pretty low level change that requires more scrutiny, because a bug here will be very difficult to debug.

cc @zsxwing too since he's one of the most detailed oriented reviewer I know!

core/src/main/java/org/apache/spark/io/NioBasedBufferedFileInputStream.java
```diff
@@ -0,0 +1,127 @@
...
+/**
+ * {@link InputStream} implementation which uses direct buffer
+ * to read a file to avoid extra copy of data between Java and
+ * native memory which happens when using {@link java.io.BufferedInputStream}.
+ * Unfortunately, this is not something already available in JDK,
+ * {@link sun.nio.ch.ChannelInputStream} supports reading a file using nio,
+ * but does not support buffering.
+ *
+ */
+public final class NioBasedBufferedFileInputStream extends InputStream {
```
mridulm: NioBasedBufferedFileInputStream  -> NioBufferedFileInputStream ?

core/src/main/java/org/apache/spark/io/NioBasedBufferedFileInputStream.java
```diff
@@ -0,0 +1,127 @@
...
+  /**
+   * Checks weather data is left to be read from the input stream.
+   * @return true if data is left, false otherwise
+   * @throws IOException
+   */
+  private boolean refill() throws IOException {
+    if (!byteBuffer.hasRemaining()) {
+      byteBuffer.clear();
+      int nRead = fileChannel.read(byteBuffer);
+      if (nRead == -1) {
```
mridulm: <= 0 ?

core/src/main/java/org/apache/spark/io/NioBasedBufferedFileInputStream.java
```diff
@@ -0,0 +1,127 @@
...
+public final class NioBasedBufferedFileInputStream extends InputStream {
+
+  private static int DEFAULT_BUFFER_SIZE_BYTES = 8192;
+
+  private final ByteBuffer byteBuffer;
+
+  private final FileChannel fileChannel;
+
+  public NioBasedBufferedFileInputStream(File file, int bufferSizeInBytes) throws IOException {
+    byteBuffer = ByteBuffer.allocateDirect(bufferSizeInBytes);
```
mridulm: @rxin Is it possible to use netty's pooled bytebuffer allocater here instead of directly allocating ?
(Given the cost of allocating/freeing direct buffers)

mridulm:
There is a behavioral change with this PR, which I am not sure is relevant.
BufferedInputStream supports mark/reset, while we are not doing so here - does deserialization and other codepaths depend on this ? (If not for functionality, then for performance ?)

sitalkedia:
>> BufferedInputStream supports mark/reset, while we are not doing so here - does deserialization and other codepaths depend on this ? (If not for functionality, then for performance ?)

@mridulm - I looked around the code and neither any compressed input stream or deserialziation stream is using mark and reset. I have put a TODO here to support in future.

core/src/main/java/org/apache/spark/io/NioBasedBufferedFileInputStream.java
sitalkedia: done

sitalkedia:
jenkins retest this please

mridulm:
Barring query to @rxin (regarding buffer pooling), I am fine with the change - pretty neat, thanks @sitalkedia !
Would be good if more eyeballs look at it though given how fundamental it is.

rxin:
Yea pooling can make sense, but we don't do it anywhere right now so it'd make more sense to defer until we have a plan to do it more broadly.

core/src/main/java/org/apache/spark/io/NioBasedBufferedFileInputStream.java
srowen: Hm, https://docs.oracle.com/javase/7/docs/api/java/nio/channels/FileChannel.html#read(java.nio.ByteBuffer) suggests that 0 doesn't mean EOF, just 0 bytes read, but, I'm also not sure what to do if the channel won't actually give any bytes at this point. I think that can only happen if the buffer is full but that won't happen here. `<= 0` seems reasonable AFAIK.

srowen (approved):

core/src/main/java/org/apache/spark/io/NioBasedBufferedFileInputStream.java
rxin: It is possible to read 0 bytes and then next time able to read more than 0 bytes.

core/src/main/java/org/apache/spark/io/NioBufferedFileInputStream.java
```diff
@@ -0,0 +1,129 @@
...
+  /**
+   * Checks weather data is left to be read from the input stream.
+   * @return true if data is left, false otherwise
+   * @throws IOException
+   */
+  private boolean refill() throws IOException {
+    if (!byteBuffer.hasRemaining()) {
+      byteBuffer.clear();
+      int nRead = fileChannel.read(byteBuffer);
+      if (nRead <= 0) {
```
srowen: The thing is, that will just result in an error on the next read(), because nothing was put in the buffer. Is that worse? I'm not even sure. But, calling it in a loop could result in an infinite loop. I'm accustomed to stopping reading when the result is 0 in cases like this.

rxin: Actually the right thing here is to put it in a loop until we see -1.

srowen: It could be 0 forever though (dunno, FS error?) and then this creates an infinite loop. I went hunting through the SDK for some examples of dealing with this, because this has always been a tricky part of even `InputStream`.

`java.nio.Files`:

```
    private static long copy(InputStream source, OutputStream sink)
        throws IOException
    {
        long nread = 0L;
        byte[] buf = new byte[BUFFER_SIZE];
        int n;
        while ((n = source.read(buf)) > 0) {
            sink.write(buf, 0, n);
            nread += n;
        }
        return nread;
    }

...

    private static byte[] read(InputStream source, int initialSize) throws IOException {
        int capacity = initialSize;
        byte[] buf = new byte[capacity];
        int nread = 0;
        int n;
        for (;;) {
            // read to EOF which may read more or less than initialSize (eg: file
            // is truncated while we are reading)
            while ((n = source.read(buf, nread, capacity - nread)) > 0)
                nread += n;

            // if last call to source.read() returned -1, we are done
            // otherwise, try to read one more byte; if that failed we're done too
            if (n < 0 || (n = source.read()) < 0)
                break;

            // one more byte was read; need to allocate a larger buffer
            if (capacity <= MAX_BUFFER_SIZE - capacity) {
                capacity = Math.max(capacity << 1, BUFFER_SIZE);
            } else {
                if (capacity == MAX_BUFFER_SIZE)
                    throw new OutOfMemoryError("Required array size too large");
                capacity = MAX_BUFFER_SIZE;
            }
            buf = Arrays.copyOf(buf, capacity);
            buf[nread++] = (byte)n;
        }
        return (capacity == nread) ? buf : Arrays.copyOf(buf, nread);
    }
```

The first instance seems to assume that 0 bytes means it should stop.
In the second instance it forces the `InputStream` to give a definitive answer by trying to read just 1 byte. That isn't available with `FileChannel` though, unfortunately.

Now `com.google.common.io.ByteStreams`:

```
  public static long copy(InputStream from, OutputStream to)
      throws IOException {
    checkNotNull(from);
    checkNotNull(to);
    byte[] buf = new byte[BUF_SIZE];
    long total = 0;
    while (true) {
      int r = from.read(buf);
      if (r == -1) {
        break;
      }
      to.write(buf, 0, r);
      total += r;
    }
    return total;
  }

  public static int read(InputStream in, byte[] b, int off, int len)
      throws IOException {
    checkNotNull(in);
    checkNotNull(b);
    if (len < 0) {
      throw new IndexOutOfBoundsException("len is negative");
    }
    int total = 0;
    while (total < len) {
      int result = in.read(b, off + total, len - total);
      if (result == -1) {
        break;
      }
      total += result;
    }
    return total;
  }
```

Guava will only stop when it sees -1.

Aside from the JDK copy method, all of these do _not_ think that 0 means reading can stop. I'd change my answer then, yes, to reading in a loop until -1 is observed.

rxin: It is pretty standard to rely only in -1. It is especially common in networking code to get 0 bytes read.

mridulm: Note: I am not sure what happens when writes interleave with teads, whether 0 can be returned for that - probably not our usecases though  ?

sitalkedia:
@rxin, @srowen, @mridulm - The consensus seems to be to treat -1 (or less than 0) as end of file and return value of 0 to continue reading the file further, I will make the necessary changes.

sitalkedia:
jenkins retest this please.

core/src/main/java/org/apache/spark/io/NioBufferedFileInputStream.java
```diff
@@ -0,0 +1,132 @@
...
+  public int read() throws IOException {
+    if (!refill()) {
+      return -1;
+    }
+    return byteBuffer.get() & 0xFF;
+  }
+
+  @Override
+  public int read(byte[] b, int offset, int len) throws IOException {
+    if (!refill()) {
```
zsxwing: nit: please add the defense codes like BufferedInputStream
```Java
        if ((off | len | (off + len) | (b.length - (off + len))) < 0) {
            throw new IndexOutOfBoundsException();
        } else if (len == 0) {
            return 0;
        }
```

zsxwing (changes requested):
- Please add `synchronized` to the public methods except `close` (the current close method is already thread-safe). The javadoc of `InputStream` says nothing about thread-safety but the implementation like `ChannelInputStream`, `ByteArrayInputStream`, `BufferedInputStream` implies that.
- A minor behavior change is NioBufferedFileInputStream doesn't close fd in `finalize` like `FileInputStream`. But that's not a big deal since the user should do it instead of counting on GC.

core/src/main/java/org/apache/spark/io/NioBufferedFileInputStream.java
sitalkedia: done.

core/src/main/java/org/apache/spark/io/NioBufferedFileInputStream.java
```diff
@@ -0,0 +1,132 @@
...
+    // Discard everything we have read in the buffer.
+    byteBuffer.position(0);
+    byteBuffer.flip();
+    return skippedFromBuffer + skipFromFileChannel(toSkipFromFileChannel);
+  }
+
+  private long skipFromFileChannel(long n) throws IOException {
+    long currentFilePosition = fileChannel.position();
+    long size = fileChannel.size();
+    if (currentFilePosition + n > size) {
```
sitalkedia: good point, changed accordingly.

core/src/main/java/org/apache/spark/io/NioBufferedFileInputStream.java
```diff
@@ -72,30 +74,35 @@ private boolean refill() throws IOException {
...
     return byteBuffer.get() & 0xFF;
   }
 
   @Override
-  public int read(byte[] b, int offset, int len) throws IOException {
+  public synchronized int read(byte[] b, int offset, int len) throws IOException {
     if (!refill()) {
       return -1;
     }
+    if ((offset | len | (offset + len) | (b.length - (offset + len))) < 0) {
```
srowen: I think this must be a typo... this is a bitwise-or of a bunch of ints. I think maybe the example was given as a sort of pseudo code. offset and len can't be negative and their sum shouldn't exceed the array lenght, but that seems like it. 'else' below isn't really needed, but that's a nit. Actually, you want to check whether len is after updating on line 94 right? if no data is available you also return 0 without a no-op read?

srowen (changes requested):
I also missed why close() shouldn't be synchronized. You wouldn't want it to take place during a read, right?

I think a finalizer wouldn't hurt here because `FileInputStream` does it. It's exactly because people may not properly close the stream. 

WDYT @zsxwing ?

Getting quite close here I think.

core/src/main/java/org/apache/spark/io/NioBufferedFileInputStream.java
zsxwing: @srowen I think this check is correct. It checks if any of these values is negative.

> offset and len can't be negative and their sum shouldn't exceed the array length.

It's possible. `offset` and `len` are set by the caller.

@sitalkedia Could you move the argument check before `refill`? If the argument is invalid, we should always throw an exception instead of returning `-1`.

zsxwing:
> I also missed why close() shouldn't be synchronized. You wouldn't want it to take place during a read, right?

Good point. I was thinking `FileChannel.close` is safe but didn't consider this case.

> I think a finalizer wouldn't hurt here because FileInputStream does it. It's exactly because people may not properly close the stream.

Agreed that it doesn't hurt the performance since FileInputStream does it.

core/src/main/java/org/apache/spark/io/NioBufferedFileInputStream.java
srowen: @zsxwing oh I get it, you're really just checking if any top bit is set. Hm, that seems fairly obscure compared to just checking the arguments. Sure it's a little more code, but it's readable and i don't think this is performance critical? I hadn't seen this before and had to think a while to get what it does.

Yes, I'm agreeing that offset and len can't be _allowed_ to be negative, not that it can't be specified by the caller as something negative.

srowen: Oh I see. Well it's reasonable to keep it then, but I'd at least add a comment I think. I didn't recognize this at all until I thought about it for a good few minutes.

sitalkedia: @srowen - Agreed, the condition was little confusing. Changed it to make it clearer.

sitalkedia:
jenkins retest this please.

core/src/main/java/org/apache/spark/io/NioBufferedFileInputStream.java
```diff
@@ -0,0 +1,142 @@
...
+ * to read a file to avoid extra copy of data between Java and
+ * native memory which happens when using {@link java.io.BufferedInputStream}.
+ * Unfortunately, this is not something already available in JDK,
+ * {@link sun.nio.ch.ChannelInputStream} supports reading a file using nio,
+ * but does not support buffering.
+ *
+ * TODO: support {@link #mark(int)}/{@link #reset()}
+ *
+ */
+@ThreadSafe
```
srowen: I only see one location in the codebase where we use this annotation, and I think we probably shouldn't use it at all if not used consistently.

core/src/main/java/org/apache/spark/io/NioBufferedFileInputStream.java
```diff
@@ -0,0 +1,142 @@
...
+public final class NioBufferedFileInputStream extends InputStream {
+
+  private static int DEFAULT_BUFFER_SIZE_BYTES = 8192;
```
srowen: Oops, forgot to say this should be `final`

core/src/main/java/org/apache/spark/io/NioBufferedFileInputStream.java
```diff
@@ -0,0 +1,142 @@
...
+
+/**
+ * {@link InputStream} implementation which uses direct buffer
+ * to read a file to avoid extra copy of data between Java and
+ * native memory which happens when using {@link java.io.BufferedInputStream}.
+ * Unfortunately, this is not something already available in JDK,
+ * {@link sun.nio.ch.ChannelInputStream} supports reading a file using nio,
+ * but does not support buffering.
+ *
+ * TODO: support {@link #mark(int)}/{@link #reset()}
```
srowen: I don't really care, but this could be a comment inside the class rather than user-facing. In fact I don't even know it's a to-do.

core/src/main/java/org/apache/spark/io/NioBufferedFileInputStream.java
```diff
@@ -0,0 +1,142 @@
...
+  public synchronized int read() throws IOException {
+    if (!refill()) {
+      return -1;
+    }
+    return byteBuffer.get() & 0xFF;
+  }
+
+  @Override
+  public synchronized int read(byte[] b, int offset, int len) throws IOException {
+    if (offset < 0 || len < 0 || (offset + len) < 0 || (b.length - (offset + len)) < 0) {
```
srowen: Hardly matters, but now that this condition has been made more explicit, then final condition is simpler as `offset + len > b.length`

core/src/main/java/org/apache/spark/io/NioBufferedFileInputStream.java
```diff
@@ -0,0 +1,142 @@
...
+
+  @Override
+  public synchronized void close() throws IOException {
+    fileChannel.close();
+    StorageUtils.dispose(byteBuffer);
+  }
+
+  @Override
+  protected void finalize() throws IOException {
+      close();
```
srowen: Nit: 2-space indent not 4

srowen (changes requested):
Now we're down to pretty trivial changes, thank you. This will be a nice win and I am glad we've scrutinized it because it will be in a lot of critical paths.

core/src/main/java/org/apache/spark/io/NioBufferedFileInputStream.java
sitalkedia: Okay, I removed the TODO here.

core/src/main/java/org/apache/spark/io/NioBufferedFileInputStream.java
sitalkedia: Alright, removed it for consistency.

core/src/main/java/org/apache/spark/io/NioBufferedFileInputStream.java
sitalkedia: done

core/src/main/java/org/apache/spark/io/NioBufferedFileInputStream.java
```diff
@@ -0,0 +1,138 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.spark.io;
+
+import org.apache.spark.storage.StorageUtils;
+
+import javax.annotation.concurrent.ThreadSafe;
```
srowen: Nit^2 : no longer needed as an import

core/src/main/java/org/apache/spark/io/NioBufferedFileInputStream.java
```diff
@@ -0,0 +1,138 @@
...
+  public synchronized int read() throws IOException {
+    if (!refill()) {
+      return -1;
+    }
+    return byteBuffer.get() & 0xFF;
+  }
+
+  @Override
+  public synchronized int read(byte[] b, int offset, int len) throws IOException {
+    if (offset < 0 || len < 0 || (b.length - (offset + len)) < 0) {
```
srowen: Ah no I think that condition was needed. I mean: `if (offset < 0 || len < 0 || offset + len < 0 || offset + len > b.length) {`

srowen (changes requested):

core/src/main/java/org/apache/spark/io/NioBufferedFileInputStream.java
sitalkedia: good eye, wonder why the checkstyle did not catch it though?

core/src/main/java/org/apache/spark/io/NioBufferedFileInputStream.java
sitalkedia: Ignore my previous comments, we still need it.

srowen (approved):
LGTM pending tests

srowen:
Going once, going twice, any more comments?

zsxwing:

PR #15409, opened 2016-10-10
[Spark-14761][SQL] Reject invalid join methods  when join columns are not specified in PySpark DataFrame join.

bkpathak:

In PySpark, the invalid join type will not throw error for the following join:
`df1.join(df2, how='not-a-valid-join-type')`

The signature of the join is:
`def join(self, other, on=None, how=None):`
The existing code completely ignores the `how` parameter when `on` is `None`. This patch will process the arguments passed to join and pass in to JVM Spark SQL Analyzer, which will validate the join type passed.
## How was this patch tested?

Used manual and existing test suites.

### DIFF
--- a/python/pyspark/sql/dataframe.py
+++ b/python/pyspark/sql/dataframe.py
@@ -640,25 +640,23 @@ class DataFrame(object):
         if on is not None and not isinstance(on, list):
             on = [on]
 
-        if on is None or len(on) == 0:
+        if on is not None:
+            if isinstance(on[0], basestring):
+                on = self._jseq(on)
+            else:
+                assert isinstance(on[0], Column), "on should be Column or list of Column"
+                if len(on) > 1:
+                    on = reduce(lambda x, y: x.__and__(y), on)
+                else:
+                    on = on[0]
+
+        if on is None and how is None:
             jdf = self._jdf.crossJoin(other._jdf)
-        elif isinstance(on[0], basestring):
-            if how is None:
-                jdf = self._jdf.join(other._jdf, self._jseq(on), "inner")
-            else:
-                assert isinstance(how, basestring), "how should be basestring"
-                jdf = self._jdf.join(other._jdf, self._jseq(on), how)
         else:
-            assert isinstance(on[0], Column), "on should be Column or list of Column"
-            if len(on) > 1:
-                on = reduce(lambda x, y: x.__and__(y), on)
-            else:
-                on = on[0]
             if how is None:
-                jdf = self._jdf.join(other._jdf, on._jc, "inner")
-            else:
-                assert isinstance(how, basestring), "how should be basestring"
-                jdf = self._jdf.join(other._jdf, on._jc, how)
+                how = "inner"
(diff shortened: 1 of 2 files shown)


bkpathak:
@holdenk @JoshRosen Can you please review this pull requst.

python/pyspark/sql/tests.py
```diff
@@ -1508,6 +1508,23 @@ def test_toDF_with_schema_string(self):
         self.assertEqual(df.schema.simpleString(), "struct<value:int>")
         self.assertEqual(df.collect(), [Row(key=i) for i in range(100)])
 
+    # Regression test for invalid join methods when on is None, Spark-14761
+    def test_invalid_join_method(self):
+        df1 = self.sqlCtx.createDataFrame([("Alice", 5), ("Bob", 8)], ["name", "age"])
+        df2 = self.sqlCtx.createDataFrame([("Alice", 80), ("Bob", 90)], ["name", "height"])
+        self.assertRaises(AnalysisException, lambda: df1.join(df2, how="invalid-join-type"))
+
+        result = df1.join(df2, how="inner").select(df1.name, df2.height).collect()
```
rxin: can we remove everything from this test onward? they are no longer testing invalid join methods.

rxin:
The change itself LGTM, but also cc @srinathshankar.

Right now Python behavior differs from Scala with respect to how crossJoin is handled.

rxin:
Oh well the test cases have issues. You can run those by `python/run-tests --module pyspark-sql`

bkpathak:
Hi @rxin, I removed those lines and fixed the errors. @srinathshankar

rxin:
Thanks!

bkpathak:
@rxin , @srinathshankar Could you please review the fix.

python/pyspark/sql/dataframe.py
```diff
@@ -640,25 +640,24 @@ def join(self, other, on=None, how=None):
...
-                jdf = self._jdf.join(other._jdf, self._jseq(on), "inner")
+        if on is not None:
+            if isinstance(on[0], basestring):
+                on = self._jseq(on)
             else:
-                assert isinstance(how, basestring), "how should be basestring"
-                jdf = self._jdf.join(other._jdf, self._jseq(on), how)
+                assert isinstance(on[0], Column), "on should be Column or list of Column"
+                if len(on) > 1:
+                    on = reduce(lambda x, y: x.__and__(y), on)
```
srinathshankar: Wouldn't reduce work even if len(on) == 1. i.e. do you need an if here ?

srinathshankar (approved):
LGTM, just a minor question

rxin:
@srinathshankar brought up a good point. Since it is a small change and doesnt' impact correctness, I'm going to merge this in master.

@srinathshankar when you add the Python API for crossjoin, can you fix this tiny thing?

python/pyspark/sql/dataframe.py
srinathshankar: I could fix this too.

PR #15412, opened 2016-10-10
[SPARK-17844] Simplify DataFrame API for defining frame boundaries in window functions

rxin:

When I was creating the example code for SPARK-10496, I realized it was pretty convoluted to define the frame boundaries for window functions when there is no partition column or ordering column. The reason is that we don't provide a way to create a WindowSpec directly with the frame boundaries. We can trivially improve this by adding rowsBetween and rangeBetween to Window object.

As an example, to compute cumulative sum using the natural ordering, before this pr:

```
df.select('key, sum("value").over(Window.partitionBy(lit(1)).rowsBetween(Long.MinValue, 0)))
```

After this pr:

```
df.select('key, sum("value").over(Window.rowsBetween(Long.MinValue, 0)))
```

Note that you could argue there is no point specifying a window frame without partitionBy/orderBy -- but it is strange that only rowsBetween and rangeBetween are not the only two APIs not available.

This also fixes https://issues.apache.org/jira/browse/SPARK-17656 (removing _root_.scala).
## How was this patch tested?

Added test cases to compute cumulative sum in DataFrameWindowSuite for Scala/Java and tests.py for Python.

### DIFF
--- a/python/pyspark/sql/tests.py
+++ b/python/pyspark/sql/tests.py
@@ -1859,6 +1859,15 @@ class HiveContextSQLTests(ReusedPySparkTestCase):
         for r, ex in zip(rs, expected):
             self.assertEqual(tuple(r), ex[:len(r)])
 
+    def test_window_functions_cumulative_sum(self):
+        df = self.spark.createDataFrame([("one", 1), ("two", 2)], ["key", "value"])
+        from pyspark.sql import functions as F
+        sel = df.select(df.key, F.sum(df.value).over(Window.rowsBetween(-sys.maxsize, 0)))
+        rs = sorted(sel.collect())
+        expected = [("one", 1), ("two", 3)]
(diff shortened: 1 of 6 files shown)


HyukjinKwon:
Hi @rxin , I just happened to look at this PR. I just want to leave a gentle reminder just in case, that there are [SPARK-17656](https://issues.apache.org/jira/browse/SPARK-17656) and two more cases in `./sql/core/src/main/scala/org/apache/spark/sql/expressions/udaf.scala` (Maybe this is not directly relevant with this PR but just when I saw the changes here, it rang a bell to me and I just wanted to let you know just in case).

rxin:
Sure I can fix those in this pull request too. Thanks for the reminder.

rxin:
cc @hvanhovell ?

hvanhovell:
LGTM - merging to master.

PR #15413, opened 2016-10-10
[SPARK-17847][ML] Reduce shuffled data size of GaussianMixture & copy the implementation from mllib to ml

yanboliang:

Copy `GaussianMixture` implementation from mllib to ml, then we can add new features to it.
I left mllib `GaussianMixture` untouched, unlike some other algorithms to wrap the ml implementation. For the following reasons:
- mllib `GaussianMixture` allows k == 1, but ml does not.
- mllib `GaussianMixture` supports setting initial model, but ml does not support currently. (We will definitely add this feature for ml in the future)

We can get around these issues to make mllib as a wrapper calling into ml, but I'd prefer to leave mllib untouched which can make ml clean.

Meanwhile, There is a big performance improvement for `GaussianMixture` in this PR. Since the covariance matrix of multivariate gaussian distribution is symmetric, we can only store the upper triangular part of the matrix and it will greatly reduce the shuffled data size. In my test, this change will reduce shuffled data size by about 50% and accelerate the job execution.

Before this PR:
image
After this PR:
image
## How was this patch tested?

Existing tests and added new tests.

### DIFF
--- a/mllib/src/main/scala/org/apache/spark/ml/clustering/GaussianMixture.scala
+++ b/mllib/src/main/scala/org/apache/spark/ml/clustering/GaussianMixture.scala
@@ -347,6 +347,7 @@ class GaussianMixture @Since("2.0.0") (
 
     val instr = Instrumentation.create(this, instances)
     instr.logParams(featuresCol, predictionCol, probabilityCol, k, maxIter, seed, tol)
+    instr.logNumFeatures(numFeatures)
 
     val shouldDistributeGaussians = GaussianMixture.shouldDistributeGaussians(
       numClusters, numFeatures)
@@ -409,8 +410,7 @@ class GaussianMixture @Since("2.0.0") (
     }
 
     val gaussianDists = gaussians.map { case (mean, covVec) =>
-      val cov = new DenseMatrix(numFeatures, numFeatures,
-        GaussianMixture.unpackUpperTriangularMatrix(numFeatures, covVec.values))
+      val cov = GaussianMixture.unpackUpperTriangularMatrix(numFeatures, covVec.values)
       new MultivariateGaussian(mean, cov)
     }
 
@@ -418,7 +418,6 @@ class GaussianMixture @Since("2.0.0") (
     val summary = new GaussianMixtureSummary(model.transform(dataset),
       $(predictionCol), $(probabilityCol), $(featuresCol), $(k))
     model.setSummary(Some(summary))
-    instr.logNumFeatures(model.gaussians.head.mean.size)
     instr.logSuccess(model)
     model
   }
@@ -439,7 +438,7 @@ class GaussianMixture @Since("2.0.0") (
    * @param numFeatures The number of features of training instance.
    * @return The initialized weights and corresponding gaussian distributions. Note the
    *         covariance matrix of multivariate gaussian distribution is symmetric and
-   *         we only save the upper triangular part as a dense vector.
+   *         we only save the upper triangular part as a dense vector (column major).
    */
   private def initRandom(
       instances: RDD[Vector],
@@ -463,8 +462,8 @@ class GaussianMixture @Since("2.0.0") (
          Construct matrix where diagonal entries are element-wise
          variance of input vectors (computes biased variance).
          Since the covariance matrix of multivariate gaussian distribution is symmetric,
-         only the upper triangular part of the matrix will be saved as a dense vector
-         in order to reduce the shuffled data size.
+         only the upper triangular part of the matrix (column major) will be saved as
+         a dense vector in order to reduce the shuffled data size.
        */
       val cov = {
         val ss = new DenseVector(new Array[Double](numFeatures)).asBreeze
@@ -505,20 +504,20 @@ object GaussianMixture extends DefaultParamsReadable[GaussianMixture] {
 
   /**
    * Convert an n * (n + 1) / 2 dimension array representing the upper triangular part of a matrix
-   * into an n * n array representing the full symmetric matrix.
+   * into an n * n array representing the full symmetric matrix (column major).
    *
    * @param n The order of the n by n matrix.
    * @param triangularValues The upper triangular part of the matrix packed in an array
    *                         (column major).
-   * @return An array which represents the symmetric matrix in column major.
+   * @return A dense matrix which represents the symmetric matrix in column major.
    */
   private[clustering] def unpackUpperTriangularMatrix(
       n: Int,
-      triangularValues: Array[Double]): Array[Double] = {
+      triangularValues: Array[Double]): DenseMatrix = {
     val symmetricValues = new Array[Double](n * n)
     var r = 0
     var i = 0
-    while(i < n) {
+    while (i < n) {
       var j = 0
       while (j <= i) {
         symmetricValues(i * n + j) = triangularValues(r)
@@ -528,7 +527,7 @@ object GaussianMixture extends DefaultParamsReadable[GaussianMixture] {
       }
       i += 1
     }
-    symmetricValues
+    new DenseMatrix(n, n, symmetricValues)
   }
 
   /**
@@ -536,7 +535,7 @@ object GaussianMixture extends DefaultParamsReadable[GaussianMixture] {
    *
    * @param mean The mean of the gaussian distribution.
    * @param cov The covariance matrix of the gaussian distribution. Note we only
-   *            save the upper triangular part as a dense vector.
+   *            save the upper triangular part as a dense vector (column major).
    * @param weight The weight of the gaussian distribution.
    * @param sumWeights The sum of weights of all clusters.
    * @return The updated weight, mean and covariance.
@@ -562,8 +561,8 @@ object GaussianMixture extends DefaultParamsReadable[GaussianMixture] {
  * @param bcWeights The broadcast weights for each Gaussian distribution in the mixture.
  * @param bcGaussians The broadcast array of Multivariate Gaussian (Normal) Distribution
  *                    in the mixture. Note only upper triangular part of the covariance
- *                    matrix of each distribution is stored as dense vector in order to
- *                    reduce shuffled data size.
+ *                    matrix of each distribution is stored as dense vector (column major)
+ *                    in order to reduce shuffled data size.
  */
 private class ExpectationAggregator(
     numFeatures: Int,
@@ -581,8 +580,7 @@ private class ExpectationAggregator(
 
   @transient private lazy val oldGaussians = {
     bcGaussians.value.map { case (mean, covVec) =>
-      val cov = new DenseMatrix(numFeatures, numFeatures,
-        GaussianMixture.unpackUpperTriangularMatrix(numFeatures, covVec.values))
+      val cov = GaussianMixture.unpackUpperTriangularMatrix(numFeatures, covVec.values)
       new MultivariateGaussian(mean, cov)
     }
   }
@@ -611,7 +609,7 @@ private class ExpectationAggregator(
     val prob = new Array[Double](k)
     var probSum = 0.0
     var i = 0
-    while(i < k) {
+    while (i < k) {
       val p = EPSILON + localWeights(i) * localOldGaussians(i).pdf(instance)
       prob(i) = p
       probSum += p
@@ -623,7 +621,7 @@ private class ExpectationAggregator(
     val localNewMeans = newMeans
     val localNewCovs = newCovs
     i = 0
-    while(i < k) {
+    while (i < k) {
       prob(i) /= probSum
       localNewWeights(i) += prob(i)
       BLAS.axpy(prob(i), instance, localNewMeans(i))
@@ -654,7 +652,7 @@ private class ExpectationAggregator(
       val localThisNewCovs = this.newCovs
       val localOtherNewCovs = other.newCovs
       var i = 0
-      while(i < k) {
+      while (i < k) {
         localThisNewWeights(i) += localOtherNewWeights(i)
         BLAS.axpy(1.0, localOtherNewMeans(i), localThisNewMeans(i))
         BLAS.axpy(1.0, localOtherNewCovs(i), localThisNewCovs(i))
--- a/mllib/src/main/scala/org/apache/spark/ml/clustering/GaussianMixture.scala
+++ b/mllib/src/main/scala/org/apache/spark/ml/clustering/GaussianMixture.scala
@@ -351,7 +351,7 @@ class GaussianMixture @Since("2.0.0") (
     val shouldDistributeGaussians = GaussianMixture.shouldDistributeGaussians(
       numClusters, numFeatures)
 
-    // TODO: Support users supplied initial GMM.
+    // TODO: SPARK-15785 Support users supplied initial GMM.
     val (weights, gaussians) = initRandom(instances, numClusters, numFeatures)
 
     var logLikelihood = Double.MinValue
@@ -429,17 +429,17 @@ class GaussianMixture @Since("2.0.0") (
   }
 
   /**
-   * Initialize weights and corresponding gaussians at random.
+   * Initialize weights and corresponding gaussian distributions at random.
    *
    * We start with uniform weights, a random mean from the data, and diagonal covariance matrices
    * using component variances derived from the samples.
    *
-   * @param instances The instances of training data.
+   * @param instances The training instances.
    * @param numClusters The number of clusters.
-   * @param numFeatures The number of features in training data.
-   * @return The initialized weights and corresponding gaussians. Note the covariance matrix of
-   *         multivariate gaussian distribution is symmetric and we only save the upper triangular
-   *         part as a dense vector.
+   * @param numFeatures The number of features of training instance.
+   * @return The initialized weights and corresponding gaussian distributions. Note the
+   *         covariance matrix of multivariate gaussian distribution is symmetric and
+   *         we only save the upper triangular part as a dense vector.
    */
   private def initRandom(
       instances: RDD[Vector],
--- a/mllib/src/main/scala/org/apache/spark/ml/clustering/GaussianMixture.scala
+++ b/mllib/src/main/scala/org/apache/spark/ml/clustering/GaussianMixture.scala
@@ -353,11 +353,16 @@ class GaussianMixture @Since("2.0.0") (
     // derived from the samples.
     // TODO: Support users supplied initial GMM.
     val samples = instances.takeSample(withReplacement = true, _k * nSamples, $(seed))
-    val weights = Array.fill(_k)(1.0 / _k)
-    val gaussians = Array.tabulate(_k) { i =>
+    val weights: Array[Double] = Array.fill(_k)(1.0 / _k)
+    /**
+     * Since the covariance matrix of multivariate gaussian distribution is symmetric,
+     * only the upper triangular part of the matrix will be stored as a dense vector
+     * in order to reduce the shuffled data size.
+     */
+    val gaussians: Array[(DenseVector, DenseVector)] = Array.tabulate(_k) { i =>
       val slice = samples.view(i * nSamples, (i + 1) * nSamples)
       val mean = {
-        val v = Vectors.zeros(numFeatures)
+        val v = new DenseVector(Array.fill[Double](numFeatures)(0.0))
         var i = 0
         while (i < nSamples) {
           BLAS.axpy(1.0, slice(i), v)
@@ -368,16 +373,21 @@ class GaussianMixture @Since("2.0.0") (
       }
       /**
        * Construct matrix where diagonal entries are element-wise
-       * variance of input vectors (computes biased variance)
+       * variance of input vectors (computes biased variance).
        */
       val cov = {
-        val ss = Vectors.zeros(numFeatures).asBreeze
+        val ss = new DenseVector(Array.fill[Double](numFeatures)(0)).asBreeze
         slice.foreach(xi => ss += (xi.asBreeze - mean.asBreeze) :^ 2.0)
         val diagVec = Vectors.fromBreeze(ss)
         BLAS.scal(1.0 / nSamples, diagVec)
-        Matrices.diag(diagVec)
+        val covVec = new DenseVector(Array.fill[Double](
+          numFeatures * (numFeatures + 1) / 2)(0.0))
+        diagVec.toArray.zipWithIndex.foreach { case (v: Double, i: Int) =>
+          covVec.values(i + i * (i + 1) / 2) = v
+        }
+        covVec
       }
-      new MultivariateGaussian(mean, cov)
+      (mean, cov)
     }
 
     var llh = Double.MinValue // current log-likelihood
@@ -434,7 +444,13 @@ class GaussianMixture @Since("2.0.0") (
       iter += 1
     }
 
-    val model = copyValues(new GaussianMixtureModel(uid, weights, gaussians)).setParent(this)
+    val gaussianDists = gaussians.map { case (mean, covVec) =>
+      val cov = new DenseMatrix(numFeatures, numFeatures,
+        GaussianMixture.unpackUpperTriangularMatrix(numFeatures, covVec.values))
+      new MultivariateGaussian(mean, cov)
+    }
+
+    val model = copyValues(new GaussianMixtureModel(uid, weights, gaussianDists)).setParent(this)
     val summary = new GaussianMixtureSummary(model.transform(dataset),
       $(predictionCol), $(probabilityCol), $(featuresCol), $(k))
     model.setSummary(Some(summary))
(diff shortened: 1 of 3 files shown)


yanboliang:
cc @sethah @jkbradley

sethah:
Do you plan to run performance tests to ensure there are no regressions?

yanboliang:
@sethah I did some performance tests actually, and found this change can improve by 1.5x ~ 2x according to different dimensions and number of clusters. I will post the test result soon. Thanks.

mllib/src/main/scala/org/apache/spark/ml/clustering/GaussianMixture.scala
```diff
@@ -316,24 +319,129 @@ class GaussianMixture @Since("2.0.0") (
...
+  private val nSamples = 5
+
   @Since("2.0.0")
   override def fit(dataset: Dataset[_]): GaussianMixtureModel = {
     transformSchema(dataset.schema, logging = true)
-    val rdd: RDD[OldVector] = dataset.select(col($(featuresCol))).rdd.map {
-      case Row(point: Vector) => OldVectors.fromML(point)
+
+    val sc = dataset.sparkSession.sparkContext
+    val _k = $(k)
```
sethah: maybe `nClusters`? I don't have a strong preference but I think it's clearer when reading the code.

mllib/src/main/scala/org/apache/spark/ml/clustering/GaussianMixture.scala
```diff
@@ -316,24 +319,129 @@ class GaussianMixture @Since("2.0.0") (
...
+    // Extract the number of features.
+    val numFeatures = instances.first().size
+
+    val shouldDistributeGaussians = GaussianMixture.shouldDistributeGaussians(_k, numFeatures)
+
+    // Determine initial weights and corresponding Gaussians.
+    // We start with uniform weights, a random mean from the data, and
+    // diagonal covariance matrices using component variances
+    // derived from the samples.
+    // TODO: Support users supplied initial GMM.
```
sethah: Let's make a JIRA for it and reference that

mllib/src/main/scala/org/apache/spark/ml/clustering/GaussianMixture.scala
```diff
@@ -316,24 +319,129 @@ class GaussianMixture @Since("2.0.0") (
...
-      case Row(point: Vector) => OldVectors.fromML(point)
+
+    val sc = dataset.sparkSession.sparkContext
+    val _k = $(k)
+
+    val instances: RDD[Vector] = dataset.select(col($(featuresCol))).rdd.map {
+      case Row(features: Vector) => features
+    }.cache()
+
+    // Extract the number of features.
```
sethah: this comment adds no value IMO

mllib/src/main/scala/org/apache/spark/ml/clustering/GaussianMixture.scala
```diff
@@ -350,6 +458,145 @@ object GaussianMixture extends DefaultParamsReadable[GaussianMixture] {
...
+   * @param n The order of the n by n matrix.
+   * @param triangular The upper triangular part of the matrix packed in an array (column major).
+   * @return An array which represents the symmetric matrix in column major.
+   */
+  private[clustering] def unpackUpperTriangularMatrix(
+      n: Int,
+      triangular: Array[Double]): Array[Double] = {
+    val symmetric = Array.fill(n * n)(0.0)
+    var r = 0
+    for (i <- 0 until n) {
```
sethah: let's use while since this is called in several places

mllib/src/main/scala/org/apache/spark/ml/clustering/GaussianMixture.scala
```diff
@@ -350,6 +458,145 @@ object GaussianMixture extends DefaultParamsReadable[GaussianMixture] {
...
+   *
+   * @param k  Number of topics
+   * @param d  Number of features
+   */
+  private[clustering] def shouldDistributeGaussians(k: Int, d: Int): Boolean = {
+    ((k - 1.0) / k) * d > 25
+  }
+
+  /**
+   * Unpack upper triangular part of a symmetric matrix.
```
sethah: "Convert an n * (n + 1) / 2 dimension array representing the upper triangular part of a matrix into an n * n array representing the full symmetric matrix". I think that's more explicit about what is happening. Also very minor nit, can we call the array `triangularValues` instead? `triangular` sounds like it should be a boolean to me.

mllib/src/main/scala/org/apache/spark/ml/clustering/GaussianMixture.scala
```diff
@@ -350,6 +458,145 @@ object GaussianMixture extends DefaultParamsReadable[GaussianMixture] {
...
+
+  /**
+   * Heuristic to distribute the computation of the [[MultivariateGaussian]]s, approximately when
+   * d > 25 except for when k is very small.
+   *
+   * @param k  Number of topics
+   * @param d  Number of features
+   */
+  private[clustering] def shouldDistributeGaussians(k: Int, d: Int): Boolean = {
+    ((k - 1.0) / k) * d > 25
```
sethah: nit: 25.0. Also, let's call it `numFeatures` instead of `d`

mllib/src/main/scala/org/apache/spark/ml/clustering/GaussianMixture.scala
```diff
@@ -316,24 +319,129 @@ class GaussianMixture @Since("2.0.0") (
...
 
-    val algo = new MLlibGM()
-      .setK($(k))
-      .setMaxIterations($(maxIter))
-      .setSeed($(seed))
-      .setConvergenceTol($(tol))
-    val parentModel = algo.run(rdd)
-    val gaussians = parentModel.gaussians.map { case g =>
-      new MultivariateGaussian(g.mu.asML, g.sigma.asML)
+    var llh = Double.MinValue // current log-likelihood
```
sethah: minor: why not `logLikelihood` and `logLikelihoodPrev` ? It's nice to have descriptive variable names, then we can remove the comments.

mllib/src/main/scala/org/apache/spark/ml/clustering/GaussianMixture.scala
```diff
@@ -350,6 +458,145 @@ object GaussianMixture extends DefaultParamsReadable[GaussianMixture] {
...
+    BLAS.spr(-weight, mean, cov)
+    BLAS.scal(1.0 / weight, cov)
+    val newWeight = weight / sumWeights
+    val newGaussian = (mean, cov)
+    (newWeight, newGaussian)
+  }
+}
+
+/**
+ * ExpectationAggregator computes the partial expectation results.
```
sethah: We have typically used this documentation as a place to explain the math used to compute updates. It would be nice to have that here as well.

sethah (commented):
Made a first pass. I haven't looked at the tests yet. 

Also, about keeping the mllib code around. It would be really quite simple to get around the issues you mentioned. We can do like `LogisticRegression` and make a `private var optInitialModel` for now. For k, we could we could make an alternate private train constructor which takes k as an argument. Still, I'm ok with leaving it for a future PR, but I don't think it's being blocked by those issues. Let me know what you think on this issue.

jkbradley:
As far as keeping the code around, I much prefer either the current approach (separate code) or having spark.mllib call into spark.ml.  That will make it easier to deprecate and eventually remove spark.mllib code in 3.0.

I like the upper-triangular matrix packing and unpacking!  Could you please add a unit test for it?

mllib/src/main/scala/org/apache/spark/ml/clustering/GaussianMixture.scala
```diff
@@ -350,6 +458,145 @@ object GaussianMixture extends DefaultParamsReadable[GaussianMixture] {
...
+ *                    reduce shuffled data size.
+ */
+private class ExpectationAggregator(
+    numFeatures: Int,
+    bcWeights: Broadcast[Array[Double]],
+    bcGaussians: Broadcast[Array[(DenseVector, DenseVector)]]) extends Serializable {
+
+  private val k: Int = bcWeights.value.length
+  private var totalCnt: Long = 0L
+  private var newLogLikelihood: Double = 0.0
```
yanboliang: ```ExpectationAggregator``` accepts the old weights, gaussians and outputs the new ones, I added prefix to make it more clear.

mllib/src/main/scala/org/apache/spark/ml/clustering/GaussianMixture.scala
```diff
@@ -350,6 +458,145 @@ object GaussianMixture extends DefaultParamsReadable[GaussianMixture] {
...
+
+    val p = localWeights.zip(oldGaussians).map { case (weight, gaussian) =>
+      EPSILON + weight * gaussian.pdf(data)
+    }
+    val pSum = p.sum
+    newLogLikelihood += math.log(pSum)
+    var i = 0
+    while(i < k) {
+      p(i) /= pSum
+      newWeights(i) += p(i)
```
yanboliang: I thought the number of iterations(equals to the number of cluster) is small enough to ignore the impact, compared with the dimension of gradient in LiR/LoR which usually as large as millions or billions, but it's better we can avoid the extra cost.

python/pyspark/ml/clustering.py
```diff
@@ -95,15 +95,10 @@ class GaussianMixture(JavaEstimator, HasFeaturesCol, HasPredictionCol, HasMaxIte
...
-    >>> model.gaussiansDF.show()
-    +--------------------+--------------------+
-    |                mean|                 cov|
-    +--------------------+--------------------+
-    |[0.82500000140229...|0.005625000000006...|
-    |[-0.4777098016092...|0.167969502720916...|
-    |[-0.4472625243352...|0.167304119758233...|
-    +--------------------+--------------------+
-    ...
+    >>> model.gaussiansDF.select("mean").head()
```
yanboliang: The test is unstable, and it's not necessary to check so many digits.

mllib/src/test/scala/org/apache/spark/ml/clustering/GaussianMixtureSuite.scala
```diff
@@ -130,4 +230,29 @@ object GaussianMixtureSuite {
     "maxIter" -> 2,
     "tol" -> 0.01
   )
+
+  val denseData = Seq(
```
yanboliang: We should copy test cases and data from mllib to ml when we copying algorithm implementation. Since the spark.mllib package would be deleted eventually(may be in 3.0?), we should keep spark.ml includes all necessary data.

yanboliang:
@sethah Yeah, I totally agree we can get around the issues I mentioned and make ```mllib.GaussianMixture``` call into ```ml.GaussianMixture```. But I'd prefer more to the current approach based on the experience of moving ```NaiveBayes```. Still, we can discuss this issue in a future task and let's this PR focus on the improved implementation.
@jkbradley I added test for upper-triangular matrix unpacking.

yanboliang:
@jkbradley @sethah Any more comments? Thanks.

jkbradley:
I'll take a look, thanks for pinging!

mllib/src/main/scala/org/apache/spark/ml/clustering/GaussianMixture.scala
```diff
@@ -356,13 +427,243 @@ class GaussianMixture @Since("2.0.0") (
...
+   *
+   * We start with uniform weights, a random mean from the data, and diagonal covariance matrices
+   * using component variances derived from the samples.
+   *
+   * @param instances The training instances.
+   * @param numClusters The number of clusters.
+   * @param numFeatures The number of features of training instance.
+   * @return The initialized weights and corresponding gaussian distributions. Note the
+   *         covariance matrix of multivariate gaussian distribution is symmetric and
+   *         we only save the upper triangular part as a dense vector.
```
jkbradley: Elsewhere too (e.g., in ExpectationAggregator)

mllib/src/main/scala/org/apache/spark/ml/clustering/GaussianMixture.scala
```diff
@@ -356,13 +427,243 @@ class GaussianMixture @Since("2.0.0") (
...
+   * @return This ExpectationAggregator object.
+   */
+  def add(instance: Vector): this.type = {
+    val localWeights = bcWeights.value
+    val localOldGaussians = oldGaussians
+
+    val prob = new Array[Double](k)
+    var probSum = 0.0
+    var i = 0
+    while(i < k) {
```
jkbradley: style: space after "while"

mllib/src/main/scala/org/apache/spark/ml/clustering/GaussianMixture.scala
```diff
@@ -356,13 +427,243 @@ class GaussianMixture @Since("2.0.0") (
...
+   * (Note that it's in place merging; as a result, `this` object will be modified.)
+   *
+   * @param other The other ExpectationAggregator to be merged.
+   * @return This ExpectationAggregator object.
+   */
+  def merge(other: ExpectationAggregator): this.type = {
+    if (other.count != 0) {
+      totalCnt += other.totalCnt
+
+      val localThisNewWeights = this.newWeights
```
jkbradley: Why do you need to make these local copies?

mllib/src/main/scala/org/apache/spark/ml/clustering/GaussianMixture.scala
```diff
@@ -323,27 +326,95 @@ class GaussianMixture @Since("2.0.0") (
...
-    val model = copyValues(new GaussianMixtureModel(uid, parentModel.weights, gaussians))
-      .setParent(this)
+
+    val gaussianDists = gaussians.map { case (mean, covVec) =>
+      val cov = new DenseMatrix(numFeatures, numFeatures,
+        GaussianMixture.unpackUpperTriangularMatrix(numFeatures, covVec.values))
+      new MultivariateGaussian(mean, cov)
+    }
+
+    val model = copyValues(new GaussianMixtureModel(uid, weights, gaussianDists)).setParent(this)
```
jkbradley: Below here, we call logNumFeatures.  This isn't part of your PR, but could you move it earlier since numFeatures is available before running the algorithm?

mllib/src/main/scala/org/apache/spark/ml/clustering/GaussianMixture.scala
```diff
@@ -356,13 +427,243 @@ class GaussianMixture @Since("2.0.0") (
...
+  /**
+   * Convert an n * (n + 1) / 2 dimension array representing the upper triangular part of a matrix
+   * into an n * n array representing the full symmetric matrix.
+   *
+   * @param n The order of the n by n matrix.
+   * @param triangularValues The upper triangular part of the matrix packed in an array
+   *                         (column major).
+   * @return An array which represents the symmetric matrix in column major.
+   */
+  private[clustering] def unpackUpperTriangularMatrix(
```
jkbradley: You always use this right away by converting it to a DenseMatrix, so how about just returning a DenseMatrix?

mllib/src/main/scala/org/apache/spark/ml/clustering/GaussianMixture.scala
jkbradley: And in several other places

python/pyspark/ml/clustering.py
jkbradley: I like the table for documentation though.  Does using fewer digits stabilize it?

mllib/src/test/scala/org/apache/spark/ml/clustering/GaussianMixtureSuite.scala
```diff
@@ -126,6 +141,106 @@ class GaussianMixtureSuite extends SparkFunSuite with MLlibTestSparkContext
...
+    assert(gmm.gaussians(0).cov ~== cov(0) absTol 1E-3)
+    assert(gmm.gaussians(1).cov ~== cov(1) absTol 1E-3)
+  }
+
+  test("univariate sparse data with two clusters") {
+    val weights = Array(2.0 / 3.0, 1.0 / 3.0)
+    val mean = Array(Vectors.dense(5.1604), Vectors.dense(-4.3673))
+    val cov = Array(Matrices.dense(1, 1, Array(0.86644)), Matrices.dense(1, 1, Array(1.1098)))
+
+    val gmm = new GaussianMixture().setK(2).fit(sparseDataset)
```
jkbradley: Set the seed in this and other tests

mllib/src/test/scala/org/apache/spark/ml/clustering/GaussianMixtureSuite.scala
```diff
@@ -18,22 +18,37 @@
...
+  @transient var decompositionDataset: Dataset[_] = _
+  @transient var rDataset: Dataset[_] = _
 
   override def beforeAll(): Unit = {
     super.beforeAll()
 
     dataset = KMeansSuite.generateKMeansData(spark, 50, 3, k)
+    denseDataset = denseData.map(FeatureData).toDF()
+    sparseDataset = denseData.map { point =>
+      FeatureData(Vectors.sparse(1, Array(0), point.toArray))
```
jkbradley: Simplify: ```point.toSparse```

mllib/src/test/scala/org/apache/spark/ml/clustering/GaussianMixtureSuite.scala
```diff
@@ -126,6 +141,106 @@ class GaussianMixtureSuite extends SparkFunSuite with MLlibTestSparkContext
...
+
+    assert(gmm.weights(0) ~== weights(0) absTol 1E-3)
+    assert(gmm.weights(1) ~== weights(1) absTol 1E-3)
+    assert(gmm.gaussians(0).mean ~== mean(0) absTol 1E-3)
+    assert(gmm.gaussians(1).mean ~== mean(1) absTol 1E-3)
+    assert(gmm.gaussians(0).cov ~== cov(0) absTol 1E-3)
+    assert(gmm.gaussians(1).cov ~== cov(1) absTol 1E-3)
+  }
+
+  test("univariate sparse data with two clusters") {
```
jkbradley: This and the previous tests are almost the same.  How about combining them via a helper function?

jkbradley (commented):
Thanks for the nice PR!  I only found small things to comment on.

mllib/src/main/scala/org/apache/spark/ml/clustering/GaussianMixture.scala
yanboliang: This is because we are actually accessing a ```getter``` method when we call ```this.newWeights```. To improve the performance of the loop at L655, we should use explicit pointers to the values rather than call ```getter``` each time. This probably isn't a big deal in this case since ```k``` is usually not very big, but I don't think it's a bad idea.

python/pyspark/ml/clustering.py
yanboliang: Using fewer digits can not solve the unstable test issue, since the output of ```gaussiansDF.show()``` is pure string, we must match them exactly. Actually ```gaussiansDF.show()``` only shows part of the whole DataFrame, others are omitted, so the original test does not check for all cells of ```gaussiansDF```.

yanboliang:
@jkbradley I addressed most of your comments. Thanks.

mllib/src/main/scala/org/apache/spark/ml/clustering/GaussianMixture.scala
jkbradley: Oh, I see.  I'm not sure about this either.  Does the JIT compiler adjust enough to make it efficient?  The way it is seems fine though.

mllib/src/test/scala/org/apache/spark/ml/clustering/GaussianMixtureSuite.scala
```diff
@@ -126,9 +143,104 @@ class GaussianMixtureSuite extends SparkFunSuite with MLlibTestSparkContext
...
+    val gaussians = means.zip(covs).map { case (mean, cov) =>
+      new MultivariateGaussian(mean, cov)
+    }
+
+    val expected = new GaussianMixtureModel("dummy", weights, gaussians)
+    val actual = new GaussianMixture().setK(2).setSeed(seed).fit(denseDataset)
+    modelEquals(expected, actual)
+  }
+
+  test("univariate sparse data with two clusters") {
```
jkbradley: This and the previous tests are almost the same. How about combining them via a helper function?

python/pyspark/ml/clustering.py
jkbradley: Right, that makes sense.

mllib/src/test/scala/org/apache/spark/ml/clustering/GaussianMixtureSuite.scala
jkbradley: Actually, don't bother.  This looks ready.

jkbradley:
This LGTM

@sethah Any further comments before we merge it?

mllib/src/main/scala/org/apache/spark/ml/clustering/GaussianMixture.scala
sethah: Yeah, I don't think it's necessary here because this is inside the merge operations which will be called far less than the add operation. It may be overkill in any event, but I'm also ok leaving it.

mllib/src/test/scala/org/apache/spark/ml/clustering/GaussianMixtureSuite.scala
sethah: I'd be in favor of merging them since they are so nearly identical

sethah:
I did a quick pass and it looks pretty good. I'll take a more thorough look at the tests this weekend, but if you want to merge it I think any of those items could be addressed in follow ups.

jkbradley:
OK, I'll just wait so @sethah can make a final pass and so @yanboliang can merge the 2 tests.

mllib/src/test/scala/org/apache/spark/ml/clustering/GaussianMixtureSuite.scala
yanboliang: Yeah, I moved on to merge them together. Thanks.

mllib/src/test/scala/org/apache/spark/ml/clustering/GaussianMixtureSuite.scala
```diff
@@ -126,9 +143,93 @@ class GaussianMixtureSuite extends SparkFunSuite with MLlibTestSparkContext
...
+    }
+    val expected = new GaussianMixtureModel("dummy", weights, gaussians)
+
+    Seq(denseDataset, sparseDataset).foreach { dataset =>
+      val actual = new GaussianMixture().setK(2).setSeed(seed).fit(dataset)
+      modelEquals(expected, actual)
+    }
+  }
+
+  test("check distributed decomposition") {
```
sethah: This test only checks that when we distribute the computation that it produces a model, i.e. that it doesn't fail. So, AFAICT we don't have any test right now that checks that when we distribute the computation it produces a _correct_ model. I think it's a good idea to have that here.

mllib/src/test/scala/org/apache/spark/ml/clustering/GaussianMixtureSuite.scala
```diff
@@ -126,9 +143,93 @@ class GaussianMixtureSuite extends SparkFunSuite with MLlibTestSparkContext
     testEstimatorAndModelReadWrite(gm, dataset,
       GaussianMixtureSuite.allParamSettings, checkModelData)
   }
+
```
sethah: In most of the other test suites in ML we have a test that checks the prediction/transform methods. For example, checking that the prediction always matches the highest probability, checking that probabilities sum to one. I don't see much reason to diverge from that pattern here, what do you think @yanboliang?

mllib/src/test/scala/org/apache/spark/ml/clustering/GaussianMixtureSuite.scala
```diff
@@ -141,4 +242,37 @@ object GaussianMixtureSuite {
...
+    Vectors.dense(10.6198257, 9.9438713), Vectors.dense(9.8442045, 8.5292476),
+    Vectors.dense(9.5218499, 10.4179416)
+  )
+
+  case class FeatureData(features: Vector)
+
+  def modelEquals(m1: GaussianMixtureModel, m2: GaussianMixtureModel): Unit = {
+    assert(m1.weights.length === m2.weights.length)
+    for (i <- m1.weights.indices) {
+      assert(m1.gaussians(i).mean ~== m2.gaussians(i).mean absTol 1E-3)
```
sethah: Why not also check the weights here?

sethah (commented):
Just a couple comments on testing. Nice work on the performance improvement :)

mllib/src/test/scala/org/apache/spark/ml/clustering/GaussianMixtureSuite.scala
yanboliang: Sounds good, updated.

mllib/src/test/scala/org/apache/spark/ml/clustering/GaussianMixtureSuite.scala
yanboliang: This is because the model is big and it's tedious to construct the model in advance. In this model, ```gaussians```(the array of ```MultivariateGaussian```) contains 5 elements and each element contains a mean array of length 50 and a covariance matrix of size 50 * 50.

mllib/src/test/scala/org/apache/spark/ml/clustering/GaussianMixtureSuite.scala
yanboliang: We have checked ```weights``` equality at L272. The ```weights``` array is belong to the model rather than each cluster.

yanboliang: Oops, forgot it, added. Thanks.

mllib/src/test/scala/org/apache/spark/ml/clustering/GaussianMixtureSuite.scala
sethah: I played with this a bit, and wrote a test to generate two very separate clusters and run with distributed computation. Then we could check that the model learns approximately the correct cluster means. However, I found that the algorithm seems incapable of learning even this very contrived example - due to the initialization method.

I still think it's a good test to have, but if you feel strongly against it then let's leave it. Otherwise it could be a follow up (along with more investigation to the initialization method, which does not seem to be effective).

sethah:
Left one small comment which isn't a blocker. LGTM otherwise.

mllib/src/test/scala/org/apache/spark/ml/clustering/GaussianMixtureSuite.scala
yanboliang: Yeah, I also suffer from bad initialization in some of my use cases. So I think we should push to commit [SPARK-15785](https://issues.apache.org/jira/browse/SPARK-15785) firstly. It's more easy to add correctness test after we support initial model. I'll leave this as follow up and open [SPARK-19144](https://issues.apache.org/jira/browse/SPARK-19144) to track. Thanks.

yanboliang:
Merged into master. Thanks for all reviews.

PR #15414, opened 2016-10-10
[SPARK-17848][ML] Move LabelCol datatype cast into Predictor.fit

zhengruifeng:

1, move cast to `Predictor`
2, and then, remove unnecessary cast
## How was this patch tested?

existing tests


zhengruifeng:
test this please

zhengruifeng:
Jenkins, retest this please

mllib/src/main/scala/org/apache/spark/ml/Predictor.scala
```diff
@@ -121,10 +122,18 @@ abstract class Predictor[
...
+    dataset.select(col($(labelCol)), col($(featuresCol))).rdd.map {
       case Row(label: Double, features: Vector) => LabeledPoint(label, features)
     }
   }
+
+  /**
+   * Return the given DataFrame, with [[labelCol]] casted to DoubleType.
+   */
+    protected def castDataSet(dataset: Dataset[_]): DataFrame = {
+      val labelMeta = dataset.schema.fields.filter(_.name == $(labelCol)).head.metadata
```
hhbyyh: Maybe simplify it: dataset.schema("value").metadata

sethah:
What do you think about adding a new suite `PredictorSuite` where we can create a mock predictor, and call train on data of various types. The train method can just require that the label column is `DoubleType`:

````scala
class MockPredictor(override val uid: String)
  extends Predictor[Vector, MockPredictor, MockPredictionModel] {

  override def train(dataset: Dataset[_]): MockPredictionModel = {
    require(dataset.schema("label").dataType == DoubleType)
    new MockPredictionModel(uid)
  }

  override def copy(extra: ParamMap): MockPredictor = defaultCopy(extra)
}

class MockPredictionModel(override val uid: String)
  extends PredictionModel[Vector, MockPredictionModel] {

  override def predict(features: Vector): Double = 1.0

  override def copy(extra: ParamMap): MockPredictionModel = defaultCopy(extra)
}
````

Then we just have a test that calls `fit` for each type of data.

File: mllib/src/main/scala/org/apache/spark/ml/Predictor.scala (hunk 1 above)
Comment (zhengruifeng) [inline mllib/src/main/scala/org/apache/spark/ml/Predictor.scala]: Thanks

**Comment by zhengruifeng** [comment]
Ok, I will create this Suite.

File: mllib/src/test/scala/org/apache/spark/ml/util/MLTestingUtils.scala [hunk 2]
```diff
@@ -117,7 +117,7 @@ object MLTestingUtils extends SparkFunSuite {
       Seq(ShortType, LongType, IntegerType, FloatType, ByteType, DoubleType, DecimalType(10, 0))
     types.map { t =>
         val castDF = df.select(col(labelColName).cast(t), col(featuresColName))
-        t -> TreeTests.setMetadata(castDF, 2, labelColName, featuresColName)
+        t -> TreeTests.setMetadata(castDF, 0, labelColName, featuresColName)
```
sethah: What is this for? If the intent is to force `getNumClasses` to infer the number of classes, then you're no longer testing the not inferred case. Further, the point of this PR is to eliminate the need to do that since it is not a robust solution, IMO. 

Also, I'd like to remove the dependence on `TreeTests` here (and `genRegressionDF`) and just explicitly set the attributes in the functions.

mllib/src/test/scala/org/apache/spark/ml/PredictorSuite.scala
```diff
@@ -0,0 +1,57 @@
...
+    extends PredictionModel[Vector, MockPredictionModel] {
+
+    override def predict(features: Vector): Double = 1.0
+
+    override def copy(extra: ParamMap): MockPredictionModel = defaultCopy(extra)
+  }
+
+  test("should support all NumericType labels and not support other types") {
+    val predictor = new MockPredictor("mock")
+    MLTestingUtils.checkNumericTypes[MockPredictionModel, MockPredictor](
```
sethah: Why don't we just cycle through the types here and call `fit`. I think it's a bit confusing the way it is now.

mllib/src/test/scala/org/apache/spark/ml/PredictorSuite.scala
```diff
@@ -0,0 +1,57 @@
...
+      new MockPredictionModel(uid)
+    }
+
+    override def copy(extra: ParamMap): MockPredictor = defaultCopy(extra)
+  }
+
+  class MockPredictionModel(override val uid: String)
+    extends PredictionModel[Vector, MockPredictionModel] {
+
+    override def predict(features: Vector): Double = 1.0
```
sethah: `override def predict(features: Vector): Double = throw new NotImplementedError()` We can do this for everything except `train`.

mllib/src/test/scala/org/apache/spark/ml/util/MLTestingUtils.scala
zhengruifeng: Ok, I will revert this

mllib/src/test/scala/org/apache/spark/ml/PredictorSuite.scala
zhengruifeng: OK, I will update this.

zhengruifeng:
@sethah I have maken some modification according to the comments

sethah:
Thanks, I'll take a more detailed look in the next couple of days. Let's also wait and see if we can get @yanboliang or @jkbradley to give an opinion.

mllib/src/test/scala/org/apache/spark/ml/PredictorSuite.scala
```diff
@@ -0,0 +1,80 @@
...
+import org.apache.spark.SparkFunSuite
+import org.apache.spark.ml.linalg._
+import org.apache.spark.ml.param.ParamMap
+import org.apache.spark.ml.util._
+import org.apache.spark.mllib.util.MLlibTestSparkContext
+import org.apache.spark.sql.Dataset
+import org.apache.spark.sql.functions._
+import org.apache.spark.sql.types._
+
+class PredictorSuite extends SparkFunSuite with MLlibTestSparkContext with DefaultReadWriteTest {
```
sethah: don't need `DefaultReadWriteTest`

zhengruifeng:
@jkbradley @yanboliang Could you please have a review of this? This PR unify usage of labelCol casting and fixs a bug described in [https://issues.apache.org/jira/browse/SPARK-17797]

zhengruifeng:
@jkbradley @yanboliang Just re-pinging for your opinions.

jkbradley:
Can you please document in Predictor that it accepts all NumericType labels?  Other than that, this LGTM.  Thanks!

sethah:
LGTM as well after adding @jkbradley's suggestion.

zhengruifeng:
@jkbradley @sethah I add a comment, thanks for reviews.

jkbradley:
LGTM
Merging with master
Thanks!

PR #15415, opened 2016-10-10
[SPARK-14503][ML] spark.ml API for FPGrowth

hhbyyh:

jira: https://issues.apache.org/jira/browse/SPARK-14503 
Function parity: Add FPGrowth and AssociationRules to ML.

design doc: https://docs.google.com/document/d/1bVhABn5DiEj8bw0upqGMJT2L4nvO_0_cXdwu4uMT6uU/pub

Currently I make FPGrowthModel a transformer. For each association rule,  it will just examine the input items against antecedents and summarize the consequents.

Update:
Thinking again, FPGrowth is only the algorithm to find the frequent itemsets, and can be replaced by other algorithms. The frequent itemsets are used by AssociationRules to generate the association rules. Then we can use the association rules to predict with other records.

drawing1

**For reviewers**, Let's first decide if the current `transform` function meets your expectation.

Current options:

1. Current implementation: Use Estimator and Transformer pattern in ML, the `transform` function will examine the input items against all the association rules and summarize the consequents. Users can also access frequent items and association rules via other model members.

2. Keep the Estimator and Transformer pattern. But AssociationRulesModel and FPGrowthModel will have empty `transform` function, meaning DataFrame has no change after transform. But users can access frequent items and association rules via other model members.

3. (mentioned by @zhengruifeng) Keep the Estimator and Transformer pattern. But `FPGrowthModel` and `AssociationRulesModel` will just return frequent itemsets and association rules DataFrame in the `transform` function. Meaning the resulting DataFrame after `transform` will not be related to the input DataFrame.

4. Discard the Estimator and Transformer pattern. Both FPGrowth and FPGrowthModel will directly extend from PipelineStage, thus we don't need to have a `transform` function.

 I'd like to hear more concrete suggestions. I would prefer option 1 or 2.

update 2:

As discussed  in the jira, we will not expose AssociationRules as a public API for now. 

## How was this patch tested?

new unit test suites


hhbyyh:
@yanboliang @jkbradley This should be ready for review.

mllib/src/main/scala/org/apache/spark/ml/fpm/AssociationRules.scala
```diff
@@ -0,0 +1,113 @@
...
+
+  /** @group setParam */
+  @Since("2.1.0")
+  def setFreqCol(value: String): this.type = set(freqCol, value)
+
+  /**
+   * Param for minimum confidence, range [0.0, 1.0].
+   * @group param
+   */
+  final val minConfidence: DoubleParam = new DoubleParam(this, "minConfidence", "min confidence")
```
zhengruifeng: there should be a `ParamValidators.inRange(...)`

mllib/src/main/scala/org/apache/spark/ml/fpm/AssociationRules.scala
```diff
@@ -0,0 +1,113 @@
...
+   * @return a DataFrame("antecedent", "consequent", "confidence") containing the association
+    *         rules.
+   *
+   */
+  @Since("2.1.0")
+  def run(freqItemsets: Dataset[_]): DataFrame = {
+    val freqItemSetRdd = freqItemsets.select($(itemsCol), $(freqCol)).rdd
+      .map(row => new FreqItemset(row.getSeq[String](0).toArray, row.getLong(1)))
+
+    val sqlContext = SparkSession.builder().getOrCreate()
```
zhengruifeng: Since val `sqlContext` is of type `SparkSession`, what about rename it `spark`?

mllib/src/main/scala/org/apache/spark/ml/fpm/AssociationRules.scala
```diff
@@ -0,0 +1,113 @@
...
+
+/**
+ * :: Experimental ::
+ *
+ * Generates association rules from frequent itemsets ("items", "freq"). This method only generates
+ * association rules which have a single item as the consequent.
+ */
+@Since("2.1.0")
+@Experimental
+class AssociationRules(override val uid: String) extends Params {
```
zhengruifeng: Since `AssociationRules` transform DataFrame `freqItemsets` to DataFrame `rules`, can it be a subclass of `Transformer`?

mllib/src/main/scala/org/apache/spark/ml/fpm/AssociationRules.scala
```diff
@@ -0,0 +1,113 @@
...
+   * Computes the association rules with confidence above [[minConfidence]].
+   * @param freqItemsets DataFrame containing frequent itemset obtained from algorithms like
+   *                     [[FPGrowth]]. Users can set itemsCol (frequent itemSet, Array[String])
+   *                     and freqCol (appearance count, Long) names in the DataFrame.
+   * @return a DataFrame("antecedent", "consequent", "confidence") containing the association
+    *         rules.
+   *
+   */
+  @Since("2.1.0")
+  def run(freqItemsets: Dataset[_]): DataFrame = {
```
zhengruifeng: If inheriting `Transform`, here should be `override def transform(dataset: Dataset[_]): DataFrame`

mllib/src/main/scala/org/apache/spark/ml/fpm/AssociationRules.scala
```diff
@@ -0,0 +1,113 @@
...
+import org.apache.spark.mllib.fpm.FPGrowth.FreqItemset
+import org.apache.spark.sql.{DataFrame, Dataset, SparkSession}
+
+/**
+ * :: Experimental ::
+ *
+ * Generates association rules from frequent itemsets ("items", "freq"). This method only generates
+ * association rules which have a single item as the consequent.
+ */
+@Since("2.1.0")
```
zhengruifeng: should be 2.2.0

mllib/src/main/scala/org/apache/spark/ml/fpm/FPGrowth.scala
```diff
@@ -0,0 +1,232 @@
...
+  }
+
+  /**
+   * the minimal support level of the frequent pattern
+   * Default: 0.3
+   * @group param
+   */
+  @Since("2.2.0")
+  val minSupport: DoubleParam = new DoubleParam(this, "minSupport",
+    "the minimal support level of the frequent pattern (Default: 0.3)")
```
zhengruifeng: also need a `ParamValidator` here

mllib/src/main/scala/org/apache/spark/ml/fpm/FPGrowth.scala
```diff
@@ -0,0 +1,232 @@
...
+  def this() = this(Identifiable.randomUID("FPGrowth"))
+
+  /** @group setParam */
+  @Since("2.2.0")
+  def setMinSupport(value: Double): this.type = set(minSupport, value)
+  setDefault(minSupport -> 0.3)
+
+  /** @group setParam */
+  @Since("2.2.0")
+  def setFeaturesCol(value: String): this.type = set(featuresCol, value)
```
zhengruifeng: I perfer use `setInputCol` and `inputCol` instead of this.

mllib/src/main/scala/org/apache/spark/ml/fpm/FPGrowth.scala
```diff
@@ -0,0 +1,232 @@
...
+
+  /** @group setParam */
+  @Since("2.2.0")
+  def setPredictionCol(value: String): this.type = set(predictionCol, value)
```
zhengruifeng: ditto, `setOutputCol`

mllib/src/main/scala/org/apache/spark/ml/fpm/FPGrowth.scala
```diff
@@ -0,0 +1,232 @@
...
+  extends Model[FPGrowthModel] with FPGrowthParams with MLWritable {
+
+  /**
+   * minimal confidence for generating Association Rule
+   * Default: 0.8
+   * @group param
+   */
+  @Since("2.2.0")
+  val minConfidence: DoubleParam = new DoubleParam(this, "minConfidence",
+    "minimal confidence for generating Association Rule (Default: 0.8)")
```
zhengruifeng: `ParamValidator.inRange(0,1,...)`

mllib/src/main/scala/org/apache/spark/ml/fpm/FPGrowth.scala
```diff
@@ -0,0 +1,232 @@
...
+  /** @group getParam */
+  @Since("2.2.0")
+  def getMinConfidence: Double = $(minConfidence)
+
+  /** @group setParam */
+  @Since("2.2.0")
+  def setMinConfidence(value: Double): this.type = set(minConfidence, value)
+
+  @Since("2.2.0")
+  override def transform(dataset: Dataset[_]): DataFrame = {
```
zhengruifeng: Is there some discussion about the behavior of `transform` here? It seems a new feature.

mllib/src/main/scala/org/apache/spark/ml/fpm/AssociationRules.scala
hhbyyh: `freqItemsets` and `rules` does not have a one-to-one mapping relation and will probably violates the primitives of Transformer.

mllib/src/main/scala/org/apache/spark/ml/fpm/FPGrowth.scala
hhbyyh: Thanks. Let's collect more feedback about it.

mllib/src/main/scala/org/apache/spark/ml/fpm/FPGrowth.scala
hhbyyh: Indeed it's a new feature for making FPGrowthModel a Transformer. I've added it to the PR description to help draw more attention.

mllib/src/main/scala/org/apache/spark/ml/fpm/FPGrowth.scala
```diff
@@ -0,0 +1,251 @@
...
+    val associationRules = getAssociationRules.rdd.map(r =>
+      (r.getSeq[String](0), r.getSeq[String](1))
+    ).collect()
```
aray: Should we guard against this being too large?

mllib/src/main/scala/org/apache/spark/ml/fpm/FPGrowth.scala
```diff
@@ -0,0 +1,251 @@
...
+
+    // For each rule, examine the input items and summarize the consequents
+    val predictUDF = udf((items: Seq[String]) => associationRules.flatMap( r =>
+      if (r._1.forall(items.contains(_))) r._2 else Array.empty[String]
+    ).distinct)
+    dataset.withColumn($(predictionCol), predictUDF(col($(featuresCol))))
```
aray: Why are we including `featuresCol` as a new column here?

mllib/src/main/scala/org/apache/spark/ml/fpm/FPGrowth.scala
```diff
@@ -0,0 +1,251 @@
...
+  @Since("2.2.0")
+  def getMinConfidence: Double = $(minConfidence)
+
+  /** @group setParam */
+  @Since("2.2.0")
+  def setMinConfidence(value: Double): this.type = set(minConfidence, value)
+
+  @Since("2.2.0")
+  override def transform(dataset: Dataset[_]): DataFrame = {
+    val associationRules = getAssociationRules.rdd.map(r =>
```
aray: Are we worried about recomputing association rules every time we do a transform?

aray (commented):
I'm not really sold on the need for the transform method, but otherwise looks good.

mllib/src/main/scala/org/apache/spark/ml/fpm/FPGrowth.scala
hhbyyh: Good point. Thanks @aray. I've tried to leverage lazy val to avoid unnecessary computation.

hhbyyh:
@aray Thanks for the comments. I'd like to collect more feedback on the current `transform` behavior. I saw your comment regarding to the necessity. Given the current ML class hierarchy (Estimator and transformer), I'd like to hear more suggestions on the expectation of `transform` in `FPGrowthModel`, or do you think we should not extend from Transformer. Thanks.

mllib/src/main/scala/org/apache/spark/ml/fpm/AssociationRules.scala
```diff
@@ -0,0 +1,234 @@
...
+  setDefault(minConfidence -> 0.8)
+
+  /**
+   * Computes the association rules with confidence above [[minConfidence]].
+   *
+   * @return a DataFrame("antecedent", "consequent", "confidence") containing the association
+   *         rules.
+   */
+  @Since("2.2.0")
+  override def fit(dataset: Dataset[_]): AssociationRulesModel = {
```
jkbradley: Definitely document the input schema since it is non-standard.  The description here can be short, with more details in the class-level Scala doc.

mllib/src/main/scala/org/apache/spark/ml/fpm/AssociationRules.scala
```diff
@@ -0,0 +1,234 @@
...
+  override def copy(extra: ParamMap): AssociationRules = defaultCopy(extra)
+
+}
+
+
+/**
+ * :: Experimental ::
+ * Model fitted by AssociationRules.
+ *
+ * @param associationRules AssociationRules
```
jkbradley: clarify

mllib/src/main/scala/org/apache/spark/ml/fpm/AssociationRules.scala
```diff
@@ -0,0 +1,234 @@
...
+ */
+@Since("2.2.0")
+@Experimental
+class AssociationRulesModel private[ml] (
+    @Since("2.2.0") override val uid: String,
+    val associationRules: DataFrame)
```
jkbradley: need Since annotation

mllib/src/main/scala/org/apache/spark/ml/fpm/AssociationRules.scala
```diff
@@ -0,0 +1,234 @@
...
+
+  /** @group setParam */
+  @Since("2.2.0")
+  def setPredictionCol(value: String): this.type = set(predictionCol, value)
+
+  @Since("2.2.0")
+  override def transform(dataset: Dataset[_]): DataFrame = {
+    val rules = associationRules.rdd.map(r =>
+      (r.getSeq[String](0), r.getSeq[String](1))
+    ).collect()
```
jkbradley: This looks like it will blow up clusters for large models.  Use a join instead of a UDF.

mllib/src/main/scala/org/apache/spark/ml/fpm/FPGrowth.scala
```diff
@@ -0,0 +1,260 @@
...
+    ParamValidators.inRange(0.0, 1.0))
+  setDefault(minSupport -> 0.3)
+
+  /** @group getParam */
+  @Since("2.2.0")
+  def getMinSupport: Double = $(minSupport)
+
+  /**
+   * Number of partitions used by parallel FP-growth
+   * @group param
```
jkbradley: Make this an expertParam

mllib/src/main/scala/org/apache/spark/ml/fpm/FPGrowth.scala
```diff
@@ -0,0 +1,260 @@
...
+ *  Recommendation]]
+ */
+@Since("2.2.0")
+@Experimental
+class FPGrowth @Since("2.2.0") (
+    @Since("2.2.0") override val uid: String)
+  extends Estimator[FPGrowthModel] with FPGrowthParams with DefaultParamsWritable {
+
+  @Since("2.2.0")
+  def this() = this(Identifiable.randomUID("FPGrowth"))
```
jkbradley: Use lowercase name to match other algs: "fpgrowth"

mllib/src/main/scala/org/apache/spark/ml/fpm/FPGrowth.scala
```diff
@@ -0,0 +1,260 @@
...
+
+  /** @group setParam */
+  @Since("2.2.0")
+  def setFeaturesCol(value: String): this.type = set(featuresCol, value)
+
+  /** @group setParam */
+  @Since("2.2.0")
+  def setPredictionCol(value: String): this.type = set(predictionCol, value)
+
+  def fit(dataset: Dataset[_]): FPGrowthModel = {
```
jkbradley: override

mllib/src/main/scala/org/apache/spark/ml/fpm/FPGrowth.scala
```diff
@@ -0,0 +1,260 @@
...
+    val data = dataset.select($(featuresCol)).rdd.map(r => r.getSeq[String](0).toArray)
+    val parentModel = new MLlibFPGrowth().setMinSupport($(minSupport)).run(data)
+    copyValues(new FPGrowthModel(uid, parentModel))
```
jkbradley: setParent

mllib/src/main/scala/org/apache/spark/ml/fpm/FPGrowth.scala
```diff
@@ -0,0 +1,260 @@
...
+    private val parentModel: MLlibFPGrowthModel[_])
+  extends Model[FPGrowthModel] with FPGrowthParams with MLWritable {
+
+  /**
+   * minimal confidence for generating Association Rule
+   * Default: 0.8
+   * @group param
+   */
+  @Since("2.2.0")
+  val minConfidence: DoubleParam = new DoubleParam(this, "minConfidence",
```
jkbradley: This must be added to the Estimator as well.  All Model params should be in the Estimator.

TODO(joseph): Should we combine these this way?

mllib/src/main/scala/org/apache/spark/ml/fpm/FPGrowth.scala
```diff
@@ -0,0 +1,260 @@
...
+  @Since("2.2.0")
+  def getAssociationRules: DataFrame = {
+     associationRulesModel.associationRules
+  }
+
+  /**
+   * Get frequent items fitted by FPGrowth, in the format of DataFrame("items", "freq")
+   */
+  @Since("2.2.0")
+  @transient lazy val getFreqItems: DataFrame = {
```
jkbradley: getFreqItems -> getFreqItemsets

mllib/src/main/scala/org/apache/spark/ml/fpm/FPGrowth.scala
```diff
@@ -0,0 +1,260 @@
...
+
+  @Since("2.2.0")
+  override def transformSchema(schema: StructType): StructType = {
+    validateAndTransformSchema(schema)
+  }
+
+  @Since("2.2.0")
+  override def copy(extra: ParamMap): FPGrowthModel = {
+    val copied = new FPGrowthModel(uid, parentModel)
+    copyValues(copied, extra)
```
jkbradley: setParent

mllib/src/test/scala/org/apache/spark/ml/fpm/FPGrowthSuite.scala
```diff
@@ -0,0 +1,110 @@
...
+    )).toDF("id", "features")
+  }
+
+  /**
+   * Mapping from all Params to valid settings which differ from the defaults.
+   * This is useful for tests which need to exercise all Params, such as save/load.
+   * This excludes input columns to simplify some tests.
+   */
+  val allParamSettings: Map[String, Any] = Map(
+    "minSupport" -> 0.3
```
jkbradley: This should include numPartitions and predictionCol

mllib/src/main/scala/org/apache/spark/ml/fpm/AssociationRules.scala
```diff
@@ -0,0 +1,234 @@
...
+
+  /** @group setParam */
+  @Since("2.2.0")
+  def setMinConfidence(value: Double): this.type = set(minConfidence, value)
+  setDefault(minConfidence -> 0.8)
+
+  /**
+   * Computes the association rules with confidence above [[minConfidence]].
+   *
+   * @return a DataFrame("antecedent", "consequent", "confidence") containing the association
```
jkbradley: returns a model, not DataFrame

jkbradley (commented):
Thanks a lot for the PR!  I did a quick pass and will leave some comments.  But I also have higher level comments which should be discussed on JIRA.  I'll post them there now.

mllib/src/main/scala/org/apache/spark/ml/fpm/FPGrowth.scala
```diff
@@ -0,0 +1,327 @@
...
+
+import org.apache.hadoop.fs.Path
+
+import org.apache.spark.annotation.{Experimental, Since}
+import org.apache.spark.ml.{Estimator, Model}
+import org.apache.spark.ml.param._
+import org.apache.spark.ml.param.shared.{HasFeaturesCol, HasPredictionCol}
+import org.apache.spark.ml.util._
+import org.apache.spark.mllib.fpm.{AssociationRules => MLlibAssociationRules,
+FPGrowth => MLlibFPGrowth}
```
jkbradley: style: indent with 2 spaces

mllib/src/main/scala/org/apache/spark/ml/fpm/FPGrowth.scala
```diff
@@ -0,0 +1,327 @@
...
+ * Common params for FPGrowth and FPGrowthModel
+ */
+private[fpm] trait FPGrowthParams extends Params with HasFeaturesCol with HasPredictionCol {
+
+  /**
+   * Validates and transforms the input schema.
+   * @param schema input schema
+   * @return output schema
+   */
+  protected def validateAndTransformSchema(schema: StructType): StructType = {
```
jkbradley: Since annotation

mllib/src/main/scala/org/apache/spark/ml/fpm/FPGrowth.scala
```diff
@@ -0,0 +1,327 @@
...
+    "the minimal support level of the frequent pattern (Default: 0.3)",
+    ParamValidators.inRange(0.0, 1.0))
+  setDefault(minSupport -> 0.3)
+
+  /** @group getParam */
+  @Since("2.2.0")
+  def getMinSupport: Double = $(minSupport)
+
+  /**
+   * Number of partitions used by parallel FP-growth
```
jkbradley: Document what it means when this is not set and that this must be >= 1.
Also say this is not set by default.

mllib/src/main/scala/org/apache/spark/ml/fpm/FPGrowth.scala
```diff
@@ -0,0 +1,327 @@
...
+  /** @group setParam */
+  @Since("2.2.0")
+  def setMinSupport(value: Double): this.type = set(minSupport, value)
+
+  /** @group expertSetParam */
+  @Since("2.2.0")
+  def setNumPartitions(value: Int): this.type = set(numPartitions, value)
+
+  /** @group setParam
+   *  minConfidence has not effect during fitting.
```
jkbradley: "has not" -> "has no"
Also, put this comment in the Scaladoc for "val minConfidence"

mllib/src/main/scala/org/apache/spark/ml/fpm/FPGrowth.scala
```diff
@@ -0,0 +1,327 @@
...
+
+  /** @group setParam */
+  @Since("2.2.0")
+  def setFeaturesCol(value: String): this.type = set(featuresCol, value)
+
+  /** @group setParam */
+  @Since("2.2.0")
+  def setPredictionCol(value: String): this.type = set(predictionCol, value)
+
+  override def fit(dataset: Dataset[_]): FPGrowthModel = {
```
jkbradley: Add Since annotation

mllib/src/main/scala/org/apache/spark/ml/fpm/FPGrowth.scala
```diff
@@ -0,0 +1,327 @@
...
+    genericFit(dataset)
```
jkbradley: Call transformSchema first to do schema validation.

mllib/src/main/scala/org/apache/spark/ml/fpm/FPGrowth.scala
```diff
@@ -0,0 +1,327 @@
...
+    val frequentItems = dataset.sparkSession.createDataFrame(rows, schema).toDF("items", "freq")
+    copyValues(new FPGrowthModel(uid, frequentItems)).setParent(this)
+  }
+
+  @Since("2.2.0")
+  override def transformSchema(schema: StructType): StructType = {
+    validateAndTransformSchema(schema)
+  }
+
+  override def copy(extra: ParamMap): FPGrowth = defaultCopy(extra)
```
jkbradley: Since annotation

mllib/src/main/scala/org/apache/spark/ml/fpm/FPGrowth.scala
```diff
@@ -0,0 +1,327 @@
...
+ * :: Experimental ::
+ * Model fitted by FPGrowth.
+ *
+ * @param freqItemsets frequent items in the format of DataFrame("items", "freq")
+ */
+@Since("2.2.0")
+@Experimental
+class FPGrowthModel private[ml] (
+    @Since("2.2.0") override val uid: String,
+    val freqItemsets: DataFrame)
```
jkbradley: Since annotation

mllib/src/main/scala/org/apache/spark/ml/fpm/FPGrowth.scala
```diff
@@ -0,0 +1,327 @@
...
+  override def copy(extra: ParamMap): FPGrowthModel = {
+    val copied = new FPGrowthModel(uid, freqItemsets)
+    copyValues(copied, extra).setParent(this.parent)
+  }
+
+  @Since("2.2.0")
+  override def write: MLWriter = new FPGrowthModel.FPGrowthModelWriter(this)
+}
+
+object FPGrowthModel extends MLReadable[FPGrowthModel] {
```
jkbradley: Add Since annotation to object

mllib/src/main/scala/org/apache/spark/ml/fpm/FPGrowth.scala
```diff
@@ -0,0 +1,327 @@
...
+      val dataPath = new Path(path, "data").toString
+      val frequentItems = sparkSession.read.load(dataPath)
+      val model = new FPGrowthModel(metadata.uid, frequentItems)
+      DefaultParamsReader.getAndSetParams(model, metadata)
+      model
+    }
+  }
+}
+
+private[fpm] object AssocaitionRules {
```
jkbradley: typo: AssocaitionRules -> AssociationRules

mllib/src/main/scala/org/apache/spark/ml/fpm/FPGrowth.scala
```diff
@@ -0,0 +1,327 @@
...
+   * Computes the association rules with confidence above minConfidence.
+   * @param dataset DataFrame("items", "freq") containing frequent itemset obtained from
+   *                algorithms like [[FPGrowth]].
+   * @param itemsCol column name for frequent itemsets
+   * @param freqCol column name for frequent itemsets count
+   * @param minConfidence minimum confidence for the result association rules
+   * @return a DataFrame("antecedent", "consequent", "confidence") containing the association
+   *         rules.
+   */
+  @Since("2.2.0")
```
jkbradley: can remove Since annotation since the object is private

mllib/src/test/scala/org/apache/spark/ml/fpm/FPGrowthSuite.scala
```diff
@@ -0,0 +1,120 @@
...
+  /**
+   * Mapping from all Params to valid settings which differ from the defaults.
+   * This is useful for tests which need to exercise all Params, such as save/load.
+   * This excludes input columns to simplify some tests.
+   */
+  val allParamSettings: Map[String, Any] = Map(
+    "minSupport" -> 0.321,
+    "minConfidence" -> 0.456,
+    "numPartitions" -> 5,
+    "featuresCol" -> "features",
```
jkbradley: Exclude input columns here; it complicates the standard tests if we don't use default values for inputs.

jkbradley (commented):
I only had time for a partial review pass today, but here are some comments.  I like the combined API; thanks!

One general comment: I'm worried about the algorithm behavior changing in the future when we port the implementation to DataFrames.  E.g., changed hashing behavior could create different partitioning of the data.  I haven't thought about this too much though...

hhbyyh:
Thanks @jkbradley . I'm also working on improving the `transform` performance and add more unit tests. I'll address the comments in a combined update.

mllib/src/main/scala/org/apache/spark/ml/fpm/FPGrowth.scala
hhbyyh: how about make it private since we already have a getter for it.

hhbyyh:
@jkbradley Sent an update to refine the transform code and address the comments.

Regarding to the behavior changing concern, I think different partition strategy will only affect the overall efficiency and maybe the order of the frequent itemsets and association rules. As long as the result set does not change, I don't think it will disturb the users. Let me know if I miss anything.

mllib/src/test/scala/org/apache/spark/ml/fpm/FPGrowthSuite.scala
```diff
@@ -0,0 +1,130 @@
...
+  @transient var dataset: Dataset[_] = _
+
+  override def beforeAll(): Unit = {
+    super.beforeAll()
+    dataset = FPGrowthSuite.getFPGrowthData(spark)
+  }
+
+  test("FPGrowth fit and transform with different data types") {
+    Array(IntegerType, StringType, ShortType, LongType, ByteType).foreach { dt =>
+      val intData = dataset.withColumn("features", col("features").cast(ArrayType(dt)))
```
jkbradley: rename intData -> data

mllib/src/test/scala/org/apache/spark/ml/fpm/FPGrowthSuite.scala
```diff
@@ -0,0 +1,130 @@
...
+      (Array("1", "2"), 3L),
+      (Array("2", "1"), 3L)
+    )).toDF("items", "freqExp")
+    val freqItems = model.getFreqItemsets
+
+    val checkDF = freqItems.join(expectedFreq, "items")
+    assert(checkDF.count() == 3 && checkDF.filter(col("freq") === col("freqExp")).count() == 3)
+  }
+
+  test("FPGrowth getFreqItems with Null") {
```
jkbradley: In FPGrowth, document that null values are treated as empty sequences.

jkbradley (commented):
Thanks for the updates!

> Regarding to the behavior changing concern

I think you're right.  I'd never read through the algorithm carefully, but now that I do, I think you're right.

> refine the transform code

I'd prefer we hold off on this.  If we're changing the transform code (rather than wrapping the old code), then we'll need to spend more time writing new tests and QAing it.  We don't really have time to do that before the 2.2 release.  Also, when we get around to optimizing it, I'd prefer to do everything using DataFrame operations if possible.  Could you please save your code but revert the transform() code to wrap the old AssociationRules code?  Thanks!

hhbyyh:
Hi @jkbradley 
We can hold the transform code. 

> wrap the old AssociationRules code

Do you mean to make transform return the Association Rules DataFrame, like the currect `getAssociationRules` ?

jkbradley:
> wrap the old AssociationRules code

Sorry, forget this comment from me; I was thinking that something like FPGrowthModel.transform had already been implemented, but it's new.

Btw, I'm adding more comments now, so please hold off on changes for an hour.  Thanks!

mllib/src/main/scala/org/apache/spark/ml/fpm/FPGrowth.scala
```diff
@@ -0,0 +1,341 @@
...
+  /**
+   * Get frequent items fitted by FPGrowth, in the format of DataFrame("items", "freq")
+   */
+  @Since("2.2.0")
+  @transient val getFreqItemsets: DataFrame = freqItemsets
+
+  /**
+   * The transform method first generates the association rules according to the frequent itemsets.
+   * Then for each association rule, it will examine the input items against antecedents and
+   * summarize the consequents as prediction.
```
jkbradley: This needs more explanation so the user knows precisely what is being done.
1. Warning that this uses a Cartesian join and may explode for large datasets.
2. Explanation of how predictions are made.  There are several options here, and I'm actually not sure myself about how rules are most commonly applied.  Things I'm wondering: Do users need confidences attached to predictions?  How should we merge multiple predictions for a single itemset (row)?  Should we use an approximate algorithm to avoid the Cartesian join?

I'll think more about this.

mllib/src/main/scala/org/apache/spark/ml/fpm/FPGrowth.scala
```diff
@@ -0,0 +1,341 @@
...
+    copyValues(copied, extra).setParent(this.parent)
+  }
+
+  @Since("2.2.0")
+  override def write: MLWriter = new FPGrowthModel.FPGrowthModelWriter(this)
+}
+
+@Since("2.2.0")
+object FPGrowthModel extends MLReadable[FPGrowthModel] {
+  @Since("2.2.0")
```
jkbradley: newline

mllib/src/main/scala/org/apache/spark/ml/fpm/FPGrowth.scala
```diff
@@ -0,0 +1,341 @@
...
+
+  @Since("2.2.0")
+  override def load(path: String): FPGrowthModel = super.load(path)
+
+  /** [[MLWriter]] instance for [[FPGrowthModel]] */
+  private[FPGrowthModel]
+  class FPGrowthModelWriter(instance: FPGrowthModel) extends MLWriter {
+
+    override protected def saveImpl(path: String): Unit = {
+      // Save metadata and Params
```
jkbradley: remove this doc line

mllib/src/main/scala/org/apache/spark/ml/fpm/FPGrowth.scala
```diff
@@ -0,0 +1,341 @@
...
+      DefaultParamsWriter.saveMetadata(instance, path, sc)
+      val dataPath = new Path(path, "data").toString
+      instance.freqItemsets.write.save(dataPath)
```
jkbradley: specify parquet explicitly

mllib/src/main/scala/org/apache/spark/ml/fpm/FPGrowth.scala
```diff
@@ -0,0 +1,341 @@
...
+
+  private class FPGrowthModelReader extends MLReader[FPGrowthModel] {
+
+    /** Checked against metadata when loading model */
+    private val className = classOf[FPGrowthModel].getName
+
+    override def load(path: String): FPGrowthModel = {
+      val metadata = DefaultParamsReader.loadMetadata(path, sc, className)
+      val dataPath = new Path(path, "data").toString
+      val frequentItems = sparkSession.read.load(dataPath)
```
jkbradley: specify parquet

mllib/src/main/scala/org/apache/spark/ml/fpm/FPGrowth.scala
```diff
@@ -0,0 +1,341 @@
...
+   * @param dataset DataFrame("items", "freq") containing frequent itemset obtained from
+   *                algorithms like [[FPGrowth]].
+   * @param itemsCol column name for frequent itemsets
+   * @param freqCol column name for frequent itemsets count
+   * @param minConfidence minimum confidence for the result association rules
+   * @return a DataFrame("antecedent", "consequent", "confidence") containing the association
+   *         rules.
+   */
+  def getAssociationRulesFromFP[T: ClassTag](dataset: Dataset[_],
+        itemsCol: String = "items",
```
jkbradley: No need to specify default arg values.  If we make this public, we will need to avoid default args to make it Java friendly.

mllib/src/main/scala/org/apache/spark/ml/fpm/FPGrowth.scala
```diff
@@ -0,0 +1,341 @@
...
+
+  @Since("2.2.0")
+  override def load(path: String): FPGrowth = super.load(path)
+}
+
+/**
+ * :: Experimental ::
+ * Model fitted by FPGrowth.
+ *
+ * @param freqItemsets frequent items in the format of DataFrame("items", "freq")
```
jkbradley: Specify schema (Seq, Long)

mllib/src/main/scala/org/apache/spark/ml/fpm/FPGrowth.scala
```diff
@@ -0,0 +1,341 @@
...
+  @Since("2.2.0")
+  def setPredictionCol(value: String): this.type = set(predictionCol, value)
+
+  /**
+   * Get association rules fitted by AssociationRules using the minConfidence. Returns a dataframe
+   * with three fields, "antecedent", "consequent" and "confidence", where "antecedent" and
+   * "consequent" are Array[String] and "confidence" is Double.
+   */
+  @Since("2.2.0")
+  @transient lazy val getAssociationRules: DataFrame = {
```
jkbradley: Also, these may not be String values

mllib/src/main/scala/org/apache/spark/ml/fpm/FPGrowth.scala
jkbradley: Also state new column's output schema

mllib/src/main/scala/org/apache/spark/ml/fpm/FPGrowth.scala
```diff
@@ -0,0 +1,341 @@
...
+  @Since("2.2.0")
+  override def fit(dataset: Dataset[_]): FPGrowthModel = {
+    transformSchema(dataset.schema, logging = true)
+    genericFit(dataset)
+  }
+
+  private def genericFit[T: ClassTag](dataset: Dataset[_]): FPGrowthModel = {
+    val data = dataset.select($(featuresCol))
+    val items = data.where(col($(featuresCol)).isNotNull).rdd.map(r => r.getSeq[T](0).toArray)
+    val parentModel = new MLlibFPGrowth().setMinSupport($(minSupport)).run(items)
```
jkbradley: Specify numPartitions

jkbradley (commented):
OK, done with another pass.  I'll think more about genericTransform too.  Thanks!

mllib/src/main/scala/org/apache/spark/ml/fpm/FPGrowth.scala
hhbyyh: Originally I thought we can collect all rules and then use broadcast to avoid Cartesian join. But this applies only to smaller set of association rules.

mllib/src/test/scala/org/apache/spark/ml/fpm/FPGrowthSuite.scala
```diff
@@ -0,0 +1,130 @@
...
+    }
+  }
+
+  test("FPGrowth getFreqItems") {
+    val model = new FPGrowth().setMinSupport(0.7).fit(dataset)
+    val expectedFreq = spark.createDataFrame(Seq(
+      (Array("1"), 4L),
+      (Array("2"), 3L),
+      (Array("1", "2"), 3L),
+      (Array("2", "1"), 3L)
```
hhbyyh: As the sequence of the frequent items is not guaranteed, a little hack..

mllib/src/main/scala/org/apache/spark/ml/fpm/FPGrowth.scala
```diff
@@ -0,0 +1,346 @@
...
+          if (antecedent.forall(itemSet.contains)) {
+            consequent.filterNot(itemSet.contains)
+          } else {
+            Seq.empty
+          }
+        } else {
+          Seq.empty
+        }
+        (id, consequents)
+    }.aggregateByKey(new ArrayBuffer[T])((ar, seq) => ar ++= seq, (ar, seq) => ar ++= seq)
```
jkbradley: How about using an OpenHashSet here to avoid collecting duplicates during aggregation?

mllib/src/main/scala/org/apache/spark/ml/fpm/FPGrowth.scala
```diff
@@ -0,0 +1,346 @@
...
+     .map { case (index, cons) => (index, cons.distinct) }
+
+    val rowAndConsequents = dataset.toDF().rdd.zipWithIndex().map(_.swap)
+      .join(indexToConsequents).sortByKey(ascending = true, dataset.rdd.getNumPartitions)
```
jkbradley: No need to sortByKey

mllib/src/main/scala/org/apache/spark/ml/fpm/FPGrowth.scala
```diff
@@ -0,0 +1,346 @@
...
+
+/**
+ * :: Experimental ::
+ * A parallel FP-growth algorithm to mine frequent itemsets. The algorithm is described in
+ * <a href="http://dx.doi.org/10.1145/1454008.1454027">Li et al., PFP: Parallel FP-Growth for Query
+ * Recommendation</a>. PFP distributes computation in such a way that each worker executes an
+ * independent group of mining tasks. The FP-Growth algorithm is described in
+ * <a href="http://dx.doi.org/10.1145/335191.335372">Han et al., Mining frequent patterns without
+ * candidate generation</a>.
+ *
```
jkbradley: Here or elsewhere, comment that null featuresCol values are ignored during fit() and are treated as empty sets during transform().

jkbradley (commented):
Thanks for the updates!  I think this is almost done.  We should try to optimize transform in the future, but I think it's OK for an MVP.  It'd be interesting to write it using DataFrames instead of RDDs and benchmark (in the future).

mllib/src/main/scala/org/apache/spark/ml/fpm/FPGrowth.scala
hhbyyh: I tried OpenHashSet and it's about %15 slower than ArrayBuffer. 

.aggregateByKey(new OpenHashSet[T])((set, seq) => {
      seq.foreach(t => set.add(t))
      set
    } , (set1, set2) => set1.union(set2))

mllib/src/main/scala/org/apache/spark/ml/fpm/FPGrowth.scala
hhbyyh: Shall we try to keep the original order of the input dataset? The time cost is about 5% of the total transform time.

hhbyyh:
I tried a few different ways to implement the transform. https://gist.github.com/hhbyyh/889b88ae2176d1263fdc9dd3e29d1c2d.

The performance actually are similiar, while the current one can maintain the original order of the input dataset. I would be glad to see an more optimized version.

mllib/src/main/scala/org/apache/spark/ml/fpm/FPGrowth.scala
jkbradley: Huh, that's surprising to me.  Maybe it depends on how many duplicates are introduced.  Let's leave it as is then.

mllib/src/main/scala/org/apache/spark/ml/fpm/FPGrowth.scala
jkbradley: DataFrames don't really have a row ordering, so we don't need to maintain one.

mllib/src/main/scala/org/apache/spark/ml/fpm/FPGrowth.scala
hhbyyh: random number... The duplicate number should be small.

mllib/src/main/scala/org/apache/spark/ml/fpm/FPGrowth.scala
hhbyyh: If that's the case, does zipWithIndex() guarantees the same result for multiple calls? 
And if can ignore the ordering, perhaps I have a faster way to write the transform.

jkbradley: I think zipWithIndex guarantees the order if the data have not been shuffled between load/creation and the zipWithIndex.  AFAIK, shuffling can reorder rows in a partition arbitrarily.

We can definitely ignore the ordering for DataFrames.

hhbyyh: Checked again and the current implementation is as quick. I will just remove the sort.

hhbyyh:
Hi @jkbradley After further performance comparison, I found using broadcast would give much better performance for the transform.

I tested with some public data from http://fimi.ua.ac.be/data/.
For kosarak (.gz) data (300K records), the current transform would take more than 3 hours for only 2 rules, while the broadcast version only cost 0.15 sec with 900 rules. ( I adjusted support and confidence)

```
    val rules = associationRules.rdd.map(r =>
      (r.getSeq[Int](0), r.getSeq[Int](1))
    ).collect()
    val brRules = dataset.sparkSession.sparkContext.broadcast(rules)

    // For each rule, examine the input items and summarize the consequents
    val predictUDF = udf((items: Seq[Int]) => brRules.value.flatMap( r =>
      if (r._1.forall(items.contains(_))) r._2 else Seq.empty[Int]
    ).distinct)
    dataset.withColumn($(predictionCol), predictUDF(col($(featuresCol))))
```
The test can be verified by the code:
https://gist.github.com/hhbyyh/06fcf3fdc8f6edda971847bcb5783d99
https://gist.github.com/hhbyyh/889b88ae2176d1263fdc9dd3e29d1c2d

Thinking again, the broadcast implementation may have much better performance in any case. The major issue is how to support generic with the UDF.

jkbradley:
I agree that, if the set of rules is small (1-2 GB max), then collecting and broadcasting it is best.  But for larger sets of rules, we'd have to keep it distributed.

I'm very surprised by the time difference in your comparison.  I'll experiment a little myself and get back soon.

jkbradley (commented):
Thinking about it more, I'm OK with keeping the broadcast.  Later on, we could write this using DataFrames and give a broadcast hint.  Can you please make the warning for transform() more distinct (maybe with a "WARNING" label) and specific about broadcasting if we're going with this option?

I got this to work:
```
  private def genericTransform4(dataset: Dataset[_]): DataFrame = {
    val rules: Array[(Seq[Any], Seq[Any])] = associationRules.rdd.map(r =>
      (r.getSeq(0), r.getSeq(1))
    ).collect().asInstanceOf[Array[(Seq[Any], Seq[Any])]]
    val brRules = dataset.sparkSession.sparkContext.broadcast(rules)

    val dt = dataset.schema($(featuresCol)).dataType

    // For each rule, examine the input items and summarize the consequents
    val predictUDF = udf((items: Seq[_]) => brRules.value.flatMap( rule =>
      if (items != null && rule._1.forall(item => items.contains(item))) {
        rule._2.filter(item => !items.contains(item))
      } else {
        Seq.empty
      }
    ).distinct, dt)
    dataset.withColumn($(predictionCol), predictUDF(col($(featuresCol))))
  }
```

Thanks!

hhbyyh:
Thanks @jkbradley for contributing the code. That helps a lot.

hhbyyh:
> Btw, I could imagine us wanting to change this later. If we're recommending items a user could add to their basket, then we might want to suggest the most frequent item rather than nothing.

Do we need to support this now? That may not be expected for all the scenarios. And it seems an Imputer can help with the issue.

jkbradley:
I don't think we need to support the default prediction (for empty/null inputs) now.  I agree we could use an inputer or add something as an option later on.

Will take a final look now

mllib/src/main/scala/org/apache/spark/ml/fpm/FPGrowth.scala
```diff
@@ -0,0 +1,339 @@
...
+      .collect().asInstanceOf[Array[(Seq[Any], Seq[Any])]]
+    val brRules = dataset.sparkSession.sparkContext.broadcast(rules)
+
+    val dt = dataset.schema($(featuresCol)).dataType
+    // For each rule, examine the input items and summarize the consequents
+    val predictUDF = udf((items: Seq[_]) => {
+      if (items != null) {
+        val itemset = items.toSet
+        brRules.value.flatMap(rule =>
+          if (items != null && rule._1.forall(item => itemset.contains(item))) {
```
jkbradley: No need to check items != null here since you put that above.

mllib/src/main/scala/org/apache/spark/ml/fpm/FPGrowth.scala
```diff
@@ -0,0 +1,339 @@
...
+      .rdd.map(r => (r.getSeq(0), r.getSeq(1)))
+      .collect().asInstanceOf[Array[(Seq[Any], Seq[Any])]]
+    val brRules = dataset.sparkSession.sparkContext.broadcast(rules)
+
+    val dt = dataset.schema($(featuresCol)).dataType
+    // For each rule, examine the input items and summarize the consequents
+    val predictUDF = udf((items: Seq[_]) => {
+      if (items != null) {
+        val itemset = items.toSet
+        brRules.value.flatMap(rule =>
```
jkbradley: style:
```
      brRules.value.flatMap { rule =>
        ...
      }
```

jkbradley (commented):
Looks good, just small comments

jkbradley:
test this please

jkbradley:
I'm going to go ahead and merge this after tests to make sure it's in 2.2, but can you please send a follow-up for my last 2 comments?  Thanks!

hhbyyh:
Sorry to miss your comments. I can send a follow-up together with document.

jkbradley:
No problem, thanks!  Could you please create a subtask for docs?

Merging with master

PR #10677, opened 2016-01-09
[SPARK-12706] [SQL] grouping() and grouping_id()

davies:
Grouping() returns a column is aggregated or not, grouping_id() returns the aggregation levels.

grouping()/grouping_id() could be used with window function, but does not work in having/sort clause, will be fixed by another PR. 

The GROUPING__ID/grouping_id() in Hive is wrong (according to docs), we also did it wrongly, this PR change that to match the behavior in most databases (also the docs of Hive).

PR #15416, opened 2016-10-10
[SPARK-17849] [SQL] Fix NPE problem when using grouping sets

yangw1234:

Prior this pr, the following code would cause an NPE:
`case class point(a:String, b:String, c:String, d: Int)`

`val data = Seq(
point("1","2","3", 1),
point("4","5","6", 1),
point("7","8","9", 1)
)`
`sc.parallelize(data).toDF().registerTempTable("table")`
`spark.sql("select a, b, c, count(d) from table group by a, b, c GROUPING SETS ((a)) ").show()`

The reason is that when the grouping_id() behavior was changed in #10677, some code (which should be changed) was left out. 

Take the above code for example, prior #10677, the bit mask for set "(a)" was `001`, while after #10677 the bit mask was changed to `011`. However, the `nonNullBitmask` was not changed accordingly.

This pr will fix this problem.
## How was this patch tested?

add integration tests


yangw1234:
cc @davies Would you help reviewing this?

yangw1234:
also cc @hvanhovell

sql/core/src/test/scala/org/apache/spark/sql/SQLQuerySuite.scala
```diff
@@ -2189,6 +2189,24 @@ class SQLQuerySuite extends QueryTest with SharedSQLContext {
     }
   }
 
+  test("SPARK-17849: grouping set throws NPE") {
```
rxin: maybe we can move this into SQLQueryTestSuite, by creating a new grouping_set.q file??

yangw1234: @rxin done

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
```diff
@@ -298,10 +298,11 @@ class Analyzer(
           case other => Alias(other, other.toString)()
         }
 
-        val nonNullBitmask = x.bitmasks.reduce(_ & _)
+        val nonNullBitmask = ~ x.bitmasks.reduce(_ | _)
```
rxin: +1

yangw1234: Ok, I'll do it.

yangw1234: @hvanhovell @rxin comments are added

davies:
@yangw1234 Thanks for working on this, could you also double check that all the places that use bitmasks are correct?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
```diff
@@ -298,10 +298,14 @@ class Analyzer(
           case other => Alias(other, other.toString)()
         }
 
-        val nonNullBitmask = x.bitmasks.reduce(_ & _)
+        // The left most bit in the bitmasks corresponds to the last expression in groupByAliases
+        // with 0 indicating this expression is in the grouping set. The following line of code
+        // calculates the bit mask representing the expressions that exist in all the grouping sets.
+        val nonNullBitmask = ~ x.bitmasks.reduce(_ | _)
```
yangw1234: Do you mean `((nonNullBitmask >> (attrLength - idx - 1)) & 1) == 1`? We can only test on `0` if we left shift `1`, right? @davies

yangw1234:
@davies Other places all seem to be correct.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
```diff
@@ -298,10 +298,14 @@ class Analyzer(
           case other => Alias(other, other.toString)()
         }
 
-        val nonNullBitmask = x.bitmasks.reduce(_ & _)
+        // The rightmost bit in the bitmasks corresponds to the last expression in groupByAliases with 0
+        // indicating this expression is in the grouping set. The following line of code calculates the
+        // bitmask representing the expressions that exist in all the grouping sets (also indicated by 0).
+        val nonNullBitmask = x.bitmasks.reduce(_ | _)
```
davies: Should we call this `nullBitmask` now? (1 means it's nullable)

rxin:
@yangw1234 can you fix the scala styles?

yangw1234:
scala style fixed.
 I didn't notice. Sorry for the delay. @rxin

hvanhovell:
retest this please

davies:
LGTM, pending tests

hvanhovell:
retest this please

hvanhovell:
I'll merge after a successfull test run

hvanhovell:
LGTM - Merging to master/2.1/2.0. Thanks!

PR #15417, opened 2016-10-10
[SPARK-17851][SQL][TESTS] Make sure all test sqls in catalyst pass checkAnalysis

jiangxb1987:

Currently we have several tens of test sqls in catalyst will fail at `SimpleAnalyzer.checkAnalysis`, we should make sure they are valid.

This PR makes the following changes:
1. Apply `checkAnalysis` on plans that tests `Optimizer` rules, but don't require the testcases for `Parser`/`Analyzer` pass `checkAnalysis`;
2. Fix testcases for `Optimizer` that would have fall.
## How was this patch tested?

Apply `SimpleAnalyzer.checkAnalysis` on plans in `PlanTest.comparePlans`, update invalid test cases.

### DIFF
--- a/sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/analysis/AnalysisTest.scala
+++ b/sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/analysis/AnalysisTest.scala
@@ -55,6 +55,14 @@ trait AnalysisTest extends PlanTest {
     comparePlans(actualPlan, expectedPlan)
   }
 
+  protected override def comparePlans(
+      plan1: LogicalPlan,
+      plan2: LogicalPlan,
+      checkAnalysis: Boolean = false): Unit = {
+    // Analysis tests may have not been fully resolved, so skip checkAnalysis.
+    super.comparePlans(plan1, plan2, checkAnalysis = false)
+  }
+
   protected def assertAnalysisSuccess(
       inputPlan: LogicalPlan,
       caseSensitive: Boolean = true): Unit = {
--- a/sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/analysis/AnalysisTest.scala
+++ b/sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/analysis/AnalysisTest.scala
@@ -60,7 +60,7 @@ trait AnalysisTest extends PlanTest {
       plan2: LogicalPlan,
       checkAnalysis: Boolean = false): Unit = {
     // Analysis tests may have not been fully resolved, so skip checkAnalysis.
-    super.comparePlans(plan1, plan2, checkAnalysis = false)
+    super.comparePlans(plan1, plan2, checkAnalysis)
   }
 
   protected def assertAnalysisSuccess(
--- a/sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/optimizer/AggregateOptimizeSuite.scala
+++ b/sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/optimizer/AggregateOptimizeSuite.scala
@@ -59,9 +59,9 @@ class AggregateOptimizeSuite extends PlanTest {
   }
 
   test("Remove aliased literals") {
-    val query = testRelation.select('a, Literal(1).as('y)).groupBy('a, 'y)(sum('b))
+    val query = testRelation.select('a, 'b, Literal(1).as('y)).groupBy('a, 'y)(sum('b))
     val optimized = Optimize.execute(analyzer.execute(query))
-    val correctAnswer = testRelation.select('a, Literal(1).as('y)).groupBy('a)(sum('b)).analyze
+    val correctAnswer = testRelation.select('a, 'b, Literal(1).as('y)).groupBy('a)(sum('b)).analyze
 
     comparePlans(optimized, correctAnswer)
   }
--- a/sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/optimizer/BooleanSimplificationSuite.scala
+++ b/sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/optimizer/BooleanSimplificationSuite.scala
@@ -41,7 +41,8 @@ class BooleanSimplificationSuite extends PlanTest with PredicateHelper {
         PruneFilters) :: Nil
   }
 
-  val testRelation = LocalRelation('a.int, 'b.int, 'c.int, 'd.string)
+  val testRelation = LocalRelation('a.int, 'b.int, 'c.int, 'd.string,
+    'e.boolean, 'f.boolean, 'g.boolean, 'h.boolean)
 
   val testRelationWithData = LocalRelation.fromExternalRows(
     testRelation.output, Seq(Row(1, 2, 3, "abc"))
@@ -101,52 +102,52 @@ class BooleanSimplificationSuite extends PlanTest with PredicateHelper {
       'a === 'b || 'b > 3 && 'a > 3 && 'a < 5)
   }
 
-  test("a && (!a || b)") {
-    checkCondition('a && (!'a || 'b ), 'a && 'b)
+  test("e && (!e || f)") {
+    checkCondition('e && (!'e || 'f ), 'e && 'f)
 
-    checkCondition('a && ('b || !'a ), 'a && 'b)
+    checkCondition('e && ('f || !'e ), 'e && 'f)
 
-    checkCondition((!'a || 'b ) && 'a, 'b && 'a)
+    checkCondition((!'e || 'f ) && 'e, 'f && 'e)
 
-    checkCondition(('b || !'a ) && 'a, 'b && 'a)
+    checkCondition(('f || !'e ) && 'e, 'f && 'e)
   }
 
-  test("a < 1 && (!(a < 1) || b)") {
-    checkCondition('a < 1 && (!('a < 1) || 'b), ('a < 1) && 'b)
-    checkCondition('a < 1 && ('b || !('a < 1)), ('a < 1) && 'b)
+  test("a < 1 && (!(a < 1) || f)") {
+    checkCondition('a < 1 && (!('a < 1) || 'f), ('a < 1) && 'f)
+    checkCondition('a < 1 && ('f || !('a < 1)), ('a < 1) && 'f)
 
-    checkCondition('a <= 1 && (!('a <= 1) || 'b), ('a <= 1) && 'b)
-    checkCondition('a <= 1 && ('b || !('a <= 1)), ('a <= 1) && 'b)
+    checkCondition('a <= 1 && (!('a <= 1) || 'f), ('a <= 1) && 'f)
+    checkCondition('a <= 1 && ('f || !('a <= 1)), ('a <= 1) && 'f)
 
-    checkCondition('a > 1 && (!('a > 1) || 'b), ('a > 1) && 'b)
-    checkCondition('a > 1 && ('b || !('a > 1)), ('a > 1) && 'b)
+    checkCondition('a > 1 && (!('a > 1) || 'f), ('a > 1) && 'f)
+    checkCondition('a > 1 && ('f || !('a > 1)), ('a > 1) && 'f)
 
-    checkCondition('a >= 1 && (!('a >= 1) || 'b), ('a >= 1) && 'b)
-    checkCondition('a >= 1 && ('b || !('a >= 1)), ('a >= 1) && 'b)
+    checkCondition('a >= 1 && (!('a >= 1) || 'f), ('a >= 1) && 'f)
+    checkCondition('a >= 1 && ('f || !('a >= 1)), ('a >= 1) && 'f)
   }
 
-  test("a < 1 && ((a >= 1) || b)") {
-    checkCondition('a < 1 && ('a >= 1 || 'b ), ('a < 1) && 'b)
-    checkCondition('a < 1 && ('b || 'a >= 1), ('a < 1) && 'b)
+  test("a < 1 && ((a >= 1) || f)") {
+    checkCondition('a < 1 && ('a >= 1 || 'f ), ('a < 1) && 'f)
+    checkCondition('a < 1 && ('f || 'a >= 1), ('a < 1) && 'f)
 
-    checkCondition('a <= 1 && ('a > 1 || 'b ), ('a <= 1) && 'b)
-    checkCondition('a <= 1 && ('b || 'a > 1), ('a <= 1) && 'b)
+    checkCondition('a <= 1 && ('a > 1 || 'f ), ('a <= 1) && 'f)
+    checkCondition('a <= 1 && ('f || 'a > 1), ('a <= 1) && 'f)
 
-    checkCondition('a > 1 && (('a <= 1) || 'b), ('a > 1) && 'b)
-    checkCondition('a > 1 && ('b || ('a <= 1)), ('a > 1) && 'b)
+    checkCondition('a > 1 && (('a <= 1) || 'f), ('a > 1) && 'f)
+    checkCondition('a > 1 && ('f || ('a <= 1)), ('a > 1) && 'f)
 
-    checkCondition('a >= 1 && (('a < 1) || 'b), ('a >= 1) && 'b)
-    checkCondition('a >= 1 && ('b || ('a < 1)), ('a >= 1) && 'b)
+    checkCondition('a >= 1 && (('a < 1) || 'f), ('a >= 1) && 'f)
+    checkCondition('a >= 1 && ('f || ('a < 1)), ('a >= 1) && 'f)
   }
 
   test("DeMorgan's law") {
-    checkCondition(!('a && 'b), !'a || !'b)
+    checkCondition(!('e && 'f), !'e || !'f)
 
-    checkCondition(!('a || 'b), !'a && !'b)
+    checkCondition(!('e || 'f), !'e && !'f)
 
-    checkCondition(!(('a && 'b) || ('c && 'd)), (!'a || !'b) && (!'c || !'d))
+    checkCondition(!(('e && 'f) || ('g && 'h)), (!'e || !'f) && (!'g || !'h))
 
-    checkCondition(!(('a || 'b) && ('c || 'd)), (!'a && !'b) || (!'c && !'d))
+    checkCondition(!(('e || 'f) && ('g || 'h)), (!'e && !'f) || (!'g && !'h))
   }
 
   private val caseInsensitiveConf = new SQLConf().copy(SQLConf.CASE_SENSITIVE -> false)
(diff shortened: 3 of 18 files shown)


jiangxb1987:
cc @gatorsmile

gatorsmile:
Is this the only one?

If you add the commands in  the function `comparePlans` of `PlanTest.scala`, you can see more. 
```Scala
SimpleAnalyzer.checkAnalysis(plan1)
SimpleAnalyzer.checkAnalysis(plan2)
```

Could you go over it and see whether you can resolve all of them? Thanks!

jiangxb1987:
Yes - there are several tens of test sql that fails checkAnalysis, should we update them all?

gatorsmile:
You can make a try.

jiangxb1987:
@gatorsmile I've updated all testcases that failed checkAnalysis in Optimizer related testsuites, for those testcases for Parser/Analyzer we don't require them to pass checkAnalysis. Please have a look at the change when you have time, thank you!

jiangxb1987:
@gatorsmile This PR is ready for review.

gatorsmile:
@jiangxb1987 Can you leave a comment on the PR changes to explain why you made these changes? You know, reviewing these changes is not easy. Thanks!

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/analysis/ResolveNaturalJoinSuite.scala
```diff
@@ -31,39 +31,46 @@ class ResolveNaturalJoinSuite extends AnalysisTest {
...
+  lazy val r4New = LocalRelation(cNotNull, bNotNullNew)
+
+  private def newAttribute(a: AttributeReference): Attribute =
+    a.withExprId(NamedExpression.newExprId)
 
   test("natural/using inner join") {
     val naturalPlan = r1.join(r2, NaturalJoin(Inner), None)
     val usingPlan = r1.join(r2, UsingJoin(Inner, Seq(UnresolvedAttribute("a"))), None)
-    val expected = r1.join(r2, Inner, Some(EqualTo(a, a))).select(a, b, c)
+    val expected = r1.join(r2New, Inner, Some(EqualTo(a, aNew))).select(a, b, c)
```
jiangxb1987: Previous `EqualTo(a, a)` introduces a conflicting attributes exception.

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/optimizer/AggregateOptimizeSuite.scala
```diff
@@ -58,9 +58,9 @@ class AggregateOptimizeSuite extends PlanTest {
   }
 
   test("Remove aliased literals") {
-    val query = testRelation.select('a, Literal(1).as('y)).groupBy('a, 'y)(sum('b))
+    val query = testRelation.select('a, 'b, Literal(1).as('y)).groupBy('a, 'y)(sum('b))
     val optimized = Optimize.execute(analyzer.execute(query))
-    val correctAnswer = testRelation.select('a, Literal(1).as('y)).groupBy('a)(sum('b)).analyze
+    val correctAnswer = testRelation.select('a, 'b, Literal(1).as('y)).groupBy('a)(sum('b)).analyze
```
jiangxb1987: Previous sql don't have `b` in projectList so `sum('b)` can't get its reference.

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/optimizer/BooleanSimplificationSuite.scala
```diff
@@ -91,51 +99,51 @@ class BooleanSimplificationSuite extends PlanTest with PredicateHelper {
   }
 
   test("a && (!a || b)") {
-    checkCondition('a && (!'a || 'b ), 'a && 'b)
```
jiangxb1987: The operator `And`/`Or` requires both children to be `Boolean`

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/optimizer/ColumnPruningSuite.scala
```diff
@@ -266,8 +266,8 @@ class ColumnPruningSuite extends PlanTest {
 
   test("Column pruning on Window with useless aggregate functions") {
     val input = LocalRelation('a.int, 'b.string, 'c.double, 'd.int)
-    val winSpec = windowSpec('a :: Nil, 'b.asc :: Nil, UnspecifiedFrame)
```
jiangxb1987: The attribute `'b` is neither in the group by expressions nor an aggregate function, so we use `'d` instead.

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/optimizer/FilterPushdownSuite.scala
```diff
@@ -1016,6 +1016,8 @@ class FilterPushdownSuite extends PlanTest {
     val correctAnswer = x.where("x.a".attr === 5).join(y.where("y.a".attr === 5),
         condition = Some("x.a".attr === Rand(10) && "y.b".attr === 5))
 
-    comparePlans(Optimize.execute(originalQuery.analyze), correctAnswer.analyze)
```
jiangxb1987: We can skip `checkAnalysis` for this testcase, or remove this one.

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/optimizer/LimitPushdownSuite.scala
```diff
@@ -70,19 +70,19 @@ class LimitPushdownSuite extends PlanTest {
   }
 
   test("Union: no limit to both sides if children having smaller limit values") {
-    val unionQuery = Union(testRelation.limit(1), testRelation2.select('d).limit(1)).limit(2)
```
jiangxb1987: the two relations for `Union` operator should have same output length.

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/optimizer/OptimizeCodegenSuite.scala
```diff
@@ -51,10 +51,10 @@ class OptimizeCodegenSuite extends PlanTest {
   test("Nested CaseWhen Codegen.") {
     assertEquivalent(
       CaseWhen(
-        Seq((CaseWhen(Seq((TrueLiteral, Literal(1))), Literal(2)), Literal(3))),
```
jiangxb1987: It should be `BooleanType` instead of `IntegerType`.

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/optimizer/OuterJoinEliminationSuite.scala
```diff
@@ -200,15 +200,15 @@ class OuterJoinEliminationSuite extends PlanTest {
 
     val originalQuery =
       x.join(y, FullOuter, Option("x.a".attr === "y.d".attr))
-        .where(Coalesce("y.e".attr :: "x.a".attr :: Nil))
```
jiangxb1987: The condition expression should be `boolean`.

jiangxb1987:
@gatorsmile I've update the PR description and left some comments on changed codes, please comment if there were anything that still unclear. Thank you!

jiangxb1987:
ping @gatorsmile

gatorsmile:
Sorry, will review it soon. Thanks!

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/analysis/ResolveNaturalJoinSuite.scala
```diff
@@ -31,39 +31,46 @@ class ResolveNaturalJoinSuite extends AnalysisTest {
...
+  lazy val aNew = newAttribute(a)
+  lazy val bNotNullNew = newAttribute(bNotNull)
   lazy val r1 = LocalRelation(b, a)
   lazy val r2 = LocalRelation(c, a)
   lazy val r3 = LocalRelation(aNotNull, bNotNull)
   lazy val r4 = LocalRelation(cNotNull, bNotNull)
+  lazy val r2New = LocalRelation(c, aNew)
+  lazy val r4New = LocalRelation(cNotNull, bNotNullNew)
+
+  private def newAttribute(a: AttributeReference): Attribute =
```
gatorsmile: `newAttribute` -> `deDuplicatedAttribute`

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/analysis/ResolveNaturalJoinSuite.scala
```diff
@@ -31,39 +31,46 @@ class ResolveNaturalJoinSuite extends AnalysisTest {
   lazy val aNotNull = a.notNull
   lazy val bNotNull = b.notNull
   lazy val cNotNull = c.notNull
+  lazy val aNew = newAttribute(a)
```
gatorsmile: `a` -> `aDeduped`

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/analysis/ResolveNaturalJoinSuite.scala
```diff
@@ -31,39 +31,46 @@ class ResolveNaturalJoinSuite extends AnalysisTest {
...
+  lazy val bNotNullNew = newAttribute(bNotNull)
   lazy val r1 = LocalRelation(b, a)
   lazy val r2 = LocalRelation(c, a)
   lazy val r3 = LocalRelation(aNotNull, bNotNull)
   lazy val r4 = LocalRelation(cNotNull, bNotNull)
+  lazy val r2New = LocalRelation(c, aNew)
```
gatorsmile: `r2New` -> `r2Deduped`

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/plans/PlanTest.scala
```diff
@@ -81,6 +91,6 @@ abstract class PlanTest extends SparkFunSuite with PredicateHelper {
 
   /** Fails the test if the two expressions do not match */
   protected def compareExpressions(e1: Expression, e2: Expression): Unit = {
-    comparePlans(Filter(e1, OneRowRelation), Filter(e2, OneRowRelation))
+    comparePlans(Filter(e1, OneRowRelation), Filter(e2, OneRowRelation), false)
```
gatorsmile: -> `checkAnalysis = false`

gatorsmile:
Thanks for your work! I finished the first round of reviews.

jiangxb1987:
Thank you @gatorsmile ! I'll address your comment ASAP!

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/analysis/AnalysisTest.scala
```diff
@@ -51,6 +51,14 @@ trait AnalysisTest extends PlanTest {
     comparePlans(actualPlan, expectedPlan)
   }
 
+  protected override def comparePlans(
+      plan1: LogicalPlan,
+      plan2: LogicalPlan,
+      checkAnalysis: Boolean = false): Unit = {
+    // Analysis tests may have not been fully resolved, so skip checkAnalysis.
+    super.comparePlans(plan1, plan2, checkAnalysis = false)
```
gatorsmile: `super.comparePlans(plan1, plan2, checkAnalysis)`

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/optimizer/BooleanSimplificationSuite.scala
```diff
@@ -40,12 +40,16 @@ class BooleanSimplificationSuite extends PlanTest with PredicateHelper {
...
-
-  private def checkCondition(input: Expression, expected: Expression): Unit = {
-    val plan = testRelation.where(input).analyze
+  val testRelation = LocalRelation(
+    'a.int, 'b.int, 'c.int, 'd.string, 'e.boolean, 'f.boolean, 'g.boolean, 'h.boolean)
+
+  private def checkCondition(
+      input: Expression,
+      expected: Expression,
+      relation: LocalRelation = testRelation): Unit = {
```
gatorsmile: You do not need to change the function interface, right?

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/optimizer/FilterPushdownSuite.scala
gatorsmile: Sorry, we are unable to merge this PR until you fix the above issue.

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/optimizer/LimitPushdownSuite.scala
gatorsmile: This is resolved.

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/optimizer/FilterPushdownSuite.scala
jiangxb1987: We have to support `PullOutNondeterministic` on `Join` operator, since`Join.condition` is where non-deterministic expressions can only appear, we can rewrite the plan by:
1、Pull out non-deterministic part of `condition`, for example, `a = b && rand() > 0 && a > 1` has deterministic part `a = b` and the rest is non-deterministic;
2、To handle the cases when multiple nondeterministic expressions exists in `condition`, we have to split conjunctive predicates, for every predicate that is nondeterministic, we insert a new `Join` operator like(`j` represents the original Join operator):
```
Join(
  j,
  LocalRelation(predicate.collect { case n: Nondeterministic => n }),
  Inner,
  Some(rewrite(predicate))
)
```
Since this added much `Join` operators, I doubt the performance will be pretty bad. Do you have any advices?

jiangxb1987: The reason I don't use `Project` operator is that might change the result correctness, for example, suppose we have join condition `rand(1) > 0 &&  rand(11) < 0`, to use `Project` operator, we will always eval the both rand functions, which will change the result.

jiangxb1987: Also cc @cloud-fan

gatorsmile: Q: We are facing the same issue in the non-deterministic pull out for `Filter`, right?

jiangxb1987: Yes you're quite right. We don't handle non-deterministic expressions in `Filter` here either.

gatorsmile:
```SQL
select actno from act where rand() < 0.5 and rand() >= 0.5
```

Based on your conclusion, this should never return any value, right? However, I checked the other DBMS and the filter conditions are not always false.

gatorsmile:
```
db2 "select a.actno from act a, act b where a.actno = b.actno and rand() < 0.5 and rand() >= 0.5"
```

In the above example, the non-deterministic functions in the join conditions are also evaluated twice in DBMS.

cloud-fan: Filter is allowed to have nondeterministic condition.

For Join, actually the rule to handle nondeterministic condition is unnecessary, as join condition is always deterministic, we can remove these logic in the optimizer rule and its corresponding tests.

jiangxb1987: @cloud-fan Perhaps we should support nondeterministic condition in `Join` operator, so that we keep the same behavior with HIVE.

cloud-fan: sounds reasonable, but this worth a new PR.(Seems we already have a PR for it?)

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/optimizer/SimplifyCastsSuite.scala
```diff
@@ -44,7 +44,9 @@ class SimplifyCastsSuite extends PlanTest {
     val input = LocalRelation('a.array(ArrayType(IntegerType, true)))
     val plan = input.select('a.cast(ArrayType(IntegerType, false)).as("casted")).analyze
     val optimized = Optimize.execute(plan)
-    comparePlans(optimized, plan)
+    // Though cast from `ArrayType(IntegerType, true)` to `ArrayType(IntegerType, false)` is not
+    // allowed, here we just ensure that `SimplifyCasts` rule respect the plan.
+    comparePlans(optimized, plan, checkAnalysis = false)
```
kiszk: LGTM. As a comment pointed out, this test do not expect the plan is executed correctly.

kiszk (approved):

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/optimizer/FilterPushdownSuite.scala
jiangxb1987: I'm afraid not, but I'd like to find out whether we could support that.

gatorsmile:
cc @jiangxb1987

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/optimizer/FilterPushdownSuite.scala
jiangxb1987: This is not easy to fix, I added a TODO here.

shaneknapp:
test this please

jiangxb1987:
This is ready for review. @gatorsmile @cloud-fan

gatorsmile:
LGTM

gatorsmile:
Thanks! Merging to master.

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/analysis/AnalysisTest.scala
```diff
@@ -55,6 +55,14 @@ trait AnalysisTest extends PlanTest {
     comparePlans(actualPlan, expectedPlan)
   }
 
+  protected override def comparePlans(
+      plan1: LogicalPlan,
+      plan2: LogicalPlan,
+      checkAnalysis: Boolean = false): Unit = {
+    // Analysis tests may have not been fully resolved, so skip checkAnalysis.
```
cloud-fan: this comment explains why we set the default value of `checkAnalysis` to false, right?

jiangxb1987: yep

PR #15421, opened 2016-10-10
[SPARK-17811] SparkR cannot parallelize data.frame with NA or NULL in Date columns

falaki:

NA date values are serialized as "NA" and NA time values are serialized as NaN from R. In the backend we did not have proper logic to deal with them. As a result we got an IllegalArgumentException for Date and wrong value for time. This PR adds support for deserializing NA as Date and Time.
## How was this patch tested?
- [x] TODO


falaki:
@shivaram can I nominate this patch for 2.0 branch?

felixcheung:
This fails on AppVeyor - any idea?
```
. Error: SPARK-17811: can create DataFrame containing NA as date and time (@test_sparkSQL.R#388) 
org.apache.spark.SparkException: Job aborted due to stage failure: Task 0 in stage 105.0 failed 1 times, most recent failure: Lost task 0.0 in stage 105.0 (TID 109, localhost): java.lang.NegativeArraySizeException
```

R/pkg/DESCRIPTION
```diff
@@ -11,7 +11,8 @@ Authors@R: c(person("Shivaram", "Venkataraman", role = c("aut", "cre"),
                     email = "felixcheung@apache.org"),
              person(family = "The Apache Software Foundation", role = c("aut", "cph")))
 URL: http://www.apache.org/ http://spark.apache.org/
-BugReports: https://cwiki.apache.org/confluence/display/SPARK/Contributing+to+Spark#ContributingtoSpark-ContributingBugReports
+BugReports: https://cwiki.apache.org/confluence/display/SPARK/Contributing+to
+    +Spark#ContributingtoSpark-ContributingBugReports
```
felixcheung: Do we know if this works if broken up into 2 lines?

R/pkg/inst/tests/testthat/test_sparkSQL.R
```diff
@@ -379,6 +379,17 @@ test_that("create DataFrame with different data types", {
...
 })
 
+test_that("SPARK-17811: can create DataFrame containing NA as date and time", {
+  df <- data.frame(
+    id = 1:2,
+    time = c(as.POSIXlt("2016-01-10"), NA),
+    date = c(as.Date("2016-10-01"), NA))
+
+  DF <- collect(createDataFrame(df))
+  expect_true(is.na(DF$date[2]))
```
felixcheung: We should check the valid dates are correctly serialized too?

falaki (commented):
Thanks @felixcheung for review. Addressed.

shivaram:
Yeah it should be fine to merge this into branch-2.0 when it is ready

felixcheung:
hmm, still the same error in the new test case in appveyor
```
1. Error: SPARK-17811: can create DataFrame containing NA as date and time (@test_sparkSQL.R#388) 
org.apache.spark.SparkException: Job aborted due to stage failure: Task 0 in stage 105.0 failed 1 times, most recent failure: Lost task 0.0 in stage 105.0 (TID 109, localhost): java.lang.NegativeArraySizeException
	at org.apache.spark.api.r.SerDe$.readStringBytes(SerDe.scala:110)
	at org.apache.spark.api.r.SerDe$.readString(SerDe.scala:119)
	at org.apache.spark.api.r.SerDe$.readDate(SerDe.scala:128)

```

wangmiao1981:
@falaki I patched your fix to a clean build. I still see the following error:

> df <- data.frame(Date = as.Date(c(rep("2016-01-10", 10), "NA", "NA")), id = 1:12)
> 
> dim(createDataFrame(df))
16/10/11 11:51:00 ERROR Executor: Exception in task 0.0 in stage 1.0 (TID 1)
java.lang.NegativeArraySizeException
	at org.apache.spark.api.r.SerDe$.readStringBytes(SerDe.scala:110)
	at org.apache.spark.api.r.SerDe$.readString(SerDe.scala:119)
	at org.apache.spark.api.r.SerDe$.readDate(SerDe.scala:128)
	at org.apache.spark.api.r.SerDe$.readTypedObject(SerDe.scala:77)
	at org.apache.spark.api.r.SerDe$.readObject(SerDe.scala:61)
	at org.apache.spark.sql.api.r.SQLUtils$$anonfun$bytesToRow$1.apply(SQLUtils.scala:161)
	at org.apache.spark.sql.api.r.SQLUtils$$anonfun$bytesToRow$1.apply(SQLUtils.scala:160)
	at scala.collection.immutable.Range.foreach(Range.scala:160)
	at scala.collection.AbstractTraversable.map(Traversable.scala:104)
	at org.apache.spark.sql.api.r.SQLUtils$.bytesToRow(SQLUtils.scala:160)
	at org.apache.spark.sql.api.r.SQLUtils$$anonfun$5.apply(SQLUtils.scala:138)
	at org.apache.spark.sql.api.r.SQLUtils$$anonfun$5.apply(SQLUtils.scala:138)
	at scala.collection.Iterator$$anon$11.next(Iterator.scala:409)
	at scala.collection.Iterator$$anon$11.next(Iterator.scala:409)
	at scala.collection.Iterator$$anon$11.next(Iterator.scala:409)
	at org.apache.spark.sql.catalyst.expressions.GeneratedClass$GeneratedIterator.agg_doAggregateWithoutKey$(Unknown Source)
	at org.apache.spark.sql.execution.BufferedRowIterator.hasNext(BufferedRowIterator.java:43)
	at org.apache.spark.sql.execution.WholeStageCodegenExec$$anonfun$8$$anon$1.hasNext(WholeStageCodegenExec.scala:372)
	at scala.collection.Iterator$$anon$11.hasNext(Iterator.scala:408)
	at org.apache.spark.shuffle.sort.BypassMergeSortShuffleWriter.write(BypassMergeSortShuffleWriter.java:126)
	at org.apache.spark.scheduler.ShuffleMapTask.runTask(ShuffleMapTask.scala:96)
	at org.apache.spark.scheduler.ShuffleMapTask.runTask(ShuffleMapTask.scala:53)
	at org.apache.spark.scheduler.Task.run(Task.scala:99)

I am still debugging. It seems that source on R side has some issue.

core/src/main/scala/org/apache/spark/api/r/SerDe.scala
```diff
@@ -125,15 +125,24 @@ private[spark] object SerDe {
   }
 
   def readDate(in: DataInputStream): Date = {
-    Date.valueOf(readString(in))
+    val inStr = readString(in)
```
wangmiao1981: readString will finally gets a negative len, which causes failure.

falaki: That is only on Windows right? I added a catch for it.

falaki:
@wangmiao1981 thanks for testing on Windows. I added a check for this. Would you please try again and let me know? Unfortunately, I don't have access to a windows box for testing.

wangmiao1981:
@falaki I saw the exception on Mac too. But I don't find the root cause of negative length in the input stream. Catching the exception will solve the problem. Do you want to explore the reason of negative index?

wangmiao1981:
MacBook Pro (Retina, 15-inch, Mid 2015)
This is the machine that I test the patch on.

wangmiao1981:
New test on Mac:

> df <- data.frame(Date = as.Date(c(rep("2016-01-10", 10), "NA", "NA")), id = 1:12)
> 
> dim(createDataFrame(df))
16/10/11 12:10:30 ERROR Executor: Exception in task 0.0 in stage 1.0 (TID 1)
java.lang.IllegalArgumentException: Invalid type N
	at org.apache.spark.api.r.SerDe$.readTypedObject(SerDe.scala:86)
	at org.apache.spark.api.r.SerDe$.readObject(SerDe.scala:61)
	at org.apache.spark.sql.api.r.SQLUtils$$anonfun$bytesToRow$1.apply(SQLUtils.scala:161)
	at org.apache.spark.sql.api.r.SQLUtils$$anonfun$bytesToRow$1.apply(SQLUtils.scala:160)
	at scala.collection.immutable.Range.foreach(Range.scala:160)
	at scala.collection.AbstractTraversable.map(Traversable.scala:104)
	at org.apache.spark.sql.api.r.SQLUtils$.bytesToRow(SQLUtils.scala:160)
	at org.apache.spark.sql.api.r.SQLUtils$$anonfun$5.apply(SQLUtils.scala:138)
	at org.apache.spark.sql.api.r.SQLUtils$$anonfun$5.apply(SQLUtils.scala:138)
	at scala.collection.Iterator$$anon$11.next(Iterator.scala:409)
	at scala.collection.Iterator$$anon$11.next(Iterator.scala:409)
	at scala.collection.Iterator$$anon$11.next(Iterator.scala:409)
	at org.apache.spark.sql.catalyst.expressions.GeneratedClass$GeneratedIterator.agg_doAggregateWithoutKey$(Unknown Source)
	at org.apache.spark.sql.execution.BufferedRowIterator.hasNext(BufferedRowIterator.java:43)
	at org.apache.spark.sql.execution.WholeStageCodegenExec$$anonfun$8$$anon$1.hasNext(WholeStageCodegenExec.scala:372)
	at scala.collection.Iterator$$anon$11.hasNext(Iterator.scala:408)
	at org.apache.spark.shuffle.sort.BypassMergeSortShuffleWriter.write(BypassMergeSortShuffleWriter.java:126)
	at org.apache.spark.scheduler.ShuffleMapTask.runTask(ShuffleMapTask.scala:96)
	at org.apache.spark.scheduler.ShuffleMapTask.runTask(ShuffleMapTask.scala:53)
	at org.apache.spark.scheduler.Task.run(Task.scala:99)

wangmiao1981:
I think we should find out the root cause of the negative length of "NA" field. Yesterday, I debugged R side and I have not found out the reason yet.

wangmiao1981:
It might not be a R specific issue. I am trying to create a test case on Scala side in SQLUtilsSuite.scala.

falaki:
My guess is that on windows R serialization behaves differently and serializes `NA` as `null`. Unfortunately, I don't have a windows machine to verify. Would you please test that?

wangmiao1981:
I did test on Mac, not on Windows. I don't have windows setup either. The very first issue that I saw is the negative index issue.

In addition, returning `null` when exception is caught is incorrect. The caller `readTypedObjects` will throw invalid datatype exception.

falaki:
@wangmiao1981 I don't get the exception that you reported on Mac. Also note that the unit test is passing on Linux. I am not sure why returning null is an issue.

wangmiao1981:
That's interesting. I patched your change to a clean checkout and simply tested against the example on the JIRA. It throws the above exception.

          val obj = (sqlSerDe._1)(dis, dataType)
          if (obj == null) {
            throw new IllegalArgumentException (s"Invalid type $dataType")<===== this line
          } else {
            obj
          } 

I have no clue why it fails on my laptop. I can test on my own server (ubuntu) tonight.

wangmiao1981:
I suspect that it could be related to my R installation:

localhost:~ mwang$ R

R version 3.3.0 (2016-05-03) -- "Supposedly Educational"
Copyright (C) 2016 The R Foundation for Statistical Computing
Platform: x86_64-apple-darwin13.4.0 (64-bit)

But I am not sure yet.

core/src/main/scala/org/apache/spark/api/r/SerDe.scala
```diff
@@ -125,15 +125,34 @@ private[spark] object SerDe {
...
-    Date.valueOf(readString(in))
+    try {
+      val inStr = readString(in)
+      if (inStr == "NA") {
+        null
+      } else {
+        Date.valueOf(inStr)
+      }
+    } catch {
+      // On windows we get NegativeArraySizeException for NAs in R
```
shivaram: I dont think we should use this exception unless we can figure out why we `NegativeSizeExceptions` -- we haven't run into this for other types right ?

felixcheung:
It's possible with R version - Jenkins is running 3.1.1 I think, the minimal supported version.
AppVeyor is running 3.3.2 I believe, which matches closer to the one @wangmiao1981 has

falaki:
@felixcheung and @wangmiao1981 thanks! This is good point. I will try testing it on different version of R.

HyukjinKwon:
FWIW, I wrote a bunch of scripts to automatically launch a build via AppVeyor via @spark-test account with a pretty string such as 

> Build started: [CORE] `org.apache.spark.storage.DiskStoreSuite` [PR-15320](https://ci.appveyor.com/project/spark-test/spark/branch/097F2F95-4748-4435-967F-98980DB2112E)

or

> Build started: [R] `ALL` [PR-15320](https://ci.appveyor.com/project/spark-test/spark/branch/097F2F95-4748-4435-967F-98980DB2112E)

Please feel free to cc me if any of you wants.

wangmiao1981:
Last night, I debugged the problem again. Now I can reproduce it as follows:

>  a <- as.Date("2016-10-01", "NA")
> b <- as.data.frame(a)
> c <- createDataFrame(b)

> a
[1] NA
> b
     a
1 <NA>
 I compared with the working case. R side function calls are the same for both cases and data type inferences are correct. R side writes data as `raw` type and the backend encounters problem when casting the byte stream based on schema. 

I continue debugging it now and hope to find the cause of why reading bytes to date type fails.

For the above example, I don't know why "2016-10-01" is missed after `as.Date`. But based on my debug message, the number of args and value (`NA`) are correctly represented on R side before serialization.

falaki:
I just tried the patch on R version 3.3.1 (2016-06-21) -- "Bug in Your Hair" on Linux and it passed tests. 

@HyukjinKwon how can kick another AppVeyor test?

HyukjinKwon:
Build started: [SparkR] `ALL` [PR-15421](https://ci.appveyor.com/project/spark-test/spark/branch/82AF3188-11E3-40D8-A71C-7806413EDAE3)

Ah, I wrote the scripts locally (separately with Spark's AppVeyor). I can't access to the Apache's AppVeyor so I made another account. I wrote some documnetations in https://github.com/HyukjinKwon/spark-appveyor to show what the scripts actually doing :)

wangmiao1981:
I modified the code on `readString`. When len is negative, I read 3 bytes. `NA` is read on scala side, which is correct field value. But the integer of len is not serialized. It seems that R side serialization doesn't add length for `NA`. I am debugging R side now.

falaki:
@wangmiao1981 you are talking about Windows right?

wangmiao1981:
No. Mac. I can always reproduce this issue on Mac. I modified code above on scala side to make sure that `NA` is correctly serialized on scala side. But the integer (i.e., length of `NA`) before `NA` is not serialized.

wangmiao1981:
I trace back to `createDataFrame.default` 
...
  jrdd <- getJRDD(lapply(rdd, function(x) x), "row")
  srdd <- callJMethod(jrdd, "rdd")
  sdf <- callJStatic("org.apache.spark.sql.api.r.SQLUtils", "createDF",
                     srdd, schema$jobj, sparkSession)
  dataFrame(sdf)

The `srdd` object should be the one used in `createDF`, which caused `def readString(in: DataInputStream): String ` negative index issue.

I checked `data` in `createDataFrame.default`, which is correct on R side. 

I get lost in the serialization part.

wangmiao1981:
r

It seems that R doesn't handle `NA` well. I suspect that the root cause is R.

felixcheung:
@wangmiao1981 do you mean for this?
```
> a <- as.Date("2016-10-11", "NA")
> a
[1] NA
> a <- as.Date(c("2016-10-11", "NA"))
> a
[1] "2016-10-11" NA
```
in the 2nd case you do have a vector of Date and NA

falaki:
@felixcheung and @shivaram AppVeyor test that @HyukjinKwon kicked passed and jenkins passed too. Do you think this is ready?

wangmiao1981:
@felixcheung Yes. I think I missed `vector` in the above example. I am building on clean checkout and test again. Thanks!

wangmiao1981:
@felixcheung I rebuild clean image and use the test case on JIRA. Still fails. 
> df <- data.frame(Date = as.Date(c(rep("2016-01-10", 10), "NA", "NA")), id = 1:12)
> 
> dim(createDataFrame(df))
16/10/14 14:28:51 ERROR Executor: Exception in task 0.0 in stage 1.0 (TID 1)
java.lang.OutOfMemoryError: Java heap space
	at org.apache.spark.api.r.SerDe$.readStringBytes(SerDe.scala:110)
	at org.apache.spark.api.r.SerDe$.readString(SerDe.scala:119)
	at org.apache.spark.api.r.SerDe$.readDate(SerDe.scala:133)
	at org.apache.spark.api.r.SerDe$.readTypedObject(SerDe.scala:77)
	at org.apache.spark.api.r.SerDe$.readObject(SerDe.scala:61)
	at org.apache.spark.sql.api.r.SQLUtils$$anonfun$bytesToRow$1.apply(SQLUtils.scala:161)
	at org.apache.spark.sql.api.r.SQLUtils$$anonfun$bytesToRow$1.apply(SQLUtils.scala:160)

Since automated tests passed, it could be related to my specific system. You can ignore the issue now. I will follow up if I find any issues. Thanks!

yhuai:
test this please

falaki:
@wandjenkins yes, I think there must be some non-standard issue on your system. I tested on Mac and Linux with different version of R and passed the test case.

wangmiao1981:
Distributor ID:	Ubuntu
Description:	Ubuntu 16.04.1 LTS
Release:	16.04
Codename:	xenial

R version 3.3.0 beta (2016-03-30 r70404) -- "Supposedly Educational"
Copyright (C) 2016 The R Foundation for Statistical Computing
Platform: x86_64-pc-linux-gnu (64-bit)

I test it on above system. It passed. 

R version 3.3.0 (2016-05-03) -- "Supposedly Educational"
Copyright (C) 2016 The R Foundation for Statistical Computing
Platform: x86_64-apple-darwin13.4.0 (64-bit)

With this version, it still fails.

It should be a R specific version issue.

falaki:
@wandjenkins thanks! It is interesting with R  3.3.1 it worked!

wangmiao1981:
@falaki As I found, for the failure cases, the serialization doesn't serialize length of `NA` and only serialize `NA` as characters (String). I just don't get the point how it is exactly serialized, because it is not serialized by field and field type. Therefore, logic like writeDate, writeTime doesn't kick in. It is good to learn how to debug serialization and R backend hook. I will keep tracking this issue at spare time. Thanks for spending time reading this.

falaki:
@shivaram and @felixcheung do I need to do more on this?

shivaram:
@falaki @wangmiao1981 Can we summarize the discussion around the `NegativeArraySizeException` ?

wangmiao1981:
@shivaram We found that the negative index error happens in some version of R. For example, on my mac, R version 3.3.0 (2016-05-03) and the previous Windows test 3.3.2. For the failed cases, I put debug message and tried to read the byte array. For the field `NA`, the failed case doesn't serialize the length as integer (i.e., the byte array only includes `NA` but no the length `3`). Therefore, when readString reads the byte array in the order of length and string, it actually reads `NA` as an integer, which is negative. 

When creating Dateframe, it serializes the data.frame as a `jobj`. I checked for both good and bad cases, but I didn't find any differences between the two. I didn't find a way to debug the `jobj`  serialization logic as it just writes binary in batch. Maybe, I can try again to dump the binary stream. On the surface, I didn't find the reason why length is not serialized for the failed case.

shivaram:
Thanks @wangmiao1981 - There are two different kinds of serializations that happen in SparkR - one is the RPC style serialization where function arguments are serialized using `writeDate`, `writeInt` etc. The other is batch or bulk serialization that we use in case of converting R `data.frame` to Spark RDDs. This is used in the `createDataFrame` case from [1].

Now the way this is supposed to work is that this is converted by the call to `lapply` and `getJRDD` [2] to be a row-wise serialized `SparkDataFrame`. To do this on the executor side you will have a `unserialize` called on the bulk data  [3] and a `writeRowSerialize` called for each row [4]. So the final byte stream to look at is the one here. But my guess is that things are going wrong somewhere before this -- i.e. the byte stream at [3] for example has some different type or something like that. Or to put it another way, are we sure `writeString` was called with `NA` or was it some other function like `writeBin` because the types were wrong ?

The other reason for such a transient bug might be that the channels are not getting flushed somewhere and this doesn't show up on some R versions. But yeah your debugging methods are in line with what I would try

[1] https://github.com/apache/spark/blob/d88a1bae6a9c975c39549ec2326d839ea93949b2/R/pkg/R/context.R#L140
[2] https://github.com/apache/spark/blob/d88a1bae6a9c975c39549ec2326d839ea93949b2/R/pkg/R/SQLContext.R#L275
[3] https://github.com/apache/spark/blob/d88a1bae6a9c975c39549ec2326d839ea93949b2/R/pkg/inst/worker/worker.R#L159
[4] https://github.com/apache/spark/blob/d88a1bae6a9c975c39549ec2326d839ea93949b2/R/pkg/inst/worker/worker.R#L78

wangmiao1981:
@shivaram  Thanks for your explanation! I can continue debugging this as you pointed and I can constantly reproduce the issue. For this PR, I think it is fine for handling the `NA` in the backend except for the unnecessary exception handling. I can submit a follow up PR on the serialization part.

shivaram:
Cool - @falaki lets see if @wangmiao1981 debugging finds anything in the next day or so ? The only thing that would be good is to get this in for 2.0.2 cut if that is happening soon. 

I'll coordinate that with @rxin

wangmiao1981:
By putting debug messages to 
[3] https://github.com/apache/spark/blob/d88a1bae6a9c975c39549ec2326d839ea93949b2/R/pkg/inst/worker/worker.R#L159
[4] https://github.com/apache/spark/blob/d88a1bae6a9c975c39549ec2326d839ea93949b2/R/pkg/inst/worker/worker.R#L78
Neither is called.

I also put debug message on
[1] https://github.com/apache/spark/blob/d88a1bae6a9c975c39549ec2326d839ea93949b2/R/pkg/R/context.R#L140

For 
> a <- as.Date(c("2016-11-11", "NA"))
> b <- as.data.frame(a)
> c <- createDataFrame(b
> a
[1] "2016-11-11" NA          
> b
           a
1 2016-11-11
2       <NA>

`slices` that is in lapply is:
slice list(17116) slice list(NA)

`"2016-11-11"` becomes `17116` and `"NA"` is `NA`.

shivaram:
Thanks - the lines in [3], [4] will be called if we do any operation on the DataFrame. i.e. something like `dim(c)`. Also can we use the same test case that is in the test file checked in ?
```
 df <- data.frame(id = 1:2,
   date = c(as.Date("2016-10-01"), NA))
DF <- collect(createDataFrame(df))
is.na(DF$date[2]) # should be TRUE
```

wangmiao1981:
I think the reason is because of the code below:

`> serialize
function (object, connection, ascii = FALSE, xdr = TRUE, version = NULL, 
    refhook = NULL) 
{
    if (!is.null(connection)) {
        if (!inherits(connection, "connection")) 
            stop("'connection' must be a connection")
        if (missing(ascii)) 
            ascii <- summary(connection)$text == "text"
    }
    if (!ascii && inherits(connection, "sockconn")) 
        .Internal(serializeb(object, connection, xdr, version, 
            refhook))
    else {
        type <- if (is.na(ascii)) 
            2L
        else if (ascii) 
            1L
        else if (!xdr) 
            3L
        else 0L
        .Internal(serialize(object, connection, type, version, 
            refhook))
    }
}
`
` is.na(list(NA))`
`[1] TRUE`
` is.na(list(17116))`
[1] FALSE

So, `"2016-11-11"` and `NA` are serialized as different types (i.e., `NA` is not serialized with my R version).

wangmiao1981:
@shivaram I think we can use that test case. Somehow, I missed the debug message of [3] and [4], but it should not be quite related. The reason should be my `serialize` function, as shown above.

shivaram:
did this change in a recent R version  ? I'm not sure why `NA` is not being serialized ? That `if` statement should only affect the value assigned to `type` right ?

wangmiao1981:
As discussed previously, R 3.3.1 works. For 3.3.0, `NA` is serialized but it is not serialized as `String`.

shivaram:
Sorry my question wasn't clear - Is there a source change that we can spot that might have caused this behavior ? I don't see this line changing recently looking at history of https://github.com/wch/r-source/blob/trunk/src/library/base/R/serialize.R

wangmiao1981:
You are right. There is no change for the two versions.

wangmiao1981:
Update R on another linux:
R version 3.3.1 (2016-06-21) -- "Bug in Your Hair"
Copyright (C) 2016 The R Foundation for Statistical Computing
Platform: x86_64-pc-linux-gnu (64-bit)

Distributor ID:	Ubuntu
Description:	Ubuntu 16.04.1 LTS
Release:	16.04
Codename:	xenial

I still see the negative index failure. It seem very consistent on my setups: 2 out of 3 setups fail; 1  out of 3 pass.

falaki:
@wangmiao1981 would you please also test the master branch?

wangmiao1981:
Master Branch:

16/10/18 16:10:15 ERROR Executor: Exception in task 0.0 in stage 1.0 (TID 1)
java.lang.NegativeArraySizeException
	at org.apache.spark.api.r.SerDe$.readStringBytes(SerDe.scala:110)
	at org.apache.spark.api.r.SerDe$.readString(SerDe.scala:119)
	at org.apache.spark.api.r.SerDe$.readDate(SerDe.scala:128)
	at org.apache.spark.api.r.SerDe$.readTypedObject(SerDe.scala:77)
	at org.apache.spark.api.r.SerDe$.readObject(SerDe.scala:61)
	at org.apache.spark.sql.api.r.SQLUtils$$anonfun$bytesToRow$1.apply(SQLUtils.scala:161)
	at org.apache.spark.sql.api.r.SQLUtils$$anonfun$bytesToRow$1.apply(SQLUtils.scala:160)
	at scala.collection.immutable.Range.foreach(Range.scala:160)
	at scala.collection.AbstractTraversable.map(Traversable.scala:104)
	at org.apache.spark.sql.api.r.SQLUtils$.bytesToRow(SQLUtils.scala:160)
	at org.apache.spark.sql.api.r.SQLUtils$$anonfun$5.apply(SQLUtils.scala:138)
	at org.apache.spark.sql.api.r.SQLUtils$$anonfun$5.apply(SQLUtils.scala:138)
	at scala.collection.Iterator$$anon$11.next(Iterator.scala:409)
	at scala.collection.Iterator$$anon$11.next(Iterator.scala:409)
	at scala.collection.Iterator$$anon$11.next(Iterator.scala:409)
	at org.apache.spark.sql.catalyst.expressions.GeneratedClass$GeneratedIterator.agg_doAggregateWithoutKey$(Unknown Source)
	at org.apache.spark.sql.execution.BufferedRowIterator.hasNext(BufferedRowIterator.java:43)
	at org.apache.spark.sql.execution.WholeStageCodegenExec$$anonfun$8$$anon$1.hasNext(WholeStageCodegenExec.scala:372)
	at scala.collection.Iterator$$anon$11.hasNext(Iterator.scala:408)
	at org.apache.spark.shuffle.sort.BypassMergeSortShuffleWriter.write(BypassMergeSortShuffleWriter.java:126)
	at org.apache.spark.scheduler.ShuffleMapTask.runTask(ShuffleMapTask.scala:96)
	at org.apache.spark.scheduler.ShuffleMapTask.runTask(ShuffleMapTask.scala:53)
	at org.apache.spark.scheduler.Task.run(Task.scala:99)

falaki:
Thanks @wangmiao1981. I thin it is best to file a separate JIRA for this issue. Thanks a lot for catching it.

shivaram:
Thats a good idea. @wangmiao1981 Can you create a new JIRA for this and @falaki can we add that JIRA as a pointer in the comment close to the `NegativeArraySizeException` ? Otherwise code change looks fine to me.

@felixcheung Any other comments ?

falaki:
@shivaram I removed catching `NegativeArraySizeException` from this PR.

core/src/main/scala/org/apache/spark/api/r/SerDe.scala
```diff
@@ -125,15 +125,34 @@ private[spark] object SerDe {
...
+        null
+      } else {
+        val sec = Math.floor(seconds).toLong
+        val t = new Timestamp(sec * 1000L)
+        t.setNanos(((seconds - sec) * 1e9).toInt)
+        t
+      }
+    } catch {
+      // On windows we get NegativeArraySizeException for NAs in R
+      case _: NegativeArraySizeException => null
```
shivaram: @falaki the catch is still in the diff ?

falaki: My bad. It was not pushed. Let's see how tests go now.

wangmiao1981:
@shivaram I will create a JIRA later today. Thanks!

wangmiao1981:
New JIRA is created: https://issues.apache.org/jira/browse/SPARK-18011

I add one comment in the JIRA to mention discussions on this PR.

I think I am close to the root cause. I will spend more time to follow up.

wangmiao1981:
appveyor also failed. Error messages below:

java.lang.NegativeArraySizeException
	at org.apache.spark.api.r.SerDe$.readStringBytes(SerDe.scala:110)
	at org.apache.spark.api.r.SerDe$.readString(SerDe.scala:119)
	at org.apache.spark.api.r.SerDe$.readDate(SerDe.scala:128)
	at org.apache.spark.api.r.SerDe$.readTypedObject(SerDe.scala:77)
	at org.apache.spark.api.r.SerDe$.readObject(SerDe.scala:61)
	at org.apache.spark.sql.api.r.SQLUtils$$anonfun$bytesToRow$1.apply(SQLUtils.scala:161)
	at org.apache.spark.sql.api.r.SQLUtils$$anonfun$bytesToRow$1.apply(SQLUtils.scala:160)
	at scala.collection.immutable.Range.foreach(Range.scala:160)
	at scala.collection.AbstractTraversable.map(Traversable.scala:104)
	at org.apache.spark.sql.api.r.SQLUtils$.bytesToRow(SQLUtils.scala:160)
	at org.apache.spark.sql.api.r.SQLUtils$$anonfun$5.apply(SQLUtils.scala:138)
	at org.apache.spark.sql.api.r.SQLUtils$$anonfun$5.apply(SQLUtils.scala:138)
	at scala.collection.Iterator$$anon$11.next(Iterator.scala:409)
	at scala.collection.Iterator$$anon$11.next(Iterator.scala:409)
	at scala.collection.Iterator$$anon$11.next(Iterator.scala:409)
	at org.apache.spark.sql.execution.SparkPlan$$anonfun$2.apply(SparkPlan.scala:232)
	at org.apache.spark.sql.execution.SparkPlan$$anonfun$2.apply(SparkPlan.scala:225)
	at org.apache.spark.rdd.RDD$$anonfun$mapPartitionsInternal$1$$anonfun$apply$24.apply(RDD.scala:803)
	at org.apache.spark.rdd.RDD$$anonfun$mapPartitionsInternal$1$$anonfun$apply$24.apply(RDD.scala:803)
	at org.apache.spark.rdd.MapPartitionsRDD.compute(MapPartitionsRDD.scala:38)
	at org.apache.spark.rdd.RDD.computeOrReadCheckpoint(RDD.scala:319)
	at org.apache.spark.rdd.RDD.iterator(RDD.scala:283)
	at org.apache.spark.scheduler.Task.run(Task.scala:99)
16/10/19 17:54:34 ERROR TaskSetManager: Task 0 in stage 108.0 failed 1 times; aborting job
16/10/19 17:54:34 ERROR RBackendHandler: dfToCols on org.apache.spark.sql.api.r.SQLUtils failed
java.lang.reflect.InvocationTargetException
	at sun.reflect.GeneratedMethodAccessor119.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.apache.spark.api.r.RBackendHandler.handleMethodCall(RBackendHandler.scala:141)
	at org.apache.spark.api.r.RBackendHandler.channelRead0(RBackendHandler.scala:86)
	at org.apache.spark.api.r.RBackendHandler.channelRead0(RBackendHandler.scala:38)
	at io.netty.channel.SimpleChannelInboundHandler.channelRead(SimpleChannelInboundHandler.java:105)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:366)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:352)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:345)
	at io.netty.handler.codec.MessageToMessageDecoder.channelRead(MessageToMessageDecoder.java:102)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:366)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:352)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:345)
	at io.netty.handler.codec.ByteToMessageDecoder.fireChannelRead(ByteToMessageDecoder.java:293)
	at io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:267)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:366)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:352)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:345)
	at io.netty.channel.DefaultChannelPipeline$HeadContext.channelRead(DefaultChannelPipeline.java:1294)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:366)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:352)
	at io.netty.channel.DefaultChannelPipeline.fireChannelRead(DefaultChannelPipeline.java:911)
	at io.netty.channel.nio.AbstractNioByteChannel$NioByteUnsafe.read(AbstractNioByteChannel.java:131)
	at io.netty.channel.nio.NioEventLoop.processSelectedKey(NioEventLoop.java:611)
	at io.netty.channel.nio.NioEventLoop.processSelectedKeysOptimized(NioEventLoop.java:552)
	at io.netty.channel.nio.NioEventLoop.processSelectedKeys(NioEventLoop.java:466)
	at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:438)
	at io.netty.util.concurrent.SingleThreadEventExecutor$2.run(SingleThreadEventExecutor.java:140)
	at io.netty.util.concurrent.DefaultThreadFactory$DefaultRunnableDecorator.run(DefaultThreadFactory.java:144)

wangmiao1981:
It seems that this test case will break windows automation test. We might need to resolve the new JIRA.

shivaram:
@wangmiao1981 is right - i think we can put back the `catch` in there and add a `TODO` referring to the JIRA. Alternately I'll try to reproduce this on a fresh Ubuntu 16.04 VM later tonight and see if I can contribute to the debugging

wangmiao1981:
I print out binary of serialized data before sending and the bytes sent by R side. They are the same. I also called `unserialize` after serialization on R side. It deserializes successfully. 

So `list(NA)` should have a special handling for serialization on R side. From R source code, I didn't get the point.

felixcheung:
Is it possible to have a test to trigger this case?

falaki:
@felixcheung you mean test for parallelizing NAs and getting them back? The patch includes that test.

falaki:
@wangmiao1981 could that behavior with `list` relate to the issue in this ticket? https://issues.apache.org/jira/browse/SPARK-17781

felixcheung:
@falaki I mean for the case the exception is raised, but it sounds it is the same test just on different R version
LGTM.

wangmiao1981:
@falaki I think it is the same issue as JIRA 17781. Your change looks good to me.

falaki:
Ping.

felixcheung:
merged to master and branch-2.0

PR #15422, opened 2016-10-10
[SPARK-17850][Core]Add a flag to ignore corrupt files

zsxwing:

Add a flag to ignore corrupt files. For Spark core, the configuration is `spark.files.ignoreCorruptFiles`. For Spark SQL, it's `spark.sql.files.ignoreCorruptFiles`.
## How was this patch tested?

The added unit tests

### DIFF
--- a/core/src/main/scala/org/apache/spark/internal/config/package.scala
+++ b/core/src/main/scala/org/apache/spark/internal/config/package.scala
@@ -170,4 +170,9 @@ package object config {
     .doc("Port to use for the block managed on the driver.")
     .fallbackConf(BLOCK_MANAGER_PORT)
 
+  private[spark] val IGNORE_CORRUPT_FILES = ConfigBuilder("spark.files.ignoreCorruptFiles")
+    .doc("Whether to ignore corrupt files. If true, the Spark jobs will continue to run when " +
+      "encountering corrupt files and contents that have been read will still be returned.")
+    .booleanConf
+    .createWithDefault(false)
 }
--- a/core/src/main/scala/org/apache/spark/rdd/NewHadoopRDD.scala
+++ b/core/src/main/scala/org/apache/spark/rdd/NewHadoopRDD.scala
@@ -186,12 +186,7 @@ class NewHadoopRDD[K, V](
           try {
             finished = !reader.nextKeyValue
           } catch {
-            case e: IOException =>
-              if (ignoreCorruptFiles) {
-                finished = true
-              } else {
-                throw e
-              }
+            case e: IOException if ignoreCorruptFiles => finished = true
           }
           if (finished) {
             // Close and release the reader here; close() will also be called when the task
--- a/core/src/main/scala/org/apache/spark/rdd/NewHadoopRDD.scala
+++ b/core/src/main/scala/org/apache/spark/rdd/NewHadoopRDD.scala
@@ -17,6 +17,7 @@
 
 package org.apache.spark.rdd
 
+import java.io.IOException
 import java.text.SimpleDateFormat
 import java.util.Date
 
@@ -33,6 +34,7 @@ import org.apache.spark._
 import org.apache.spark.annotation.DeveloperApi
 import org.apache.spark.deploy.SparkHadoopUtil
 import org.apache.spark.internal.Logging
+import org.apache.spark.internal.config.IGNORE_CORRUPT_FILES
 import org.apache.spark.rdd.NewHadoopRDD.NewHadoopMapPartitionsWithSplitRDD
 import org.apache.spark.storage.StorageLevel
 import org.apache.spark.util.{SerializableConfiguration, ShutdownHookManager}
@@ -84,6 +86,8 @@ class NewHadoopRDD[K, V](
   @transient protected val jobId = new JobID(jobTrackerId, id)
 
   private val shouldCloneJobConf = sparkContext.conf.getBoolean("spark.hadoop.cloneConf", false)
+
+  private val ignoreCorruptFiles = sparkContext.conf.get(IGNORE_CORRUPT_FILES)
 
   def getConf: Configuration = {
     val conf: Configuration = confBroadcast.value.value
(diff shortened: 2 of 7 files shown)


zsxwing:
retest this please

mridulm:
Why would corrupt record cause EOFException to be thrown ?

srowen:
If this happens it isn't clear that anything that was read is valid. It doesn't seem like something to ignore. Log, at least. I know people differ on this but I think continuing with partial and maybe corrupt read even with a warning seems more likely to cause tears.

mridulm:
@srowen The tuples already returned would have been valid, it is the subsequent block decompression which has failed. For example, in a 1gb file, the last few bytes missing (or corrupt) will cause the last block to be decompressed incorrectly - but all previous tuples already returned would be fine and valid.

The 'current' key/value resulted in exception.
Note that the returned value from the method is not used when finished = true (what is returned is ignored).

marmbrus:
I agree that the data that was already read is probably good.  I also think that this is a pretty big behavior change where there are legitimate cases (i.e. tons of data and it is fine to miss some) where you'd only want a warning.  Can we add a flag for failing on unexpected EOF?  (probably set to `true` in `master` and `false` in `branch-1.6`).

zsxwing:
@mridulm This fix just makes HadoopRDD consistent with NewHadoopRDD and the current behavior of Spark SQL in 2.0.

For 1.6, that's another story since Spark SQL uses HadoopRDD directly.

mridulm:
@zsxwing You are right, NewHadoopRDD is not handling this case.
Probably would be good to add exception handling there when nextKeyValue throws exception ?

Context is, for large jobs/data, it is not unexpected to see some data corruption at times. We dont want to throw out the entire job due to a few bad records.
For example, in MR you have the ability to even set the percentage of bad records you want to tolerate (we dont have that in spark).

mridulm:
@marmbrus +1 on logging, that is definitely something which was probably missed here.

srowen:
@mridulm for the scenario you're imagining, maybe the data is OK, sure. That doesn't mean it's true in all cases. Yeah, this is really to work around bad input, which you can to some degree do at the user level. Other parts of Spark don't work this way. I'm neutral on whether this is a good idea at all, but would prefer consistency more than anything.

mridulm:
@zsxwing The map task is run by https://github.com/apache/hadoop/blob/trunk/hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-core/src/main/java/org/apache/hadoop/mapred/MapTask.java , no ?

You can take a look at skipping bad records here : https://hadoop.apache.org/docs/r1.2.1/mapred_tutorial.html#Skipping+Bad+Records

mridulm:
@srowen Since this is happening 'below' the user code (in the hadoop rdd), is there a way around how to handle this ?
I agree that for a lot of usecases where it is critical to work off the entire data, we should abort rather than process incomplete (and corrupt) data; not sure if we want to customize ability to do this though.

EOFException being thrown instead of IOException is an implementation detail of the codec if I am not wrong (I dont think the contract specifies this) - @zsxwing can confirm though, I am not very familiar with that.

setting finished = true should probably be done not just for EOFException, but also for IOException.

I am fine with that as well, though this is a behavior change.

sql/core/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala
```diff
@@ -588,6 +588,12 @@ object SQLConf {
       .doubleConf
       .createWithDefault(0.05)
 
+  val IGNORE_CORRUPT_FILES = SQLConfigBuilder("spark.sql.files.ignoreCorruptFiles")
+    .doc("Whether to ignore corrupt files. If true, the Spark jobs will continue to run when " +
+      "encountering corrupt files and contents that have been read will still be returned.")
+    .booleanConf
+    .createWithDefault(false)
+
```
mridulm: Curious why we are duplicating the parameter in sql namespace. Wont spark.files.ignoreCorruptFiles not do ?

core/src/main/scala/org/apache/spark/rdd/NewHadoopRDD.scala
```diff
@@ -179,7 +183,16 @@ class NewHadoopRDD[K, V](
...
+          try {
+            finished = !reader.nextKeyValue
+          } catch {
+            case e: IOException =>
+              if (ignoreCorruptFiles) {
+                finished = true
+              } else {
+                throw e
+              }
+          }
```
mridulm: Thanks for changing this too !

core/src/main/scala/org/apache/spark/internal/config/package.scala
```diff
@@ -170,4 +170,9 @@ package object config {
     .doc("Port to use for the block managed on the driver.")
     .fallbackConf(BLOCK_MANAGER_PORT)
 
+  private[spark] val IGNORE_CORRUPT_FILES = ConfigBuilder("spark.files.ignoreCorruptFiles")
+    .doc("Whether to ignore corrupt files. If true, the Spark jobs will continue to run when " +
+      "encountering corrupt files and contents that have been read will still be returned.")
+    .booleanConf
+    .createWithDefault(false)
```
mridulm: So either way we will have a behavioral change - if NewHadoopRDD vs HadoopRDD.
IMO that is fine, given that we are standardizing on the behavior and this is something which was a corner case anyway.

Setting default to false makes sense.

mridulm (approved):

sql/core/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala
zsxwing: A sql conf can appear in the following command:
```
sql("set -v").filter('key contains "files").show(truncate = false)
```

mridulm:
Merged - had issue with pip (new laptop, sigh), and so jira and pr did not get closed.

zsxwing:
I will work on a patch for 1.6.

zsxwing:
PR for 1.6: #15454

rxin:
@zsxwing shouldn't we at least log the exception?

PR #15423, opened 2016-10-10
[SPARK-17860][SQL] SHOW COLUMN's database conflict check should respect case sensitivity configuration

dilipbiswal:

SHOW COLUMNS command validates the user supplied database 
name with database name from qualified table name name to make
sure both of them are consistent. This comparison should respect
case sensitivity.
## How was this patch tested?

Added tests in DDLSuite and existing tests were moved to use new sql based test infrastructure.


sql/core/src/test/scala/org/apache/spark/sql/SQLQueryTestSuite.scala
```diff
@@ -207,6 +208,7 @@ class SQLQueryTestSuite extends QueryTest with SharedSQLContext {
     // Returns true if the plan is supposed to be sorted.
     def isSorted(plan: LogicalPlan): Boolean = plan match {
       case _: Join | _: Aggregate | _: Generate | _: Sample | _: Distinct => false
+      case _: ShowColumnsCommand => true
```
viirya: Why `ShowColumnsCommand` is sorted?

dilipbiswal: @viirya I added the case here because of
https://github.com/dilipbiswal/spark/blob/3acd08f9431d6cdfe4d043aa342d09fc0ebfa497/sql/core/src/test/scala/org/apache/spark/sql/SQLQueryTestSuite.scala#L222

I didn't want the output of ShowColumnsCommand to be sorted before comparison.

viirya: Will it affect comparison result? I think the result is generated, right?

dilipbiswal: @viirya So it seemed odd to have the generated output files to have column names sorted which didn't reflect the columns. In the test case i had the table create like following.
```SQL
CREATE TABLE showcolumn2 (price int, qty int) partitioned by (year int, month int)
```
It seemed odd to me to have the generated output file report the columns as month, price, qty and year as opposed to price, qty, year and month.

viirya: Personally I don't think it is odd because we just want to compare the results. Adding `ShowColumnsCommand` to sorted op looks more odd to me. cc @cloud-fan

sql/core/src/test/scala/org/apache/spark/sql/execution/command/DDLSuite.scala
```diff
@@ -1713,4 +1713,19 @@ class DDLSuite extends QueryTest with SharedSQLContext with BeforeAndAfterEach {
...
     }
   }
+
+  test("show columns - negative test") {
+    // When case sensitivity is true, the user supplied database name in table identifier
+    // should match the supplied database name in case sensitive way.
+    withSQLConf(SQLConf.CASE_SENSITIVE.key -> "true") {
+      val tabName = "showcolumn"
+      withTable(tabName) {
+        sql(s"CREATE TABLE $tabName(col1 int, col2 string) USING parquet ")
```
viirya: Do we need to explicitly set the current database here? If the current database is not `default`, the test is not making sense anymore.

dilipbiswal: @viirya OK.. I agree. I will make the change

sql/core/src/main/scala/org/apache/spark/sql/execution/SparkSqlParser.scala
```diff
@@ -168,17 +168,7 @@ class SparkSqlAstBuilder(conf: SQLConf) extends AstBuilder {
...
-    val lookupTable = Option(ctx.db) match {
-      case None => table
-      case Some(db) if table.database.exists(_ != db) =>
-        operationNotAllowed(
-          s"SHOW COLUMNS with conflicting databases: '$db' != '${table.database.get}'",
-          ctx)
-      case Some(db) => TableIdentifier(table.identifier, Some(db.getText))
-    }
-    ShowColumnsCommand(lookupTable)
+    ShowColumnsCommand(Option(ctx.db).map(_.getText), visitTableIdentifier(ctx.tableIdentifier))
```
cloud-fan: FYI, MySQL will treat `SHOW COLUMNS FROM db1.tbl1 FROM db2` as `SHOW COLUMNS FROM tbl1 FROM db2`, i.e. if `FROM database` is specified, it will just ignore the database specified in table name, instead of reporting error.

cloud-fan: we should investigate more databases to see how they handle this case.

viirya: Good  point!

viirya: BTW, what @dilipbiswal does in this is following previous behavior, do we want to break it?

viirya: Seems Hive doesn't allow specifying duplicate databases no matter they are the same or not.

https://github.com/apache/hive/blob/21a0142f333fba231f2648db53a48dc41384ad72/ql/src/java/org/apache/hadoop/hive/ql/parse/DDLSemanticAnalyzer.java#L2215

sql/core/src/test/scala/org/apache/spark/sql/SQLQueryTestSuite.scala
cloud-fan: Does it break test if we don't mark it as sorted?

dilipbiswal: @cloud-fan @viirya Actually it does not break the test if we don't mark it as sorted. What happens is that, when we generate the expected output file, the results appear sorted like following:

```SQL
-- !query 7
SHOW COLUMNS IN showcolumn2 IN showdb
-- !query 7 schema
struct<col_name:string>
-- !query 7 output
month
price
qty
year
```
When i was going through the expected output file to make sure its correct, i noticed this as the above output would not be how it would be shown if i cut-paste the SQLs snippets from the test file and ran it in spark-sql shell. 

If you guys think its okay to have the output in sorted form in the expected file, then i will change it back.

sql/core/src/main/scala/org/apache/spark/sql/execution/SparkSqlParser.scala
dilipbiswal: Thanks @viirya for checking hive. Yeah, when we implemented the SHOW columns native command, wde went through the the above hive code. We decided to improve up on the above check and report an error only when the database names are different.

sql/core/src/test/scala/org/apache/spark/sql/SQLQueryTestSuite.scala
cloud-fan: marking `ShowColumnsCommand` as sorted is more weird, I'd like to leave the result sorted.

viirya: +1 as mentioned in previous comment.

dilipbiswal: @cloud-fan @viirya Thanks :-) I will change it.

dilipbiswal:
@viirya @cloud-fan I have incorporated the review comments. Could we please look at this again ?

sql/core/src/main/scala/org/apache/spark/sql/execution/SparkSqlParser.scala
viirya: I have no strong option towards this. From my point, MySQL's way might be little confusing users if they don't notice the database name is different.

viirya:
LGTM, see if @cloud-fan has more comments on this or not?

sql/core/src/main/scala/org/apache/spark/sql/execution/command/tables.scala
```diff
@@ -651,14 +651,28 @@ case class ShowTablePropertiesCommand(table: TableIdentifier, propertyKey: Optio
...
   }
 
+  private def nameEqual(name1: String, name2: String, caseSensitive: Boolean): Boolean = {
+    if (caseSensitive) name1 == name2 else name1.equalsIgnoreCase(name2)
+  }
+
   override def run(sparkSession: SparkSession): Seq[Row] = {
     val catalog = sparkSession.sessionState.catalog
-    val table = catalog.getTempViewOrPermanentTableMetadata(tableName)
+    val caseSensitive = sparkSession.sessionState.conf.caseSensitiveAnalysis
```
cloud-fan: nit: we can simplify it to
```
val resolver = sparkSession.sessionState.conf.resolver
...
case Some(db) if tableName.database.exists(!resolver(_, db))
```

sql/core/src/test/resources/sql-tests/inputs/show_columns.sql
```diff
@@ -0,0 +1,35 @@
+CREATE DATABASE showdb;
+
+USE showdb;
+
+CREATE TABLE showcolumn1 (col1 int, `col 2` int);
+CREATE TABLE showcolumn2 (price int, qty int) partitioned by (year int, month int);
```
cloud-fan: can we also test temp view?

dilipbiswal:
@cloud-fan Hi wenchen, i have added the test cases for temp view. Could we please look at this again? Thanks !

sql/core/src/test/scala/org/apache/spark/sql/execution/command/DDLSuite.scala
```diff
@@ -1713,4 +1713,21 @@ class DDLSuite extends QueryTest with SharedSQLContext with BeforeAndAfterEach {
...
+  test("show columns - negative test") {
+    // When case sensitivity is true, the user supplied database name in table identifier
+    // should match the supplied database name in case sensitive way.
+    withSQLConf(SQLConf.CASE_SENSITIVE.key -> "true") {
+      withTempDatabase { db =>
+        val tabName = s"$db.showcolumn"
+        withTable(tabName) {
+          sql(s"CREATE TABLE $tabName(col1 int, col2 string) USING parquet ")
+          val message = intercept[AnalysisException] {
+          sql(s"SHOW COLUMNS IN $db.showcolumn FROM ${db.toUpperCase}")
```
cloud-fan: nit: wrong ident here.

cloud-fan:
LGTM

viirya:
The tests are passed but the results are failed to post back to github...

viirya:
@cloud-fan Need to run tests again?

cloud-fan:
it's fine, merging to master!

dilipbiswal:
@cloud-fan @viirya Thank you very much !!

PR #14897, opened 2016-08-31
[SPARK-17338][SQL] add global temp view

cloud-fan:
## What changes were proposed in this pull request?

Global temporary view is a cross-session temporary view, which means it's shared among all sessions. Its lifetime is the lifetime of the Spark application, i.e. it will be automatically dropped when the application terminates. It's tied to a system preserved database `global_temp`(configurable via SparkConf), and we must use the qualified name to refer a global temp view, e.g. SELECT \* FROM global_temp.view1.

changes for `SessionCatalog`:
1. add a new field `gloabalTempViews: GlobalTempViewManager`, to access the shared global temp views, and the global temp db name.
2. `createDatabase` will fail if users wanna create `global_temp`, which is system preserved.
3. `setCurrentDatabase` will fail if users wanna set `global_temp`, which is system preserved.
4. add `createGlobalTempView`, which is used in `CreateViewCommand` to create global temp views.
5. add `dropGlobalTempView`, which is used in `CatalogImpl` to drop global temp view.
6. add `alterTempViewDefinition`, which is used in `AlterViewAsCommand` to update the view definition for local/global temp views.
7. `renameTable`/`dropTable`/`isTemporaryTable`/`lookupRelation`/`getTempViewOrPermanentTableMetadata`/`refreshTable` will handle global temp views.

changes for SQL commands:
1. `CreateViewCommand`/`AlterViewAsCommand` is updated to support global temp views
2. `ShowTablesCommand` outputs a new column `database`, which is used to distinguish global and local temp views.
3. other commands can also handle global temp views if they call `SessionCatalog` APIs which accepts global temp views, e.g. `DropTableCommand`, `AlterTableRenameCommand`, `ShowColumnsCommand`, etc.

changes for other public API
1. add a new method `dropGlobalTempView` in `Catalog`
2. `Catalog.findTable` can find global temp view
3. add a new method `createGlobalTempView` in `Dataset`
## How was this patch tested?

new tests in `SQLViewSuite`

PR #15424, opened 2016-10-11
[SPARK-17338][SQL][follow-up] add global temp view

cloud-fan:

address post hoc review comments for https://github.com/apache/spark/pull/14897
## How was this patch tested?

N/A

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/catalog/Catalog.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/catalog/Catalog.scala
@@ -270,9 +270,10 @@ abstract class Catalog {
    * tied to any databases, i.e. we can't use `db1.view1` to reference a local temporary view.
    *
    * @param viewName the name of the view to be dropped.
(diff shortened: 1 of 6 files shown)


cloud-fan:
cc @rxin @yhuai

sql/core/src/main/scala/org/apache/spark/sql/catalog/Catalog.scala
```diff
@@ -270,9 +270,10 @@ abstract class Catalog {
    * tied to any databases, i.e. we can't use `db1.view1` to reference a local temporary view.
    *
```
rxin: can you add a line saying the return type was unit in Spark 2.0, but changed to boolean in Spark 2.1?

rxin:
LGTM other than the two minor comments.

We also need a Python API for this, don't we?

rxin:
LGTM pending Jenkins.

cloud-fan:
merging to master!

PR #15427, opened 2016-10-11
[SPARK-17866][SPARK-17867][SQL] Fix Dataset.dropduplicates

viirya:

Two issues regarding Dataset.dropduplicates:
1. Dataset.dropDuplicates should consider the columns with same column name
   
   We find and get the first resolved attribute from output with the given column name in `Dataset.dropDuplicates`. When we have the more than one columns with the same name. Other columns are put into aggregation columns, instead of grouping columns.
2. Dataset.dropDuplicates should not change the output of child plan
   
   We create new `Alias` with new exprId in `Dataset.dropDuplicates` now. However it causes problem when we want to select the columns as follows:
   
   ```
   val ds = Seq(("a", 1), ("a", 2), ("b", 1), ("a", 1)).toDS()
   // ds("_2") will cause analysis exception
   ds.dropDuplicates("_1").select(ds("_1").as[String], ds("_2").as[Int])
   ```

Because the two issues are both related to `Dataset.dropduplicates` and the code changes are not big, so submitting them together as one PR.
## How was this patch tested?

Jenkins tests.

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/Dataset.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/Dataset.scala
@@ -1893,7 +1893,9 @@ class Dataset[T] private[sql](
       if (groupColExprIds.contains(attr.exprId)) {
         attr
       } else {
-        // We should keep the original exprId of the attribute.
+        // Removing duplicate rows should not change output attributes. We should keep
+        // the original exprId of the attribute. Otherwise, to select a column in original
+        // dataset will cause analysis exception due to unresolved attribute.
         Alias(new First(attr).toAggregateExpression(), attr.name)(exprId = attr.exprId)
       }
     }
--- a/sql/core/src/main/scala/org/apache/spark/sql/Dataset.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/Dataset.scala
@@ -1878,17 +1878,23 @@ class Dataset[T] private[sql](
   def dropDuplicates(colNames: Seq[String]): Dataset[T] = withTypedPlan {
     val resolver = sparkSession.sessionState.analyzer.resolver
     val allColumns = queryExecution.analyzed.output
-    val groupCols = colNames.map { colName =>
-      allColumns.find(col => resolver(col.name, colName)).getOrElse(
+    val groupCols = colNames.flatMap { colName =>
+      // It is possibly there are more than one columns with the same name,
+      // so we call filter instead of find.
+      val cols = allColumns.filter(col => resolver(col.name, colName))
+      if (cols.isEmpty) {
         throw new AnalysisException(
-          s"""Cannot resolve column name "$colName" among (${schema.fieldNames.mkString(", ")})"""))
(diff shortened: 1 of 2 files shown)


viirya:
cc @cloud-fan @hvanhovell

sql/core/src/main/scala/org/apache/spark/sql/Dataset.scala
```diff
@@ -1878,17 +1878,23 @@ class Dataset[T] private[sql](
...
+      }
+      cols
     }
     val groupColExprIds = groupCols.map(_.exprId)
     val aggCols = logicalPlan.output.map { attr =>
       if (groupColExprIds.contains(attr.exprId)) {
         attr
       } else {
-        Alias(new First(attr).toAggregateExpression(), attr.name)()
+        // We should keep the original exprId of the attribute.
```
rxin: can you explain why in the comment why we should keep the original exprId? Otherwise this comment is redundant with the code itself.

cloud-fan:
My thoughts:

1. `Dataset.dropDuplicates()` should drop duplicates for all columns, the current implementation is wrong, this PR fixed it.
2. `Dataset.dropDuplicates(col: String)` should drop the first column matching the given name, or all matched columns? `Dataset.drop(col: String)` also drops all matched columns, the new behaviour seems reasonable. But we should be very careful, as this is a breaking change. cc @rxin

rxin:
1. `Dataset.dropDuplicates()` should definitely drop duplicates for all columns.

2. `Dataset.dropDuplicates(col: String)` should also drop duplicates for all columns matching the name.

sql/core/src/main/scala/org/apache/spark/sql/Dataset.scala
```diff
@@ -1878,17 +1878,25 @@ class Dataset[T] private[sql](
...
     val resolver = sparkSession.sessionState.analyzer.resolver
     val allColumns = queryExecution.analyzed.output
-    val groupCols = colNames.map { colName =>
-      allColumns.find(col => resolver(col.name, colName)).getOrElse(
+    val groupCols = colNames.flatMap { colName =>
+      // It is possibly there are more than one columns with the same name,
+      // so we call filter instead of find.
+      val cols = allColumns.filter(col => resolver(col.name, colName))
+      if (cols.isEmpty) {
         throw new AnalysisException(
```
cloud-fan: `Dataset.drop` is a no-op if the given name doesn't match any column. Should we follow it?

viirya: My thought is:

When an user mistakenly gives wrong column to `Dataset.drop`, it can be easily found out.

But for `Dataset.dropDuplicates`, it might be harder to figure out duplicate rows are still there. So to throw an explicit exception looks more proper to me.

cloud-fan:
LGTM, merging to master!

viirya:
Thanks for review! @rxin @cloud-fan

PR #15428, opened 2016-10-11
[SPARK-17219][ML] enhanced NaN value handling in Bucketizer

VinceShieh:

This PR is an enhancement of PR with commit ID:57dc326bd00cf0a49da971e9c573c48ae28acaa2.
NaN is a special type of value which is commonly seen as invalid. But We find that there are certain cases where NaN are also valuable, thus need special handling. We provided user when dealing NaN values with 3 options, to either reserve an extra bucket for NaN values, or remove the NaN values, or report an error, by setting handleNaN "keep", "skip", or "error"(default) respectively.

'''Before:
val bucketizer: Bucketizer = new Bucketizer()
          .setInputCol("feature")
          .setOutputCol("result")
          .setSplits(splits)
'''After:
val bucketizer: Bucketizer = new Bucketizer()
          .setInputCol("feature")
          .setOutputCol("result")
          .setSplits(splits)
          .setHandleNaN("keep")
## How was this patch tested?

Tests added in QuantileDiscretizerSuite, BucketizerSuite and DataFrameStatSuite

Signed-off-by: VinceShieh vincent.xie@intel.com

### DIFF
--- a/docs/ml-features.md
+++ b/docs/ml-features.md
@@ -1103,13 +1103,16 @@ for more details on the API.
 
 `QuantileDiscretizer` takes a column with continuous features and outputs a column with binned
 categorical features. The number of bins is set by the `numBuckets` parameter. It is possible
-that the number of buckets used will be less than this value, for example, if there are too few
-distinct values of the input to create enough distinct quantiles. Note also that QuantileDiscretizer
-will raise an error when it finds NaN value in the dataset, but user can also choose to either
-keep or remove NaN values within the dataset by setting handleInvalid. If user chooses to keep
+that the number of buckets used will be smaller than this value, for example, if there are too few
+distinct values of the input to create enough distinct quantiles.
+
+NaN values: Note also that QuantileDiscretizer
+will raise an error when it finds NaN values in the dataset, but the user can also choose to either
+keep or remove NaN values within the dataset by setting `handleInvalid`. If the user chooses to keep
 NaN values, they will be handled specially and placed into their own bucket, for example, if 4 buckets
 are used, then non-NaN data will be put into buckets[0-3], but NaNs will be counted in a special bucket[4].
-The bin ranges are chosen using an approximate algorithm (see the documentation for
+
+Algorithm: The bin ranges are chosen using an approximate algorithm (see the documentation for
 [approxQuantile](api/scala/index.html#org.apache.spark.sql.DataFrameStatFunctions) for a
 detailed description). The precision of the approximation can be controlled with the
 `relativeError` parameter. When set to zero, exact quantiles are calculated
--- a/docs/ml-features.md
+++ b/docs/ml-features.md
@@ -1104,9 +1104,11 @@ for more details on the API.
 `QuantileDiscretizer` takes a column with continuous features and outputs a column with binned
 categorical features. The number of bins is set by the `numBuckets` parameter. It is possible
 that the number of buckets used will be less than this value, for example, if there are too few
-distinct values of the input to create enough distinct quantiles. Note also that NaN values are
-handled specially and placed into their own bucket. For example, if 4 buckets are used, then
-non-NaN data will be put into buckets[0-3], but NaNs will be counted in a special bucket[4].
+distinct values of the input to create enough distinct quantiles. Note also that QuantileDiscretizer
+will raise an error when it finds NaN value in the dataset, but user can also choose to either
+keep or remove NaN values within the dataset by setting handleInvalid. If user chooses to keep
+NaN values, they will be handled specially and placed into their own bucket, for example, if 4 buckets
+are used, then non-NaN data will be put into buckets[0-3], but NaNs will be counted in a special bucket[4].
 The bin ranges are chosen using an approximate algorithm (see the documentation for
 [approxQuantile](api/scala/index.html#org.apache.spark.sql.DataFrameStatFunctions) for a
 detailed description). The precision of the approximation can be controlled with the
--- a/mllib/src/main/scala/org/apache/spark/ml/feature/Bucketizer.scala
+++ b/mllib/src/main/scala/org/apache/spark/ml/feature/Bucketizer.scala
@@ -47,6 +47,9 @@ final class Bucketizer @Since("1.4.0") (@Since("1.4.0") override val uid: String
    * also includes y. Splits should be of length >= 3 and strictly increasing.
    * Values at -inf, inf must be explicitly provided to cover all Double values;
    * otherwise, values outside the splits specified will be treated as errors.
+   *
+   * See also [[handleInvalid]], which can optionally create an additional bucket for NaN values.
+   *
    * @group param
    */
   @Since("1.4.0")
@@ -75,37 +78,36 @@ final class Bucketizer @Since("1.4.0") (@Since("1.4.0") override val uid: String
   def setOutputCol(value: String): this.type = set(outputCol, value)
 
   /**
-   * Param for how to handle invalid entries. Options are skip (which will filter out rows with
-   * invalid values), or error (which will throw an error), or keep (which will keep the invalid
-   * values in certain way).
+   * Param for how to handle invalid entries. Options are skip (filter out rows with
+   * invalid values), error (throw an error), or keep (keep invalid values in a special additional
+   * bucket).
    * Default: "error"
    * @group param
    */
   @Since("2.1.0")
   val handleInvalid: Param[String] = new Param[String](this, "handleInvalid", "how to handle" +
-    "invalid entries. Options are skip (which will filter out rows with invalid values), or" +
-    "error (which will throw an error), or keep (which will keep the invalid values" +
-    " in certain way). Default behaviour is to report an error for invalid entries.",
-    ParamValidators.inArray(Array("skip", "error", "keep")))
+    "invalid entries. Options are skip (filter out rows with invalid values), " +
+    "error (throw an error), or keep (keep invalid values in a special additional bucket).",
+    ParamValidators.inArray(Bucketizer.supportedHandleInvalid))
 
   /** @group getParam */
   @Since("2.1.0")
-  def gethandleInvalid: String = $(handleInvalid)
+  def getHandleInvalid: String = $(handleInvalid)
 
   /** @group setParam */
   @Since("2.1.0")
-  def sethandleInvalid(value: String): this.type = set(handleInvalid, value)
-  setDefault(handleInvalid, "error")
+  def setHandleInvalid(value: String): this.type = set(handleInvalid, value)
+  setDefault(handleInvalid, Bucketizer.ERROR_INVALID)
 
   @Since("2.0.0")
   override def transform(dataset: Dataset[_]): DataFrame = {
     transformSchema(dataset.schema)
     val (filteredDataset, keepInvalid) = {
-      if ("skip" == gethandleInvalid) {
+      if (getHandleInvalid == Bucketizer.SKIP_INVALID) {
         // "skip" NaN option is set, will filter out NaN values in the dataset
-        (dataset.na.drop.toDF(), false)
+        (dataset.na.drop().toDF(), false)
       } else {
-        (dataset.toDF(), "keep" == gethandleInvalid)
+        (dataset.toDF(), getHandleInvalid == Bucketizer.KEEP_INVALID)
       }
     }
 
@@ -140,6 +142,12 @@ final class Bucketizer @Since("1.4.0") (@Since("1.4.0") override val uid: String
 @Since("1.6.0")
 object Bucketizer extends DefaultParamsReadable[Bucketizer] {
 
+  private[feature] val SKIP_INVALID: String = "skip"
+  private[feature] val ERROR_INVALID: String = "error"
+  private[feature] val KEEP_INVALID: String = "keep"
+  private[feature] val supportedHandleInvalid: Array[String] =
+    Array(SKIP_INVALID, ERROR_INVALID, KEEP_INVALID)
+
   /**
    * We require splits to be of length >= 3 and to be in strictly increasing order.
    * No NaN split should be accepted.
@@ -173,9 +181,13 @@ object Bucketizer extends DefaultParamsReadable[Bucketizer] {
       splits: Array[Double],
       feature: Double,
       keepInvalid: Boolean): Double = {
-    if (feature.isNaN && keepInvalid) {
-      // NaN data point found plus "keep" NaN option is set
-      splits.length - 1
+    if (feature.isNaN) {
+      if (keepInvalid) {
+        splits.length - 1
+      } else {
+        throw new SparkException("Bucketizer encountered NaN value. To handle or skip NaNs," +
+          " try setting Bucketizer.handleInvalid.")
+      }
     } else if (feature == splits.last) {
       splits.length - 2
     } else {
--- a/mllib/src/main/scala/org/apache/spark/ml/feature/Bucketizer.scala
+++ b/mllib/src/main/scala/org/apache/spark/ml/feature/Bucketizer.scala
@@ -77,8 +77,8 @@ final class Bucketizer @Since("1.4.0") (@Since("1.4.0") override val uid: String
   /**
    * Param for how to handle invalid entries. Options are skip (which will filter out rows with
    * invalid values), or error (which will throw an error), or keep (which will keep the invalid
-   * values in certain way). Default behaviour is to report an error for invalid entries.
-   *
+   * values in certain way).
+   * Default: "error"
    * @group param
    */
   @Since("2.1.0")
@@ -90,11 +90,7 @@ final class Bucketizer @Since("1.4.0") (@Since("1.4.0") override val uid: String
 
   /** @group getParam */
   @Since("2.1.0")
-  def gethandleInvalid: Option[Boolean] = $(handleInvalid) match {
-    case "keep" => Some(true)
-    case "skip" => Some(false)
-    case _ => None
-  }
+  def gethandleInvalid: String = $(handleInvalid)
 
   /** @group setParam */
   @Since("2.1.0")
@@ -104,19 +100,19 @@ final class Bucketizer @Since("1.4.0") (@Since("1.4.0") override val uid: String
   @Since("2.0.0")
   override def transform(dataset: Dataset[_]): DataFrame = {
     transformSchema(dataset.schema)
-    val keepInvalid = gethandleInvalid.isDefined && gethandleInvalid.get
+    val (filteredDataset, keepInvalid) = {
+      if ("skip" == gethandleInvalid) {
+        // "skip" NaN option is set, will filter out NaN values in the dataset
+        (dataset.na.drop.toDF(), false)
+      } else {
+        (dataset.toDF(), "keep" == gethandleInvalid)
+      }
+    }
 
     val bucketizer: UserDefinedFunction = udf { (feature: Double) =>
       Bucketizer.binarySearchForBuckets($(splits), feature, keepInvalid)
     }
-    val filteredDataset = {
-      if (!keepInvalid) {
-        // "skip" NaN option is set, will filter out NaN values in the dataset
-        dataset.na.drop.toDF()
-      } else {
-        dataset.toDF()
-      }
-    }
+
     val newCol = bucketizer(filteredDataset($(inputCol)))
     val newField = prepOutputField(filteredDataset.schema)
     filteredDataset.withColumn($(outputCol), newCol, newField.metadata)
--- a/mllib/src/main/scala/org/apache/spark/ml/feature/Bucketizer.scala
+++ b/mllib/src/main/scala/org/apache/spark/ml/feature/Bucketizer.scala
@@ -27,6 +27,7 @@ import org.apache.spark.ml.param._
 import org.apache.spark.ml.param.shared.{HasInputCol, HasOutputCol}
 import org.apache.spark.ml.util._
 import org.apache.spark.sql._
+import org.apache.spark.sql.expressions.UserDefinedFunction
 import org.apache.spark.sql.functions._
 import org.apache.spark.sql.types.{DoubleType, StructField, StructType}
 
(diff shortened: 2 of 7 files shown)


mllib/src/main/scala/org/apache/spark/ml/param/shared/sharedParams.scala
```diff
@@ -270,10 +270,10 @@ private[ml] trait HasFitIntercept extends Params {
 private[ml] trait HasHandleInvalid extends Params {
 
   /**
-   * Param for how to handle invalid entries. Options are skip (which will filter out rows with bad values), or error (which will throw an error). More options may be added later.
+   * Param for how to handle invalid entries. Options are skip (which will filter out rows with bad values), or error (which will throw an error), or keep (which will keep the bad values in certain way). More options may be added later.
```
srowen: I'm neutral on the complexity that this adds, but not against it. It gets a little funny to say "keep invalid data" but I think we discussed that on the JIRA

mllib/src/main/scala/org/apache/spark/ml/feature/Bucketizer.scala
```diff
@@ -73,15 +78,27 @@ final class Bucketizer @Since("1.4.0") (@Since("1.4.0") override val uid: String
...
+
   @Since("2.0.0")
   override def transform(dataset: Dataset[_]): DataFrame = {
     transformSchema(dataset.schema)
-    val bucketizer = udf { feature: Double =>
-      Bucketizer.binarySearchForBuckets($(splits), feature)
+    val bucketizer: UserDefinedFunction = udf { (feature: Double, flag: String) =>
+      Bucketizer.binarySearchForBuckets($(splits), feature, flag)
+    }
+    val filteredDataset = {
```
srowen: Doesn't this need to try to handle "error"?

```
val filteredDataSet = getHandleInvalid match {
  case "skip" => dataset.na.drop
  case "keep" => dataset
  case "error" => 
    if (...dataset contains NaN...) {
      throw new IllegalArgumentException(...)
   } else {
      dataset
   }
}
```

mllib/src/main/scala/org/apache/spark/ml/feature/Bucketizer.scala
```diff
@@ -128,8 +145,9 @@ object Bucketizer extends DefaultParamsReadable[Bucketizer] {
    * Binary searching in several buckets to place each data point.
    * @throws SparkException if a feature is < splits.head or > splits.last
    */
-  private[feature] def binarySearchForBuckets(splits: Array[Double], feature: Double): Double = {
-    if (feature.isNaN) {
+  private[feature] def binarySearchForBuckets
+  (splits: Array[Double], feature: Double, flag: String): Double = {
```
srowen: Nit: I think the convention is to leave the open paren on the previous line

Doesn't this need to handle "skip" and "error"? throw an exception on NaN if "error" or ignore it if "skip"?

srowen (changes requested):

mllib/src/main/scala/org/apache/spark/ml/param/shared/sharedParams.scala
VinceShieh: yeah, usually we treat NaN as a type of invalid value, but we know there are cases they are proved useful, so, it's a bit dilemma here, also with the API naming.

mllib/src/main/scala/org/apache/spark/ml/feature/Bucketizer.scala
srowen: I don't see that the method handles `NaN` below. What binarySearch returns is undefined. One place or the other I think this has to be explicitly handled.

srowen (changes requested):
I'll let @thunterdb or @jkbradley mostly review from here to see if this is what they intended

jkbradley:
Thanks! I'll take a look.

Could you please fix the typo in the title? "enchanced" -> "enhanced"

VinceShieh:
typo corrected. Thank you all. @srowen @jkbradley

docs/ml-features.md
```diff
@@ -1104,9 +1104,11 @@ for more details on the API.
 `QuantileDiscretizer` takes a column with continuous features and outputs a column with binned
 categorical features. The number of bins is set by the `numBuckets` parameter. It is possible
 that the number of buckets used will be less than this value, for example, if there are too few
-distinct values of the input to create enough distinct quantiles. Note also that NaN values are
-handled specially and placed into their own bucket. For example, if 4 buckets are used, then
-non-NaN data will be put into buckets[0-3], but NaNs will be counted in a special bucket[4].
+distinct values of the input to create enough distinct quantiles. Note also that QuantileDiscretizer
```
jkbradley: (same as below) Is "possible that the number of buckets used will be less than this value" true?  It was true before this used Dataset.approxQuantiles, but I don't think it is true any longer.

mllib/src/main/scala/org/apache/spark/ml/feature/Bucketizer.scala
```diff
@@ -73,15 +78,27 @@ final class Bucketizer @Since("1.4.0") (@Since("1.4.0") override val uid: String
...
+  @Since("2.1.0")
+  def setHandleInvalid(value: String): this.type = set(handleInvalid, value)
+  setDefault(handleInvalid, "error")
+
   @Since("2.0.0")
   override def transform(dataset: Dataset[_]): DataFrame = {
     transformSchema(dataset.schema)
-    val bucketizer = udf { feature: Double =>
-      Bucketizer.binarySearchForBuckets($(splits), feature)
+    val bucketizer: UserDefinedFunction = udf { (feature: Double, flag: String) =>
```
jkbradley: Also, at this point, you already know the value for flag, so you can just use it here, rather than making the UDF take an extra argument.

mllib/src/main/scala/org/apache/spark/ml/feature/Bucketizer.scala
```diff
@@ -73,15 +78,27 @@ final class Bucketizer @Since("1.4.0") (@Since("1.4.0") override val uid: String
...
+      Bucketizer.binarySearchForBuckets($(splits), feature, flag)
+    }
+    val filteredDataset = {
+      if ("skip" == getHandleInvalid) {
+        dataset.na.drop
+      } else {
+        dataset
```
jkbradley: I'd put ```dataset.toDF()``` here and not import existentials.

mllib/src/main/scala/org/apache/spark/ml/feature/Bucketizer.scala
```diff
@@ -128,8 +145,11 @@ object Bucketizer extends DefaultParamsReadable[Bucketizer] {
    * Binary searching in several buckets to place each data point.
    * @throws SparkException if a feature is < splits.head or > splits.last
    */
-  private[feature] def binarySearchForBuckets(splits: Array[Double], feature: Double): Double = {
-    if (feature.isNaN) {
+  private[feature] def binarySearchForBuckets(
+                                               splits: Array[Double],
+                                               feature: Double,
+                                               flag: String): Double = {
```
jkbradley: Use a better name than "flag"  How about "keepNaNs"?
Also, please document this argument.

mllib/src/main/scala/org/apache/spark/ml/feature/QuantileDiscretizer.scala
```diff
@@ -66,11 +67,13 @@ private[feature] trait QuantileDiscretizerBase extends Params
 /**
  * `QuantileDiscretizer` takes a column with continuous features and outputs a column with binned
  * categorical features. The number of bins can be set using the `numBuckets` parameter. It is
- * possible that the number of buckets used will be less than this value, for example, if there
- * are too few distinct values of the input to create enough distinct quantiles. Note also that
- * NaN values are handled specially and placed into their own bucket. For example, if 4 buckets
- * are used, then non-NaN data will be put into buckets(0-3), but NaNs will be counted in a special
- * bucket(4).
+ * possible that the number of buckets used will be less than this value, for example, if there are
```
jkbradley: Is this true?  It was true before this used Dataset.approxQuantiles, but I don't think it is true any longer.

mllib/src/main/scala/org/apache/spark/ml/feature/QuantileDiscretizer.scala
```diff
@@ -66,11 +67,13 @@ private[feature] trait QuantileDiscretizerBase extends Params
...
+ * too few distinct values of the input to create enough distinct quantiles. Note also that
+ * QuantileDiscretizer will raise an error when it finds NaN value in the dataset, but user can
+ * also choose to either keep or remove NaN values within the dataset by calling setHandleInvalid.
```
jkbradley: "calling setHandleInvalid" -> "setting `handleInvalid`"

mllib/src/main/scala/org/apache/spark/ml/feature/QuantileDiscretizer.scala
```diff
@@ -66,11 +67,13 @@ private[feature] trait QuantileDiscretizerBase extends Params
...
+ * If user chooses to keep NaN values, they will be handled specially and placed into their own
```
jkbradley: This documentation should go in the handleInvalid Param doc string.

mllib/src/test/scala/org/apache/spark/ml/feature/BucketizerSuite.scala
```diff
@@ -114,6 +115,7 @@ class BucketizerSuite extends SparkFunSuite with MLlibTestSparkContext with Defa
           .setInputCol("feature")
```
jkbradley: Just set splits here; there's no need to set the other Params for this test.

mllib/src/test/scala/org/apache/spark/ml/feature/QuantileDiscretizerSuite.scala
```diff
@@ -83,10 +83,20 @@ class QuantileDiscretizerSuite
       .setOutputCol("result")
       .setNumBuckets(numBuckets)
 
-    // Reserve extra one bucket for NaN
-    val expectedNumBuckets = discretizer.fit(df).getSplits.length - 1
-    val result = discretizer.fit(df).transform(df)
-    val observedNumBuckets = result.select("result").distinct.count
+    // Reserve extra one bucket for NaN values
+    discretizer.setHandleInvalid("keep")
+    var expectedNumBuckets = discretizer.fit(df).getSplits.length - 1
```
jkbradley: I'd also recommend directly testing the values of the splits.  The current test makes sure that handleInvalid is passed to the bucketizer correctly, which is important but separate.

Also, please use vals (not vars) for clarity.  I'd recommend making a helper method for lines 87-93, which can be reused for the test of handleInvalid = "skip"

mllib/src/test/scala/org/apache/spark/ml/feature/QuantileDiscretizerSuite.scala
```diff
@@ -83,10 +83,20 @@ class QuantileDiscretizerSuite
...
+    var result = discretizer.fit(df).transform(df)
+    var observedNumBuckets = result.select("result").distinct.count
+    assert(observedNumBuckets == expectedNumBuckets,
+      s"Observed number of buckets are not correct." +
+        s" Expected $expectedNumBuckets but found $observedNumBuckets")
+
+    // Remove NaN values
+    discretizer.setHandleInvalid("skip")
+    expectedNumBuckets = discretizer.fit(df).getSplits.length - 2
```
jkbradley: Same here; I'd also recommend directly testing the values of the splits.

python/pyspark/ml/feature.py
```diff
@@ -1157,9 +1157,11 @@ class QuantileDiscretizer(JavaEstimator, HasInputCol, HasOutputCol, JavaMLReadab
     categorical features. The number of bins can be set using the :py:attr:`numBuckets` parameter.
     It is possible that the number of buckets used will be less than this value, for example, if
```
jkbradley: Here too: no longer the case

jkbradley (commented):
Done with review.  Thanks for the PR!

Other comments:
- Add unit test for `DataFrameStatFunctions.approxQuantile` to check for handling of NaN values.
- Create JIRA for adding handleInvalid to Python API.

docs/ml-features.md
VinceShieh: In cases when the number of buckets requested by the user is greater than the number of distinct splits generated from Bucketizer, the returned number of buckets will be less than requested.

mllib/src/main/scala/org/apache/spark/ml/feature/QuantileDiscretizer.scala
VinceShieh: same as the comment above

python/pyspark/ml/feature.py
VinceShieh: same as the comment above.

VinceShieh:
Thanks for your valuable suggestions. @jkbradley @srowen

mllib/src/main/scala/org/apache/spark/ml/feature/Bucketizer.scala
```diff
@@ -73,15 +74,51 @@ final class Bucketizer @Since("1.4.0") (@Since("1.4.0") override val uid: String
...
+  def setHandleNaN(value: String): this.type = set(handleNaN, value)
+  setDefault(handleNaN, "error")
+
   @Since("2.0.0")
   override def transform(dataset: Dataset[_]): DataFrame = {
     transformSchema(dataset.schema)
-    val bucketizer = udf { feature: Double =>
-      Bucketizer.binarySearchForBuckets($(splits), feature)
+
+    val bucketizer: UserDefinedFunction = udf { (feature: Double) =>
```
jkbradley: Ah, sorry, one more comment.  I'm not quite sure how closure capture behaves currently, but it might be good to define local vals for ```$(splits)``` and ```getHandleNaN.isDefined && getHandleNaN.get```.  Since these reference methods in the Bucketizer class, I believe the UDF may capture the whole Bucketizer class instead of just those vals.

After you define them in local vals here, you can use those vals in this UDF.

docs/ml-features.md
jkbradley: Yep, you're right

mllib/src/test/scala/org/apache/spark/ml/feature/QuantileDiscretizerSuite.scala
```diff
@@ -82,14 +82,23 @@ class QuantileDiscretizerSuite
...
-    val result = discretizer.fit(df).transform(df)
-    val observedNumBuckets = result.select("result").distinct.count
-    assert(observedNumBuckets == expectedNumBuckets,
-      s"Observed number of buckets are not correct." +
-        s" Expected $expectedNumBuckets but found $observedNumBuckets")
+    val splits = Array(Double.NegativeInfinity, 1.0, Double.PositiveInfinity)
+
+    for (handleNaN <- Array("keep", "skip")) {
+      discretizer.setHandleNaN(handleNaN)
+      val expectedNumBuckets = {
```
jkbradley: This is kind of complex logic just to get the expected number of splits.  I'd recommend just putting the expected bucket *values* in the original DataFrame:
```
val df = sc.parallelize(Seq(
  (1.0, /*expected value for option "keep"*/, /*expected value for option "skip"*/),
  ...
)).toDF("input", "expectedKeep", "expectedSkip")
```
Then you can compare with the actual values.  That'll be an easier test to understand IMO.

python/pyspark/ml/feature.py
```diff
@@ -1157,9 +1157,11 @@ class QuantileDiscretizer(JavaEstimator, HasInputCol, HasOutputCol, JavaMLReadab
...
     there are too few distinct values of the input to create enough distinct quantiles. Note also
-    that NaN values are handled specially and placed into their own bucket. For example, if 4
-    buckets are used, then non-NaN data will be put into buckets(0-3), but NaNs will be counted in
-    a special bucket(4).
+    that QuantileDiscretizer will raise an error when it finds NaN value in the dataset, but user
```
jkbradley: Actually no need to update Python API until it is updated to include handleNaN

mllib/src/main/scala/org/apache/spark/ml/feature/QuantileDiscretizer.scala
```diff
@@ -100,6 +102,24 @@ final class QuantileDiscretizer @Since("1.6.0") (@Since("1.6.0") override val ui
...
 
+  /**
+   * Param for how to handle NaN entries. Options are skip (which will filter out rows with
+   * NaN values), or error (which will throw an error), or keep (which will make the NaN
+   * values an extra bucket). More options may be added later.
+   *
+   * @group param
+   */
+  @Since("2.1.0")
+  val handleNaN: Param[String] = new Param[String](this, "handleNaNs", "how to handle NaN" +
```
jkbradley: Add getHandleNaN

mllib/src/main/scala/org/apache/spark/ml/feature/QuantileDiscretizer.scala
```diff
@@ -100,6 +102,24 @@ final class QuantileDiscretizer @Since("1.6.0") (@Since("1.6.0") override val ui
   @Since("1.6.0")
   def setOutputCol(value: String): this.type = set(outputCol, value)
 
+  /**
+   * Param for how to handle NaN entries. Options are skip (which will filter out rows with
+   * NaN values), or error (which will throw an error), or keep (which will make the NaN
+   * values an extra bucket). More options may be added later.
+   *
```
jkbradley: State default

mllib/src/main/scala/org/apache/spark/ml/feature/Bucketizer.scala
```diff
@@ -73,15 +74,51 @@ final class Bucketizer @Since("1.4.0") (@Since("1.4.0") override val uid: String
   @Since("1.4.0")
   def setOutputCol(value: String): this.type = set(outputCol, value)
 
+  /**
+   * Param for how to handle NaN entries. Options are skip (which will filter out rows with
+   * NaN values), or error (which will throw an error), or keep (which will make the NaN
+   * values an extra bucket). More options may be added later.
```
jkbradley: Remove "More options may be added later."

Also state default.

mllib/src/main/scala/org/apache/spark/ml/feature/Bucketizer.scala
```diff
@@ -73,15 +74,51 @@ final class Bucketizer @Since("1.4.0") (@Since("1.4.0") override val uid: String
...
+   *
+   * @group param
+   */
+  @Since("2.1.0")
+  val handleNaN: Param[String] = new Param[String](this, "handleNaNs", "how to handle NaN" +
+    "entries. Options are skip (which will filter out rows with NaN values), or error" +
+    "(which will throw an error), or keep (which will make the NaN values an extra bucket)." +
+    "More options may be added later", ParamValidators.inArray(Array("skip", "error", "keep")))
```
jkbradley: Remove "More options may be added later."

jkbradley (commented):
Thanks for the updates!  A few more comments.

Also, I see you renamed the param to handleNaN.  Do you think we will ever want to handle null values too?  It might be good to stick with "handleInvalid" in case we add null support.

mllib/src/main/scala/org/apache/spark/ml/feature/Bucketizer.scala
```diff
@@ -73,15 +74,52 @@ final class Bucketizer @Since("1.4.0") (@Since("1.4.0") override val uid: String
   @Since("1.4.0")
   def setOutputCol(value: String): this.type = set(outputCol, value)
 
+  /**
+   * Param for how to handle invalid entries. Options are skip (which will filter out rows with
+   * invalid values), or error (which will throw an error), or keep (which will keep the invalid
+   * values in certain way). Default behaviour is to report an error for invalid entries.
```
jkbradley: I'd just write: ```default: "error"```
Rewording as "report" instead of "throw" could confuse people.

mllib/src/main/scala/org/apache/spark/ml/feature/Bucketizer.scala
```diff
@@ -73,15 +74,52 @@ final class Bucketizer @Since("1.4.0") (@Since("1.4.0") override val uid: String
...
+  @Since("2.1.0")
+  val handleInvalid: Param[String] = new Param[String](this, "handleInvalid", "how to handle" +
+    "invalid entries. Options are skip (which will filter out rows with invalid values), or" +
+    "error (which will throw an error), or keep (which will keep the invalid values" +
+    " in certain way). Default behaviour is to report an error for invalid entries.",
+    ParamValidators.inArray(Array("skip", "error", "keep")))
+
+  /** @group getParam */
+  @Since("2.1.0")
+  def gethandleInvalid: Option[Boolean] = $(handleInvalid) match {
```
jkbradley: This should just return ```$(handleInvalid)```, just like any other Param getter method.

mllib/src/main/scala/org/apache/spark/ml/feature/QuantileDiscretizer.scala
```diff
@@ -100,6 +102,33 @@ final class QuantileDiscretizer @Since("1.6.0") (@Since("1.6.0") override val ui
   @Since("1.6.0")
   def setOutputCol(value: String): this.type = set(outputCol, value)
 
+  /**
+   * Param for how to handle invalid entries. Options are skip (which will filter out rows with
+   * invalid values), or error (which will throw an error), or keep (which will keep the invalid
+   * values in certain way). Default behaviour is to report an error for invalid entries.
```
jkbradley: I'd just write: ```default: "error"```
Rewording as "report" instead of "throw" could confuse people.

python/pyspark/ml/feature.py
jkbradley: This isn't available in Python yet, so can you please revert this change to feature.py?

jkbradley (commented):
Thanks for the updates.
Also, when you push further updates, could you please just push a new commit?  Rebasing makes it harder to review since it makes it impossible to connect your most recent changes with the recent reviewer comments.  Thanks!

jkbradley:
Thanks, you still need to remove the change in feature.py, but other than that, this should be ready.

VinceShieh:
sorry, I must have forgotten to commit the changes.
All done now. Thanks for reviewing.

jkbradley:
Thanks for the update!  On a last glance through, I spotted a few things to fix up.  Since some had to do with wording, I thought it'd be easier just to send a PR to your PR.  Can you please review this and merge it if it looks Ok to you?  Thanks!  Here it is: [https://github.com/VinceShieh/spark/pull/2]

jkbradley:
Thanks for merging that!  This LGTM
Merging with master
Thanks a lot @VinceShieh !

PR #15432, opened 2016-10-11
[SPARK-17854][SQL] rand/randn allows null/long as input seed

HyukjinKwon:

This PR proposes `rand`/`randn` accept `null` as input in Scala/SQL and `LongType` as input in SQL. In this case, it treats the values as `0`.

So, this PR includes both changes below:
- `null` support
  
  It seems MySQL also accepts this.
  
  ``` sql
  mysql> select rand(0);
  +---------------------+
  | rand(0)             |
  +---------------------+
  | 0.15522042769493574 |
  +---------------------+
  1 row in set (0.00 sec)
  
  mysql> select rand(NULL);
  +---------------------+
  | rand(NULL)          |
  +---------------------+
  | 0.15522042769493574 |
  +---------------------+
  1 row in set (0.00 sec)
  ```
  
  and also Hive does according to [HIVE-14694](https://issues.apache.org/jira/browse/HIVE-14694)
  
  So the codes below:
  
  ``` scala
  spark.range(1).selectExpr("rand(null)").show()
  ```
  
  prints..
  
  **Before**
  
  ```
    Input argument to rand must be an integer literal.;; line 1 pos 0
  org.apache.spark.sql.AnalysisException: Input argument to rand must be an integer literal.;; line 1 pos 0
  at org.apache.spark.sql.catalyst.analysis.FunctionRegistry$$anonfun$5.apply(FunctionRegistry.scala:465)
  at org.apache.spark.sql.catalyst.analysis.FunctionRegistry$$anonfun$5.apply(FunctionRegistry.scala:444)
  ```
  
  **After**
  
  ```
    +-----------------------+
    |rand(CAST(NULL AS INT))|
    +-----------------------+
    |    0.13385709732307427|
    +-----------------------+
  ```
- `LongType` support in SQL.
  
  In addition, it make the function allows to take `LongType` consistently within Scala/SQL.
  
  In more details, the codes below:
  
  ``` scala
  spark.range(1).select(rand(1), rand(1L)).show()
  spark.range(1).selectExpr("rand(1)", "rand(1L)").show()
  ```
  
  prints..
  
  **Before**
  
  ```
  +------------------+------------------+
  |           rand(1)|           rand(1)|
  +------------------+------------------+
  |0.2630967864682161|0.2630967864682161|
  +------------------+------------------+
  
  
  Input argument to rand must be an integer literal.;; line 1 pos 0
  org.apache.spark.sql.AnalysisException: Input argument to rand must be an integer literal.;; line 1 pos 0
  at org.apache.spark.sql.catalyst.analysis.FunctionRegistry$$anonfun$5.apply(FunctionRegistry.scala:465)
  at
  ```
  
  **After**
  
  ```
  +------------------+------------------+
  |           rand(1)|           rand(1)|
  +------------------+------------------+
  |0.2630967864682161|0.2630967864682161|
  +------------------+------------------+
  
  +------------------+------------------+
  |           rand(1)|           rand(1)|
  +------------------+------------------+
  |0.2630967864682161|0.2630967864682161|
  +------------------+------------------+
  ```
## How was this patch tested?

Unit tests in `DataFrameSuite.scala` and `RandomSuite.scala`.


rxin:
hm - maybe we should just cast any NullType input into some concrete type defined by an ExpectsInputTypes expression?

HyukjinKwon:
@rxin yes, I just wanted to avoid changing a lot. Will try to fix it in that way (at least) to show how it actually look like.

HyukjinKwon:
@rxin yes, I just wanted to avoid changing a lot. Will try to fix it in that way (at least) to show how it actually look like.

HyukjinKwon:
@rxin, I updated the codes and also updated the PR description. Could you please check if my change makes sense?

HyukjinKwon:
Oh, my bad. Will fix it up.

HyukjinKwon:
FWIW, the cases below are fine in MySQL:

```sql
mysql> SELECT RAND(CAST(2 AS UNSIGNED));
+---------------------------+
| RAND(CAST(2 AS UNSIGNED)) |
+---------------------------+
|        0.6555866465490187 |
+---------------------------+
1 row in set (0.00 sec)

mysql> SELECT RAND(CAST(NULL AS UNSIGNED));
+------------------------------+
| RAND(CAST(NULL AS UNSIGNED)) |
+------------------------------+
|          0.15522042769493574 |
+------------------------------+
1 row in set (0.00 sec)
```

gatorsmile:
Not sure whether you realize it. Since this PR changes the input parm of `Rand` and `Randn`, you also changes the external support. 

Now, users can do something like
```SQL
select rand(cast(9 / 4 as int)) from src
```

My suggestion is to always add new test cases whenever you made an external change like this. It can help reviewers decide whether this PR is good or not.

gatorsmile:
Since you are running mysql, the output of rand(0) is the same as rand(null)?

HyukjinKwon:
@gatorsmile Yes (for https://github.com/apache/spark/pull/15432#issuecomment-253291901), it is and sure, I should add more tests. I actually intended to show how it looks like.

HyukjinKwon:
I should have added `[WIP]` maybe. If it look okay in general, I will try to follow your suggestions and also the case you gave.

HyukjinKwon:
@rxin How does it look like? Let me add some cases for constant folding as @gatorsmile has shown, `RAND(CAST(2 AS INT))` and `CAST(NULL AS INT)` too if it looks okay.

gatorsmile:
I have a very general comment about the work you are working. Like what we are doing for the `LIKE` operation, we did an investigation on ANSI standard, and all the mainstream data stores, including Oracle, MySQL, SQL Server, Hive, DB2, Informix and Postgres. Could you do a similar thing here? Thank you!

HyukjinKwon:
@gatorsmile Thanks for your feedback. Could you please be a little bit specific? Do you expect some researches on the argument of `rand` function in a standard and checking the DBMSs you listed above maybe?

If so, it'd be nicer if we have such reseaches on other JIRAs or PRs so that I (and other contributoers) can refer when we make a change on such thing. Is there a great example we already have maybe?

HyukjinKwon:
(Oh, I am making a comment via my phone. Sorry for occasional  closing and reopening here..)

gatorsmile:
Let me show you an example:
https://www.ibm.com/support/knowledgecenter/SSEPEK_11.0.0/sqlref/src/tpc/db2z_bif_rand.html

This is the official document of `rand` in DB2 z/OS. Below is about the input parameter:

1. If numeric-expression is specified, it is used as the seed value. The argument must be an expression that returns a value of a built-in integer data type (SMALLINT or INTEGER). The value must be between 0 and 2,147,483,646.

2. The argument must be an expression that returns a value of a built-in integer data type (SMALLINT or INTEGER). The value must be between 0 and 2,147,483,646.

3. The result can be null; if the argument is null, the result is the null value.

4. RAND(0) is processed the same as RAND().

HyukjinKwon:
That is a great reference. However, is this the function described in a standard? I guess it is different for each implementation of database. For example, 

> The result can be null; if the argument is null, the result is the null value.

MySQL treats it as 0 rather than returning `null` value. Also, I gave both references of MySQL and Hive in the PR description. Can we define the behaviour here? Do we have a target DBMS to follow? I guess it is usually Hive, PostgreSQL and MySQL as I recall.

In case of PostgreSQL, it seems there is both functions for this, `random()` and `setseed()`. This works differently with MySQL also DB2 (assuming from the comment you left). So, I got rid of this.

I think I have checked other examples enough. Do we usually have such explanations and tests of all the DBMS, Oracle, MySQL, SQL Server, Hive, DB2, Informix and PostgresSQL and mentions in ANSI standard? It can be problematic if we don't comply the standard which all other implementations follow but I think it'd be fine if other databases have different implementations.

I am sure I am taking every look for other PRs time to time and trying to make mine sensible but I don't think we always have references from all other DBMS and explanations from ANSI standard.

It is hard to change it again and that is why I am asking to review.

HyukjinKwon:
Initially, this JIRA was only handling `null` as seed. If you both worry the change here, I would like to make the PR smaller as suggested initially.

gatorsmile:
Unfortunately, not all the things have a standard to follow. That is why I suggested you to do a research about it. Like Oracle, it does not have such a function in their SQL-function list: https://docs.oracle.com/cd/B19306_01/server.102/b14200/functions001.htm

Since you are doing the change in `rand`, I think you can check whether the existing `rand` behaves as expected and adds the missing test cases if needed. This JIRA is just trying to cover an edge case of a seed number. Why not checking whether we appropriately handle all the cases? Then, we do not need to submit more small fixes for `rand`, right?

HyukjinKwon:
Strictly, the JIRA describes handling `null` and we might not have to generalize the cases further.

> it will failed when do select rand(null)

Also, I would like to add the edge cases here but I'd like to avoid PR is being hold. 

As not all the things have a standard to follow, we can define the behaviour here. I don't have access to Oracle and DB2. Do you think Hive, PostgreSQL and MySQL examples are not enough?

gatorsmile:
At first, we do not strictly follow Hive. You can easily find many in Spark. I do not think this is an urgent JIRA, right? Like what @srowen replied in the JIRA, he does not think this is a bug. The existing output message looks reasonable to me too.  
```
Input argument to rand must be an integer literal.;; line 1 pos 0
```
Setting the seed as `null` also looks weird to me. 

DB2 and Oracle have free versions to download. You can easily install the docker versions. You also can google their documentation. What we need to do at first is to do an investigation to save the times of all the other reviewers; otherwise, they have to do it too.

HyukjinKwon:
Not urgent but in my experience such PR have been being hold. So, I am trying to fix the problem specified in the JIRA only rather than fixing others together.
 
@srowen said "I'm not even sure that's a bug.." but "... reasonable to try to follow it.". At least, all the implementations of DB2, MySQL and PostgreSQL do not throw an exception but it defines its own behaviour.

gatorsmile:
What is the behavior of `PostgreSQL`? Treating `NULL` as zero?

HyukjinKwon:
Ah, PostgreSQL seems unsetting the seed.

```
postgres=# SELECT setseed(0), random(), random();
 setseed |      random       |      random
---------+-------------------+-------------------
         | 0.840187716763467 | 0.394382926635444
(1 row)

postgres=# SELECT setseed(0), random(), random();
 setseed |      random       |      random
---------+-------------------+-------------------
         | 0.840187716763467 | 0.394382926635444
(1 row)

postgres=# SELECT setseed(null), random(), random();
 setseed |      random       |      random
---------+-------------------+-------------------
         | 0.783099223393947 | 0.798440033104271
(1 row)

postgres=# SELECT setseed(null), random(), random();
 setseed |      random       |      random
---------+-------------------+-------------------
         | 0.911647357512265 | 0.197551369201392
(1 row)
```

gatorsmile:
Now, at least we have four options, when users setting `NULL` as a seed number for `rand`:
1. Hive/MySQL - `NULL` is equivalent to `0`
2. DB2 - when the seed is `NULL`, `rand` returns `NULL`
3. PostgreSQL - when the seed is `NULL`, ignore it. 
4. SparkSQL - does not allow it.

I do not have a strong opinion. Maybe @rxin need to make a decision.

HyukjinKwon:
Oh, strictly, it does not ignore in case of `PostgreSQL`. It unsets. 

```sql
postgres=# SELECT setseed(0);
 setseed
---------

(1 row)

postgres=# SELECT random();
      random
-------------------
 0.840187716763467
(1 row)

postgres=# SELECT random();
      random
-------------------
 0.394382926635444
(1 row)

postgres=# SELECT setseed(null);
 setseed
---------

(1 row)

postgres=# SELECT random();
      random
-------------------
 0.783099223393947
(1 row)

postgres=# SELECT random();
      random
-------------------
 0.798440033104271
(1 row)

postgres=# SELECT setseed(0);
 setseed
---------

(1 row)

postgres=# SELECT random();
      random
-------------------
 0.840187716763467
(1 row)

postgres=# SELECT random();
      random
-------------------
 0.394382926635444
(1 row)
```

rxin:
Given different databases diverge (they don't even have the same function names), I think it's fine to just have null be treated as 0 like Hive/MySQL.

gatorsmile:
@HyukjinKwon Could you document the behavior in the description of `rand` function? Checked whether we have any missing test case? Not sure whether you are also can check whether `rand` in R and Python can also correctly handle `NULL`. Thank you!

HyukjinKwon:
@gatorsmile Thanks. I intended to not describe this behaviour first but just thought it'd be kind of handling exceptional cases and it's discouraged because Hive, MySQL and PostgreSQL at least do not describe this behaviour. I will document this if you strongly feel it should be.

Yes, I can test Python and R but would be more specific about what  shouldI test please? This would affect only SQL expression not the actual functions. Do you expect the tests as below and do you usually check?

```r
df <- sql("SELECT RAND(NULL)")
```

or 

```python
spark.sql("SELECT RAND(NULL)")
df.selectExpr("RAND(NULL)")
```

For any missing case, I will check it out. Can I assume this way looks okay then?

gatorsmile:
Like what I mentioned above, the current change in this PR also allows `rand` to take an expression that returns a value of a built-in integer/long data type. Thus, you also need to test them too, right?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/randomExpressions.scala
```diff
@@ -77,18 +76,18 @@ case class Rand(seed: Long) extends RDG {
   }
 }
 
+object Rand {
+  def apply(seed: Long): Rand = Rand(Literal(seed))
+}
+
 /** Generate a random column with i.i.d. gaussian random distribution. */
 @ExpressionDescription(
   usage = "_FUNC_(a) - Returns a random column with i.i.d. gaussian random distribution.")
```
gatorsmile: Same here.

HyukjinKwon:
I didn't understand. I would like to be very sure on this to avoid mistakes in my future PRs.

> the current change in this PR also allows `rand` to take an expression that returns a value of a built-in integer/long data type. Thus, you also need to test them too, right?

This PR allows `rand` to take an expression but does not affect any other related language specific functions. Those functions are even already being tested.

> Not sure whether you are also can check whether `rand` in R and Python can also correctly handle `NULL`

Assuming from your comment, I can't think of more cases other than the ones as below (like I said above).

**SparkR**

```r
df <- sql("SELECT RAND(NULL)")
```

**PySpark** 

```python
spark.sql("SELECT RAND(NULL)")
df.selectExpr("RAND(NULL)")
```

I see the tests only check if the string is gone well via JVM or not.

IIUC, it is handled in the exact same path with Scala/R/Python within JVM in this case.

As far as I know, we haven't added/tested such cases in PySpark and SparkR In history unless it affects each related language specific functions.

Am I missing something?

gatorsmile:
Here, I am talking about the black box testing. If you add the new capability to any external function, you should add it in the test cases. This is very fundamental when we developing enterprise-level software. Note, Spark is an infrastructure software, not an application software.

The quality of Spark depends on the test case coverage. The code base of Spark SQL is rapidly increasing. We keep refactoring it. Even if your code does not have any bug, it does not mean the future code changes will not break your logics. Thus, we should do our best to improve the test case coverage in each PR.

gatorsmile:
One more suggestion to your future PR. Whenever you submit a PR, please try to improve the test case coverage. This can help you find bugs in your codes and also benefit the whole community.

gatorsmile:
Regarding the test cases of R and PythonR, I am fine if you do not add them into the code base. However, please at least run them manually. We hit many surprise bugs in the past just because nobody even tried it.

HyukjinKwon:
I see. That makes perfect sense in general. It is good to improve test coverage. Though, we might have to avoid adding test cases here and there when there are obviously no side-effect. Running tests already takes few hours.

At least, I am trying to submit a PR as sensible as I can, considering all the comments I received before, re-reading documentation and referring other good PRs. I left some comments above because it seems usually not or not like what I was told before (at least in my experience).

Maybe we should mention it somewhere like wiki and make this official because I guess others might have a very different view for this. We may say, "please at least run them manually and then include the results in the PR description".

HyukjinKwon:
Added `extend` is printed as below:

```sql
spark-sql> DESCRIBE FUNCTION EXTENDED rand;
Function: rand
Class: org.apache.spark.sql.catalyst.expressions.Rand
Usage: rand(a) - Returns a random column with i.i.d. uniformly distributed values in [0, 1].
Extended Usage:
> SELECT rand();
 0.9629742951434543
> SELECT rand(0);
 0.8446490682263027
> SELECT rand(NULL);
 0.8446490682263027
Time taken: 0.022 seconds, Fetched 4 row(s)
```
```sql
spark-sql> DESCRIBE FUNCTION EXTENDED randn;
Function: randn
Class: org.apache.spark.sql.catalyst.expressions.Randn
Usage: randn(a) - Returns a random column with i.i.d. gaussian random distribution.
Extended Usage:
> SELECT randn();
 -0.3254147983080288
> SELECT randn(0);
 1.1164209726833079
> SELECT randn(NULL);
 1.1164209726833079
Time taken: 0.02 seconds, Fetched 4 row(s)
```

srowen:
@gatorsmile @HyukjinKwon As a general comment Spark SQL doesn't claim a particular level of ANSI SQL compatibility. If anything it tries to match "whatever Hive does" and that's probably the best goal. 

It's always interesting to see what other RDBMSes do, especially where Hive's behavior seems ambiguous, but I would not describe this as required research. "What Hive does" is the important question in most cases.

This one is funny because Hive accepts behavior I wouldn't expect, even after reading its docs. I'm neutral on changing it if nobody is suggesting it's actually a problem in practice. I'm for changing it to match Hive if @rxin slightly favors it.

Documenting the behavior... sounds good. I suppose that commits to the behavior, but I can't see that Hive would change and we'd want to follow that change.

For a change that's so clearly inside the SQL engine and not language-specific, I personally wouldn't imagine we need language-specific tests. The new tests seem to cover the possibilities from a SQL perspective. I don't think we're here testing whether R/Python correctly serialize "null". This is also, keep in mind, a corner case of the behavior to begin with.

HyukjinKwon:
Documentation is printed as below now:

```sql
spark-sql> DESCRIBE FUNCTION EXTENDED RAND;
Function: rand
Class: org.apache.spark.sql.catalyst.expressions.Rand
Usage:
      rand() - Returns a random column with i.i.d. uniformly distributed values in [0, 1].
        seed is given randomly.

      rand(seed) - Returns a random column with i.i.d. uniformly distributed values in [0, 1].
        seed should be an integer/long/NULL literal.

Extended Usage:
> SELECT rand();
 0.9629742951434543
> SELECT rand(0);
 0.8446490682263027
> SELECT rand(NULL);
 0.8446490682263027
Time taken: 3.799 seconds, Fetched 4 row(s)
spark-sql> DESCRIBE FUNCTION EXTENDED RANDN;
Function: randn
Class: org.apache.spark.sql.catalyst.expressions.Randn
Usage:
      randn() - Returns a random column with i.i.d. gaussian random distribution.
        seed is given randomly.

      randn(seed) - Returns a random column with i.i.d. gaussian random distribution.
        seed should be an integer/long/NULL literal.

Extended Usage:
> SELECT randn();
 -0.3254147983080288
> SELECT randn(0);
 1.1164209726833079
> SELECT randn(NULL);
 1.1164209726833079
Time taken: 0.073 seconds, Fetched 4 row(s)
```

gatorsmile:
@srowen Previously, I also assumed Spark SQL is following the behaviors defined by Hive. However, we already merged a few PRs in Spark SQL that do not follow Hive. Especially, Hive has so many [configuration flags](https://cwiki.apache.org/confluence/display/Hive/Configuration+Properties), but we do not want to add them for better usability.

In the long term, when enterprise customers start using Spark more and more, I expect they might not even care whether we are Hive compliant or not.

HyukjinKwon:
retest this please

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/randomExpressions.scala
```diff
@@ -50,22 +54,27 @@ abstract class RDG extends LeafExpression with Nondeterministic {
...
-  override protected def evalInternal(input: InternalRow): Double = rng.nextDouble()
+  usage =
+    """
+      _FUNC_() - Returns a random column with i.i.d. uniformly distributed values in [0, 1].
+        seed is given randomly.
+
+      _FUNC_(seed) - Returns a random column with i.i.d. uniformly distributed values in [0, 1].
+        seed should be an integer/long/NULL literal.
+    """,
+  extended = "> SELECT _FUNC_();\n 0.9629742951434543\n> SELECT _FUNC_(0);\n 0.8446490682263027\n> SELECT _FUNC_(NULL);\n 0.8446490682263027")
```
srowen: Given the discussion in the other PR, should the 'usage' message be shorter and push some of this into 'extended'? whatever would be consistent.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/randomExpressions.scala
```diff
@@ -77,18 +86,28 @@ case class Rand(seed: Long) extends RDG {
...
+
 /** Generate a random column with i.i.d. gaussian random distribution. */
+// scalastyle:off line.size.limit
 @ExpressionDescription(
-  usage = "_FUNC_(a) - Returns a random column with i.i.d. gaussian random distribution.")
-case class Randn(seed: Long) extends RDG {
-  override protected def evalInternal(input: InternalRow): Double = rng.nextGaussian()
+  usage =
+    """
+      _FUNC_() - Returns a random column with i.i.d. gaussian random distribution.
```
srowen: Nit: gaussian -> Gaussian. Well, really we need to specify that it has mean 0 and stdev 1. And that's called the standard normal distribution. And I assume the 'n' in 'randn' is for normal. So maybe better to say "Returns a random column with i.i.d. values drawn from the standard normal distribution" everywhere including the comment.

srowen (changes requested):
Only minor changes that I can see. @gatorsmile how do you feel about this?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/randomExpressions.scala
HyukjinKwon: Sure, I will make it consistent.

srowen:
I assume this will proceed after https://github.com/apache/spark/pull/15513

HyukjinKwon:
I definitely will.

HyukjinKwon:
I will rebase and clean up this tomorrow.

sql/core/src/test/scala/org/apache/spark/sql/DataFrameSuite.scala
```diff
@@ -1728,4 +1728,29 @@ class DataFrameSuite extends QueryTest with SharedSQLContext {
     val df = spark.createDataFrame(spark.sparkContext.makeRDD(rows), schema)
     assert(df.filter($"array1" === $"array2").count() == 1)
   }
+
+  test("SPARK-17854: rand/randn allows null and long as input seed") {
```
rxin: move this into sql query test suite

rxin (changes requested):

sql/core/src/test/scala/org/apache/spark/sql/DataFrameSuite.scala
HyukjinKwon: Sure!

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/randomExpressions.scala
```diff
@@ -97,17 +101,15 @@ case class Rand(seed: Long) extends RDG {
...
+      > SELECT _FUNC_(null);
+       1.1164209726833079
   """)
 // scalastyle:on line.size.limit
-case class Randn(seed: Long) extends RDG {
-  override protected def evalInternal(input: InternalRow): Double = rng.nextGaussian()
+case class Randn(child: Expression) extends RDG {
 
-  def this() = this(Utils.random.nextLong())
+  def this() = this(Literal(Utils.random.nextLong()))
```
gatorsmile: Why not specifying the data type here?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/randomExpressions.scala
```diff
@@ -87,6 +87,10 @@ case class Rand(seed: Long) extends RDG {
   }
 }
 
+object Rand {
+  def apply(seed: Long): Rand = Rand(Literal(seed))
```
gatorsmile: The same here?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/randomExpressions.scala
```diff
@@ -64,17 +66,15 @@ abstract class RDG extends LeafExpression with Nondeterministic {
...
+      > SELECT _FUNC_(null);
+       0.8446490682263027
   """)
 // scalastyle:on line.size.limit
-case class Rand(seed: Long) extends RDG {
-  override protected def evalInternal(input: InternalRow): Double = rng.nextDouble()
+case class Rand(child: Expression) extends RDG {
 
-  def this() = this(Utils.random.nextLong())
+  def this() = this(Literal(Utils.random.nextLong()))
```
gatorsmile: The same here?

gatorsmile:
LGTM except a few minor comments.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/randomExpressions.scala
HyukjinKwon: The complier seems complaining if we specify the return type in `def this`.

HyukjinKwon: Oh, do you mean the type for literal for example as below?

```scala
Literal(Utils.random.nextLong(), LongType)
```

If you think it is beneficial because it at least does not do the type dispatch once, will fix here. Also, I can sweep the usages in `functions.scala` in another PR.

HyukjinKwon: If you meant something like `Literal(Long: Utils.random.nextLong()))`, I guess both are fine assuming from the discussion in the PR 12452.

HyukjinKwon: Yes, makes sense. I will fix them here first.

HyukjinKwon:
Thanks @gatorsmile!

HyukjinKwon:
retest this please

HyukjinKwon:
retest this please

srowen:
Merged to master/2.1

PR #15433, opened 2016-10-11
[SPARK-17822][SPARKR] Use weak reference in JVMObjectTracker.objMap because it may leak JVM objects

techaddict:

Use weak reference in JVMObjectTracker.objMap because it may leak JVM objects
## How was this patch tested?

existing tests


shivaram:
@techaddict Thanks for the PR. The test failures seem related to this patch -- Could you check why they are failing ?

core/src/main/scala/org/apache/spark/api/r/RBackendHandler.scala
```diff
@@ -284,7 +286,7 @@ private[r] object JVMObjectTracker {
     objId
   }
 
-  def remove(id: String): Option[Object] = {
+  def remove(id: String) {
```
srowen: Although it's normal to return the object that was removed from map methods, if this isn't desired, I'd declare this as `: Unit` to be clear

core/src/main/scala/org/apache/spark/api/r/RBackendHandler.scala
```diff
@@ -263,18 +264,19 @@ private[r] object JVMObjectTracker {
 
   // TODO: This map should be thread-safe if we want to support multiple
   // connections at the same time
-  private[this] val objMap = new HashMap[String, Object]
+  private[this] val objMap: ConcurrentMap[String, Object] =
```
srowen: Not that it matters a lot, but do you need the reference type?

core/src/main/scala/org/apache/spark/api/r/RBackendHandler.scala
```diff
@@ -263,18 +264,19 @@ private[r] object JVMObjectTracker {
...
-    objMap(id)
+    objMap.get(id)
   }
 
   def get(id: String): Option[Object] = {
-    objMap.get(id)
+    Option(objMap.get(id))
   }
 
   def put(obj: Object): String = {
```
srowen: It kind of seems like this class is meant to be thread-safe, but this isn't. You could use `AtomicLong` and `getAndIncrement`?

srowen (changes requested):

core/src/main/scala/org/apache/spark/api/r/RBackendHandler.scala
srowen: Agree, this can just be changed a little more to also handle that issue. They end up being interrelated here.

techaddict:
@shivaram @srowen not sure why its failing, will try to fix this ASAP.

yhuai:
test this please

yhuai:
test this please

techaddict:
closing this since, its maybe not the right way to do this

PR #8197, opened 2015-08-14
[SPARK-9906] [ML] User guide for LogisticRegressionSummary

MechCoder:
User guide for LogisticRegression summaries

PR #7538, opened 2015-07-20
[SPARK-9112] [ML] Implement Stats for LogisticRegression

MechCoder:
I have added support for stats in LogisticRegression. The API is similar to that of LinearRegression with LogisticRegressionTrainingSummary and LogisticRegressionSummary

I have some queries and asked them inline.

PR #15435, opened 2016-10-11
[SPARK-17139][ML] Add model summary for MultinomialLogisticRegression

WeichenXu123:

Add 4 traits, using the following hierarchy:
LogisticRegressionSummary
LogisticRegressionTrainingSummary: LogisticRegressionSummary
BinaryLogisticRegressionSummary: LogisticRegressionSummary
BinaryLogisticRegressionTrainingSummary: LogisticRegressionTrainingSummary, BinaryLogisticRegressionSummary

and the public method such as `def summary` only return trait type listed above.

and then implement 4 concrete classes:
LogisticRegressionSummaryImpl (multiclass case)
LogisticRegressionTrainingSummaryImpl (multiclass case)
BinaryLogisticRegressionSummaryImpl (binary case).
BinaryLogisticRegressionTrainingSummaryImpl (binary case).

## How was this patch tested?

Existing tests & added tests.

### DIFF
--- a/mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala
+++ b/mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala
@@ -884,7 +884,7 @@ class LogisticRegression @Since("1.2.0") (
       numClasses, isMultinomial))
 
     val (summaryModel, probabilityColName, predictionColName) = model.findSummaryModel()
-    val logRegSummary = if (!isMultinomial) {
+    val logRegSummary = if (!isMultinomial || (isMultinomial && numClasses <= 2)) {
       new BinaryLogisticRegressionTrainingSummaryImpl(
         summaryModel.transform(dataset),
         probabilityColName,
--- a/mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala
+++ b/mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala
@@ -1456,9 +1456,11 @@ sealed trait LogisticRegressionSummary extends Serializable {
  */
 sealed trait LogisticRegressionTrainingSummary extends LogisticRegressionSummary {
 
+  @Since("1.5.0")
   /** objective function (scaled loss + regularization) at each iteration. */
   def objectiveHistory: Array[Double]
 
+  @Since("1.5.0")
   /** Number of training iterations. */
   def totalIterations: Int = objectiveHistory.length
 
@@ -1565,15 +1567,13 @@ sealed trait BinaryLogisticRegressionTrainingSummary extends BinaryLogisticRegre
  * @param featuresCol field in "predictions" which gives the features of each instance as a vector.
  * @param objectiveHistory objective function (scaled loss + regularization) at each iteration.
  */
-@Experimental
-@Since("2.2.0")
 private class LogisticRegressionTrainingSummaryImpl(
     override val predictions: DataFrame,
     override val probabilityCol: String,
     override val predictionCol: String,
     override val labelCol: String,
     override val featuresCol: String,
-    @Since("1.5.0") val objectiveHistory: Array[Double])
+    val objectiveHistory: Array[Double])
   extends LogisticRegressionSummaryImpl(
     predictions, probabilityCol, predictionCol, labelCol, featuresCol)
   with LogisticRegressionTrainingSummary
@@ -1590,14 +1590,12 @@ private class LogisticRegressionTrainingSummaryImpl(
  * @param labelCol field in "predictions" which gives the true label of each instance.
  * @param featuresCol field in "predictions" which gives the features of each instance as a vector.
  */
-@Experimental
-@Since("2.2.0")
 private class LogisticRegressionSummaryImpl(
-    @Since("2.2.0") @transient override val predictions: DataFrame,
-    @Since("2.2.0") override val probabilityCol: String,
-    @Since("2.2.0") override val predictionCol: String,
-    @Since("2.2.0") override val labelCol: String,
-    @Since("2.2.0") override val featuresCol: String)
+    @transient override val predictions: DataFrame,
+    override val probabilityCol: String,
+    override val predictionCol: String,
+    override val labelCol: String,
+    override val featuresCol: String)
   extends LogisticRegressionSummary
 
 /**
@@ -1613,15 +1611,13 @@ private class LogisticRegressionSummaryImpl(
  * @param featuresCol field in "predictions" which gives the features of each instance as a vector.
  * @param objectiveHistory objective function (scaled loss + regularization) at each iteration.
  */
-@Experimental
-@Since("2.2.0")
 private class BinaryLogisticRegressionTrainingSummaryImpl(
     override val predictions: DataFrame,
     override val probabilityCol: String,
     override val predictionCol: String,
     override val labelCol: String,
     override val featuresCol: String,
-    @Since("1.5.0") override val objectiveHistory: Array[Double])
+    override val objectiveHistory: Array[Double])
   extends BinaryLogisticRegressionSummaryImpl(
     predictions, probabilityCol, predictionCol, labelCol, featuresCol)
   with BinaryLogisticRegressionTrainingSummary
@@ -1638,14 +1634,12 @@ private class BinaryLogisticRegressionTrainingSummaryImpl(
  * @param labelCol field in "predictions" which gives the true label of each instance.
  * @param featuresCol field in "predictions" which gives the features of each instance as a vector.
  */
-@Experimental
-@Since("2.2.0")
 private class BinaryLogisticRegressionSummaryImpl(
-    @Since("2.2.0") @transient override val predictions: DataFrame,
-    @Since("2.2.0") override val probabilityCol: String,
-    @Since("2.2.0") override val predictionCol: String,
-    @Since("2.2.0") override val labelCol: String,
-    @Since("2.2.0") override val featuresCol: String)
+    @transient override val predictions: DataFrame,
+    override val probabilityCol: String,
+    override val predictionCol: String,
+    override val labelCol: String,
+    override val featuresCol: String)
   extends LogisticRegressionSummaryImpl(
     predictions, probabilityCol, predictionCol, labelCol, featuresCol)
   with BinaryLogisticRegressionSummary
--- a/mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala
+++ b/mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala
@@ -884,7 +884,7 @@ class LogisticRegression @Since("1.2.0") (
       numClasses, isMultinomial))
 
     val (summaryModel, probabilityColName, predictionColName) = model.findSummaryModel()
-    val logRegSummary = if (numClasses <= 2) {
+    val logRegSummary = if (!isMultinomial) {
       new BinaryLogisticRegressionTrainingSummaryImpl(
         summaryModel.transform(dataset),
         probabilityColName,
@@ -1384,45 +1384,50 @@ sealed trait LogisticRegressionSummary extends Serializable {
         .rdd.map { case Row(prediction: Double, label: Double) => (prediction, label) })
   }
 
-  /** Returns true positive rate for each label. */
+  /** Returns true positive rate for each label (category). */
   @Since("2.3.0")
   def truePositiveRateByLabel: Array[Double] = recallByLabel
 
-  /** Returns false positive rate for each label. */
+  /** Returns false positive rate for each label (category). */
   @Since("2.3.0")
   def falsePositiveRateByLabel: Array[Double] = {
     multiclassMetrics.labels.map(label => multiclassMetrics.falsePositiveRate(label))
   }
 
-  /** Returns precision for each label. */
+  /** Returns precision for each label (category). */
   @Since("2.3.0")
   def precisionByLabel: Array[Double] = {
     multiclassMetrics.labels.map(label => multiclassMetrics.precision(label))
   }
 
-  /** Returns recall for each label. */
+  /** Returns recall for each label (category). */
   @Since("2.3.0")
   def recallByLabel: Array[Double] = {
     multiclassMetrics.labels.map(label => multiclassMetrics.recall(label))
   }
 
-  /**
-   * Returns f-measure for each label.
-   */
+  /** Returns f-measure for each label (category). */
   @Since("2.3.0")
   def fMeasureByLabel(beta: Double): Array[Double] = {
     multiclassMetrics.labels.map(label => multiclassMetrics.fMeasure(label, beta))
   }
 
-  /** Returns f1-measure for each label. */
+  /** Returns f1-measure for each label (category). */
   @Since("2.3.0")
   def fMeasureByLabel: Array[Double] = fMeasureByLabel(1.0)
 
-  /** Returns accuracy. */
+  /**
+   * Returns accuracy.
+   * (equals to the total number of correctly classified instances
+   * out of the total number of instances.)
+   */
   @Since("2.3.0")
   def accuracy: Double = multiclassMetrics.accuracy
 
-  /** Returns weighted true positive rate. */
+  /**
+   * Returns weighted true positive rate.
+   * (equals to precision, recall and f-measure)
+   */
   @Since("2.3.0")
   def weightedTruePositiveRate: Double = weightedRecall
 
@@ -1430,7 +1435,10 @@ sealed trait LogisticRegressionSummary extends Serializable {
   @Since("2.3.0")
   def weightedFalsePositiveRate: Double = multiclassMetrics.weightedFalsePositiveRate
 
-  /** Returns weighted averaged recall. */
+  /**
+   * Returns weighted averaged recall.
+   * (equals to precision, recall and f-measure)
+   */
   @Since("2.3.0")
   def weightedRecall: Double = multiclassMetrics.weightedRecall
 
@@ -1438,9 +1446,7 @@ sealed trait LogisticRegressionSummary extends Serializable {
   @Since("2.3.0")
   def weightedPrecision: Double = multiclassMetrics.weightedPrecision
 
-  /**
-   * Returns weighted averaged f-measure.
-   */
+  /** Returns weighted averaged f-measure. */
   @Since("2.3.0")
   def weightedFMeasure(beta: Double): Double = multiclassMetrics.weightedFMeasure(beta)
 
@@ -1456,12 +1462,12 @@ sealed trait LogisticRegressionSummary extends Serializable {
  */
 sealed trait LogisticRegressionTrainingSummary extends LogisticRegressionSummary {
 
-  @Since("1.5.0")
   /** objective function (scaled loss + regularization) at each iteration. */
+  @Since("1.5.0")
   def objectiveHistory: Array[Double]
 
-  @Since("1.5.0")
   /** Number of training iterations. */
+  @Since("1.5.0")
   def totalIterations: Int = objectiveHistory.length
 
 }
--- a/mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala
+++ b/mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala
@@ -884,7 +884,7 @@ class LogisticRegression @Since("1.2.0") (
       numClasses, isMultinomial))
 
     val (summaryModel, probabilityColName, predictionColName) = model.findSummaryModel()
-    val logRegSummary = if (numClasses <=2) {
+    val logRegSummary = if (numClasses <= 2) {
       new BinaryLogisticRegressionTrainingSummaryImpl(
         summaryModel.transform(dataset),
         probabilityColName,
@@ -1017,15 +1017,19 @@ class LogisticRegressionModel private[spark] (
   private var trainingSummary: Option[LogisticRegressionTrainingSummary] = None
 
   /**
-   * Gets summary of model on training set. An exception is
-   * thrown if `trainingSummary == None`.
+   * Gets summary of model on training set. An exception is thrown
+   * if `trainingSummary == None`.
    */
   @Since("1.5.0")
   def summary: LogisticRegressionTrainingSummary = trainingSummary.getOrElse {
     throw new SparkException("No training summary available for this LogisticRegressionModel")
   }
 
-  @Since("2.2.0")
+  /**
+   * Gets summary of model on training set. An exception is thrown
+   * if `trainingSummary == None` or it is a multiclass model.
+   */
+  @Since("2.3.0")
   def binarySummary: BinaryLogisticRegressionTrainingSummary = summary match {
     case b: BinaryLogisticRegressionTrainingSummary => b
     case _ =>
@@ -1357,11 +1361,11 @@ sealed trait LogisticRegressionSummary extends Serializable {
   /**
    * Dataframe output by the model's `transform` method.
    */
-  @Since("2.3.0")
+  @Since("1.5.0")
   def predictions: DataFrame
 
   /** Field in "predictions" which gives the probability of each class as a vector. */
-  @Since("2.3.0")
+  @Since("1.5.0")
   def probabilityCol: String
 
   /** Field in "predictions" which gives the prediction of each class. */
@@ -1369,11 +1373,11 @@ sealed trait LogisticRegressionSummary extends Serializable {
   def predictionCol: String
 
   /** Field in "predictions" which gives the true label of each instance (if available). */
-  @Since("2.3.0")
+  @Since("1.5.0")
   def labelCol: String
 
   /** Field in "predictions" which gives the features of each instance as a vector. */
-  @Since("2.3.0")
+  @Since("1.6.0")
   def featuresCol: String
 
   @transient private val multiclassMetrics = {
@@ -1383,6 +1387,17 @@ sealed trait LogisticRegressionSummary extends Serializable {
         col(labelCol).cast(DoubleType))
         .rdd.map { case Row(prediction: Double, label: Double) => (prediction, label) })
   }
+
+  /**
+   * Returns the sequence of labels in ascending order
+   *
+   * Note: In most cases, it will be values {0.0, 1.0, ..., numClasses-1}, However, if the
+   * training set is missing a label, then all of the arrays over labels
+   * (e.g., from truePositiveRateByLabel) will be of length numClasses-1 instead of the
+   * expected numClasses.
+   */
+  @Since("2.3.0")
+  def labels: Array[Double] = multiclassMetrics.labels
 
   /** Returns true positive rate for each label (category). */
   @Since("2.3.0")
@@ -1561,7 +1576,6 @@ sealed trait BinaryLogisticRegressionTrainingSummary extends BinaryLogisticRegre
   with LogisticRegressionTrainingSummary
 
 /**
- * :: Experimental ::
  * Multiclass logistic regression training results.
  *
  * @param predictions dataframe output by the model's `transform` method.
@@ -1574,18 +1588,17 @@ sealed trait BinaryLogisticRegressionTrainingSummary extends BinaryLogisticRegre
  * @param objectiveHistory objective function (scaled loss + regularization) at each iteration.
  */
 private class LogisticRegressionTrainingSummaryImpl(
-    override val predictions: DataFrame,
-    override val probabilityCol: String,
-    override val predictionCol: String,
-    override val labelCol: String,
-    override val featuresCol: String,
-    val objectiveHistory: Array[Double])
+    predictions: DataFrame,
+    probabilityCol: String,
+    predictionCol: String,
+    labelCol: String,
+    featuresCol: String,
+    override val objectiveHistory: Array[Double])
   extends LogisticRegressionSummaryImpl(
     predictions, probabilityCol, predictionCol, labelCol, featuresCol)
   with LogisticRegressionTrainingSummary
 
 /**
- * :: Experimental ::
  * Multiclass logistic regression results for a given model.
  *
  * @param predictions dataframe output by the model's `transform` method.
@@ -1605,7 +1618,6 @@ private class LogisticRegressionSummaryImpl(
   extends LogisticRegressionSummary
 
 /**
- * :: Experimental ::
  * Binary logistic regression training results.
  *
  * @param predictions dataframe output by the model's `transform` method.
@@ -1618,18 +1630,17 @@ private class LogisticRegressionSummaryImpl(
  * @param objectiveHistory objective function (scaled loss + regularization) at each iteration.
  */
 private class BinaryLogisticRegressionTrainingSummaryImpl(
-    override val predictions: DataFrame,
-    override val probabilityCol: String,
-    override val predictionCol: String,
-    override val labelCol: String,
-    override val featuresCol: String,
+    predictions: DataFrame,
+    probabilityCol: String,
+    predictionCol: String,
+    labelCol: String,
+    featuresCol: String,
     override val objectiveHistory: Array[Double])
   extends BinaryLogisticRegressionSummaryImpl(
     predictions, probabilityCol, predictionCol, labelCol, featuresCol)
   with BinaryLogisticRegressionTrainingSummary
 
 /**
- * :: Experimental ::
  * Binary logistic regression results for a given model.
  *
  * @param predictions dataframe output by the model's `transform` method.
@@ -1641,11 +1652,11 @@ private class BinaryLogisticRegressionTrainingSummaryImpl(
  * @param featuresCol field in "predictions" which gives the features of each instance as a vector.
  */
 private class BinaryLogisticRegressionSummaryImpl(
-    @transient override val predictions: DataFrame,
-    override val probabilityCol: String,
-    override val predictionCol: String,
-    override val labelCol: String,
-    override val featuresCol: String)
+    predictions: DataFrame,
+    probabilityCol: String,
+    predictionCol: String,
+    labelCol: String,
+    featuresCol: String)
   extends LogisticRegressionSummaryImpl(
     predictions, probabilityCol, predictionCol, labelCol, featuresCol)
   with BinaryLogisticRegressionSummary
--- a/mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala
+++ b/mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala
@@ -1357,23 +1357,23 @@ sealed trait LogisticRegressionSummary extends Serializable {
   /**
    * Dataframe output by the model's `transform` method.
    */
-  @Since("2.2.0")
+  @Since("2.3.0")
   def predictions: DataFrame
 
   /** Field in "predictions" which gives the probability of each class as a vector. */
-  @Since("2.2.0")
+  @Since("2.3.0")
   def probabilityCol: String
 
   /** Field in "predictions" which gives the prediction of each class. */
-  @Since("2.2.0")
+  @Since("2.3.0")
   def predictionCol: String
 
   /** Field in "predictions" which gives the true label of each instance (if available). */
-  @Since("2.2.0")
+  @Since("2.3.0")
   def labelCol: String
 
   /** Field in "predictions" which gives the features of each instance as a vector. */
-  @Since("2.2.0")
+  @Since("2.3.0")
   def featuresCol: String
 
   @transient private val multiclassMetrics = {
@@ -1385,23 +1385,23 @@ sealed trait LogisticRegressionSummary extends Serializable {
   }
 
   /** Returns true positive rate for each label. */
-  @Since("2.2.0")
+  @Since("2.3.0")
   def truePositiveRateByLabel: Array[Double] = recallByLabel
 
   /** Returns false positive rate for each label. */
-  @Since("2.2.0")
+  @Since("2.3.0")
   def falsePositiveRateByLabel: Array[Double] = {
     multiclassMetrics.labels.map(label => multiclassMetrics.falsePositiveRate(label))
   }
 
   /** Returns precision for each label. */
-  @Since("2.2.0")
+  @Since("2.3.0")
   def precisionByLabel: Array[Double] = {
     multiclassMetrics.labels.map(label => multiclassMetrics.precision(label))
   }
 
   /** Returns recall for each label. */
-  @Since("2.2.0")
+  @Since("2.3.0")
   def recallByLabel: Array[Double] = {
     multiclassMetrics.labels.map(label => multiclassMetrics.recall(label))
   }
@@ -1409,43 +1409,43 @@ sealed trait LogisticRegressionSummary extends Serializable {
   /**
    * Returns f-measure for each label.
    */
-  @Since("2.2.0")
+  @Since("2.3.0")
   def fMeasureByLabel(beta: Double): Array[Double] = {
     multiclassMetrics.labels.map(label => multiclassMetrics.fMeasure(label, beta))
   }
 
   /** Returns f1-measure for each label. */
-  @Since("2.2.0")
+  @Since("2.3.0")
   def fMeasureByLabel: Array[Double] = fMeasureByLabel(1.0)
 
   /** Returns accuracy. */
-  @Since("2.2.0")
+  @Since("2.3.0")
   def accuracy: Double = multiclassMetrics.accuracy
 
   /** Returns weighted true positive rate. */
-  @Since("2.2.0")
+  @Since("2.3.0")
   def weightedTruePositiveRate: Double = weightedRecall
 
   /** Returns weighted false positive rate. */
-  @Since("2.2.0")
+  @Since("2.3.0")
   def weightedFalsePositiveRate: Double = multiclassMetrics.weightedFalsePositiveRate
 
   /** Returns weighted averaged recall. */
-  @Since("2.2.0")
+  @Since("2.3.0")
   def weightedRecall: Double = multiclassMetrics.weightedRecall
 
   /** Returns weighted averaged precision. */
-  @Since("2.2.0")
+  @Since("2.3.0")
   def weightedPrecision: Double = multiclassMetrics.weightedPrecision
 
   /**
    * Returns weighted averaged f-measure.
    */
-  @Since("2.2.0")
+  @Since("2.3.0")
   def weightedFMeasure(beta: Double): Double = multiclassMetrics.weightedFMeasure(beta)
 
   /** Returns weighted averaged f1-measure. */
-  @Since("2.2.0")
+  @Since("2.3.0")
   def weightedFMeasure: Double = multiclassMetrics.weightedFMeasure(1.0)
 }
 
--- a/mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala
+++ b/mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala
@@ -1576,13 +1576,11 @@ private class LogisticRegressionTrainingSummaryImpl(
     @Since("1.5.0") val objectiveHistory: Array[Double])
   extends LogisticRegressionSummaryImpl(
     predictions, probabilityCol, predictionCol, labelCol, featuresCol)
-  with LogisticRegressionTrainingSummary {
-
-}
+  with LogisticRegressionTrainingSummary
 
 /**
  * :: Experimental ::
- * Multiclass Logistic regression results for a given model.
+ * Multiclass logistic regression results for a given model.
  *
  * @param predictions dataframe output by the model's `transform` method.
  * @param probabilityCol field in "predictions" which gives the probability of
--- a/mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala
+++ b/mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala
@@ -35,7 +35,7 @@ import org.apache.spark.ml.optim.loss.{L2Regularization, RDDLossFunction}
 import org.apache.spark.ml.param._
 import org.apache.spark.ml.param.shared._
 import org.apache.spark.ml.util._
-import org.apache.spark.mllib.evaluation.BinaryClassificationMetrics
+import org.apache.spark.mllib.evaluation.{BinaryClassificationMetrics, MulticlassMetrics}
 import org.apache.spark.mllib.linalg.VectorImplicits._
 import org.apache.spark.mllib.stat.MultivariateOnlineSummarizer
 import org.apache.spark.mllib.util.MLUtils
@@ -882,21 +882,28 @@ class LogisticRegression @Since("1.2.0") (
 
     val model = copyValues(new LogisticRegressionModel(uid, coefficientMatrix, interceptVector,
       numClasses, isMultinomial))
-    // TODO: implement summary model for multinomial case
-    val m = if (!isMultinomial) {
-      val (summaryModel, probabilityColName) = model.findSummaryModelAndProbabilityCol()
-      val logRegSummary = new BinaryLogisticRegressionTrainingSummary(
+
+    val (summaryModel, probabilityColName, predictionColName) = model.findSummaryModel()
+    val logRegSummary = if (numClasses <= 2) {
+      new BinaryLogisticRegressionTrainingSummaryImpl(
         summaryModel.transform(dataset),
         probabilityColName,
+        predictionColName,
         $(labelCol),
         $(featuresCol),
         objectiveHistory)
-      model.setSummary(Some(logRegSummary))
     } else {
-      model
+      new LogisticRegressionTrainingSummaryImpl(
+        summaryModel.transform(dataset),
+        probabilityColName,
+        predictionColName,
+        $(labelCol),
+        $(featuresCol),
+        objectiveHistory)
     }
-    instr.logSuccess(m)
-    m
+    model.setSummary(Some(logRegSummary))
+    instr.logSuccess(model)
+    model
   }
 
   @Since("1.4.0")
@@ -1018,19 +1025,33 @@ class LogisticRegressionModel private[spark] (
     throw new SparkException("No training summary available for this LogisticRegressionModel")
   }
 
+  @Since("2.2.0")
+  def binarySummary: BinaryLogisticRegressionTrainingSummary = summary match {
+    case b: BinaryLogisticRegressionTrainingSummary => b
+    case _ =>
+      throw new RuntimeException("Cannot create a binary summary for a non-binary model" +
+        s"(numClasses=${numClasses}), use summary instead.")
+  }
+
   /**
-   * If the probability column is set returns the current model and probability column,
-   * otherwise generates a new column and sets it as the probability column on a new copy
-   * of the current model.
+   * If the probability and prediction columns are set, this method returns the current model,
+   * otherwise it generates new columns for them and sets them as columns on a new copy of
+   * the current model
    */
-  private[classification] def findSummaryModelAndProbabilityCol():
-      (LogisticRegressionModel, String) = {
-    $(probabilityCol) match {
-      case "" =>
-        val probabilityColName = "probability_" + java.util.UUID.randomUUID.toString
-        (copy(ParamMap.empty).setProbabilityCol(probabilityColName), probabilityColName)
-      case p => (this, p)
+  private[classification] def findSummaryModel():
+      (LogisticRegressionModel, String, String) = {
+    val model = if ($(probabilityCol).isEmpty && $(predictionCol).isEmpty) {
+      copy(ParamMap.empty)
+        .setProbabilityCol("probability_" + java.util.UUID.randomUUID.toString)
+        .setPredictionCol("prediction_" + java.util.UUID.randomUUID.toString)
+    } else if ($(probabilityCol).isEmpty) {
+      copy(ParamMap.empty).setProbabilityCol("probability_" + java.util.UUID.randomUUID.toString)
+    } else if ($(predictionCol).isEmpty) {
+      copy(ParamMap.empty).setPredictionCol("prediction_" + java.util.UUID.randomUUID.toString)
+    } else {
+      this
     }
+    (model, model.getProbabilityCol, model.getPredictionCol)
   }
 
   private[classification]
@@ -1051,9 +1072,14 @@ class LogisticRegressionModel private[spark] (
   @Since("2.0.0")
   def evaluate(dataset: Dataset[_]): LogisticRegressionSummary = {
     // Handle possible missing or invalid prediction columns
-    val (summaryModel, probabilityColName) = findSummaryModelAndProbabilityCol()
-    new BinaryLogisticRegressionSummary(summaryModel.transform(dataset),
-      probabilityColName, $(labelCol), $(featuresCol))
+    val (summaryModel, probabilityColName, predictionColName) = findSummaryModel()
+    if (numClasses > 2) {
+      new LogisticRegressionSummaryImpl(summaryModel.transform(dataset),
+        probabilityColName, predictionColName, $(labelCol), $(featuresCol))
+    } else {
+      new BinaryLogisticRegressionSummaryImpl(summaryModel.transform(dataset),
+        probabilityColName, predictionColName, $(labelCol), $(featuresCol))
+    }
   }
 
   /**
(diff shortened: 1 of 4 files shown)


WeichenXu123:
@sethah This pr seems to be discussed about several details, I am pleasure to hear your opinion, thanks!

sethah:
I'll try to take a look before too long. For now, I see there are no tests, could you please add tests, using the summary tests for binary classification as a guide? Thanks!

sethah:
So I've been reading through some of the history with logistic regression summaries. There was a lot of discussion on how to design the abstractions for this, [here](https://github.com/apache/spark/pull/7538) and [here](https://github.com/apache/spark/pull/8197).

I'm reposting some of the relevant snippets (I will comment on them in a follow up):

"We'll need to use traits to fix the multiple inheritance issue:"

````
sealed trait LogisticRegressionSummary
sealed trait LogisticRegressionTrainingSummary
class BinaryLogisticRegressionSummary extends LogisticRegressionSummary
class BinaryLogisticRegressionTrainingSummary extends BinaryLogisticRegressionSummary with LogisticRegressionTrainingSummary
````

"Are we planning to have a MulticlassLogisticRegressionSummary inheriting from LogisticRegressionSummary in the future because without that I'm unable to understand how using a trait would help since there is no access to the predictions dataframe."

"Yes, MulticlassLogisticRegressionSummary should be analogous to the binary version, with both inheriting from LogisticRegressionSummary."

...

"Synced with @jkbradley offline. Summary:

We should not require end users to perform any sort of downcasting in the stabilized API. This is OK for now since the API is still experimental.

Eventually we could provide two methods, a summary : LogisticRegressionSummary and a binarySummary : BInaryLogisticRegressionSummary which errors when called on a multiclass LRModel. This will be easy to implement because summary is returning the base LogisticRegressionSummary class so will not require any public API change."

**Comment by sethah** [comment]
So, based on my interpretation of this and how this can actually work, we need to have:

````scala
sealed trait LogisticRegressionSummary
sealed trait LogisticRegressionTrainingSummary
class MulticlassLogisticRegressionSummary extends LogisticRegressionSummary
class MulticlassLogisticRegressionTrainingSummary extends MulticlassLogisticRegressionSummary with LogisticRegressionTrainingSummary
class BinaryLogisticRegressionSummary extends MulticlassLogisticRegressionSummary
class BinaryLogisticRegressionTrainingSummary extends BinaryLogisticRegressionSummary with LogisticRegressionTrainingSummary
````

Then, in `LogisticRegressionModel` we have:

````scala
def summary: LogisticRegressionTrainingSummary
def binarySummary: BinaryLogisticRegressionTrainingSummary = summary match {
  case b: BinaryLogisticRegressionTrainingSummary => b
  case _ => throw new Exception()
}
````

And we avoid downcasting in the summary case since `MulticlassLogisticRegressionSummary` only implements the methods defined in the trait. Otherwise, we would have to downcast to get access to those methods. Then if the summary is binary, you can just call binary summary. Anyway, I got this to compile, and if there is some other way, I'm not seeing it. Would really like to get some clarification from @jkbradley. Not sure if @feynmanliang is still involved with Spark.

**Comment by WeichenXu123** [comment]
@sethah Good suggestion. code updated, thanks!

File: mllib/src/test/scala/org/apache/spark/ml/classification/LogisticRegressionSuite.scala [hunk 1]
```diff
@@ -1756,55 +1765,105 @@ class LogisticRegressionSuite
...
-    val model = lr.fit(smallBinaryDataset)
-    val summary = model.evaluate(smallBinaryDataset).asInstanceOf[BinaryLogisticRegressionSummary]
+    val blorModel = blor.fit(smallBinaryDataset)
+    val blorSummary = blorModel.evaluate(smallBinaryDataset)
+      .asInstanceOf[BinaryLogisticRegressionSummary]
 
-    val longLabelData = smallBinaryDataset.select(col(model.getLabelCol).cast(LongType),
-      col(model.getFeaturesCol))
-    val longSummary = model.evaluate(longLabelData).asInstanceOf[BinaryLogisticRegressionSummary]
+    val blorLongLabelData = smallBinaryDataset.select(col(blorModel.getLabelCol).cast(LongType),
```
zhengruifeng: Non-double numeric datatypes other than LongType maybe also needed to test. Any thoughts? @sethah

WeichenXu123: @zhengruifeng Thanks for carefule review, because in the Summary class code we use `labelValue.toDouble` to do the cast so that we use one type `LongType` as the test is OK IMO, is there some possible case that `LongType` cast succeed but other legal numeric value will fails?

sethah:
re-ping @jkbradley Would be great to get your thoughts on the above discussion since you were involved in the original design

WeichenXu123:
@sethah jkbradley seems not online recently we can invite @yanboliang to give some advice.

jkbradley:
@sethah Thanks for pinging, and for bringing up those old design discussions (which are still the best options I know of).  @WeichenXu123 sorry I've been slow to respond; I'm trying to keep up!

I'll take a look at the PR.

mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala
```diff
@@ -777,6 +787,20 @@ class LogisticRegressionModel private[spark] (
     throw new SparkException("No training summary available for this LogisticRegressionModel")
   }
 
+  @Since("2.1.0")
+  def binarySummary: BinaryLogisticRegressionTrainingSummary = summary match {
+    case b: BinaryLogisticRegressionTrainingSummary => b
+    case _ => throw new SparkException("Summary type dismatch, " +
```
jkbradley: Also, the current error doesn't really help the user.  How about "Cannot create a binarySummary for a non-binary model (with numClasses = $numClasses).  Use multinomialSummary instead."

mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala
```diff
@@ -777,6 +787,20 @@ class LogisticRegressionModel private[spark] (
...
+      "BinaryLogisticRegressionTrainingSummary expected.")
+  }
+
+  @Since("2.1.0")
+  def multinomialSummary: MultinomialLogisticRegressionTrainingSummary = summary match {
+    case m: MultinomialLogisticRegressionTrainingSummary => m
+    case _ => throw new SparkException("Summary type dismatch, " +
```
jkbradley: We actually could return a valid multinomial summary, even for binary problems.  I'd recommend we do that.

mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala
```diff
@@ -919,6 +948,10 @@ class LogisticRegressionModel private[spark] (
     }
   }
 
+  private[classification] val probability2predictionUdf = udf {
```
jkbradley: This UDF might pull this entire class into the closure and thus into the summaries.  I believe we can eliminate this UDF; I'll comment more below.

mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala
```diff
@@ -1141,6 +1174,21 @@ class BinaryLogisticRegressionTrainingSummary private[classification] (
 
 }
 
+@Experimental
```
jkbradley: Add scala doc

mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala
```diff
@@ -1243,6 +1291,73 @@ class BinaryLogisticRegressionSummary private[classification] (
...
+    @Since("1.6.0") override val featuresCol: String,
+    private val probability2PredictionUdf: UserDefinedFunction) extends LogisticRegressionSummary {
+
+  private val sparkSession = predictions.sparkSession
+  import sparkSession.implicits._
+
+  @transient private val multinomialMetrics = {
+    new MulticlassMetrics(
+      predictions.select(
+        probability2PredictionUdf(col(probabilityCol)),
```
jkbradley: This is the only place where you need the probability2predictionUDF.  Rather than passing a UDF or the model, I'd like us to modify findSummaryModelAndProbabilityCol to also set predictionCol if needed.

mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala
```diff
@@ -1243,6 +1291,73 @@ class BinaryLogisticRegressionSummary private[classification] (
   }
 }
 
+@Experimental
+@Since("2.1.0")
+class MultinomialLogisticRegressionSummary private[classification] (
+    @Since("1.5.0") @transient override val predictions: DataFrame,
```
jkbradley: These since tags should all be 2.1.0.

mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala
```diff
@@ -1243,6 +1291,73 @@ class BinaryLogisticRegressionSummary private[classification] (
...
+    @Since("2.1.0") override val probabilityCol: String,
+    @Since("1.5.0") override val labelCol: String,
+    @Since("1.6.0") override val featuresCol: String,
+    private val probability2PredictionUdf: UserDefinedFunction) extends LogisticRegressionSummary {
+
+  private val sparkSession = predictions.sparkSession
```
jkbradley: Do you need these 2 lines?

mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala
```diff
@@ -1243,6 +1291,73 @@ class BinaryLogisticRegressionSummary private[classification] (
...
+  @transient private val multinomialMetrics = {
+    new MulticlassMetrics(
+      predictions.select(
+        probability2PredictionUdf(col(probabilityCol)),
+        col(labelCol).cast(DoubleType))
+        .rdd.map { case Row(prediction: Double, label: Double) => (prediction, label) })
+  }
+
+  @Since("2.1.0")
+  @transient lazy val labels: Array[Double] = multinomialMetrics.labels
```
jkbradley: Add scala doc for all of these

mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala
```diff
@@ -1390,7 +1505,8 @@ class BinaryLogisticRegressionSummary private[classification] (
  *    $$
  * </blockquote></p>
  *
- * @param bcCoefficients The broadcast coefficients corresponding to the features.
+  *
+  * @param bcCoefficients The broadcast coefficients corresponding to the features.
```
jkbradley: fix indentation

sethah:
@jkbradley Thanks for your input. I'm happy to review this, but wanted to get clarification before proceeding. I can take a look in the next couple of days.

jkbradley (commented):
@sethah Sorry, didn't want to stomp on your review.  I had started reviewing, but I'll just go ahead and submit these partial review comments.  Thanks!
And thanks @WeichenXu123 for adding this.

sethah:
@jkbradley No problems at all, it's always better to have more reviewers.

WeichenXu123:
@sethah @jkbradley 

Now I am considering elemating the `probabilityToPredictionUDF`,
There are some fussy problems about the `predictionCol` and the summary class hierarchy....

currently, the base interface `LogisticRegressionSummary` do not have the member `predictionCol`, which is needed in `MLOR summary`.
if I add the `predictionCol` into the base interface `LogisticRegressionSummary` then `BinaryLogisticRegressionTrainingSummary` will also need to be modified, it seems break API compatibility, should we avoid such thing ?

And, whether should we make `BinaryLogisticRegressionTrainingSummary` be the subclass of MLOR summary ? I would like let @jkbradley decide it.

zhengruifeng:
It seems that many metrics in `MultinomialLogisticRegressionSummary` are generic to other classification algos. So what about create a new class `MultiClassificationSummary` and put it in a new file? Just like what @yanboliang does in https://github.com/apache/spark/pull/15555

jkbradley:
@WeichenXu123 About breaking APIs: I'm OK adding predictionCol to LogisticRegressionSummary, even though I agree it technically breaks a public API:
* The traits are OK to change since they are sealed.
* The subclasses inheriting from the traits are public, but they all have private constructors, so users cannot have extended them (except by using hacks).
So I don't see a way this could break a non-hack use case, but let me know if I'm not thinking of an edge case.

@zhengruifeng I'm OK refactoring to create a MulticlassClassificationSummary (or MulticlassSummary?) now or later.  Now does sound better.  Let's keep it ```private[ml]``` for now though.

WeichenXu123:
All right. I will create a new class `MulticlassClassificationSummary` it looks better.
And I will change `BinaryLogisticRegressionTrainingSummary` be the subclass of MLOR summary, it looks more reasonable. If there is some problem in such hierarchy let me know it. @jkbradley 

And thanks @zhengruifeng for good suggestion~

mllib/src/test/scala/org/apache/spark/ml/classification/LogisticRegressionSuite.scala
```diff
@@ -149,15 +149,34 @@ class LogisticRegressionSuite
     assert(copiedModel.hasSummary)
   }
 
-  test("empty probabilityCol") {
-    val lr = new LogisticRegression().setProbabilityCol("")
-    val model = lr.fit(smallBinaryDataset)
+  test("empty probabilityCol or predictionCol") {
+    var lr = new LogisticRegression().setProbabilityCol("")
```
sethah: I think we can get rid of the vars and shorten this up. Also, we should probably test both binary and multiclass. How does this look?:

````scala
val lr = new LogisticRegression().setMaxIter(1)
    val datasetFieldNames = smallBinaryDataset.schema.fieldNames.toSet
    def checkSummarySchema(model: LogisticRegressionModel, columns: Seq[String]): Unit = {
      val fieldNames = model.summary.predictions.schema.fieldNames
      assert(model.hasSummary)
      assert(datasetFieldNames.subsetOf(fieldNames.toSet))
      columns.foreach { c => assert(fieldNames.exists(_.startsWith(c))) }
    }
    Seq("binomial", "multinomial").foreach { family =>
      lr.setProbabilityCol("")
      val modelNoProb = lr.fit(smallBinaryDataset)
      checkSummarySchema(modelNoProb, Seq("probability_"))

      lr.setProbabilityCol("probability").setPredictionCol("")
      val modelNoPred = lr.fit(smallBinaryDataset)
      checkSummarySchema(modelNoPred, Seq("prediction_"))

      lr.setProbabilityCol("").setPredictionCol("")
      val modelNoPredNoProb = lr.fit(smallBinaryDataset)
      checkSummarySchema(modelNoPredNoProb, Seq("prediction_", "probability_"))
    }
````

File: mllib/src/test/scala/org/apache/spark/ml/classification/LogisticRegressionSuite.scala [hunk 12]
```diff
@@ -1762,51 +1781,101 @@ class LogisticRegressionSuite
...
-    assert(summary.recallByThreshold.collect() === sameSummary.recallByThreshold.collect())
+      blorSummary.fMeasureByThreshold.collect() === sameBlorSummary.fMeasureByThreshold.collect())
     assert(
-      summary.precisionByThreshold.collect() === sameSummary.precisionByThreshold.collect())
+      blorSummary.recallByThreshold.collect() === sameBlorSummary.recallByThreshold.collect())
+    assert(
+      blorSummary.precisionByThreshold.collect()
+        === sameBlorSummary.precisionByThreshold.collect())
+
+    val mlor = new LogisticRegression()
```
sethah: This is never used since you fit with `blor` below. We can probably make a single estimator called `lr` and then just reuse it for both tests.

mllib/src/test/scala/org/apache/spark/ml/classification/LogisticRegressionSuite.scala
```diff
@@ -1762,51 +1781,101 @@ class LogisticRegressionSuite
...
+    assert(mlorSummary.weightedRecall === mlorSameSummary.weightedRecall)
+    assert(mlorSummary.weightedFMeasure === mlorSameSummary.weightedFMeasure)
   }
 
   test("evaluate with labels that are not doubles") {
     // Evaluate a test set with Label that is a numeric type other than Double
-    val lr = new LogisticRegression()
+    val blor = new LogisticRegression()
       .setMaxIter(1)
       .setRegParam(1.0)
```
sethah: add `.setFamily("binomial")`

mllib/src/test/scala/org/apache/spark/ml/classification/LogisticRegressionSuite.scala
sethah: Well, from a completeness standpoint I agree that it's better to test all the types that it's intended to work for. However, since it's just calling `cast` under the hood, it does seem a bit redundant. I'm ok leaving it as is, but I don't feel strongly about it.

mllib/src/test/scala/org/apache/spark/ml/classification/LogisticRegressionSuite.scala
```diff
@@ -2006,7 +2075,7 @@ class LogisticRegressionSuite
         assert(prob === Vectors.dense(Array(0.0, 0.0, 0.0, 0.0, 1.0, 0.0)))
         assert(pred === 4.0)
     }
-    // TODO: check num iters is zero when it become available in the model
+    require(modelWithMetadata.summary.totalIterations == 0)
```
sethah: Also, add `assert(model.summary.totalIterations === 0` after the first of these three tests and add `modelZeroLabel.summary.totalIterations > 0` after the second of these tests.

sethah (commented):
Made a pass. We should really have this in 2.2, will you have time, @WeichenXu123?

sethah:
I still don't know what the plan _was_ for the downcasting issue, and it doesn't seem to me like there's really a good solution. `evaluate` returns a `LogisticRegressionSummary` so, we have to downcast that in the binary case regardless. We still have the issue of downcasting in the multiclass case if we ever have methods defined outside of the `LogisticRegressionSummary` trait.

As far as the inheritance goes, I think a binary summary can do anything that a multiclass summary can do, so I think it's good to have `BinaryLogisticRegressionSummary extends MulticlassLogisticRegressionSummary`. 

@jkbradley Is there any clarification you can provide?

WeichenXu123:
@sethah Oh, thanks very much! I will update code immediately. As I commit this PR I will work on it in the first priority. Happy new year!

WeichenXu123:
@sethah Thanks your careful review! code updated. And about the downcast issue, is there a better solution? I think it is difficult to solve ideally. cc @jkbradley  @yanboliang thanks!

sethah:
sorry for the delay, hope to get to it soon.

sethah:
So, looking at the design, I'm a bit concerned. Since we're adding summaries in several places around ML, I think we'd ideally design a hierarchy like we did for the estimators and models: 

* `Summary`
* `PredictionSummary extends Summary`
* `ClassificationSummary extends PredictionSummary`
* `ProbabilisticClassificationSummary extends ClassificationSummary`
* `MulticlassClassificationSummary extends ClassificationSummary`
* `BinaryClassificationSummary extends MulticlassClassificationSummary`

Then we could implement model summaries like: `LogisticRegressionTrainingSummary extends MulticlassClassificationSummary with ProbabilisticClassificationSummary with TrainingSummary`. However, I don't see a way to accomplish this without breaking API compatibility. Regarding what we have now, a couple things:

* Why are we creating a class `MulticlassClassificationSummary` but not also `BinaryClassificationSummary` ?
* Why is there a one-to-one overlap between `MulticlassClassificationSummary` and `LogisticRegressionSummary`, and `MulticlassLogisticRegressionSummary` inherits from them both?

Maybe @jkbradley can help shed some light on this and what was originally intended? Open to other thoughts...

WeichenXu123:
sethah 

About this issue:
Why is there a one-to-one overlap between MulticlassClassificationSummary and LogisticRegressionSummary, and MulticlassLogisticRegressionSummary inherits from them both?

If I merge the MulticlassLogisticRegressionSummary into LogisticRegressionSummary to remove the  one-to-one overlap between MulticlassClassificationSummary and LogisticRegressionSummary, it will cause **more API breaking**, because in this way it will make BinaryLogisticRegressionTrainingSummary cannot extends LogisticRegressionTrainingSummary and it will break some other public API such as `def summary`.
you can try to modify it and compile the code and will find this problem...
Maybe there is some better way but I haven't think out.

WeichenXu123:
@sethah 

If I merge the MulticlassLogisticRegressionSummary into LogisticRegressionSummary,
then, according to the hierarchy currently designed, it became:

class LogisticRegressionSummary extends MulticlassSummary with LogisticRegressionSummary
class LogisticRegressionTrainingSummary extends LogisticRegressionSummary with 

** Note that now LogisticRegressionTrainingSummary must become a class, not a trait, if merge the MulticlassLogisticRegressionSummary into LogisticRegressionSummary, it has to be class...**

Now consider the `BinaryLogisticRegressionSummary`:

class BinaryLogisticRegressionSummary extends LogisticRegressionSummary
class BinaryLogisticRegressionTrainingSummary extends BinaryLogisticRegressionSummary

** Now new problem occur: BinaryLogisticRegressionTrainingSummary cannot extend LogisticRegressionTrainingSummary, because `LogisticRegressionTrainingSummary` has changed into a class, not a trait... **

** BinaryLogisticRegressionTrainingSummary cannot extend LogisticRegressionTrainingSummary cause more API breaking, such as `def summary`...**

So these problems are troublesome... for causing so many API breaking...

WeichenXu123:
@sethah 

About your new design,

```
Summary
PredictionSummary extends Summary
ClassificationSummary extends PredictionSummary
ProbabilisticClassificationSummary extends ClassificationSummary
MulticlassClassificationSummary extends ClassificationSummary
BinaryClassificationSummary extends MulticlassClassificationSummary

Then we could implement model summaries like: LogisticRegressionTrainingSummary extends MulticlassClassificationSummary with ProbabilisticClassificationSummary with TrainingSummary

```
suppose the `def summary` return `LogisticRegressionTrainingSummary` type,
how to deal with `BinaryClassificationSummary` ?
In this hierarchy, `BinaryClassificationSummary` cannot inherit from `LogisticRegressionTrainingSummary`,
but `def summary` method must support `Binary` case...
so it meets the similar problem I mentioned above...

jkbradley:
Sorry for the delay!  This sounds like an involved discussion, so I put my thoughts on the JIRA.  Let me know what you think.

WeichenXu123:
I read  jkbradley's thoughts here, so I will modify this as following:

first we need 4 traits, using the following hierarchy:
LogisticRegressionSummary
LogisticRegressionTrainingSummary: LogisticRegressionSummary
BinaryLogisticRegressionSummary: LogisticRegressionSummary
BinaryLogisticRegressionTrainingSummary: LogisticRegressionTrainingSummary, BinaryLogisticRegressionSummary

and the public method such as `def summary` only return trait type listed above.

and then implement 4 concrete classes:
`LogisticRegressionSummaryImpl` (multiclass case) 
`LogisticRegressionTrainingSummaryImpl` (multiclass case)
`BinaryLogisticRegressionSummaryImpl` (binary case).
`BinaryLogisticRegressionTrainingSummaryImpl` (binary case).

Is that right ? @jkbradley @sethah

sethah:
Yes I think that's the gist of it, thanks a lot @WeichenXu123 !

WeichenXu123:
cc @sethah @jkbradley Thanks!

sethah:
@WeichenXu123 Thanks for the updates. I'm wondering why the `Impl` classes have the method implementations? I thought we should just put the implementations (e.g. `falsePositiveRateByLabel`, etc...) into the traits.

WeichenXu123:
@sethah I think `truePositiveRate` is equivalent to `recall` so I directly implement another one in the trait, but if you don't like this way I can modify it.
Is there anything else need update?
Thx!

sethah:
I mean that I thought we would put all the method implementations in the traits, so that the `Impl` classes only have to extend them and not implement anything themselves. So, the impl class would just be

````scala
class LogisticRegressionSummaryImpl private[classification](
    @Since("2.2.0") @transient override val predictions: DataFrame,
    @Since("2.2.0") override val probabilityCol: String,
    @Since("2.2.0") override val predictionCol: String,
    @Since("2.2.0") override val labelCol: String,
    @Since("2.2.0") override val featuresCol: String)
  extends LogisticRegressionSummary
````

without a class body.

**Comment by WeichenXu123** [comment]
@sethah OK no problem! I can move the method implementations into trait.

**Comment by jkbradley** [comment]
+1 for moving implementation to traits, as long as the public methods are still Java-friendly.  (Methods which are implemented in traits often can't be called from Java.)

**Comment by WeichenXu123** [comment]
jenkins, test please.

**Comment by WeichenXu123** [comment]
Done. cc @sethah @jkbradley thanks!

File: mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala [hunk 15]
```diff
@@ -644,21 +644,29 @@ class LogisticRegression @Since("1.2.0") (
...
     val model = copyValues(new LogisticRegressionModel(uid, coefficientMatrix, interceptVector,
       numClasses, isMultinomial))
-    // TODO: implement summary model for multinomial case
-    val m = if (!isMultinomial) {
-      val (summaryModel, probabilityColName) = model.findSummaryModelAndProbabilityCol()
-      val logRegSummary = new BinaryLogisticRegressionTrainingSummary(
+
+    val (summaryModel, probabilityColName, predictionColName)
+      = model.findSummaryModel()
+    val logRegSummary = if (!isMultinomial) {
```
sethah: We should change this to `if (numClasses <= 2)` since you can train a multinomial model on 2 classes, and we should still get a binary summary in that case. And of course, please add tests.

mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala
```diff
@@ -813,9 +835,16 @@ class LogisticRegressionModel private[spark] (
   @Since("2.0.0")
   def evaluate(dataset: Dataset[_]): LogisticRegressionSummary = {
     // Handle possible missing or invalid prediction columns
-    val (summaryModel, probabilityColName) = findSummaryModelAndProbabilityCol()
-    new BinaryLogisticRegressionSummary(summaryModel.transform(dataset),
-      probabilityColName, $(labelCol), $(featuresCol))
+    val (summaryModel, probabilityColName, predictionColName)
+      = findSummaryModel()
+    if (isMultinomial) {
```
sethah: same as above comment, check number of classes, not family

mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala
```diff
@@ -780,19 +788,33 @@ class LogisticRegressionModel private[spark] (
     throw new SparkException("No training summary available for this LogisticRegressionModel")
   }
 
+  @Since("2.2.0")
+  def binarySummary: BinaryLogisticRegressionTrainingSummary = summary match {
```
sethah: The behavior we are implementing is non-trivial. We need to add tests to ensure that everything happens as expected. This is an example:

````scala
test("binary and multiclass summary") {
    val lr = new LogisticRegression()
      .setFamily("binomial")

    val blorModel = lr.fit(smallBinaryDataset)
    assert(blorModel.summary.isInstanceOf[BinaryLogisticRegressionTrainingSummaryImpl])
    assert(blorModel.binarySummary.isInstanceOf[BinaryLogisticRegressionTrainingSummaryImpl])

    val mlorModel = lr.setFamily("multinomial").fit(smallMultinomialDataset)
    assert(mlorModel.summary.isInstanceOf[LogisticRegressionTrainingSummaryImpl])
    withClue("cannot get binary summary for multiclass model") {
      intercept[RuntimeException] {
        mlorModel.binarySummary
      }
    }

    val blorSummary = blorModel.evaluate(smallBinaryDataset)
    val mlorSummary = mlorModel.evaluate(smallMultinomialDataset)
    assert(blorSummary.isInstanceOf[BinaryLogisticRegressionSummaryImpl])
    assert(mlorSummary.isInstanceOf[LogisticRegressionSummaryImpl])
  }
````

File: mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala [hunk 18]
```diff
@@ -1242,11 +1312,122 @@ class BinaryLogisticRegressionSummary private[classification] (
...
+ * @param predictions dataframe output by the model's `transform` method.
+ * @param probabilityCol field in "predictions" which gives the probability of
+ *                       each class as a vector.
+ * @param predictionCol field in "predictions" which gives the prediction of
+ *                      each class as a vector.
+ * @param labelCol field in "predictions" which gives the true label of each instance.
+ * @param featuresCol field in "predictions" which gives the features of each instance as a vector.
+ */
+@Experimental
+@Since("1.5.0")
```
sethah: Please review all since tags for correctness.

mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala
```diff
@@ -1394,13 +1575,13 @@ class BinaryLogisticRegressionSummary private[classification] (
  *    $$
  * </blockquote>
  *
+ *
```
sethah: revert this one and the next one

mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala
```diff
@@ -813,9 +835,16 @@ class LogisticRegressionModel private[spark] (
   @Since("2.0.0")
   def evaluate(dataset: Dataset[_]): LogisticRegressionSummary = {
     // Handle possible missing or invalid prediction columns
-    val (summaryModel, probabilityColName) = findSummaryModelAndProbabilityCol()
-    new BinaryLogisticRegressionSummary(summaryModel.transform(dataset),
-      probabilityColName, $(labelCol), $(featuresCol))
+    val (summaryModel, probabilityColName, predictionColName)
```
sethah: put this on a single line. also, it happens above so do it there as well

mllib/src/test/scala/org/apache/spark/ml/classification/LogisticRegressionSuite.scala
```diff
@@ -1786,51 +1793,98 @@ class LogisticRegressionSuite
...
-    val model = lr.fit(smallBinaryDataset)
-    val summary = model.evaluate(smallBinaryDataset).asInstanceOf[BinaryLogisticRegressionSummary]
+    val blorModel = blor.fit(smallBinaryDataset)
+    val blorSummary = blorModel.evaluate(smallBinaryDataset)
+      .asInstanceOf[BinaryLogisticRegressionSummary]
+
+    val blorLongLabelData = smallBinaryDataset.select(col(blorModel.getLabelCol).cast(LongType),
+      col(blorModel.getFeaturesCol))
+    val blorLongSummary = blorModel.evaluate(blorLongLabelData)
+      .asInstanceOf[BinaryLogisticRegressionSummary]
```
sethah: This is quite annoying. I wonder if we should add a `asBinary` method to `LogisticRegressionSummary` that casts it or throws an error to make it easier. Probably to do in a separate PR anyway.

mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala
```diff
@@ -1086,83 +1115,124 @@ private[classification] class MultiClassSummarizer extends Serializable {
...
- * :: Experimental ::
- * Logistic regression training results.
- *
- * @param predictions dataframe output by the model's `transform` method.
- * @param probabilityCol field in "predictions" which gives the probability of
- *                       each class as a vector.
- * @param labelCol field in "predictions" which gives the true label of each instance.
- * @param featuresCol field in "predictions" which gives the features of each instance as a vector.
- * @param objectiveHistory objective function (scaled loss + regularization) at each iteration.
+ * Abstraction for multinomial Logistic Regression Training results.
```
sethah: don't use capital letters here and elsewhere, only at the beginning of a sentence.

mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala
```diff
@@ -780,19 +788,33 @@ class LogisticRegressionModel private[spark] (
...
+    case _ =>
+      throw new RuntimeException("Cannot create a binarySummary for a non-binary model" +
+        s"(numClasses=${numClasses}), use summary instead.")
+  }
+
   /**
-   * If the probability column is set returns the current model and probability column,
-   * otherwise generates a new column and sets it as the probability column on a new copy
-   * of the current model.
+   * If the probability column and prediction column is set returns the current model, otherwise
```
sethah: "If the probability and prediction columns are set, this method returns the current model, otherwise it generates new columns for them and sets them as columns on a new copy of the current model."

mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala
```diff
@@ -780,19 +788,33 @@ class LogisticRegressionModel private[spark] (
     throw new SparkException("No training summary available for this LogisticRegressionModel")
   }
 
+  @Since("2.2.0")
+  def binarySummary: BinaryLogisticRegressionTrainingSummary = summary match {
+    case b: BinaryLogisticRegressionTrainingSummary => b
+    case _ =>
+      throw new RuntimeException("Cannot create a binarySummary for a non-binary model" +
```
sethah: "binarySummary" -> "binary summary"

sethah (commented):
Made a pass, thanks for making those changes! When we implement things like this where the behavior varies depending on parameters of model fits, etc... we should be sure to "define" the behavior implicitly through unit tests. That helps make it clear what we expect to happen in different scenarios, and makes sure the code actually implements that behavior.

sethah:
@jkbradley @MLnick @yanboliang Is there any committer support for getting this into 2.2? We'd need to really make a blitz on it I think. I can help, but if we don't have a committer supporting it then we will just have to wait until 2.3.

WeichenXu123:
@sethah Don't worry, I will update code ASAP and @yanboliang will also help review it. Thanks!

yanboliang:
Yeah, I can support for getting this into 2.2 if others are busy. Thanks!

mllib/src/test/scala/org/apache/spark/ml/classification/LogisticRegressionSuite.scala
```diff
@@ -194,15 +207,9 @@ class LogisticRegressionSuite
...
-        lr2.fit(smallBinaryDataset,
-          lr2.thresholds -> Array(0.3, 0.7), lr2.threshold -> (expectedThreshold / 2.0))
-      }
-    }
-    withClue("fit with ParamMap should throw error if threshold, thresholds do not match.") {
-      intercept[IllegalArgumentException] {
-        val lr2model = lr2.fit(smallBinaryDataset,
-          lr2.thresholds -> Array(0.3, 0.7), lr2.threshold -> (expectedThreshold / 2.0))
-        lr2model.getThreshold
+        val lr3 = lr2.copy(
```
WeichenXu123: I will revert this. (previously has a bug which cause this test case fail but now fixed)

mllib/src/test/scala/org/apache/spark/ml/classification/LogisticRegressionSuite.scala
WeichenXu123: OK. I agree to do it in separate PR.

WeichenXu123:
@sethah Code updated.
cc @yanboliang Pls help boost this PR thanks!

WeichenXu123:
Jenkins, test this please

mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala
sethah: It doesn't appear you added any tests for this? Also, it is better for reviewers if you do not squash your commits since we can then see what has changed since the last pass was made.

sethah:
@WeichenXu123 I made a PR to your branch. Can you check it? I think you'll still need to update the Mima file. Also, this may not make 2.2, so then you'd have to update the since tags.

WeichenXu123:
@sethah Thanks! I have merged your updates and fix mima file conflicts.
@yanboliang has just come back from trip and will help review and merge it into 2.2 so don't worry about it!

mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala
```diff
@@ -1070,90 +1096,128 @@ private[classification] class MultiClassSummarizer extends Serializable {
...
-/**
- * Abstraction for Logistic Regression Results for a given model.
+ * Abstraction for logistic regression results for a given model.
  */
 sealed trait LogisticRegressionSummary extends Serializable {
 
   /**
    * Dataframe output by the model's `transform` method.
    */
+  @Since("2.2.0")
```
sethah: We have avoided putting since tags on methods inside traits. But since this is sealed and I don't think we intend to add to it, it should be ok

mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala
```diff
@@ -1231,6 +1295,109 @@ class BinaryLogisticRegressionSummary private[classification] (
...
+    override val probabilityCol: String,
+    override val predictionCol: String,
+    override val labelCol: String,
+    override val featuresCol: String,
+    @Since("1.5.0") val objectiveHistory: Array[Double])
+  extends LogisticRegressionSummaryImpl(
+    predictions, probabilityCol, predictionCol, labelCol, featuresCol)
+  with LogisticRegressionTrainingSummary {
+
+}
```
sethah: remove braces

mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala
```diff
@@ -1231,6 +1295,109 @@ class BinaryLogisticRegressionSummary private[classification] (
...
+
+/**
+ * :: Experimental ::
+ * Multiclass Logistic regression results for a given model.
```
sethah: "Multiclass Logistic" -> "Multiclass logistic"

sethah (commented):
A couple nits. Also please fix MIMA.  And can you please stop squashing your commits? :)

sethah:
ping @yanboliang @jkbradley This LGTM.

mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala
WeichenXu123: @sethah I have a small problem, if I do not squash commits, it seems cannot solve the conflicts...

WeichenXu123:
`update v7` fix previous `LogisticRegressionSuite` conflicts and `fix nits` commit for some nits update.

mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala
sethah: have you tried just rebasing? `git pull --rebase upstream master`?

WeichenXu123: yeah, I already rebase it. (If I do not squash the commit to fix conflicts , the rebase always fail, maybe there are some better way?)

sethah:
ping @jkbradley @srowen. Any hope/interest for 2.2? Probably too late, but wanted to check.

sethah:
btw @WeichenXu123 you just have to fix merge conflicts while rebasing. This is always possible. Squashing commits is rarely necessary and rarely good practice for an open PR IMO.

yanboliang:
I'll make a final pass tomorrow, and merge it if possible. Thanks for all your efforts.

mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala
```diff
@@ -1449,6 +1513,107 @@ class BinaryLogisticRegressionSummary private[classification] (
...
+ * @param probabilityCol field in "predictions" which gives the probability of
+ *                       each class as a vector.
+ * @param predictionCol field in "predictions" which gives the prediction for a data instance as a
+ *                      double.
+ * @param labelCol field in "predictions" which gives the true label of each instance.
+ * @param featuresCol field in "predictions" which gives the features of each instance as a vector.
+ * @param objectiveHistory objective function (scaled loss + regularization) at each iteration.
+ */
+@Experimental
+@Since("2.2.0")
```
yanboliang: Remove ```Experimental``` and ```Since``` tag for private class.

mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala
```diff
@@ -1449,6 +1513,107 @@ class BinaryLogisticRegressionSummary private[classification] (
...
+ * @param probabilityCol field in "predictions" which gives the probability of
+ *                       each class as a vector.
+ * @param predictionCol field in "predictions" which gives the prediction for a data instance as a
+ *                      double.
+ * @param labelCol field in "predictions" which gives the true label of each instance.
+ * @param featuresCol field in "predictions" which gives the features of each instance as a vector.
+ */
+@Experimental
+@Since("2.2.0")
+private class LogisticRegressionSummaryImpl(
```
yanboliang: Ditto.

mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala
```diff
@@ -982,19 +989,33 @@ class LogisticRegressionModel private[spark] (
     throw new SparkException("No training summary available for this LogisticRegressionModel")
   }
 
+  @Since("2.2.0")
+  def binarySummary: BinaryLogisticRegressionTrainingSummary = summary match {
```
yanboliang: Could I ask why we need a separate ```binarySummary```? If we get a binary classification model, the ```trainingSummary ``` should be the corresponding binary summary. Is that correct?

sethah: For convenience. Otherwise, users need to call `model.summary.asInstanceOf[BinaryLogisticRegressionTrainingSummary]` in order to access the extra, binary-specific methods.

WeichenXu123:
The 4 summary impl classes are all private, I remove all `since` tag in them and put the since tags in the traits. cc @yanboliang

WeichenXu123:
Jenkins test this please.

WeichenXu123:
Jenkins, test this please.

WeichenXu123:
jenkins test please

WeichenXu123:
jenkins test please

felixcheung:
please rebase to pick up the fix for R tests

WeichenXu123:
@felixcheung allready updated..

WeichenXu123:
jenkins test please

WeichenXu123:
jenkins test please

sethah:
ping! @jkbradley @yanboliang

yanboliang:
LGTM for me. @WeichenXu123 Could you bump up @since version to 2.3.0? I'd like to merge this to master only, since 2.2 is under voting now. Thanks for all reviews and comments.

MLnick:
From what I see here, this is not aiming to put in the more generic trait hierarchy mentioned in the [JIRA comment](https://issues.apache.org/jira/browse/SPARK-17139?focusedCommentId=15856612&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-15856612) and Seth's [gist](https://gist.github.com/sethah/83c57fd77385979579cb44f3d5730e67)?

Is the plan to merge this in and then update with the generic hierarchy in a later PR? Because #17586 would also build of the generic hierarchy as it's practically the same as binary LoR summary.

WeichenXu123:
@MLnick I think those hierarchy design can be put into later PR to avoid this PR too complex.. The things current PR considering are satisfying blor/mlor summary hierarchy and training summary interface and keep api compatibility, and @jkbradley and @yanboliang also agree with this design so that we can merge it first and other hierarchy improvement can be discussed later.

sethah:
ping?? @yanboliang @MLnick

WeichenXu123:
cc @jkbradley I think it's OK now.

WeichenXu123:
jenkins, test this please.

WeichenXu123:
cc @jkbradley @MrBago thanks!

jkbradley:
I'll do a review pass now, despite the merge conflicts

mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala
```diff
@@ -1324,90 +1350,130 @@ private[ml] class MultiClassSummarizer extends Serializable {
...
+  @Since("2.3.0")
+  def weightedTruePositiveRate: Double = weightedRecall
+
+  /** Returns weighted false positive rate. */
+  @Since("2.3.0")
+  def weightedFalsePositiveRate: Double = multiclassMetrics.weightedFalsePositiveRate
+
+  /** Returns weighted averaged recall. */
+  @Since("2.3.0")
+  def weightedRecall: Double = multiclassMetrics.weightedRecall
```
jkbradley: For all of these, can you please make sure to copy all of the info from the MulticlassMetrics Scala doc?  There is some info missing, and users cannot be expected to find it on their own.

mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala
```diff
@@ -1324,90 +1350,130 @@ private[ml] class MultiClassSummarizer extends Serializable {
...
-    predictions: DataFrame,
-    probabilityCol: String,
-    labelCol: String,
-    featuresCol: String,
-    @Since("1.5.0") val objectiveHistory: Array[Double])
-  extends BinaryLogisticRegressionSummary(predictions, probabilityCol, labelCol, featuresCol)
-  with LogisticRegressionTrainingSummary {
+sealed trait LogisticRegressionTrainingSummary extends LogisticRegressionSummary {
+
+  @Since("1.5.0")
```
jkbradley: style: put annotations below the Scala doc

jkbradley (commented):
I'm not done with the review yet but have to pause.  Feel free to update this, and I'll finish reviewing later.

mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala
```diff
@@ -882,21 +882,28 @@ class LogisticRegression @Since("1.2.0") (
 
     val model = copyValues(new LogisticRegressionModel(uid, coefficientMatrix, interceptVector,
       numClasses, isMultinomial))
-    // TODO: implement summary model for multinomial case
-    val m = if (!isMultinomial) {
-      val (summaryModel, probabilityColName) = model.findSummaryModelAndProbabilityCol()
-      val logRegSummary = new BinaryLogisticRegressionTrainingSummary(
+
+    val (summaryModel, probabilityColName, predictionColName) = model.findSummaryModel()
+    val logRegSummary = if (numClasses <= 2) {
```
WeichenXu123: @jkbradley The logic here isn't `if (!isMultinomial) ... else ...`
it is `if (!isMultinomial || (isMultinomial && numClasses <= 2)) ... else ...`
see testcase `test("check summary types for binary and multiclass")`

mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala
```diff
@@ -882,21 +882,28 @@ class LogisticRegression @Since("1.2.0") (
 
     val model = copyValues(new LogisticRegressionModel(uid, coefficientMatrix, interceptVector,
       numClasses, isMultinomial))
-    // TODO: implement summary model for multinomial case
-    val m = if (!isMultinomial) {
-      val (summaryModel, probabilityColName) = model.findSummaryModelAndProbabilityCol()
-      val logRegSummary = new BinaryLogisticRegressionTrainingSummary(
+
+    val (summaryModel, probabilityColName, predictionColName) = model.findSummaryModel()
+    val logRegSummary = if (!isMultinomial || (isMultinomial && numClasses <= 2)) {
```
jkbradley: Oh sorry!  I didn't think about this case where a model is specified as multinomial but has 2 classes.  What you had before is better: ```numClasses <=2``` (without using ```isMultinomial```).  Feel free to question my suggestions :P

mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala
```diff
@@ -1484,3 +1556,96 @@ class BinaryLogisticRegressionSummary private[classification] (
     binaryMetrics.recallByThreshold().toDF("threshold", "recall")
   }
 }
+
+sealed trait BinaryLogisticRegressionTrainingSummary extends BinaryLogisticRegressionSummary
+  with LogisticRegressionTrainingSummary
+
+/**
+ * :: Experimental ::
```
jkbradley: No need for Experimental tags on private classes

mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala
```diff
@@ -1484,3 +1556,96 @@ class BinaryLogisticRegressionSummary private[classification] (
...
+ * @param probabilityCol field in "predictions" which gives the probability of
+ *                       each class as a vector.
+ * @param predictionCol field in "predictions" which gives the prediction for a data instance as a
+ *                      double.
+ * @param labelCol field in "predictions" which gives the true label of each instance.
+ * @param featuresCol field in "predictions" which gives the features of each instance as a vector.
+ * @param objectiveHistory objective function (scaled loss + regularization) at each iteration.
+ */
+private class LogisticRegressionTrainingSummaryImpl(
+    override val predictions: DataFrame,
```
jkbradley: I think you can leave out the "override val" since the value is passed to LogisticRegressionSummaryImpl, which provides it as a val.  Same for the other vals provided by LogisticRegressionSummaryImpl.

jkbradley: Same for BinaryLogisticRegressionTrainingSummaryImpl

mllib/src/test/scala/org/apache/spark/ml/classification/LogisticRegressionSuite.scala
```diff
@@ -199,15 +199,57 @@ class LogisticRegressionSuite
...
+    val datasetFieldNames = smallBinaryDataset.schema.fieldNames.toSet
+    def checkSummarySchema(model: LogisticRegressionModel, columns: Seq[String]): Unit = {
+      val fieldNames = model.summary.predictions.schema.fieldNames
+      assert(model.hasSummary)
+      assert(datasetFieldNames.subsetOf(fieldNames.toSet))
+      columns.foreach { c => assert(fieldNames.exists(_.startsWith(c))) }
+    }
+    // check that the summary model adds the appropriate columns
+    Seq(("binomial", smallBinaryDataset), ("multinomial", smallMultinomialDataset)).foreach {
+      case (family, dataset) =>
```
jkbradley: You are not using "dataset" in this.  However, this logic should be the same for both binomial and multinomial families, so I'm OK if you just test the binomial case here.

mllib/src/test/scala/org/apache/spark/ml/classification/LogisticRegressionSuite.scala
```diff
@@ -199,15 +199,57 @@ class LogisticRegressionSuite
...
+        checkSummarySchema(modelNoPred, Seq("prediction_"))
+
+        lr.setProbabilityCol("").setPredictionCol("")
+        val modelNoPredNoProb = lr.fit(smallBinaryDataset)
+        checkSummarySchema(modelNoPredNoProb, Seq("prediction_", "probability_"))
+    }
+  }
+
+  test("check summary types for binary and multiclass") {
+    val lr = new LogisticRegression()
```
jkbradley: Set maxIter = 1 for speed

mllib/src/test/scala/org/apache/spark/ml/classification/LogisticRegressionSuite.scala
```diff
@@ -199,15 +199,57 @@ class LogisticRegressionSuite
...
+      .setFamily("binomial")
+
+    val blorModel = lr.fit(smallBinaryDataset)
+    assert(blorModel.summary.isInstanceOf[BinaryLogisticRegressionTrainingSummaryImpl])
```
jkbradley: nit: You can test for the public trait BinaryLogisticRegressionTrainingSummary instead of the private implementation class.

jkbradley (commented):
OK, done with a full review pass!  My comments are pretty much all small items.

mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala
```diff
@@ -1324,90 +1350,136 @@ private[ml] class MultiClassSummarizer extends Serializable {
...
-/**
- * Abstraction for Logistic Regression Results for a given model.
+ * Abstraction for logistic regression results for a given model.
  */
 sealed trait LogisticRegressionSummary extends Serializable {
 
   /**
    * Dataframe output by the model's `transform` method.
    */
+  @Since("2.3.0")
```
jkbradley: I confirmed: These annotations are inherited, so users may be confused about why some Since annotations have changed.  E.g., BinaryLogisticRegressionSummary.featuresCol will change from 1.5.0 to 2.3.0.  I'd just remove the ones in the trait.

WeichenXu123: @jkbradley 
I think about this, currently the concrete impl class are marked as private, so it seems add @Since in `trait` is better (this is discussed with @sethah before) So what do you think about  this:
```
sealed trait LogisticRegressionSummary extends Serializable {

  /**
   * Dataframe output by the model's `transform` method.
   */
  @Since("1.5.0")
  def predictions: DataFrame

  /** Field in "predictions" which gives the probability of each class as a vector. */
  @Since("1.5.0")
  def probabilityCol: String

  /** Field in "predictions" which gives the prediction of each class. */
  @Since("2.3.0")
  def predictionCol: String

  /** Field in "predictions" which gives the true label of each instance (if available). */
  @Since("1.5.0")
  def labelCol: String

  /** Field in "predictions" which gives the features of each instance as a vector. */
  @Since("1.6.0")
  def featuresCol: String
  ...
}
```
I also check current master code and `featuresCol` is marked "1.6.0", others marked "1.5.0", and "predictionCol" is new added so mark it "2.3.0"

mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala
```diff
@@ -1357,23 +1361,23 @@ sealed trait LogisticRegressionSummary extends Serializable {
...
   def predictionCol: String
 
   /** Field in "predictions" which gives the true label of each instance (if available). */
-  @Since("2.3.0")
+  @Since("1.5.0")
   def labelCol: String
 
   /** Field in "predictions" which gives the features of each instance as a vector. */
-  @Since("2.3.0")
+  @Since("1.6.0")
```
WeichenXu123: current master code featuresCol is marked "1.6.0".
I do not remove the `@since` into concrete impl class because they are all private.

mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala
```diff
@@ -1574,18 +1588,17 @@ sealed trait BinaryLogisticRegressionTrainingSummary extends BinaryLogisticRegre
...
-    override val predictionCol: String,
-    override val labelCol: String,
-    override val featuresCol: String,
-    val objectiveHistory: Array[Double])
+    predictions: DataFrame,
+    probabilityCol: String,
+    predictionCol: String,
+    labelCol: String,
+    featuresCol: String,
+    override val objectiveHistory: Array[Double])
```
WeichenXu123: This `val` cannot be removed because in base class it is a `def` so override here is required.

WeichenXu123 (commented):
@jkbradley Thanks for careful review! code updated.

WeichenXu123:
Jenkins, test this please.

WeichenXu123:
Jenkins, test this please.

WeichenXu123:
Jenkins, test this please.

WeichenXu123:
Jenkins, test this please.

mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala
```diff
@@ -1324,90 +1354,147 @@ private[ml] class MultiClassSummarizer extends Serializable {
...
+  @transient private val multiclassMetrics = {
+    new MulticlassMetrics(
+      predictions.select(
+        col(predictionCol),
+        col(labelCol).cast(DoubleType))
+        .rdd.map { case Row(prediction: Double, label: Double) => (prediction, label) })
+  }
+
+  /**
+   * Returns the sequence of labels in ascending order
```
jkbradley: Clarify: "Returns the sequence of labels in ascending order.  This order matches the order used in metrics which are specified as arrays over labels, e.g., truePositiveRateByLabel."

mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala
jkbradley: I see; I think this should work.

jkbradley (commented):
Has there been discussion of removing the Experimental tags from the summary types?  I'd prefer to leave them since there is still talk of further refactoring (e.g., generalizing out a ClassificationSummary across different models).

Other than that + my 1 small comment, this looks ready.  @sethah and @yanboliang any more comments?

Thanks everyone!

MLnick:
I agree keeping the `@Experimental` tags for now is best.

jkbradley:
OK, then let's reinstate the Experimental tags.  @weichen could you please mark the 4 public summary traits Experimental?

WeichenXu123:
Jenkins test this please

WeichenXu123:
Jenkins test this please

jkbradley:
LGTM
Merging with master
Thanks a lot @WeichenXu123 and everyone who reviewed!

PR #15436, opened 2016-10-11
[SPARK-17875] [BUILD] Remove unneeded direct dependence on Netty 3.x

srowen:

Remove unneeded direct dependency on Netty 3.x. I left the `dependencyManagement` entry because some Hadoop libs still use an older version of Netty 3, and I thought it would be weird if the transitive version we reference went backwards. (Note too that Flume declares a direct separate dependency in test scope on Netty 3.4.x)
## How was this patch tested?

Existing tests

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -162,7 +162,7 @@ Please visit the Netty web site for more information:
 
   * http://netty.io/
 
-Copyright 2011 The Netty Project
+Copyright 2014 The Netty Project
 
 The Netty Project licenses this file to you under the Apache License,
 version 2.0 (the "License"); you may not use this file except in compliance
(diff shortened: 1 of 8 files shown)


NOTICE
```diff
@@ -162,7 +162,7 @@ Please visit the Netty web site for more information:
 
   * http://netty.io/
 
-Copyright 2011 The Netty Project
+Copyright 2014 The Netty Project
```
srowen: The license changes are overdue update of the Netty 4.x license from 3.x's version

dev/deps/spark-deps-hadoop-2.3
```diff
@@ -130,7 +130,6 @@ metrics-json-3.1.2.jar
 metrics-jvm-3.1.2.jar
 minlog-1.3.0.jar
 mx4j-3.0.2.jar
-netty-3.8.0.Final.jar
```
srowen: I can't quite figure this out: Hadoop 2.2 and 2.6-2.7 do transitively depend on Netty 3.6.x. 2.3 and 2.4 do not. _shrug_

srowen (commented):
CC @zsxwing

zsxwing:
retest this please

rabbitonweb:
My question would be: what about updating netty 4.x version as well? Right now it's `4.0.29.Final` if I recall correctly, but we could update it to `4.1.3.Final`

srowen:
@rabbitonweb We're on 4.0.41 already. 4.1 won't work; see https://issues.apache.org/jira/browse/SPARK-17379

srowen:
@zsxwing I'm getting failures like 

```
ERROR
test_flume_polling_multiple_hosts (pyspark.streaming.tests.FlumePollingStreamTests) ... Traceback (most recent call last):
  File "/home/jenkins/workspace/SparkPullRequestBuilder/python/pyspark/streaming/tests.py", line 1367, in _testMultipleTimes
    f()
  File "/home/jenkins/workspace/SparkPullRequestBuilder/python/pyspark/streaming/tests.py", line 1387, in _testFlumePollingMultipleHosts
    port = self._utils.startSingleSink()
  File "/home/jenkins/workspace/SparkPullRequestBuilder/python/lib/py4j-0.10.3-src.zip/py4j/java_gateway.py", line 1133, in __call__
    answer, self.gateway_client, self.target_id, self.name)
  File "/home/jenkins/workspace/SparkPullRequestBuilder/python/lib/py4j-0.10.3-src.zip/py4j/protocol.py", line 319, in get_return_value
    format(target_id, ".", name), value)
Py4JJavaError: An error occurred while calling o24677.startSingleSink.
: java.lang.NoClassDefFoundError: org/jboss/netty/channel/ChannelPipelineFactory
	at org.apache.spark.streaming.flume.sink.SparkSink.start(SparkSink.scala:90)
	at org.apache.spark.streaming.flume.PollingFlumeTestUtils.startSingleSink(PollingFlumeTestUtils.scala:68)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.jav
```

I wonder if you encountered this and this is why you had to add Netty 3 back into the overall assembly? I'm still debugging why it isn't finding this from the flume assembly.

srowen:
Well, I'm sure the flume assembly has the classes in question here, and I'm sure it's being added to `--jars` when pyspark is run for the tests. I'm still trying to figure out what's wrong here.

srowen:
Well, heck. This all works fine except with Python 3.x tests. I don't know why yet. Still debugging.

zsxwing:
@srowen agreed. Since Hadoop still depends on Netty 3, it hurts little if we still keep it.

PR #15437, opened 2016-10-11
[SPARK-17876] Write StructuredStreaming WAL to a stream instead of materializing all at once

brkyvz:

The CompactibleFileStreamLog materializes the whole metadata log in memory as a String. This can cause issues when there are lots of files that are being committed, especially during a compaction batch.
You may come across stacktraces that look like:

```
java.lang.OutOfMemoryError: Requested array size exceeds VM limit
at java.lang.StringCoding.encode(StringCoding.java:350)
at java.lang.String.getBytes(String.java:941)
at org.apache.spark.sql.execution.streaming.FileStreamSinkLog.serialize(FileStreamSinkLog.scala:127)

```

The safer way is to write to an output stream so that we don't have to materialize a huge string.
## How was this patch tested?

Existing unit tests

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/CompactibleFileStreamLog.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/CompactibleFileStreamLog.scala
@@ -106,7 +106,7 @@ abstract class CompactibleFileStreamLog[T: ClassTag](
   }
 
   override def deserialize(in: InputStream): Array[T] = {
-    val lines = IOUtils.lineIterator(in, "UTF-8").asScala
+    val lines = IOUtils.lineIterator(in, UTF_8).asScala
     if (!lines.hasNext) {
       throw new IllegalStateException("Incomplete log file")
     }
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/CompactibleFileStreamLog.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/CompactibleFileStreamLog.scala
@@ -62,7 +62,7 @@ abstract class CompactibleFileStreamLog[T: ClassTag](
   /**
    * Serialize the data into encoded string.
    */
-  protected def serializeData(t: T, out: OutputStream): Unit
+  protected def serializeData(t: T): String
 
   /**
    * Deserialize the string into data object.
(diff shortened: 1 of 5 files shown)


brkyvz:
cc @tdas @zsxwing Would one of you want to look at this?

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/HDFSMetadataLog.scala
```diff
@@ -17,19 +17,17 @@
...
+import java.io.{FileNotFoundException, InputStream, IOException, OutputStream}
 import java.util.{ConcurrentModificationException, EnumSet, UUID}
 
 import scala.reflect.ClassTag
 
-import org.apache.commons.io.IOUtils
 import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.fs._
 import org.apache.hadoop.fs.permission.FsPermission
+import org.apache.hadoop.io.IOUtils
```
zsxwing: nit: use org.apache.commons.io.IOUtils instead. Hadoop's IOUtils is `@InterfaceStability.Evolving` which can break compatibility at minor release

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/CompactibleFileStreamLog.scala
```diff
@@ -93,20 +95,26 @@ abstract class CompactibleFileStreamLog[T: ClassTag](
...
+      out.write(serializeData(data).getBytes(UTF_8))
+    }
+    out.flush()
   }
 
-  override def deserialize(bytes: Array[Byte]): Array[T] = {
-    val lines = new String(bytes, UTF_8).split("\n")
-    if (lines.length == 0) {
+  override def deserialize(in: InputStream): Array[T] = {
+    val lines = IOUtils.lineIterator(in, UTF_8).asScala
```
zsxwing: nit: why not use `Source.getLines` since this is Scala.

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/CompactibleFileStreamLog.scala
```diff
@@ -93,20 +95,26 @@ abstract class CompactibleFileStreamLog[T: ClassTag](
...
-  override def serialize(logData: Array[T]): Array[Byte] = {
-    (metadataLogVersion +: logData.map(serializeData)).mkString("\n").getBytes(UTF_8)
+  override def serialize(logData: Array[T], out: OutputStream): Unit = {
+    // called inside a try-finally where the underlying stream is closed in the caller
+    out.write(metadataLogVersion.getBytes(UTF_8))
+    logData.foreach { data =>
+      out.write('\n')
+      out.write(serializeData(data).getBytes(UTF_8))
+    }
+    out.flush()
```
zsxwing: nit: no need to flush since it will be closed at once.

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/FileStreamSourceLog.scala
```diff
@@ -17,6 +17,7 @@
 
 package org.apache.spark.sql.execution.streaming
 
+import java.io.OutputStream
```
zsxwing: nit: unused import

zsxwing (changes requested):
Looks pretty good. Just some nits

brkyvz:
Thanks @zsxwing addressed your comments

zsxwing:

PR #15440, opened 2016-10-11
Fix hadoop.version in building-spark.md

apivovarov:
Couple of mvn build examples use `-Dhadoop.version=VERSION` instead of actual version number


rxin:
Thanks - merging in master/branch-2.0.

PR #4823, opened 2015-02-28
[SPARK-4411][UI]Add kill link for jobs in the UI

lianhuiwang:
We should have a "kill" link for each job, similar to what we have for each stage, so it's easier for users to kill  jobs in the UI. @kayousterhout can you take a look at this? thanks.

PR #15441, opened 2016-10-12
[SPARK-4411] [Web UI] Add "kill" link for jobs in the UI

ajbozarth:

Currently users can kill stages via the web ui but not jobs directly (jobs are killed if one of their stages is). I've added the ability to kill jobs via the web ui. This code change is based on #4823 by @lianhuiwang and updated to work with the latest code matching how stages are currently killed. In general I've copied the kill stage code warning and note comments and all. I also updated applicable tests and documentation.
## How was this patch tested?

Manually tested and dev/run-tests

screen shot 2016-10-11 at 4 49 43 pm


ajbozarth:
@srowen @kayousterhout @tgravescs You have had input or the JIRA or previous PR, could you take a look?

core/src/main/scala/org/apache/spark/ui/jobs/JobsTab.scala
```diff
@@ -35,4 +37,18 @@ private[ui] class JobsTab(parent: SparkUI) extends SparkUITab(parent, "jobs") {
...
   attachPage(new AllJobsPage(this))
   attachPage(new JobPage(this))
+
+  def handleKillRequest(request: HttpServletRequest): Unit = {
+    if (killEnabled && (parent.securityManager.checkModifyPermissions(request.getRemoteUser))) {
+      val killFlag = Option(request.getParameter("terminate")).getOrElse("false").toBoolean
+      val jobId = Option(request.getParameter("id")).getOrElse("-1").toInt
+      if (jobId >= 0 && killFlag && jobProgresslistener.activeJobs.contains(jobId)) {
+        sc.get.cancelJob(jobId)
+      }
```
markhamstra: Creating an `Option` only to immediately `get` the value out of it is poor style, and unnecessary.
```scala
val jobId = Option(request.getParameter("id"))
jobId.foreach { id =>
  if (killFlag && jobProgresslistener.activeJobs.contains(id)) {
    sc.get.cancelJob(id)
  }
}
```

markhamstra: Similarly, there is no need for `sc.get`.  In fact, that's a bug if `parent.sc` really should be an `Option` and thus could be `None`.
```scala
sc.foreach(_.cancelJob(id))
```

markhamstra: And if the Job isn't actually going to be canceled, then there is no need to delay the page refresh (which I'm not entirely happy with, but I'm not going to try to resolve that issue right now.)  So...
```scala
sc.foreach { sparkContext =>
  sparkContext.cancelJob(id)
  Thread.sleep(100)
}
```

And we better make that `jobId` is an `Option[Int]` instead of an `Option[String]`, so...
```scala
val jobId = Option(request.getParameter("id")).map(_.toInt)
```

ajbozarth: I copied this code from `StagesTab.scala` so I'll fix these issues for both in the morning

ajbozarth:
Addressed @markhamstra comments for both `JobsTab` and `StagesTab`

core/src/main/scala/org/apache/spark/ui/jobs/JobsTab.scala
```diff
@@ -35,4 +37,20 @@ private[ui] class JobsTab(parent: SparkUI) extends SparkUITab(parent, "jobs") {
...
+    if (killEnabled && parent.securityManager.checkModifyPermissions(request.getRemoteUser)) {
+      val killFlag = Option(request.getParameter("terminate")).getOrElse("false").toBoolean
+      val jobId = Option(request.getParameter("id")).map(_.toInt)
+      jobId.foreach { id =>
+        if (killFlag && jobProgresslistener.activeJobs.contains(id)) {
+          sc.foreach(_.cancelJob(id))
+          // Do a quick pause here to give Spark time to kill the job so it shows up as
+          // killed after the refresh. Note that this will block the serving thread so the
+          // time should be limited in duration.
+          Thread.sleep(100)
```
markhamstra: Your doing the sleep here even if `sc` is `None`, but I wouldn't expect that that is even possible or that it is handled completely correctly elsewhere in the Web UI if it actually is possible, so this is likely fine in practice.

ajbozarth: That was my thinking, that in the case of sc being None that this code shouldn't be able to be called. Plus by moving the sleep into the if statement it at least only run when an actual kill is attempted.

ajbozarth:
@srowen @tgravescs any chance you'd be able to take a look at this before spark summit?

core/src/main/scala/org/apache/spark/ui/jobs/JobsTab.scala
srowen: Hm, in both this and the kill-stage endpoint I wonder why we have a 'terminate' flag at all. What would it mean to call this with terminate=false? it seems like this can be checked once, if at all, at the start of the method. Or just removed, if I'm not missing something.

core/src/test/scala/org/apache/spark/ui/UISeleniumSuite.scala
```diff
@@ -651,6 +671,15 @@ class UISeleniumSuite extends SparkFunSuite with WebBrowser with Matchers with B
     }
   }
 
+  def getResponseCode(url: URL, method: String): Int = {
+    val connection = url.openConnection().asInstanceOf[HttpURLConnection]
+    connection.setRequestMethod(method)
+    connection.connect()
+    val code = connection.getResponseCode()
+    connection.disconnect()
```
srowen: For total tidiness you might use try-finally here to disconnect on all paths.

docs/configuration.md
```diff
@@ -624,7 +624,7 @@ Apart from these, the following properties are also available, and may be useful
   <td><code>spark.ui.killEnabled</code></td>
   <td>true</td>
   <td>
-    Allows stages and corresponding jobs to be killed from the web ui.
+    Allows jobs and stages to be killed from the web ui.
```
srowen: Nit: web ui -> web UI

srowen (changes requested):
I think this is pretty OK, pending a few minor changes.

core/src/main/scala/org/apache/spark/ui/jobs/JobsTab.scala
ajbozarth: I just did some looking into the git history and thats left over from when there wasn't a separate `/kill` endpoint, so I'll remove that logic in my next commit

core/src/test/scala/org/apache/spark/ui/UISeleniumSuite.scala
ajbozarth: It might just because it's late and I'm tired but I'm not quite sure where you think the try-finally should be

ajbozarth:
@srowen I addressed most of your comments except the one about the try-finally I commented on above

core/src/test/scala/org/apache/spark/ui/UISeleniumSuite.scala
ajbozarth: Thanks, I didn't know scala could do that (how it returns), still learning new things every day

ajbozarth:
@srowen Addressed the try-finally, this look good to go now?

core/src/main/scala/org/apache/spark/ui/jobs/AllJobsPage.scala
```diff
@@ -586,12 +592,28 @@ private[ui] class JobPagedTable(
   override def row(jobTableRow: JobTableRowData): Seq[Node] = {
     val job = jobTableRow.jobData
 
+    val killLink = if (killEnabled) {
```
srowen: Hm, does this work? there's no 'else' so this becomes an `Any` and I don't know what happens below when it's rendered as a string. Should it be an empty string otherwise?

core/src/main/scala/org/apache/spark/ui/SparkUI.scala
```diff
@@ -59,21 +59,25 @@ private[spark] class SparkUI private (
...
+  val jobsTab = new JobsTab(this)
+
   val stagesTab = new StagesTab(this)
 
   var appId: String = _
 
   /** Initialize all components of the server. */
   def initialize() {
-    attachTab(new JobsTab(this))
+    attachTab(jobsTab)
```
srowen: See https://github.com/apache/spark/pull/15603 which reminds me that these two tabs don't need to be members. They can be locals. You can make the change for both here.

srowen (changes requested):
A couple minor extra comments

core/src/main/scala/org/apache/spark/ui/jobs/AllJobsPage.scala
ajbozarth: Nice catch, since it's supposed to be html I'll add a `else Seq.empty` to each instance of this

ajbozarth:
@srowen I addressed your latest comments

srowen (approved):
I think that looks good pending tests.

ajbozarth:
Jenkins, retest this please

ajbozarth:
Good to merge? @srowen

srowen:
Merged to master

PR #15442, opened 2016-10-12
[SPARK-17853][STREAMING][KAFKA][DOC] make it clear that reusing group.id is bad

koeninger:

Documentation fix to make it clear that reusing group id for different streams is super duper bad, just like it is with the underlying Kafka consumer.
## How was this patch tested?

I built jekyll doc and made sure it looked ok.


rxin:
Merging in master/branch-2.0.

PR #15443, opened 2016-10-12
[SPARK-17881] [SQL] Aggregation function for generating string histograms

wzhfy:

This work is a part of [SPARK-17074](https://issues.apache.org/jira/browse/SPARK-17074) to generate histogram statistics.
This agg function generates equi-width histograms in the form of Map(value: String, frequency: Long) for string type columns, with a maximum number of histogram bins. It returns a empty result if the ndv(number of distinct value) of the column exceeds the maximum number allowed.
## How was this patch tested?

add test cases

### DIFF
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/StringHistogram.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/StringHistogram.scala
@@ -1,0 +1,209 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.spark.sql.catalyst.expressions.aggregate
+
+import java.nio.ByteBuffer
+
+import scala.collection.immutable.TreeMap
+import scala.collection.mutable
+
+import com.google.common.primitives.{Ints, Longs}
+
+import org.apache.spark.sql.catalyst.InternalRow
+import org.apache.spark.sql.catalyst.analysis.TypeCheckResult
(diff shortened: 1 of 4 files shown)


wzhfy:
cc @cloud-fan @hvanhovell

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/StringHistogram.scala
```diff
@@ -0,0 +1,209 @@
...
+object StringHistogram {
+
+  /**
+   * StringHistogramInfo maintains frequency of each distinct value.
+   * @param invalid This is to mark the HashMap as invalid when its size (ndv of the column)
+   *                exceeds numBins, because we won't generate string histograms in that case.
+   * @param bins A HashMap to maintain frequency of each distinct value.
+   */
+  case class StringHistogramInfo(
+      var invalid: Boolean = false,
```
cloud-fan: it's really weird to see a `var` in the constructor of case class, how about
```
case class xxx {
  var invalide = false
}
```

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/StringHistogram.scala
```diff
@@ -0,0 +1,209 @@
...
+/**
+ * The StringHistogram function returns the histogram bins - (distinct value, frequency) pairs
+ * for a string column.
+ * @param child child expression that can produce column value with `child.eval(inputRow)`
+ * @param numBinsExpression The maximum number of bins. When the number of distinct values is
+ *                          larger than this, the function will return an empty result.
+ */
+@ExpressionDescription(
+  usage = "_FUNC_(col, numBins) - Returns histogram bins with the maximum number of bins allowed.")
+case class StringHistogram(
```
wzhfy: We will support histogram for numeric types, but the logics in that agg function will be very different from this. You can refer to discussion on the [jira](https://issues.apache.org/jira/browse/SPARK-17074).

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/StringHistogram.scala
```diff
@@ -0,0 +1,209 @@
...
+
+  // Mark as lazy so that numBinsExpression is not evaluated during tree transformation.
+  private lazy val numBins: Int = numBinsExpression.eval().asInstanceOf[Int]
+
+  override def checkInputDataTypes(): TypeCheckResult = {
+    val defaultCheck = super.checkInputDataTypes()
+    if (defaultCheck.isFailure) {
+      defaultCheck
+    } else if (!numBinsExpression.foldable) {
+      TypeCheckFailure(s"The maximum number of bins provided must be a constant literal")
```
wzhfy: remove "s"

wzhfy:
This pr is included in [a new pr](https://github.com/apache/spark/pull/15637), so I'll close this one.

PR #15444, opened 2016-10-12
[SPARK-17870][MLLIB][ML]Change statistic to pValue for SelectKBest and SelectPercentile because of DoF difference

mpjlu:

For feature selection method ChiSquareSelector, it is based on the ChiSquareTestResult.statistic (ChiSqure value) to select the features. It select the features with the largest ChiSqure value. But the Degree of Freedom (df) of ChiSqure value is different in Statistics.chiSqTest(RDD), and for different df, you cannot base on ChiSqure value to select features.

So we change statistic to pValue for SelectKBest and SelectPercentile
## How was this patch tested?

change existing test


srowen (approved):

srowen:
Merged to master

PR #15449, opened 2016-10-12
[SPARK-17884][SQL] To resolve Null pointer exception when casting from empty string to interval type.

priyankagargnitk:

This change adds a check in castToInterval method of Cast expression , such that if converted value is null , then isNull variable should be set to true.

Earlier, the expression Cast(Literal(), CalendarIntervalType) was throwing NullPointerException because of the above mentioned reason.
## How was this patch tested?

Added test case in CastSuite.scala

jira entry for detail: https://issues.apache.org/jira/browse/SPARK-17884


rxin:
Thanks - merging in master/branch-2.0.

priyankagargnitk:
Thanks rxin

priyankagargnitk:
Hi rxin, Can we merge the same change in branch 1.6 as well... As we are still using spark 1.6 and need this change?

rxin:
can you submit a pull request against that branch?

priyankagargnitk:
Done. PR #15479

PR #15450, opened 2016-10-12
[SPARK-3261] [MLLIB] KMeans clusterer can return duplicate cluster centers

srowen:

Return potentially fewer than k cluster centers in cases where k distinct centroids aren't available or aren't selected.
## How was this patch tested?

Existing tests

### DIFF
--- a/mllib/src/main/scala/org/apache/spark/mllib/clustering/KMeans.scala
+++ b/mllib/src/main/scala/org/apache/spark/mllib/clustering/KMeans.scala
@@ -56,7 +56,8 @@ class KMeans private (
   def this() = this(2, 20, KMeans.K_MEANS_PARALLEL, 2, 1e-4, Utils.random.nextLong())
 
   /**
-   * Number of clusters to create (k).
+   * Number of clusters to create (k). Note that if the input has fewer than k elements,
+   * then it's possible that fewer than k clusters are created.
    */
   @Since("1.4.0")
   def getK: Int = k
@@ -323,7 +324,7 @@ class KMeans private (
    * Initialize a set of cluster centers at random.
    */
   private def initRandom(data: RDD[VectorWithNorm]): Array[VectorWithNorm] = {
-    data.takeSample(true, k, new XORShiftRandom(this.seed).nextInt()).map(_.toDense)
+    data.takeSample(false, k, new XORShiftRandom(this.seed).nextInt()).map(_.toDense)
   }
 
   /**
@@ -378,10 +379,10 @@ class KMeans private (
     costs.unpersist(blocking = false)
     bcNewCentersList.foreach(_.destroy(false))
 
-    if (centers.size == k) {
+    if (centers.size <= k) {
       centers.toArray
     } else {
-      // Finally, we might have a set of more or less than k candidate centers; weight each
+      // Finally, we might have a set of more than k candidate centers; weight each
       // candidate by the number of points in the dataset mapping to it and run a local k-means++
       // on the weighted centers to pick k of them
       val bcCenters = data.context.broadcast(centers)
--- a/mllib/src/main/scala/org/apache/spark/mllib/clustering/KMeans.scala
+++ b/mllib/src/main/scala/org/apache/spark/mllib/clustering/KMeans.scala
@@ -326,8 +326,8 @@ class KMeans private (
   private def initRandom(data: RDD[VectorWithNorm]): Array[VectorWithNorm] = {
     // Select without replacement; may still produce duplicates if the data has < k distinct
     // points, so deduplicate the centroids to match the behavior of k-means|| in the same situation
-    data.takeSample(false, k, new XORShiftRandom(this.seed).nextInt()).
-      map(_.vector).distinct.map(new VectorWithNorm(_))
+    data.takeSample(false, k, new XORShiftRandom(this.seed).nextInt())
+      .map(_.vector).distinct.map(new VectorWithNorm(_))
   }
 
   /**
--- a/mllib/src/main/scala/org/apache/spark/mllib/clustering/KMeans.scala
+++ b/mllib/src/main/scala/org/apache/spark/mllib/clustering/KMeans.scala
@@ -324,7 +324,10 @@ class KMeans private (
    * Initialize a set of cluster centers at random.
    */
   private def initRandom(data: RDD[VectorWithNorm]): Array[VectorWithNorm] = {
-    data.takeSample(false, k, new XORShiftRandom(this.seed).nextInt()).map(_.toDense)
+    // Select without replacement; may still produce duplicates if the data has < k distinct
+    // points, so deduplicate the centroids to match the behavior of k-means|| in the same situation
+    data.takeSample(false, k, new XORShiftRandom(this.seed).nextInt()).
+      map(_.vector).distinct.map(new VectorWithNorm(_))
   }
 
   /**
--- a/mllib/src/main/scala/org/apache/spark/mllib/clustering/KMeans.scala
+++ b/mllib/src/main/scala/org/apache/spark/mllib/clustering/KMeans.scala
@@ -56,14 +56,15 @@ class KMeans private (
   def this() = this(2, 20, KMeans.K_MEANS_PARALLEL, 2, 1e-4, Utils.random.nextLong())
 
   /**
-   * Number of clusters to create (k). Note that if the input has fewer than k elements,
-   * then it's possible that fewer than k clusters are created.
+   * Number of clusters to create (k). Note that it is possible for fewer than k clusters to
+   * be returned, for example, if there are fewer than k distinct points to cluster.
    */
   @Since("1.4.0")
   def getK: Int = k
 
   /**
-   * Set the number of clusters to create (k). Default: 2.
+   * Set the number of clusters to create (k). Note that it is possible for fewer than k clusters to
+   * be returned, for example, if there are fewer than k distinct points to cluster. Default: 2.
    */
   @Since("0.8.0")
   def setK(k: Int): this.type = {
@@ -339,7 +340,7 @@ class KMeans private (
    *
    * The original paper can be found at http://theory.stanford.edu/~sergei/papers/vldb12-kmpar.pdf.
    */
-  private def initKMeansParallel(data: RDD[VectorWithNorm]): Array[VectorWithNorm] = {
+  private[clustering] def initKMeansParallel(data: RDD[VectorWithNorm]): Array[VectorWithNorm] = {
     // Initialize empty centers and point costs.
     var costs = data.map(_ => Double.PositiveInfinity)
 
@@ -382,19 +383,21 @@ class KMeans private (
     costs.unpersist(blocking = false)
     bcNewCentersList.foreach(_.destroy(false))
 
-    if (centers.size <= k) {
-      centers.toArray
+    val distinctCenters = centers.map(_.vector).distinct.map(new VectorWithNorm(_))
+
+    if (distinctCenters.size <= k) {
+      distinctCenters.toArray
     } else {
-      // Finally, we might have a set of more than k candidate centers; weight each
+      // Finally, we might have a set of more than k distinct candidate centers; weight each
       // candidate by the number of points in the dataset mapping to it and run a local k-means++
       // on the weighted centers to pick k of them
-      val bcCenters = data.context.broadcast(centers)
+      val bcCenters = data.context.broadcast(distinctCenters)
       val countMap = data.map(KMeans.findClosest(bcCenters.value, _)._1).countByValue()
 
       bcCenters.destroy(blocking = false)
 
-      val myWeights = centers.indices.map(countMap.getOrElse(_, 0L).toDouble).toArray
-      LocalKMeans.kMeansPlusPlus(0, centers.toArray, myWeights, k, 30)
+      val myWeights = distinctCenters.indices.map(countMap.getOrElse(_, 0L).toDouble).toArray
+      LocalKMeans.kMeansPlusPlus(0, distinctCenters.toArray, myWeights, k, 30)
     }
   }
 }
--- a/mllib/src/test/scala/org/apache/spark/mllib/clustering/KMeansSuite.scala
+++ b/mllib/src/test/scala/org/apache/spark/mllib/clustering/KMeansSuite.scala
@@ -75,7 +75,7 @@ class KMeansSuite extends SparkFunSuite with MLlibTestSparkContext {
 
     // Make sure code runs.
     var model = KMeans.train(data, k = 2, maxIterations = 1)
-    assert(model.clusterCenters.size === 2)
+    assert(model.clusterCenters.size === 1)
   }
 
   test("more clusters than points") {
@@ -87,7 +87,7 @@ class KMeansSuite extends SparkFunSuite with MLlibTestSparkContext {
 
     // Make sure code runs.
     var model = KMeans.train(data, k = 3, maxIterations = 1)
-    assert(model.clusterCenters.size === 3)
+    assert(model.clusterCenters.size === 2)
   }
 
   test("deterministic initialization") {
--- a/mllib/src/test/scala/org/apache/spark/mllib/clustering/KMeansSuite.scala
+++ b/mllib/src/test/scala/org/apache/spark/mllib/clustering/KMeansSuite.scala
@@ -64,18 +64,34 @@ class KMeansSuite extends SparkFunSuite with MLlibTestSparkContext {
     assert(model.clusterCenters.head ~== center absTol 1E-5)
   }
 
-  test("no distinct points") {
+  test("fewer distinct points than clusters") {
     val data = sc.parallelize(
       Array(
         Vectors.dense(1.0, 2.0, 3.0),
         Vectors.dense(1.0, 2.0, 3.0),
         Vectors.dense(1.0, 2.0, 3.0)),
       2)
-    val center = Vectors.dense(1.0, 2.0, 3.0)
 
-    // Make sure code runs.
-    var model = KMeans.train(data, k = 2, maxIterations = 1)
-    assert(model.clusterCenters.size === 1)
+    var model = KMeans.train(data, k = 2, maxIterations = 1, initializationMode = "random")
+    assert(model.clusterCenters.length === 1)
+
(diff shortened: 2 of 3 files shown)


srowen:
@sethah I wanted to check how strongly against this kind of change you might be, and continue to discussion here.

sethah:
@srowen I'm not against the change per se, I was just hoping to understand how duplicate centers arise. In the case of `initRandom` sampling with replacement makes it possible to select the same initial centers, but it should be quite unlikely if there are much more unique data points than requested centers. Even when this happens, the algorithm should move the centers unless they never have any data assigned to them. Since the centers are double-valued points in the feature space, when we say duplicate centers do we mean literally duplicate or that `|c1 - c2|_p < eps` for some norm? 

It seems to me that the problem of duplicate centers would not arise in most real-world use cases, but from comments on the JIRA it appears that assumption could be false. I think it's easier to assess the change if we understand what causes this situation. Do you have any insight?

srowen:
@sethah I agree that when there are lots of unique points (>> k) then this is almost certain to not happen, and that's most real-world use cases, but the question indeed is what should happen when this is not the case. In that sense, this change only affects corner cases so isn't really a big deal either way.

Yes the one case is clear: sampling with replacement when the data set has < k (unique) points. It will always return k centroids, so must return duplicates. In this case, every point will be at distance 0 from some centroid and so I don't think the centroids can move apart. It stops in 1 iteration with the degenerate solution, with some centroids assigned 0 points. Not the end of the world but not exactly meaningful.

The more interesting case is k-means ||. Of course, again, if there are < k unique points to start, in this case as well, returning k centroids means returning duplicates. Same argument there -- seems to be no value in returning k centroids.

This is really the sum of the argument to me, regardless of what Derrick's case is.

A twist: it's possible, but quite improbable, for k-means || to choose fewer than k unique centroids, when there are >= k distinct points. This is most likely when there are barely more than k distinct points. In that case it's possible that duplicated centroids do get pulled apart and do end up doing something meaningful. I am arguing this case is not worth dealing with because it's rare and it doesn't meaningfully harm the quality of the resulting clustering, but, that point is arguable.

I am about 7/10 in favor of the change, certainly the bit about sampling without replacement, but the rest I could drop if there's any significant objection to it.

sethah:
The cases you enumerated are the ones I was thinking of. The changes introduced here would alleviate those problems, I agree. What I'm wondering is if this problem still exists in other cases. If Derrick had 1.3M data points and asked for 10k clusters, and got only 1k unique cluster centers, why did it happen? Is it common/possible for clusters that start in different locations to converge to the same point? I'd be interested to replicate this issue, not sure if I will have the time.

Also, I'm slightly inclined to match scikit or R unless we're certain of a clear benefit not to.

srowen:
@sethah I should say I am not trying to handle cases where clusters start separate and converge to nearly the same point. I don't that's something we should even try to do. 

To elaborate, here are the relevant cases, I think. I'm ignoring cases where the data size is >> k because it all behaves as desired with high probability already.

*Case 1*. Data = [A, B], k = 3, init = random
Now, you'll get 3 centroids and definitely have a duplicate. After this change, you'd get 2.

*Case 2*. Data = [A, A, B, B], k = 3, init = random
Now, you'll get 3 centroids and definitely have a duplicate. After this change, same thing.

*Case 3*. Data = [A, B, C, D], k = 3, init = random
Now, you'll get 3 centroids and occasionally have a duplicate. After, this change, you'd get 3 distinct centroids always.

*Case 4*. Data = [A, B], k = 3, init = parallel
Now, you'll get 3 centroids and definitely have a duplicate. After this change, you should get 2 if I read the code right (that it won't pick a centroid at distance 0 from another).

*Case 5*. Data = [A, A, B, B], k = 3, init = parallel
Now, you'll get 3 centroids and definitely have a duplicate. After this change, you should get 2 for the same reason above.

*Case 6*. Data = [A, B, C, D], k = 3, init = parallel
Now, you'll get 3 centroids and occasionally have a duplicate. After this change, same thing.

Case 1/4/5 are, I think, positive changes. Case 3 seems like a clear win. Case 2 could be made consistent with 1/4/5 easily, I suppose, by calling `.distinct`. Case 6 is the interesting one -- can we make it manage to return 3 distinct centroids always because that's clearly possible? like case 3? I punted on that though.

I think Derrick's case is like 2 or 5. There are duplicates in the data and not many unique points. Filtering the duplicates early is a performance win. At least -- that is the case I am trying to solve. If that's not what is described in the JIRA then I think it's a different issue or question.

I think there's a benefit to speed and meaningfulness of the model in case 1/2/4/5.

sethah:
Aren't all these cases sort of non-sensical anyway? What good is performing clustering on a dataset where the result has (approximately) the same number of clusters as unique data points?

The reason I'm clarifying this is because if this arguably beneficial patch only applies to useless applications of clustering, then I would lean towards maintaining the current behavior.

srowen:
Heh, I believe the PC term is 'corner cases'. I agree. There's not much point in clustering data to k centroids when there are <= k distinct points. I think that's all the more reasons to not make the output sillier by returning k centroids. That's really the extent of the question here. I am not sure I understand why you believe the current behavior is better?

I'd like to resolve this at last, so unless you would -1 this, I'm going to proceed.

sethah:
I don't feel strongly either way, but I don't like the potential of this:

````scala
model.getK
scala> 3
model.clusterCenters.length
scala> 1
````

Should we consider setting the model `k` to the number of clusters at the end of training? Or is there some utility to knowing that the number of clusters was less than the requested `k`?

**Comment by srowen** [comment]
_k_ is a parameter to the model building process, and I don't think it should change based on the model that comes out. It's the requested or maximum number of centroids, if you like. Or, weigh that against printing the centroids and finding duplicates. Does that make more sense?

File: mllib/src/test/scala/org/apache/spark/mllib/clustering/KMeansSuite.scala [hunk 1]
```diff
@@ -75,7 +75,7 @@ class KMeansSuite extends SparkFunSuite with MLlibTestSparkContext {
 
     // Make sure code runs.
```
sethah: while we're here, can we make `var model = ...` into `val model = ...` and delete L74 `val center = ...` which is not used.

mllib/src/test/scala/org/apache/spark/mllib/clustering/KMeansSuite.scala
```diff
@@ -75,7 +75,7 @@ class KMeansSuite extends SparkFunSuite with MLlibTestSparkContext {
...
     var model = KMeans.train(data, k = 2, maxIterations = 1)
-    assert(model.clusterCenters.size === 2)
+    assert(model.clusterCenters.size === 1)
```
sethah: these tests don't cover `initRandom`. We should add both initialization methods to be complete

mllib/src/main/scala/org/apache/spark/mllib/clustering/KMeans.scala
```diff
@@ -323,7 +324,10 @@ class KMeans private (
    * Initialize a set of cluster centers at random.
    */
   private def initRandom(data: RDD[VectorWithNorm]): Array[VectorWithNorm] = {
-    data.takeSample(true, k, new XORShiftRandom(this.seed).nextInt()).map(_.toDense)
+    // Select without replacement; may still produce duplicates if the data has < k distinct
+    // points, so deduplicate the centroids to match the behavior of k-means|| in the same situation
+    data.takeSample(false, k, new XORShiftRandom(this.seed).nextInt()).
+      map(_.vector).distinct.map(new VectorWithNorm(_))
```
sethah: I think we generally put the `.` on the next line, but I'm not sure it's a strict style requirement.

Also, we end up recomputing the norm of each sample. For large `k` this can be inefficient. I'm not sure of a non-ugly way to implement the distinct otherwise.

sethah (commented):
I left a few comments. I will take another look in more detail later on.

mllib/src/test/scala/org/apache/spark/mllib/clustering/KMeansSuite.scala
srowen: I will add more tests, yes.

mllib/src/main/scala/org/apache/spark/mllib/clustering/KMeans.scala
srowen: I'll move the dot. Agree, though computing k norms once seems negligible in the context of the overall computation.

mllib/src/main/scala/org/apache/spark/mllib/clustering/KMeans.scala
```diff
@@ -378,10 +382,10 @@ class KMeans private (
     costs.unpersist(blocking = false)
     bcNewCentersList.foreach(_.destroy(false))
 
-    if (centers.size == k) {
+    if (centers.size <= k) {
```
sethah: There's nothing that enforces these centers to be distinct. The following test(s) fail (IMO these are much more thorough than the existing tests):

````scala
  test("unique cluster centers") {
    val rng = new scala.util.Random(42)
    val points = (0 until 10).map(i => Vectors.dense(Array.fill(3)(rng.nextDouble)))
    val data = sc.parallelize(
      points.flatMap { point =>
        Array.fill(rng.nextInt(4))(point)
      }, 2
    )
    val norms = data.map(Vectors.norm(_, 2.0))
    val zippedData = data.zip(norms).map { case (v, norm) =>
      new VectorWithNorm(v, norm)
    }
    // less centers than k
    val km = new KMeans().setK(50)
      .setMaxIterations(10)
      .setInitializationMode("k-means||")
      .setInitializationSteps(10)
      .setSeed(42)
    val initialCenters = km.initKMeansParallel(zippedData).map(_.vector)
    assert(initialCenters.length === initialCenters.distinct.length)

    val model = km.run(data)
    val finalCenters = model.clusterCenters
    assert(finalCenters.length === finalCenters.distinct.length)

    // run local k-means
    val km2 = new KMeans().setK(10)
      .setMaxIterations(10)
      .setInitializationMode("k-means||")
      .setInitializationSteps(10)
      .setSeed(42)
    val initialCenters2 = km2.initKMeansParallel(zippedData).map(_.vector)
    assert(initialCenters2.length === initialCenters2.distinct.length)

    val model2 = km.run(data)
    val finalCenters2 = model2.clusterCenters
    assert(finalCenters2.length === finalCenters2.distinct.length)
````

I guess we can call distinct on the centers before this if statement?

**Comment by sethah** [comment]
Also, if we're going to make this change, we should document in the ML estimator that the algorithm can return fewer than `k` centers.

File: mllib/src/main/scala/org/apache/spark/mllib/clustering/KMeans.scala (hunk 4 above)
Comment (srowen) [inline mllib/src/main/scala/org/apache/spark/mllib/clustering/KMeans.scala]: That's right, although it's not a goal to guarantee distinct centers, though that's a nice to have where possible. I think my goal is improving the straightforward cases, and, maintaining as much consistency as possible. I agree with adding a test like this and adding a call to `.distinct`. Might as well take this to a logical conclusion.

It also raises the interesting question: if you have >= k distinct points, and happen to pick < k distinct centroids, should you go back and replenish the set of centroids? I am punting on that right now but it's a legitimate point. It's quite a corner case though.

**Comment by srowen** [comment]
@sethah let me know how you feel about it at this stage

File: mllib/src/test/scala/org/apache/spark/mllib/clustering/KMeansSuite.scala [hunk 5]
```diff
@@ -64,18 +64,34 @@ class KMeansSuite extends SparkFunSuite with MLlibTestSparkContext {
...
-    assert(model.clusterCenters.size === 2)
+    var model = KMeans.train(data, k = 2, maxIterations = 1, initializationMode = "random")
+    assert(model.clusterCenters.length === 1)
+
+    model = KMeans.train(data, k = 2, maxIterations = 1, initializationMode = "k-means||")
+    assert(model.clusterCenters.length === 1)
+  }
+
+
+  test("fewer clusters than points") {
```
sethah: I'd prefer to remove these two tests since we've added a more thorough test below. We can check the "random" init method in that test as well, then we can eliminate these two.

mllib/src/test/scala/org/apache/spark/mllib/clustering/KMeansSuite.scala
```diff
@@ -85,9 +101,50 @@ class KMeansSuite extends SparkFunSuite with MLlibTestSparkContext {
...
+      new VectorWithNorm(v, norm)
+    }
+    // less centers than k
+    val km = new KMeans().setK(50)
+      .setMaxIterations(10)
+      .setInitializationMode("k-means||")
+      .setInitializationSteps(10)
+      .setSeed(42)
+    val initialCenters = km.initKMeansParallel(zippedData).map(_.vector)
+    assert(initialCenters.length === initialCenters.distinct.length)
```
sethah: also check `initialCenters.length <= numDistinctPoints` where `numDistinctPoints` is 10 (defined above) at the moment.

mllib/src/test/scala/org/apache/spark/mllib/clustering/KMeansSuite.scala
```diff
@@ -85,9 +101,50 @@ class KMeansSuite extends SparkFunSuite with MLlibTestSparkContext {
...
+    assert(finalCenters.length === finalCenters.distinct.length)
+
+    // run local k-means
+    val km2 = new KMeans().setK(10)
+      .setMaxIterations(10)
+      .setInitializationMode("k-means||")
+      .setInitializationSteps(10)
+      .setSeed(42)
+    val initialCenters2 = km2.initKMeansParallel(zippedData).map(_.vector)
+    assert(initialCenters2.length === initialCenters2.distinct.length)
```
sethah: also check `initialCenters2.length === 2`

mllib/src/test/scala/org/apache/spark/mllib/clustering/KMeansSuite.scala
```diff
@@ -85,9 +101,50 @@ class KMeansSuite extends SparkFunSuite with MLlibTestSparkContext {
...
+
+    val model2 = km.run(data)
```
sethah: should be `val model2 = km2.run(data)`

sethah (commented):
Left some comments, just some cleanup changes to make to the tests really.

mllib/src/main/scala/org/apache/spark/mllib/clustering/KMeans.scala
sethah: Ok, I think you're probably right about it being negligible.

mllib/src/test/scala/org/apache/spark/mllib/clustering/KMeansSuite.scala
```diff
@@ -64,30 +66,55 @@ class KMeansSuite extends SparkFunSuite with MLlibTestSparkContext {
...
+
+    // run local k-means
+    val km2 = new KMeans().setK(10)
+      .setMaxIterations(5)
+      .setInitializationMode("k-means||")
+      .setInitializationSteps(10)
+      .setSeed(seed)
+    val initialCenters2 = km2.initKMeansParallel(normedData).map(_.vector)
+    assert(initialCenters2.length === initialCenters2.distinct.length)
+    assert(initialCenters2.length === 10)
```
sethah: minor/nit: maybe make `k` a val here and use that instead. Since we use 10 for something else above, this can be obfuscated in the future.

mllib/src/test/scala/org/apache/spark/mllib/clustering/KMeansSuite.scala
```diff
@@ -64,30 +66,55 @@ class KMeansSuite extends SparkFunSuite with MLlibTestSparkContext {
...
-        Vectors.dense(1.0, 3.0, 4.0)),
-      2)
+    model = KMeans.train(data, k = 2, maxIterations = 1, initializationMode = "k-means||")
+    assert(model.clusterCenters.length === 1)
+  }
 
-    // Make sure code runs.
-    var model = KMeans.train(data, k = 3, maxIterations = 1)
-    assert(model.clusterCenters.size === 3)
+    test("unique cluster centers") {
```
sethah: nit: indentation

mllib/src/test/scala/org/apache/spark/mllib/clustering/KMeansSuite.scala
```diff
@@ -64,30 +66,55 @@ class KMeansSuite extends SparkFunSuite with MLlibTestSparkContext {
...
+    val rng = new Random(seed)
+    val numDistinctPoints = 10
+    val points = (0 until numDistinctPoints).map(i => Vectors.dense(Array.fill(3)(rng.nextDouble)))
+    val data = sc.parallelize(points.flatMap(Array.fill(1 + rng.nextInt(3))(_)), 2)
+    val normedData = data.map(new VectorWithNorm(_))
+
+    // less centers than k
```
sethah: Can we also test the "random" method here? I was going to suggest putting the test cases inside `Seq("k-means||", "random").foreach { initMode =>`, but we run the specific parallel case. Maybe just manually add it?

sethah (commented):
LGTM pending added test case. Thanks a lot!

srowen:
@sethah done. I also removed references to the runs parameter, which has no effect (and was triggering deprecation warnings). I should have done that last time.

mllib/src/test/scala/org/apache/spark/mllib/clustering/KMeansSuite.scala
```diff
@@ -257,11 +246,6 @@ class KMeansSuite extends SparkFunSuite with MLlibTestSparkContext {
     model = KMeans.train(rdd, k = 5, maxIterations = 10)
     assert(model.clusterCenters.sortBy(VectorWithCompare(_))
       .zip(points.sortBy(VectorWithCompare(_))).forall(x => x._1 ~== (x._2) absTol 1E-5))
-
-    // Neither should more runs
```
sethah: Let's fix the other comments that reference runs. e.g. "No matter how many iterations or runs we use ...."

sethah (commented):
One last minor thing :)

srowen (commented):
Done. Let me see one more test pass and then I'll merge. TYVM

srowen:
merged to master

PR #15404, opened 2016-10-09
Branch 2.0

yintengfei:
## What changes were proposed in this pull request?

(Please fill in changes proposed in this fix)
## How was this patch tested?

(Please explain how this patch was tested. E.g. unit tests, integration tests, manual tests)

(If this patch involves UI changes, please attach a screenshot; otherwise, remove this)

PR #15303, opened 2016-09-29
[SPARK-17671] Changed implementation of HistoryServer.getApplicationInfoList for lazy evaluation

wgtmac:
## What changes were proposed in this pull request?

Changed implementation of HistoryServer.getApplicationInfoList() to lazily evaluation the time-consuming transformation. In this way, ApplicationListResource can return an iterator without evaluating the whole input iterator.
## How was this patch tested?

No API added. Just manual unit tests.

PR #15135, opened 2016-09-18
[pyspark][group]pyspark GroupedData can't apply agg functions on all left numeric columns.

citoubest:
## What changes were proposed in this pull request?

With pyspark dataframe, the agg method just support two ways, one is to give the column and agg method maps and another one is to use agg functions in package functions  to apply on specific columns names. The two approach both ask us to asign a method on specific columns names. But if
I want to apply the agg method on all other numeric columns, I should list all method-column combinations. such as, suppose the df has to columns province,age,income, I want to groupby the province and calculate the min, max and average values on age and income, before change I have to approach: df.groupby('province').agg({'age':'max','age':'min','age':'avg','income':'max','income':'min','income':'avg'})
df.groupby('province').agg({F.min('age'),F.max('age'),F.avg('age'),F.min('income'),F.max('income'),F.avg('income')}), which are both redundant.

with this change, we can simply replace the code with df.groupby('province').agg('max','min','avg')
## How was this patch tested?

manual tests

PR #15080, opened 2016-09-13
[SPARK-17526][Web UI]: Display the executor log links with the job failure message on Spark UI and Console

zhzhan:
## What changes were proposed in this pull request?

Add the executor log links with the job failure message on Spark UI and Console
## How was this patch tested?

The patch is manually tested. 

In Console:

ERROR TaskSetManager: Task 0 in stage 0.0 failed 4 times; stdout: stdout_log_link stderr: stderr_log_link; aborting job ...

In WebUI 

screen shot 2016-09-13 at 9 33 06 am

(If this patch involves UI changes, please attach a screenshot; otherwise, remove this)

PR #15451, opened 2016-10-12
[BUILD] Closing stale PRs

srowen:
Closes #15303
Closes #15078
Closes #15080
Closes #15135
Closes #14565
Closes #12355
Closes #15404


rxin:
Merging in master.

rxin:
Actually I'm getting error when merging. Not sure why.

vanzin:
Let me try...

vanzin:
Done. I had to temporarily add `--allow-empty` to the `git commit` command in `./dev/merge_spark_pr.py`. Perhaps we should fix the script to always add that option (or have some command line switch for it).

srowen:
Yes you have to add `--allow-empty` temporarily. I hadn't added it to the script because usually we don't want to allow an empty commit. Yes if this becomes pretty regular, probably worth some kind of check and prompt during the script that would allow this to proceed without temporarily changing the script.

PR #15453, opened 2016-10-12
[SPARK-17770] [CATALYST] making ObjectType public

bdrillard:

In order to facilitate the writing of additional Encoders, I proposed opening up the ObjectType SQL DataType. This DataType is used extensively in the JavaBean Encoder, but would also be useful in writing other custom encoders. 

As mentioned by @marmbrus, it is understood that the Expressions API is subject to potential change.
## How was this patch tested?

The change only affects the visibility of the ObjectType class, and the existing SQL test suite still runs without error.

### DIFF
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/types/ObjectType.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/types/ObjectType.scala
@@ -19,7 +19,10 @@ package org.apache.spark.sql.types
 
 import scala.language.existentials
 
-private[sql] object ObjectType extends AbstractDataType {
+import org.apache.spark.annotation.InterfaceStability
+
+@InterfaceStability.Evolving
+object ObjectType extends AbstractDataType {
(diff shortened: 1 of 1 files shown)


marmbrus:
ok to test

sql/catalyst/src/main/scala/org/apache/spark/sql/types/ObjectType.scala
```diff
@@ -19,7 +19,7 @@ package org.apache.spark.sql.types
 
 import scala.language.existentials
 
-private[sql] object ObjectType extends AbstractDataType {
+object ObjectType extends AbstractDataType {
```
rxin: I would put a InterfaceStability.Evolving to this.

bdrillard: Sure, I've updated the pull-request for the annotation.

marmbrus:
Thanks, merging to master!

PR #15454, opened 2016-10-12
[SPARK-17850][Core]Add a flag to ignore corrupt files (branch 1.6)

zsxwing:

This is the patch for 1.6. It only adds Spark conf `spark.files.ignoreCorruptFiles` because SQL just uses HadoopRDD directly in 1.6. `spark.files.ignoreCorruptFiles` is `true` by default.
## How was this patch tested?

The added test.

### DIFF
--- a/core/src/main/scala/org/apache/spark/rdd/HadoopRDD.scala
+++ b/core/src/main/scala/org/apache/spark/rdd/HadoopRDD.scala
@@ -139,6 +139,9 @@ class HadoopRDD[K, V](
 
   private val shouldCloneJobConf = sparkContext.conf.getBoolean("spark.hadoop.cloneConf", false)
 
+  private val ignoreCorruptFiles =
+    sparkContext.conf.getBoolean("spark.files.ignoreCorruptFiles", true)
+
   // Returns a JobConf that will be used on slaves to obtain input splits for Hadoop reads.
   protected def getJobConf(): JobConf = {
     val conf: Configuration = broadcastedConf.value.value
@@ -245,8 +248,7 @@ class HadoopRDD[K, V](
         try {
           finished = !reader.next(key, value)
         } catch {
-          case eof: EOFException =>
-            finished = true
+          case _: EOFException if ignoreCorruptFiles => finished = true
         }
         if (!finished) {
           inputMetrics.incRecordsRead(1)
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/apache/spark/rdd/NewHadoopRDD.scala
```diff
@@ -171,7 +175,11 @@ class NewHadoopRDD[K, V](
 
       override def hasNext: Boolean = {
         if (!finished && !havePair) {
-          finished = !reader.nextKeyValue
+          try {
+            finished = !reader.nextKeyValue
+          } catch {
+            case _: EOFException if ignoreCorruptFiles => finished = true
```
zsxwing: This is a behavior change to `NewHadoopRDD`, which may surprise the existing 1.6 users.

core/src/main/scala/org/apache/spark/rdd/HadoopRDD.scala
```diff
@@ -245,8 +248,7 @@ class HadoopRDD[K, V](
         try {
           finished = !reader.next(key, value)
         } catch {
-          case eof: EOFException =>
-            finished = true
+          case _: EOFException if ignoreCorruptFiles => finished = true
```
zsxwing: I didn't use `IOException` to keep the default behavior is same as before.

zsxwing (commented):
The problem here is should HadoopRDD and NewHadoopRDD be consistent. If so, it means we have to break the current behavior.

core/src/main/scala/org/apache/spark/rdd/HadoopRDD.scala
mridulm: sounds good

core/src/main/scala/org/apache/spark/rdd/NewHadoopRDD.scala
mridulm: Yeah, I am slightly worried about this change of behavior too.
Though I think it should be fine.

mridulm (approved):
LGTM

zsxwing:
Since I have not yet heard complaints about this for 1.6, and this one may break some user's job, I'm going to close it now.

PR #15456, opened 2016-10-13
[SPARK-17686][Core] Support printing out scala and java version with spark-submit --version command

jerryshao:

In our universal gateway service we need to specify different jars to Spark according to scala version. For now only after launching Spark application can we know which version of Scala it depends on. It makes hard for us to support different Scala + Spark versions to pick the right jars.

So here propose to print out Scala version according to Spark version in "spark-submit --version", so that user could leverage this output to make the choice without needing to launching application.
## How was this patch tested?

Manually verified in local environment.

### DIFF
--- a/core/src/main/scala/org/apache/spark/deploy/SparkSubmit.scala
+++ b/core/src/main/scala/org/apache/spark/deploy/SparkSubmit.scala
@@ -24,6 +24,7 @@ import java.security.PrivilegedExceptionAction
 
 import scala.annotation.tailrec
 import scala.collection.mutable.{ArrayBuffer, HashMap, Map}
+import scala.util.Properties
 
 import org.apache.commons.lang3.StringUtils
 import org.apache.hadoop.fs.Path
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/apache/spark/deploy/SparkSubmit.scala
```diff
@@ -104,6 +104,8 @@ object SparkSubmit {
    /___/ .__/\_,_/_/ /_/\_\   version %s
       /_/
                         """.format(SPARK_VERSION))
+    printStream.println("Using Scala %s (%s, Java %s)".format(
```
rxin: How about just ...

"Using Scala %s, %s, %s"

rxin: Basically Properties.javaVmName already says "Java".

jerryshao: Thanks Reynold for your comments. I will change it.

rxin:
LGTM pending Jenkins.

andrewor14:
Merging into master

jerryshao:
Thanks @rxin and @andrewor14 .

jerryshao:
Looks like unrelated test failure, it can be passed in my local test.

PR #15457, opened 2016-10-13
[SPARK-17830][SQL] Annotate remaining SQL APIs with InterfaceStability

rxin:

This patch annotates all the remaining APIs in SQL (excluding streaming) with InterfaceStability.
## How was this patch tested?

N/A - just annotation change.


sql/core/src/main/java/org/apache/spark/sql/api/java/UDF1.java
```diff
@@ -19,14 +19,12 @@
...
-// **************************************************
+import org.apache.spark.annotation.InterfaceStability;
 
 /**
  * A Spark SQL UDF that has 1 arguments.
  */
+@InterfaceStability.Stable
 public interface UDF1<T1, R> extends Serializable {
-  public R call(T1 t1) throws Exception;
+  R call(T1 t1) throws Exception;
```
rxin: methods in interface in Java are by default public.

hvanhovell:
LGTM - pending jenkins

sql/core/src/main/java/org/apache/spark/sql/api/java/UDF1.java
```diff
@@ -19,14 +19,12 @@
 
 import java.io.Serializable;
 
-// **************************************************
```
viirya: +1

hvanhovell:
Merging to master. Thanks!

PR #15458, opened 2016-10-13
[SPARK-17899][SQL] add a debug mode to keep raw table properties in HiveExternalCatalog

cloud-fan:

Currently `HiveExternalCatalog` will filter out the Spark SQL internal table properties, e.g. `spark.sql.sources.provider`, `spark.sql.sources.schema`, etc. This is reasonable for external users as they don't want to see these internal properties in `DESC TABLE`.

However, as a Spark developer, sometimes we do wanna see the raw table properties. This PR adds a new internal SQL conf, `spark.sql.debug`, to enable debug mode and keep these raw table properties.

This config can also be used in similar places where we wanna retain debug information in the future.
## How was this patch tested?

new test in MetastoreDataSourcesSuite

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala
@@ -915,4 +915,9 @@ object StaticSQLConf {
     .internal()
     .intConf
     .createWithDefault(4000)
+
+  val DEBUG_MODE = buildConf("spark.sql.debug")
+    .internal()
+    .booleanConf
+    .createWithDefault(false)
 }
--- a/sql/core/src/test/scala/org/apache/spark/sql/internal/SQLConfSuite.scala
+++ b/sql/core/src/test/scala/org/apache/spark/sql/internal/SQLConfSuite.scala
@@ -19,6 +19,7 @@ package org.apache.spark.sql.internal
 
 import org.apache.hadoop.fs.Path
 
+import org.apache.spark.SparkContext
 import org.apache.spark.sql._
 import org.apache.spark.sql.execution.WholeStageCodegenExec
 import org.apache.spark.sql.internal.StaticSQLConf._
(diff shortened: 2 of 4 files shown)


cloud-fan:
cc @yhuai

rxin:
LGTM pending Jenkins.

sql/core/src/test/scala/org/apache/spark/sql/internal/SQLConfSuite.scala
```diff
@@ -254,18 +255,21 @@ class SQLConfSuite extends QueryTest with SharedSQLContext {
...
-  test("global SQL conf comes from SparkConf") {
-    val newSession = SparkSession.builder()
-      .config(SCHEMA_STRING_LENGTH_THRESHOLD.key, "2000")
-      .getOrCreate()
-
-    assert(newSession.conf.get(SCHEMA_STRING_LENGTH_THRESHOLD.key) == "2000")
-    checkAnswer(
-      newSession.sql(s"SET ${SCHEMA_STRING_LENGTH_THRESHOLD.key}"),
-      Row(SCHEMA_STRING_LENGTH_THRESHOLD.key, "2000"))
+  test("static SQL conf comes from SparkConf") {
```
cloud-fan: this change is not related to this PR.

`SparkSession.builder()....getOrCreate()` may not set the static SQL conf if there is an active or global default session. To make this test more robust, I explicitly create a new SparkSession.

This is the same approach I used in the new test, so I include this unrelated change in this PR.

andrewor14:
Cool beans. Merging into master 2.0.

andrewor14:
JK, actually it doesn't merge in 2.0.

gatorsmile:
If we turn this flag on, I am just afraid that some DDL statements might not work well. I assume this flag is really used for the our Spark SQL developers.

andrewor14:
Yes that's why it's `internal`

gatorsmile:
Just FYI. https://github.com/apache/spark/pull/15478 shows at least three DDL commands do not work properly when `spark.sql.debug` is set to `true`: `ALTER TABLE`, `ANALYZE TABLE` and `CREATE TABLE LIKE`

andrewor14:
I see. Then maybe we should add a comment above the config to note that several commands don't work (e.g. ALTER TABLE) if this is turned on, even if it's only internal.

sql/core/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala
```diff
@@ -915,4 +915,9 @@ object StaticSQLConf {
     .internal()
     .intConf
     .createWithDefault(4000)
+
+  val DEBUG_MODE = buildConf("spark.sql.debug")
+    .internal()
+    .booleanConf
+    .createWithDefault(false)
```
gatorsmile: Sure, I can do a quick PR to fix it.

PR #15460, opened 2016-10-13
[SPARK-17903][SQL] MetastoreRelation should talk to external catalog instead of hive client

cloud-fan:

`HiveExternalCatalog` should be the only interface to talk to the hive metastore. In `MetastoreRelation` we can just use `ExternalCatalog` instead of `HiveClient` to interact with hive metastore,  and add missing API in `ExternalCatalog`.
## How was this patch tested?

existing tests.

### DIFF
--- a/sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveExternalCatalog.scala
+++ b/sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveExternalCatalog.scala
@@ -32,6 +32,7 @@ import org.apache.spark.sql.AnalysisException
 import org.apache.spark.sql.catalyst.TableIdentifier
 import org.apache.spark.sql.catalyst.analysis.TableAlreadyExistsException
 import org.apache.spark.sql.catalyst.catalog._
+import org.apache.spark.sql.catalyst.expressions.Expression
 import org.apache.spark.sql.catalyst.plans.logical.{ColumnStat, Statistics}
 import org.apache.spark.sql.execution.command.{ColumnStatStruct, DDLUtils}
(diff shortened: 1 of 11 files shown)


cloud-fan:
cc @yhuai @gatorsmile

rxin:
cc @ericl

ericl:
LGTM

rxin:
A bunch of mysterious failures ...

ericl:
LGTM if tests pass :)

yhuai:
`Caused by: sbt.ForkMain$ForkError: org.apache.spark.sql.catalyst.analysis.NoSuchTableException: Table or view 'default' not found in database 'srcpart';` Seems we swapped the db and table name somehow.

sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveExternalCatalog.scala
```diff
@@ -636,6 +637,13 @@ private[spark] class HiveExternalCatalog(conf: SparkConf, hadoopConf: Configurat
     client.getPartitions(db, table, partialSpec)
   }
 
+  override def listPartitionsByFilter(
```
gatorsmile: Need a test case in `HiveExternalCatalogSuite`, just like the other APIs?

gatorsmile:
LGTM except one comment about test case

sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveExternalCatalog.scala
cloud-fan: added

cloud-fan:

PR #15464, opened 2016-10-13
[SPARK-17827][SQL]maxColLength type should be Int for String and Binary

robbinspg:

correct the expected type from Length function to be Int
## How was this patch tested?

Test runs on little endian and big endian platforms

### DIFF
--- a/sql/core/src/test/scala/org/apache/spark/sql/StatisticsColumnSuite.scala
+++ b/sql/core/src/test/scala/org/apache/spark/sql/StatisticsColumnSuite.scala
@@ -150,7 +150,7 @@ class StatisticsColumnSuite extends StatisticsTest {
       val colStat = ColumnStat(InternalRow(
         values.count(_.isEmpty).toLong,
         nonNullValues.map(_.length).sum / nonNullValues.length.toDouble,
-        nonNullValues.map(_.length).max.toLong,
+        nonNullValues.map(_.length).max.toInt,
         nonNullValues.distinct.length.toLong))
       (f, colStat)
     }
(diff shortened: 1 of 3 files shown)


sql/core/src/test/scala/org/apache/spark/sql/StatisticsColumnSuite.scala
```diff
@@ -150,7 +150,7 @@ class StatisticsColumnSuite extends StatisticsTest {
       val colStat = ColumnStat(InternalRow(
         values.count(_.isEmpty).toLong,
         nonNullValues.map(_.length).sum / nonNullValues.length.toDouble,
-        nonNullValues.map(_.length).max.toLong,
+        nonNullValues.map(_.length).max.toInt,
```
srowen: Does it even need `toInt`? I would presume it's already an int, but not sure.

srowen (commented):
OK by me if @wzhfy is OK with it

sql/core/src/test/scala/org/apache/spark/sql/StatisticsColumnSuite.scala
wzhfy: I think we can keep `toInt`, it explicitly shows the type.

wzhfy:
LGTM

wzhfy:
Did you check hive/StatisticsSuite? There's also a test case for column stats.

robbinspg:
This PR contains a change to  o.a.s.sql.hive.StatisticsSuite which I believe should fix that issue (awaiting big-endian build to complete)

robbinspg:
Tests all pass on big-endian with this PR

hvanhovell:
Merging to master. Thanks!

PR #15465, opened 2016-10-13
[SPARK-17909][SQL] we should create table before writing out the data in CTAS

cloud-fan:

In `CreateDataSourceTableAsSelectCommand`, we write out the data to the table location before create table, because we assume the data schema may change after it has been written out. However, we may only change the schema's nullability. We can still create table first, with regarding all columns are nullable.
## How was this patch tested?

existing tests.

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/command/createDataSourceTables.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/command/createDataSourceTables.scala
@@ -17,10 +17,11 @@
 
 package org.apache.spark.sql.execution.command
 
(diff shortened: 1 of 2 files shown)


cloud-fan:
cc @yhuai @gatorsmile

gatorsmile:
Not sure what is the reason in the original design. 

If we write data before creating the table, users will never get partial results. If we first create a table, and then write data. Users might get the partial result when querying the table before finishing the writing.

cloud-fan:
@gatorsmile that's a good point, although not documented. I'm closing this PR and will add this explanation in another PR.

PR #15466, opened 2016-10-13
[SPARK-13983][SQL] HiveThriftServer2 can not get "--hiveconf" or ''--hivevar" variables since 1.6 version (both multi-session and single session)

wangyum:

We can't get variables like [this way](https://github.com/pwendell/hive/blob/release-1.2.1-spark/ql/src/java/org/apache/hadoop/hive/ql/parse/VariableSubstitution.java#L38) after [SPARK-10810](https://issues.apache.org/jira/browse/SPARK-10810). but based on one SQLContext per session,  we can set variables to SQLContext. this is similar to [SPARK-15730](https://issues.apache.org/jira/browse/SPARK-15730)
 Thus, this PR is to set `--hivevar` and `--hiveconf` variables to `SQLContext` and then `VariableSubstitution` can substitution these variables.
## How was this patch tested?

Added test cases for spark-sql
and manual test beeline


wangyum:
@liancheng, Please help me review this PR, I have apply this patch to our spark cluster. 
spark-hive-var

wangyum:
@srowen please help to review, thanks a lot.

PR #15470, opened 2016-10-13
[SPARK-17921] failfast on checkpointLocation specified for memory streams

brkyvz:

The checkpointLocation option in memory streams in StructuredStreaming is not used during recovery. However, it can use this location if it is being set. However, during recovery, if this location was set, we get an exception saying that we will not use this location for recovery, please delete it.
It's better to just fail before you start the stream in the first place
## How was this patch tested?

Unit tests


brkyvz:
cc @zsxwing

sql/core/src/main/scala/org/apache/spark/sql/streaming/DataStreamWriter.scala
```diff
@@ -222,9 +222,13 @@ final class DataStreamWriter[T] private[sql](ds: Dataset[T]) {
 
       val sink = new MemorySink(df.schema, outputMode)
       val resultDf = Dataset.ofRows(df.sparkSession, new MemoryPlan(sink))
+      if (extraOptions.get("checkpointLocation").isDefined) {
+        throw new AnalysisException("Memory streams do not recover from checkpoints. Please " +
```
zsxwing: nit: this is memory sink

PR #15471, opened 2016-10-13
[SPARK-17919] Make timeout to RBackend configurable in SparkR

falaki:

This patch makes RBackend connection timeout configurable by user.
## How was this patch tested?

N/A

### DIFF
--- a/R/pkg/R/backend.R
+++ b/R/pkg/R/backend.R
@@ -111,7 +111,7 @@ invokeJava <- function(isStatic, objId, methodName, ...) {
   returnStatus <- readInt(conn)
   handleErrors(returnStatus, conn)
 
-  # Backend will send -1 as keep alive value to prevent various connection timeouts
+  # Backend will send +1 as keep alive value to prevent various connection timeouts
   # on very long running jobs. See spark.r.heartBeatInterval
   while (returnStatus == 1) {
     returnStatus <- readInt(conn)
--- a/R/pkg/R/backend.R
+++ b/R/pkg/R/backend.R
@@ -108,13 +108,27 @@ invokeJava <- function(isStatic, objId, methodName, ...) {
   conn <- get(".sparkRCon", .sparkREnv)
   writeBin(requestMessage, conn)
 
-  # TODO: check the status code to output error information
   returnStatus <- readInt(conn)
+  handleErrors(returnStatus, conn)
+
+  # Backend will send -1 as keep alive value to prevent various connection timeouts
+  # on very long running jobs. See spark.r.heartBeatInterval
+  while (returnStatus == -1) {
+    returnStatus <- readInt(conn)
+    handleErrors(returnStatus, conn)
+  }
+
+  readObject(conn)
+}
+
+# Helper function to check for returned errors and print appropriate error message to user
+handleErrors <- function(returnStatus, conn) {
   if (length(returnStatus) == 0) {
     stop("No status is returned. Java SparkR backend might have failed.")
   }
-  if (returnStatus != 0) {
+
+  # 0 is success and -1 is reserved for heartbeats. Positive values indicate errors.
+  if (returnStatus > 0) {
     stop(readString(conn))
   }
-  readObject(conn)
 }
--- a/R/pkg/R/backend.R
+++ b/R/pkg/R/backend.R
@@ -113,7 +113,7 @@ invokeJava <- function(isStatic, objId, methodName, ...) {
 
   # Backend will send -1 as keep alive value to prevent various connection timeouts
   # on very long running jobs. See spark.r.heartBeatInterval
-  while (returnStatus == -1) {
+  while (returnStatus == 1) {
     returnStatus <- readInt(conn)
     handleErrors(returnStatus, conn)
   }
@@ -127,8 +127,8 @@ handleErrors <- function(returnStatus, conn) {
     stop("No status is returned. Java SparkR backend might have failed.")
   }
 
-  # 0 is success and -1 is reserved for heartbeats. Positive values indicate errors.
-  if (returnStatus > 0) {
+  # 0 is success and +1 is reserved for heartbeats. Other negative values indicate errors.
+  if (returnStatus < 0) {
     stop(readString(conn))
   }
 }
--- a/R/pkg/inst/worker/worker.R
+++ b/R/pkg/inst/worker/worker.R
@@ -90,6 +90,7 @@ bootTime <- currentTimeSecs()
 bootElap <- elapsedSecs()
 
 rLibDir <- Sys.getenv("SPARKR_RLIBDIR")
+connectionTimeout <- Sys.getenv("SPARKR_BACKEND_CONNECTION_TIMEOUT")
 dirs <- strsplit(rLibDir, ",")[[1]]
 # Set libPaths to include SparkR package as loadNamespace needs this
 # TODO: Figure out if we can avoid this by not loading any objects that require
@@ -98,8 +99,10 @@ dirs <- strsplit(rLibDir, ",")[[1]]
 suppressPackageStartupMessages(library(SparkR))
 
 port <- as.integer(Sys.getenv("SPARKR_WORKER_PORT"))
-inputCon <- socketConnection(port = port, blocking = TRUE, open = "rb")
-outputCon <- socketConnection(port = port, blocking = TRUE, open = "wb")
+inputCon <- socketConnection(
+    port = port, blocking = TRUE, open = "rb", timeout = connectionTimeout)
+outputCon <- socketConnection(
+    port = port, blocking = TRUE, open = "wb", timeout = connectionTimeout)
 
 # read the index of the current partition inside the RDD
 partition <- SparkR:::readInt(inputCon)
--- a/R/pkg/inst/worker/worker.R
+++ b/R/pkg/inst/worker/worker.R
@@ -90,7 +90,7 @@ bootTime <- currentTimeSecs()
 bootElap <- elapsedSecs()
 
 rLibDir <- Sys.getenv("SPARKR_RLIBDIR")
-connectionTimeout <- Sys.getenv("SPARKR_BACKEND_CONNECTION_TIMEOUT")
+connectionTimeout <- as.integer(Sys.getenv("SPARKR_BACKEND_CONNECTION_TIMEOUT", "6000"))
 dirs <- strsplit(rLibDir, ",")[[1]]
 # Set libPaths to include SparkR package as loadNamespace needs this
 # TODO: Figure out if we can avoid this by not loading any objects that require
--- a/core/src/main/scala/org/apache/spark/api/r/RBackend.scala
+++ b/core/src/main/scala/org/apache/spark/api/r/RBackend.scala
@@ -22,7 +22,7 @@ import java.net.{InetAddress, InetSocketAddress, ServerSocket}
 import java.util.concurrent.TimeUnit
 
 import io.netty.bootstrap.ServerBootstrap
-import io.netty.channel.{ChannelFuture, ChannelInitializer, EventLoopGroup, ChannelOption}
+import io.netty.channel.{ChannelFuture, ChannelInitializer, ChannelOption, EventLoopGroup}
 import io.netty.channel.nio.NioEventLoopGroup
 import io.netty.channel.socket.SocketChannel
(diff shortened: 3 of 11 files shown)


felixcheung:
would it be possible to pull "6000" into a constant so it could be found/changed in one shot?

falaki:
@felixcheung that is a good suggestion. I will try to use a single constant.
I changed the label to WIP because something is still timing out the connection in my tests. Maybe very long timeouts are not properly implemented in R? If you noticed I am setting timeout on all the connections that are every opened. 

@shivaram I am going to experiment with a helper thread inside the main netty thread that sends heartbeats (-1 as result) and then `invokeJava()` would try reading again if it sees the heartbeat value. What do you think?

shivaram:
@falaki Thats an interesting approach to try -- the other thing to try might be to use a separate connection to send back results. i.e. as soon as JVM gets the request it returns `OK` or some such status and then whenever the result is ready the JVM pushes data on the reader socket that is created in R. We might need to try a couple of options here to see which looks best.

BTW the monitorConn is used - its used in cases where the JVM comes up first, say in YARN / spark-submit and the JVM uses that to detect if the R process has crashed.

falaki:
@shivaram thanks for clarification. 
I realized we were not setting socket timeout on Netty socket. So I added that as well.
I also introduced the heartbeat mechanism and tested it locally. Next I am going to test if this works on real workload.

falaki:
@shivaram this worked on my stress tests. The question is how to unit test this?

yhuai:
test this please

shivaram:
@falaki Do we know if the tests timeout are due to this change ? Or are they unrelated ?

falaki:
@shivaram I think they are unrelated. Can you trigger another test?

shivaram:
Jenkins, retest this please

shivaram:
@falaki looks like the SparkR MLlib unit tests are timing out on Jenkins. Do they pass on your machine ?

falaki:
@shivaram it was indeed my fault. I did not run local tests after I added the heartbeat. I am now using +1 for heartbeat.

core/src/main/scala/org/apache/spark/api/r/RBackendHandler.scala
```diff
@@ -83,7 +86,29 @@ private[r] class RBackendHandler(server: RBackend)
...
+        override def run(): Unit = {
+          val pingBaos = new ByteArrayOutputStream()
+          val pingDaos = new DataOutputStream(pingBaos)
+          writeInt(pingDaos, +1)
+          ctx.write(pingBaos.toByteArray)
+        }
+      }
+      val conf = new SparkConf()
+      val heartBeatInterval = conf.getInt(
+        "spark.r.heartBeatInterval", SparkRDefaults.DEFAULT_HEARTBEAT_INTERVAL)
```
felixcheung: should this be documented too?

core/src/main/scala/org/apache/spark/api/r/RBackendHandler.scala
```diff
@@ -83,7 +86,29 @@ private[r] class RBackendHandler(server: RBackend)
           writeString(dos, s"Error: unknown method $methodName")
       }
     } else {
+      // To avoid timeouts when reading results in SparkR driver, we will be regularly sending
+      // heartbeat responses. We use special character -1 to signal the client that backend is
```
felixcheung: `-1` -> `+1`?

R/pkg/R/backend.R
```diff
@@ -108,13 +108,27 @@ invokeJava <- function(isStatic, objId, methodName, ...) {
   conn <- get(".sparkRCon", .sparkREnv)
   writeBin(requestMessage, conn)
 
-  # TODO: check the status code to output error information
   returnStatus <- readInt(conn)
+  handleErrors(returnStatus, conn)
+
+  # Backend will send -1 as keep alive value to prevent various connection timeouts
```
felixcheung: `-1` -> `+1`?

core/src/main/scala/org/apache/spark/api/r/RBackendHandler.scala
```diff
@@ -95,9 +120,15 @@ private[r] class RBackendHandler(server: RBackend)
   }
 
   override def exceptionCaught(ctx: ChannelHandlerContext, cause: Throwable): Unit = {
-    // Close the connection when an exception is raised.
-    cause.printStackTrace()
-    ctx.close()
+    cause match {
+      case timeout: ReadTimeoutException =>
+        // Do nothing. We don't want to timeout on read
+        logInfo("Ignoring read timeout in RBackendHandler")
```
felixcheung: logWarning?

felixcheung (changes requested):
Looks good aside from minor comments.
This goes to 2.1.0?

falaki:
Thanks @felixcheung addressed your comments.

core/src/main/scala/org/apache/spark/api/r/RBackendHandler.scala
falaki: Done.

felixcheung:
LGTM

shivaram:
@falaki @felixcheung Since this is a big change I'd like to also take a look at this once - Will try to get to it tonight.

felixcheung:
sure. I think the target is 2.0.2 - it will be good to review this more closely.

falaki:
Thanks @shivaram. I ran a real workload consisting of long running parallel simulations that took about 3.5 hours. I also tested it by calling `Sys.sleep()` inside workers with dapply and spark.lapply. Would be good to torture it in new ways. 

Also would be interesting to think about ways of unit-testing this.

R/pkg/inst/worker/worker.R
```diff
@@ -90,6 +90,7 @@ bootTime <- currentTimeSecs()
 bootElap <- elapsedSecs()
 
 rLibDir <- Sys.getenv("SPARKR_RLIBDIR")
+connectionTimeout <- Sys.getenv("SPARKR_BACKEND_CONNECTION_TIMEOUT")
```
shivaram: Default value here as well

core/src/main/scala/org/apache/spark/api/r/RBackend.scala
```diff
@@ -110,6 +115,11 @@ private[spark] object RBackend extends Logging {
       val boundPort = sparkRBackend.init()
       val serverSocket = new ServerSocket(0, 1, InetAddress.getByName("localhost"))
       val listenPort = serverSocket.getLocalPort()
+      // Connection timeout is set by socket client. To make it configurable we will pass the
+      // timeout value to client inside the temp file
+      val conf = new SparkConf()
+      val backendConnectionTimeout = conf.getInt(
```
shivaram: Are we sure SparkConf has been initialized successfully at this point ? Or to to put it another way, in which cases does this code path get called from ? Is this in the `spark-submit` case or the shell etc. ?

core/src/main/scala/org/apache/spark/api/r/RRunner.scala
```diff
@@ -333,6 +333,8 @@ private[r] object RRunner {
     var rCommand = sparkConf.get("spark.sparkr.r.command", "Rscript")
     rCommand = sparkConf.get("spark.r.command", rCommand)
 
+    val rConnectionTimeout = SparkEnv.get.conf.getInt(
```
shivaram: can we just use `sparkConf.get` similar to the line above ?

core/src/main/scala/org/apache/spark/api/r/RBackendHandler.scala
```diff
@@ -83,7 +86,29 @@ private[r] class RBackendHandler(server: RBackend)
           writeString(dos, s"Error: unknown method $methodName")
       }
     } else {
+      // To avoid timeouts when reading results in SparkR driver, we will be regularly sending
+      // heartbeat responses. We use special code +1 to signal the client that backend is
+      // alive and it should continue blocking for result.
+      val execService = ThreadUtils.newDaemonSingleThreadScheduledExecutor("SparkRKeepAliveThread")
```
shivaram: I'm not sure how expensive it is to create and destroy an executor service each time. Can we just schedule at fixed rate when we get the request and then cancel the schedule at the end of the request ?

shivaram (commented):
Thanks @falaki -- This is a very useful change. I just did a pass of comments.

core/src/main/scala/org/apache/spark/api/r/RBackend.scala
falaki: This is for `spark-submit`. Basically the JVM starts before the R process. As a result the only way for R process to get these configuration parameters from the JVM. In this case, `RBackend` sets the environment variables based on configs.

For the other mode where JVM is started after the R process, we are sending this timeout value through the TCP connection.

At least that is my current understanding of how deploy modes work. In our production environment we launch the R process from the JVM.

core/src/main/scala/org/apache/spark/api/r/RBackendHandler.scala
falaki: I was not sure about this either. I used this method based on advice from @zsxwing.

falaki:
@shivaram sorry for delay getting back to this. Please take another look.

core/src/main/scala/org/apache/spark/api/r/RBackendHandler.scala
shivaram: Hmm - my question on whether we can reuse this still stands. @zsxwing do you think thats possible ?

zsxwing:
retest this please

falaki:
retest this please

falaki:
@shivaram is there a chance this makes it to the 2.0.2 release?

shivaram:
Taking another look now

core/src/main/scala/org/apache/spark/api/r/RBackendHandler.scala
shivaram: I took a closer look at this and (a) it looks like the executor just calls cancel on the tasks during shutdown, so that part of the behavior is the same as calling cancel on the task we have [1]. But you are right that if we wanted to wait for termination we'd need to do some extra work.  We could use the `get` call but its unclear what the semantics of that are. It might be more easier to just setup a semaphore or mutex that is shared by the runnable and the outside thread.

But overall it looks like thread pool creation is only around 100 micro-seconds [2] and I also benchmarked this locally.

[1] http://www.docjar.com/html/api/java/util/concurrent/ScheduledThreadPoolExecutor.java.html line 367
[2] http://stackoverflow.com/a/5483467/4577954

shivaram:
@falaki The code change looks pretty good to me, but I'm still a bit worried about introducing a big change in a minor release. Can we have this disabled by default and flip the flag only in the master branch ?

@HyukjinKwon Is there any way to retrigger the AppVeyor build ?

HyukjinKwon:
Build started: [SparkR] `ALL` [PR-15471](https://ci.appveyor.com/project/spark-test/spark/branch/1175779D-A053-45AF-BC6C-EA34931CFC37)
Diff: https://github.com/apache/spark/compare/master...spark-test:1175779D-A053-45AF-BC6C-EA34931CFC37

@shivaram I first thought committers can access to AppVeyor account but it seems not. We should go to the Web UI and click the rebuild button.. So, I just made a bunch of scripts to launch a build via @spark-test account in such cases. Please cc me (treat me like a bot) then I will leave such comments above until we find a good way to do it.

falaki:
@shivaram that is fine. We can merge it to 2.1 (or whatever the next major release is going to be).

shivaram:
Thanks @HyukjinKwon - The appveyor tests seem to pass as well

Change LGTM to merge to master. @felixcheung Any other comments ?

felixcheung:
LGTM.

felixcheung:
merged to master.

R/pkg/R/backend.R
```diff
@@ -108,13 +108,27 @@ invokeJava <- function(isStatic, objId, methodName, ...) {
...
+  # on very long running jobs. See spark.r.heartBeatInterval
+  while (returnStatus == 1) {
```
shivaram: @falaki @felixcheung any thoughts on this ?

felixcheung: +1 I think it's a good idea to avoid infinite loop in general.
how is toree calling this?
could you open a JIRA?

PR #15473, opened 2016-10-14
[SPARK-17925][SQL] Break fileSourceInterfaces.scala into multiple pieces

rxin:

This patch does a few changes to the file structure of data sources:
- Break fileSourceInterfaces.scala into multiple pieces (HadoopFsRelation, FileFormat, OutputWriter)
- Move ParquetOutputWriter into its own file

I created this as a separate patch so it'd be easier to review my future PRs that focus on refactoring this internal logic. This patch only moves code around, and has no logic changes.
## How was this patch tested?

N/A - should be covered by existing tests.


rxin:
cc @ericl @mallman  there's another small conflict here with your pr, but should be easy to resolve.

rxin:
cc @hvanhovell

cloud-fan:
LGTM, merging to master!

PR #15478, opened 2016-10-14
[SPARK-17899][SQL][DO-NOT-MERGE] Check the Impact When Turning On spark.sql.debug

gatorsmile:

**DO NOT MERGE**

This PR is just to do a quick check and see the impact of `spark.sql.debug`. Expect no actual failure should be seen.

N/A


gatorsmile:
`ALTER TABLE`, `ANALYZE TABLE` and `CREATE TABLE LIKE` do not work well when `spark.sql.debug` is set to true.

cloud-fan:
I don't think it's a problem, we will only set this flag when debugging.

gatorsmile:
Yeah, I see. This is just for showing the impact. Thanks!

PR #15479, opened 2016-10-14
[SPARK-17884][SQL] To resolve Null pointer exception when casting from  empty string to interval type

priyankagargnitk:

This change adds a check in castToInterval method of Cast expression , such that if converted value is null , then isNull variable should be set to true.

Earlier, the expression Cast(Literal(), CalendarIntervalType) was throwing NullPointerException because of the above mentioned reason.
## How was this patch tested?

Added test case in CastSuite.scala

jira entry for detail: https://issues.apache.org/jira/browse/SPARK-17884


rxin:
Thanks - merging. Can you close the ticket?

HyukjinKwon:
@priyankagargnitk Could maybe we then close this?

priyankagargnitk:
Yeah.. closing this.

PR #15480, opened 2016-10-14
[SPARK-16845][SQL] `GeneratedClass$SpecificOrdering` grows beyond 64 KB

lw-lin:

Prior to this patch, we'll generate `compare(...)` for `GeneratedClass$SpecificOrdering` like below, leading to Janino exceptions saying the code grows beyond 64 KB.

``` scala
/* 005 */ class SpecificOrdering extends o.a.s.sql.catalyst.expressions.codegen.BaseOrdering {
/* ..... */   ...
/* 10969 */   private int compare(InternalRow a, InternalRow b) {
/* 10970 */     InternalRow i = null;  // Holds current row being evaluated.
/* 10971 */
/* 1.... */     code for comparing field0
/* 1.... */     code for comparing field1
/* 1.... */     ...
/* 1.... */     code for comparing field449
/* 15012 */
/* 15013 */     return 0;
/* 15014 */   }
/* 15015 */ }
```

This patch would break `compare(...)` into smaller `compare_xxx(...)` methods when necessary; then we'll get generated `compare(...)` like:

``` scala
/* 001 */ public SpecificOrdering generate(Object[] references) {
/* 002 */   return new SpecificOrdering(references);
/* 003 */ }
/* 004 */
/* 005 */ class SpecificOrdering extends o.a.s.sql.catalyst.expressions.codegen.BaseOrdering {
/* 006 */
/* 007 */     ...
/* 1.... */
/* 11290 */   private int compare_0(InternalRow a, InternalRow b) {
/* 11291 */     InternalRow i = null;  // Holds current row being evaluated.
/* 11292 */
/* 11293 */     i = a;
/* 11294 */     boolean isNullA;
/* 11295 */     UTF8String primitiveA;
/* 11296 */     {
/* 11297 */
/* 11298 */       Object obj = ((Expression) references[0]).eval(null);
/* 11299 */       UTF8String value = (UTF8String) obj;
/* 11300 */       isNullA = false;
/* 11301 */       primitiveA = value;
/* 11302 */     }
/* 11303 */     i = b;
/* 11304 */     boolean isNullB;
/* 11305 */     UTF8String primitiveB;
/* 11306 */     {
/* 11307 */
/* 11308 */       Object obj = ((Expression) references[0]).eval(null);
/* 11309 */       UTF8String value = (UTF8String) obj;
/* 11310 */       isNullB = false;
/* 11311 */       primitiveB = value;
/* 11312 */     }
/* 11313 */     if (isNullA && isNullB) {
/* 11314 */       // Nothing
/* 11315 */     } else if (isNullA) {
/* 11316 */       return -1;
/* 11317 */     } else if (isNullB) {
/* 11318 */       return 1;
/* 11319 */     } else {
/* 11320 */       int comp = primitiveA.compare(primitiveB);
/* 11321 */       if (comp != 0) {
/* 11322 */         return comp;
/* 11323 */       }
/* 11324 */     }
/* 11325 */
/* 11326 */
/* 11327 */     i = a;
/* 11328 */     boolean isNullA1;
/* 11329 */     UTF8String primitiveA1;
/* 11330 */     {
/* 11331 */
/* 11332 */       Object obj1 = ((Expression) references[1]).eval(null);
/* 11333 */       UTF8String value1 = (UTF8String) obj1;
/* 11334 */       isNullA1 = false;
/* 11335 */       primitiveA1 = value1;
/* 11336 */     }
/* 11337 */     i = b;
/* 11338 */     boolean isNullB1;
/* 11339 */     UTF8String primitiveB1;
/* 11340 */     {
/* 11341 */
/* 11342 */       Object obj1 = ((Expression) references[1]).eval(null);
/* 11343 */       UTF8String value1 = (UTF8String) obj1;
/* 11344 */       isNullB1 = false;
/* 11345 */       primitiveB1 = value1;
/* 11346 */     }
/* 11347 */     if (isNullA1 && isNullB1) {
/* 11348 */       // Nothing
/* 11349 */     } else if (isNullA1) {
/* 11350 */       return -1;
/* 11351 */     } else if (isNullB1) {
/* 11352 */       return 1;
/* 11353 */     } else {
/* 11354 */       int comp = primitiveA1.compare(primitiveB1);
/* 11355 */       if (comp != 0) {
/* 11356 */         return comp;
/* 11357 */       }
/* 11358 */     }
/* 1.... */
/* 1.... */   ...
/* 1.... */
/* 12652 */     return 0;
/* 12653 */   }
/* 1.... */
/* 1.... */   ...
/* 15387 */
/* 15388 */   public int compare(InternalRow a, InternalRow b) {
/* 15389 */
/* 15390 */     int comp_0 = compare_0(a, b);
/* 15391 */     if (comp_0 != 0) {
/* 15392 */       return comp_0;
/* 15393 */     }
/* 15394 */
/* 15395 */     int comp_1 = compare_1(a, b);
/* 15396 */     if (comp_1 != 0) {
/* 15397 */       return comp_1;
/* 15398 */     }
/* 1.... */
/* 1.... */     ...
/* 1.... */
/* 15450 */     return 0;
/* 15451 */   }
/* 15452 */ }
```
## How was this patch tested?
- a new added test case which
  - would fail prior to this patch
  - would pass with this patch
- ordering correctness should already be covered by existing tests like those in `OrderingSuite`

## Acknowledgement

A major part of this PR - the refactoring work of `splitExpression()` - has been done by @ueshin.

### DIFF
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/codegen/CodeGenerator.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/codegen/CodeGenerator.scala
@@ -537,7 +537,6 @@ class CodegenContext {
       val funcCode: String =
         s"""
           public int $compareFunc(InternalRow a, InternalRow b) {
-            InternalRow i = null;
             $comparisons
             return 0;
           }
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/codegen/CodeGenerator.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/codegen/CodeGenerator.scala
@@ -610,6 +610,17 @@ class CodegenContext {
     splitExpressions(expressions, "apply", ("InternalRow", row) :: Nil)
   }
 
+  /**
+   * Splits the generated code of expressions into multiple functions, because function has
+   * 64kb code size limit in JVM
+   *
+   * @param expressions the codes to evaluate expressions.
+   * @param funcName the split function name base.
+   * @param arguments the list of (type, name) of the arguments of the split function.
+   * @param returnType the return type of the split function.
+   * @param makeSplitFunction makes split function body, e.g. add preparation or cleanup.
+   * @param foldFunctions folds the split function calls.
+   */
   def splitExpressions(
       expressions: Seq[String],
       funcName: String,
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/codegen/CodeGenerator.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/codegen/CodeGenerator.scala
@@ -610,8 +610,13 @@ class CodegenContext {
     splitExpressions(expressions, "apply", ("InternalRow", row) :: Nil)
   }
 
-  private def splitExpressions(
-      expressions: Seq[String], funcName: String, arguments: Seq[(String, String)]): String = {
+  def splitExpressions(
+      expressions: Seq[String],
+      funcName: String,
+      arguments: Seq[(String, String)],
+      returnType: String = "void",
+      makeSplitFunction: String => String = identity,
+      foldFunctions: Seq[String] => String = _.mkString("", ";\n", ";")): String = {
     val blocks = new ArrayBuffer[String]()
     val blockBuilder = new StringBuilder()
     for (code <- expressions) {
@@ -632,18 +637,19 @@ class CodegenContext {
       blocks.head
     } else {
       val func = freshName(funcName)
+      val argString = arguments.map { case (t, name) => s"$t $name" }.mkString(", ")
       val functions = blocks.zipWithIndex.map { case (body, i) =>
         val name = s"${func}_$i"
         val code = s"""
-           |private void $name(${arguments.map { case (t, name) => s"$t $name" }.mkString(", ")}) {
-           |  $body
+           |private $returnType $name($argString) {
+           |  ${makeSplitFunction(body)}
            |}
          """.stripMargin
         addNewFunction(name, code)
         name
       }
 
-      functions.map(name => s"$name(${arguments.map(_._2).mkString(", ")});").mkString("\n")
+      foldFunctions(functions.map(name => s"$name(${arguments.map(_._2).mkString(", ")})"))
     }
   }
 
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/codegen/CodeGenerator.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/codegen/CodeGenerator.scala
@@ -537,6 +537,7 @@ class CodegenContext {
       val funcCode: String =
         s"""
           public int $compareFunc(InternalRow a, InternalRow b) {
+            InternalRow i = null;
             $comparisons
             return 0;
           }
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/codegen/GenerateOrdering.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/codegen/GenerateOrdering.scala
@@ -72,7 +72,7 @@ object GenerateOrdering extends CodeGenerator[Seq[SortOrder], Ordering[InternalR
    * Generates the code for ordering based on the given order.
    */
   def genComparisons(ctx: CodegenContext, ordering: Seq[SortOrder]): String = {
-    val comparisons = ordering.map { order =>
+    def comparisons(orderingGroup: Seq[SortOrder]) = orderingGroup.map { order =>
       val eval = order.child.genCode(ctx)
       val asc = order.isAscending
       val isNullA = ctx.freshName("isNullA")
@@ -118,7 +118,45 @@ object GenerateOrdering extends CodeGenerator[Seq[SortOrder], Ordering[InternalR
           }
       """
     }.mkString("\n")
-    comparisons
+
+    /*
+     * 40 = 7000 bytes / 170 (around 170 bytes per ordering comparison).
+     * The maximum byte code size to be compiled for HotSpot is 8000 bytes.
+     * We should keep less than 8000 bytes.
+     */
+    val numberOfComparisonsThreshold = 40
+
+    if (ordering.size <= numberOfComparisonsThreshold) {
+      s"""
+         |  InternalRow ${ctx.INPUT_ROW} = null;  // Holds current row being evaluated.
+         |  ${comparisons(ordering)}
+      """.stripMargin
+    } else {
+      val groupedOrderingItr = ordering.grouped(numberOfComparisonsThreshold)
+      var groupedOrderingLength = 0
+      groupedOrderingItr.zipWithIndex.foreach { case (orderingGroup, i) =>
+        groupedOrderingLength += 1
+        val funcName = s"compare_$i"
+        val funcCode =
+          s"""
+             |private int $funcName(InternalRow a, InternalRow b) {
+             |  InternalRow ${ctx.INPUT_ROW} = null;  // Holds current row being evaluated.
+             |  ${comparisons(orderingGroup)}
+             |  return 0;
+             |}
+          """.stripMargin
+        ctx.addNewFunction(funcName, funcCode)
+      }
+
+      (0 to groupedOrderingLength - 1).map { i =>
+        s"""
+           |int comp_$i = compare_$i(a, b);
+           |if (comp_$i != 0) {
+           |  return comp_$i;
+           |}
+        """.stripMargin
+      }.mkString
+    }
   }
 
   protected def create(ordering: Seq[SortOrder]): BaseOrdering = {
@@ -142,7 +180,6 @@ object GenerateOrdering extends CodeGenerator[Seq[SortOrder], Ordering[InternalR
         ${ctx.declareAddedFunctions()}
 
         public int compare(InternalRow a, InternalRow b) {
-          InternalRow ${ctx.INPUT_ROW} = null;  // Holds current row being evaluated.
           $comparisons
           return 0;
         }
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/codegen/GenerateOrdering.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/codegen/GenerateOrdering.scala
@@ -133,10 +133,9 @@ object GenerateOrdering extends CodeGenerator[Seq[SortOrder], Ordering[InternalR
       """.stripMargin
     } else {
       val groupedOrderingItr = ordering.grouped(numberOfComparisonsThreshold)
-      var groupedOrderingLength = 0
-      groupedOrderingItr.zipWithIndex.foreach { case (orderingGroup, i) =>
-        groupedOrderingLength += 1
-        val funcName = s"compare_$i"
+      val funcNamePrefix = ctx.freshName("compare")
+      val funcNames = groupedOrderingItr.zipWithIndex.map { case (orderingGroup, i) =>
+        val funcName = s"${funcNamePrefix}_$i"
         val funcCode =
           s"""
              |private int $funcName(InternalRow a, InternalRow b) {
@@ -146,11 +145,12 @@ object GenerateOrdering extends CodeGenerator[Seq[SortOrder], Ordering[InternalR
              |}
           """.stripMargin
         ctx.addNewFunction(funcName, funcCode)
+        funcName
       }
 
-      (0 to groupedOrderingLength - 1).map { i =>
+      funcNames.zipWithIndex.map { case (funcName, i) =>
         s"""
-           |int comp_$i = compare_$i(a, b);
+           |int comp_$i = ${funcName}(a, b);
            |if (comp_$i != 0) {
            |  return comp_$i;
            |}
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/codegen/GenerateOrdering.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/codegen/GenerateOrdering.scala
@@ -132,12 +132,12 @@ object GenerateOrdering extends CodeGenerator[Seq[SortOrder], Ordering[InternalR
         """
       },
       foldFunctions = { funCalls =>
-        val comp = ctx.freshName("comp")
         funCalls.zipWithIndex.map { case (funCall, i) =>
+          val comp = ctx.freshName("comp")
           s"""
-            int ${comp}_$i = $funCall;
-            if (${comp}_$i != 0) {
-              return ${comp}_$i;
+            int $comp = $funCall;
+            if ($comp != 0) {
+              return $comp;
             }
           """
         }.mkString
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/codegen/GenerateOrdering.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/codegen/GenerateOrdering.scala
@@ -72,7 +72,7 @@ object GenerateOrdering extends CodeGenerator[Seq[SortOrder], Ordering[InternalR
    * Generates the code for ordering based on the given order.
    */
   def genComparisons(ctx: CodegenContext, ordering: Seq[SortOrder]): String = {
-    def comparisons(orderingGroup: Seq[SortOrder]) = orderingGroup.map { order =>
+    val comparisons = ordering.map { order =>
       val eval = order.child.genCode(ctx)
       val asc = order.isAscending
       val isNullA = ctx.freshName("isNullA")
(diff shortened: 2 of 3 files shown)


HyukjinKwon:
Maybe we could run retest this uf you are very sure on this failure is due to a flaky test.

lw-lin:
Flaky test I think not related to this PR. Thanks, @HyukjinKwon !

Jenkins retest this please.

lw-lin:
Jenkins retest this please

lw-lin:
@davies @kiszk it'd be great if you can take a look

rxin:
cc @ueshin want to help review this?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/codegen/GenerateOrdering.scala
```diff
@@ -118,7 +118,45 @@ object GenerateOrdering extends CodeGenerator[Seq[SortOrder], Ordering[InternalR
           }
       """
     }.mkString("\n")
-    comparisons
+
+    /*
```
kiszk: Instead of this implementation, is it possible to use [`this function`](https://github.com/apache/spark/blob/b1b47274bfeba17a9e4e9acebd7385289f31f6c8/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/codegen/CodeGenerator.scala#L613) by adding `return` as a default argument?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/codegen/GenerateOrdering.scala
```diff
@@ -118,7 +118,45 @@ object GenerateOrdering extends CodeGenerator[Seq[SortOrder], Ordering[InternalR
...
+      s"""
+         |  InternalRow ${ctx.INPUT_ROW} = null;  // Holds current row being evaluated.
+         |  ${comparisons(ordering)}
+      """.stripMargin
+    } else {
+      val groupedOrderingItr = ordering.grouped(numberOfComparisonsThreshold)
+      var groupedOrderingLength = 0
+      groupedOrderingItr.zipWithIndex.foreach { case (orderingGroup, i) =>
+        groupedOrderingLength += 1
+        val funcName = s"compare_$i"
```
ueshin: We need to use fresh name for `funcName` or its prefix (see [here](https://github.com/apache/spark/blob/b1b47274bfeba17a9e4e9acebd7385289f31f6c8/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/codegen/CodeGenerator.scala#L634)).

ueshin (changes requested):

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/codegen/GenerateOrdering.scala
```diff
@@ -118,7 +118,45 @@ object GenerateOrdering extends CodeGenerator[Seq[SortOrder], Ordering[InternalR
...
+             |private int $funcName(InternalRow a, InternalRow b) {
+             |  InternalRow ${ctx.INPUT_ROW} = null;  // Holds current row being evaluated.
+             |  ${comparisons(orderingGroup)}
+             |  return 0;
+             |}
+          """.stripMargin
+        ctx.addNewFunction(funcName, funcCode)
+      }
+
+      (0 to groupedOrderingLength - 1).map { i =>
```
ueshin: nit: use `(0 until groupedOrderingLength)`.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/codegen/GenerateOrdering.scala
lw-lin: Thanks @kiszk for bringing this up! @ueshin @kiszk any comments please on do we want to expand `CodeGenerator#private splitExpressions` to:
 - add one more argument to specific how we want to "return" things?
 - add arguments to specific in which manner do we want to break expression? E.g. in the pr's case, we might want to break it according to the `numberOfComparisonsThreshold = 40` rather than on the string length

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/codegen/GenerateOrdering.scala
lw-lin: let me fix this, thanks!

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/codegen/GenerateOrdering.scala
ueshin: I agree with @kiszk that we would use the function if possible, but I have no idea to expand the function to apply to this case simply.

kiszk: @lw-lin , good points.
For the second issue, can we use the string length as a proxy?
For the first issue, how about the following approach? In advance, I am sorry that I have not compiled it myself.

```java
// CodeGenerator.scalar
def splitExpressions(expressions: Seq[String], funcName: String, arguments: Seq[(String, String), returns: (String, String] =("void", "")): String = {
  ..
  val code = s"""
    |private $(returns._1) $name(${arguments.map { case (t, name) => s"$t $name" }.mkString(", ")}) {
    |  $body
    |  ${returns._2)
    |}
  """.stripMargin
...
}
```

```java
// GenerateOrdering.scala 
  val groupedOrderingItr = ordering.grouped(numberOfComparisonsThreshold)
  // var groupedOrderingLength = 0
  val functions = ctx.splitExpressions {
    ..
  }
  val comp = freshName("comp")
  functions.zipWithIndex.map { case(func , i) =>
    val name = s"${comp}_$i"
    s"""
      |int $name = $func(a, b);
      |if ($name != 0) {
      |  return $name;
      |}
    """.stripMargin
  }.mkString
```

lw-lin: @kiszk @ueshin
i kind of implemented a prove of concept (https://github.com/lw-lin/spark/commit/d0c11984e9aa84f68b9ebdd21317f629d759ae86) on how to extend `splitExpressions` to rewrite this PR
could you comment?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/codegen/CodeGenerator.scala
```diff
@@ -537,7 +537,6 @@ class CodegenContext {
       val funcCode: String =
         s"""
           public int $compareFunc(InternalRow a, InternalRow b) {
-            InternalRow i = null;
```
ueshin: I think this line should not be moved into `GenerateOrdering.genComparisons()` if there is a special reason.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/codegen/GenerateOrdering.scala
lw-lin: @davies could you take a quick look here:

this PR tries to break a huge generated method into smaller pieces. @kiszk @ueshin and I were discussing whether we should:
- (a) just do the breaking-ups case by case, like in this PR
- or (b) expand `CodeGenerator.splitExpressions(...)` to generally support this breaking-ups, like in the POC https://github.com/lw-lin/spark/commit/d0c11984e9aa84f68b9ebdd21317f629d759ae86

@davies could you advise on this? thanks!

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/codegen/CodeGenerator.scala
lw-lin: thanks -- but not removing it from here would lead to declaring a local variable `i` that we might not use when `(ordering.size > numberOfComparisonsThreshold) `. what do you think?

lw-lin:
Jenkins retest this please

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/codegen/CodeGenerator.scala
ueshin: I think `GenerateOrdering.genComparisons()` is expected to generate a PART of comparison method. If the declaration of the variable `i` is in the part, we can't use `GenerateOrdering.genComparisons()` twice or more for the same method.
Declaration of unused variable will not be a problem.

lw-lin: Oh I see, makes a lot of sense. Let me fix this, thanks!

lw-lin:
Jenkins retest this please

kiszk:
Jenkins retest this please

lw-lin:
@hvanhovell it'd be great if you can take a look at this, thanks!

lw-lin:
@cloud-fan @hvanhovell would you take a look at this? Seems like it's targeted for 2.1. Thanks!

lw-lin:
Jenkins retest this please

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/codegen/GenerateOrdering.scala
```diff
@@ -118,7 +118,42 @@ object GenerateOrdering extends CodeGenerator[Seq[SortOrder], Ordering[InternalR
...
+      val groupedOrderingItr = ordering.grouped(numberOfComparisonsThreshold)
+      val funcNamePrefix = ctx.freshName("compare")
+      val funcNames = groupedOrderingItr.zipWithIndex.map { case (orderingGroup, i) =>
+        val funcName = s"${funcNamePrefix}_$i"
+        val funcCode =
+          s"""
+             |private int $funcName(InternalRow a, InternalRow b) {
+             |  InternalRow ${ctx.INPUT_ROW} = null;  // Holds current row being evaluated.
+             |  ${comparisons(orderingGroup)}
+             |  return 0;
```
cloud-fan: if we make the comparison result a member variable, then we don't need `return` in the comparison code right?

lw-lin: Yea that's right -- here we're returning ints because comparisons(ordering/orderingGroup)(L75 of this same file) is returning ints. Should we change that all along?

cloud-fan: for performance concerns, we should avoid using member variables. If there is no easy way to reuse `splitExpressions`, I'm ok with the current approach.

lw-lin: Ah then maybe this is ready to go. I recall @ueshin or @kiszk might be interested in refactoring `splitExpressions` in future PRs.

Thanks for the review!

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/codegen/GenerateOrdering.scala
```diff
@@ -118,7 +118,42 @@ object GenerateOrdering extends CodeGenerator[Seq[SortOrder], Ordering[InternalR
           }
       """
     }.mkString("\n")
-    comparisons
+
+    /*
+     * 40 = 7000 bytes / 170 (around 170 bytes per ordering comparison).
```
cloud-fan: how do you get the `170`?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/codegen/GenerateOrdering.scala
ueshin: I'm interested in refactoring approach, which will be useful for more general case.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/codegen/GenerateOrdering.scala
lw-lin: Ah sorry this is vague.

Actually I ran all the 36 test cases in `OrderingSuite` and logged each of the generated `comparisonXXX` method's size. Each method:
- ＜ 4.9 KB when `numberOfComparisonsThreshold = 40`
- ＜ 6.4 KB when `numberOfComparisonsThreshold = 50`
- ＜ 7.8 KB when `numberOfComparisonsThreshold = 60`

I thought it'd be safer if we pick 40 (taking minor future changes into account). Thus 170 should be considered as some kind of a safe assumption (or not?).

Would you share your thoughts on this?

cloud-fan: test cases are not real world workloads, we can't estimate comparison code size based on test cases. My idea is that, first we generate the comparison code like before, and check the code size, if it exceeds 1024(see https://github.com/apache/spark/pull/15620#discussion_r90185362), go to the splitting branch.

ueshin: If you use `ctx.splitExpressions()` approach, we don't need to calculate the size because the method will split based on the code size of 1024.

ueshin: I tried to refactor `ctx.splitExpressions()` like 0aedc47fa94366a39e66da82b70c79dccac7d375 based on @lw-lin's PoC.
Unfortunately, the max number of comparisons we can support declined to around `5000` (current approach can support `10000` or more) because `ctx.splitExpressions()` splits comparisons into smaller size than current rule.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/codegen/GenerateOrdering.scala
kiszk: I am also interested in refactoring. However, it would be better to do in another PR.

mallman:
Hi @lw-lin. Just FYI we use this patch at VideoAmp and would love to see it merged in. I notice this PR has gone a little cold. I'm sorry I can't offer much concrete help, but I wanted to check with you to see if you'll be able to pick this up again soon. Cheers.

lw-lin:
Hi @mallman, I'll pick up this within this week. Thanks for the feedback! :)

lw-lin:
Jenkins retest this please

lw-lin:
I've cherry-picked the refactoring work of `splitExpression` (from @ueshin -- thank you!) into this. Also test passed.
So @cloud-fan would you take a look at this again at your convenience? Thanks!

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/codegen/GenerateOrdering.scala
```diff
@@ -117,8 +117,31 @@ object GenerateOrdering extends CodeGenerator[Seq[SortOrder], Ordering[InternalR
...
+          InternalRow ${ctx.INPUT_ROW} = null;  // Holds current row being evaluated.
+          $body
+          return 0;
+        """
+      },
+      foldFunctions = { funCalls =>
+        val comp = ctx.freshName("comp")
+        funCalls.zipWithIndex.map { case (funCall, i) =>
+          s"""
+            int ${comp}_$i = $funCall;
```
cloud-fan: nit: `ctx.freshName` already adds postfix to the name, you don't need to add `_$i` again.

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/OrderingSuite.scala
```diff
@@ -127,4 +127,17 @@ class OrderingSuite extends SparkFunSuite with ExpressionEvalHelper {
...
   }
+
+  test("SPARK-16845: GeneratedClass$SpecificOrdering grows beyond 64 KB") {
+    val sortOrder = Literal("abc").asc
+
+    // this is passing prior to SPARK-16845, and it should also be passing after SPARK-16845
+    GenerateOrdering.generate(Array.fill(40)(sortOrder))
+
+    // this is FAILING prior to SPARK-16845, but it should be passing after SPARK-16845
+    GenerateOrdering.generate(Array.fill(450)(sortOrder))
```
cloud-fan: This is unnecessary, it's covered by the `5000` test case.

cloud-fan:
LGTM

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/codegen/GenerateOrdering.scala
lw-lin: It's `comp_0`, `comp_1` in the following:
```scala
/* 15388 */   public int compare(InternalRow a, InternalRow b) {
/* 15389 */
/* 15390 */     int comp_0 = compare_0(a, b);
/* 15391 */     if (comp_0 != 0) {
/* 15392 */       return comp_0;
/* 15393 */     }
/* 15394 */
/* 15395 */     int comp_1 = compare_1(a, b);
/* 15396 */     if (comp_1 != 0) {
/* 15397 */       return comp_1;
/* 15398 */     }
/* 1.... */
/* 1.... */     ...
/* 1.... */
/* 15450 */     return 0;
/* 15451 */   }
```
so maybe we should keep this `_$i`?

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/OrderingSuite.scala
lw-lin: Sure, I'll remove the `450` test case

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/codegen/GenerateOrdering.scala
cloud-fan: can you double check? the implementation of `freshName` is
```
if (freshNameIds.contains(fullName)) {
      val id = freshNameIds(fullName)
      freshNameIds(fullName) = id + 1
      s"$fullName$id"
    } else {
      freshNameIds += fullName -> 1
      fullName
    }
```
it already adds an id postfix.

lw-lin: Ah I mis-understood it. You meant moving `ctx.freshName("comp")` into the `funCalls.zipWithIndex.map {...}`, right? like the following:

```scala
// val comp = ctx.freshName("comp")                 // moved this into the map {...}
funCalls.zipWithIndex.map { case (funCall, i) =>
  s"""
    int ${ctx.freshName("comp")} = $funCall;
 ...
}
```

cloud-fan: I mean
```
val comp = ctx.freshName("comp")
funCalls.zipWithIndex.map { case (funCall, i) =>
  s"""
    int $comp = $funCall;
 ...
}
```
just remove all the `_$i` here

lw-lin: thanks for clarifying on this. but we'll get something like (suppose we get a fresh name `comp_1`):
```java
/* 15388 */   public int compare(InternalRow a, InternalRow b) {
/* 15389 */
/* 15390 */     int comp_1 = compare_0(a, b);     // comp_1 
/* 15391 */     if (comp_1 != 0) {
/* 15392 */       return comp_1;
/* 15393 */     }
/* 15394 */
/* 15395 */     int comp_1 = compare_1(a, b);     // still comp_1 
/* 15396 */     if (comp_1 != 0) {
/* 15397 */       return comp_1;
/* 15398 */     }
/* 1.... */
/* 1.... */     ...
/* 1.... */
/* 15450 */     return 0;
/* 15451 */   }
```

cloud-fan: oh sorry, we should put the `freshName` in the map function
```
funCalls.zipWithIndex.map { case (funCall, i) =>
  val comp = ctx.freshName("comp")
  s"""
    int $comp = $funCall;
 ...
}
```

lw-lin: Ah i see - let me update this :-)

lw-lin:
Jenkins retest this please

cloud-fan:

dsimmie:
Is there any plan to apply this fix to 1.6? (am using the Cloudera version `spark-1.6.0-cdh5.9.0`) and am seeing this problem with running a `countDistinct` over a `DF` with 600 columns.

MGwynne:
Same question as @dsimmie - we are having the same issues for 1.6.2 and can't upgrade to 2.0 immediately - would be good to get this backported.

PR #9963, opened 2015-11-25
[SPARK-10582][Yarn][Core] Fix AM failure situation for dynamic allocation

jerryshao:
Because of AM failure, the target executor number between driver and AM will be different, which will lead to unexpected behavior in dynamic allocation. So when AM is re-registered with driver, state in `ExecutorAllocationManager` and `CoarseGrainedSchedulerBacked` should be reset.

This issue is originally addressed in #8737 , here re-opened again. Thanks a lot @KaiXinXiaoLei for finding this issue.

@andrewor14 and @vanzin would you please help to review this, thanks a lot.

PR #15481, opened 2016-10-14
[SPARK-17929] [CORE] Fix deadlock when CoarseGrainedSchedulerBackend reset

scwf:

https://issues.apache.org/jira/browse/SPARK-17929

Now `CoarseGrainedSchedulerBackend` reset will get the lock,

```
  protected def reset(): Unit = synchronized {
    numPendingExecutors = 0
    executorsPendingToRemove.clear()

    // Remove all the lingering executors that should be removed but not yet. The reason might be
    // because (1) disconnected event is not yet received; (2) executors die silently.
    executorDataMap.toMap.foreach { case (eid, _) =>
      driverEndpoint.askWithRetry[Boolean](
        RemoveExecutor(eid, SlaveLost("Stale executor after cluster manager re-registered.")))
    }
  }
```

 but on removeExecutor also need the lock "CoarseGrainedSchedulerBackend.this.synchronized", this will cause deadlock.

```
   private def removeExecutor(executorId: String, reason: ExecutorLossReason): Unit = {
      logDebug(s"Asked to remove executor $executorId with reason $reason")
      executorDataMap.get(executorId) match {
        case Some(executorInfo) =>
          // This must be synchronized because variables mutated
          // in this block are read when requesting executors
          val killed = CoarseGrainedSchedulerBackend.this.synchronized {
            addressToExecutorId -= executorInfo.executorAddress
            executorDataMap -= executorId
            executorsPendingLossReason -= executorId
            executorsPendingToRemove.remove(executorId).getOrElse(false)
          }
     ...

## How was this patch tested?

manual test.

```

### DIFF
--- a/core/src/main/scala/org/apache/spark/scheduler/cluster/CoarseGrainedSchedulerBackend.scala
+++ b/core/src/main/scala/org/apache/spark/scheduler/cluster/CoarseGrainedSchedulerBackend.scala
@@ -92,18 +92,18 @@ class CoarseGrainedSchedulerBackend(scheduler: TaskSchedulerImpl, val rpcEnv: Rp
   // The num of current max ExecutorId used to re-register appMaster
   @volatile protected var currentExecutorIdCounter = 0
 
+  // Executors that have been lost, but for which we don't yet know the real exit reason.
+  protected val executorsPendingLossReason = new HashSet[String]
+
+  protected val addressToExecutorId = new HashMap[RpcAddress, String]
+
   class DriverEndpoint(override val rpcEnv: RpcEnv, sparkProperties: Seq[(String, String)])
     extends ThreadSafeRpcEndpoint with Logging {
-
-    // Executors that have been lost, but for which we don't yet know the real exit reason.
-    protected val executorsPendingLossReason = new HashSet[String]
 
     // If this DriverEndpoint is changed to support multiple threads,
     // then this may need to be changed so that we don't share the serializer
     // instance across threads
     private val ser = SparkEnv.get.closureSerializer.newInstance()
-
-    protected val addressToExecutorId = new HashMap[RpcAddress, String]
 
     private val reviveThread =
       ThreadUtils.newDaemonSingleThreadScheduledExecutor("driver-revive-thread")
@@ -392,9 +392,23 @@ class CoarseGrainedSchedulerBackend(scheduler: TaskSchedulerImpl, val rpcEnv: Rp
 
     // Remove all the lingering executors that should be removed but not yet. The reason might be
     // because (1) disconnected event is not yet received; (2) executors die silently.
-    executorDataMap.toMap.foreach { case (eid, _) =>
-      driverEndpoint.send(
-        RemoveExecutor(eid, SlaveLost("Stale executor after cluster manager re-registered.")))
+    // Note: here copy the code of remove executor from DriverEndpoint to avoid deadlock(reset
+    // and removeExecutor both to get the lock of schedulerbackend.)
+    val reason = SlaveLost("Stale executor after cluster manager re-registered.")
+    executorDataMap.toMap.foreach { case (executorId, executorInfo) =>
+      executorInfo.executorEndpoint.send(StopExecutor)
+      logDebug(s"Asked to remove executor $executorId with reason $reason")
+      val killed = {
+        addressToExecutorId -= executorInfo.executorAddress
+        executorDataMap -= executorId
+        executorsPendingLossReason -= executorId
+        executorsPendingToRemove.remove(executorId).getOrElse(false)
+      }
+      totalCoreCount.addAndGet(-executorInfo.totalCores)
+      totalRegisteredExecutors.addAndGet(-1)
+      scheduler.executorLost(executorId, if (killed) ExecutorKilled else reason)
+      listenerBus.post(
+        SparkListenerExecutorRemoved(System.currentTimeMillis(), executorId, reason.toString))
     }
   }
 
--- a/core/src/main/scala/org/apache/spark/scheduler/cluster/CoarseGrainedSchedulerBackend.scala
+++ b/core/src/main/scala/org/apache/spark/scheduler/cluster/CoarseGrainedSchedulerBackend.scala
@@ -393,7 +393,7 @@ class CoarseGrainedSchedulerBackend(scheduler: TaskSchedulerImpl, val rpcEnv: Rp
     // Remove all the lingering executors that should be removed but not yet. The reason might be
     // because (1) disconnected event is not yet received; (2) executors die silently.
     executorDataMap.toMap.foreach { case (eid, _) =>
-      driverEndpoint.askWithRetry[Boolean](
+      driverEndpoint.send(
         RemoveExecutor(eid, SlaveLost("Stale executor after cluster manager re-registered.")))
     }
   }
--- a/core/src/main/scala/org/apache/spark/scheduler/cluster/CoarseGrainedSchedulerBackend.scala
+++ b/core/src/main/scala/org/apache/spark/scheduler/cluster/CoarseGrainedSchedulerBackend.scala
@@ -145,6 +145,9 @@ class CoarseGrainedSchedulerBackend(scheduler: TaskSchedulerImpl, val rpcEnv: Rp
             // Ignoring the task kill since the executor is not registered.
             logWarning(s"Attempted to kill task $taskId for unknown executor $executorId.")
         }
+
+      case RemoveExecutor(executorId, reason) =>
+        removeExecutor(executorId, reason)
     }
 
     override def receiveAndReply(context: RpcCallContext): PartialFunction[Any, Unit] = {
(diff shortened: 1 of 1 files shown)


srowen:
@zsxwing or @andrewor14 might know best on this one.

mridulm:
Would be cleaner to simply copy executorDataMap.keys and works off that to ensure there is no coupling between actor thread and invoker.

jerryshao:
LGTM, sorry to bring in deadlock issue.

scwf:
retest this please.

scwf:
retest this please.

core/src/main/scala/org/apache/spark/scheduler/cluster/CoarseGrainedSchedulerBackend.scala
```diff
@@ -392,9 +392,23 @@ class CoarseGrainedSchedulerBackend(scheduler: TaskSchedulerImpl, val rpcEnv: Rp
 
     // Remove all the lingering executors that should be removed but not yet. The reason might be
     // because (1) disconnected event is not yet received; (2) executors die silently.
-    executorDataMap.toMap.foreach { case (eid, _) =>
-      driverEndpoint.askWithRetry[Boolean](
-        RemoveExecutor(eid, SlaveLost("Stale executor after cluster manager re-registered.")))
+    // Note: here copy the code of remove executor from DriverEndpoint to avoid deadlock(reset
+    // and removeExecutor both to get the lock of schedulerbackend.)
+    val reason = SlaveLost("Stale executor after cluster manager re-registered.")
+    executorDataMap.toMap.foreach { case (executorId, executorInfo) =>
```
zsxwing: `executorDataMap` should not be modified outside DriverEndpoint. See the comment for `executorDataMap`.

core/src/main/scala/org/apache/spark/scheduler/cluster/CoarseGrainedSchedulerBackend.scala
```diff
@@ -92,19 +92,19 @@ class CoarseGrainedSchedulerBackend(scheduler: TaskSchedulerImpl, val rpcEnv: Rp
   // The num of current max ExecutorId used to re-register appMaster
   @volatile protected var currentExecutorIdCounter = 0
 
+  // Executors that have been lost, but for which we don't yet know the real exit reason.
+  protected val executorsPendingLossReason = new HashSet[String]
```
zsxwing: they were in `DriverEndpoint` to make sure they won't be accessed outside `DriverEndpoint` since they are not thread-safe. This changes break it.

zsxwing (changes requested):
`reset` is called in YarnSchedulerEndpoint which ideally should not be a blocking action.

@jerryshao Can we just make it fire and forget?

mridulm:
@scwf I think the initial fix with a small change might be sufficient.
What I meant was something like this :

```
protected def reset(): Unit = {
  val executors = synchronized {
    Set() ++ executorDataMap.keys
  }

  executors.foreach ( eid => driverEndpoint.askWithRetry[Boolean](
    RemoveExecutor(eid, SlaveLost("Stale executor after cluster manager re-registered.")))
}
```


This will keep the synchronized block restricted to copying the executor keys, and leave the rest as-is removing the deadlock. Thoughts ?

zsxwing:
> reset is called in YarnSchedulerEndpoint which ideally should not be a blocking action.

I'm wondering if we can also fix this.

mridulm:
@zsxwing Ah, then simply making it send() instead of askWithRetry() should do, no ?
That was actually in the initial PR - I was not sure if we want to change the behavior from askWithRetry to send ...

jerryshao:
Seems it could be changed to `send` instead.

scwf:
ok, i will revert to the initial commit.

core/src/main/scala/org/apache/spark/scheduler/cluster/CoarseGrainedSchedulerBackend.scala
```diff
@@ -393,7 +393,7 @@ class CoarseGrainedSchedulerBackend(scheduler: TaskSchedulerImpl, val rpcEnv: Rp
     // Remove all the lingering executors that should be removed but not yet. The reason might be
     // because (1) disconnected event is not yet received; (2) executors die silently.
     executorDataMap.toMap.foreach { case (eid, _) =>
-      driverEndpoint.askWithRetry[Boolean](
+      driverEndpoint.send(
```
viirya: When changed to `send`, we don't do retry anymore. It may become less tolerant.

jerryshao: This will only be invoked in yarn client mode when AM is failed and some lingering executors are existed, such situation may not be happened normally as I know. So I think it should be OK to call `send`.

zsxwing: This call is actually just sending the message to the same process. retry is not needed.

jerryshao: Also IIUC `driverEndpoint` is a in-process endpoint, so it should be safe.

mridulm: This is synchronizing on the entire method which is not required - we need to restrict synchronized block to the subset which is actually required - the copying of the executors to remove.

The snippet I posted earlier does the same - can you please modify it accordingly ?
If we had coded it that way initially, this bug wouldn't have existed to begin with (with or without send/askWithRetry)

viirya: Once we change this to `send`, don't we need to move the handling of `RemoveExecutor` to `receive` in `DriverEndpoint`?

zsxwing: Good catch. I think here we can just call `removeExecutor(...)`.

scwf:
Updated, can you review again?

mridulm:
LGTM, @zsxwing any comments ?

mridulm:
BTW, it was interesting that the earlier change did not trigger a test failure (the issue @viirya pointed out - about needing to move RemoveExecutor to receive)

viirya:
@mridulm I checked #9963 and looks like we don't test against `CoarseGrainedSchedulerBackend.reset`.

core/src/main/scala/org/apache/spark/scheduler/cluster/CoarseGrainedSchedulerBackend.scala
```diff
@@ -145,6 +145,9 @@ class CoarseGrainedSchedulerBackend(scheduler: TaskSchedulerImpl, val rpcEnv: Rp
             // Ignoring the task kill since the executor is not registered.
             logWarning(s"Attempted to kill task $taskId for unknown executor $executorId.")
         }
+
+      case RemoveExecutor(executorId, reason) =>
+        removeExecutor(executorId, reason)
```
viirya: Why remove `executorDataMap.get(executorId).foreach(_.executorEndpoint.send(StopExecutor))`?

zsxwing:
Sorry that my comment was unclear. I meant:
```Scala
  protected def reset(): Unit = synchronized {
    numPendingExecutors = 0
    executorsPendingToRemove.clear()

    // Remove all the lingering executors that should be removed but not yet. The reason might be
    // because (1) disconnected event is not yet received; (2) executors die silently.
    executorDataMap.toMap.foreach { case (eid, _) =>
      removeExecutor(eid, SlaveLost("Stale executor after cluster manager re-registered."))
    }
  }
```

mridulm:
@zsxwing To minimize scope of synchronized block.
The way @scwf has now, the synchronized block is limited to duplicating key and setting some state.
Remaining can happen outside of the lock.

zsxwing:
@mridulm `ask` is very cheap. It just puts the serialized message into a buffer. The current codes now need to duplicate the codes and as @viirya pointed out, it misses `executorDataMap.get(executorId).foreach(_.executorEndpoint.send(StopExecutor))`.

How about just the following changes? I agreed that `removeExecutor` doesn't need to be in the `synchronized` block.
```
  protected def reset(): Unit = {
    val executors = synchronized {
      numPendingExecutors = 0
      executorsPendingToRemove.clear()
      Set() ++ executorDataMap.keys
    }

    // Remove all the lingering executors that should be removed but not yet. The reason might be
    // because (1) disconnected event is not yet received; (2) executors die silently.
    executors.foreach { eid =>
      removeExecutor(eid, SlaveLost("Stale executor after cluster manager re-registered."))
    }
  }
```

mridulm:
@zsxwing I think the issue is that case RemoveExecutor() is not identical to what exists in receiveAndReply
Any reason 'executorDataMap.get(executorId).foreach(_.executorEndpoint.send(StopExecutor))' is missing from case RemoveExecutor(executorId, reason) in receive ?

If we add that in there, will it not be sufficient ?
I dont want to call removeExecutor from outside the event loop.

zsxwing:
I meant `CoarseGrainedSchedulerBackend.removeExecutor` not `DriverEndpoint.removeExecutor`. It's confusing that we have two methods having the same name :(

mridulm:
Ah ! Apologies, I got confused. Yes, I agree, that is a better approach.

It also means we can get rid of the RemoveExecutor pattern match from receive right ? As it stands now, that looks buggy.

zsxwing:
> It also means we can get rid of the RemoveExecutor pattern match from receive right ?

yep

scwf:
`CoarseGrainedSchedulerBackend.removeExecutor` also use ask,  but it does not matter right?  because it just send msg once and log the error if failure

zsxwing:
retest this please

viirya:
LGTM now.

zsxwing:
LGTM. Merging to master and 2.0. Thanks!

PR #15482, opened 2016-10-14
[Hot fix] ignore("single listener, check trigger statuses") for now

lw-lin:


srowen:
Is this actually failing builds? I don't see that consistently at https://amplab.cs.berkeley.edu/jenkins/view/Spark%20QA%20Test%20(Dashboard)/

HyukjinKwon:
I think he met this in https://amplab.cs.berkeley.edu/jenkins/job/SparkPullRequestBuilder/66954/testReport/ which seems a possibly flaky test. It seems several latest tests gone fine though. Maybe it'd be great if we identify the cause and a proper fix.

lw-lin:
Now that td has issued a commit https://github.com/apache/spark/commit/05800b4b4e7873ebc445dfcd020b76d7539686e1 exactly like this, I'm closing this. Sure `flaky-test` is more proper than `hot-fix` -- thanks @srowen @HyukjinKwon !

PR #15483, opened 2016-10-14
[SPARK-17935][SQL]Add KafkaForeachWriter in external kafka-0.8.0 for structured streaming module

zhangxinyu1:

I propose a `KafkaForeachWriter.scala` in external kafka-0.8.0. 
It's useful for someone who requires to write results to kafka in structured streaming module. It can be used with `ForeachSink`, and the usage is as below: 

 `val query = input.writeStream`
                   `.outputMode("append")`
                   `.foreach(new KafkaForeachWriter[Row](producerProps, topics, input.schema)`
                   `.start()` 
## How was this patch tested?

Unit test in `KafkaForeachWriterSuite.scala`

### DIFF
--- a/external/kafka-0-8/pom.xml
+++ b/external/kafka-0-8/pom.xml
@@ -46,6 +46,13 @@
       <version>${project.version}</version>
       <scope>provided</scope>
     </dependency>
+	<dependency>
+      <groupId>org.apache.spark</groupId>
+      <artifactId>spark-sql_${scala.binary.version}</artifactId>
+      <version>${project.version}</version>
+      <type>test-jar</type>
+      <scope>test</scope>
+    </dependency>
     <dependency>
       <groupId>org.apache.spark</groupId>
       <artifactId>spark-core_${scala.binary.version}</artifactId>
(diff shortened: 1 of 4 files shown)


zhangxinyu1:
@marmbrus Could you take a look? Please tell me if you hava any suggestions?
Thanks!

marmbrus:
Thanks for working on this!  However, I'm not sure that this is something that we should merge into the core repository (Though I think its an awesome example of how to use the `ForeachWriter` interface!  You should consider making a blog post or even a spark package).  Specifically, when we add support for writing to kafka, I think we'll want to do it both through the DataSource API, as well as the `Sink` API.  This would let users seamlessly write batch and streaming jobs to Kafka in java, scala, python, R and SQL using the same code.

zhangxinyu1:
@marmbrus 
Yes, I also think `KafkaSink` is better than `KafkaForeachWriter` for users. I 'm trying to complete `KafkaSink` these days. Like `KafkaSource` in external kafka-0-10-sql, there is a procuer cache in executor to store kafkaProducers, which avoids repeatly creating and closing producers.

zhangxinyu1:
@marmbrus 
I write a short deasign doc of KafkaSink in jira https://issues.apache.org/jira/browse/SPARK-17935 
Could you please take a look?

marmbrus:
I will try to take a look soon.  Can we close this PR until we have an updated implementation?

zhangxinyu1:
Sure. Thanks!

PR #15484, opened 2016-10-14
[SPARK-17868][SQL] Do not use bitmasks during parsing and analysis of CUBE/ROLLUP/GROUPING SETS

jiangxb1987:

We generate bitmasks for grouping sets during the parsing process, and use these during analysis. These bitmasks are difficult to work with in practice and have lead to numerous bugs. This PR removes these and use actual sets instead, however we still need to generate these offsets for the grouping_id.

This PR does the following works:
1. Replace bitmasks by actual grouping sets durning Parsing/Analysis stage of CUBE/ROLLUP/GROUPING SETS;
2. Add new testsuite `ResolveGroupingAnalyticsSuite` to test the `Analyzer.ResolveGroupingAnalytics` rule directly;
3. Fix a minor bug in `ResolveGroupingAnalytics`.
## How was this patch tested?

By existing test cases, and add new testsuite `ResolveGroupingAnalyticsSuite` to test directly.

### DIFF
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
@@ -306,18 +306,18 @@ class Analyzer(
       // GROUPING SETS ((a,b), a), we do not need to change the nullability of a, but we
       // should change the nullabilty of b to be TRUE.
       // TODO: For Cube/Rollup just set nullability to be `true`.
-      val expandedAttributes = groupByAliases.zipWithIndex.map { case (a, idx) =>
-        if (selectedGroupByExprs.exists(!_.contains(a.child))) {
-          a.toAttribute.withNullability(true)
+      val expandedAttributes = groupByAliases.map { alias =>
+        if (selectedGroupByExprs.exists(!_.contains(alias.child))) {
+          alias.toAttribute.withNullability(true)
         } else {
-          a.toAttribute
+          alias.toAttribute
         }
       }
 
       val groupingSetsAttributes = selectedGroupByExprs.map { groupingSetExprs =>
         groupingSetExprs.map { expr =>
           val alias = groupByAliases.find(_.child.semanticEquals(expr)).getOrElse(
-            failAnalysis(s"$expr doesn't show up in the GROUP BY list"))
+            failAnalysis(s"$expr doesn't show up in the GROUP BY list $groupByAliases"))
           // Map alias to expanded attribute.
           expandedAttributes.find(_.semanticEquals(alias.toAttribute)).getOrElse(
             alias.toAttribute)
@@ -335,7 +335,7 @@ class Analyzer(
         aggregations: Seq[NamedExpression],
         groupByAliases: Seq[Alias],
         groupingAttrs: Seq[Expression],
-        gid: Attribute): Seq[NamedExpression] = aggregations.map { case expr =>
+        gid: Attribute): Seq[NamedExpression] = aggregations.map {
       // collect all the found AggregateExpression, so we can check an expression is part of
       // any AggregateExpression or not.
       val aggsBuffer = ArrayBuffer[Expression]()
@@ -343,7 +343,7 @@ class Analyzer(
       def isPartOfAggregation(e: Expression): Boolean = {
         aggsBuffer.exists(a => a.find(_ eq e).isDefined)
       }
-      replaceGroupingFunc(expr, groupByExprs, gid).transformDown {
+      replaceGroupingFunc(_, groupByExprs, gid).transformDown {
         // AggregateExpression should be computed on the unmodified value of its argument
         // expressions, so we should not replace any references to grouping expression
         // inside it.
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
@@ -219,14 +219,14 @@ class Analyzer(
      *  We need to get all of its subsets for the rule described above, the subset is
      *  represented as sequence of expressions.
      */
-    def selectGroupExprsRollup(exprs: Seq[Expression]): Seq[Seq[Expression]] = {
-      if (exprs.length == 0) {
-        Seq(Seq.empty[Expression])
-      } else {
-        selectGroupExprsRollup(exprs.drop(1)).map { expandExprs =>
-          exprs.take(1) ++ expandExprs
-        } ++ Seq(Seq.empty[Expression])
+    def rollupExprs(exprs: Seq[Expression]): Seq[Seq[Expression]] = {
+      val buffer = ArrayBuffer.empty[Seq[Expression]]
+      var current = exprs
+      while (current.nonEmpty) {
+        buffer += current
+        current = current.init
       }
+      buffer
     }
 
     /*
@@ -238,15 +238,12 @@ class Analyzer(
      *  We need to get all of its subsets for a given GROUPBY expression, the subsets are
      *  represented as sequence of expressions.
      */
-    def selectGroupExprsCube(exprs: Seq[Expression]): Seq[Seq[Expression]] = {
-      if (exprs.length == 0) {
-        Seq(Seq.empty[Expression])
-      } else {
-        val expandExprsList = selectGroupExprsCube(exprs.drop(1))
-        expandExprsList.map { expandExprs =>
-          exprs.take(1) ++ expandExprs
-        } ++ expandExprsList
-      }
+    def cubeExprs(exprs: Seq[Expression]): Seq[Seq[Expression]] = exprs match {
+      case x :: xs =>
+        val initial = cubeExprs(xs)
+        initial.map(x +: _) ++ initial
+      case Nil =>
+        Seq(Seq.empty)
     }
 
     private def hasGroupingAttribute(expr: Expression): Boolean = {
@@ -296,10 +293,10 @@ class Analyzer(
 
       case Aggregate(Seq(c @ Cube(groupByExprs)), aggregateExpressions, child) =>
         GroupingSets(
-          selectGroupExprsCube(c.groupByExprs), groupByExprs, child, aggregateExpressions)
+          cubeExprs(c.groupByExprs), groupByExprs, child, aggregateExpressions)
       case Aggregate(Seq(r @ Rollup(groupByExprs)), aggregateExpressions, child) =>
         GroupingSets(
-          selectGroupExprsRollup(r.groupByExprs), groupByExprs, child, aggregateExpressions)
+          rollupExprs(r.groupByExprs), groupByExprs, child, aggregateExpressions)
 
       // Ensure all the expressions have been resolved.
       case x: GroupingSets if x.expressions.forall(_.resolved) =>
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
@@ -226,7 +226,7 @@ class Analyzer(
         buffer += current
         current = current.init
       }
-      buffer
+      buffer += Seq.empty
     }
 
     /*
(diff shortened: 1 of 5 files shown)


sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/parser/AstBuilder.scala
```diff
@@ -488,23 +488,13 @@ class AstBuilder extends SqlBaseBaseVisitor[AnyRef] with Logging {
...
-      val expressionMap = groupByExpressions.zipWithIndex.toMap
-      val numExpressions = expressionMap.size
-      val mask = (1 << numExpressions) - 1
-      val masks = ctx.groupingSet.asScala.map {
-        _.expression.asScala.foldLeft(mask) {
-          case (bitmap, eCtx) =>
-            // Find the index of the expression.
-            val e = typedVisit[Expression](eCtx)
-            val index = expressionMap.find(_._1.semanticEquals(e)).map(_._2).getOrElse(
-              throw new ParseException(
```
jiangxb1987: We don't check whether expression is in the GROUP BY list, moved this to Analysis stage.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
```diff
@@ -335,13 +366,16 @@ class Analyzer(
 
         Aggregate(groupingAttrs, aggregations, expand)
 
-      case f @ Filter(cond, child) if hasGroupingFunction(cond) =>
+      // We should make sure all expressions in condition have been resolved.
```
jiangxb1987: expression in condition could be unresolved, we should check this and respect the operator until `Filter.condition` is resolved. A testcase in `ResolveGroupingAnalyticsSuite` would fail before this PR.

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/analysis/ResolveGroupingAnalyticsSuite.scala
```diff
@@ -0,0 +1,244 @@
...
+        Seq(unresolved_a, unresolved_b), r1, Seq(unresolved_a, unresolved_b)))
+    val expected2 = Project(Seq(a, b), Sort(
+      Seq(SortOrder('aggOrder.int.withNullability(false), Ascending)), true,
+      Aggregate(Seq(a, b, gid),
+        Seq(a, b, gid.as("aggOrder")),
+        Expand(
+          Seq(Seq(a, b, c, nulInt, nulStr, 3), Seq(a, b, c, a, nulStr, 1), Seq(a, b, c, a, b, 0)),
+          Seq(a, b, c, a, b, gid),
+          Project(Seq(a, b, c, a.as("a"), b.as("b")), r1)))))
+    checkAnalysis(originalPlan2, expected2)
```
jiangxb1987: Will add negative cases for this.

jiangxb1987:
@hvanhovell Could you look at this please? Thank you!

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/parser/AstBuilder.scala
hvanhovell: That is fine.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/basicLogicalOperators.scala
```diff
@@ -518,59 +517,64 @@ case class Window(
...
-    while (bit >= 0) {
-      if (((bitmask >> bit) & 1) == 1) nonSelect += attrs(attrs.length - bit - 1)
-      bit -= 1
+  private def buildBitmask(
+    groupingSetAttrs: Seq[Attribute],
+    attrMap: Map[Attribute, Int]): Int = {
+    val numAttributes = attrMap.size
+    val mask = (1 << numAttributes) - 1
+
+    groupingSetAttrs.foldLeft(mask) {
```
hvanhovell: You only need fold left when you want to traverse the collection in a certain order. Folds are typically not the easiest to understand, so use them sparingly and prefer to use more imperative constructs.

Could you rewrite this using a more imperative approach?

Just for kicks & giggles:
```scala
groupingSetAttrs.map(attrMap).map(index => ~(1 << (numAttributes - 1 - index))).reduce(_ && _)
```

hvanhovell (changes requested):

jiangxb1987:
@hvanhovell Thank you for your comments - They are awesome! I've made some changes according to your advice, I hope the code looks better now. Thanks a lot!

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/basicLogicalOperators.scala
```diff
@@ -518,59 +517,58 @@ case class Window(
...
-      bit -= 1
-    }
-
-    AttributeSet(nonSelect)
+  private def buildBitmask(
+    groupingSetAttrs: Seq[Attribute],
+    attrMap: Map[Attribute, Int]): Int = {
+    val numAttributes = attrMap.size
+    val mask = (1 << numAttributes) - 1
+    (mask +: groupingSetAttrs.map(attrMap).map(index =>
```
hvanhovell: Could you add a little bit of documentation on the mask? It is non-trivial to understand.

It might also be a good idea to split this into two separate statements. One to calculate the the attribute masks and one to reduce them.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
```diff
@@ -255,98 +265,125 @@ class Analyzer(
...
+    private def constructExpand(
+        selectedGroupByExprs: Seq[Seq[Expression]],
+        child: LogicalPlan,
+        groupByAliases: Seq[Alias],
+        gid: Attribute): LogicalPlan = {
+      // Change the nullability of group by aliases if necessary. For example, if we have
+      // GROUPING SETS ((a,b), a), we do not need to change the nullability of a, but we
+      // should change the nullabilty of b to be TRUE.
+      // TODO: For Cube/Rollup just set nullability to be `true`.
+      val expandedAttributes = groupByAliases.zipWithIndex.map { case (a, idx) =>
```
rxin: idx is not used?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
```diff
@@ -255,98 +265,125 @@ class Analyzer(
...
-            aggsBuffer.exists(a => a.find(_ eq e).isDefined)
+    /*
+     * Construct new aggregate expressions by replacing grouping functions.
+     */
+    private def constructAggregateExprs(
+        groupByExprs: Seq[Expression],
+        aggregations: Seq[NamedExpression],
+        groupByAliases: Seq[Alias],
+        groupingAttrs: Seq[Expression],
+        gid: Attribute): Seq[NamedExpression] = aggregations.map { case expr =>
```
rxin: don't need the "case" here.

rxin:
cc @davies too

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
```diff
@@ -216,10 +216,16 @@ class Analyzer(
      *  Group Count: N + 1 (N is the number of group expressions)
      *
      *  We need to get all of its subsets for the rule described above, the subset is
-     *  represented as the bit masks.
+     *  represented as sequence of expressions.
      */
-    def bitmasks(r: Rollup): Seq[Int] = {
-      Seq.tabulate(r.groupByExprs.length + 1)(idx => (1 << idx) - 1)
+    def rollupExprs(exprs: Seq[Expression]): Seq[Seq[Expression]] = {
+      val buffer = ArrayBuffer.empty[Seq[Expression]]
```
tejasapatil: Avoid using `ArrayBuffer` as insertions would lead to expansion of underlying array and copying of data to the new one. Since you know the size upfront, you could create an `Array` of required size.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
tejasapatil: +1. Looking at it more, I feel `zipWithIndex` is not needed at all and the `map` would suffice.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
```diff
@@ -255,98 +265,125 @@ class Analyzer(
...
+      }
 
-        val nonNullBitmask = x.bitmasks.reduce(_ & _)
-
-        val expandedAttributes = groupByAliases.zipWithIndex.map { case (a, idx) =>
-          a.toAttribute.withNullability((nonNullBitmask & 1 << idx) == 0)
+      val groupingSetsAttributes = selectedGroupByExprs.map { groupingSetExprs =>
+        groupingSetExprs.map { expr =>
+          val alias = groupByAliases.find(_.child.semanticEquals(expr)).getOrElse(
+            failAnalysis(s"$expr doesn't show up in the GROUP BY list"))
```
tejasapatil: can you also display the GROUP BY list in the message ?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
jiangxb1987: The use of `ArrayBuffer` will make this piece of code more concise, since the sequence of `exprs` is not usually very long, maybe performance is not the major concern here, I'd prefer to keep this one, is it OK? @hvanhovell

jiangxb1987:
@tejasapatil @rxin I've addressed most of your comments, thanks for reviewing this!

jiangxb1987:
@davies Would you please have a look at this PR? Thank you!

davies:
@jiangxb1987 Could you say a little bit more about the "minor bug"? that help us to decide this patch should be backported or not.

hvanhovell:
It is not really a bug. Bitmap manipulation has bitten us quite a few time in the past, so I would rather use expressions.

davies:
I see, this PR just improve the readability, we don't need to backport it, looks good to me overall.

jiangxb1987:
What else should I update on this PR? Please don't be hesitate to require any change, thanks!

jiangxb1987:
ping @hvanhovell

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
rxin: Is this just `exprs.inits` ?

to be honest this is the first time I've seen the use of init/inits on a trait.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
```diff
@@ -229,10 +235,14 @@ class Analyzer(
      *  Group Count: 2 ^ N (N is the number of group expressions)
      *
      *  We need to get all of its subsets for a given GROUPBY expression, the subsets are
-     *  represented as the bit masks.
+     *  represented as sequence of expressions.
      */
-    def bitmasks(c: Cube): Seq[Int] = {
-      Seq.tabulate(1 << c.groupByExprs.length)(i => i)
+    def cubeExprs(exprs: Seq[Expression]): Seq[Seq[Expression]] = exprs.toList match {
```
rxin: I'd also write unit tests specifically for cubeExprs and rollupExprs

rxin: Also I think you can just use subsets? e.g.

```
scala> Seq(1, 2, 3).toSet.subsets.foreach(println)
Set()
Set(1)
Set(2)
Set(3)
Set(1, 2)
Set(1, 3)
Set(2, 3)
Set(1, 2, 3)
```

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
jiangxb1987: `exprs.inits` is much more concise.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
jiangxb1987: I'm afraid we can't just map the `exprs` to a set because we want to keep the original order.

jiangxb1987:
Does this version looks good now?

jiangxb1987:
ping @hvanhovell

hvanhovell:

PR #15485, opened 2016-10-14
[Doc] Fix typo in sql hive doc

dhruve:
Change is too trivial to file a JIRA.

### DIFF
--- a/docs/sql-programming-guide.md
+++ b/docs/sql-programming-guide.md
@@ -998,7 +998,7 @@ The following options can be used to configure the version of Hive that is used 
         <li>A classpath in the standard format for the JVM. This classpath must include all of Hive
         and its dependencies, including the correct version of Hadoop. These jars only need to be
         present on the driver, but if you are running in yarn cluster mode then you must ensure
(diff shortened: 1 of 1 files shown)


HyukjinKwon:
Maybe I think you could try to find other typoes together (at least in other docs as you fix a typo in documentation). It seems too minor in my opinion.

dhruve:
I wasn't scanning explicitly for these typoes. I have seen individual commits addressing such typoes earlier so submitted the super trivial change. If unnecessary we can close this.

HyukjinKwon:
Yeap, it is true and I think it is acceptable but maybe we should avoid. I guess we don't want flooding small fixes for typoes.

srowen:
Yeah, I'd prefer to encourage fixing a bunch of stuff at once, and discourage truly non-functional changes that don't meaningfully clarify something. If you add up the time everyone spends looking at it it's, shockingly, like hundreds of dollars of time. OK, I can merge it as the fastest way to resolve this but it's not worth it in general.

PR #15487, opened 2016-10-14
[SPARK-17940][SQL] Fixed a typo in LAST function and improved its usage string

lins05:
- Fixed a a typo in the LAST function error message
- Also improved its usage string to match the FIRST function
## How was this patch tested?

Existing tests.

### DIFF
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/Last.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/Last.scala
@@ -29,10 +29,14 @@ import org.apache.spark.sql.types._
  * a single partition, and we use a single reducer to do the aggregation.).
  */
 @ExpressionDescription(
-  usage = """_FUNC_(expr) - Returns the last value of `child` for a group of rows.
-    _FUNC_(expr,isIgnoreNull=false) - Returns the last value of `child` for a group of rows.
-      If isIgnoreNull is true, returns only non-null values. Note that in most cases the
-      result is nondeterministic.
+  usage =
+    """
+      _FUNC_(expr) - Returns the last value of `child` for a group of rows.
+
+      _FUNC_(expr, isIgnoreNull) - Returns the last value of `child` for a group of rows.
+      If isIgnoreNull is true, returns only non-null values.
+
+      Note that in most cases the result is nondeterministic.
     """)
 case class Last(child: Expression, ignoreNullsExpr: Expression) extends DeclarativeAggregate {
 
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/Last.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/Last.scala
@@ -29,9 +29,10 @@ import org.apache.spark.sql.types._
  * a single partition, and we use a single reducer to do the aggregation.).
  */
 @ExpressionDescription(
-  usage = """_FUNC_(expr,isIgnoreNull) - Returns the last value of `child` for a group of rows.
+  usage = """_FUNC_(expr) - Returns the last value of `child` for a group of rows.
     _FUNC_(expr,isIgnoreNull=false) - Returns the last value of `child` for a group of rows.
-      If isIgnoreNull is true, returns only non-null values.
+      If isIgnoreNull is true, returns only non-null values. Note that in most cases the
+      result is nondeterministic.
     """)
 case class Last(child: Expression, ignoreNullsExpr: Expression) extends DeclarativeAggregate {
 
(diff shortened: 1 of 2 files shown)


sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/Last.scala
```diff
@@ -29,15 +29,18 @@ import org.apache.spark.sql.types._
  * a single partition, and we use a single reducer to do the aggregation.).
  */
 @ExpressionDescription(
-  usage = "_FUNC_(expr,isIgnoreNull) - Returns the last value of `child` for a group of rows.")
+  usage = """_FUNC_(expr,isIgnoreNull) - Returns the last value of `child` for a group of rows.
+    _FUNC_(expr,isIgnoreNull=false) - Returns the last value of `child` for a group of rows.
```
hvanhovell: Why repeat this?

hvanhovell: While you are added could you also add a line about this being a non-deterministic function

HyukjinKwon: I just happened to take a look at this PR. I manually built and tested. It seems printing the message as below:

```sql
spark-sql> DESCRIBE FUNCTION last;
Function: last
Class: org.apache.spark.sql.catalyst.expressions.aggregate.Last
Usage: last(expr,isIgnoreNull) - Returns the last value of `child` for a group of rows.
    last(expr,isIgnoreNull=false) - Returns the last value of `child` for a group of rows.
      If isIgnoreNull is true, returns only non-null values.
```

Maybe, it'd be nicer if the indentation is pretty.

lins05: @hvanhovell @HyukjinKwon Thanks for the review. I'm simply following the usage string of other functions, e.g:

```
spark-sql> describe function first;
Function: first
Class: org.apache.spark.sql.catalyst.expressions.aggregate.First
Usage: first(expr) - Returns the first value of `child` for a group of rows.
    first(expr,isIgnoreNull=false) - Returns the first value of `child` for a group of rows.
      If isIgnoreNull is true, returns only non-null values.
   
spark-sql> describe function approx_count_distinct;
Function: approx_count_distinct
Class: org.apache.spark.sql.catalyst.expressions.aggregate.HyperLogLogPlusPlus
Usage: approx_count_distinct(expr) - Returns the estimated cardinality by HyperLogLog++.
    approx_count_distinct(expr, relativeSD=0.05) - Returns the estimated cardinality by HyperLogLog++
      with relativeSD, the maximum estimation error allowed.
```

So it seems the current convention is that: the first line is a short one-line description, followed by a detail description. Do we have any explicit "usage string style" to follow?

@hvanhovell I'll add the note about nondeterministic.

HyukjinKwon: Ah, I thought it should be as below if it dose not affect anything but only this:

```sql
spark-sql> DESCRIBE FUNCTION last;
Function: last
Class: org.apache.spark.sql.catalyst.expressions.aggregate.Last
Usage: last(expr,isIgnoreNull) - Returns the last value of `child` for a group of rows.
       last(expr,isIgnoreNull=false) - Returns the last value of `child` for a group of rows.
         If isIgnoreNull is true, returns only non-null values.
```

This was just my personal opinion.

HyukjinKwon:
@hvanhovell @lins05 I just have a confirmation for the documentation format and will proceed to sweep it ([SPARK-17963](https://issues.apache.org/jira/browse/SPARK-17963)). So, maybe you could do as below:

```scala
@ExpressionDescription(
  usage =
    """
      _FUNC_() - Returns a random column with i.i.d. uniformly distributed values in [0, 1].
        seed is given randomly.
      _FUNC_(seed) - Returns a random column with i.i.d. uniformly distributed values in [0, 1].
        seed should be an integer/long/NULL literal.
    """,
  extended = "...")
```

In the PR I am going to submit, I will exclude both functions, `first` and `last` and `ran` which would be handled in https://github.com/apache/spark/pull/15432

lins05:
@HyukjinKwon thanks, I'll update the PR accordingly.

lins05:
@HyukjinKwon I've updated the usage string. Now it looks like this:
```
spark-sql> describe function first;
Function: first
Class: org.apache.spark.sql.catalyst.expressions.aggregate.First
Usage: 
      first(expr) - Returns the first value of `child` for a group of rows.

      first(expr, isIgnoreNull=false) - Returns the first value of `child` for a group of rows.
      If isIgnoreNull is true, returns only non-null values.

      Note that in most cases the result is nondeterministic.
    
spark-sql> describe function last;
Function: last
Class: org.apache.spark.sql.catalyst.expressions.aggregate.Last
Usage: 
      last(expr) - Returns the last value of `child` for a group of rows.

      last(expr, isIgnoreNull) - Returns the last value of `child` for a group of rows.
      If isIgnoreNull is true, returns only non-null values.

      Note that in most cases the result is nondeterministic.
```

HyukjinKwon:
@lins05 This is being discussed in https://github.com/apache/spark/pull/15513. One idea is, we might have to a bit wait until it is confirmed (although I guess it is a soft-yes). Alternatively, I can do the indentation stuff in the PR above together but let this PR fix only the typo and extra mention for `nundeterministic`. I can resolve the conflicts.

Let us please follow @hvanhovell's decision here.

lins05:
@HyukjinKwon OK, please fix all these in your PR. I'll close this small one.

PR #15488, opened 2016-10-14
[SPARK-17941][ML][TEST] Logistic regression tests should use sample weights.

sethah:

The sample weight testing for logistic regressions is not robust. Logistic regression suite already has many test cases comparing results to R glmnet. Since both libraries support sample weights, we should use sample weights in the test to increase coverage for sample weighting. This patch doesn't really add any code and makes the testing more complete.

Also fixed some errors with the R code that was referenced in the test suit. Changed `standardization=T` to `standardize=T` since the former is invalid.
## How was this patch tested?

Existing unit tests are modified. No non-test code is touched.

### DIFF
--- a/mllib/src/test/scala/org/apache/spark/ml/classification/LogisticRegressionSuite.scala
+++ b/mllib/src/test/scala/org/apache/spark/ml/classification/LogisticRegressionSuite.scala
@@ -104,7 +104,7 @@ class LogisticRegressionSuite
    * Enable the ignored test to export the dataset into CSV format,
    * so we can validate the training accuracy compared with R's glmnet package.
    */
-  test("export test data into CSV format") {
+  ignore("export test data into CSV format") {
     binaryDataset.rdd.map { case Row(label: Double, features: Vector, weight: Double) =>
       label + "," + weight + "," + features.toArray.mkString(",")
     }.repartition(1).saveAsTextFile("target/tmp/LogisticRegressionSuite/binaryDataset")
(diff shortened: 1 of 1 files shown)


sethah:
cc @yanboliang @dbtsai

mllib/src/test/scala/org/apache/spark/ml/classification/LogisticRegressionSuite.scala
```diff
@@ -519,31 +520,35 @@ class LogisticRegressionSuite
...
-       Using the following R code to load the data and train the model using glmnet package.
-
-       library("glmnet")
-       data <- read.csv("path", header=FALSE)
-       label = factor(data$V1)
-       features = as.matrix(data.frame(data$V2, data$V3, data$V4, data$V5))
-       coefficients = coef(glmnet(features,label, family="binomial", alpha = 0, lambda = 0))
-       coefficients
+      Use the following R code to load the data and train the model using glmnet package.
+      library("glmnet")
```
sethah: I pasted every R code snippet into an R shell, so we can be reasonably certain of its correctness

mllib/src/test/scala/org/apache/spark/ml/classification/LogisticRegressionSuite.scala
```diff
@@ -1190,92 +1169,95 @@ class LogisticRegressionSuite
...
     val model2 = trainer2.fit(multinomialDataset)
 
     /*
-       Use the following R code to load the data and train the model using glmnet package.
-       library("glmnet")
-       data <- read.csv("path", header=FALSE)
-       label = as.factor(data$V1)
-       features = as.matrix(data.frame(data$V2, data$V3, data$V4, data$V5))
-       coefficientsStd = coef(glmnet(features, label, family="multinomial", alpha = 1,
-        lambda = 0.05, standardization=T))
```
sethah: `standardization` is an invalid argument. I changed these to be correct `standardize`

dbtsai:
LGTM.

dbtsai:
Merged into master. Thanks.

PR #15489, opened 2016-10-14
[SPARK-17863]  [SQL] should not add column into Distinct

davies:

We are trying to resolve the attribute in sort by pulling up some column for grandchild into child, but that's wrong when the child is Distinct, because the added column will change the behavior of Distinct, we should not do that. 
## How was this patch tested?

Added regression test.

### DIFF
--- a/sql/core/src/test/scala/org/apache/spark/sql/SQLQuerySuite.scala
+++ b/sql/core/src/test/scala/org/apache/spark/sql/SQLQuerySuite.scala
@@ -1106,6 +1106,30 @@ class SQLQuerySuite extends QueryTest with SharedSQLContext {
     )
   }
 
+  test("SPARK-17863: SELECT distinct does not work correctly if order by missing attribute") {
+    checkAnswer(
+      sql("""select distinct struct.a, struct.b
+          |from (
+          |  select named_struct('a', 1, 'b', 2, 'c', 3) as struct
+          |  union all
+          |  select named_struct('a', 1, 'b', 2, 'c', 4) as struct) tmp
+          |order by a, b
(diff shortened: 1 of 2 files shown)


sql/core/src/test/scala/org/apache/spark/sql/SQLQuerySuite.scala
```diff
@@ -1106,6 +1106,30 @@ class SQLQuerySuite extends QueryTest with SharedSQLContext {
...
+    val error = intercept[AnalysisException] {
+      sql("""select distinct struct.a, struct.b
+            |from (
+            |  select named_struct('a', 1, 'b', 2, 'c', 3) as struct
+            |  union all
+            |  select named_struct('a', 1, 'b', 2, 'c', 4) as struct) tmp
+            |order by struct.a, struct.b
+            |""".stripMargin)
+    }
+    assert(error.message contains "cannot resolve '`struct.a`' given input columns: [a, b]")
```
yhuai: Is it possible to suggest the workaround in the error message?

davies: What's the suggestion looks like?

yhuai: Oh, i see. Seems this error is thrown by check analysis, which does not have info about what is the query. So, it is hard to figure out what messages (for the workaround) to put at here.

yhuai:
LGTM.

yhuai:
Thanks! Merging to master and branch 2.0.

PR #15490, opened 2016-10-14
[SPARK-10541] [Web UI] Allow ApplicationHistoryProviders to provide their own text when there aren't any complete apps

ajbozarth:

I've added a method to `ApplicationHistoryProvider` that returns the html paragraph to display when there are no applications. This allows providers other than `FsHistoryProvider` to determine what is printed. The current hard coded text is now moved into `FsHistoryProvider` since it assumed that's what was being used before.

I chose to make the function return html rather than text because the current text block had inline html in it and it allows a new implementation of `ApplicationHistoryProvider` more versatility. I did not see any security issues with this since injecting html here requires implementing `ApplicationHistoryProvider` and can't be done outside of code.
## How was this patch tested?

Manual testing and dev/run-tests

No visible changes to the UI

### DIFF
--- a/core/src/main/scala/org/apache/spark/deploy/history/FsHistoryProvider.scala
+++ b/core/src/main/scala/org/apache/spark/deploy/history/FsHistoryProvider.scala
@@ -267,7 +267,7 @@ private[history] class FsHistoryProvider(conf: SparkConf, clock: Clock)
     <p>
       Did you specify the correct logging directory? Please verify your setting of
       <span style="font-style:italic">spark.history.fs.logDirectory</span>
-      and whether you have the permissions to access it.
+      listed above and whether you have the permissions to access it.
       <br/>
       It is also possible that your application did not run to
       completion or did not stop the SparkContext.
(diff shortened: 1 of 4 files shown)


ajbozarth:
Jenkins, retest this please

ajbozarth:
@steveloughran Since you opened the JIRA do you mind taking a look?

steveloughran:
@ajbozarth I'm not a spark committer, I'm not capaclbe of getting stuff in. I did dd one comment to some of the code, otherwise nothing I have issues with. LGTM

ajbozarth:
Thanks @steveloughran, I'm not seeing your code comment though. Also @srowen mind taking a look?

steveloughran:
Oh, I see, new UI had meant I'd left the comment partially incomplete. Sorry. Just the one: printing out the actual log dir location. That makes it much easier to identify a configuration problem and allow anyone with access to the cluster FS just to take a quick ls of the directory to see what's up

core/src/main/scala/org/apache/spark/deploy/history/FsHistoryProvider.scala
```diff
@@ -262,6 +263,17 @@ private[history] class FsHistoryProvider(conf: SparkConf, clock: Clock)
     }
   }
 
+  override def getEmptyListingHtml(): Seq[Node] = {
+    <p>
+      Did you specify the correct logging directory? Please verify your setting of
+      <span style="font-style:italic">spark.history.fs.logDirectory</span>
```
ajbozarth: That's a great idea, I'll update it

ajbozarth:
The dir is actually already displayed so I just added a few words to clarify that the current setting is displayed above.

screen shot 2016-10-18 at 3 48 39 pm

vanzin:
LGTM. Merging to master.

PR #15493, opened 2016-10-14
[SPARK-17946][PYSPARK] Python crossJoin API similar to Scala

srinathshankar:

Add a crossJoin function to the DataFrame API similar to that in Scala. Joins with no condition (cartesian products) must be specified with the crossJoin API
## How was this patch tested?

Added python tests to ensure that an AnalysisException if a cartesian product is specified without crossJoin(), and that cartesian products can execute if specified via crossJoin()

(If this patch involves UI changes, please attach a screenshot; otherwise, remove this)

### DIFF
--- a/python/pyspark/sql/dataframe.py
+++ b/python/pyspark/sql/dataframe.py
@@ -627,7 +627,7 @@ class DataFrame(object):
         return DataFrame(getattr(self._jdf, "as")(alias), self.sql_ctx)
 
     @ignore_unicode_prefix
-    @since(2.0)
+    @since(2.1)
     def crossJoin(self, other):
         """Returns the cartesian product with another :class:`DataFrame`
 
--- a/python/pyspark/sql/dataframe.py
+++ b/python/pyspark/sql/dataframe.py
@@ -627,6 +627,24 @@ class DataFrame(object):
         return DataFrame(getattr(self._jdf, "as")(alias), self.sql_ctx)
 
     @ignore_unicode_prefix
+    @since(2.0)
+    def crossJoin(self, other):
+        """Returns the cartesian product with another :class:`DataFrame`
+
+        :param other: Right side of the cartesian product
+
+        >>> df.select("age", "name").collect()
+        [Row(age=2, name=u'Alice'), Row(age=5, name=u'Bob')]
+        >>> df2.select("name", "height").collect()
+        [Row(name=u'Tom', height=80), Row(name=u'Bob', height=85)]
+        >>> df.crossJoin(df2.select("height")).select("age", "name", "height").collect()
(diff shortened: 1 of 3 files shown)


srinathshankar:
cc @sameeragarwal @hvanhovell @davies @JoshRosen

python/pyspark/sql/dataframe.py
```diff
@@ -627,6 +627,25 @@ def alias(self, alias):
         return DataFrame(getattr(self._jdf, "as")(alias), self.sql_ctx)
 
     @ignore_unicode_prefix
+    @since(2.0)
```
sameeragarwal: shouldn't this be 2.1?

sameeragarwal (approved):
LGTM

python/pyspark/sql/dataframe.py
```diff
@@ -627,6 +627,25 @@ def alias(self, alias):
         return DataFrame(getattr(self._jdf, "as")(alias), self.sql_ctx)
 
     @ignore_unicode_prefix
+    @since(2.1)
+    def crossJoin(self, other):
+        """Returns the cartesian product with another :class:`DataFrame`
```
rxin: nit: add a period.

srinathshankar: done.

rxin:
LGTM pending Jenkins.

rxin:
Merging in master.

rxin:
cc @felixcheung do we need some change for R?

felixcheung:
@rxin In R, CrossJoin is the default when expr is empty (https://github.com/apache/spark/blob/master/R/pkg/R/DataFrame.R#L2304)
I reviewed the code and documentation I think it is sufficient

rxin:
The issue is that we want to prevent users from shooting themselves in the foot, i.e. we want to avoid accidental cross joins. The idea is unless the user explicitly says crossJoin, we should disallow crossjoins.

felixcheung:
That's a great point. Currently R is the same as Python in that when `joinExpr` is `NULL` (R) or `on` is `None` (Python), CrossJoin is assumed. (Python [here](https://github.com/apache/spark/pull/15493/files#diff-6fc344560230bf0ef711bb9b5573f1faR650))

Problem is by default `joinExpr = NULL` (R) and `on = None` (Python) - one approach is to change both defaults, so that when omitted, they don't default to CrossJoin.
Another approach is to additional ask for `joinType` or `how` of a new `cross_join` - but this could be a bigger change.

rxin:
Why not just introduce a crossJoin function in R, similar to Python/Scala/Java?

We don't want to change the default join type, because it is still valid to run an inner join by specifying a predicate later using the filter operator.

rxin:
cc @felixcheung 

We still need this. I'm going to create an upstream ticket. Can one of you take it?

srinathshankar:
Created https://issues.apache.org/jira/browse/SPARK-18013
@felixcheung @rxin

felixcheung:
I will take that and add my note.

rxin:
Thanks!

PR #14531, opened 2016-08-08
[SPARK-17353] [SPARK-16943] [SPARK-16942] [SQL] Fix multiple bugs in CREATE TABLE LIKE command

gatorsmile:
### What changes were proposed in this pull request?

The existing `CREATE TABLE LIKE` command has multiple issues:
- The generated table is non-empty when the source table is a data source table. The major reason is the data source table is using the table property `path` to store the location of table contents. Currently, we keep it unchanged. Thus, we still create the same table with the same location.
- The table type of the generated table is `EXTERNAL` when the source table is an external Hive Serde table. Currently, we explicitly set it to `MANAGED`, but Hive is checking the table property `EXTERNAL` to decide whether the table is `EXTERNAL` or not. (See https://github.com/apache/hive/blob/master/metastore/src/java/org/apache/hadoop/hive/metastore/ObjectStore.java#L1407-L1408) Thus, the created table is still `EXTERNAL`. 
- When the source table is a `VIEW`, the metadata of the generated table contains the original view text and view original text. So far, this does not break anything, but it could cause something wrong in Hive. (For example, https://github.com/apache/hive/blob/master/metastore/src/java/org/apache/hadoop/hive/metastore/ObjectStore.java#L1405-L1406)
- The issue regarding the table `comment`. To follow what Hive does, the table comment should be cleaned, but the column comments should be still kept.
- The `INDEX` table is not supported. Thus, we should throw an exception in this case. 
- `owner` should not be retained. `ToHiveTable` set it [here](https://github.com/apache/spark/blob/e679bc3c1cd418ef0025d2ecbc547c9660cac433/sql/hive/src/main/scala/org/apache/spark/sql/hive/client/HiveClientImpl.scala#L793) no matter which value we set in `CatalogTable`. We set it to an empty string for avoiding the confusing output in Explain.
- Add a support for temp tables
- Like Hive, we should not copy the table properties from the source table to the created table, especially for the statistics-related properties, which could be wrong in the created table.
- `unsupportedFeatures` should not be copied from the source table. The created table does not have these unsupported features.
- When the type of source table is a view, the target table is using the default format of data source tables: `spark.sql.sources.default`.

This PR is to fix the above issues. 
### How was this patch tested?

Improve the test coverage by adding more test cases

PR #15494, opened 2016-10-14
[SPARK-17947] [SQL] Add Doc and Comment about spark.sql.debug

gatorsmile:

Just document the impact of `spark.sql.debug`:

When enabling the debug, Spark SQL internal table properties are not filtered out; however, some related DDL commands (e.g., Analyze Table and CREATE TABLE LIKE) might not work properly.

N/A

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala
@@ -917,7 +917,7 @@ object StaticSQLConf {
     .createWithDefault(4000)
 
   // When enabling the debug, Spark SQL internal table properties are not filtered out; however,
-  // Some related DDL commands (e.g., Analyze Table and CREATE TABLE LIKE) might not work properly.
+  // some related DDL commands (e.g., Analyze Table and CREATE TABLE LIKE) might not work properly.
   val DEBUG_MODE = buildConf("spark.sql.debug")
     .internal()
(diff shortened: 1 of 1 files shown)


rxin:
Why do they not work properly?

gatorsmile:
You can see the failed test cases in the PR: https://github.com/apache/spark/pull/15478

`ANALYZE TABLE` will fail due to the [checking](https://github.com/apache/spark/blob/6ce1b675ee9fc9a6034439c3ca00441f9f172f84/sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveExternalCatalog.scala#L111-L115
) when we try to alter the table properties. 

`SHOW CREATE TABLE` also outputs the table properties that should not be part of the output CREAT TABLE statement.

`CREATE TABLE LIKE` always excludes all the table properties of the source table. However, we might make a change in this part. This is still waiting for your input in another PR. See the [discussion](https://github.com/apache/spark/pull/14531#issuecomment-252147424)

cloud-fan:
@rxin because we are assuming that all the hive hacks are isolated inside `HiveExternalCatalog`, but the debug mode will break it and spread the hive metastore specific table properties outside of `HiveExternalCatalog`. Then some commands like `SHOW CREATE TABLE`, `ALTER TABLE` will be broken, but should be ok as it's a "debug" mode anyway.

cloud-fan:
thanks, merging to master!

PR #15492, opened 2016-10-14
[DO NOT MERGE][TEST] Testing flakiness of StreamingQueryListener

tdas:

PR #15497, opened 2016-10-15
[Test][SPARK-16002][Follow-up] Fix flaky test in StreamingQueryListenerSuite

lw-lin:

`StreamingQueryListenerSuite #test(s"single listener, check trigger statuses")` is flaky; the following is one possible flaky execution:

```
+-----------------------------------+--------------------------------+
|      StreamExecution thread       |         testing thread         |
+-----------------------------------+--------------------------------+
|  ManualClock.waitTillTime(100) {  |                                |
|        _isWaiting = true          |                                |
|            wait(10)               |                                |
|        still in wait(10)          |  if (_isWaiting) advance(100)  |
|        still in wait(10)          |  if (_isWaiting) advance(200)  | <- this should be disallowed !
|        still in wait(10)          |  if (_isWaiting) advance(300)  | <- this should be disallowed !
|      wake up from wait(10)        |                                |
|       current time is 600         |                                |
|       _isWaiting = false          |                                |
|  }                                |                                |
+-----------------------------------+--------------------------------+
```

This patch's fix is, disallowing `advance(...)` for **more than once** when there's some thread -- e.g. `StreamExecution` -- is `still in wait(10)`. In other words, we do not want to `advance()` "too early" -- in this sense, this is a follow-up to SPARK-16002.
## How was this patch tested?

Ran the flaky test for 1000 times, and all passed.

### DIFF
--- a/core/src/main/scala/org/apache/spark/util/ManualClock.scala
+++ b/core/src/main/scala/org/apache/spark/util/ManualClock.scala
@@ -27,7 +27,6 @@ package org.apache.spark.util
 private[spark] class ManualClock(private var time: Long) extends Clock {
 
   private var _isWaiting = false
-  private var _readyForFirstPeek = false
 
   /**
    * @return `ManualClock` with initial time 0
@@ -44,19 +43,14 @@ private[spark] class ManualClock(private var time: Long) extends Clock {
    */
   def setTime(timeToSet: Long): Unit = synchronized {
     time = timeToSet
-    _readyForFirstPeek = false
     notifyAll()
   }
 
   /**
-   * Note: if we don't want to advance too early, we should place a `isWaitingAndReadyForFirstPeek`
-   * guard before we call this. See SPARK-16002. See also `[sql] StreamTest#AdvanceManualClock`.
-   *
    * @param timeToAdd time (in milliseconds) to add to the clock's time
    */
   def advance(timeToAdd: Long): Unit = synchronized {
     time += timeToAdd
-    _readyForFirstPeek = false
     notifyAll()
   }
 
@@ -66,7 +60,6 @@ private[spark] class ManualClock(private var time: Long) extends Clock {
    */
   def waitTillTime(targetTime: Long): Long = synchronized {
     _isWaiting = true
-    _readyForFirstPeek = true
     try {
       while (time < targetTime) {
         wait(10)
@@ -74,13 +67,11 @@ private[spark] class ManualClock(private var time: Long) extends Clock {
       getTimeMillis()
     } finally {
       _isWaiting = false
-      _readyForFirstPeek = false
     }
   }
 
   /**
-   * Returns whether there is any thread being blocked in `waitTillTime`, and this will be the first
-   * time it's been peeked in the blocking state.
+   * Returns whether there is any thread being blocked in `waitTillTime`.
    */
-  def isWaitingAndReadyForFirstPeek: Boolean = synchronized { _isWaiting && _readyForFirstPeek }
+  def isWaiting: Boolean = synchronized { _isWaiting }
 }
--- a/core/src/main/scala/org/apache/spark/util/ManualClock.scala
+++ b/core/src/main/scala/org/apache/spark/util/ManualClock.scala
@@ -26,7 +26,7 @@ package org.apache.spark.util
  */
 private[spark] class ManualClock(private var time: Long) extends Clock {
 
-  private var _isWaiting = false
+  private var timeWhenWaitStarted: Option[Long] = None
 
   /**
    * @return `ManualClock` with initial time 0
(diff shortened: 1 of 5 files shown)


lw-lin:
@tdas @zsxwing could you take a look

lw-lin:
Jenkins retest this please

core/src/main/scala/org/apache/spark/util/ManualClock.scala
```diff
@@ -27,6 +27,7 @@ package org.apache.spark.util
 private[spark] class ManualClock(private var time: Long) extends Clock {
 
   private var _isWaiting = false
+  private var _readyForFirstPeek = false
```
tdas: Can you add more docs on what does peek mean, what does this variable supposed to do?

tdas: Its really not obvious on what purpose does _readyForFirstPeek server over _isWaiting.

core/src/main/scala/org/apache/spark/util/ManualClock.scala
```diff
@@ -60,18 +66,21 @@ private[spark] class ManualClock(private var time: Long) extends Clock {
...
     }
   }
 
   /**
-   * Returns whether there is any thread being blocked in `waitTillTime`.
+   * Returns whether there is any thread being blocked in `waitTillTime`, and this will be the first
+   * time it's been peeked in the blocking state.
    */
-  def isWaiting: Boolean = synchronized { _isWaiting }
+  def isWaitingAndReadyForFirstPeek: Boolean = synchronized { _isWaiting && _readyForFirstPeek }
```
tdas: I dont like this name. Why not keep it simple like isWaiting.

tdas: Guess it boils down to what is the purpose of _readyForPeek, which is not clear to me.

tdas (commented):
Thank you very much for identifying this potential source of flakiness. But its not clear to me why another variable _readyForFirstPeek is needed over _isWaiting. Could you add more docs? Or at least explain why a single variable _isWaiting not sufficient to get this logic right?

core/src/main/scala/org/apache/spark/util/ManualClock.scala
lw-lin: `_isWaiting` indicates the main `StreamExecution` thread is waiting, while `_readyForFirstPeek` is supposed to indicate whether it's **the first time** the **test thread** knows about the main thread is waiting.

When it's a second time or a third time the test thread knows the main thread is waiting via `isWaitingAndReadyForFirstPeek()`, the test thread itself should block to prevent `advance()`ing too early.

tdas: I think semantically this ManualClock is becoming a very complex and confusing API to use. The caller has to 

How about changing the API to something like this. 

- ManualClock: It has a method called `isThreadWaitingAt(timeWhenWaitStarted: Long): Boolean`. It returns true when another thread has started waiting when the time was `timeWhenWaitStarted`. This replaces `isWaiting`.

- StreamTest: It keeps track of the expected time when manual clock is being used. For every AdvaneManualClock is first calls `isThreadWaitingAt(expectedTime)` until it returns true, then advances manual clock as well as the expected time.

This will ensure that successive AdvanceManualClock will wait for the StreamExecution thread to be unblocked from the previous wait, and start a new wait on the expected time.

How does this sound?

If this sounds good, then there should be unit tests to test this Manual Clock behavior thoroughly.

-

sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamingQueryListenerSuite.scala
```diff
@@ -81,7 +81,7 @@ class StreamingQueryListenerSuite extends StreamTest with BeforeAndAfter {
       AssertOnLastQueryStatus { status: StreamingQueryStatus =>
         // Check the correctness of the trigger info of the last completed batch reported by
         // onQueryProgress
-        assert(status.triggerDetails.get("triggerId") == "0")
+        assert(Seq("-1", "0").contains(status.triggerDetails.get("triggerId")))
```
tdas: Look at my PR, I replaced `.get` with `.containsKey` because all we need to test here is that the triggerId key is set. The exact trigger id in which the data was found, etc. is not important. 

https://github.com/apache/spark/pull/15492

Could you change this PR to do the same.

core/src/main/scala/org/apache/spark/util/ManualClock.scala
lw-lin: This sounds good!

One thing to confirm: we are assuming only **one** main thread and **one** test thread, right? Because:
- given two main threads m1, m2, and one testing thread t, should `isThreadWaitingAt(time)` return true or false if m1 reaches `waitUtil(time)` but m2 does not yet?
- given one main threads m, and two testing threads t1, t2, should `isThreadWaitingAt(time)` return true for only one of t1, t2 or both?

tdas: I could reproduce this issue as well in a Jenkins run in my PR - https://amplab.cs.berkeley.edu/jenkins/job/NewSparkPullRequestBuilder/3351/consoleFull

sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamingQueryListenerSuite.scala
lw-lin: Sure -- let me do the change, thanks!

lw-lin:
Jenkins retest this please

tdas:
I thought about it, and I still dont like this design. This is adding more complexity in a general class ManualClock, for functionality needed only by StreamExecution. And that leads to these sort of question - should the general feature like `isThreadWaiting` work with multiple threads, etc. 

I think we need to do it differently. I think its best to create a custom ManualClock for StreamExecution, which adds the functionality necessary for StreamExecution.

Mind if I take over this PR and work this out (in the interest of time, 2.0.2 cutoff is imminent)?

tdas:
I opened a PR after modifying your branch #15519. Since you did the initial investigation, I will mark you as the author when I merge it.

lw-lin:
Please go ahead and take over -- let's fix this before 2.0.2, thanks!

PR #15500, opened 2016-10-15
[SPARK-17956][SQL] Fix projection output ordering

viirya:

Currently `ProjectExec` simply takes child plan's `outputOrdering` as its `outputOrdering`. In some cases, this might lead to incorrect `outputOrdering`. This argument applies to `TakeOrderedAndProjectExec` too.

For example, if a plan has output ['a, 'b, 'c], and sort order ['a.asc, 'b.asc, 'c.asc]. After a projection ['b, 'c], the output attribute 'a is not existing now. Then the original sort order ['a.asc, 'b.asc, 'c.asc] is not correct for the projection.
## How was this patch tested?

Jenkins tests.

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/basicPhysicalOperators.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/basicPhysicalOperators.scala
@@ -77,9 +77,40 @@ case class ProjectExec(projectList: Seq[NamedExpression], child: SparkPlan)
     }
   }
 
-  override def outputOrdering: Seq[SortOrder] = child.outputOrdering
+  override def outputOrdering: Seq[SortOrder] =
+    ProjectHelper.outputOrdering(projectList, child.outputOrdering, child)
 }
 
+object ProjectHelper {
+  /**
+   * Determins the outputOrdering property for [[ProjectExec]] and [[TakeOrderedAndProjectExec]]
+   * in a conservative way. We assume any different expresssion to child's sort order expression
+   * will change the sort order. In this case, we drop the sort order. E.g., if the child has the
+   * sort order of 'a.asc and a project expression is UnaryMinus('a), we drop the sort order after
+   * this projection.
+   */
+  def outputOrdering(
+      projectList: Seq[NamedExpression],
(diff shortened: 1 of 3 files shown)


sql/core/src/main/scala/org/apache/spark/sql/execution/basicPhysicalOperators.scala
```diff
@@ -77,9 +77,40 @@ case class ProjectExec(projectList: Seq[NamedExpression], child: SparkPlan)
...
   }
 
-  override def outputOrdering: Seq[SortOrder] = child.outputOrdering
+  override def outputOrdering: Seq[SortOrder] =
+    ProjectHelper.outputOrdering(projectList, child.outputOrdering, child)
 }
 
+object ProjectHelper {
+  /**
+   * Determins the outputOrdering property for [[ProjectExec]] and [[TakeOrderedAndProjectExec]]
```
viirya: If the attribute `a is not existing in the output anymore, isn't sortorder('a) meaningless?

viirya:
also cc @cloud-fan @yhuai

sql/core/src/main/scala/org/apache/spark/sql/execution/basicPhysicalOperators.scala
cloud-fan: yea, it doesn't make sense, but does it cause any problems? I checked the code in `EnsureRequirements`, looks like it's ok to have some useless sort orders.

viirya: Yea, looks like it is. If keeping meaningless sort order seems no harm and we don't require it strictly, we can skip this.

tejasapatil: This is not a correctness issue nor does it buy any performance.

viirya: @tejasapatil Thanks. I just think having a meaningless sort order looks weird. Since it has no harm, I closed this.

PR #15504, opened 2016-10-16
[SPARK-17812][SQL][KAFKA] Assign and specific startingOffsets for structured stream

koeninger:

startingOffsets takes specific per-topicpartition offsets as a json argument, usable with any consumer strategy

assign with specific topicpartitions as a consumer strategy 
## How was this patch tested?

Unit tests


external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/KafkaSource.scala
```diff
@@ -232,6 +232,42 @@ private[kafka010] case class KafkaSource(
...
+  private def fetchSpecificStartingOffsets(
+      partitionOffsets: Map[TopicPartition, Long]): Map[TopicPartition, Long] =
+    withRetriesWithoutInterrupt {
+      // Poll to get the latest assigned partitions
+      consumer.poll(0)
+      val partitions = consumer.assignment()
+      consumer.pause(partitions)
+      assert(partitions.asScala == partitionOffsets.keySet,
+        "If startingOffsets contains specific offsets, you must specify all TopicPartitions.\n" +
+          "Use -1 for latest, -2 for earliest, if you don't care.\n" +
```
ofir-manor: `<if you don't care>`... maybe replace with
`<You can also use -1 for latest, -2 for earliest>` - less judgmental

koeninger: That was pretty much direct from Michael's suggestion on the -1 -2 feature.  I don't care either way about "don't care" ;)

external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/KafkaSource.scala
```diff
@@ -232,6 +232,42 @@ private[kafka010] case class KafkaSource(
...
+        case (tp, -2) => consumer.seekToBeginning(ju.Arrays.asList(tp))
+        case (tp, off) => consumer.seek(tp, off)
+      }
+      val result = partitionOffsets.map {
+        case (tp, _) => tp -> consumer.position(tp)
+      }
+      partitionOffsets.foreach {
+        case (tp, off) if off != -1 && off != -2 =>
+          if (result(tp) != off) {
+            reportDataLoss(
```
zsxwing: move this check out of `withRetriesWithoutInterrupt` to avoid unnecessary retry.

external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/KafkaSource.scala
```diff
@@ -317,6 +353,8 @@ private[kafka010] case class KafkaSource(
               try {
                 result = Some(body)
               } catch {
+                case x: OffsetOutOfRangeException =>
```
zsxwing: When this one will be thrown? I think it only happens when fetching the data. So it's not supposed to happen here. Right?

zsxwing (changes requested):
Just finished my review. The API looks good to me. The major comment from me is just using json4s instead.

external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/JsonUtils.scala
```diff
@@ -0,0 +1,115 @@
...
+import scala.util.control.NonFatal
+
+import com.fasterxml.jackson.databind.ObjectMapper
+import com.fasterxml.jackson.databind.node._
+import org.apache.kafka.common.TopicPartition
+
+/**
+ * Utilities for converting Kafka related objects to and from json.
+ */
+private object JsonUtils {
```
koeninger: Regarding json4s, I haven't benchmarked this particular usage, but my practical experience has been that json4s is noticeably slower than just using jackson directly.  Is this just a code conciseness concern, or is there some other reason to prefer json4s?

external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/KafkaSource.scala
koeninger: OffsetOutOfRange exeception can happen on any poll.  I agree that with the way things are now, it shouldn't happen, but that has more to do with us currently having an explictly enforced value for auto.offset.reset, not with whether data is being fetched. This change is basically paranoia about catching all nonfatal errors without respecting user settings regarding data loss.

external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/JsonUtils.scala
zsxwing: > Is this just a code conciseness concern, or is there some other reason to prefer json4s?

The performance won't be an issue for parsing `assign` and `offsets` since we only parse them once.

> Regarding json4s, I haven't benchmarked this particular usage, but my practical experience has been that json4s is noticeably slower than just using jackson directly.

Just less codes to maintain and easy to understand.

external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/KafkaSource.scala
koeninger: good call, will fix that

external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/JsonUtils.scala
koeninger: Sorry, looks like I edited while you were commenting.  I was thinking towards SPARK-17829, where this same format might get used more frequently throughout a job.

zsxwing: The metadata log only stores `KafkaSourceOffset` which has a different format and is unrelated to this PR.

koeninger: KafkaSourceOffset is a Map[TopicPartition, Long].  This PR has an api for a user to provide a Map[TopicPartition, Long].  There's an existing ticket for making the metadata log store user-readable json.  This PR uses json.  Are they really unrelated?  Why would we want a second, different, json format for the same information....

zsxwing: The input and output of these APIs are string. So they cannot be reused. I would expect SPARK-17829 will send a TreeNode and expect a TreeNode, or we may just use json4s. Since that's unknown until there is a decision, let's just avoid premature optimization in this PR.

zsxwing:
LGTM. Merging to master and 2.0. Thanks!

PR #14151, opened 2016-07-12
[SPARK-16496][SQL] Add wholetext as option for reading text in SQL.

ScrapCodes:
## What changes were proposed in this pull request?

In multiple text analysis problems, it is not often desirable for the rows to be split by "\n". There exists a wholeText reader for RDD API, and this JIRA just adds the same support for Dataset API.
## How was this patch tested?

Added relevant new tests for both scala and Java APIs

PR #15511, opened 2016-10-17
[SPARK-17969]I think it's user unfriendly to process standard json file with DataFrame

codlife:

Currently, with DataFrame API, we can't load standard json file directly, so we can provide an override method to process this.
## How was this patch tested?

manual tests


srowen:
I don't quite understand this -- what does "standard" mean? This still doesn't load a 'standard JSON' file.

codlife:
In standard json file, multi lines json object is legal, but currently, we can just load single-line json obejct directly.

sql/core/src/main/scala/org/apache/spark/sql/DataFrameReader.scala
```diff
@@ -240,16 +240,35 @@ class DataFrameReader private[sql](sparkSession: SparkSession) extends Logging {
...
+  def json(path: String, isStandard: Boolean = false ): DataFrame = {
     // This method ensures that calls that explicit need single argument works, see SPARK-16009
-    json(Seq(path): _*)
+    if (!isStandard) {
+      json(Seq(path): _*)
+    } else {
+      val jsonRDD = sparkSession.sparkContext.wholeTextFiles(path)
+        .map(line => line.toString().replaceAll("\\s+", ""))
+        .map { jsonLine =>
+          val index = jsonLine.indexOf(",")
```
HyukjinKwon: Do you mind if I ask what this line means?

HyukjinKwon:
I guess it'd be nicer if this PR resembles https://github.com/apache/spark/pull/14151
The suggested change is to read each JSON object per file which I guess we can share some codes in the PR.

Also, as we have a `JSONOptions` and `DataFrameReader.option(...)` API, I think it'd be nicer if this one is added as an option rather than introducing another API.

HyukjinKwon:
BTW, I guess per-line JSON also complies a standard - https://tools.ietf.org/html/rfc7159#section-4. We should add a test, fix the title to summarise what the PR proposes and fill the PR description. I think also we can also alternatively close this, wait until 14151 is merged and then open again whan you are ready to start working on this..

codlife:
Compile is ok, but when we call show(), we will get a _corrupt_record, besides when we call select on this df, we will get an exception.

srowen:
OK, I think in both cases "standard" JSON is read, and in both cases, each record is a JSON document. These aren't different cases. If you mean to read small JSON files as records, you just use wholeTextFiles, as you show. I do not think wrapping this up with an extra flag helps enough to justify this because callers can easily implement this. There are a hundred other variations on this, and the reason we don't implement them all is exactly because there are so many variations to bottle up like this.

codlife:
@srowen , you are right! I propose this method just to make it more user friendly,  With this method, user can load a standard json file directly.

sql/core/src/main/scala/org/apache/spark/sql/DataFrameReader.scala
codlife: maybe this code is bad, I just want to get the json contents
such as: ("filename",json_contents)

srowen:
I think we should close this. I don't believe it's worth a new API method.

PR #15512, opened 2016-10-17
[SPARK-17930][CORE]The SerializerInstance instance used when deserializing a TaskResult is not reused 

witgo:

The following code is called when the DirectTaskResult instance is deserialized

``` scala

  def value(): T = {
    if (valueObjectDeserialized) {
      valueObject
    } else {
      // Each deserialization creates a new instance of SerializerInstance, which is very time-consuming
      val resultSer = SparkEnv.get.serializer.newInstance()
      valueObject = resultSer.deserialize(valueBytes)
      valueObjectDeserialized = true
      valueObject
    }
  }

```

In the case of stage has a lot of tasks, reuse SerializerInstance instance can improve the scheduling performance of three times

The test data is TPC-DS 2T (Parquet) and  SQL statement as follows (query 2):

``` sql

select  i_item_id, 
        avg(ss_quantity) agg1,
        avg(ss_list_price) agg2,
        avg(ss_coupon_amt) agg3,
        avg(ss_sales_price) agg4 
 from store_sales, customer_demographics, date_dim, item, promotion
 where ss_sold_date_sk = d_date_sk and
       ss_item_sk = i_item_sk and
       ss_cdemo_sk = cd_demo_sk and
       ss_promo_sk = p_promo_sk and
       cd_gender = 'M' and 
       cd_marital_status = 'M' and
       cd_education_status = '4 yr Degree' and
       (p_channel_email = 'N' or p_channel_event = 'N') and
       d_year = 2001 
 group by i_item_id
 order by i_item_id
 limit 100;

```

`spark-defaults.conf` file:

```
spark.master                           yarn-client
spark.executor.instances               20
spark.driver.memory                    16g
spark.executor.memory                  30g
spark.executor.cores                   5
spark.default.parallelism              100 
spark.sql.shuffle.partitions           100000 
spark.serializer                       org.apache.spark.serializer.KryoSerializer
spark.driver.maxResultSize              0
spark.rpc.netty.dispatcher.numThreads   8
spark.executor.extraJavaOptions          -XX:+UseG1GC -XX:+UseStringDeduplication -XX:G1HeapRegionSize=16M -XX:MetaspaceSize=256M 
spark.cleaner.referenceTracking.blocking true
spark.cleaner.referenceTracking.blocking.shuffle true

```

Performance test results are as follows 

| [SPARK-17930](https://github.com/witgo/spark/tree/SPARK-17930) | [ed14633](https://github.com/witgo/spark/commit/ed1463341455830b8867b721a1b34f291139baf3]) |
| --- | --- |
| 54.5 s | 231.7 s |
## How was this patch tested?

Existing tests.

### DIFF
--- a/core/src/main/scala/org/apache/spark/scheduler/TaskResult.scala
+++ b/core/src/main/scala/org/apache/spark/scheduler/TaskResult.scala
@@ -23,6 +23,7 @@ import java.nio.ByteBuffer
 import scala.collection.mutable.ArrayBuffer
 
 import org.apache.spark.SparkEnv
+import org.apache.spark.serializer.SerializerInstance
 import org.apache.spark.storage.BlockId
 import org.apache.spark.util.{AccumulatorV2, Utils}
 
@@ -77,14 +78,12 @@ private[spark] class DirectTaskResult[T](
    *
    * After the first time, `value()` is trivial and just returns the deserialized `valueObject`.
    */
-  def value(): T = {
+  def value(resultSer: SerializerInstance = null): T = {
     if (valueObjectDeserialized) {
       valueObject
     } else {
-      // This should not run when holding a lock because it may cost dozens of seconds for a large
-      // value.
-      val resultSer = SparkEnv.get.serializer.newInstance()
-      valueObject = resultSer.deserialize(valueBytes)
+      valueObject = Option(resultSer).getOrElse(SparkEnv.get.serializer.newInstance())
+        .deserialize(valueBytes)
       valueObjectDeserialized = true
       valueObject
     }
--- a/core/src/main/scala/org/apache/spark/scheduler/TaskResult.scala
+++ b/core/src/main/scala/org/apache/spark/scheduler/TaskResult.scala
@@ -82,8 +82,10 @@ private[spark] class DirectTaskResult[T](
     if (valueObjectDeserialized) {
       valueObject
     } else {
-      valueObject = Option(resultSer).getOrElse(SparkEnv.get.serializer.newInstance())
-        .deserialize(valueBytes)
+      // This should not run when holding a lock because it may cost dozens of seconds for a large
+      // value
+      val ser = if (resultSer == null) SparkEnv.get.serializer.newInstance() else resultSer
+      valueObject = ser.deserialize(valueBytes)
       valueObjectDeserialized = true
       valueObject
     }
--- a/core/src/main/scala/org/apache/spark/scheduler/TaskResultGetter.scala
+++ b/core/src/main/scala/org/apache/spark/scheduler/TaskResultGetter.scala
@@ -90,6 +90,7 @@ private[spark] class TaskResultGetter(sparkEnv: SparkEnv, scheduler: TaskSchedul
               }
               val deserializedResult = serializer.get().deserialize[DirectTaskResult[_]](
                 serializedTaskResult.get.toByteBuffer)
+              // force deserialization of referenced value
               deserializedResult.value(taskResultSerializer.get())
               sparkEnv.blockManager.master.removeBlock(blockId)
               (deserializedResult, size)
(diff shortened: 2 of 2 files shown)


hvanhovell:
ok to test

srowen:
Hm, if the benchmark you give generalizes much that is certainly compelling. I guess I'm surprised that instantiating the object can be so expensive relative to deserialization since it just happens once per task. 

But it is a fairly simple change. `ThreadLocal` avoids thread safety issues though I do wonder if the serializers can hold on to state that would make this a source of memory leak?

Maybe ... @squito or @zsxwing has a thought?

squito:
serializing will create buffers, but since these are only used for deserializing, I don't think there should even be any buffers created.  I guess the time saved is all the registration which can be skipped?  https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/serializer/KryoSerializer.scala#L85.  I suppose in this case, this is the result of `spark.sql.shuffle.partitions=100000`, so you end up calling this method a lot.

The only wrinkle I can see here is if reference-tracking is turned on (which it is, by default).  But I think this is taken care of anyway by the way `borrowKryo`.

squito:
Jenkins, retest this please

core/src/main/scala/org/apache/spark/scheduler/TaskResult.scala
```diff
@@ -77,14 +78,12 @@ private[spark] class DirectTaskResult[T](
    *
    * After the first time, `value()` is trivial and just returns the deserialized `valueObject`.
    */
-  def value(): T = {
+  def value(resultSer: SerializerInstance = null): T = {
     if (valueObjectDeserialized) {
       valueObject
     } else {
-      // This should not run when holding a lock because it may cost dozens of seconds for a large
```
zsxwing: nit: please keep the comment.

core/src/main/scala/org/apache/spark/scheduler/TaskResult.scala
```diff
@@ -77,14 +78,12 @@ private[spark] class DirectTaskResult[T](
...
-      // value.
-      val resultSer = SparkEnv.get.serializer.newInstance()
-      valueObject = resultSer.deserialize(valueBytes)
+      valueObject = Option(resultSer).getOrElse(SparkEnv.get.serializer.newInstance())
```
zsxwing: nit: I prefer the following codes because `Option(...).getOrElse` doesn't improve the readability but creates unnecessary objects.
```Scala
val resultSer = if (resultSer == null) SparkEnv.get.serializer.newInstance() else resultSer	
valueObject = resultSer.deserialize(valueBytes)
```

zsxwing (changes requested):
Looks pretty good just some nits.

core/src/main/scala/org/apache/spark/scheduler/TaskResultGetter.scala
```diff
@@ -84,6 +90,7 @@ private[spark] class TaskResultGetter(sparkEnv: SparkEnv, scheduler: TaskSchedul
               }
               val deserializedResult = serializer.get().deserialize[DirectTaskResult[_]](
                 serializedTaskResult.get.toByteBuffer)
+              deserializedResult.value(taskResultSerializer.get())
```
mridulm: nit: Would be nice to add a comment here saying "force deserialization of referenced value" or some such

core/src/main/scala/org/apache/spark/scheduler/TaskResult.scala
witgo: Done.

core/src/main/scala/org/apache/spark/scheduler/TaskResult.scala
witgo: Done.

core/src/main/scala/org/apache/spark/scheduler/TaskResultGetter.scala
witgo: Done.

witgo:
I also think that the time saved is all the registration which can be skipped, but did not verify.

zsxwing:

PR #15514, opened 2016-10-17
[SPARK-17960][PySpark] [Upgrade to Py4J 0.10.4]

jagadeesanas2:

1) Upgrade the Py4J version on the Java side
2) Update the py4j src zip file we bundle with Spark
## How was this patch tested?

Existing doctests & unit tests pass


holdenk:
Awesome, thanks for taking this on @jagadeesanas2, this looks pretty reasonable although it might be helpful if we mentioned briefly the PR description what the Py4J changes were in this release since that is what ends up as the commit description once its merged. But the change its self looks good to me - cc @srowen / @davies for the final review.

davies:
cc @JoshRosen

srowen:
I'm OK with it. For the record, was there any fix here that impacts Spark? updating just to keep updated isn't a bad idea but wouldn't want to update through every version of everything. Maybe at least once per minor release it's worth review.

srowen:
Merged to master

PR #15515, opened 2016-10-17
[SPARK-17970][SQL] store partition spec in metastore for data source table

cloud-fan:

We should follow hive table and also store partition spec in metastore for data source table.
This brings 2 benefits:
1. It's more flexible to manage the table data files, as users can use `ADD PARTITION`, `DROP PARTITION` and `RENAME PARTITION`
2. We don't need to cache all file status for data source table anymore.
## How was this patch tested?

existing tests.


cloud-fan:
cc @yhuai

yhuai:
Also cc @ericl

ericl:
Also cc @mallman

ericl:
Could we add a feature flag for this? Then it would be easier to write tests for backwards compatibility against old data-source tables.

ericl:
@cloud-fan I think this is pretty close once https://github.com/apache/spark/pull/15539 is merged.

The only remaining issues I know about are (1) supporting mixed case columns, and (2) optimizing the insert command to not call repair table. But as discussed we can address these separately.

ericl:
Looks like two failing tests in FileStreamSourceSuite

ericl:
Ok, I think this is now ready to review. I've pushed commits that add a feature flag (consolidated with prior feature flags to make it easy to revert back to 2.0 behavior), backwards compatibility, and tests cases.

The proposal for backwards compatibility is as follows.
- Newly created datasource tables will have the table property `partitionProvider=hive` and will use the metastore for partition discovery.
- Older Spark versions will not respect the partitionProvider property, and will always discover partitions from remote storage.
- When partitionProvider is set to hive, partition related hive DDLs (e.g. SHOW PARTITIONS) will work. Otherwise, an exception will be thrown telling the user to run `msck repair table` to import the table partitions into the hive metastore (this will set partitionProvider = hive).
- When the feature flag is disabled, tables with hive partition provider will behave like legacy tables. The partition DDLs will also be disabled for all datasource tables.

sql/core/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala
```diff
@@ -272,18 +272,20 @@ object SQLConf {
       .booleanConf
       .createWithDefault(true)
 
-  val HIVE_FILESOURCE_PARTITION_PRUNING =
-    SQLConfigBuilder("spark.sql.hive.filesourcePartitionPruning")
```
davies: Is this one public? For public one, should we deprecate it first (not just remove it)?

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSourceStrategy.scala
```diff
@@ -188,6 +188,15 @@ case class DataSourceAnalysis(conf: CatalystConf) extends Rule[LogicalPlan] {
         t.options,
         query,
         mode)
+
+      if (l.catalogTable.isDefined && l.catalogTable.get.partitionColumnNames.nonEmpty &&
+          l.catalogTable.get.partitionProviderIsHive) {
+        // TODO(ekl) we should be more efficient here and only recover the newly added partitions
```
davies: Should we use addPartition? (recover partition is known slow)

davies (commented):
Had quick look, looks good to me overall. It's better to have another people looking at the details.

sql/core/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala
ericl: It's only in master, so changing it should be ok here.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSourceStrategy.scala
ericl: This seems non-trivial to implement, so I'll probably leave it to a follow-up pr.

ericl:
Rebased

sql/core/src/main/scala/org/apache/spark/sql/execution/command/AnalyzeColumnCommand.scala
```diff
@@ -50,7 +50,8 @@ case class AnalyzeColumnCommand(
           AnalyzeTableCommand.calculateTotalSize(sessionState, catalogRel.catalogTable))
 
       case logicalRel: LogicalRelation if logicalRel.catalogTable.isDefined =>
-        updateStats(logicalRel.catalogTable.get, logicalRel.relation.sizeInBytes)
+        updateStats(logicalRel.catalogTable.get,
+          AnalyzeTableCommand.calculateTotalSize(sessionState, logicalRel.catalogTable.get))
```
cloud-fan: Why change this? `AnalyzeTableCommand.calculateTotalSize` reads a special hive property `hive.exec.stagingdir`, does data source table also need to take care of it?

sql/core/src/main/scala/org/apache/spark/sql/execution/command/createDataSourceTables.scala
```diff
@@ -91,13 +91,25 @@ case class CreateDataSourceTableCommand(table: CatalogTable, ignoreIfExists: Boo
       table.storage.properties
     }
 
+    val newProps = if (partitionColumnNames.nonEmpty &&
+        sparkSession.sqlContext.conf.manageFilesourcePartitions) {
+      // Start off with partition provider hive, but no partitions in the metastore. The user
+      // has to call `msck repair table` to populate the table partitions.
+      table.properties ++
+        Map(CatalogTable.PARTITION_PROVIDER_KEY -> CatalogTable.PARTITION_PROVIDER_HIVE)
```
cloud-fan: Can users set/unset this special property? I guess no? and `DESC TABLE`, `SHOW CREATE TABLE` should also exclude it right?

sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveExternalCatalog.scala
```diff
@@ -581,13 +582,30 @@ private[spark] class HiveExternalCatalog(conf: SparkConf, hadoopConf: Configurat
   // Partitions
   // --------------------------------------------------------------------------
 
+  // Hive metastore is not case preserving and the partition columns are always lower cased. We need
+  // to lower case the column names in partition specification before calling partition related Hive
+  // APIs, to match this behaviour.
+  private def lowerCasePartitionSpec(spec: TablePartitionSpec): TablePartitionSpec = {
```
cloud-fan: cc @ericl @yhuai, I have isolated these hive partition hacks in `HiveExternalCatalog`, so that `HiveExternalCatalog` can provides case preserving partition related interfaces.

sql/hive/src/test/scala/org/apache/spark/sql/hive/PartitionProviderCompatibilitySuite.scala
```diff
@@ -0,0 +1,138 @@
...
+import org.apache.spark.sql.{AnalysisException, QueryTest}
+import org.apache.spark.sql.hive.test.TestHiveSingleton
+import org.apache.spark.sql.internal.SQLConf
+import org.apache.spark.sql.test.SQLTestUtils
+
+class PartitionProviderCompatibilitySuite
+  extends QueryTest with TestHiveSingleton with SQLTestUtils {
+
+  private def setupPartitionedDatasourceTable(tableName: String, dir: File): Unit = {
+    // TODO(ekl) make these mixed-case fields once support for that is fixed
```
cloud-fan: try it again?

sql/hive/src/test/scala/org/apache/spark/sql/hive/PartitionProviderCompatibilitySuite.scala
```diff
@@ -0,0 +1,138 @@
...
+          assert(spark.sql("select * from test").count() == 0)
+        }
+        // disabled
+        withSQLConf(SQLConf.HIVE_MANAGE_FILESOURCE_PARTITIONS.key -> "false") {
+          val e = intercept[AnalysisException] {
+            spark.sql(s"show partitions test")
+          }
+          assert(e.getMessage.contains("filesource partition management is disabled"))
+          spark.sql("refresh table test")
+          assert(spark.sql("select * from test").count() == 5)
```
cloud-fan: what's the result if we run this without `refresh table`?

sql/hive/src/test/scala/org/apache/spark/sql/hive/PartitionProviderCompatibilitySuite.scala
```diff
@@ -0,0 +1,138 @@
...
+        }
+        // then enabled again
+        withSQLConf(SQLConf.HIVE_MANAGE_FILESOURCE_PARTITIONS.key -> "true") {
+          spark.sql("refresh table test")
+          assert(spark.sql("select * from test").count() == 0)
```
cloud-fan: what's the result if we run this without `refresh table`?

sql/core/src/main/scala/org/apache/spark/sql/execution/command/AnalyzeColumnCommand.scala
ericl: This is needed since the logical relation `sizeInBytes` is no longer the latest size. Previously ListingFileCatalog would compute the size, but TableFileCatalog just reads whatever size was present in the table metadata last.

sql/core/src/main/scala/org/apache/spark/sql/execution/command/createDataSourceTables.scala
ericl: I'll remove it from `SHOW CREATE TABLE`, along with `transient_lastDdlTime`. For desc table, I think it would be misleading to not show it, since otherwise users could not debug the partition provider if needed. We can rename it to `org.apache.spark/partitionProvider` to make it more clear Spark is managing this property.

sql/hive/src/test/scala/org/apache/spark/sql/hive/PartitionProviderCompatibilitySuite.scala
ericl: Will do

sql/hive/src/test/scala/org/apache/spark/sql/hive/PartitionProviderCompatibilitySuite.scala
ericl: You'll get the old cached results. I think this is kind of fair since otherwise we'd have to include the flag value in the cache key, which would be kind of odd.

sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveExternalCatalog.scala
ericl: cool

sql/core/src/main/scala/org/apache/spark/sql/DataFrameWriter.scala
```diff
@@ -387,7 +388,15 @@ final class DataFrameWriter[T] private[sql](ds: Dataset[T]) {
           partitionColumnNames = partitioningColumns.getOrElse(Nil),
           bucketSpec = getBucketSpec
         )
-        val cmd = CreateTable(tableDesc, mode, Some(df.logicalPlan))
+        val createCmd = CreateTable(tableDesc, mode, Some(df.logicalPlan))
+        val cmd = if (tableDesc.partitionColumnNames.nonEmpty &&
+            df.sparkSession.sqlContext.conf.manageFilesourcePartitions) {
+          // Need to recover partitions into the metastore so our saved data is visible.
+          val recoverPartitionCmd = AlterTableRecoverPartitionsCommand(tableDesc.identifier)
+          Union(createCmd, recoverPartitionCmd)
```
ericl: I think this ok, otherwise you'd need a lot of special logic just for this.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/interface.scala
```diff
@@ -217,8 +218,24 @@ case class CatalogTable(
...
+   */
+  def partitionProviderIsHive: Boolean = {
+    import CatalogTable._
+    provider == Some("hive") ||
+      properties.get(PARTITION_PROVIDER_KEY) == Some(PARTITION_PROVIDER_HIVE)
+  }
 }
 
+object CatalogTable {
+  val PARTITION_PROVIDER_KEY = "org.apache.spark/partitionProvider"
```
cloud-fan: add some comments to explain this format?

cloud-fan:
LGTM, I added a new commit to make `partitionProviderIsHive` a field and put it in table properties in `HiveExternalCatalog`, to simplify the code a little bit. Feel free to revert it if it breaks something.

ericl:
That LGTM as well. I pushed a small fix for the failing debug mode test.

ericl:
```
[info] - SPARK-10562: partition by column with mixed case name *** FAILED *** (605 milliseconds)
[info]   java.lang.reflect.InvocationTargetException:
[info]   at sun.reflect.GeneratedMethodAccessor156.invoke(Unknown Source)
[info]   at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[info]   at java.lang.reflect.Method.invoke(Method.java:497)
[info]   at org.apache.spark.sql.hive.client.Shim_v0_13.getPartitionsByFilter(HiveShim.scala:588)
```

Odd that the last commit could cause this, maybe it's a flake? jenkins test this please

cloud-fan:
retest this please

cloud-fan:
retest this please

sql/core/src/main/scala/org/apache/spark/sql/DataFrameWriter.scala
yhuai: Let's add a special node for running a sequence of commands. We are relying on the implementation of Union at here. Let's address this in a follow-up PR.

sql/core/src/main/scala/org/apache/spark/sql/execution/command/AnalyzeColumnCommand.scala
yhuai: How's the cost of `AnalyzeTableCommand.calculateTotalSize`? Also, why is `sizeInBytes` not the latest size?

sql/core/src/main/scala/org/apache/spark/sql/execution/command/createDataSourceTables.scala
```diff
@@ -232,6 +238,15 @@ case class CreateDataSourceTableAsSelectCommand(
       sessionState.catalog.createTable(newTable, ignoreIfExists = false)
     }
 
+    result match {
+      case fs: HadoopFsRelation if table.partitionColumnNames.nonEmpty &&
+          sparkSession.sqlContext.conf.manageFilesourcePartitions =>
+        // Need to recover partitions into the metastore so our saved data is visible.
+        sparkSession.sessionState.executePlan(
+          AlterTableRecoverPartitionsCommand(table.identifier)).toRdd
```
yhuai: Seems we have a similar logic at https://github.com/apache/spark/pull/15515/files#diff-94fbd986b04087223f53697d4b6cab24R396. Are we recovering partitions twice?

sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
```diff
@@ -531,6 +529,11 @@ case class AlterTableRecoverPartitionsCommand(
     logInfo(s"Finished to gather the fast stats for all $total partitions.")
 
     addPartitions(spark, table, partitionSpecsAndLocs, partitionStats)
+    // Updates the table to indicate that its partition metadata is stored in the Hive metastore.
+    // This is always the case for Hive format tables, but is not true for Datasource tables created
+    // before Spark 2.1 unless they are converted via `msck repair table`.
+    spark.sessionState.catalog.alterTable(table.copy(partitionProviderIsHive = true))
+    catalog.refreshTable(tableName)
```
yhuai: Let's also update the doc.

yhuai:
Looks good. I left a few questions. Let me know if you want to address them in follow-up prs.

sql/core/src/main/scala/org/apache/spark/sql/execution/command/createDataSourceTables.scala
ericl: I just double checked with `create table test_sel2 USING parquet PARTITIONED BY (fieldone, fieldtwo) AS SELECT id as fieldzero, id as fieldone, id as fieldtwo from range(100)` and it is using a different path, so the recovery is not duplicated.

sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
ericl: Yeah, this is a pretty major change for 2.1. Shall we do it in a followup once the patches for 2.1 are finalized?

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSourceStrategy.scala
ericl: I have a follow-up pr for this. Will cc you.

ericl (approved):

yhuai:
Cool. I am merging this pr to unblock other tasks.

sql/hive/src/test/scala/org/apache/spark/sql/hive/PartitionProviderCompatibilitySuite.scala
```diff
@@ -0,0 +1,137 @@
...
+        withSQLConf(SQLConf.HIVE_MANAGE_FILESOURCE_PARTITIONS.key -> "false") {
+          setupPartitionedDatasourceTable("test", dir)
+          verifyIsLegacyTable("test")
+          assert(spark.sql("select * from test").count() == 5)
+        }
+      }
+    }
+  }
+
+  test("when partition management is disabled, we preserve the old behavior even for new tables") {
```
gatorsmile: I just checked the old behavior. It is different from the existing behavior in our Spark 2.0 build. Let me do a quick fix to resolve it.

PR #13780, opened 2016-06-20
[SPARK-16063][SQL] Add storageLevel to Dataset

MLnick:
[SPARK-11905](https://issues.apache.org/jira/browse/SPARK-11905) added support for `persist`/`cache` for `Dataset`. However, there is no user-facing API to check if a `Dataset` is cached and if so what the storage level is. This PR adds `getStorageLevel` to `Dataset`, analogous to `RDD.getStorageLevel`.
## How was this patch tested?

Updated `DatasetCacheSuite`.

PR #15516, opened 2016-10-17
[SPARK-17961][SparkR][SQL] Add storageLevel to DataFrame for SparkR

WeichenXu123:

Add storageLevel to DataFrame for SparkR.
This is similar to this RP:  https://github.com/apache/spark/pull/13780

but in R I do not make a class for `StorageLevel`
but add a method `storageToString`
## How was this patch tested?

test added.


R/pkg/R/DataFrame.R
```diff
@@ -654,6 +654,33 @@ setMethod("unpersist",
...
           })
 
+#' StorageLevel
+#'
+#' Get storage level of this SparkDataFrame.
+#'
+#' @param x the SparkDataFrame to get the storage level.
+#'
+#' @family SparkDataFrame functions
+#' @rdname storageLevel-methods
```
felixcheung: change this storageLevel - to match the method name

R/pkg/inst/tests/testthat/test_sparkSQL.R
```diff
@@ -795,6 +795,8 @@ test_that("cache(), persist(), and unpersist() on a DataFrame", {
   persist(df, "MEMORY_AND_DISK")
   expect_true(df@env$isCached)
 
+  expect_equal(storageLevel(df), "StorageLevel(disk, memory, deserialized, 1 replicas)")
```
felixcheung: so the output of this doesn't say "MEMORY_AND_DISK"? Should we have that in addition to "StorageLevel(disk, memory, deserialized, 1 replicas)"? It might be confusing to set "MEMORY_AND_DISK" and get "StorageLevel(disk, memory, deserialized, 1 replicas)" back?

WeichenXu123: good suggestion, I'll update the code later. thanks!

WeichenXu123:
@felixcheung code updated. thanks!

R/pkg/R/utils.R
```diff
@@ -385,6 +385,41 @@ getStorageLevel <- function(newLevel = c("DISK_ONLY",
                          "OFF_HEAP" = callJStatic(storageLevelClass, "OFF_HEAP"))
 }
 
+storageLevelToString <- function(levelObj) {
+  useDisk <- callJMethod(levelObj, "useDisk")
+  useMemory <- callJMethod(levelObj, "useMemory")
+  useOffHeap <- callJMethod(levelObj, "useOffHeap")
+  deserialized <- callJMethod(levelObj, "deserialized")
+  replication <- callJMethod(levelObj, "replication")
+  if (!useDisk && !useMemory && !useOffHeap && !deserialized && replication == 1) {
```
felixcheung: hardcoding the variations in R could be hard to maintain or easily get out of sync. is there anyway to do this?
Python seems to be able to get the enum name as a string

WeichenXu123: python has itself `StorageLevel` class, and the python side code about `storageLevel`  also exists  duplicated code problem...
and if we make an r-side `StorageLevel` class may cause the code more complex and seems won't help solving the duplicated code problem.
What do you think about it ?

WeichenXu123: and, about the R-side String constant, is there better way to avoid duplicated literal constant in code ? such as "MEMORY_AND_DISK", does we need to define some global vars, such as
MEMORY_AND_DISK_CONSTANT <- "MEMORY_AND_DISK" ?
and where could we put the definition above?  if use this way.

felixcheung: I see. Class in R wouldn't help much in this case.
You could have a look up table - check out https://github.com/apache/spark/blob/master/R/pkg/R/types.R and how it is used

felixcheung:
I think you've committed a jar file by accident

WeichenXu123:
@felixcheung
Remove the unrelated jar file.
and about the `String`  look up table, here seems there are not the mapping relationship between these String constant, so that the code I thinks the code just keep it current status is fine, no need to add some look-up table.

felixcheung:
I see. I understand the constraint here. I'd hold for a bit to see if anyone else has any thought on this?

Also, I'd think it would be useful to output both the short name + long description (from toString)
eg.
```
MEMORY_AND_DISK - Serialized 1x Replicated
```
or similar. Perhaps later on we could deprecate the `MEMORY_AND_DISK` short names like in Scala or Python.

WeichenXu123:
@felixcheung @yanboliang thanks!

felixcheung:
did you get my comment previously here https://github.com/apache/spark/pull/15516#discussion_r83920746

R/pkg/R/DataFrame.R
felixcheung: this should be 
`@rdname storageLevel` instead of 
`@rdname storageLevel-methods`

WeichenXu123:
@felixcheung update rdname, `unpersited-method` also updated by the way.

felixcheung:
To clarify, we have the method name in the @rdname tag.

In this case it should be
@rdname storageLevel

And not storageLevel-methods

felixcheung:
merged to master.

PR #15519, opened 2016-10-17
[SQL][STREAMING][TEST] Fix flaky tests in StreamingQueryListenerSuite

tdas:
This work has largely been done by @lw-lin in his PR #15497. This is a slight refactoring of it.

There were two sources of flakiness in StreamingQueryListener test.
- When testing with manual clock, consecutive attempts to advance the clock can occur without the stream execution thread being unblocked and doing some work between the two attempts. Hence the following can happen with the current ManualClock.

```
+-----------------------------------+--------------------------------+
|      StreamExecution thread       |         testing thread         |
+-----------------------------------+--------------------------------+
|  ManualClock.waitTillTime(100) {  |                                |
|        _isWaiting = true          |                                |
|            wait(10)               |                                |
|        still in wait(10)          |  if (_isWaiting) advance(100)  |
|        still in wait(10)          |  if (_isWaiting) advance(200)  | <- this should be disallowed !
|        still in wait(10)          |  if (_isWaiting) advance(300)  | <- this should be disallowed !
|      wake up from wait(10)        |                                |
|       current time is 600         |                                |
|       _isWaiting = false          |                                |
|  }                                |                                |
+-----------------------------------+--------------------------------+
```
- Second source of flakiness is that the adding data to memory stream may get processing in any trigger, not just the first trigger.

My fix is to make the manual clock wait for the other stream execution thread to start waiting for the clock at the right wait start time. That is, `advance(200)` (see above) will wait for stream execution thread to complete the wait that started at time 0, and start a new wait at time 200 (i.e. time stamp after the previous `advance(100)`).

In addition, since this is a feature that is solely used by StreamExecution, I removed all the non-generic code from ManualClock and put them in StreamManualClock inside StreamTest.
## How was this patch tested?

Ran existing unit test MANY TIME in Jenkins


tdas:
@lw-lin  please take a look.

tdas:
I have tested enough. There was a single failure in a different flaky test, not in StreamingQueryListenerSuite.

lw-lin:
This looks good to me, thanks!

sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamSuite.scala
```diff
@@ -199,7 +199,7 @@ class StreamSuite extends StreamTest {
 
       /* Stop then restart the Stream  */
       StopStream,
-      StartStream(ProcessingTime("10 seconds"), new ManualClock),
+      StartStream(ProcessingTime("10 seconds"), new ManualClock(60 * 1000)),
```
tdas: Huh! I wonder how the test passed without this change.

tdas: Oh I never ran the StreamSuite in jenkins till now.

tdas:
@lw-lin Fixed the bug.

tdas:
Merging this to master and branch 2.0

PR #15520, opened 2016-10-17
[SPARK-13747][SQL]Fix concurrent executions in ForkJoinPool for SQL

zsxwing:

Calling `Await.result` will allow other tasks to be run on the same thread when using ForkJoinPool. However, SQL uses a `ThreadLocal` execution id to trace Spark jobs launched by a query, which doesn't work perfectly in ForkJoinPool.

This PR just uses `Awaitable.result` instead to  prevent ForkJoinPool from running other tasks in the current waiting thread.
## How was this patch tested?

Jenkins

### DIFF
--- a/scalastyle-config.xml
+++ b/scalastyle-config.xml
@@ -200,6 +200,7 @@ This file is divided into 3 sections:
       // scalastyle:off awaitresult
       Await.result(...)
       // scalastyle:on awaitresult
+      If your codes use ThreadLocal and run in a user's thread, use ThreadUtils.awaitResultInForkJoinSafely instead.
     ]]></customMessage>
   </check>
 
--- a/scalastyle-config.xml
+++ b/scalastyle-config.xml
@@ -200,7 +200,7 @@ This file is divided into 3 sections:
       // scalastyle:off awaitresult
       Await.result(...)
       // scalastyle:on awaitresult
-      If your codes use ThreadLocal and run in a user's thread, use ThreadUtils.awaitResultInForkJoinSafely instead.
+      If your codes use ThreadLocal and may run in threads created by the user, use ThreadUtils.awaitResultInForkJoinSafely instead.
     ]]></customMessage>
   </check>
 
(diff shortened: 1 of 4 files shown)


jodersky:
I'm not sure I completely understand why you have to use `Awaitable` instead of `Await`. Is it to avoid the [`blocking()` call in `Await`](https://github.com/scala/scala/blob/v2.12.0-RC1/src/library/scala/concurrent/package.scala#L168)?

zsxwing:
> Is it to avoid the blocking() call in Await?

Yep.

jodersky:
That makes sense, thanks for explaining. Would installing a custom `BlockingContext` on threads that run tasks also work? I'm not very familiar with the internals of the Awaitable api so I'm just throwing this out there in the hopes you can tell me more :)

zsxwing:
> Would installing a custom BlockingContext on threads that run tasks also work?

Yes, I think so. It just needs more lines :)

zsxwing:
retest this please

jodersky:
> Yes, I think so. It just needs more lines :)

I think the solution you propose is sound. Do you think that a quick mention about avoiding the `BlockingContext` in a comment would help in case someone sees this code in the future and is puzzled as to what it does?

zsxwing:
cc @andrewor14

scalastyle-config.xml
```diff
@@ -200,6 +200,7 @@ This file is divided into 3 sections:
       // scalastyle:off awaitresult
       Await.result(...)
       // scalastyle:on awaitresult
+      If your codes use ThreadLocal and run in a user's thread, use ThreadUtils.awaitResultInForkJoinSafely instead.
```
yhuai: What does the `user's thread` mean at here?

zsxwing: I meant threads created by the user. Fixed the description.

yhuai:
lgtm

zsxwing:

zsxwing:
There are conflicts with 2.0. #15646 is the backport for 2.0.

PR #15523, opened 2016-10-17
[SPARK-17981] [SPARK-17957] [SQL] Fix Incorrect Nullability Setting to False in FilterExec

gatorsmile:

When `FilterExec` contains `isNotNull`, which could be inferred and pushed down or users specified, we convert the nullability of the involved columns if the top-layer expression is null-intolerant. However, this is not correct, if the top-layer expression is not a leaf expression, it could still tolerate the null when it has null-tolerant child expressions. 

For example, `cast(coalesce(a#5, a#15) as double)`. Although `cast` is a null-intolerant expression, but obviously`coalesce` is null-tolerant. Thus, it could eat null.

When the nullability is wrong, we could generate incorrect results in different cases. For example, 

``` Scala
    val df1 = Seq((1, 2), (2, 3)).toDF("a", "b")
    val df2 = Seq((2, 5), (3, 4)).toDF("a", "c")
    val joinedDf = df1.join(df2, Seq("a"), "outer").na.fill(0)
    val df3 = Seq((3, 1)).toDF("a", "d")
    joinedDf.join(df3, "a").show
```

The optimized plan is like 

```
Project [a#29, b#30, c#31, d#42]
+- Join Inner, (a#29 = a#41)
   :- Project [cast(coalesce(cast(coalesce(a#5, a#15) as double), 0.0) as int) AS a#29, cast(coalesce(cast(b#6 as double), 0.0) as int) AS b#30, cast(coalesce(cast(c#16 as double), 0.0) as int) AS c#31]
   :  +- Filter isnotnull(cast(coalesce(cast(coalesce(a#5, a#15) as double), 0.0) as int))
   :     +- Join FullOuter, (a#5 = a#15)
   :        :- LocalRelation [a#5, b#6]
   :        +- LocalRelation [a#15, c#16]
   +- LocalRelation [a#41, d#42]
```

Without the fix, it returns an empty result. With the fix, it can return a correct answer:

```
+---+---+---+---+
|  a|  b|  c|  d|
+---+---+---+---+
|  3|  0|  4|  1|
+---+---+---+---+
```

Added test cases to verify the nullability changes in FilterExec. Also added a test case for verifying the reported incorrect result.


gatorsmile:
cc @cloud-fan @davies @sameeragarwal

sql/core/src/main/scala/org/apache/spark/sql/execution/basicPhysicalOperators.scala
```diff
@@ -87,7 +87,14 @@ case class FilterExec(condition: Expression, child: SparkPlan)
...
   private val (notNullPreds, otherPreds) = splitConjunctivePredicates(condition).partition {
-    case IsNotNull(a: NullIntolerant) if a.references.subsetOf(child.outputSet) => true
+    case IsNotNull(a) => isNullIntolerant(a) && a.references.subsetOf(child.outputSet)
+    case _ => false
+  }
+
+  // One expression is null intolerant iff it and its children are null intolerant
+  private def isNullIntolerant(expr: Expression): Boolean = expr match {
+    case e: NullIntolerant =>
+      if (e.isInstanceOf[LeafExpression]) true else e.children.forall(isNullIntolerant)
```
gatorsmile: Just realized the original code was from your PR. Then, in your above code, why you still need to keep `a.references.subsetOf(child.outputSet)`? It looks confusing to me.

viirya: Even `a` passed the check of `isNullIntolerant`, i.e., it has not non `NullIntolerant` which wraps output attributes. If it doesn't refer to any output attributes, we don't need it.

gatorsmile: Could you show me an example?

gatorsmile: uh, I see. 

First, we definitely need test cases to cover each positive and negative scenario. Previously, we did not have any test case to check the validity of nullability changes. Second, the code needs more comments when the variable/function names are not able to explain the codes.

gatorsmile:
@viirya I have not changed the algorithm. I just tried to improve the test case coverage. 

Thanks to `constructIsNotNullConstraints`, the existing solution already covers all the cases, right?

Can you help me check anything scenario is missing? Thanks!

viirya:
@gatorsmile A predicate like `IsNotNull(a + b + Rand())` will let this change to wrongly set the nullability of `a` and `b` to true. Isn't it?

gatorsmile:
The parm name of the verification function is wrong. It should be `expectedNonNullableColumns`. Please check the test case again.

gatorsmile:
cc @cloud-fan @davies @sameeragarwal @viirya

sql/core/src/test/scala/org/apache/spark/sql/DataFrameSuite.scala
```diff
@@ -1617,6 +1617,72 @@ class DataFrameSuite extends QueryTest with SharedSQLContext {
...
+      new java.lang.Integer(2) -> new java.lang.Integer(4))).toDF()
+
+    verifyNullabilityInFilterExec(df,
+      expr = "_1 + _2 * 3", expectedNonNullableColumns = Seq("_1", "_2"))
+    verifyNullabilityInFilterExec(df,
+      expr = "_1 + _2", expectedNonNullableColumns = Seq("_1", "_2"))
+    verifyNullabilityInFilterExec(df,
+      expr = "_1", expectedNonNullableColumns = Seq("_1"))
+    verifyNullabilityInFilterExec(df,
+      expr = "_2 + Rand()", expectedNonNullableColumns = Seq("_2"))
```
viirya: This should not work in current approach. It works now because we infer redundant IsNotNull constraints. E.g., if Filter has a constraint IsNotNull(_2 + Rand()), we will infer another IsNotNull(_2) from it. Your approach is working on IsNotNull(_2) to decide _2 is non-nullable column, not IsNotNull(_2 + Rand()).

I submitted another PR #15653 for redundant IsNotNull constraints. But I am not sure if we want to fix it since it doesn't affect correctness. I left that to @cloud-fan or @hvanhovell to decide it.

gatorsmile: I already explained why my current solution works in [my previous statement](https://github.com/apache/spark/pull/15523#issuecomment-254931776). Personally, I like simple code, which is **easy to understand and maintain**, especially when it can cover all the cases. Result correctness and code maintainability are alwasy more important.

If `constructIsNotNullConstraints ` is changed by somebody else (i.e., it does not provide the expected `IsNotNull` constraints), the test cases added by this PR will fail. Then, we can modify the codes.

viirya: Yeah, so I said I will left that to @cloud-fan or others to decide...

viirya:
LGTM and wait for @cloud-fan or others to do second check.

sql/core/src/main/scala/org/apache/spark/sql/execution/basicPhysicalOperators.scala
```diff
@@ -89,7 +89,14 @@ case class FilterExec(condition: Expression, child: SparkPlan)
...
   private val (notNullPreds, otherPreds) = splitConjunctivePredicates(condition).partition {
-    case IsNotNull(a: NullIntolerant) if a.references.subsetOf(child.outputSet) => true
+    case IsNotNull(a) => isNullIntolerant(a) && a.references.subsetOf(child.outputSet)
+    case _ => false
+  }
+
+  // If one expression and its children are null intolerant, it is null intolerant.
+  private def isNullIntolerant(expr: Expression): Boolean = expr match {
+    case e: NullIntolerant =>
+      if (e.isInstanceOf[LeafExpression]) true else e.children.forall(isNullIntolerant)
```
sameeragarwal: nit: How about something like this for better readability:

```scala
  private def isNullIntolerant(expr: Expression): Boolean = expr match {
    case e: NullIntolerant with LeafExpression => true
    case e: NullIntolerant => e.children.forall(isNullIntolerant)
    case _ => false
  }
```

sql/core/src/test/scala/org/apache/spark/sql/DataFrameSuite.scala
sameeragarwal: I agree with the simplicity argument but can you please add a comment here explaining why this particular case is working due to the null inference rule?

sameeragarwal (approved):
Thanks, this LGTM too with just a couple of extremely minor comments. As far as the change is concerned, I tend to side with simplicity argument for now (especially if we're planning to target 2.1).

sql/core/src/main/scala/org/apache/spark/sql/execution/basicPhysicalOperators.scala
gatorsmile: `forall` will return `true` if the `children` is empty. Thus, we can remove the first case. Now it becomes simpler.

```Scala
  private def isNullIntolerant(expr: Expression): Boolean = expr match {
    case e: NullIntolerant => e.children.forall(isNullIntolerant)
    case _ => false
  }
```

sameeragarwal: nice

hvanhovell:

hvanhovell:
@gatorsmile can you open a PR for 2.0 if we also need to port it to that branch?

gatorsmile:
Sure, @hvanhovell  will do it. Thanks!

PR #15526, opened 2016-10-18
[SPARK-17986] [ML] SQLTransformer should remove temporary tables

drewrobb:

A call to the method `SQLTransformer.transform` previously would create a temporary table and never delete it. This change adds a call to `dropTempView()` that deletes this temporary table before returning the result so that the table will not remain in spark's table catalog. Because `tableName` is randomized and not exposed, there should be no expected use of this table outside of the `transform` method.
## How was this patch tested?

A single new assertion was added to the existing test of the `SQLTransformer.transform` method that all temporary tables are removed. Without the corresponding code change, this new assertion fails. I am not aware of any circumstances in which removing this temporary view would be bad for performance or correctness in other ways, but some expertise here would be helpful.

### DIFF
--- a/mllib/src/test/scala/org/apache/spark/ml/feature/SQLTransformerSuite.scala
+++ b/mllib/src/test/scala/org/apache/spark/ml/feature/SQLTransformerSuite.scala
@@ -43,6 +43,7 @@ class SQLTransformerSuite
     assert(result.schema.toString == resultSchema.toString)
     assert(resultSchema == expected.schema)
     assert(result.collect().toSeq == expected.collect().toSeq)
+    assert(original.sparkSession.catalog.listTables().count() == 0)
   }
 
   test("read/write") {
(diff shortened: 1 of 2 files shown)


mllib/src/test/scala/org/apache/spark/ml/feature/SQLTransformerSuite.scala
```diff
@@ -43,6 +43,7 @@ class SQLTransformerSuite
     assert(result.schema.toString == resultSchema.toString)
     assert(resultSchema == expected.schema)
     assert(result.collect().toSeq == expected.collect().toSeq)
+    assert(original.sparkSession.catalog.listTables().count() == 0)
```
drewrobb: No, I believe `isEmpty` is only currently defined for an RDD, not a Dataset, you could do `collect().isEmpty` but that would be equally verbose I think.

drewrobb:
@yanboliang would you be able to review this patch?

yanboliang:
Jenkins, test this please.

yanboliang:
ok to test

yanboliang:
Add to whitelist.

yanboliang:
ok to test.

yanboliang:
cc @srowen @MLnick @jkbradley @rxin Could you add @drewrobb to whitelist? I can not trigger the Jenkins job. Thanks.

MLnick:
Jenkins add to whitelist

yanboliang:
Good catch! LGTM, merged into master and branch-2.0. Thanks! @drewrobb @leoromanovsky

PR #15531, opened 2016-10-18
[SQL][STREAMING][TEST] Follow up to remove Option.contains for Scala 2.10 compatibility

tdas:

Scala 2.10 does not have Option.contains, which broke Scala 2.10 build. 
## How was this patch tested?

Locally compiled and ran sql/core unit tests in 2.10

### DIFF
--- a/sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamTest.scala
+++ b/sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamTest.scala
@@ -216,7 +216,9 @@ trait StreamTest extends QueryTest with SharedSQLContext with Timeouts {
       }
     }
 
-    def isStreamWaitingAt(time: Long): Boolean = synchronized { waitStartTime.contains(time) }
+    def isStreamWaitingAt(time: Long): Boolean = synchronized {
+      waitStartTime == Some(time)
+    }
   }
 
 


tdas:
I will merge this as soon as compilation passes.

srowen:
Ha we have been bit by this like 3 times in the past month.
Time to begin agitating to remove 2.10 support.

tdas:
We should ping the dev list for something like this. 
For now, I am merging this fix. Compilation has passed in 2.11 in Jenkins, and I locally compiled and tested in 2.10.

rxin:
Hm we should ban Option.contains. It's not a good idea anyway.

markhamstra:
@rxin Nonsense; it's just as good an idea as are List.contains and Set.contains.  The only problem with it is that it doesn't exist in Scala 2.10.

rxin:
I disagree. It makes it very confusing, especially without explicit types, to know what exactly is happening. Not everything should be a monad.

srowen:
Let the religious wars commence. If Option is a "collection of 0 or 1 things" then `.contains` at least makes some sense, if a bit obtuse. I still think `== Some(x)` is clearer anyway.

mridulm:
@srowen Unfortunately the 'Some' in the == is usually missed out, resulting in bugs (and we have had a fair share of them in the past - some more severe than others).
Given that there is no support for contains in 2.10, the alternative is moot ofcourse.

PR #15534, opened 2016-10-18
[SPARK-17917][Scheduler] Fire SparkListenerTasksStarved event when submitted tasks are starved beyond the configured starvation timeout

mariobriggs:

Provide applications that are interested in knowing when submitted tasks are starved with that ability
## How was this patch tested?

Added testcase SparkListenerStarvedSuite in file     org/apache/spark/scheduler/SparkListenerWithClusterSuite.java

### DIFF
--- a/core/src/main/scala/org/apache/spark/scheduler/SparkListener.scala
+++ b/core/src/main/scala/org/apache/spark/scheduler/SparkListener.scala
@@ -131,6 +131,13 @@ case class SparkListenerApplicationStart(
 @DeveloperApi
 case class SparkListenerApplicationEnd(time: Long) extends SparkListenerEvent
 
+@DeveloperApi
+case class SparkListenerTasksStarved(starvationStartTime: Long) extends SparkListenerEvent
+
+@DeveloperApi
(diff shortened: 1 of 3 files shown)


tejasapatil:
You are introducing 2 new events but they are not being consumed. Do you plan to use those anywhere ? Whats is use case that you are trying to solve ?

rxin:
The user can track this pretty easily themselves, can't they?

mariobriggs:
>>
The user can track this pretty easily themselves, can't they?
<<
Can you explain a little more on your line of thinking here?  I am more than happy to not have this code added if there exists other ways to track this.

rxin:
If a job is submitted and there is no task start update for a while?

mariobriggs:
thanks. Sounds possible. Let me try it that way with my app and i can then close this as required.

mariobriggs:
@rxin while doing this the user way, had to copy the exact StarvationTimer thread code/logic and flags that already exists in TaskSchedulerImpl.scala to the app Listener. 

I am fine if opinion is not to merge. My 2 cents is that simpler for user since TaskSchedulerImpl already has the boilerplate code.

mariobriggs:
Closing, since can implement in the app.

PR #15538, opened 2016-10-18
[SPARK-17993][SQL] Fix Parquet log output redirection

mallman:
(Link to Jira issue: https://issues.apache.org/jira/browse/SPARK-17993)

PR #14690 broke parquet log output redirection for converted partitioned Hive tables. For example, when querying parquet files written by Parquet-mr 1.6.0 Spark prints a torrent of (harmless) warning messages from the Parquet reader:

```
Oct 18, 2016 7:42:18 PM WARNING: org.apache.parquet.CorruptStatistics: Ignoring statistics because created_by could not be parsed (see PARQUET-251): parquet-mr version 1.6.0
org.apache.parquet.VersionParser$VersionParseException: Could not parse created_by: parquet-mr version 1.6.0 using format: (.+) version ((.*) )?\(build ?(.*)\)
    at org.apache.parquet.VersionParser.parse(VersionParser.java:112)
    at org.apache.parquet.CorruptStatistics.shouldIgnoreStatistics(CorruptStatistics.java:60)
    at org.apache.parquet.format.converter.ParquetMetadataConverter.fromParquetStatistics(ParquetMetadataConverter.java:263)
    at org.apache.parquet.hadoop.ParquetFileReader$Chunk.readAllPages(ParquetFileReader.java:583)
    at org.apache.parquet.hadoop.ParquetFileReader.readNextRowGroup(ParquetFileReader.java:513)
    at org.apache.spark.sql.execution.datasources.parquet.VectorizedParquetRecordReader.checkEndOfRowGroup(VectorizedParquetRecordReader.java:270)
    at org.apache.spark.sql.execution.datasources.parquet.VectorizedParquetRecordReader.nextBatch(VectorizedParquetRecordReader.java:225)
    at org.apache.spark.sql.execution.datasources.parquet.VectorizedParquetRecordReader.nextKeyValue(VectorizedParquetRecordReader.java:137)
    at org.apache.spark.sql.execution.datasources.RecordReaderIterator.hasNext(RecordReaderIterator.scala:39)
    at org.apache.spark.sql.execution.datasources.FileScanRDD$$anon$1.hasNext(FileScanRDD.scala:102)
    at org.apache.spark.sql.execution.datasources.FileScanRDD$$anon$1.nextIterator(FileScanRDD.scala:162)
    at org.apache.spark.sql.execution.datasources.FileScanRDD$$anon$1.hasNext(FileScanRDD.scala:102)
    at org.apache.spark.sql.catalyst.expressions.GeneratedClass$GeneratedIterator.scan_nextBatch$(Unknown Source)
    at org.apache.spark.sql.execution.BufferedRowIterator.hasNext(BufferedRowIterator.java:43)
    at org.apache.spark.sql.execution.WholeStageCodegenExec$$anonfun$8$$anon$1.hasNext(WholeStageCodegenExec.scala:372)
    at org.apache.spark.sql.execution.SparkPlan$$anonfun$2.apply(SparkPlan.scala:231)
    at org.apache.spark.sql.execution.SparkPlan$$anonfun$2.apply(SparkPlan.scala:225)
    at org.apache.spark.rdd.RDD$$anonfun$mapPartitionsInternal$1$$anonfun$apply$24.apply(RDD.scala:803)
    at org.apache.spark.rdd.RDD$$anonfun$mapPartitionsInternal$1$$anonfun$apply$24.apply(RDD.scala:803)
    at org.apache.spark.rdd.MapPartitionsRDD.compute(MapPartitionsRDD.scala:38)
    at org.apache.spark.rdd.RDD.computeOrReadCheckpoint(RDD.scala:319)
    at org.apache.spark.rdd.RDD.iterator(RDD.scala:283)
    at org.apache.spark.scheduler.Task.run(Task.scala:99)
```

This only happens during execution, not planning, and it doesn't matter what log level the `SparkContext` is set to. That's because Parquet (versions < 1.9) doesn't use slf4j for logging. Note, you can tell that log redirection is not working here because the log message format does not conform to the default Spark log message format.

This is a regression I noted as something we needed to fix as a follow up.

It appears that the problem arose because we removed the call to `inferSchema` during Hive table conversion. That call is what triggered the output redirection.

## How was this patch tested?

I tested this manually in four ways:
1. Executing `spark.sqlContext.range(10).selectExpr("id as a").write.mode("overwrite").parquet("test")`.
2. Executing `spark.read.format("parquet").load(legacyParquetFile).show` for a Parquet file `legacyParquetFile` written using Parquet-mr 1.6.0.
3. Executing `select * from legacy_parquet_table limit 1` for some unpartitioned Parquet-based Hive table written using Parquet-mr 1.6.0.
4. Executing `select * from legacy_partitioned_parquet_table where partcol=x limit 1` for some partitioned Parquet-based Hive table written using Parquet-mr 1.6.0.

I ran each test with a new instance of `spark-shell` or `spark-sql`.

Incidentally, I found that test case 3 was not a regression—redirection was not occurring in the master codebase prior to #14690.

I spent some time working on a unit test, but based on my experience working on this ticket I feel that automated testing here is far from feasible.

cc @ericl @dongjoon-hyun

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetFileFormat.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetFileFormat.scala
@@ -146,7 +146,14 @@ class ParquetFileFormat
     }
 
     new OutputWriterFactory {
-      override def newInstance(
+        // OutputWriterFactory is deserialized in the write path on the executor side before any
+        // output is actually written. Redirect Parquet logs at this time.
+        private def readObject(in: ObjectInputStream): Unit = {
+          in.defaultReadObject
+          ParquetFileFormat.ensureParquetLogRedirection
+        }
+
+        override def newInstance(
           path: String,
           dataSchema: StructType,
           context: TaskAttemptContext): OutputWriter = {
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetFileFormat.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetFileFormat.scala
@@ -56,7 +56,13 @@ class ParquetFileFormat
   with DataSourceRegister
   with Logging
   with Serializable {
-  ParquetFileFormat.redirectParquetLogs
+  // Poor man's "static initializer". Scala doesn't have language support for static initializers,
+  // and it's important that we initialize `ParquetFileFormat.redirectParquetLogsViaSLF4J` before
+  // doing anything with the Parquet libraries. Rather than expect clients to initialize the
+  // `ParquetFileFormat` singleton object at the right time, we put that initialization in the
+  // constructor of this class. This method is idempotent, and essentially a no-op after its first
+  // call.
+  ParquetFileFormat.ensureParquetLogRedirection
 
   override def shortName(): String = "parquet"
 
@@ -709,8 +715,9 @@ object ParquetFileFormat extends Logging {
   }
 
   /**
-   * The ParquetFileFormat constructor calls this function to initialize
-   * `redirectParquetLogsViaSLF4J`.
+   * The `ParquetFileFormat` constructor calls this method to ensure that Parquet library log
+   * output is redirected through the SLF4J JUL bridge handler. This method is a no-op because we
+   * only require that `redirectParquetLogsViaSLF4J` is initialized. This accomplishes that task.
    */
-  private def redirectParquetLogs(): Unit = {}
+  private def ensureParquetLogRedirection(): Unit = {}
 }
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetFileFormat.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetFileFormat.scala
@@ -716,8 +716,7 @@ object ParquetFileFormat extends Logging {
 
   /**
    * The `ParquetFileFormat` constructor calls this method to ensure that Parquet library log
-   * output is redirected through the SLF4J JUL bridge handler. This method is a no-op because we
-   * only require that `redirectParquetLogsViaSLF4J` is initialized. This accomplishes that task.
+   * output is redirected through the SLF4J JUL bridge handler.
    */
-  private def ensureParquetLogRedirection(): Unit = {}
+  private def ensureParquetLogRedirection(): Unit = redirectParquetLogsViaSLF4J
 }
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetFileFormat.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetFileFormat.scala
@@ -56,6 +56,7 @@ class ParquetFileFormat
   with DataSourceRegister
   with Logging
   with Serializable {
+  ParquetFileFormat.redirectParquetLogs
 
   override def shortName(): String = "parquet"
 
@@ -128,8 +129,6 @@ class ParquetFileFormat
     if (conf.get(ParquetOutputFormat.ENABLE_JOB_SUMMARY) == null) {
       conf.setBoolean(ParquetOutputFormat.ENABLE_JOB_SUMMARY, false)
     }
-
-    ParquetFileFormat.redirectParquetLogs()
 
     new OutputWriterFactory {
       override def newInstance(
@@ -682,7 +681,7 @@ object ParquetFileFormat extends Logging {
 
   // Parquet initializes its own JUL logger in a static block which always prints to stdout.  Here
   // we redirect the JUL logger via SLF4J JUL bridge handler.
-  val redirectParquetLogsViaSLF4J: Unit = {
+  private val redirectParquetLogsViaSLF4J: Unit = {
     def redirect(logger: JLogger): Unit = {
       logger.getHandlers.foreach(logger.removeHandler)
       logger.setUseParentHandlers(false)
@@ -710,7 +709,8 @@ object ParquetFileFormat extends Logging {
   }
 
   /**
-   * ParquetFileFormat.prepareWrite calls this function to initialize `redirectParquetLogsViaSLF4J`.
+   * The ParquetFileFormat constructor calls this function to initialize
+   * `redirectParquetLogsViaSLF4J`.
    */
-  def redirectParquetLogs(): Unit = {}
+  private def redirectParquetLogs(): Unit = {}
 }
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetFileFormat.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetFileFormat.scala
@@ -17,9 +17,7 @@
 
 package org.apache.spark.sql.execution.datasources.parquet
 
-import java.io.ObjectInputStream
 import java.net.URI
-import java.util.logging.{Logger => JLogger}
 
 import scala.collection.JavaConverters._
 import scala.collection.mutable
@@ -30,14 +28,12 @@ import org.apache.hadoop.fs.{FileStatus, Path}
 import org.apache.hadoop.mapreduce._
 import org.apache.hadoop.mapreduce.lib.input.FileSplit
 import org.apache.hadoop.mapreduce.task.TaskAttemptContextImpl
-import org.apache.parquet.{Log => ApacheParquetLog}
 import org.apache.parquet.filter2.compat.FilterCompat
 import org.apache.parquet.filter2.predicate.FilterApi
 import org.apache.parquet.hadoop._
 import org.apache.parquet.hadoop.codec.CodecConfig
 import org.apache.parquet.hadoop.util.ContextUtil
 import org.apache.parquet.schema.MessageType
-import org.slf4j.bridge.SLF4JBridgeHandler
 
 import org.apache.spark.{SparkException, TaskContext}
 import org.apache.spark.internal.Logging
@@ -57,21 +53,11 @@ class ParquetFileFormat
   with DataSourceRegister
   with Logging
   with Serializable {
-  // Poor man's "static initializer". Scala doesn't have language support for static initializers,
-  // and it's important that we initialize `ParquetFileFormat.redirectParquetLogsViaSLF4J` before
-  // doing anything with the Parquet libraries. Rather than expect clients to initialize the
-  // `ParquetFileFormat` singleton object at the right time, we put that initialization in the
-  // constructor of this class. This method is idempotent, and essentially a no-op after its first
-  // call.
-  ParquetFileFormat.ensureParquetLogRedirection
-
-  // Java serialization will not call the default constructor. Make sure we call
-  // ParquetFileFormat.ensureParquetLogRedirection in deserialization by implementing this hook
-  // method.
-  private def readObject(in: ObjectInputStream): Unit = {
-    in.defaultReadObject
-    ParquetFileFormat.ensureParquetLogRedirection
-  }
+  // Hold a reference to the (serializable) singleton instance of ParquetLogRedirector. This
+  // ensures the ParquetLogRedirector class is initialized whether an instance of ParquetFileFormat
+  // is constructed or deserialized. Do not heed the Scala compiler's warning about an unused field
+  // here.
+  private val parquetLogRedirector = ParquetLogRedirector.INSTANCE
 
   override def shortName(): String = "parquet"
 
@@ -146,12 +132,11 @@ class ParquetFileFormat
     }
 
     new OutputWriterFactory {
-        // OutputWriterFactory is deserialized in the write path on the executor side before any
-        // output is actually written. Redirect Parquet logs at this time.
-        private def readObject(in: ObjectInputStream): Unit = {
-          in.defaultReadObject
-          ParquetFileFormat.ensureParquetLogRedirection
-        }
+      // This OutputWriterFactory instance is deserialized when writing Parquet files on the
+      // executor side without constructing or deserializing ParquetFileFormat. Therefore, we hold
+      // another reference to ParquetLogRedirector.INSTANCE here to ensure the latter class is
+      // initialized.
+      private val parquetLogRedirector = ParquetLogRedirector.INSTANCE
 
         override def newInstance(
           path: String,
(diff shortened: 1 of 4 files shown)


sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetFileFormat.scala
```diff
@@ -56,6 +56,7 @@ class ParquetFileFormat
   with DataSourceRegister
   with Logging
   with Serializable {
+  ParquetFileFormat.redirectParquetLogs
```
ericl: Should we only invoke this once per jvm?

mallman: Essentially, yes.

mallman: The tricky bits are _when_, _where_ and _how_.

ericl: If you moved this into an object that was referenced by this class on init, would that suffice?

mallman: I'm not sure I understand what you're suggesting. Can you elaborate further?

srowen: You should just be able to put this in the class init for `ParquetFileFormat`, by adding this to the body of `object ParquetFileFormat` below.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetFileFormat.scala
```diff
@@ -718,7 +717,8 @@ object ParquetFileFormat extends Logging {
   }
 
   /**
-   * ParquetFileFormat.prepareWrite calls this function to initialize `redirectParquetLogsViaSLF4J`.
+   * The ParquetFileFormat constructor calls this function to initialize
+   * `redirectParquetLogsViaSLF4J`.
    */
-  def redirectParquetLogs(): Unit = {}
+  private def redirectParquetLogs(): Unit = {}
```
srowen: Probably a dumb question but this is a no-op? how does this work?

srowen (changes requested):

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetFileFormat.scala
mallman: That doesn't work. To wit: http://stackoverflow.com/questions/2347107/what-is-scala-equivalent-of-javas-static-block.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetFileFormat.scala
mallman: If I just call `redirectParquetLogsViaSLF4J` I get a compilation error:

```
[error] [warn] /Volumes/VideoAmpCS/msa/workspace/spark-master/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetFileFormat.scala:59: a pure expression does nothing in statement position; you may be omitting necessary parentheses
[error] [warn]   ParquetFileFormat.redirectParquetLogsViaSLF4J
```

mallman:
I could use some advice on writing a unit test for this. Do you guys know if there is a precedent in the codebase that covers a situation like this? I'd like to reuse existing code if possible.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetFileFormat.scala
srowen: Yes, because it's not a method. I'm wondering what this method does, and I think it exists only to cause the `object` to initialize. That's awkward, but it already exists I guess. Is there really no way to implement a static init block in Scala? hm. If so then I support this change. It's actually being moved to a spot where it will be called earlier, and, be called fewer times, right?

ericl: Maybe it would be clearer to call it once with an atomic boolean guard:

```
if (ParquetFileFormat.logsRedirected.compareAndSwap(false, true)) {
  ParquetFileFormat.redirectParquetLogsViaSLF4J()
}
```

Then you could test that the flag is set after constructing a ParquetFileFormat. That's obviously not a great test but probably good enough.

mallman: I'll try to clean things up in another commit. If nothing else, it's clear I need to add some code comments.

mallman:
I pushed a commit to improve the documentation. I also removed a couple of unused imports (boy scout rule).

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetFileFormat.scala
```diff
@@ -718,7 +722,9 @@ object ParquetFileFormat extends Logging {
   }
 
   /**
-   * ParquetFileFormat.prepareWrite calls this function to initialize `redirectParquetLogsViaSLF4J`.
+   * The `ParquetFileFormat` constructor calls this method to ensure that Parquet library log
+   * output is redirected through the SLF4J JUL bridge handler. This method is a no-op because we
+   * only require that `redirectParquetLogsViaSLF4J` is initialized. This accomplishes that task.
    */
-  def redirectParquetLogs(): Unit = {}
+  private def ensureParquetLogRedirection(): Unit = {}
```
srowen: OK this is better, though on a second look, it seems less indirect to either:

a) use a `lazy val` above and simply do anything with that reference that compiles, like log it or something. It didn't before because just referencing a field and doing nothing else will cause a warning, or

b) make it a method that executes once with a simple `AtomicBoolean` guard or boolean + synchronization

I suppose I'm wondering whether even this mechanism is guaranteed to work.

mallman:
I've spent a couple hours today working on a unit test which captures stdout during a parquet write operation to validate that it has no parquet logging output. I haven't got it working yet, but I'll spend some more time on it tomorrow.

ericl:
also LGTM as-is; it's polluting test output which is kind of annoying. Testing the log output might be a little overkill since this is not critical functionality and has an upstream fix.

mallman:
@ericl What do you mean it's polluting test output?

ericl:
I see ~10k lines of "CorruptStatistics" stack traces in the jenkins log. Though, it also seems to show up in the test logs for this pr, so maybe it wouldn't affect this.

mallman:
I'll take a closer look at that.

mallman:
The "CorruptStatistics" stack traces (which I agree are really annoying) are being logged because parquet logs them at the WARN level, and Spark's default logging threshold when running tests is WARN.

I've pushed a commit to set the parquet log threshold to ERROR for testing. This is actually the default configuration for a Spark deployment: https://github.com/apache/spark/blob/v2.0.1/conf/log4j.properties.template#L35.

mallman:
I like this test failure:

```
org.apache.spark.sql.sources.CreateTableAsSelectSuite.(It is not a test)
```

Anyway, I don't think this is related to this PR.

mallman:
So raising the log threshold looks like it didn't do anything for Jenkins, but when I run the tests locally it does just the trick. \*sigh\*

Anyway, might as well push a rebase and see what happens.

mallman:
I'm still seeing the torrent of `CorruptStatistics` errors in the Jenkins build log, even though I don't see them running the tests locally with sbt. Maybe it's a maven versus sbt build issue.

Anyway, having spent a few hours trying to get a reliable, effective unit test for this functionality, I don't think it's going to happen.

Can you guys let me know if you have a blocking concern around this PR? Otherwise, can we please merge this to master? I'm reluctant to spend more time on this.

Cheers.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetFileFormat.scala
srowen: I suppose it's no worse than the existing setup, though surely there's a better way to get once-per-class execution. Would you try one more thing: 

`private def ensureParquetLogRedirection(): Unit = redirectParquetLogsViaSLF4J`

_That_ would at least connect the things that are supposed to be connected though it's all a no-op and would look less wonky to me.

If that still generates a warning OK let's merge this.

mallman: I believe the log output redirection is working. The mystery here is why the test log level threshold is not having an effect in Jenkins.

Nevertheless I'll make the change and push it.

mallman:
Looks like the test failed for reasons unrelated to this PR. Can someone trigger a retest, please?

ericl:
Jenkins retest this please

On Thu, Oct 27, 2016, 12:34 PM Michael Allman <notifications@github.com>
wrote:

Looks like the test failed for reasons unrelated to this PR. Can someone
trigger a retest, please?

—
<https://github.com/apache/spark/pull/15538#issuecomment-256745918>, or mute
the thread

srowen:
Jenkins add to whitelist

srowen (approved):
Great, if this still works, this looks more understandable to me.

mallman:
@srowen I ran my manual tests for this build and they worked as expected. Can you merge this PR?

mallman:
Actually, this may not be working for remote executors. I tested this patch running in local mode, but in running a version of this with actual remote executors I'm seeing the original parquet log output.

Let me take a closer look before merging this PR.

srowen:
OK, well whatever works here, even if somehow the previous version is what's needed to get it to work.

mallman:
Getting this redirection to work on remote executors was quite involved. The additional complexities derive from the following:

1. Java doesn't call the default (or any) constructor in deserialization.
2. Spark doesn't use `ParquetFileFormat` at all in the write path.

To address the first issue, I added a deserialization hook to `ParquetFileFormat` which calls `ParquetFileFormat.ensureParquetLogRedirection`. To address the second, I added deserialization hooks to a couple of classes the driver serializes when it sends the output job to the executors.

This is messy, but it's the simplest, most direct way I found to tackle these complexities.

Hopefully all of this awfulness can be removed when Spark updates its dependency on parquet-mr to the 1.9 release, which uses slf4j for logging.

srowen:
If this is necessary, then isn't it simpler to leave the log configuration call where it was, so that it doesn't depend on the constructor? that wasn't the actual problem was it, just the logging? if the problem is that it fires repeatedly, then a flag could be used to record that it has executed?

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetFileFormat.scala
```diff
@@ -55,6 +56,21 @@ class ParquetFileFormat
...
+  // constructor of this class. This method is idempotent, and essentially a no-op after its first
+  // call.
+  ParquetFileFormat.ensureParquetLogRedirection
+
+  // Java serialization will not call the default constructor. Make sure we call
+  // ParquetFileFormat.ensureParquetLogRedirection in deserialization by implementing this hook
+  // method.
+  private def readObject(in: ObjectInputStream): Unit = {
+    in.defaultReadObject
+    ParquetFileFormat.ensureParquetLogRedirection
```
ericl: You could also call `ensureParquetLogRedirection` from some main class right? e.g. `class Executor`.

mallman:
Are we planning to incorporate the Parquet 1.9 libraries into Spark 2.1? If so, then this PR should be unnecessary.

Hopefully.

mallman:
In fact, if no one else is working on the Parquet upgrade it probably makes more sense for me to contribute that then continue working on this PR. I'll check with the dev mailing list.

srowen:
Yeah there's a ticket for it to solve a different parquet bug. That might be the path of least resistance.

mallman:
I found two such tickets. How should we organize this in Jira?

srowen:
Ah right. Let me redirect 18140 to 13127 because all of these end up being resolved by "upgrade to 1.9"

rxin:
@mallman are you working on Parquet 1.9 upgrade? Would be great to get that in sooner than later because in the past Parquet upgrades tend to bring a lot of issues, so it'd be good to go through more QA testing.

mallman:
Yes, I am working on it. I'm planning to have a PR in no later than EOW.

mallman:
@rxin One of our weekly Spark jobs is choking, and fixing it could take the rest of the week. I suggest someone else take the lead on the Parquet 1.9 upgrade if they can devote their time to it.

cc @rdblue

mallman:
Hi Guys,

Unfortunately debugging our Spark job sucked up all my Spark time last week, and I still have more to do on it this week. Because of that, it doesn't look like I'll have time to work on the Parquet 1.9 upgrade after all.

FWIW, I did some work on it, got everything to compile and all the unit tests pass. However, I simply have to change my focus this week and can't devote time to another PR.

I'll rebase this patch and hope to get it in in case a 1.9 upgrade doesn't happen for Spark 2.1.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetFileFormat.scala
mallman: I've tried to keep this code isolated to the parquet implementation.

mallman:
> If this is necessary, then isn't it simpler to leave the log configuration call where it was, so that it doesn't depend on the constructor? that wasn't the actual problem was it, just the logging? if the problem is that it fires repeatedly, then a flag could be used to record that it has executed?

Sorry @srowen, I don't understand what you're asking here. However, I can say that calling this method repeatedly is not the problem. The problem is getting it to be called before Parquet starts logging.

mallman:
I've pushed a rebase. I re-tested this PR using the methodology I describe in the description for both local and remote executors.

rxin:
@mallman are you also working on Parquet upgrade? If we don't get it in in the next day or two, we shouldn't merge that into branch-2.1 anymore since it is way past feature freeze.

srowen:
Yeah I see, but this is getting to be quite hacky just to turn off log messages, trying to make sure static init is triggered from a custom deserializer. It feels like this should be punted until Parquet 1.9 takes care of it.

mallman:
@rxin I'm not working on the Parquet upgrade this week. I think we'll have to punt on it.

mallman:
> Yeah I see, but this is getting to be quite hacky **just to turn off log messages**

This isn't just a few annoying log messages. This is an _avalanche_ of log messages, each of which contains a deep stack trace.

You can read https://issues.apache.org/jira/browse/SPARK-17993?focusedCommentId=15645322&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-15645322 for another perspective.

I agree this patch is ugly. I don't like writing code like this. I just don't see a feasible alternative for Spark 2.1. It can be removed as part of a future Parquet 1.9 upgrade.

I abbreviated the console output when I created SPARK-17993. Running a `spark-sql` query with 10 results on a table written with Spark 1.5 results in 7,052 lines of console output. (The last 10 are the actual results.)

srowen:
OK, maybe. It is not just about being hacky but being likely to break. If this really calls for a static initializer, how about a Java helper class to do the work? How about piggy backing on the Logging init routines? I don't see this as something without alternatives that needs to be pushed through

mallman:
Sorry, I don't see how Java code will make this patch any better. We're not really missing a static initializer. We just need _some_ initializer to run early enough.

If we can tolerate putting parquet-specific code in datasource-agnostic code there are a number of places where we can put a call to the log redirection code. I've tried to avoid putting parquet specific code outside the parquet specific codebase.

srowen:
If `ParquetFileFormat` had a static init block somehow, we'd be done right? because the logging config in that static initializer would have to execute, once, during classloading and therefore before any usage in a constructor or serialization. But per above, you can't write such a block in Scala. So, make a dummy `ParquetLogHelper.java` class, and make `ParquetFileFormat` have a dummy field of its type or something. `ParquetLogHelper` can have a static init, and must load when `ParquetFileFormat` loads. I might miss a reason that doesn't work, but seems like the sort of thing to rule out before settling on a complicated solution. It would be a little simpler than the current code.

I don't think it's so wrong to add logic directly to Logging.scala as a fairly temporary workaround. It wouldn't entail an actual compile or runtime dependency on Parquet in core. At least for me that seems simpler than the heroic attempt to intercept serialization and constructor paths. I'd hope it's sufficient to add it to Logging, which should have to init very early, everywhere, once.

mallman:
I'll work on a revision and try to push something today.

mallman:
Pushed a new version which I think is cleaner than before. I tested all 8 scenarios manually.

srowen:
I like this. It centralizes and yet elaborates the handling of the issue, does seem to be the right mechanism to use, and I presume, does work reliably.

mallman:
@ericl Does this LGTY?

ericl (approved):

rxin:
I don't think we are going to upgrade Parquet for branch-2.1, since it's way past that point.  Let's merge this for 2.1.

Merging in master/branch-2.1.

PR #15539, opened 2016-10-19
[SPARK-17994] [SQL] Add back a file status cache for catalog tables

ericl:

In SPARK-16980, we removed the full in-memory cache of table partitions in favor of loading only needed partitions from the metastore. This greatly improves the initial latency of queries that only read a small fraction of table partitions.

However, since the metastore does not store file statistics, we need to discover those from remote storage. With the loss of the in-memory file status cache this has to happen on each query, increasing the latency of repeated queries over the same partitions.

The proposal is to add back a per-table cache of partition contents, i.e. Map[Path, Array[FileStatus]]. This cache would be retained per-table, and can be invalidated through refreshTable() and refreshByPath(). Unlike the prior cache, it can be incrementally updated as new partitions are read.
## How was this patch tested?

Existing tests and new tests in `HiveTablePerfStatsSuite`.

cc @mallman


core/src/main/scala/org/apache/spark/util/Utils.scala
```diff
@@ -1486,10 +1485,10 @@ private[spark] object Utils extends Logging {
       val gzInputStream = new GZIPInputStream(new FileInputStream(file))
       val bufSize = 1024
```
ericl: +1

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/FileStatusCache.scala
```diff
@@ -0,0 +1,78 @@
...
+
+  override def invalidateAll(): Unit = {
+    cache.clear()
+  }
+}
+
+/**
+ * A non-caching implementation used when partition file status caching is disabled.
+ */
+class NoopCache extends FileStatusCache {
```
mallman: It looks like this can be a singleton object instead of a class.

ericl: good idea

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/PartitioningAwareFileCatalog.scala
```diff
@@ -276,15 +290,15 @@ object PartitioningAwareFileCatalog extends Logging {
...
     val jobConf = new JobConf(hadoopConf, this.getClass)
     val filter = FileInputFormat.getInputPathFilter(jobConf)
 
-    paths.flatMap { path =>
+    paths.map { path =>
       val fs = path.getFileSystem(hadoopConf)
-      listLeafFiles0(fs, path, filter)
-    }
+      (path, listLeafFiles0(fs, path, filter))
+    }.toMap
```
mallman: Based on how you're using this return value above, it looks like you can omit this call to `toMap` and just return a sequence of tuples.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/ListingFileCatalog.scala
```diff
@@ -38,14 +38,16 @@ class ListingFileCatalog(
     sparkSession: SparkSession,
     override val rootPaths: Seq[Path],
     parameters: Map[String, String],
-    partitionSchema: Option[StructType])
-  extends PartitioningAwareFileCatalog(sparkSession, parameters, partitionSchema) {
+    partitionSchema: Option[StructType],
+    fileStatusCache: FileStatusCache = new NoopCache)
```
mallman: I still think you should just use a singleton object, but if you want to use `FileStatusCache.noop`, you forgot to update this code.

mallman: It just occurred to me I may not have been clear in my suggestion. Just in case, I'm suggesting changing `class NoopCache` to `object NoopCache`.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/FileStatusCache.scala
```diff
@@ -0,0 +1,82 @@
...
+
+/**
+ * A cache of the leaf files of partition directories. We cache these files in order to speed
+ * up iterated queries over the same set of partitions. Otherwise, each query would have to
+ * hit remote storage in order to gather file statistics for physical planning.
+ *
+ * Each resolved catalog table has its own FileStatusCache. When the backing relation for the
+ * table is refreshed via refreshTable() or refreshByPath(), this cache will be invalidated.
+ */
+abstract class FileStatusCache {
```
cloud-fan: should we have a limit for max cache entries?

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/TableFileCatalog.scala
```diff
@@ -32,13 +32,21 @@ import org.apache.spark.sql.types.StructType
...
+ * @param enableFileStatusCache whether to enable file status caching
  */
 class TableFileCatalog(
     sparkSession: SparkSession,
     db: String,
     table: String,
     partitionSchema: Option[StructType],
-    override val sizeInBytes: Long) extends FileCatalog {
+    override val sizeInBytes: Long,
+    enableFileStatusCache: Boolean) extends FileCatalog {
```
cloud-fan: when will users want to turn it off?

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/FileStatusCache.scala
mallman: I have similar concerns. I think that the proposed, simple solution may suffice for a single-user or short-lived spark session. On the other hand, an unbounded cache here will probable lead to trouble for some users in a multi-user or long-running Spark app such as the thriftserver.

Right now I'm thinking the `InMemoryCache` is okay as-is because this feature is behind a configuration flag, but I think we'll need a resource-sensitive solution in order to support this cache in the thriftserver.

mallman:
@ericl I took this PR for a test drive with some large-ish tables. Everything appeared to work as expected.

As far as performance goes, planning a simple select on a partitioned table (with cached partition metadata) took less than 1 second up to about 3k partitions. It was about 1.5 seconds with 5k partitions, 2 seconds with about 7.5k partitions and about 3.2 seconds with 10k partitions. Does that align with your findings?

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/FileStatusCache.scala
ericl: There's some extra complexity here since each table has its own cache, so you could still exceed the amount of available memory if there were multiple tables.

Some options here:
1) A large shared cache that includes table name in the key
2) Pick an arbitrary size limit for each table cache
3) Do nothing

ericl:
@mallman those numbers seem about right. I think as long as planning time is not that much worse than with the old ListingFileCatalog we are good.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/FileStatusCache.scala
mallman: A single shared cache with some kind of upper limit on entries (or memory, the configuration of which is worth another discussion) and automatic eviction would simplify global resource management. I think we will need something like this to support the thriftserver.

ericl: It turns out a global cache is fairly complicated if you want to correctly garbage-collect expired entries. I opted for a simpler per-table max cache size, defaulting to 50MB per table. This should fit a few tens of thousands of partitions (1 filestatus object is ~1k in size).

mallman: Sounds like a plan.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/FileStatusCache.scala
```diff
@@ -0,0 +1,98 @@
...
+  def putLeafFiles(path: Path, leafFiles: Array[FileStatus]): Unit
+
+  /**
+   * Invalidates all data held by this cache.
+   */
+  def invalidateAll(): Unit
+}
+
+/**
+ * An implementation that caches all partition file statuses in memory forever.
```
mallman: This comment is out of date.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/ListingFileCatalog.scala
```diff
@@ -64,11 +66,18 @@ class ListingFileCatalog(
   }
 
   override def refresh(): Unit = {
+    refresh0(true)
+  }
+
+  private def refresh0(invalidateSharedCache: Boolean): Unit = {
```
cloud-fan: the cache is shared?

sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveMetastoreCatalog.scala
```diff
@@ -231,11 +231,16 @@ private[hive] class HiveMetastoreCatalog(sparkSession: SparkSession) extends Log
         val sizeInBytes = metastoreRelation.statistics.sizeInBytes.toLong
         val fileCatalog = {
           val catalog = new TableFileCatalog(
-            sparkSession, db, table, Some(partitionSchema), sizeInBytes)
+            sparkSession, db, table, Some(partitionSchema), sizeInBytes,
+            fileStatusCacheSize = if (lazyPruningEnabled) {
```
cloud-fan: shall we inline this logic in `TableFileCatalog`? `TableFileCatalog` can access `SparkSession` so it can know these 2 flags: `filesourcePartitionPruning` and `filesourcePartitionFileCacheSize`

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/TableFileCatalog.scala
```diff
@@ -87,21 +85,13 @@ class TableFileCatalog(
...
-          sparkSession, new Path(baseLocation.get), partitionSpec)
+          sparkSession, new Path(baseLocation.get), fileStatusCache, partitionSpec)
       case None =>
-        new ListingFileCatalog(sparkSession, rootPaths, parameters, None)
-    }
-  }
-
-  // Not used in the hot path of queries when metastore partition pruning is enabled
-  def allPartitions: ListingFileCatalog = synchronized {
-    if (cachedAllPartitions == null) {
```
cloud-fan: `cachedAllPartitions` is useless now.

sql/hive/src/test/scala/org/apache/spark/sql/hive/HiveTablePerfStatsSuite.scala
```diff
@@ -103,11 +92,103 @@ class HiveDataFrameSuite extends QueryTest with TestHiveSingleton with SQLTestUt
...
+          assert(HiveCatalogMetrics.METRIC_FILE_CACHE_HITS.getCount() == 0)
+        }
+      }
+    }
+  }
+
+  test("file status cache respects size limit") {
+    withSQLConf(
+        "spark.sql.hive.filesourcePartitionPruning" -> "true",
+        "spark.sql.hive.filesourcePartitionFileCacheSize" -> "1" /* 1 byte */) {
```
cloud-fan: should we test the evicting behaviour?

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/FileStatusCache.scala
rxin: hm this is pretty risky, because the number of tables in a real environment can be very large.

mallman: That's a good point.

@ericl, do you want to have a go at a global cache and push something for review?

ericl: This is already better than what we have in 2.0 - which is caching everything with no limit.

I think a global cache is really fairly complicated since we can no longer rely on the Java GC to manage the lifetime of the cache entries. Unless we want to do cache GC in a finalizer, I don't think this is worth it.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/ListingFileCatalog.scala
ericl: It's shared within a single table. Confusingly, you can have views over the table that are cached in CacheManager, so we need this to invalidate the table cache as well on refresh.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/TableFileCatalog.scala
ericl: Removed

sql/hive/src/test/scala/org/apache/spark/sql/hive/HiveTablePerfStatsSuite.scala
ericl: Yeah this test covers that

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/FileStatusCache.scala
mallman: What about using the cache's built-in eviction process? That would enforce an upper limit on memory usage.

ericl: That's a good idea. I've pushed a change that implements this with a global cache size limit of 250MB.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/PartitioningAwareFileCatalog.scala
```diff
@@ -294,7 +308,7 @@ object PartitioningAwareFileCatalog extends Logging {
   private def listLeafFilesInParallel(
       paths: Seq[Path],
       hadoopConf: Configuration,
-      sparkSession: SparkSession): Seq[FileStatus] = {
+      sparkSession: SparkSession): Map[Path, Seq[FileStatus]] = {
```
mallman: I believe this should return the same type as `listLeafFilesInSerial`, i.e. `Seq[(Path, Seq[FileStatus])]`.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/TableFileCatalog.scala
```diff
@@ -42,24 +43,21 @@ class TableFileCatalog(
 
   protected val hadoopConf = sparkSession.sessionState.newHadoopConf
 
+  private val fileStatusCache = FileStatusCache.getOrInitializeShared(new Object(), sparkSession)
```
mallman: Does the file status cache need to be private to an instance of `TableFileCatalog`? Can we use the table's qualified name as the key here instead?

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/ListingFileCatalog.scala
```diff
@@ -64,11 +66,18 @@ class ListingFileCatalog(
   }
 
   override def refresh(): Unit = {
+    refresh0(true)
```
cloud-fan: nit: instead of having a parameter in `refresh0`, can we just call `refresh0` here, and then invalidate the file status cache?

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/FileStatusCache.scala
```diff
@@ -0,0 +1,149 @@
...
+
+  /**
+   * Invalidates all data held by this cache.
+   */
+  def invalidateAll(): Unit
+}
+
+object FileStatusCache {
+  // Opaque object that uniquely identifies a shared cache user
+  type ClientId = Object
```
cloud-fan: Why do we need a client id? It looks to me that if a client refers to a same path of another client, it won't hit the cache. Is it an expected behaviour?

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/PartitioningAwareFileCatalog.scala
ericl: Updated

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/TableFileCatalog.scala
ericl: The way we currently refresh tables is by dropping the reference to TableFileCatalog and letting it get GC'ed. Given this strategy, making it private is the simplest way to ensure refresh actually works correctly -- otherwise you have to carefully test that refresh also invalidates shared cache entries.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/ListingFileCatalog.scala
ericl: Good idea

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/FileStatusCache.scala
ericl: I think it's better to err on the side of isolation here. Otherwise, it is harder to reason about what is actually invalidated when a table is refreshed.

cloud-fan:
We have REFRESH TABLE/PATH because we cache things, so I think we should consider caching and refreshing together. Currently we have 4 caches:

1. **table name to `LogicalRelation` cache**: It's kept in `HiveMetastoreCatalog`, we won't touch it in this PR.
2. **table name to partition directories**: It's kept in `ListingFileCatalog` and is data source table only, as hive table persists this information in metastore. #15515 is going to follow hive and remove this cache for data source table.
3. **partition directory to file status**: It's also kept in `ListingFileCatalog` and is data source table only. This PR is going to add it to data source tables that are converted from hive tables, and make it global.
4. **table relation to `InMemoryRelation`**: It's kept in `CacheManager`, and will only be cached when users call `CACHE TABLE`. We won't touch it in this PR.

The current refresh logic:

1. **REFRESH TABLE**: it invalidates all 4 caches, by removing the first cache, as `FileCatalog` is a member of `HadoopFsRelation` and `HadoopFsRelation` is a member of `LogicalRelation`. Then it asks `CacheManage` to invalidate the `InMemoryRelation` cache if it's cached.
2. **REFRESH PATH**: it finds a table that uses the given path as table path or partition paths, and then refresh this table. Actually I don't think this is the right behaviour, REFRESH PATH should be more fine-grained, and only remove this path from the third cache.

However, if we make the third cache global, we need a little more work to make REFRESH TABLE work, because just removing the first cache is not enough. We need a way to know which partition directories are cached for the given table.

Proposal 1(the current implementation of this PR): In the global "partition directory to file status" cache, we use `clientId -> partition dir` as key instead of `partition dir`. When we refresh a table, we go through the whole map and find cache entries match the `clientId` and remove them. Note that, every `TableFileCatalog` has a unique `clientId`.

The biggest problem of this proposal is, invalidating the cache may be slow if there are a lot of cache entries.

Proposal 2: we just use `partition dir` as the cache entry key in the global cache, and keep the list of cached partition dirs in `TableFileCatalog`. When refreshing a table, we get the list of cached partition dir first, and then invalidete them one by one in the "partition directory to file status" cache.

The biggest problem of this proposal is, if 2 tables share some paths(it's possible if we use same paths to add parittions to 2 tables), refreshing one table will partially refresh the other table. What is worse, if the other table is explicitly cached via CACHE TABLE, we won't invalidate the `InMemoryRelation` cache and brings inconsistency.

Proposal 3: do not make the third cache in this PR, and think about it later. At least it's not worse than spark 2.0

Any ideas?

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/TableFileCatalog.scala
mallman: Odd, I figured we'd call the `refresh` method to refresh the table.

The reason I suggested using the table's name is so that all references to the same table can use the same cache—avoid redundant cache entries and provide the ability to inherit cached entries for new instances of the same logical table. Aside from implementation issues, does that make sense?

Anyway, this is something we can explore in a follow up PR if desired.

ericl:
> The biggest problem of this proposal is, invalidating the cache may be slow if there are a lot of cache entries.

I don't think this is really an issue. Conservatively assuming ~1us per eviction and one file per partition, we can evict all 250MB of file entries in ~160ms.

I also cleaned up the code a little to better hide the ids from cache users.

sql/hive/src/test/scala/org/apache/spark/sql/hive/HiveTablePerfStatsSuite.scala
```diff
@@ -104,11 +107,103 @@ class HiveDataFrameSuite extends QueryTest with TestHiveSingleton with SQLTestUt
...
+          // cache should be disabled
+          assert(HiveCatalogMetrics.METRIC_FILE_CACHE_HITS.getCount() == 0)
+        }
+      }
+    }
+  }
+
+  test("lazy partition pruning with file status caching enabled") {
+    withSQLConf(
+        "spark.sql.hive.filesourcePartitionPruning" -> "true",
```
cloud-fan: nit: please use SQLConf.HIVE_FILESOURCE_PARTITION_PRUNING.key

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/ListingFileCatalog.scala
```diff
@@ -38,14 +38,16 @@ class ListingFileCatalog(
...
     parameters: Map[String, String],
-    partitionSchema: Option[StructType])
-  extends PartitioningAwareFileCatalog(sparkSession, parameters, partitionSchema) {
+    partitionSchema: Option[StructType],
+    fileStatusCache: FileStatusCache = NoopCache)
+  extends PartitioningAwareFileCatalog(
+    sparkSession, parameters, partitionSchema, fileStatusCache) {
 
   @volatile private var cachedLeafFiles: mutable.LinkedHashMap[Path, FileStatus] = _
   @volatile private var cachedLeafDirToChildrenFiles: Map[Path, Array[FileStatus]] = _
```
cloud-fan: is this cache still useful?

cloud-fan:
LGTM, next maybe we can refactor the `PartitionAwareFileCatalog` and make it use the new global cache better. I'm going to merge it to unblock other works. thanks!

cloud-fan:
I had some network problems, I'll ask @yhuai to merge it.

cloud-fan:
merging to master!

mallman:
@ericl Great work on this. I don't know how I got an author credit in the commit...

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/FileStatusCache.scala
```diff
@@ -0,0 +1,149 @@
...
+  private var sharedCache: SharedInMemoryCache = null
+
+  /**
+   * @return a new FileStatusCache based on session configuration. Cache memory quota is
+   *         shared across all clients.
+   */
+  def newCache(session: SparkSession): FileStatusCache = {
+    synchronized {
+      if (session.sqlContext.conf.filesourcePartitionPruning &&
+          session.sqlContext.conf.filesourcePartitionFileCacheSize > 0) {
```
gatorsmile: These two conf have to be global and static; otherwise, we might face weird behaviors when different sessions are using/changing different values. Let me submit a PR to improve it.

ericl: Sounds good, as long as you can still change it at runtime.

gatorsmile: If we allow users to change it at runtime, users might get strange results due to the global sharing cache. Some sessions might skip cache to directly change the metastore, some sessions are still using the cached values. Even in a single session, we might still face similar issues if we allow users to change it at runtime.

gatorsmile: Or can we make the cache session-specific?

ericl: Oh I see, the sizing can be problematic. I find it sometimes useful for debugging to turn it off at runtime by setting it to zero though -- it will take effect after a refresh table.

Perhaps we can just add a comment that adjusting the value at runtime may produce unexpected behavior when there are multiple sessions?

gatorsmile: How about `HIVE_MANAGE_FILESOURCE_PARTITIONS`?

ericl:
That one is safe to make global but mutable right? It will take effect
after a table is refreshed.

Most of these anomalies seem OK to me provided we document them -- it seems
to solve them we would have to cache per session which would significantly
increase memory required.

gatorsmile:
`(ClientId, Path), Array[FileStatus]` uh... `FileStatusCache` does not share any entries with any other client, but does share memory resources for the purpose of cache eviction.

Sorry, I misunderstood the code. I think we need to evict all the cache that are associated with the specific client when users make a switch. Do you think this makes sense?

gatorsmile:
Yeah, I think we should document the behavior issues when different sessions are using different conf values. Will do it. I think we also need to evict all the cache that are associated with the specific client when users make a switch. Do you think this makes sense?

ericl:
Hm, what use cases are we trying to address? As I understand, the worst that can happen if the cache size flag is toggled at runtime is that the old settings might still apply. And when the filesource partition management flag is changed, it only takes effect after the table entry in `cachedDataSourceTables` in HiveMetastoreCatalog is refreshed.

If we document the above, is that enough?

gatorsmile:
It could return incorrect results, but I need to prove it using a use case. We always call [refresh0](https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/InMemoryFileIndex.scala#L50) in [`filterPartitions`](https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/CatalogFileIndex.scala#L78-L79). Thus, it covers most cases. Let me try more in the next few days. 

Tonight, let me first fix the cache inconsistence issue caused by `SET LOCATION`.

ericl:
Hmm, I don't think fileStatusCache can ever return incorrect results, only stale ones. Furthermore, its scoped by client-id to particular instances of tables, so refresh table is guaranteed to wipe out cache state.

gatorsmile:
Just submitted a PR for `set location`: https://github.com/apache/spark/pull/16514 That issue is caused by the cache for mapping the table name to LogicalRelation. We need to refresh it after changing the table location.

PR #15540, opened 2016-10-19
[MINOR][Documentation] Mention that short names can be used in data sources not only for built-in sources

HyukjinKwon:

This PR proposes to fix the documentation in order to mention the short names can be used for other external data sources as well (not only built-in ones).
## How was this patch tested?

Manually built the docs.

**Before**

**After**

### DIFF
--- a/docs/sql-programming-guide.md
+++ b/docs/sql-programming-guide.md
@@ -421,9 +421,10 @@ In the simplest form, the default data source (`parquet` unless otherwise config
 
 You can also manually specify the data source that will be used along with any extra options
 that you would like to pass to the data source. Data sources are specified by their fully qualified
-name (i.e., `org.apache.spark.sql.parquet`), but for built-in sources you can also use their short
-names (`json`, `parquet`, `jdbc`, `orc`, `libsvm`, `csv`, `text`). DataFrames loaded from any data
-source type can be converted into other types using this syntax.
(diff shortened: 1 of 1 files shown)


HyukjinKwon:
Please let me cc @weiqingy and @rxin as we talk about this.

docs/sql-programming-guide.md
```diff
@@ -421,9 +421,10 @@ In the simplest form, the default data source (`parquet` unless otherwise config
 
 You can also manually specify the data source that will be used along with any extra options
 that you would like to pass to the data source. Data sources are specified by their fully qualified
-name (i.e., `org.apache.spark.sql.parquet`), but for built-in sources you can also use their short
-names (`json`, `parquet`, `jdbc`, `orc`, `libsvm`, `csv`, `text`). DataFrames loaded from any data
-source type can be converted into other types using this syntax.
+name (i.e., `org.apache.spark.sql.parquet`) or short name specified in
+`DataSourceRegister.shortName()` in their implementation. For built-in sources you can use their
```
rxin: hm this isn't clear to the end user at all, since there is no class called DataSourceRegister even in the source code for data sources. I'd just leave it out. It should be the job of the data source to tell users how to use it.

HyukjinKwon: Makes sense. Thank you for your quick response.

PR #15218, opened 2016-09-23
[SPARK-17637][Scheduler]Packed scheduling for Spark tasks across executors

zhzhan:
## What changes were proposed in this pull request?

Restructure the code and implement two new task assigner.
PackedAssigner: try to allocate tasks to the executors with least available cores, so that spark can release reserved executors when dynamic allocation is enabled.

BalancedAssigner: try to allocate tasks to the executors with more available cores in order to balance the workload across all executors.

By default, the original round robin assigner is used.

We test a pipeline, and new PackedAssigner  save around 45% regarding the reserved cpu and memory with dynamic allocation enabled.
## How was this patch tested?

(Please explain how this patch was tested. E.g. unit tests, integration tests, manual tests)
Both unit test in TaskSchedulerImplSuite and manual tests in production pipeline.

PR #15541, opened 2016-10-19
[SPARK-17637][Scheduler]Packed scheduling for Spark tasks across executors

zhzhan:

Restructure the code and implement two new task assigner.
PackedAssigner: try to allocate tasks to the executors with least available cores, so that spark can release reserved executors when dynamic allocation is enabled.

BalancedAssigner: try to allocate tasks to the executors with more available cores in order to balance the workload across all executors.

By default, the original round robin assigner is used.

We test a pipeline, and new PackedAssigner save around 45% regarding the reserved cpu and memory with dynamic allocation enabled.
## How was this patch tested?

Both unit test in TaskSchedulerImplSuite and manual tests in production pipeline.

### DIFF
--- a/core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
+++ b/core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
@@ -133,8 +133,8 @@ object TaskAssigner extends Logging {
     val className = {
       val name = assignerMap.get(assignerName.toLowerCase())
       name.getOrElse {
-        throw new SparkException(s"Task Assigner $assignerName is not available. " +
-          s"Please choose roundrobin, packed, or balanced.  roundrobin is used by default")
+        throw new SparkException(s"Task Assigner $assignerName is invalid. Available assigners " +
+          s"are roundrobin, packed, and balanced.  roundrobin is the default")
 
       }
     }
--- a/core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
+++ b/core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
@@ -26,7 +26,7 @@ import org.apache.spark.SparkConf
 import org.apache.spark.util.Utils
 
 /** Tracks the current state of the workers with available cores and assigned task list. */
-class OfferState(val workOffer: WorkerOffer) {
+private[scheduler] class OfferState(val workOffer: WorkerOffer) {
   /** The current remaining cores that can be allocated to tasks. */
   var coresAvailable: Int = workOffer.cores
   /** The list of tasks that are assigned to this WorkerOffer. */
@@ -38,7 +38,7 @@ class OfferState(val workOffer: WorkerOffer) {
  * extended to implement different task scheduling algorithms.
  * Together with [[org.apache.spark.scheduler.TaskScheduler TaskScheduler]], TaskAssigner
  * is used to assign tasks to workers with available cores. Internally, when TaskScheduler
- * perform task assignment given available workers, it first sorts the candidate tasksets,
+ * performs task assignment given available workers, it first sorts the candidate tasksets,
  * and then for each taskset, it takes a number of rounds to request TaskAssigner for task
  * assignment with different locality restrictions until there is either no qualified
  * workers or no valid tasks to be assigned.
@@ -50,7 +50,7 @@ class OfferState(val workOffer: WorkerOffer) {
  * First, TaskScheduler invokes construct() of TaskAssigner to initialize the its internal
  * worker states at the beginning of resource offering.
  *
- * Second, before each round of task assignment for a taskset, TaskScheduler invoke the init()
+ * Second, before each round of task assignment for a taskset, TaskScheduler invokes the init()
  * of TaskAssigner to initialize the data structure for the round.
  *
  * Third, when performing real task assignment, hasNext()/getNext() is used by TaskScheduler
@@ -83,9 +83,9 @@ private[scheduler] abstract class TaskAssigner {
   def init(): Unit
 
   /**
-   * Tests Whether there is offer available to be used inside of one round of Taskset assignment.
-   *  @return  `true` if a subsequent call to `next` will yield an element,
-   *           `false` otherwise.
+   * Tests whether there is offer available to be used inside of one round of Taskset assignment.
+   * @return  `true` if a subsequent call to `next` will yield an element,
+   *          `false` otherwise.
    */
   def hasNext: Boolean
 
@@ -152,7 +152,7 @@ class RoundRobinAssigner extends TaskAssigner {
 }
 
 /**
- * Assign the task to workers with the most available cores. It other words, BalancedAssigner tries
+ * Assign the task to workers with the most available cores. In other words, BalancedAssigner tries
  * to distribute the task across workers in a balanced way. Potentially, it may alleviate the
  * workers' memory pressure as less tasks running on the same workers, which also indicates that
  * the task itself can make use of more computation resources, e.g., hyper-thread, across clusters.
--- a/core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
+++ b/core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
@@ -76,7 +76,7 @@ private[scheduler] abstract class TaskAssigner {
 
   /** Invoked at the beginning of resource offering to construct the offer with the workoffers. */
   def construct(workOffer: Seq[WorkerOffer]): Unit = {
-    offer = workOffer.map(o => new OfferState(o))
+    offer = Random.shuffle(workOffer.map(o => new OfferState(o)))
   }
 
   /** Invoked at each round of Taskset assignment to initialize the internal structure. */
@@ -136,10 +136,6 @@ object TaskAssigner extends Logging {
 class RoundRobinAssigner extends TaskAssigner {
   private var currentOfferIndex = 0
 
-  override def construct(workOffer: Seq[WorkerOffer]): Unit = {
-    offer = Random.shuffle(workOffer.map(o => new OfferState(o)))
-  }
-
   override def init(): Unit = {
     currentOfferIndex = 0
   }
@@ -169,10 +165,6 @@ class BalancedAssigner extends TaskAssigner {
   }
   private val maxHeap: PriorityQueue[OfferState] = new PriorityQueue[OfferState]()
   private var currentOffer: OfferState = _
-
-  override def construct(workOffer: Seq[WorkerOffer]): Unit = {
-    offer = Random.shuffle(workOffer.map(o => new OfferState(o)))
-  }
 
   override def init(): Unit = {
     maxHeap.clear()
--- a/core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
+++ b/core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
@@ -25,11 +25,11 @@ import org.apache.spark.internal.{config, Logging}
 import org.apache.spark.SparkConf
 import org.apache.spark.util.Utils
 
-/** Tracking the current state of the workers with available cores and assigned task list. */
+/** Tracks the current state of the workers with available cores and assigned task list. */
 class OfferState(val workOffer: WorkerOffer) {
-  // The current remaining cores that can be allocated to tasks.
+  /** The current remaining cores that can be allocated to tasks. */
   var coresAvailable: Int = workOffer.cores
-  // The list of tasks that are assigned to this worker.
+  /** The list of tasks that are assigned to this WorkerOffer. */
   val tasks = new ArrayBuffer[TaskDescription](coresAvailable)
 }
 
@@ -37,60 +37,70 @@ class OfferState(val workOffer: WorkerOffer) {
  * TaskAssigner is the base class for all task assigner implementations, and can be
  * extended to implement different task scheduling algorithms.
  * Together with [[org.apache.spark.scheduler.TaskScheduler TaskScheduler]], TaskAssigner
- * is used to assign tasks to workers with available cores. Internally, TaskScheduler, requested
- * to perform task assignment given available workers, first sorts the candidate tasksets,
+ * is used to assign tasks to workers with available cores. Internally, when TaskScheduler
+ * perform task assignment given available workers, it first sorts the candidate tasksets,
  * and then for each taskset, it takes a number of rounds to request TaskAssigner for task
- * assignment with different the locality restrictions until there is either no qualified
+ * assignment with different locality restrictions until there is either no qualified
  * workers or no valid tasks to be assigned.
  *
  * TaskAssigner is responsible to maintain the worker availability state and task assignment
  * information. The contract between [[org.apache.spark.scheduler.TaskScheduler TaskScheduler]]
- * and TaskAssigner is as follows. First, TaskScheduler invokes construct() of TaskAssigner to
- * initialize the its internal worker states at the beginning of resource offering. Before each
- * round of task assignment for a taskset, TaskScheduler invoke the init() of TaskAssigner to
- * initialize the data structure for the round. When performing real task assignment,
- * hasNext()/getNext() is used by TaskScheduler to check the worker availability and retrieve
- * current offering from TaskAssigner. Then offerAccepted is used by TaskScheduler to notify
- * the TaskAssigner so that TaskAssigner can decide whether the current offer is valid or not for
- * the next request. After task assignment is done, TaskScheduler invokes the tasks() to
- * retrieve all the task assignment information, and eventually, invokes reset() method so that
- * TaskAssigner can cleanup its internal maintained resources.
+ * and TaskAssigner is as follows.
+ *
+ * First, TaskScheduler invokes construct() of TaskAssigner to initialize the its internal
+ * worker states at the beginning of resource offering.
+ *
+ * Second, before each round of task assignment for a taskset, TaskScheduler invoke the init()
+ * of TaskAssigner to initialize the data structure for the round.
+ *
+ * Third, when performing real task assignment, hasNext()/getNext() is used by TaskScheduler
+ * to check the worker availability and retrieve current offering from TaskAssigner.
+ *
+ * Fourth, then offerAccepted is used by TaskScheduler to notify the TaskAssigner so that
+ * TaskAssigner can decide whether the current offer is valid or not for the next request.
+ *
+ * Fifth, After task assignment is done, TaskScheduler invokes the tasks() to
+ * retrieve all the task assignment information.
  */
 
 private[scheduler] abstract class TaskAssigner {
-  var offer: Seq[OfferState] = _
-  var CPUS_PER_TASK = 1
+  protected var offer: Seq[OfferState] = _
+  protected var cpuPerTask = 1
 
-  def withCpuPerTask(CPUS_PER_TASK: Int): Unit = {
-    this.CPUS_PER_TASK = CPUS_PER_TASK
+  protected def withCpuPerTask(cpuPerTask: Int): Unit = {
+    this.cpuPerTask = cpuPerTask
   }
 
-  // The final assigned offer returned to TaskScheduler.
+  /** The final assigned offer returned to TaskScheduler. */
   final def tasks: Seq[ArrayBuffer[TaskDescription]] = offer.map(_.tasks)
 
-  // Invoked at the beginning of resource offering to construct the offer with the workoffers.
+  /** Invoked at the beginning of resource offering to construct the offer with the workoffers. */
   def construct(workOffer: Seq[WorkerOffer]): Unit = {
     offer = workOffer.map(o => new OfferState(o))
   }
 
-  // Invoked at each round of Taskset assignment to initialize the internal structure.
+  /** Invoked at each round of Taskset assignment to initialize the internal structure. */
   def init(): Unit
 
-  // Whether there is offer available to be used inside of one round of Taskset assignment.
+  /**
+   * Tests Whether there is offer available to be used inside of one round of Taskset assignment.
+   *  @return  `true` if a subsequent call to `next` will yield an element,
+   *           `false` otherwise.
+   */
   def hasNext: Boolean
 
-  // Returned the next assigned offer based on the task assignment strategy.
-  def getNext(): OfferState
+  /**
+   * Produces next worker offer based on the task assignment strategy.
+   * @return  the next available offer, if `hasNext` is `true`,
+   *          undefined behavior otherwise.
+   */
+  def next(): OfferState
 
-  // Invoked by the TaskScheduler to indicate whether the current offer is accepted or not so that
-  // the assigner can decide whether the current worker is valid for the next offering.
+  /**
+   * Invoked by the TaskScheduler to indicate whether the current offer is accepted or not so that
+   * the assigner can decide whether the current worker is valid for the next offering.
+   */
   def offerAccepted(assigned: Boolean): Unit
-
-  // Invoked at the end of resource offering to release internally maintained resources.
-  // Subclass is responsible to release its own private resources.
-  def reset(): Unit = {
-    offer = null
-  }
 }
 
 object TaskAssigner extends Logging {
@@ -104,19 +114,18 @@ object TaskAssigner extends Logging {
 
   def init(conf: SparkConf): TaskAssigner = {
     val assignerName = conf.get(config.SPARK_SCHEDULER_TASK_ASSIGNER.key, "roundrobin")
-      .toLowerCase()
-    val className = assignerMap.getOrElse(assignerName, roundrobin)
-    val CPUS_PER_TASK = conf.getInt("spark.task.cpus", 1)
-    val assigner = try {
-      logInfo(s"Constructing an assigner as $className")
-      Utils.classForName(className).getConstructor()
-        .newInstance().asInstanceOf[TaskAssigner]
-    } catch {
-      case _: Throwable =>
-        logInfo(s"$assignerName cannot be constructed, fallback to default $roundrobin.")
-        new RoundRobinAssigner()
+    val className = {
+      val name = assignerMap.get(assignerName.toLowerCase())
+      name.getOrElse {
+        logWarning(s"$assignerName cannot be constructed, fallback to default $roundrobin.")
+        roundrobin
+      }
     }
-    assigner.withCpuPerTask(CPUS_PER_TASK)
+    // The className is valid. No need to catch exceptions.
+    logInfo(s"Constructing TaskAssigner as $className")
+    val assigner = Utils.classForName(className)
+      .getConstructor().newInstance().asInstanceOf[TaskAssigner]
+    assigner.withCpuPerTask(cpuPerTask = conf.getInt("spark.task.cpus", 1))
     assigner
   }
 }
@@ -125,71 +134,62 @@ object TaskAssigner extends Logging {
  * Assign the task to workers with available cores in roundrobin manner.
  */
 class RoundRobinAssigner extends TaskAssigner {
-  private var idx = 0
+  private var currentOfferIndex = 0
 
   override def construct(workOffer: Seq[WorkerOffer]): Unit = {
     offer = Random.shuffle(workOffer.map(o => new OfferState(o)))
   }
 
   override def init(): Unit = {
-    idx = 0
+    currentOfferIndex = 0
   }
 
-  override def hasNext: Boolean = idx < offer.size
+  override def hasNext: Boolean = currentOfferIndex < offer.size
 
-  override def getNext(): OfferState = {
-    offer(idx)
+  override def next(): OfferState = {
+    offer(currentOfferIndex)
   }
 
   override def offerAccepted(assigned: Boolean): Unit = {
-    idx += 1
-  }
-
-  override def reset(): Unit = {
-    super.reset
-    idx = 0
+    currentOfferIndex += 1
   }
 }
 
 /**
- * Assign the task to workers with the most available cores.
+ * Assign the task to workers with the most available cores. It other words, BalancedAssigner tries
+ * to distribute the task across workers in a balanced way. Potentially, it may alleviate the
+ * workers' memory pressure as less tasks running on the same workers, which also indicates that
+ * the task itself can make use of more computation resources, e.g., hyper-thread, across clusters.
  */
 class BalancedAssigner extends TaskAssigner {
-  private var maxHeap: PriorityQueue[OfferState] = _
+  implicit val ord: Ordering[OfferState] = new Ordering[OfferState] {
+    def compare(x: OfferState, y: OfferState): Int = {
+      return Ordering[Int].compare(x.coresAvailable, y.coresAvailable)
+    }
+  }
+  private val maxHeap: PriorityQueue[OfferState] = new PriorityQueue[OfferState]()
   private var currentOffer: OfferState = _
 
   override def construct(workOffer: Seq[WorkerOffer]): Unit = {
     offer = Random.shuffle(workOffer.map(o => new OfferState(o)))
   }
 
-  implicit val ord: Ordering[OfferState] = new Ordering[OfferState] {
-    def compare(x: OfferState, y: OfferState): Int = {
-      return Ordering[Int].compare(x.coresAvailable, y.coresAvailable)
-    }
-  }
-
   override def init(): Unit = {
-    maxHeap = new PriorityQueue[OfferState]()
-    offer.filter(_.coresAvailable >= CPUS_PER_TASK).foreach(maxHeap.enqueue(_))
+    maxHeap.clear()
+    offer.filter(_.coresAvailable >= cpuPerTask).foreach(maxHeap.enqueue(_))
   }
 
   override def hasNext: Boolean = maxHeap.nonEmpty
 
-  override def getNext(): OfferState = {
+  override def next(): OfferState = {
     currentOffer = maxHeap.dequeue()
     currentOffer
   }
 
   override def offerAccepted(assigned: Boolean): Unit = {
-    if (currentOffer.coresAvailable >= CPUS_PER_TASK && assigned) {
+    if (currentOffer.coresAvailable >= cpuPerTask && assigned) {
       maxHeap.enqueue(currentOffer)
     }
-  }
-
-  override def reset(): Unit = {
-    super.reset
-    maxHeap = null
-    currentOffer = null
   }
 }
 
@@ -199,35 +199,28 @@ class BalancedAssigner extends TaskAssigner {
  * assigned if more than required workers are reserved. If the dynamic allocator is enabled,
  * these idle workers will be released by driver. The released resources can then be allocated to
  * other jobs by underling resource manager. This assigner can potentially reduce the resource
- * reservation for a job.
+ * reservation for jobs, which over allocate resources than they need.
  */
 class PackedAssigner extends TaskAssigner {
   private var sorted: Seq[OfferState] = _
-  private var idx = 0
+  private var currentOfferIndex = 0
   private var currentOffer: OfferState = _
 
   override def init(): Unit = {
-    idx = 0
-    sorted = offer.filter(_.coresAvailable >= CPUS_PER_TASK).sortBy(_.coresAvailable)
+    currentOfferIndex = 0
+    sorted = offer.filter(_.coresAvailable >= cpuPerTask).sortBy(_.coresAvailable)
   }
 
-  override def hasNext: Boolean = idx < sorted.size
+  override def hasNext: Boolean = currentOfferIndex < sorted.size
 
-  override def getNext(): OfferState = {
-    currentOffer = sorted(idx)
+  override def next(): OfferState = {
+    currentOffer = sorted(currentOfferIndex)
     currentOffer
   }
 
   override def offerAccepted(assigned: Boolean): Unit = {
-    if (currentOffer.coresAvailable < CPUS_PER_TASK || !assigned) {
-      idx += 1
+    if (currentOffer.coresAvailable < cpuPerTask || !assigned) {
+      currentOfferIndex += 1
     }
-  }
-
-  override def reset(): Unit = {
-    super.reset
-    sorted = null
-    currentOffer = null
-    idx = 0
   }
 }
--- a/core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
+++ b/core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
@@ -133,8 +133,9 @@ object TaskAssigner extends Logging {
     val className = {
       val name = assignerMap.get(assignerName.toLowerCase())
       name.getOrElse {
-        logWarning(s"$assignerName cannot be constructed, fallback to default $roundrobin.")
-        roundrobin
+        throw new SparkException(s"Task Assigner $assignerName is not available. " +
+          s"Please choose roundrobin, packed, or balanced.  roundrobin is used by default")
+
       }
     }
     // The className is valid. No need to catch exceptions.
--- a/core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
+++ b/core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
@@ -31,6 +31,12 @@ private[scheduler] class OfferState(val workOffer: WorkerOffer) {
   var coresAvailable: Int = workOffer.cores
   /** The list of tasks that are assigned to this WorkerOffer. */
   val tasks = new ArrayBuffer[TaskDescription](coresAvailable)
+
+  def assignTask(task: TaskDescription, cpu: Int): Unit = {
+    tasks += task
+    coresAvailable -= cpu
+    assert(coresAvailable >= 0)
+  }
 }
 
 /**
@@ -39,7 +45,7 @@ private[scheduler] class OfferState(val workOffer: WorkerOffer) {
  * Together with [[org.apache.spark.scheduler.TaskScheduler TaskScheduler]], TaskAssigner
  * is used to assign tasks to workers with available cores. Internally, when TaskScheduler
  * performs task assignment given available workers, it first sorts the candidate tasksets,
- * and then for each taskset, it takes a number of rounds to request TaskAssigner for task
+ * and then for each taskset, it takes multiple rounds to request TaskAssigner for task
  * assignment with different locality restrictions until there is either no qualified
  * workers or no valid tasks to be assigned.
  *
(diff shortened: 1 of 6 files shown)


zhzhan:
@rxin @gatorsmile Can you please take a look, and kindly provide your comments.

core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
```diff
@@ -0,0 +1,233 @@
...
+
+  def init(conf: SparkConf): TaskAssigner = {
+    val assignerName = conf.get(config.SPARK_SCHEDULER_TASK_ASSIGNER.key, "roundrobin")
+      .toLowerCase()
+    val className = assignerMap.getOrElse(assignerName, roundrobin)
+    val CPUS_PER_TASK = conf.getInt("spark.task.cpus", 1)
+    val assigner = try {
+      logInfo(s"Constructing an assigner as $className")
+      Utils.classForName(className).getConstructor()
+        .newInstance().asInstanceOf[TaskAssigner]
```
gatorsmile: one line?

core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
```diff
@@ -0,0 +1,233 @@
...
+
+  override def init(): Unit = {
+    idx = 0
+    sorted = offer.filter(_.coresAvailable >= CPUS_PER_TASK).sortBy(_.coresAvailable)
+  }
+
+  override def hasNext: Boolean = idx < sorted.size
+
+  override def getNext(): OfferState = {
+    currentOffer = sorted(idx)
```
gatorsmile: If I just read this class, my first question is how we can ensure this will not be out of boundary? We need to leave a comment to explain this. Or add a safety check for avoiding any bug we could add in the future

zhzhan: will change to make it similar to Iterator.next() method and add comments with similar comments to the Iterator.next()

core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
```diff
@@ -0,0 +1,233 @@
...
+    currentOffer
+  }
+
+  override def offerAccepted(assigned: Boolean): Unit = {
+    if (currentOffer.coresAvailable < CPUS_PER_TASK || !assigned) {
```
viirya: If the current offer is not assigned, why we need to step to next offer if `coresAvailable` is still enough?

core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
```diff
@@ -0,0 +1,233 @@
...
+ * Assign the task to workers with the least available cores. In other words, PackedAssigner tries
+ * to schedule tasks to fewer workers.  As a result, there will be idle workers without any tasks
+ * assigned if more than required workers are reserved. If the dynamic allocator is enabled,
+ * these idle workers will be released by driver. The released resources can then be allocated to
+ * other jobs by underling resource manager. This assigner can potentially reduce the resource
+ * reservation for a job.
+ */
+class PackedAssigner extends TaskAssigner {
+  private var sorted: Seq[OfferState] = _
+  private var idx = 0
```
tejasapatil: nit: `idx` => `currentOfferIndex`

core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
mridulm: There are two cases :
a) There is no (or insufficient) locality information - in which case, what you describe will hold.
All subsequent requests will also not result in assignment.

b) If there are other executors for which sufficient locality affinity holds, then a 'later' executor in the iteration order can satisfy the locality preference.

The assignment is decided by TaskSetManager eventually - the Assigner is simply specifying the order in which iteration proceeds.

zhzhan: If the current offer is rejected, it is not valid for the current taskset, (probably due to locality restriction). Each scheduling algorithm has to respect the locality restriction, and in the meantime provide next available offer to the taskset.

core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
```diff
@@ -0,0 +1,233 @@
...
+    currentOffer = null
+  }
+}
+
+/**
+ * Assign the task to workers with the least available cores. In other words, PackedAssigner tries
+ * to schedule tasks to fewer workers.  As a result, there will be idle workers without any tasks
+ * assigned if more than required workers are reserved. If the dynamic allocator is enabled,
+ * these idle workers will be released by driver. The released resources can then be allocated to
+ * other jobs by underling resource manager. This assigner can potentially reduce the resource
```
tejasapatil: `This assigner can potentially reduce the resource reservation for jobs which overallocate resources than they need.`

core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
```diff
@@ -109,6 +108,85 @@ class TaskSchedulerImplSuite extends SparkFunSuite with LocalSparkContext with B
...
+    val taskScheduler = setupScheduler(("spark.scheduler.taskAssigner", "RoUndrObin"))
+    roundrobin(taskScheduler)
+  }
+
+  test("Fallback to roundrobin when the task assigner provided is not valid") {
+    val taskScheduler = setupScheduler("spark.scheduler.taskAssigner" -> "invalid")
+    roundrobin(taskScheduler)
+  }
+
+  test("Scheduler balance the assignment to the worker with more free cores") {
```
tejasapatil: I think you should structure this so that tests for the same task assigner are packed together. Right now, one has the look into each test case to figure out which policy is being tested.

docs/configuration.md
```diff
@@ -1342,6 +1342,19 @@ Apart from these, the following properties are also available, and may be useful
     Should be greater than or equal to 1. Number of allowed retries = this value - 1.
   </td>
 </tr>
+<tr>
+  <td><code>spark.scheduler.taskAssigner</code></td>
+  <td>roundrobin</td>
+  <td>
+    The strategy of how to allocate tasks among workers with free cores. By default, roundrobin
```
tejasapatil: nit: create a list for each policy and explain inline instead of saying `former`, `latter` below.

core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
```diff
@@ -0,0 +1,233 @@
...
+/**
+ * Assign the task to workers with the most available cores.
+ */
+class BalancedAssigner extends TaskAssigner {
+  private var maxHeap: PriorityQueue[OfferState] = _
+  private var currentOffer: OfferState = _
+
+  override def construct(workOffer: Seq[WorkerOffer]): Unit = {
+    offer = Random.shuffle(workOffer.map(o => new OfferState(o)))
+  }
```
viirya: As you will put offers into the `PriorityQueue`, is it still necessary to do shuffling?

core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
```diff
@@ -0,0 +1,233 @@
...
+    super.reset
+    idx = 0
+  }
+}
+
+/**
+ * Assign the task to workers with the most available cores.
+ */
+class BalancedAssigner extends TaskAssigner {
+  private var maxHeap: PriorityQueue[OfferState] = _
```
viirya: Use `val` and call `clear` in `init` below?

core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
```diff
@@ -60,7 +58,7 @@ private[spark] class TaskSchedulerImpl(
   def this(sc: SparkContext) = this(sc, sc.conf.get(config.MAX_TASK_FAILURES))
 
   val conf = sc.conf
-
+  private  val taskAssigner: TaskAssigner = TaskAssigner.init(conf)
```
tejasapatil: nit: extra space in `private  val`

core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
```diff
@@ -0,0 +1,233 @@
...
+
+  override def offerAccepted(assigned: Boolean): Unit = {
+    if (currentOffer.coresAvailable >= CPUS_PER_TASK && assigned) {
+      maxHeap.enqueue(currentOffer)
+    }
+  }
+
+  override def reset(): Unit = {
+    super.reset
+    maxHeap = null
```
viirya: As pointed above, call `clear`?

core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
```diff
@@ -250,24 +248,26 @@ private[spark] class TaskSchedulerImpl(
...
-      availableCpus: Array[Int],
-      tasks: IndexedSeq[ArrayBuffer[TaskDescription]]) : Boolean = {
+      taskAssigner: TaskAssigner) : Boolean = {
     var launchedTask = false
-    for (i <- 0 until shuffledOffers.size) {
-      val execId = shuffledOffers(i).executorId
-      val host = shuffledOffers(i).host
-      if (availableCpus(i) >= CPUS_PER_TASK) {
+    taskAssigner.init()
+    while(taskAssigner.hasNext) {
```
tejasapatil: nit: space after `while`

core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
```diff
@@ -0,0 +1,233 @@
...
+
+/**
+ * Assign the task to workers with the least available cores. In other words, PackedAssigner tries
+ * to schedule tasks to fewer workers.  As a result, there will be idle workers without any tasks
+ * assigned if more than required workers are reserved. If the dynamic allocator is enabled,
+ * these idle workers will be released by driver. The released resources can then be allocated to
+ * other jobs by underling resource manager. This assigner can potentially reduce the resource
+ * reservation for a job.
+ */
+class PackedAssigner extends TaskAssigner {
```
viirya: As this assigner will try to pack as more as possible tasks into the same worker, the concern would be the increasing memory pressure on the worker. Do you have experienced such issue in your practical usage?

core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
zhzhan: The comments from @mridulm in last PR. I think it is reasonable, but don't have concrete answer in my mind.

"Would be good to shuffle workOffset's for this class too.
Practically, this ensures that initial heap will be randomized when cores are the same. "

core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
```diff
@@ -0,0 +1,233 @@
...
+
+  override def hasNext: Boolean = maxHeap.nonEmpty
+
+  override def getNext(): OfferState = {
+    currentOffer = maxHeap.dequeue()
+    currentOffer
+  }
+
+  override def offerAccepted(assigned: Boolean): Unit = {
+    if (currentOffer.coresAvailable >= CPUS_PER_TASK && assigned) {
```
tejasapatil: Shouldn't you add it to the heap any ways despite of what `assigned` is set to ?

core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
zhzhan: Your concern is valid. Each scheduling algorithms has its pros and cons, and which one is chosen depends on user's requirement. We mainly want to use this to save reserved resources combined with dynamic allocation. In our pipeline, we didn't observe the problem. If it happens, we need to investigate the memory allocation part to see whether it has problem.

core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
```diff
@@ -0,0 +1,233 @@
...
+  // Returned the next assigned offer based on the task assignment strategy.
+  def getNext(): OfferState
+
+  // Invoked by the TaskScheduler to indicate whether the current offer is accepted or not so that
+  // the assigner can decide whether the current worker is valid for the next offering.
+  def offerAccepted(assigned: Boolean): Unit
+
+  // Invoked at the end of resource offering to release internally maintained resources.
+  // Subclass is responsible to release its own private resources.
+  def reset(): Unit = {
```
viirya: Do we need `reset()`? Looks like we only need `init()`. As we will call `init` before each assignment, it should be complete in resetting the status to initial.

core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
viirya: It sounds correct. However, I don't think it has real effect. Once the cores are the same, meaning no task gets assigned in previous run. So it doesn't matter if we begin with different order of offers.

core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
tejasapatil: same as before. Shouldn't you add it to the heap any ways despite of what `assigned` is set to ?

core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
```diff
@@ -0,0 +1,233 @@
...
+ * the next request. After task assignment is done, TaskScheduler invokes the tasks() to
+ * retrieve all the task assignment information, and eventually, invokes reset() method so that
+ * TaskAssigner can cleanup its internal maintained resources.
+ */
+
+private[scheduler] abstract class TaskAssigner {
+  var offer: Seq[OfferState] = _
+  var CPUS_PER_TASK = 1
+
+  def withCpuPerTask(CPUS_PER_TASK: Int): Unit = {
```
zhzhan: I tried to pass as a construct param, but int is not an object,and does not work.

core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
```diff
@@ -0,0 +1,233 @@
...
+  private val balanced = classOf[BalancedAssigner].getCanonicalName
+  private val assignerMap: Map[String, String] =
+    Map("roundrobin" -> roundrobin,
+      "packed" -> packed,
+      "balanced" -> balanced)
+
+  def init(conf: SparkConf): TaskAssigner = {
+    val assignerName = conf.get(config.SPARK_SCHEDULER_TASK_ASSIGNER.key, "roundrobin")
+      .toLowerCase()
+    val className = assignerMap.getOrElse(assignerName, roundrobin)
```
zhzhan: After discussing, it is better to keep this restriction. Please refer to last PR
https://github.com/apache/spark/pull/15218#issuecomment-249237597

core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
zhzhan: If it is rejected, it is not valid for this round of assignment for this specific task set anymore. Because it means it is not valid for all tasks in the task set.

core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
zhzhan: Similar to the reason above, if the offer is rejected, we have to move forward

core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
zhzhan: Can you please clarify?

docs/configuration.md
```diff
@@ -1342,6 +1342,19 @@ Apart from these, the following properties are also available, and may be useful
...
+    with randomness is used, which tries to allocate task to workers with available cores in
+    roundrobin manner. In addition, packed and balanced is provided. The former tries to
```
viirya: nit: packed and balanced are provided.

core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
viirya: Looks like you don't have big object which posts serious concern on this.

core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
viirya: Put a log info or warn when the given `assignerName` is not correct instead of slightly turning to default one.

core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
zhzhan: After talking to several other people, they don't feel the shuffle is strongly needed. @mridulm  If you don't mind, I will remove it in my next patch.

core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
zhzhan: Will change the test case name to make it more explicit.

core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
mridulm: @zhzhan Can you elaborate what the concern with shuffle'ing are ?
There were various reasons why we started shuffling offers.

zhzhan: @mridulm I am not sure how much the shuffle can impact the scheduling, and thus don't have strong opinion on this.
@viirya  Even if the cores are the same, it does not mean that "no task gets assigned in previous run". Shuffling does take effect here. For example, the previous round may be (5, 4, 3), and one core is allocated, then the current round would be (4, ,4 3).

mridulm: @zhzhan Let is preserve the behavior - for any application using this assigner, all tasksets will be executed based on the ordering of offers (both with and without good locality info).
The impact can be fairly non trivial - which is why shuffle'ing was initially added.

core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
```diff
@@ -0,0 +1,226 @@
...
+ * reservation for jobs, which over allocate resources than they need.
+ */
+class PackedAssigner extends TaskAssigner {
+  private var sorted: Seq[OfferState] = _
+  private var currentOfferIndex = 0
+  private var currentOffer: OfferState = _
+
+  override def init(): Unit = {
+    currentOfferIndex = 0
+    sorted = offer.filter(_.coresAvailable >= cpuPerTask).sortBy(_.coresAvailable)
```
viirya: Sounds like the shuffling reason of `BalancedAssigner` can be applied here too? If no shuffling, this ensures that initial sorted offers will be randomized when cores are the same, right?

mridulm: @viirya Agree, good point.

zhzhan:
@rxin Can you please take a look, and let me know if you have any concern?

gatorsmile:
Accidentally, I deleted all my comments. You might need to check the emails to find all my comments. :)

core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
```diff
@@ -0,0 +1,218 @@
...
+import scala.collection.mutable.ArrayBuffer
+import scala.collection.mutable.PriorityQueue
+import scala.util.Random
+
+import org.apache.spark.internal.{config, Logging}
+import org.apache.spark.SparkConf
+import org.apache.spark.util.Utils
+
+/** Tracks the current state of the workers with available cores and assigned task list. */
+class OfferState(val workOffer: WorkerOffer) {
```
wangmiao1981: Is this class private to `scheduler`?

zhzhan:
@gatorsmile I didn't see your new comments

core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
```diff
@@ -0,0 +1,218 @@
...
+  /** The list of tasks that are assigned to this WorkerOffer. */
+  val tasks = new ArrayBuffer[TaskDescription](coresAvailable)
+}
+
+/**
+ * TaskAssigner is the base class for all task assigner implementations, and can be
+ * extended to implement different task scheduling algorithms.
+ * Together with [[org.apache.spark.scheduler.TaskScheduler TaskScheduler]], TaskAssigner
+ * is used to assign tasks to workers with available cores. Internally, when TaskScheduler
+ * perform task assignment given available workers, it first sorts the candidate tasksets,
```
wangmiao1981: `perform` -> `performs`

core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
```diff
@@ -0,0 +1,218 @@
...
+ * workers or no valid tasks to be assigned.
+ *
+ * TaskAssigner is responsible to maintain the worker availability state and task assignment
+ * information. The contract between [[org.apache.spark.scheduler.TaskScheduler TaskScheduler]]
+ * and TaskAssigner is as follows.
+ *
+ * First, TaskScheduler invokes construct() of TaskAssigner to initialize the its internal
+ * worker states at the beginning of resource offering.
+ *
+ * Second, before each round of task assignment for a taskset, TaskScheduler invoke the init()
```
wangmiao1981: `invoke` -> `invokes`

core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
```diff
@@ -0,0 +1,218 @@
...
+  def construct(workOffer: Seq[WorkerOffer]): Unit = {
+    offer = Random.shuffle(workOffer.map(o => new OfferState(o)))
+  }
+
+  /** Invoked at each round of Taskset assignment to initialize the internal structure. */
+  def init(): Unit
+
+  /**
+   * Tests Whether there is offer available to be used inside of one round of Taskset assignment.
+   *  @return  `true` if a subsequent call to `next` will yield an element,
```
wangmiao1981: `@return` is not aligned with the line above.

core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
```diff
@@ -0,0 +1,218 @@
...
+  /** Invoked at the beginning of resource offering to construct the offer with the workoffers. */
+  def construct(workOffer: Seq[WorkerOffer]): Unit = {
+    offer = Random.shuffle(workOffer.map(o => new OfferState(o)))
+  }
+
+  /** Invoked at each round of Taskset assignment to initialize the internal structure. */
+  def init(): Unit
+
+  /**
+   * Tests Whether there is offer available to be used inside of one round of Taskset assignment.
```
wangmiao1981: `Whether` -> `whether`

core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
```diff
@@ -0,0 +1,218 @@
...
+    offer(currentOfferIndex)
+  }
+
+  override def offerAccepted(assigned: Boolean): Unit = {
+    currentOfferIndex += 1
+  }
+}
+
+/**
+ * Assign the task to workers with the most available cores. It other words, BalancedAssigner tries
```
wangmiao1981: `It` -> `In`

core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
```diff
@@ -0,0 +1,218 @@
...
+   * @return  the next available offer, if `hasNext` is `true`,
+   *          undefined behavior otherwise.
+   */
+  def next(): OfferState
+
+  /**
+   * Invoked by the TaskScheduler to indicate whether the current offer is accepted or not so that
+   * the assigner can decide whether the current worker is valid for the next offering.
+   */
+  def offerAccepted(assigned: Boolean): Unit
```
wangmiao1981: I remembered that in your original PR, there is a resource release method. Do you still need it?

zhzhan: Based on the review comments, we do not need it anymore. The resource will be released in the init method.

docs/configuration.md
```diff
@@ -1342,6 +1342,20 @@ Apart from these, the following properties are also available, and may be useful
...
   </td>
 </tr>
+<tr>
+  <td><code>spark.scheduler.taskAssigner</code></td>
+  <td>roundrobin</td>
+  <td>
+    The strategy of how to allocate tasks among workers with free cores. There are three task
+    assigners (roundrobin, packed, and balanced) are supported currently. By default, roundrobin
+    with randomness is used, which tries to allocate task to workers with available cores in
+    roundrobin manner.The packed task assigner tries to allocate tasks to workers with the least
```
wangmiao1981: missed space between . and `The`

core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
```diff
@@ -0,0 +1,218 @@
...
+  val tasks = new ArrayBuffer[TaskDescription](coresAvailable)
+}
+
+/**
+ * TaskAssigner is the base class for all task assigner implementations, and can be
+ * extended to implement different task scheduling algorithms.
+ * Together with [[org.apache.spark.scheduler.TaskScheduler TaskScheduler]], TaskAssigner
+ * is used to assign tasks to workers with available cores. Internally, when TaskScheduler
+ * performs task assignment given available workers, it first sorts the candidate tasksets,
+ * and then for each taskset, it takes a number of rounds to request TaskAssigner for task
```
gatorsmile: `a number of` -> `multiple`

core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
```diff
@@ -0,0 +1,218 @@
...
+ * to check the worker availability and retrieve current offering from TaskAssigner.
+ *
+ * Fourth, then offerAccepted is used by TaskScheduler to notify the TaskAssigner so that
+ * TaskAssigner can decide whether the current offer is valid or not for the next request.
+ *
+ * Fifth, After task assignment is done, TaskScheduler invokes the tasks() to
+ * retrieve all the task assignment information.
+ */
+
+private[scheduler] abstract class TaskAssigner {
```
gatorsmile: `abstract class` -> `sealed abstract class`

core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
```diff
@@ -0,0 +1,218 @@
...
+ * information. The contract between [[org.apache.spark.scheduler.TaskScheduler TaskScheduler]]
+ * and TaskAssigner is as follows.
+ *
+ * First, TaskScheduler invokes construct() of TaskAssigner to initialize the its internal
+ * worker states at the beginning of resource offering.
+ *
+ * Second, before each round of task assignment for a taskset, TaskScheduler invokes the init()
+ * of TaskAssigner to initialize the data structure for the round.
+ *
+ * Third, when performing real task assignment, hasNext()/getNext() is used by TaskScheduler
```
gatorsmile: `hasNext()` -> `hasNext`

gatorsmile: `getNext()` -> `next()`

core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
```diff
@@ -0,0 +1,218 @@
...
+ * to check the worker availability and retrieve current offering from TaskAssigner.
+ *
+ * Fourth, then offerAccepted is used by TaskScheduler to notify the TaskAssigner so that
+ * TaskAssigner can decide whether the current offer is valid or not for the next request.
+ *
+ * Fifth, After task assignment is done, TaskScheduler invokes the tasks() to
```
gatorsmile: `After` -> `after`

gatorsmile: `the tasks()` -> `the function `tasks`

core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
```diff
@@ -0,0 +1,218 @@
...
+ * First, TaskScheduler invokes construct() of TaskAssigner to initialize the its internal
+ * worker states at the beginning of resource offering.
+ *
+ * Second, before each round of task assignment for a taskset, TaskScheduler invokes the init()
+ * of TaskAssigner to initialize the data structure for the round.
+ *
+ * Third, when performing real task assignment, hasNext()/getNext() is used by TaskScheduler
+ * to check the worker availability and retrieve current offering from TaskAssigner.
+ *
+ * Fourth, then offerAccepted is used by TaskScheduler to notify the TaskAssigner so that
```
gatorsmile: `then offerAccepted is used by TaskScheduler` -> `TaskScheduler calls offerAccepted()`

core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
```diff
@@ -0,0 +1,218 @@
...
+
+private[scheduler] abstract class TaskAssigner {
+  protected var offer: Seq[OfferState] = _
+  protected var cpuPerTask = 1
+
+  protected def withCpuPerTask(cpuPerTask: Int): Unit = {
+    this.cpuPerTask = cpuPerTask
+  }
+
+  /** The final assigned offer returned to TaskScheduler. */
```
gatorsmile: Here, the implementation of this function does not guarantee the offer assignment is final or not. Thus, maybe we can change the description to `The currently assigned offers`

core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
```diff
@@ -0,0 +1,218 @@
...
+    offer(currentOfferIndex)
+  }
+
+  override def offerAccepted(assigned: Boolean): Unit = {
+    currentOfferIndex += 1
+  }
+}
+
+/**
+ * Assign the task to workers with the most available cores. In other words, BalancedAssigner tries
```
gatorsmile: `Assign` -> `Assigns`

core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
```diff
@@ -0,0 +1,218 @@
...
+   * @return  the next available offer, if `hasNext` is `true`,
+   *          undefined behavior otherwise.
+   */
+  def next(): OfferState
+
+  /**
+   * Invoked by the TaskScheduler to indicate whether the current offer is accepted or not so that
+   * the assigner can decide whether the current worker is valid for the next offering.
+   */
+  def offerAccepted(assigned: Boolean): Unit
```
gatorsmile: `assigned ` is a very critical parameter. Could you explain it in the function description (the code comment)?

core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
```diff
@@ -250,24 +251,26 @@ private[spark] class TaskSchedulerImpl(
...
-            tasks(i) += task
+            currentOffer.tasks += task
             val tid = task.taskId
             taskIdToTaskSetManager(tid) = taskSet
             taskIdToExecutorId(tid) = execId
             executorIdToTaskCount(execId) += 1
-            availableCpus(i) -= CPUS_PER_TASK
-            assert(availableCpus(i) >= 0)
+            currentOffer.coresAvailable -= CPUS_PER_TASK
+            assert(currentOffer.coresAvailable >= 0)
```
gatorsmile: Could we add a function in `class OfferState` for including the following three lines of codes? 
```Scala
currentOffer.tasks += task
currentOffer.coresAvailable -= CPUS_PER_TASK
assert(currentOffer.coresAvailable >= 0)
```

core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
```diff
@@ -305,12 +309,8 @@ private[spark] class TaskSchedulerImpl(
         hostsByRack.getOrElseUpdate(rack, new HashSet[String]()) += o.host
       }
     }
+    taskAssigner.construct(offers)
 
-    // Randomly shuffle offers to avoid always placing tasks on the same set of workers.
```
gatorsmile: Could we add this comment to [the code in `construct`](https://github.com/zhzhan/spark/blob/dd2b2077430bbb07047e928d20c1ad8fe940827a/core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala#L79)?

sql/core/src/test/scala/org/apache/spark/sql/DataFrameWindowFunctionsSuite.scala
```diff
@@ -31,7 +31,10 @@ class DataFrameWindowFunctionsSuite extends QueryTest with SharedSQLContext {
   test("reuse window partitionBy") {
     val df = Seq((1, "1"), (2, "2"), (1, "1"), (2, "2")).toDF("key", "value")
     val w = Window.partitionBy("key").orderBy("value")
-
+    val df1 = df.select(
+      lead("key", 1).over(w),
+      lead("value", 1).over(w))
+      df1.explain(true)
```
gatorsmile: Unrelated code changes. Please remove them. Thanks!

docs/configuration.md
```diff
@@ -1342,6 +1342,20 @@ Apart from these, the following properties are also available, and may be useful
     Should be greater than or equal to 1. Number of allowed retries = this value - 1.
   </td>
 </tr>
+<tr>
+  <td><code>spark.scheduler.taskAssigner</code></td>
+  <td>roundrobin</td>
+  <td>
+    The strategy of how to allocate tasks among workers with free cores. There are three task
```
gatorsmile: `There are three task` -> `Three task`

docs/configuration.md
```diff
@@ -1342,6 +1342,20 @@ Apart from these, the following properties are also available, and may be useful
...
+    assigners (roundrobin, packed, and balanced) are supported currently. By default, roundrobin
+    with randomness is used, which tries to allocate task to workers with available cores in
+    roundrobin manner. The packed task assigner tries to allocate tasks to workers with the least
```
gatorsmile: `in roundrobin manner` ->  `in a round-robin manner `

docs/configuration.md
```diff
@@ -1342,6 +1342,20 @@ Apart from these, the following properties are also available, and may be useful
...
+    free cores, resulting in tasks assigned to few workers, which may help driver to release the
```
gatorsmile: `few` -> `fewer`

docs/configuration.md
```diff
@@ -1342,6 +1342,20 @@ Apart from these, the following properties are also available, and may be useful
...
+    reserved idle workers when dynamic allocation(spark.dynamicAllocation.enabled) is enabled.
+    The balanced task assigner tries to assign tasks across workers in a balance way (allocating
+    tasks to workers with most free cores).
```
gatorsmile: `most` -> `the most`

docs/configuration.md
```diff
@@ -1342,6 +1342,20 @@ Apart from these, the following properties are also available, and may be useful
...
+  <td><code>spark.scheduler.taskAssigner</code></td>
+  <td>roundrobin</td>
+  <td>
+    The strategy of how to allocate tasks among workers with free cores. There are three task
+    assigners (roundrobin, packed, and balanced) are supported currently. By default, roundrobin
+    with randomness is used, which tries to allocate task to workers with available cores in
+    roundrobin manner. The packed task assigner tries to allocate tasks to workers with the least
+    free cores, resulting in tasks assigned to few workers, which may help driver to release the
+    reserved idle workers when dynamic allocation(spark.dynamicAllocation.enabled) is enabled.
+    The balanced task assigner tries to assign tasks across workers in a balance way (allocating
```
gatorsmile: `balance ` -> `balanced `

core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
```diff
@@ -0,0 +1,229 @@
...
+private[scheduler] class OfferState(val workOffer: WorkerOffer) {
+  /** The current remaining cores that can be allocated to tasks. */
+  var coresAvailable: Int = workOffer.cores
+  /** The list of tasks that are assigned to this WorkerOffer. */
+  val tasks = new ArrayBuffer[TaskDescription](coresAvailable)
+
+  def assignTask(task: TaskDescription, cpu: Int): Unit = {
+    tasks += task
+    coresAvailable -= cpu
+    assert(coresAvailable >= 0)
```
gatorsmile: Can we check the eligibility at the beginning of this function? Issue an exception instead of using assert?

core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
```diff
@@ -0,0 +1,229 @@
...
+    }
+    // The className is valid. No need to catch exceptions.
+    logInfo(s"Constructing TaskAssigner as $className")
+    Utils.classForName(className).getConstructor().newInstance().asInstanceOf[TaskAssigner]
+      .withCpuPerTask(cpuPerTask = conf.getInt("spark.task.cpus", 1))
+  }
+}
+
+/**
+ * Assigns the task to workers with available cores in roundrobin manner.
```
gatorsmile: `in a roundrobin manner`

core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
```diff
@@ -0,0 +1,229 @@
...
+
+  override def offerAccepted(isAccepted: Boolean): Unit = {
+    if (currentOffer.coresAvailable >= cpuPerTask && isAccepted) {
+      maxHeap.enqueue(currentOffer)
+    }
+  }
+}
+
+/**
+ * Assign the task to workers with the least available cores. In other words, PackedAssigner tries
```
gatorsmile: `Assign` -> `Assigns`

docs/configuration.md
```diff
@@ -1350,6 +1350,20 @@ Apart from these, the following properties are also available, and may be useful
     Should be greater than or equal to 1. Number of allowed retries = this value - 1.
   </td>
 </tr>
+<tr>
+  <td><code>spark.scheduler.taskAssigner</code></td>
+  <td>roundrobin</td>
+  <td>
+    The strategy of how to allocate tasks among workers with free cores. Three task
+    assigners (roundrobin, packed, and balanced) are supported currently. By default, roundrobin
+    with randomness is used to allocate task to workers with available cores in a
```
gatorsmile: `allocate task` -> `allocate tasks`

gatorsmile:
LGTM except a few minor comments. Could @rxin do the final check? 

Thank you for your work! @zhzhan

core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
```diff
@@ -0,0 +1,229 @@
...
+ * Fifth, after task assignment is done, TaskScheduler invokes the function tasks to
+ * retrieve all the task assignment information.
+ */
+
+private[scheduler] sealed abstract class TaskAssigner {
+  protected var offer: Seq[OfferState] = _
+  protected var cpuPerTask = 1
+
+  protected def withCpuPerTask(cpuPerTask: Int): TaskAssigner = {
+    this.cpuPerTask = cpuPerTask
```
viirya: Do we need to add sanity check for `cpuPerTask`?

zhzhan: Please clarify what you mean. Thanks.

zhzhan: You mean cpuPerTask >= 1? I don't  think we need this check.

viirya: I mean checking if `cpuPerTask` is a reasonable value like `cpuPerTask > 0`.

zhzhan: @viirya  Thanks for the quick response. Personally I don't think we need it. You can refer to other places in the existing spark code.

core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
```diff
@@ -250,24 +251,24 @@ private[spark] class TaskSchedulerImpl(
...
         try {
           for (task <- taskSet.resourceOffer(execId, host, maxLocality)) {
-            tasks(i) += task
+            currentOffer.assignTask(task, CPUS_PER_TASK)
             val tid = task.taskId
             taskIdToTaskSetManager(tid) = taskSet
             taskIdToExecutorId(tid) = execId
             executorIdToTaskCount(execId) += 1
-            availableCpus(i) -= CPUS_PER_TASK
-            assert(availableCpus(i) >= 0)
```
viirya: This question might not be directly related to your change here. But looks like we do check current offer has enough cores in above `if`. Then we go loop tasks in this task set. If the task is assigned to the current offer, we minus `availableCpus`. Looks like we don't additionally check if `availableCpus` is enough to hold the task? Is it possibly we assign tasks to the current offer until it runs of of `availableCpus` and this assert goes failed?

core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
viirya: @zhzhan ok.

core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
viirya: This is not questioning about your change but towards the legacy code. I am curious if this would cause a problem or not.

viirya:
LGTM. I have a question for the legacy code, but not related to this change.

viirya:
One concern I have is the cluster of machines with different codes. It is possibly if you construct your cluster with solution like ec2 and buy different kinds of nodes. In this case, the balance assigner would first consume machines with higher cores, and packed assigner would consume machines with less codes. I don't know if this is an issue to most of you. Even it is, this might not be able to solve in this change and we can consider this later.

core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
zhzhan: @viirya The assert will not fail even in the legacy code, because  taskSet.resourceOffer(execId, host, maxLocality) return an Option. The for loop at most run once.

viirya: @zhzhan ah, thanks for explanation! I don't notice it returns an Option. :)

zhzhan:
@rxin Would you like to take a look and let you know if you have any concern? Thanks.

gatorsmile:
retest this please

rxin:
Sure will take a look in the next couple of days to get this into 2.1 if possible.

core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
```diff
@@ -0,0 +1,232 @@
...
+  }
+}
+
+/**
+ * Assigns the task to workers with the most available cores. In other words, BalancedAssigner tries
+ * to distribute the task across workers in a balanced way. Potentially, it may alleviate the
+ * workers' memory pressure as less tasks running on the same workers, which also indicates that
+ * the task itself can make use of more computation resources, e.g., hyper-thread, across clusters.
+ */
+class BalancedAssigner extends TaskAssigner {
```
zhzhan: @rxin These two do the opposite thing. The packed scheduler tries to schedule tasks to workers as few as possible so that some workers without task running can be released. The balanced assigner tries to schedule the tasks to workers with least work load.

If user wants optimal resource reservation, they may want the packer assigner. If user want observe some memory pressure, they may want to try the balanced assigner.

rxin: Sorry I made a mistake -- I meant to ask the difference between balance and round robin. Isn't the two similar?

rxin:
To be honest, I find the current API pretty weird (it is some stateful object that has to be reset every time). I suspect you designed this API by just abstracting out the logic you wanted to change from the existing implementation, but that doesn't necessarily lead to intuitive apis. It's been a while since I last checked the scheduler code, so it'd take me a while to page back in.

core/src/main/scala/org/apache/spark/scheduler/TaskAssigner.scala
zhzhan: These two assigner may behave similar in practice. The difference is that the balanced assigner tries to distribute the work load more aggressively.

zhzhan:
@rxin Thanks for the feedback regarding the TaskAssigner API. The current API is designed based on the current logic of TaskSchedulerImp, where the scheduler takes many rounds to assign the tasks for each task set. I have not figured out a better way yet. Any suggestions are welcome.

jiangxb1987:
@rxin @zhzhan Is there any chance that we could reach any consensus on the API design and move forward with this PR?

PR #15542, opened 2016-10-19
[SPARK-17996][SQL] Fix unqualified catalog.getFunction(...)

hvanhovell:

Currently an unqualified `getFunction(..)`call returns a wrong result; the returned function is shown as temporary function without a database. For example:

```
scala> sql("create function fn1 as 'org.apache.hadoop.hive.ql.udf.generic.GenericUDFAbs'")

scala> spark.catalog.getFunction("fn1")
res1: org.apache.spark.sql.catalog.Function = Function[name='fn1', className='org.apache.hadoop.hive.ql.udf.generic.GenericUDFAbs', isTemporary='true']
```

This PR fixes this by adding database information to ExpressionInfo (which is used to store the function information).
## How was this patch tested?

Added more thorough tests to `CatalogSuite`.


hvanhovell:
cc @srinathshankar

sql/core/src/main/scala/org/apache/spark/sql/internal/CatalogImpl.scala
```diff
@@ -133,11 +133,11 @@ class CatalogImpl(sparkSession: SparkSession) extends Catalog {
...
     val metadata = sessionCatalog.lookupFunctionInfo(funcIdent)
     new Function(
-      name = funcIdent.identifier,
-      database = funcIdent.database.orNull,
+      name = metadata.getName,
+      database = metadata.getDb,
       description = null, // for now, this is always undefined
       className = metadata.getClassName,
-      isTemporary = funcIdent.database.isEmpty)
+      isTemporary = metadata.getDb == null)
```
srinathshankar: Is this still the right way to do this ? What about global temp tables ?

srinathshankar (approved):
Some minor comments

sql/core/src/main/scala/org/apache/spark/sql/internal/CatalogImpl.scala
hvanhovell: We do not support global temp functions. So this is the right way to do it.

sql/catalyst/src/main/java/org/apache/spark/sql/catalyst/expressions/ExpressionInfo.java
```diff
@@ -42,14 +43,23 @@ public String getExtended() {
         return extended;
     }
 
-    public ExpressionInfo(String className, String name, String usage, String extended) {
+    public String getDb() {
+        return db;
+    }
+
+    public ExpressionInfo(String className, String name, String usage, String extended, String db) {
```
hvanhovell: Done

sql/core/src/main/scala/org/apache/spark/sql/execution/command/functions.scala
```diff
@@ -118,14 +118,16 @@ case class DescribeFunctionCommand(
       case _ =>
         try {
           val info = sparkSession.sessionState.catalog.lookupFunctionInfo(functionName)
+          val db = if (info.getDb != null) info.getDb + "." else ""
```
hvanhovell: Done.

hvanhovell:
Merging to master. Thanks for the review.

PR #15544, opened 2016-10-19
[SPARK-17997] [SQL] Add an aggregation function for counting distinct values for multiple intervals

wzhfy:

This work is a part of [SPARK-17074](https://issues.apache.org/jira/browse/SPARK-17074) to compute equi-height histograms. Equi-height histogram is an array of bins. A bin consists of two endpoints which form an interval of values and the ndv in that interval.

This PR creates a new aggregate function, given an array of endpoints, counting distinct values (ndv) in intervals among those endpoints.

This PR also refactors `HyperLogLogPlusPlus` by extracting a helper class `HyperLogLogPlusPlusHelper`, where the underlying HLLPP algorithm locates.

## How was this patch tested?

Add new test cases.


wzhfy:
@hvanhovell @yhuai @cloud-fan Can you review this? Thanks!

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/HyperLogLogPlusPlus.scala
```diff
@@ -142,319 +84,37 @@ case class HyperLogLogPlusPlus(
 
   override def inputTypes: Seq[AbstractDataType] = Seq(AnyDataType)
 
-  override def aggBufferSchema: StructType = StructType.fromAttributes(aggBufferAttributes)
-
-  /** Allocate enough words to store all registers. */
-  override val aggBufferAttributes: Seq[AttributeReference] = Seq.tabulate(numWords) { i =>
-    AttributeReference(s"MS[$i]", LongType)()
+  override def createAggregationBuffer(): HyperLogLogPlusPlusAlgo = {
+    new HyperLogLogPlusPlusAlgo(relativeSD)
```
wzhfy: @hvanhovell Do you mean we put some digest object (like how PercentileDigest does in ApproximatePercentile) in the buffer instead of putting HLL++ directly in it?

wzhfy: @hvanhovell Do you mean we should not use `TypedImperativeAggregate` instead of `ImperativeAggregate`?

wzhfy:
retest this please

wzhfy:
retest this please

wzhfy:
@hvanhovell Now HyperLogLogPlusPlus is kept as an ImperativeAggregate and the algorithm is separated from it. I also modify the newly added IntervalDistinctApprox as an ImperativeAggregate too.

wzhfy:
retest this please

wzhfy:
@rxin @hvanhovell @cloud-fan Do you think this is a useful aggregate function?
It used to compute ndv for bins in a histogram, but we are not using histogram for now.
It still can be used to compute ndv for a sequence of intervals.
If you think it's useful, I'll update it, otherwise I'll close it. Thanks.

rxin:
@wzhfy let's close this one for now. I think we can revisit it in the future when we introduce more advanced statistics.

wzhfy:
@cloud-fan @gatorsmile  @sameeragarwal Because we are going to add equi-height histogram, I reopened this PR and updated description and code. Please help review it. Thanks.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/HyperLogLogPlusPlusHelper.scala
```diff
@@ -0,0 +1,373 @@
...
+
+import java.lang.{Long => JLong}
+import java.util
+
+import org.apache.spark.sql.catalyst.InternalRow
+import org.apache.spark.sql.catalyst.expressions.XxHash64Function
+import org.apache.spark.sql.types._
+
+// A helper class for HyperLogLogPlusPlus.
+class HyperLogLogPlusPlusHelper(relativeSD: Double) extends Serializable {
```
viirya: Is this code mainly copied from `HyperLogLogPlusPlus`?

wzhfy: Yes.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/FunctionRegistry.scala
```diff
@@ -270,6 +270,7 @@ object FunctionRegistry {
     expression[Remainder]("%"),
 
     // aggregate functions
+    expression[ApproxCountDistinctForIntervals]("approx_count_distinct_for_intervals"),
```
cloud-fan: If it's only used internally, we don't need to register it here.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/ApproxCountDistinctForIntervals.scala
```diff
@@ -0,0 +1,248 @@
...
+/**
+ * This function counts the approximate number of distinct values (ndv) in
+ * intervals constructed from endpoints specified in `endpointsExpression`. The endpoints will be
+ * sorted into ascending order. To count ndv's in these intervals, apply the HyperLogLogPlusPlus
+ * algorithm in each of them.
+ * @param child to estimate the ndv's of.
+ * @param endpointsExpression to construct the intervals.
+ * @param relativeSD The maximum estimation error allowed in the HyperLogLogPlusPlus algorithm.
+ */
+@ExpressionDescription(
```
cloud-fan: we don't need this if it's an internal func

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/ApproxCountDistinctForIntervals.scala
```diff
@@ -0,0 +1,248 @@
...
+  // Mark as lazy so that endpointsExpression is not evaluated during tree transformation.
+  lazy val endpoints: Array[Double] = {
+    val doubleArray = (endpointsExpression.dataType, endpointsExpression.eval()) match {
+      case (ArrayType(baseType: NumericType, _), arrayData: ArrayData) =>
+        val numericArray = arrayData.toObjectArray(baseType)
+        numericArray.map { x =>
+          baseType.numeric.toDouble(x.asInstanceOf[baseType.InternalType])
+        }
+    }
+    util.Arrays.sort(doubleArray)
```
cloud-fan: again, if it's only used internally, we can require the caller side to pass the endpoints sorted.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/ApproxCountDistinctForIntervals.scala
```diff
@@ -0,0 +1,248 @@
...
+  // N endpoints construct N-1 intervals, creating a HLLPP for each interval
+  private lazy val hllppArray = {
+    val array = new Array[HyperLogLogPlusPlusHelper](endpoints.length - 1)
+    for (i <- array.indices) {
+      array(i) = new HyperLogLogPlusPlusHelper(relativeSD)
+    }
+    array
+  }
+
+  private lazy val numWordsPerHllpp = hllppArray.head.numWords
```
cloud-fan: do all hllpphelper have same value of `numWords`?

wzhfy: yes, `numWords` is determined by parameter `relativeSD`, which is the same for all HLLPPHelper.

wzhfy:
retest this please

wzhfy:
retest this please

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/ApproxCountDistinctForIntervals.scala
```diff
@@ -0,0 +1,232 @@
...
+    }
+  }
+
+  // N endpoints construct N-1 intervals, creating a HLLPP for each interval
+  private lazy val hllppArray = {
+    val array = new Array[HyperLogLogPlusPlusHelper](endpoints.length - 1)
+    for (i <- array.indices) {
+      array(i) = new HyperLogLogPlusPlusHelper(relativeSD)
+    }
+    array
```
cloud-fan: add `assert(array.map(_.numWords).distinct == 1)`

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/aggregate/ApproxCountDistinctForIntervalsSuite.scala
```diff
@@ -0,0 +1,207 @@
...
+    aggFunc.initialize(buffer)
+    buffer
+  }
+
+  test("merging ApproxCountDistinctForIntervals instances") {
+    val (aggFunc, input, buffer1a) = createEstimator(Array[Double](0, 10, 2000, 345678, 1000000))
+    val buffer1b = createBuffer(aggFunc)
+    val buffer2 = createBuffer(aggFunc)
+
+    // Create the
```
cloud-fan: typo?

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/aggregate/ApproxCountDistinctForIntervalsSuite.scala
```diff
@@ -0,0 +1,207 @@
...
+    checkHllppIndex(endpoints = endpoints, value = 3, expectedIntervalIndex = 0)
+    checkHllppIndex(endpoints = endpoints, value = 6, expectedIntervalIndex = 1)
+    checkHllppIndex(endpoints = endpoints, value = 10, expectedIntervalIndex = 2)
+    // value is not found
+    checkHllppIndex(endpoints = endpoints, value = 2, expectedIntervalIndex = 0)
+    checkHllppIndex(endpoints = endpoints, value = 4, expectedIntervalIndex = 1)
+    checkHllppIndex(endpoints = endpoints, value = 8, expectedIntervalIndex = 2)
+
+    // value is the same as multiple boundaries
+    checkHllppIndex(endpoints = Array(7, 7, 7, 9), value = 7, expectedIntervalIndex = 0)
```
cloud-fan: are `Array(7, 7, 7, 9)` valid end points?

wzhfy: yes, it means 7 is the min value and it occupies more than 2/3 buckets.

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/aggregate/ApproxCountDistinctForIntervalsSuite.scala
```diff
@@ -0,0 +1,206 @@
...
+import org.apache.spark.sql.catalyst.InternalRow
+import org.apache.spark.sql.catalyst.analysis.TypeCheckResult.TypeCheckFailure
+import org.apache.spark.sql.catalyst.expressions.{AttributeReference, BoundReference, CreateArray, Literal, SpecificInternalRow}
+import org.apache.spark.sql.catalyst.util.ArrayData
+import org.apache.spark.sql.types._
+
+class ApproxCountDistinctForIntervalsSuite extends SparkFunSuite {
+
+  test("fails analysis if parameters are invalid") {
+    def assertEqual[T](left: T, right: T): Unit = {
```
cloud-fan: why do we need this?

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/aggregate/ApproxCountDistinctForIntervalsSuite.scala
```diff
@@ -0,0 +1,206 @@
...
+    // value is the same as multiple boundaries
+    checkHllppIndex(endpoints = Array(7, 7, 7, 9), value = 7, expectedIntervalIndex = 0)
+    checkHllppIndex(endpoints = Array(3, 5, 7, 7, 7), value = 7, expectedIntervalIndex = 1)
+    checkHllppIndex(endpoints = Array(1, 3, 5, 7, 7, 9), value = 7, expectedIntervalIndex = 2)
+  }
+
+  test("basic operations: update, merge, eval...") {
+    val endpoints = Array[Double](0, 0.33, 0.6, 0.6, 0.6, 1.0)
+    val data: Seq[Double] = Seq(0, 0.6, 0.3, 1, 0.6, 0.5, 0.6, 0.33)
+    val expectedNdvs = Array[Long](3, 2, 1, 1, 1)
```
cloud-fan: this can be inlined

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/aggregate/ApproxCountDistinctForIntervalsSuite.scala
```diff
@@ -0,0 +1,206 @@
...
+      }
+      // before eval(), for intervals with the same endpoints, only the first interval counts the
+      // value
+      checkNDVs(
+        ndvs = aggFunc.hllppResults(buffer),
+        expectedNdvs = Array(3, 2, 0, 0, 1),
+        rsd = relativeSD)
+
+      // A value out of the whole range will not change the buffer
+      input.setInt(0, 2)
```
cloud-fan: the input is double type, shall we do `input.setDouble(0, 2.0)` here?

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/aggregate/ApproxCountDistinctForIntervalsSuite.scala
wzhfy: oh, I'll remove this. Previously I put some other logic here, but we should remove it now.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/ApproxCountDistinctForIntervals.scala
```diff
@@ -0,0 +1,235 @@
...
+ * (endpoint_1, endpoint_2, ... endpoint_N), returns the approximate ndv's for intervals
+ * [endpoint_1, endpoint_2], (endpoint_2, endpoint_3], ... (endpoint_N-1, endpoint_N].
+ * To count ndv's in these intervals, apply the HyperLogLogPlusPlus algorithm in each of them.
+ * @param child to estimate the ndv's of.
+ * @param endpointsExpression to construct the intervals, should be sorted into ascending order.
+ * @param relativeSD The maximum estimation error allowed in the HyperLogLogPlusPlus algorithm.
+ */
+case class ApproxCountDistinctForIntervals(
+    child: Expression,
+    endpointsExpression: Expression,
```
viirya: `endpointsExpression`  is foldable?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/ApproxCountDistinctForIntervals.scala
```diff
@@ -0,0 +1,235 @@
...
+
+/**
+ * This function counts the approximate number of distinct values (ndv) in
+ * intervals constructed from endpoints specified in `endpointsExpression`. The endpoints should be
+ * sorted into ascending order. E.g., given an array of endpoints
+ * (endpoint_1, endpoint_2, ... endpoint_N), returns the approximate ndv's for intervals
+ * [endpoint_1, endpoint_2], (endpoint_2, endpoint_3], ... (endpoint_N-1, endpoint_N].
+ * To count ndv's in these intervals, apply the HyperLogLogPlusPlus algorithm in each of them.
+ * @param child to estimate the ndv's of.
+ * @param endpointsExpression to construct the intervals, should be sorted into ascending order.
```
viirya: `An array expression with `NumericType` element to construct the intervals .... Must be foldable.`

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/ApproxCountDistinctForIntervals.scala
```diff
@@ -0,0 +1,235 @@
...
+          baseType.numeric.toDouble(x.asInstanceOf[baseType.InternalType])
+        }
+    }
+
+  override def checkInputDataTypes(): TypeCheckResult = {
+    val defaultCheck = super.checkInputDataTypes()
+    if (defaultCheck.isFailure) {
+      defaultCheck
+    } else if (!endpointsExpression.foldable) {
+      TypeCheckFailure("The intervals provided must be constant literals")
```
viirya: Should we also check element type of `endpointsExpression`?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/ApproxCountDistinctForIntervals.scala
```diff
@@ -0,0 +1,235 @@
...
+  }
+
+  override def inputTypes: Seq[AbstractDataType] = {
+    Seq(TypeCollection(NumericType, TimestampType, DateType), ArrayType)
+  }
+
+  // Mark as lazy so that endpointsExpression is not evaluated during tree transformation.
+  lazy val endpoints: Array[Double] =
+    (endpointsExpression.dataType, endpointsExpression.eval()) match {
+      case (ArrayType(baseType: NumericType, _), arrayData: ArrayData) =>
```
viirya: The type of `child` can be `TimestampType` and `DateType`, but endpoints can only be `ArrayType` of `NumericType`. It may not be convenient to set up numeric endpoints for a timestamp or date child column.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/ApproxCountDistinctForIntervals.scala
```diff
@@ -0,0 +1,235 @@
...
+  def findHllppIndex(value: Double): Int = {
+    var index = util.Arrays.binarySearch(endpoints, value)
+    if (index >= 0) {
+      // The value is found.
+      if (index == 0) {
+        0
+      } else {
+        // If the endpoints contains multiple elements with the specified value, there is no
+        // guarantee which one binarySearch will return. We remove this uncertainty by moving the
+        // index to the first position of these elements.
```
viirya: Can we just ask there no duplicate points in the endpoints?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/ApproxCountDistinctForIntervals.scala
wzhfy: Yes. I'll improve the comment for `endpointsExpression` as you suggested.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/ApproxCountDistinctForIntervals.scala
wzhfy: duplicate points mean skewed value, i.e. the value occupies several buckets.

viirya: hmm, e.g. an endpoints (1, 5, 5, 10), what you mean skewed value for this case? Is it meaningful to compute a ndv for bin(5, 5]?

wzhfy: duplicate points mean skewed value, i.e. the value occupies several buckets.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/ApproxCountDistinctForIntervals.scala
wzhfy: Good point. I think we should also allow date/timestamp endpoints.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/ApproxCountDistinctForIntervals.scala
wzhfy: This function is used in computing equi-height histograms. First, we compute equi-height bucket endpoints (percentiles). Next, we compute ndv's for intervals constructed by these endpoints. So (1, 5, 5, 10) means each bucket (1, 5), (5, 5), (5, 10) have 1/3 of records.
Mathematically it's not necessary for compute ndv for (5, 5], but it would simplify implementation when we generate histogram. Besides, it also makes this function more general.

viirya: Oh. I see. So the endpoints are not arbitrary but computed for equi-height histograms?

wzhfy: Yes.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/ApproxCountDistinctForIntervals.scala
wzhfy: This aggregate function is only used internally, the expression will use internal types, so there is no case using date/timestamp endpoints.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/ApproxCountDistinctForIntervals.scala
wzhfy: I'll check element type of `endpointsExpression`, thanks.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/ApproxCountDistinctForIntervals.scala
viirya: We'd better explain this in the expression comment.

cloud-fan:
LGTM, merging to master!

PR #15525, opened 2016-10-18
[SPARK-17985][CORE] Bump commons-lang3 version to 3.5.

ueshin:
## What changes were proposed in this pull request?

`SerializationUtils.clone()` of commons-lang3 (<3.5) has a bug that breaks thread safety, which gets stack sometimes caused by race condition of initializing hash map.
See https://issues.apache.org/jira/browse/LANG-1251.
## How was this patch tested?

Existing tests.

PR #15548, opened 2016-10-19
[SPARK-17985][CORE] Bump commons-lang3 version to 3.5.

ueshin:

`SerializationUtils.clone()` of commons-lang3 (<3.5) has a bug that breaks thread safety, which gets stack sometimes caused by race condition of initializing hash map.
See https://issues.apache.org/jira/browse/LANG-1251.
## How was this patch tested?

Existing tests.


srowen:
Given the discussion in https://github.com/apache/spark/pull/15525 indicates this change didn't cause the error, I'll merge it again.

ueshin:
The build for hadoop-2.2 has finished successfully.
https://amplab.cs.berkeley.edu/jenkins/job/spark-master-test-sbt-hadoop-2.2/1956/
Thanks.

PR #15550, opened 2016-10-19
[SPARK-18003][Spark Core] Fix bug of RDD zipWithIndex & zipWithUniqueId index value overflowing

WeichenXu123:
- Fix bug of RDD `zipWithIndex` generating wrong result when one partition contains more than 2147483647 records.
- Fix bug of RDD `zipWithUniqueId` generating wrong result when one partition contains more than 2147483647 records.
## How was this patch tested?

test added.

### DIFF
--- a/core/src/main/scala/org/apache/spark/rdd/ZippedWithIndexRDD.scala
+++ b/core/src/main/scala/org/apache/spark/rdd/ZippedWithIndexRDD.scala
@@ -64,8 +64,7 @@ class ZippedWithIndexRDD[T: ClassTag](prev: RDD[T]) extends RDD[(T, Long)](prev)
 
   override def compute(splitIn: Partition, context: TaskContext): Iterator[(T, Long)] = {
     val split = splitIn.asInstanceOf[ZippedWithIndexRDDPartition]
-    firstParent[T].iterator(split.prev, context).zipWithIndex.map { x =>
-      (x._1, split.startIndex + x._2)
-    }
+    val parentIter = firstParent[T].iterator(split.prev, context)
+    Utils.getIteratorZipWithIndex(parentIter, split.startIndex)
   }
 }
--- a/core/src/main/scala/org/apache/spark/rdd/RDD.scala
+++ b/core/src/main/scala/org/apache/spark/rdd/RDD.scala
@@ -1278,7 +1278,7 @@ abstract class RDD[T: ClassTag](
   def zipWithUniqueId(): RDD[(T, Long)] = withScope {
     val n = this.partitions.length.toLong
     this.mapPartitionsWithIndex { case (k, iter) =>
-      iter.zipWithIndex.map { case (item, i) =>
+      Utils.getIteratorZipWithIndex(iter, 0L).map { case (item, i) =>
         (item, i * n + k)
       }
     }
--- a/core/src/main/scala/org/apache/spark/util/Utils.scala
+++ b/core/src/main/scala/org/apache/spark/util/Utils.scala
@@ -1699,6 +1699,21 @@ private[spark] object Utils extends Logging {
   }
 
   /**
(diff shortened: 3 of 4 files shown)


core/src/main/scala/org/apache/spark/rdd/ZippedWithIndexRDD.scala
```diff
@@ -64,8 +64,14 @@ class ZippedWithIndexRDD[T: ClassTag](prev: RDD[T]) extends RDD[(T, Long)](prev)
 
   override def compute(splitIn: Partition, context: TaskContext): Iterator[(T, Long)] = {
     val split = splitIn.asInstanceOf[ZippedWithIndexRDDPartition]
-    firstParent[T].iterator(split.prev, context).zipWithIndex.map { x =>
-      (x._1, split.startIndex + x._2)
+    val parentIter = firstParent[T].iterator(split.prev, context)
```
rxin: we should add a line of comment saying we don't use Scala's zipWithIndex to avoid overflowing.

core/src/main/scala/org/apache/spark/rdd/ZippedWithIndexRDD.scala
```diff
@@ -64,8 +64,14 @@ class ZippedWithIndexRDD[T: ClassTag](prev: RDD[T]) extends RDD[(T, Long)](prev)
...
+    new Iterator[(T, Long)] {
+      var idxAcc: Long = -1L
```
rxin: private[this]

rxin: also why don't we initialize this to split.startIndex?

I'd also rename this just "index"

core/src/test/scala/org/apache/spark/rdd/RDDSuite.scala
```diff
@@ -833,6 +833,30 @@ class RDDSuite extends SparkFunSuite with SharedSparkContext {
     }
   }
 
+  test("zipWithIndex with partition size exceeding MaxInt") {
+    val result = sc.parallelize(Seq(1), 1).mapPartitions(
```
WeichenXu123: yeah, here I run the test, because the loop of CPU is very fast, running this testcase on my machine is about 3s. Does it still need optimized ?

rxin: hm that's about 299s longer than I'd like for a unit test ...

WeichenXu123: all right. I'll update the testcase.

core/src/main/scala/org/apache/spark/rdd/ZippedWithIndexRDD.scala
mridulm: or rather split.startIndex - 1 given the current code for idxAcc.

tejasapatil:
Have you looked for other places in the codebase which would also produce wrong result by using scala's `zipWithIndex()` ?

WeichenXu123:
@tejasapatil I check the code where reference `RDD.zipWithIndex`, there are 7 usages currently. Because current test data won't generate big enough partition so they don't triggers the bug. The referencing code   are all correct no need to update them.

tejasapatil:
I meant scala's `zipWithIndex()` and not `RDD.zipWithIndex`.

rxin:
The current change LGTM (pending Jenkins). It would be great to check the usage as @tejasapatil said.

WeichenXu123:
@tejasapatil check other reference to scala's `zipWithIndex` and fix similar case in `RDD.zipWithUniqueId`, thanks!

rxin:
Cool - LGTM! (I will merge once Jenkins comes back positive)

rxin:
Merging in master / branch-2.0.

PR #15551, opened 2016-10-19
[SPARK-18012][SQL] Simplify WriterContainer

rxin:

This patch refactors WriterContainer to simplify the logic and make control flow more obvious.The previous code setup made it pretty difficult to track the actual dependencies on variables and setups because the driver side and the executor side were using the same set of variables.
## How was this patch tested?

N/A - this should be covered by existing tests.


rxin:
cc @cloud-fan, @liancheng, @hvanhovell 

What do you think about the high level idea? The best way to get an idea of this is to look at the source code directly, rather than looking at the diff.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/WriteOutput.scala
```diff
@@ -0,0 +1,478 @@
...
+
+
+/**
+ * A helper object for writing data out to an existing partition. This is a work-in-progress
+ * refactoring. The goal is to eventually remove the class hierarchy in WriterContainer.
+ */
+object WriteOutput extends Logging {
+
+  def write(
+      sparkSession: SparkSession,
```
rxin: i'm not too happy this function has a lot of arguments

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/WriteOutput.scala
```diff
@@ -0,0 +1,478 @@
...
+      } catch { case cause: Throwable =>
+        logError(s"Aborting job ${job.getJobID}.", cause)
+        committer.abortJob(job, JobStatus.State.FAILED)
+
+        throw new SparkException("Job aborted.", cause)
+      }
+    }
+  }
+
+  class WriteJobDescription(
```
rxin: this includes the all description needed in each task for the write job.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/WriteOutput.scala
```diff
@@ -0,0 +1,478 @@
...
+    committer.setupTask(taskAttemptContext)
+
+    if (description.partitionColumns.isEmpty && description.bucketSpec.isEmpty) {
+      executeSingleDirectoryWriteTask(
+        description,
+        taskAttemptContext,
+        committer,
+        iterator)
+    } else {
+      executeDynamicPartitionWriteTask(
```
rxin: note to self: it'd be better if we can move the task commit and abort logic into this function

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/WriteOutput.scala
```diff
@@ -0,0 +1,478 @@
...
+    hadoopConf.set("mapred.task.id", taskAttemptId.toString)
+    hadoopConf.setBoolean("mapred.task.is.map", true)
+    hadoopConf.setInt("mapred.task.partition", 0)
+
+    val taskAttemptContext = new TaskAttemptContextImpl(hadoopConf, taskAttemptId)
+    val committer = newOutputCommitter(
+      description.outputFormatClass, taskAttemptContext, description.path, description.isAppend)
+    committer.setupTask(taskAttemptContext)
+
+    if (description.partitionColumns.isEmpty && description.bucketSpec.isEmpty) {
```
cloud-fan: Now we will run this if-else at executor side right? Not a big deal, and worth it to make the code simpler.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/WriteOutput.scala
```diff
@@ -0,0 +1,514 @@
...
+      Utils.tryWithSafeFinallyAndFailureCallbacks(block = {
+        // Execute the task to write rows out
+        writeTask.execute(iterator)
+        writeTask.releaseResources()
+
+        // Commit the task
+        SparkHadoopMapRedUtil.commitTask(committer, taskAttemptContext, jobId.getId, taskId.getId)
+      })(catchBlock = {
+        // If there is an error, release resource and then abort the task
+        try {
```
liancheng: Can we eliminate this `try` and just put the `finally` clause into `finallyBlock` of `tryWithSafeFinallyAndFailureCallbacks`?

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/WriteOutput.scala
```diff
@@ -0,0 +1,514 @@
...
+          writeTask.releaseResources()
+        } finally {
+          committer.abortTask(taskAttemptContext)
+          logError(s"Job $jobId aborted.")
+        }
+      })
+    } catch {
+      case t: Throwable =>
+        throw new SparkException("Task failed while writing rows", t)
+    }
```
liancheng: Seems that the only purpose of this outermost `try` is to wrap `SparkException` over any thrown exception.

Curious about the contract here, when do we want a `SparkException` rather than some random exception class?

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/WriteOutput.scala
```diff
@@ -0,0 +1,512 @@
...
+
+      val sortingKeySchema = StructType(sortingExpressions.map {
+        case a: Attribute => StructField(a.name, a.dataType, a.nullable)
+        // The sorting expressions are all `Attribute` except bucket id.
+        case _ => StructField("bucketId", IntegerType, nullable = false)
+      })
+
+      // Returns the data columns to be written given an input row
+      val getOutputRow =
+      UnsafeProjection.create(description.nonPartitionColumns, description.allColumns)
```
liancheng: Nit: Indentation is off.

liancheng (commented):
Overall LGTM, left some minor comments and questions.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/WriteOutput.scala
rxin: No because we can't abort the task if there is no error (which would happen if we put it in finally).

liancheng:
LGTM

rxin:
cc @tejasapatil fyi on the change

tejasapatil:
@rxin : Thanks for notifying me.

PR #15552, opened 2016-10-19
[SPARK-18007][SparkR][ML] update SparkR MLP - add initalWeights parameter

WeichenXu123:

update SparkR MLP, add initalWeights parameter.
## How was this patch tested?

test added.

### DIFF
--- a/R/pkg/R/mllib.R
+++ b/R/pkg/R/mllib.R
@@ -665,6 +665,7 @@ setMethod("predict", signature(object = "KMeansModel"),
 #' @param tol convergence tolerance of iterations.
 #' @param stepSize stepSize parameter.
 #' @param seed seed parameter for weights initialization.
+#' @param initialWeights initialWeights parameter for weights initialization.
 #' @param ... additional arguments passed to the method.
 #' @return \code{spark.mlp} returns a fitted Multilayer Perceptron Classification Model.
 #' @rdname spark.mlp
@@ -695,7 +696,7 @@ setMethod("predict", signature(object = "KMeansModel"),
 #' @note spark.mlp since 2.1.0
 setMethod("spark.mlp", signature(data = "SparkDataFrame"),
           function(data, layers, blockSize = 128, solver = "l-bfgs", maxIter = 100,
-                   tol = 1E-6, stepSize = 0.03, seed = NULL) {
+                   tol = 1E-6, stepSize = 0.03, seed = NULL, initialWeights = NULL) {
             if (is.null(layers)) {
               stop ("layers must be a integer vector with length > 1.")
             }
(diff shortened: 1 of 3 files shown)


R/pkg/R/mllib.R
```diff
@@ -706,10 +707,13 @@ setMethod("spark.mlp", signature(data = "SparkDataFrame"),
             if (!is.null(seed)) {
               seed <- as.character(as.integer(seed))
             }
+            if (!is.null(initialWeights)) {
+              initialWeights <- as.array(as.numeric(na.omit(initialWeights)))
```
felixcheung: is the `as.array` required here?

R/pkg/R/mllib.R
```diff
@@ -665,6 +665,7 @@ setMethod("predict", signature(object = "KMeansModel"),
 #' @param tol convergence tolerance of iterations.
 #' @param stepSize stepSize parameter.
 #' @param seed seed parameter for weights initialization.
+#' @param initialWeights initialWeights parameter for weights initialization.
```
felixcheung: could you clarify that this should be a vector - also is there any restrictions on the length of this?

R/pkg/inst/tests/testthat/test_mllib.R
```diff
@@ -410,6 +410,16 @@ test_that("spark.mlp", {
   model <- spark.mlp(df, layers = c(4, 5, 4, 3), maxIter = 10, seed = 10)
   mlpPredictions <- collect(select(predict(model, mlpTestDF), "prediction"))
   expect_equal(head(mlpPredictions$prediction, 12), c(1, 1, 1, 1, 2, 1, 2, 2, 1, 0, 0, 1))
+
+  # test initialWeights
+  model <- spark.mlp(df, layers = c(4, 3), maxIter = 2, initialWeights =
+    c(0, 0, 0, 0, 0, 5, 5, 5, 5, 5, 9, 9, 9, 9, 9))
```
felixcheung: if this should be numeric/double, could you add a test with weight as numeric?

R/pkg/R/mllib.R
```diff
@@ -695,7 +696,7 @@ setMethod("predict", signature(object = "KMeansModel"),
 #' @note spark.mlp since 2.1.0
```
felixcheung: could you update the example with weights

R/pkg/R/mllib.R
WeichenXu123: scala-side wrapper need Array[Double] param so use as.array here, like other parameters such `layers`.

R/pkg/R/mllib.R
WeichenXu123: OK.

mllib/src/main/scala/org/apache/spark/ml/r/MultilayerPerceptronClassifierWrapper.scala
```diff
@@ -73,6 +75,8 @@ private[r] object MultilayerPerceptronClassifierWrapper
       .setStepSize(stepSize)
       .setPredictionCol(PREDICTED_LABEL_COL)
     if (seed != null && seed.length > 0) mlp.setSeed(seed.toInt)
+    if (initialWeights != null) mlp.setInitialWeights(Vectors.dense(initialWeights))
```
felixcheung: that seems somewhat involved, so it is likely better to not duplicate the logic on the R side.
we should check here if initialWeights != null its length > 0

WeichenXu123:
@felixcheung
updated following places:
1. clarify `initialWeights` parameter should be a vector
2. add require check in the mlp scala wrapper that if initialWeights != null its length > 0
3. update the example with initial weights

felixcheung:
LGTM.

felixcheung:
merged to master.

PR #15555, opened 2016-10-19
[Minor][ML] Refactor clustering summary.

yanboliang:

Abstract `ClusteringSummary` from `KMeansSummary`, `GaussianMixtureSummary` and `BisectingSummary`, and eliminate duplicated pieces of code.
## How was this patch tested?

Existing tests.

### DIFF
--- a/mllib/src/main/scala/org/apache/spark/ml/clustering/GaussianMixture.scala
+++ b/mllib/src/main/scala/org/apache/spark/ml/clustering/GaussianMixture.scala
@@ -365,33 +365,20 @@ object GaussianMixture extends DefaultParamsReadable[GaussianMixture] {
 @Since("2.0.0")
 @Experimental
 class GaussianMixtureSummary private[clustering] (
-    @Since("2.0.0") @transient val predictions: DataFrame,
-    @Since("2.0.0") val predictionCol: String,
-    @Since("2.0.0") val probabilityCol: String,
-    @Since("2.0.0") val featuresCol: String,
-    @Since("2.0.0") val k: Int) extends Serializable {
-
-  /**
-   * Cluster centers of the transformed data.
-   */
-  @Since("2.0.0")
-  @transient lazy val cluster: DataFrame = predictions.select(predictionCol)
+    predictions: DataFrame,
+    predictionCol: String,
+    val probabilityCol: String,
+    featuresCol: String,
+    k: Int)
+  extends ClusteringSummary (
+    predictions,
+    predictionCol,
+    featuresCol,
+    k) {
 
   /**
    * Probability of each cluster.
    */
   @Since("2.0.0")
   @transient lazy val probability: DataFrame = predictions.select(probabilityCol)
-
-  /**
-   * Size of (number of data points in) each cluster.
-   */
-  @Since("2.0.0")
-  lazy val clusterSizes: Array[Long] = {
-    val sizes = Array.fill[Long](k)(0)
-    cluster.groupBy(predictionCol).count().select(predictionCol, "count").collect().foreach {
-      case Row(cluster: Int, count: Long) => sizes(cluster) = count
-    }
-    sizes
-  }
 }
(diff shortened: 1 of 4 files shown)


yanboliang:
cc @zhengruifeng @srowen

mllib/src/main/scala/org/apache/spark/ml/clustering/KMeans.scala
```diff
@@ -354,21 +354,41 @@ object KMeans extends DefaultParamsReadable[KMeans] {
...
+/**
+ * :: Experimental ::
+ * Summary of clustering.
+ *
+ * @param predictions  [[DataFrame]] produced by model.transform()
+ * @param predictionCol  Name for column of predicted clusters in `predictions`
+ * @param featuresCol  Name for column of features in `predictions`
+ * @param k  Number of clusters
+ */
+@Experimental
```
zhengruifeng: what about adding `@Since("2.1.0")` here? 
Create a new scala file named `Clustering.scala` and move `ClusteringSummary` into it?

yanboliang: ```ClusteringSummary``` will be succeeded by summaries who were added in different version, so I think we should not add since version here. To the issue for a new file, I think ```ClusteringSummary``` is a small class, we can place it here temporarily.

yanboliang:
@srowen Would you mind to have a look when you available? Thanks.

mllib/src/main/scala/org/apache/spark/ml/clustering/KMeans.scala
sethah: I'm not entirely certain on the official policy for the `@Since` tags, but it seems better to me to put `@Since("2.1.0")` here for the class and the methods. It will be correct for some and will at least not be incorrect for others. I'm not positive though.

mllib/src/main/scala/org/apache/spark/ml/clustering/KMeans.scala
```diff
@@ -354,21 +354,41 @@ object KMeans extends DefaultParamsReadable[KMeans] {
...
+  extends ClusteringSummary (
+    predictions,
+    predictionCol,
+    featuresCol,
+    k
+  )
+
+/**
+ * :: Experimental ::
+ * Summary of clustering.
```
sethah: "Summary of clustering algorithms." ?

sethah (commented):
Left only minor comments.

yanboliang:
jenkins test this please

zhengruifeng:
Looks good to me.

sethah:
@yanboliang What do you think about adding the `@Since` tags?

jkbradley:
I'd say add Since tags wherever applicable in concrete classes.  But if it's an abstract method, we probably should not since they will be incorrect for new child classes later on.

mllib/src/main/scala/org/apache/spark/ml/clustering/GaussianMixture.scala
```diff
@@ -356,42 +356,25 @@ object GaussianMixture extends DefaultParamsReadable[GaussianMixture] {
...
-    @Since("2.0.0") val k: Int) extends Serializable {
-
-  /**
-   * Cluster centers of the transformed data.
-   */
-  @Since("2.0.0")
-  @transient lazy val cluster: DataFrame = predictions.select(predictionCol)
+    predictions: DataFrame,
+    predictionCol: String,
+    val probabilityCol: String,
```
jkbradley: could do a Since tag here

mllib/src/main/scala/org/apache/spark/ml/clustering/KMeans.scala
```diff
@@ -346,29 +346,43 @@ object KMeans extends DefaultParamsReadable[KMeans] {
...
+ * :: Experimental ::
+ * Summary of clustering algorithms.
+ *
+ * @param predictions  [[DataFrame]] produced by model.transform().
+ * @param predictionCol  Name for column of predicted clusters in `predictions`.
+ * @param featuresCol  Name for column of features in `predictions`.
+ * @param k  Number of clusters.
+ */
+@Experimental
+class ClusteringSummary private[clustering] (
```
jkbradley: If this is generic to clustering, how about putting it in a new file?

jkbradley:
LGTM
I'll merge this with master
Thanks!

PR #15556, opened 2016-10-19
[SPARK-18010][Core] Reduce work performed for building up the application list for the History Server app list UI page

vijoshi:

allow ReplayListenerBus to skip deserialising and replaying certain events using an inexpensive check of the event log entry. Use this to ensure that when event log replay is triggered for building the application list, we get the ReplayListenerBus to skip over all but the few events needed for our immediate purpose. Refer [SPARK-18010] for the motivation behind this change.
## How was this patch tested?

Tested with existing HistoryServer and ReplayListener unit test suites. All tests pass.

### DIFF
--- a/core/src/main/scala/org/apache/spark/deploy/history/FsHistoryProvider.scala
+++ b/core/src/main/scala/org/apache/spark/deploy/history/FsHistoryProvider.scala
@@ -81,6 +81,10 @@ private[history] class FsHistoryProvider(conf: SparkConf, clock: Clock)
 
   private val SPARK_HISTORY_FS_NUM_REPLAY_THREADS = "spark.history.fs.numReplayThreads"
 
+  private val AppStartEvent = "{\"Event\":\"SparkListenerApplicationStart\""
+
+  private val AppEndEvent = "{\"Event\":\"SparkListenerApplicationEnd\""
+
   // Interval between safemode checks.
   private val SAFEMODE_CHECK_INTERVAL_S = conf.getTimeAsSeconds(
     "spark.history.fs.safemodeCheck.interval", "5s")
@@ -399,14 +403,14 @@ private[history] class FsHistoryProvider(conf: SparkConf, clock: Clock)
     }
   }
 
-
   /**
    * Replay the log files in the list and merge the list of old applications with new ones
    */
   private def mergeApplicationListing(fileStatus: FileStatus): Unit = {
     val newAttempts = try {
         val bus = new ReplayListenerBus()
-        val res = replay(fileStatus, bus)
+        val res = replay(fileStatus, bus, (eventString => eventString.startsWith(AppStartEvent)
+          || eventString.startsWith(AppEndEvent)))
         res match {
           case Some(r) => logDebug(s"Application log ${r.logPath} loaded successfully: $r")
           case None => logWarning(s"Failed to load application log ${fileStatus.getPath}. " +
@@ -557,7 +561,8 @@ private[history] class FsHistoryProvider(conf: SparkConf, clock: Clock)
    */
   private def replay(
       eventLog: FileStatus,
-      bus: ReplayListenerBus): Option[FsApplicationAttemptInfo] = {
+      bus: ReplayListenerBus,
+      eventsFilter: (String) => Boolean = (s: String) => true): Option[FsApplicationAttemptInfo] = {
     val logPath = eventLog.getPath()
     logInfo(s"Replaying log path: $logPath")
     // Note that the eventLog may have *increased* in size since when we grabbed the filestatus,
@@ -571,7 +576,7 @@ private[history] class FsHistoryProvider(conf: SparkConf, clock: Clock)
       val appListener = new ApplicationEventListener
       val appCompleted = isApplicationCompleted(eventLog)
       bus.addListener(appListener)
-      bus.replay(logInput, logPath.toString, !appCompleted)
+      bus.replay(logInput, logPath.toString, !appCompleted, eventsFilter)
 
       // Without an app ID, new logs will render incorrectly in the listing page, so do not list or
       // try to show their UI.
--- a/core/src/main/scala/org/apache/spark/deploy/history/FsHistoryProvider.scala
+++ b/core/src/main/scala/org/apache/spark/deploy/history/FsHistoryProvider.scala
@@ -77,10 +77,6 @@ private[history] class FsHistoryProvider(conf: SparkConf, clock: Clock)
 
   import FsHistoryProvider._
 
-  private val NOT_STARTED = "<Not Started>"
-
-  private val SPARK_HISTORY_FS_NUM_REPLAY_THREADS = "spark.history.fs.numReplayThreads"
-
   // Interval between safemode checks.
   private val SAFEMODE_CHECK_INTERVAL_S = conf.getTimeAsSeconds(
     "spark.history.fs.safemodeCheck.interval", "5s")
@@ -684,6 +680,10 @@ private[history] class FsHistoryProvider(conf: SparkConf, clock: Clock)
 
 private[history] object FsHistoryProvider {
   val DEFAULT_LOG_DIR = "file:/tmp/spark-events"
+
+  private val NOT_STARTED = "<Not Started>"
+
+  private val SPARK_HISTORY_FS_NUM_REPLAY_THREADS = "spark.history.fs.numReplayThreads"
 
   val APPL_START_EVENT_PREFIX = "{\"Event\":\"SparkListenerApplicationStart\""
 
--- a/core/src/main/scala/org/apache/spark/deploy/history/FsHistoryProvider.scala
+++ b/core/src/main/scala/org/apache/spark/deploy/history/FsHistoryProvider.scala
@@ -81,10 +81,6 @@ private[history] class FsHistoryProvider(conf: SparkConf, clock: Clock)
 
   private val SPARK_HISTORY_FS_NUM_REPLAY_THREADS = "spark.history.fs.numReplayThreads"
 
-  private val AppStartEvent = "{\"Event\":\"SparkListenerApplicationStart\""
-
-  private val AppEndEvent = "{\"Event\":\"SparkListenerApplicationEnd\""
-
   // Interval between safemode checks.
   private val SAFEMODE_CHECK_INTERVAL_S = conf.getTimeAsSeconds(
     "spark.history.fs.safemodeCheck.interval", "5s")
@@ -247,7 +243,7 @@ private[history] class FsHistoryProvider(conf: SparkConf, clock: Clock)
           val appListener = new ApplicationEventListener()
           replayBus.addListener(appListener)
           val appAttemptInfo = replay(fs.getFileStatus(new Path(logDir, attempt.logPath)),
-            replayBus)
+                                      replayBus, ReplayListenerBus.SELECT_ALL_FILTER)
           appAttemptInfo.map { info =>
             val uiAclsEnabled = conf.getBoolean("spark.history.ui.acls.enable", false)
             ui.getSecurityManager.setAcls(uiAclsEnabled)
@@ -409,8 +405,12 @@ private[history] class FsHistoryProvider(conf: SparkConf, clock: Clock)
   private def mergeApplicationListing(fileStatus: FileStatus): Unit = {
     val newAttempts = try {
         val bus = new ReplayListenerBus()
-        val res = replay(fileStatus, bus, (eventString => eventString.startsWith(AppStartEvent)
-          || eventString.startsWith(AppEndEvent)))
+        val res = replay(fileStatus, bus,
+                         { eventString =>
+          eventString.startsWith (APPL_START_EVENT_PREFIX) ||
+          eventString.startsWith (APPL_END_EVENT_PREFIX)
+                         })
+
         res match {
           case Some(r) => logDebug(s"Application log ${r.logPath} loaded successfully: $r")
           case None => logWarning(s"Failed to load application log ${fileStatus.getPath}. " +
@@ -562,7 +562,7 @@ private[history] class FsHistoryProvider(conf: SparkConf, clock: Clock)
   private def replay(
       eventLog: FileStatus,
       bus: ReplayListenerBus,
-      eventsFilter: (String) => Boolean = (s: String) => true): Option[FsApplicationAttemptInfo] = {
+      eventsFilter: (String) => Boolean): Option[FsApplicationAttemptInfo] = {
     val logPath = eventLog.getPath()
     logInfo(s"Replaying log path: $logPath")
     // Note that the eventLog may have *increased* in size since when we grabbed the filestatus,
@@ -682,6 +682,10 @@ private[history] class FsHistoryProvider(conf: SparkConf, clock: Clock)
 
 private[history] object FsHistoryProvider {
   val DEFAULT_LOG_DIR = "file:/tmp/spark-events"
+
+  val APPL_START_EVENT_PREFIX = "{\"Event\":\"SparkListenerApplicationStart\""
+
+  val APPL_END_EVENT_PREFIX = "{\"Event\":\"SparkListenerApplicationEnd\""
 }
 
 /**
--- a/core/src/main/scala/org/apache/spark/deploy/history/FsHistoryProvider.scala
+++ b/core/src/main/scala/org/apache/spark/deploy/history/FsHistoryProvider.scala
@@ -243,7 +243,7 @@ private[history] class FsHistoryProvider(conf: SparkConf, clock: Clock)
           val appListener = new ApplicationEventListener()
           replayBus.addListener(appListener)
           val appAttemptInfo = replay(fs.getFileStatus(new Path(logDir, attempt.logPath)),
-                                      replayBus, ReplayListenerBus.SELECT_ALL_FILTER)
+            replayBus, ReplayListenerBus.SELECT_ALL_FILTER)
           appAttemptInfo.map { info =>
             val uiAclsEnabled = conf.getBoolean("spark.history.ui.acls.enable", false)
             ui.getSecurityManager.setAcls(uiAclsEnabled)
@@ -405,11 +405,13 @@ private[history] class FsHistoryProvider(conf: SparkConf, clock: Clock)
   private def mergeApplicationListing(fileStatus: FileStatus): Unit = {
     val newAttempts = try {
         val bus = new ReplayListenerBus()
-        val res = replay(fileStatus, bus,
-                         { eventString =>
-          eventString.startsWith (APPL_START_EVENT_PREFIX) ||
-          eventString.startsWith (APPL_END_EVENT_PREFIX)
-                         })
+
+        val eventsFilter: (String => Boolean) = { eventString =>
+          eventString.startsWith(APPL_START_EVENT_PREFIX) ||
+            eventString.startsWith(APPL_END_EVENT_PREFIX)
+        }
+
+        val res = replay(fileStatus, bus, eventsFilter)
 
         res match {
           case Some(r) => logDebug(s"Application log ${r.logPath} loaded successfully: $r")
--- a/core/src/main/scala/org/apache/spark/deploy/history/FsHistoryProvider.scala
+++ b/core/src/main/scala/org/apache/spark/deploy/history/FsHistoryProvider.scala
@@ -238,20 +238,24 @@ private[history] class FsHistoryProvider(conf: SparkConf, clock: Clock)
             // Do not call ui.bind() to avoid creating a new server for each application
           }
 
-          val res = replay(fs.getFileStatus(new Path(logDir, attempt.logPath)),
-            replayBus)
+          val fileStatus = fs.getFileStatus(new Path(logDir, attempt.logPath))
 
-          res.map { case (_, envInfo) =>
+          val appListener = replay(fileStatus, isApplicationCompleted(fileStatus), replayBus)
+
+          if (appListener.appId.isDefined) {
             val uiAclsEnabled = conf.getBoolean("spark.history.ui.acls.enable", false)
             ui.getSecurityManager.setAcls(uiAclsEnabled)
             // make sure to set admin acls before view acls so they are properly picked up
-            ui.getSecurityManager.setAdminAcls(envInfo.adminAcls.getOrElse(""))
+            ui.getSecurityManager.setAdminAcls(appListener.adminAcls.getOrElse(""))
             ui.getSecurityManager.setViewAcls(attempt.sparkUser,
-              envInfo.viewAcls.getOrElse(""))
-            ui.getSecurityManager.setAdminAclsGroups(envInfo.adminAclsGroups.getOrElse(""))
-            ui.getSecurityManager.setViewAclsGroups(envInfo.viewAclsGroups.getOrElse(""))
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/apache/spark/deploy/history/FsHistoryProvider.scala
```diff
@@ -81,6 +81,10 @@ private[history] class FsHistoryProvider(conf: SparkConf, clock: Clock)
 
   private val SPARK_HISTORY_FS_NUM_REPLAY_THREADS = "spark.history.fs.numReplayThreads"
 
+  private val AppStartEvent = "{\"Event\":\"SparkListenerApplicationStart\""
```
srowen: Nit: use `CONSTANT_CASE` and maybe call it a `_PREFIX`? that's just my taste.

srowen: These should really be in a companion object anyway, including existing constants.

core/src/main/scala/org/apache/spark/deploy/history/FsHistoryProvider.scala
```diff
@@ -557,7 +561,8 @@ private[history] class FsHistoryProvider(conf: SparkConf, clock: Clock)
    */
   private def replay(
       eventLog: FileStatus,
-      bus: ReplayListenerBus): Option[FsApplicationAttemptInfo] = {
+      bus: ReplayListenerBus,
+      eventsFilter: (String) => Boolean = (s: String) => true): Option[FsApplicationAttemptInfo] = {
```
srowen: Does this need to be parameterized? it seems like this method always can apply the filter you pass here. It's called two places.

srowen (changes requested):

core/src/main/scala/org/apache/spark/deploy/history/FsHistoryProvider.scala
vijoshi: Did you mean this way ?

```
  private def replay(
      eventLog: FileStatus,
      bus: ReplayListenerBus,
      eventsFilter: (String) => Boolean): Option[FsApplicationAttemptInfo] = {
```

and have both the places that we call into it provide a filter:

```
val appAttemptInfo = replay(fs.getFileStatus(new Path(logDir, attempt.logPath)),
            replayBus, (s)=> true)
```

and 

```
        val res = replay(fileStatus, bus, (eventString => eventString.startsWith(AppStartEvent)
          || eventString.startsWith(AppEndEvent)))
```

core/src/main/scala/org/apache/spark/scheduler/ReplayListenerBus.scala
```diff
@@ -43,19 +43,25 @@ private[spark] class ReplayListenerBus extends SparkListenerBus with Logging {
...
   def replay(
       logData: InputStream,
       sourceName: String,
-      maybeTruncated: Boolean = false): Unit = {
+      maybeTruncated: Boolean = false,
+      eventsFilter: (String) => Boolean = (s: String) => true): Unit = {
     var currentLine: String = null
     var lineNumber: Int = 1
     try {
       val lines = Source.fromInputStream(logData).getLines()
```
srowen: Ah right. Still it seems like a case where `zipWithIndex` would be easier, followed by `filter`? While we're at it here.

core/src/main/scala/org/apache/spark/deploy/history/FsHistoryProvider.scala
vijoshi: Moved the new constants to a companion object. did you want the existing unrelated ones moved into it as well?

core/src/main/scala/org/apache/spark/deploy/history/FsHistoryProvider.scala
```diff
@@ -399,14 +399,18 @@ private[history] class FsHistoryProvider(conf: SparkConf, clock: Clock)
...
    */
   private def mergeApplicationListing(fileStatus: FileStatus): Unit = {
     val newAttempts = try {
         val bus = new ReplayListenerBus()
-        val res = replay(fileStatus, bus)
+        val res = replay(fileStatus, bus,
+                         { eventString =>
+          eventString.startsWith (APPL_START_EVENT_PREFIX) ||
+          eventString.startsWith (APPL_END_EVENT_PREFIX)
+                         })
```
ajbozarth: Though this passes scalastyle I think this could be styled cleaner

ajbozarth (commented):
I've checked out your code and it looks good and works well, huge speed up. A couple of nits below but otherwise I really like this, it solved the speed issue without added/updated metadata. I'd like to see some more committer feedback but overall LGTM.

core/src/main/scala/org/apache/spark/deploy/history/FsHistoryProvider.scala
srowen: No, I mean it looked like the method would never need to use any filter except for the start/end event one. So it doesn't need it as a parameter. That is, why is it up to the caller? the method filters in what it needs to produce the info it needs. I could be missing something.

core/src/main/scala/org/apache/spark/scheduler/ReplayListenerBus.scala
srowen: I don't think that logic changes. It already doesn't actually know if there are more lines in the file, because the error might cause the iterator to close and not indicate any more lines available. It's pretty best-effort. I guess the scenario is: error occurs, and there are no more matching lines, but there are some unmatching lines available? Now we'd continue instead of fail, which seems equally reasonable given the purpose.

core/src/main/scala/org/apache/spark/deploy/history/FsHistoryProvider.scala
vijoshi: oh ok - this internal replay helper method is called in two places in FsHistoryProvider. The two places  want different filtering - one wants all events replayed (call from `getAppUI()` ), the other (call from `mergeApplicationListing`) just the app start/end so they need to pass in the respective filters.

core/src/main/scala/org/apache/spark/deploy/history/FsHistoryProvider.scala
vijoshi: Done

core/src/main/scala/org/apache/spark/scheduler/ReplayListenerBus.scala
vijoshi: Okay, agree - updated the replay implementation

core/src/main/scala/org/apache/spark/scheduler/ReplayListenerBus.scala
```diff
@@ -43,38 +43,56 @@ private[spark] class ReplayListenerBus extends SparkListenerBus with Logging {
...
+            throw e // quit processing the event log
         }
-        lineNumber += 1
       }
     } catch {
       case ioe: IOException =>
         throw ioe
       case e: Exception =>
         logError(s"Exception parsing Spark event log: $sourceName", e)
-        logError(s"Malformed line #$lineNumber: $currentLine\n")
```
steveloughran: might still be useful to include the line number in there, if relevant. Not that it helps debug things much: if your file is corrupt then it will take effort to fix

vijoshi: note that if the Exception occurs when iterating over lines, then I've added code  to the inner catch( ) block to log out the line number etc. had to move things around for the way due to the way i had to refactor this method. now when we get to this outer catch block:
- either the line number has already been logged or,
- there was a problem building the iterator over the event stream in which case there is no line number / current line to display.

tgravescs:
So this is still reading the entire file, its just filtering out just the start and end log entries?   so basically  the listener bus only replays those events vs all of them. 

Have you run this on history files with GBs?  I'll give it a try here when I have a chance and do a further review.

If it a good speedup without having the metadata separate its definitely good  to have an if needed we can always do the separate metadata later.

srowen:
That's right, it still reads the whole file but skips posting most of it.

vijoshi:
@tgravescs I have tried it with a 150MB file and speedup is very noticeable (refer the spark issue for results I see). I have noticed the JSON deserialization is particularly slow, while simply running through the file not so much. I would expect proportionate benefit on larger files wherever reading the file/event stream is much less expensive than the Json deserialization + replay.

tgravescs:
I tried this out on one of our larger clusters. Normally history UI shows around 4500-5000 applications and takes a long time to load all of them.  I don't have an exact time but I'm sure its over an hour.  We have history files that can be 10's of GBs.

It is loading noticeably faster but to me its still not as good as it could be for just listing the apps.  It took 30 minutes for it to create the entire listing.  That is the initial startup.

anyway I'm fine with it as its definitely an improvement but I think we should leave the other one open.   Or if no one else has this problem leave it closed and I'll reopen when I have time to work on it. I won't have time to code review in detail today but if others have already you don't have to wait for me.

ajbozarth:
@tgravescs I've been looking into (and will continue to) a metadata solution for this, but for now I think this fixes the "Bug" that was reported. I'd consider any metadata update an "improvement" at this point and a new JIRA could be opened for it at that time.

Also @srowen and @tgravescs, a side note, I've noticed that work done on the History Server is usually labeled as [Core] not [Web UI], is there a reason we don't consider the History Server part of the Web UI on the code side? (It can make tracking UI related JIRAs and PRs a little difficult)

vanzin:
ok to test

vanzin:
>  is there a reason we don't consider the History Server part of the Web UI 

It would probably be better to have a separate component altogether for it.

core/src/main/scala/org/apache/spark/scheduler/ReplayListenerBus.scala
```diff
@@ -43,38 +43,56 @@ private[spark] class ReplayListenerBus extends SparkListenerBus with Logging {
...
+      maybeTruncated: Boolean = false,
+      eventsFilter: (String) => Boolean = ReplayListenerBus.SELECT_ALL_FILTER): Unit = {
     try {
-      val lines = Source.fromInputStream(logData).getLines()
-      while (lines.hasNext) {
-        currentLine = lines.next()
+      val lineEntries = Source.fromInputStream(logData)
+        .getLines()
+        .zipWithIndex
+        .filter(entry => eventsFilter(entry._1))
```
vanzin: nit: `..filter { case (line, _) => eventsFilter(line) }`

core/src/main/scala/org/apache/spark/scheduler/ReplayListenerBus.scala
```diff
@@ -43,38 +43,56 @@ private[spark] class ReplayListenerBus extends SparkListenerBus with Logging {
...
+
+      var entry: (String, Int) = ("", 0)
+
+      while (lineEntries.hasNext) {
         try {
-          postToAll(JsonProtocol.sparkEventFromJson(parse(currentLine)))
+          entry = lineEntries.next()
```
vanzin: nit: `val (line, lineno) = lineEntries.next()`

core/src/main/scala/org/apache/spark/scheduler/ReplayListenerBus.scala
```diff
@@ -43,38 +43,56 @@ private[spark] class ReplayListenerBus extends SparkListenerBus with Logging {
...
+            // as such in a best effort to replay the given input
+            if (!maybeTruncated || lineEntries.hasNext) {
               throw jpe
             } else {
               logWarning(s"Got JsonParseException from log file $sourceName" +
-                s" at line $lineNumber, the file might not have finished writing cleanly.")
+                s" at line number ${entry._2}, the file might not have finished writing cleanly.")
             }
+
+          case e: Exception =>
```
vanzin: This exception handler looks weird. It's logging then re-throwing the exception, which is later caught again and another log message is written for it. So this can probably just be removed.

core/src/main/scala/org/apache/spark/scheduler/ReplayListenerBus.scala
```diff
@@ -43,38 +43,56 @@ private[spark] class ReplayListenerBus extends SparkListenerBus with Logging {
...
         logError(s"Exception parsing Spark event log: $sourceName", e)
-        logError(s"Malformed line #$lineNumber: $currentLine\n")
     }
   }
+}
+
+private[spark] object ReplayListenerBus {
 
+  // utility filter that selects all event logs during replay
+  val SELECT_ALL_FILTER = (eventString: String) => true
```
vanzin: Instead of this I have a slight preference for an overloaded `replay` method that does not take a filter. Or a default value for the filter argument.

ajbozarth: Having this as a default value is what I was trying to get at with some of my previous comments as well

vanzin:
BTW I agree with Tom that SPARK-6951 should remain open since this issue helps, but does not fix the issue completely.

core/src/main/scala/org/apache/spark/scheduler/ReplayListenerBus.scala
vijoshi: needed to have the line number at which failure happened, since we had that in the earlier implementation. the line/line no values are not in scope in the outer catch so can't get them logged there. open for a better way to do this if any.

vijoshi: changed it anyway to remove that exception handler.

core/src/main/scala/org/apache/spark/scheduler/ReplayListenerBus.scala
vijoshi: updated with latest review comments.

core/src/main/scala/org/apache/spark/deploy/history/FsHistoryProvider.scala
```diff
@@ -557,7 +560,8 @@ private[history] class FsHistoryProvider(conf: SparkConf, clock: Clock)
    */
   private def replay(
       eventLog: FileStatus,
-      bus: ReplayListenerBus): Option[FsApplicationAttemptInfo] = {
+      bus: ReplayListenerBus,
+      eventsFilter: ReplayEventsFilter = SELECT_ALL_FILTER): Option[FsApplicationAttemptInfo] = {
```
vijoshi: @srowen while you're right the return value could be empty if the filter specified filtered out the app start/end events, but then I also see that ideally the way to parse out data from replayed events is for callers to set appropriate listeners of their choice on the ReplayListenerBus instance specified for event replay. And the listeners capture event data etc. This particular replay () method in FsHistoryProvider looks like a helper for the two call sites and not much more, so maybe not so much of a problem?

srowen: I think I agree, but then, the other call site actually uses the return value of this method, but, also supplies a listener/bus to collect other info, and then also parses that. It seems like this method could simply do all that work for both callers and not even need an external bus?

vijoshi: This helper replay method will need to be supplied a listener bus externally because the two call sites need very different sets of listeners on the bus. Only thing it looked we could do was enrich the return value from the helper to prevent the second callsite (getAppUI() ) from having to add and parse from an additional listener of its own. I've pushed the updated commit.

core/src/main/scala/org/apache/spark/deploy/history/FsHistoryProvider.scala
```diff
@@ -399,20 +396,26 @@ private[history] class FsHistoryProvider(conf: SparkConf, clock: Clock)
...
-            "The application may have not started.")
+
+        val eventsFilter: ReplayEventsFilter = { eventString =>
+          eventString.startsWith(APPL_START_EVENT_PREFIX) ||
+            eventString.startsWith(APPL_END_EVENT_PREFIX)
+        }
+
+        val res = replay(fileStatus, bus, eventsFilter)
+
+        res.map {
```
srowen: Is the idea here to try to get rid of some of the redundancy, because the `None` case is kind of addressed twice? 

If it doesn't work to just `return` from the existing `None` case, that's simpler. Otherwise, pull up the check? to before this line? otherwise no real point in mapping nothing.

And then: map should be foreach, and there's no need to return a value in the statement below

core/src/main/scala/org/apache/spark/deploy/history/FsHistoryProvider.scala
```diff
@@ -399,20 +396,26 @@ private[history] class FsHistoryProvider(conf: SparkConf, clock: Clock)
     }
   }
 
-
   /**
    * Replay the log files in the list and merge the list of old applications with new ones
    */
   private def mergeApplicationListing(fileStatus: FileStatus): Unit = {
     val newAttempts = try {
         val bus = new ReplayListenerBus()
```
srowen: Do you even need a reference to this here now? can just pass `new ReplayListenerBus()` to the replay method

core/src/main/scala/org/apache/spark/deploy/history/FsHistoryProvider.scala
```diff
@@ -677,6 +693,14 @@ private[history] class FsHistoryProvider(conf: SparkConf, clock: Clock)
 
 private[history] object FsHistoryProvider {
   val DEFAULT_LOG_DIR = "file:/tmp/spark-events"
+
+  private val NOT_STARTED = "<Not Started>"
+
+  private val SPARK_HISTORY_FS_NUM_REPLAY_THREADS = "spark.history.fs.numReplayThreads"
+
+  val APPL_START_EVENT_PREFIX = "{\"Event\":\"SparkListenerApplicationStart\""
```
srowen: Can these be private? not 100% sure

core/src/main/scala/org/apache/spark/scheduler/ReplayListenerBus.scala
```diff
@@ -43,38 +44,62 @@ private[spark] class ReplayListenerBus extends SparkListenerBus with Logging {
...
-      while (lines.hasNext) {
-        currentLine = lines.next()
+      val lineEntries = Source.fromInputStream(logData)
+        .getLines()
+        .zipWithIndex
+        .filter { case (line, _) => eventsFilter(line) }
+
+      while (lineEntries.hasNext) {
         try {
+          val entry = lineEntries.next()
```
srowen: Echoing Marcelo's comment that this can be `val (currentLine, lineNumber) = ...`, I believe.

srowen (changes requested):

core/src/main/scala/org/apache/spark/deploy/history/FsHistoryProvider.scala
```diff
@@ -240,19 +237,19 @@ private[history] class FsHistoryProvider(conf: SparkConf, clock: Clock)
...
           }
-          val appListener = new ApplicationEventListener()
-          replayBus.addListener(appListener)
-          val appAttemptInfo = replay(fs.getFileStatus(new Path(logDir, attempt.logPath)),
+
+          val res = replay(fs.getFileStatus(new Path(logDir, attempt.logPath)),
             replayBus)
-          appAttemptInfo.map { info =>
+
+          res.map { case (attemptInfo, envInfo) =>
```
vanzin: nit: seems like `attemptInfo` is not used, so it should be `_`.

core/src/main/scala/org/apache/spark/scheduler/ReplayListenerBus.scala
```diff
@@ -43,38 +44,62 @@ private[spark] class ReplayListenerBus extends SparkListenerBus with Logging {
...
             // We can only ignore exception from last line of the file that might be truncated
-            if (!maybeTruncated || lines.hasNext) {
+            // the last entry may not be the very last line in the event log, but we treat it
+            // as such in a best effort to replay the given input
+            if (!maybeTruncated || lineEntries.hasNext) {
               throw jpe
             } else {
               logWarning(s"Got JsonParseException from log file $sourceName" +
-                s" at line $lineNumber, the file might not have finished writing cleanly.")
+                s" at line ${lineNumber + 1}, the file might not have finished writing cleanly.")
```
vanzin: Since you're adjusting line number every time you use it, might as well do it when you assign the variable. (Also because without the adjustment the variable name is wrong - it's the 0-based index, not the "line number".)

core/src/main/scala/org/apache/spark/deploy/history/FsHistoryProvider.scala
```diff
@@ -422,6 +425,8 @@ private[history] class FsHistoryProvider(conf: SparkConf, clock: Clock)
       }
 
     if (newAttempts.isEmpty) {
+      logWarning(s"Failed to load application log ${fileStatus.getPath}. " +
```
vanzin: I think this was clearer where it was before. (Which might be the reason for Sean's other comment.)

In fact, I think this log message is scarier than it is useful. An application without an ID is a valid case (it may be still starting up), so printing warnings here is a little misleading.

core/src/main/scala/org/apache/spark/deploy/history/FsHistoryProvider.scala
```diff
@@ -554,10 +559,15 @@ private[history] class FsHistoryProvider(conf: SparkConf, clock: Clock)
...
+   *
+   * Depending on the eventsFilter, return values may not be populated or get returned as `None`
+   * because the events carrying the needed data are filtered out and therefore not replayed.
    */
   private def replay(
       eventLog: FileStatus,
-      bus: ReplayListenerBus): Option[FsApplicationAttemptInfo] = {
+      bus: ReplayListenerBus,
+      eventsFilter: ReplayEventsFilter = SELECT_ALL_FILTER)
+  : (Option[(FsApplicationAttemptInfo, FsApplicationEnvACLInfo)]) = {
```
vanzin: What about returning the `ApplicationEventListener` instead of creating this new type (`FsApplicationEnvACLInfo`) for the info that exists there?

vanzin (commented):
A few minor things left to take care of.

core/src/main/scala/org/apache/spark/deploy/history/FsHistoryProvider.scala
vijoshi: given that the other value returned is an FsApplicationAttemptInfo. For a replay, the ApplicationEventListener is always there, the  FsApplicationAttemptInfo may or may not get generated. So the return value would probably have to be (Option[FsApplicationAttemptInfo], ApplicationEventListener) - which could get awkward to deal with ?  i.e you couldn't simply say that the method returned an Option[(FsApplicationAttemptInfo, ApplicationEventListener)]

srowen (approved):
I'm OK with it if @vanzin is

vanzin:
LGTM. Merging to master.

vijoshi:
@vanzin @srowen @ajbozarth - can this go into branch-2.0 as well? is there a deciding criteria? i can open a pull request.

srowen:
Hm, usually it's more for important or small fixes. This is a moderate performance improvement, in a place it's important. I don't think it changes behavior. I'd be OK back-porting it. Anyone else?

vanzin:
If it's a clean backport I'm ok with it.

vijoshi:
@vanzin @srowen - thanks, i've opened pull request #15655 on branch-2.0

PR #15557, opened 2016-10-19
[SPARK-17944] [DEPLOY] sbin/start-* scripts use of `hostname -f` fail with Solaris

JnyJny:

Modify sbin/start-master.sh, sbin/start-mesos-dispatcher.sh and sbin/start-slaves.sh to use the output of 'uname' to select which OS-specific command-line is used to determine the host's fully qualified host name.  
## How was this patch tested?

Tested by hand; starting on Solaris, Linux and macOS.

### DIFF
--- a/sbin/start-master.sh
+++ b/sbin/start-master.sh
@@ -48,7 +48,14 @@ if [ "$SPARK_MASTER_PORT" = "" ]; then
 fi
 
 if [ "$SPARK_MASTER_HOST" = "" ]; then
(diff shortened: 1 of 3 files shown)


srowen:
Jenkins test this please

srowen (approved):
That seems OK to me, in that it will only affect Solaris anyway, and the bash formulation looks OK.

srowen:
Merged to master

PR #15562, opened 2016-10-20
[SPARK-18021][SQL] Refactor file name specification for data sources

rxin:

Currently each data source OutputWriter is responsible for specifying the entire file name for each file output. This, however, does not make any sense because we rely on file naming schemes for certain behaviors in Spark SQL, e.g. bucket id. The current approach allows individual data sources to break the implementation of bucketing.

On the flip side, we also don't want to move file naming entirely out of data sources, because different data sources do want to specify different extensions.

This patch divides file name specification into two parts: the first part is a prefix specified by the caller of OutputWriter (in WriteOutput), and the second part is the suffix that can be specified by the OutputWriter itself. Note that a side effect of this change is that now all file based data sources also support bucketing automatically.

There are also some other minor cleanups:
- Removed the UUID passed through generic Configuration string
- Some minor rewrites for better clarity
- Renamed "path" in multiple places to "stagingDir", to more accurately reflect its meaning
## How was this patch tested?

This should be covered by existing data source tests.


rxin:
cc @liancheng @cloud-fan @tejasapatil 

also @marmbrus -- you have always wanted this to happen.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetOutputWriter.scala
```diff
@@ -122,47 +122,29 @@ private[parquet] class ParquetOutputWriterFactory(
...
-    bucketId: Option[Int],
+    stagingDir: String,
+    fileNamePrefix: String,
     context: TaskAttemptContext)
   extends OutputWriter {
 
   private val recordWriter: RecordWriter[Void, InternalRow] = {
     val outputFormat = {
       new ParquetOutputFormat[InternalRow]() {
-        // Here we override `getDefaultWorkFile` for two reasons:
```
cloud-fan: why remove this comment?

rxin: Basically the only contract now is that prefix needs to be enforced, and it is not the concern of these classes to think about dynamic partitioning or appending.

cloud-fan:
LGTM

liancheng (commented):
LGTM pending Jenkins. Left one minor comment though.

rxin:
Alright merging this in master. I will submit more patches to continue this work.

PR #15563, opened 2016-10-20
[SPARK-16759][CORE] Add a configuration property to pass caller contexts of upstream applications into Spark

weiqingy:

Many applications take Spark as a computing engine and run on it. This PR adds a configuration property `spark.log.callerContext` that can be used by Spark's upstream applications (e.g. Oozie) to set up their caller contexts into Spark. In the end, Spark will combine its own caller context with the caller contexts of its upstream applications, and write them into Yarn RM log and HDFS audit log.

The audit log has a config to truncate the caller contexts passed in (default 128). The caller contexts will be sent over rpc, so it should be concise. The call context written into HDFS log and Yarn log consists of two parts: the information `A` specified by Spark itself and the value `B` of `spark.log.callerContext` property.  Currently `A` typically takes 64 to 74 characters,  so `B` can have up to 50 characters (mentioned in the doc `running-on-yarn.md`)
## How was this patch tested?

Manual tests. I have run some Spark applications with `spark.log.callerContext` configuration in Yarn client/cluster mode, and verified that the caller contexts were written into Yarn RM log and HDFS audit log correctly.

The ways to configure `spark.log.callerContext` property:
- In spark-defaults.conf:

```
spark.log.callerContext  infoSpecifiedByUpstreamApp
```
- In app's source code:

```
val spark = SparkSession
      .builder
      .appName("SparkKMeans")
      .config("spark.log.callerContext", "infoSpecifiedByUpstreamApp")
      .getOrCreate()
```

When running on Spark Yarn cluster mode, the driver is unable to pass 'spark.log.callerContext' to Yarn client and AM since Yarn client and AM have already started before the driver performs `.config("spark.log.callerContext", "infoSpecifiedByUpstreamApp")`.

The following  example shows the command line used to submit a SparkKMeans application and the corresponding records in Yarn RM log and HDFS audit log.

Command:

```
./bin/spark-submit --verbose --executor-cores 3 --num-executors 1 --master yarn --deploy-mode client --class org.apache.spark.examples.SparkKMeans examples/target/original-spark-examples_2.11-2.1.0-SNAPSHOT.jar hdfs://localhost:9000/lr_big.txt 2 5
```

Yarn RM log:

screen shot 2016-10-19 at 9 12 03 pm

HDFS audit log:

screen shot 2016-10-19 at 10 18 14 pm

### DIFF
--- a/core/src/main/scala/org/apache/spark/deploy/SparkSubmit.scala
+++ b/core/src/main/scala/org/apache/spark/deploy/SparkSubmit.scala
@@ -462,6 +462,8 @@ object SparkSubmit {
       OptionAssigner(args.archives, YARN, ALL_DEPLOY_MODES, sysProp = "spark.yarn.dist.archives"),
       OptionAssigner(args.principal, YARN, ALL_DEPLOY_MODES, sysProp = "spark.yarn.principal"),
       OptionAssigner(args.keytab, YARN, ALL_DEPLOY_MODES, sysProp = "spark.yarn.keytab"),
+      OptionAssigner(args.callerContext, YARN, ALL_DEPLOY_MODES,
+        sysProp = "spark.hadoop.callerContext"),
 
       // Other options
       OptionAssigner(args.executorCores, STANDALONE | YARN, ALL_DEPLOY_MODES,
--- a/core/src/main/scala/org/apache/spark/deploy/SparkSubmit.scala
+++ b/core/src/main/scala/org/apache/spark/deploy/SparkSubmit.scala
@@ -462,8 +462,6 @@ object SparkSubmit {
       OptionAssigner(args.archives, YARN, ALL_DEPLOY_MODES, sysProp = "spark.yarn.dist.archives"),
       OptionAssigner(args.principal, YARN, ALL_DEPLOY_MODES, sysProp = "spark.yarn.principal"),
       OptionAssigner(args.keytab, YARN, ALL_DEPLOY_MODES, sysProp = "spark.yarn.keytab"),
-      OptionAssigner(args.callerContext, YARN, ALL_DEPLOY_MODES,
-        sysProp = "spark.hadoop.callerContext"),
 
       // Other options
       OptionAssigner(args.executorCores, STANDALONE | YARN, ALL_DEPLOY_MODES,
--- a/core/src/main/scala/org/apache/spark/internal/config/package.scala
+++ b/core/src/main/scala/org/apache/spark/internal/config/package.scala
@@ -210,4 +210,17 @@ package object config {
   private[spark] val APP_CALLER_CONTEXT = ConfigBuilder("spark.log.callerContext")
     .stringConf
     .createOptional
+
+  private[spark] val FILES_MAX_PARTITION_BYTES = ConfigBuilder("spark.files.maxPartitionBytes")
+    .doc("The maximum number of bytes to pack into a single partition when reading files.")
+    .longConf
+    .createWithDefault(128 * 1024 * 1024)
+
+  private[spark] val FILES_OPEN_COST_IN_BYTES = ConfigBuilder("spark.files.openCostInBytes")
+    .doc("The estimated cost to open a file, measured by the number of bytes could be scanned in" +
+      " the same time. This is used when putting multiple files into a partition. It's better to" +
+      " over estimate, then the partitions with small files will be faster than partitions with" +
+      " bigger files.")
+    .longConf
+    .createWithDefault(4 * 1024 * 1024)
 }
--- a/core/src/main/scala/org/apache/spark/internal/config/package.scala
+++ b/core/src/main/scala/org/apache/spark/internal/config/package.scala
@@ -206,4 +206,8 @@ package object config {
       "encountering corrupt files and contents that have been read will still be returned.")
     .booleanConf
     .createWithDefault(false)
+
+  private[spark] val APP_CALLER_CONTEXT = ConfigBuilder("spark.upstreamApp.callerContext")
+    .stringConf
+    .createOptional
 }
--- a/core/src/main/scala/org/apache/spark/internal/config/package.scala
+++ b/core/src/main/scala/org/apache/spark/internal/config/package.scala
@@ -207,7 +207,7 @@ package object config {
     .booleanConf
     .createWithDefault(false)
 
-  private[spark] val APP_CALLER_CONTEXT = ConfigBuilder("spark.upstreamApp.callerContext")
+  private[spark] val APP_CALLER_CONTEXT = ConfigBuilder("spark.log.callerContext")
     .stringConf
     .createOptional
 }
--- a/core/src/main/scala/org/apache/spark/scheduler/Task.scala
+++ b/core/src/main/scala/org/apache/spark/scheduler/Task.scala
@@ -26,6 +26,7 @@ import scala.collection.mutable.HashMap
 
 import org.apache.spark._
 import org.apache.spark.executor.TaskMetrics
+import org.apache.spark.internal.config.APP_CALLER_CONTEXT
 import org.apache.spark.memory.{MemoryMode, TaskMemoryManager}
 import org.apache.spark.metrics.MetricsSystem
 import org.apache.spark.serializer.SerializerInstance
@@ -94,7 +95,7 @@ private[spark] abstract class Task[T](
 
     new CallerContext(
       "TASK",
-      SparkEnv.get.conf.getOption("spark.hadoop.callerContext"),
+      SparkEnv.get.conf.get(APP_CALLER_CONTEXT),
       appId,
       appAttemptId,
       jobId,
--- a/core/src/main/scala/org/apache/spark/scheduler/Task.scala
+++ b/core/src/main/scala/org/apache/spark/scheduler/Task.scala
@@ -92,8 +92,16 @@ private[spark] abstract class Task[T](
       kill(interruptThread = false)
     }
 
-    new CallerContext("TASK", appId, appAttemptId, jobId, Option(stageId), Option(stageAttemptId),
-      Option(taskAttemptId), Option(attemptNumber)).setCurrentContext()
+    new CallerContext(
+      "TASK",
+      SparkEnv.get.conf.getOption("spark.hadoop.callerContext"),
+      appId,
+      appAttemptId,
+      jobId,
+      Option(stageId),
+      Option(stageAttemptId),
+      Option(taskAttemptId),
+      Option(attemptNumber)).setCurrentContext()
 
     try {
       runTask(context)
--- a/core/src/main/scala/org/apache/spark/util/Utils.scala
+++ b/core/src/main/scala/org/apache/spark/util/Utils.scala
@@ -2579,6 +2579,7 @@ private[util] object CallerContext extends Logging {
  */
 private[spark] class CallerContext(
    from: String,
+   upstreamCallerContext: Option[String] = None,
    appId: Option[String] = None,
    appAttemptId: Option[String] = None,
    jobId: Option[Int] = None,
@@ -2587,6 +2588,8 @@ private[spark] class CallerContext(
    taskId: Option[Long] = None,
    taskAttemptNumber: Option[Int] = None) extends Logging {
 
+   val upstreamCallerContextStr =
+     if (upstreamCallerContext.isDefined) s"_${upstreamCallerContext.get}" else ""
    val appIdStr = if (appId.isDefined) s"_${appId.get}" else ""
    val appAttemptIdStr = if (appAttemptId.isDefined) s"_${appAttemptId.get}" else ""
    val jobIdStr = if (jobId.isDefined) s"_JId_${jobId.get}" else ""
@@ -2597,7 +2600,8 @@ private[spark] class CallerContext(
      if (taskAttemptNumber.isDefined) s"_${taskAttemptNumber.get}" else ""
 
    val context = "SPARK_" + from + appIdStr + appAttemptIdStr +
-     jobIdStr + stageIdStr + stageAttemptIdStr + taskIdStr + taskAttemptNumberStr
+     jobIdStr + stageIdStr + stageAttemptIdStr + taskIdStr + taskAttemptNumberStr +
+     upstreamCallerContextStr
 
   /**
    * Set up the caller context [[context]] by invoking Hadoop CallerContext API of
--- a/core/src/main/scala/org/apache/spark/util/Utils.scala
+++ b/core/src/main/scala/org/apache/spark/util/Utils.scala
@@ -2569,6 +2569,7 @@ private[util] object CallerContext extends Logging {
  * @param from who sets up the caller context (TASK, CLIENT, APPMASTER)
  *
  * The parameters below are optional:
+ * @param upstreamCallerContext caller context the upstream application passes in
  * @param appId id of the app this task belongs to
  * @param appAttemptId attempt id of the app this task belongs to
  * @param jobId id of the job this task belongs to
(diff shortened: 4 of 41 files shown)


rxin:
Can't this just be a normal sparkconf? Why do we need to specialize the spark-submit arguments for this?

core/src/main/scala/org/apache/spark/util/Utils.scala
```diff
@@ -2523,6 +2524,8 @@ private[spark] class CallerContext(
    taskId: Option[Long] = None,
    taskAttemptNumber: Option[Int] = None) extends Logging {
 
+   val upstreamCallerContextStr =
+     if (upstreamCallerContext.isDefined) s"_${upstreamCallerContext.get}" else ""
```
jerryshao: We could simplify with `getOrElse` here and below.

core/src/main/scala/org/apache/spark/scheduler/Task.scala
```diff
@@ -92,8 +92,16 @@ private[spark] abstract class Task[T](
       kill(interruptThread = false)
     }
 
-    new CallerContext("TASK", appId, appAttemptId, jobId, Option(stageId), Option(stageAttemptId),
-      Option(taskAttemptId), Option(attemptNumber)).setCurrentContext()
+    new CallerContext(
+      "TASK",
+      SparkEnv.get.conf.getOption("spark.hadoop.callerContext"),
```
jerryshao: It would be better to define "spark.hadoop.callerContext" in `internal/config` and use that instead.

core/src/main/scala/org/apache/spark/deploy/SparkSubmit.scala
```diff
@@ -465,6 +465,8 @@ object SparkSubmit {
       OptionAssigner(args.archives, YARN, ALL_DEPLOY_MODES, sysProp = "spark.yarn.dist.archives"),
       OptionAssigner(args.principal, YARN, ALL_DEPLOY_MODES, sysProp = "spark.yarn.principal"),
       OptionAssigner(args.keytab, YARN, ALL_DEPLOY_MODES, sysProp = "spark.yarn.keytab"),
+      OptionAssigner(args.callerContext, YARN, ALL_DEPLOY_MODES,
+        sysProp = "spark.hadoop.callerContext"),
```
weiqingy: Thanks for this info. I have changed the property name to "spark.upstreamApp.callercontext".

core/src/main/scala/org/apache/spark/util/Utils.scala
weiqingy: `upstreamCallerContextStr` needs to combine some characters/words with `upstreamCallerContext.get`, so `getOrElse` is not used here and below.

core/src/main/scala/org/apache/spark/scheduler/Task.scala
weiqingy: Yes. Done.

weiqingy:
@rxin Thanks for reviewing this. I have updated the PR not to make this as a spark-submit argument.

weiqingy:
Hadoop verifies caller contexts passed in [here](https://github.com/apache/hadoop/blob/trunk/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/server/namenode/FSNamesystem.java#L6918). It limits the length.

weiqingy:
Hi, @tgravescs Could you please review this PR? Thanks.

core/src/main/scala/org/apache/spark/util/Utils.scala
```diff
@@ -2551,6 +2551,7 @@ private[util] object CallerContext extends Logging {
  */
 private[spark] class CallerContext(
    from: String,
+   upstreamCallerContext: Option[String] = None,
```
tgravescs: I'd prefer to see new parameters added as the last param to the class.

core/src/main/scala/org/apache/spark/util/Utils.scala
```diff
@@ -2591,6 +2595,16 @@ private[spark] class CallerContext(
       }
     }
   }
+
+  def prepareContext(context: String): String = {
+    // The default max size of Hadoop caller context is 128
+    lazy val len = SparkHadoopUtil.get.conf.getInt("hadoop.caller.context.max.size", 128)
```
tgravescs: I'm not sure we want to do this.  the 128 here is hardcoded and if hadoop changes its default then this is wrong.  It can change the default in the code without the config being present in the hadoop xml files.
  
while it would be nice to truncate to what hadoop actually is going to write I think its better to just send it and let hadoop do that.

docs/running-on-yarn.md
```diff
@@ -495,6 +495,15 @@ To use a custom metrics.properties for the application master and executors, upd
   name matches both the include and the exclude pattern, this file will be excluded eventually.
   </td>
 </tr>
+<tr>
+  <td><code>spark.hadoop.callerContext</code></td>
```
tgravescs: doesn't match above actual config spark.upstreamApp.callerContext.  how about:  spark.log.callerContext

If I'm running spark in standalonde mode with master/worker and reading from hdfs, the caller context would still work on the hdfs side, right?  So this isn't just a  spark on yarn config and should move to general configuration section but mention applies to yarn/hdfs.

core/src/main/scala/org/apache/spark/util/Utils.scala
mridulm: upstreamCallerContext.map("_" + _).getOrElse("") ?

core/src/main/scala/org/apache/spark/util/Utils.scala
```diff
@@ -2582,7 +2586,7 @@ private[spark] class CallerContext(
         val callerContext = Class.forName("org.apache.hadoop.ipc.CallerContext")
         val builder = Class.forName("org.apache.hadoop.ipc.CallerContext$Builder")
         // scalastyle:on classforname
-        val builderInst = builder.getConstructor(classOf[String]).newInstance(context)
+        val builderInst = builder.getConstructor(classOf[String]).newInstance(prepareContext(context))
```
mridulm: Should prepareContext be done once as part of initialization of context or should it need to be done for each invocation of setCurrentContext ?

core/src/main/scala/org/apache/spark/util/Utils.scala
```diff
@@ -2591,6 +2595,16 @@ private[spark] class CallerContext(
       }
     }
   }
+
+  def prepareContext(context: String): String = {
```
mridulm: private

core/src/main/scala/org/apache/spark/util/Utils.scala
```diff
@@ -2569,7 +2572,8 @@ private[spark] class CallerContext(
      if (taskAttemptNumber.isDefined) s"_${taskAttemptNumber.get}" else ""
 
    val context = "SPARK_" + from + appIdStr + appAttemptIdStr +
-     jobIdStr + stageIdStr + stageAttemptIdStr + taskIdStr + taskAttemptNumberStr
+     jobIdStr + stageIdStr + stageAttemptIdStr + taskIdStr + taskAttemptNumberStr +
+     upstreamCallerContextStr
```
mridulm: We can inline upstreamCallerContextStr and remove the variable ?

core/src/main/scala/org/apache/spark/util/Utils.scala
weiqingy: I agree with you that new parameters should be added as last parameters so that no client code would be broken. It’s just in this case, there are three callers in total and they are all under my control. The new optional parameter would be used much more frequently than other optional parameters, so I think I shall change the parameter orders before the class is used more widely. If I put the new parameter as the last one, users (including the existing three) would have to pass may “None”s as parameters.

core/src/main/scala/org/apache/spark/util/Utils.scala
```diff
@@ -2591,6 +2595,16 @@ private[spark] class CallerContext(
...
     }
   }
+
+  def prepareContext(context: String): String = {
+    // The default max size of Hadoop caller context is 128
+    lazy val len = SparkHadoopUtil.get.conf.getInt("hadoop.caller.context.max.size", 128)
+    if (context == null || context.length <= len) {
+      context
+    } else {
+      context.substring(0, len)
```
weiqingy: Yes. Done.

docs/running-on-yarn.md
weiqingy: Yes. I have changed the config to spark.log.callerContext and moved the documentation to `spark/docs/configuration.md`.

core/src/main/scala/org/apache/spark/util/Utils.scala
weiqingy: Yes. I have updated the PR to inline the variables.

core/src/main/scala/org/apache/spark/util/Utils.scala
weiqingy: Yes, I have updated the PR to use this.

core/src/main/scala/org/apache/spark/util/Utils.scala
weiqingy: Yes. Done.

tgravescs:
Jenkins, test this please

core/src/main/scala/org/apache/spark/util/Utils.scala
tgravescs: ok

core/src/main/scala/org/apache/spark/util/Utils.scala
tgravescs: why?  we set the actual context when we create it and there is no way to change it.  set is just actually calling the hadoop routine.  I don't know that it matters to much right now since we always create and set but its better to do.

core/src/main/scala/org/apache/spark/util/Utils.scala
```diff
@@ -2587,17 +2589,16 @@ private[spark] class CallerContext(
...
-   val jobIdStr = if (jobId.isDefined) s"_JId_${jobId.get}" else ""
-   val stageIdStr = if (stageId.isDefined) s"_SId_${stageId.get}" else ""
-   val stageAttemptIdStr = if (stageAttemptId.isDefined) s"_${stageAttemptId.get}" else ""
-   val taskIdStr = if (taskId.isDefined) s"_TId_${taskId.get}" else ""
-   val taskAttemptNumberStr =
-     if (taskAttemptNumber.isDefined) s"_${taskAttemptNumber.get}" else ""
-
-   val context = "SPARK_" + from + appIdStr + appAttemptIdStr +
-     jobIdStr + stageIdStr + stageAttemptIdStr + taskIdStr + taskAttemptNumberStr
+   val context = "SPARK_" +
```
tgravescs: make this private

tgravescs:
please look at jenkins failures

core/src/main/scala/org/apache/spark/util/Utils.scala
mridulm: It is possible to use named parameters, but I agree with the point of private scope + frequency of use here to be mitigating factor.

core/src/main/scala/org/apache/spark/util/Utils.scala
weiqingy: The reason prepareContext is called in each setCallerContext is that currently every client only need to set the caller context once, and each CallerContext object only call setCallerContext once. Yes, I agree with your point, that will make the implementation better and benefit future invocations.

core/src/main/scala/org/apache/spark/util/Utils.scala
weiqingy: Done.

mridulm:
Jenkins, test this please

tgravescs:
Jenkins, test this please

weiqingy:
@mridulm @tgravescs It seems Jenkins doesn't work?

tdas:
jenkins, ok to test

tgravescs:
+1.  @mridulm  you have any further comments?

mridulm:
Looks good to me

mridulm:
Merging into master

weiqingy:
Thanks a lot for the review, @tgravescs @mridulm

core/src/main/scala/org/apache/spark/util/Utils.scala
```diff
@@ -2578,26 +2579,38 @@ private[util] object CallerContext extends Logging {
...
-   val jobIdStr = if (jobId.isDefined) s"_JId_${jobId.get}" else ""
-   val stageIdStr = if (stageId.isDefined) s"_SId_${stageId.get}" else ""
-   val stageAttemptIdStr = if (stageAttemptId.isDefined) s"_${stageAttemptId.get}" else ""
-   val taskIdStr = if (taskId.isDefined) s"_TId_${taskId.get}" else ""
-   val taskAttemptNumberStr =
-     if (taskAttemptNumber.isDefined) s"_${taskAttemptNumber.get}" else ""
-
-   val context = "SPARK_" + from + appIdStr + appAttemptIdStr +
-     jobIdStr + stageIdStr + stageAttemptIdStr + taskIdStr + taskAttemptNumberStr
+  from: String,
```
rxin: this is off?

core/src/main/scala/org/apache/spark/internal/config/package.scala
```diff
@@ -207,6 +207,10 @@ package object config {
     .booleanConf
     .createWithDefault(false)
 
+  private[spark] val APP_CALLER_CONTEXT = ConfigBuilder("spark.log.callerContext")
```
rxin: shouldn't this option be called spark.yarn.log.callerContext?

weiqingy: This is not just for Yarn, if running spark apps in standalonde mode with master and workers and reading/writing from/to hdfs, the caller context would still work on the hdfs side. (PS. we also can not use spark.hadoop prefix that will be treated as Hadoop configuration and set to `Configuration`.)

core/src/main/scala/org/apache/spark/util/Utils.scala
weiqingy: What do you mean "off"?

core/src/main/scala/org/apache/spark/internal/config/package.scala
rxin: Thanks for the comment. Would it be a problem if we use spark.hadoop.log.callerContext? I know it gets passed into Configuration, but why would that be a problem? Is it overriding some common configuration in Hadoop?

tgravescs: It shouldn't hurt anything as I don't think there is a hadoop config called that but if there was it would conflict.  In general I think that is a bad idea for us to purposely do this, there is no reason to put it into the hadoop configuration also.  I saw this as something that could apply to other things (then yarn/hdfs) if they supplied the api.  For instance is aws s3 or any other filesystem provided the similar api you could use same config.  But I don't know if it does now.

What is your concern with the current name?

rxin: My concern is that this is a very hadoop specific thing at the moment, and it is unclear other environments will support it, so it'd make more sense to have hadoop in the name. A lot of Spark users don't run Hadoop.

tgravescs: Personally I don't like using the spark.hadoop. for this if we know its already being used and these configs will always be added to the hadoop config when not needed.  I guess perhaps we messed up naming that to put things into hadoop configuration.

what about spark.log.hadoop.callerContext?  Although perhaps we should set a policy for this in general.

I would say we should change spark.hadoop being applied to hadoop conf (to like spark.hadoopConf.) but even though I don't see it documented anywhere I think it would be to painful to change  as I know people are using it.

PR #15564, opened 2016-10-20
[SPARK-17331][FOLLOWUP][ML][CORE] Avoid allocating 0-length arrays 

zhengruifeng:

`Array[T]()` -> `Array.empty[T]` to avoid allocating 0-length arrays.
Use regex `find . -name '*.scala' | xargs -i bash -c 'egrep "Array\[[A-Za-z]+\]\(\)" -n {} && echo {}'` to find modification candidates.

cc @srowen 
## How was this patch tested?

existing tests


srowen:
Yes, I'm OK with this. I intentionally didn't change these in the original change because they're just in test code and wanted to keep the change slimmer. But this shouldn't hurt.

zhengruifeng:
@srowen Yes. Only two changes are in `main`, others are all in testsuites.

srowen:
Merged to master

PR #15567, opened 2016-10-20
[SPARK-14393][SQL] values generated by non-deterministic functions shouldn't change after coalesce or union

mengxr:

When a user appended a column using a "nondeterministic" function to a DataFrame, e.g., `rand`, `randn`, and `monotonically_increasing_id`, the expected semantic is the following:
- The value in each row should remain unchanged, as if we materialize the column immediately, regardless of later DataFrame operations.

However, since we use `TaskContext.getPartitionId` to get the partition index from the current thread, the values from nondeterministic columns might change if we call `union` or `coalesce` after. `TaskContext.getPartitionId` returns the partition index of the current Spark task, which might not be the corresponding partition index of the DataFrame where we defined the column.

See the unit tests below or JIRA for examples.

This PR uses the partition index from `RDD.mapPartitionWithIndex` instead of `TaskContext` and fixes the partition initialization logic in whole-stage codegen, normal codegen, and codegen fallback. `initializeStatesForPartition(partitionIndex: Int)` was added to `Projection`, `Nondeterministic`, and `Predicate` (codegen) and initialized right after object creation in `mapPartitionWithIndex`. `newPredicate` now returns a `Predicate` instance rather than a function for proper initialization.
## How was this patch tested?

Unit tests. (Actually I'm not very confident that this PR fixed all issues without introducing new ones ...)

cc: @rxin @davies


core/src/main/scala/org/apache/spark/rdd/RDD.scala
```diff
@@ -787,14 +787,26 @@ abstract class RDD[T: ClassTag](
...
+    new MapPartitionsRDD(
+      this,
+      (context: TaskContext, index: Int, iter: Iterator[T]) => f(index, iter),
+      preservesPartitioning)
+  }
+
+  /**
+   * [performance] Spark's internal mapPartitions method that skips closure cleaning.
+   */
   private[spark] def mapPartitionsInternal[U: ClassTag](
```
rxin: can we get rid of this?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/Expression.scala
```diff
@@ -274,12 +274,12 @@ trait Nondeterministic extends Expression {
 
   private[this] var initialized = false
 
-  final def setInitialValues(): Unit = {
-    initInternal()
+  final def initializeStatesForPartition(partitionIndex: Int): Unit = {
```
rxin: while you are at it, it'd be great to add some comments documenting the function ...

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/codegen/CodeGenerator.scala
```diff
@@ -185,6 +185,20 @@ class CodegenContext {
   }
 
   /**
+   * Code statements to initialize states that depend on the partition index.
+   * An integer `partitionIndex` will be made available within the scope.
+   */
+  val partitionInitializationStatements: mutable.ArrayBuffer[String] = mutable.ArrayBuffer.empty
+
+  def addPartitionInitializationStatement(statement: String): Unit = {
```
rxin: any reason you are creating this rather than just using addMutableState?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/Expression.scala
```diff
@@ -274,12 +274,12 @@ trait Nondeterministic extends Expression {
 
   private[this] var initialized = false
```
rxin: should we change this to transient? then it will always get reset to false on a new partition.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/SparkPartitionID.scala
```diff
@@ -17,16 +17,15 @@
...
 import org.apache.spark.sql.catalyst.expressions.codegen.{CodegenContext, ExprCode}
 import org.apache.spark.sql.types.{DataType, IntegerType}
 
 /**
- * Expression that returns the current partition id of the Spark task.
+ * Expression that returns the current partition id.
  */
 @ExpressionDescription(
-  usage = "_FUNC_() - Returns the current partition id of the Spark task",
+  usage = "_FUNC_() - Returns the current partition id",
```
rxin: hmmm this is behavior changing, and there is some value to the old partition id ...

rxin: I'd consider introducing a new expression for the proper id and leave the old one as is.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/codegen/GeneratePredicate.scala
```diff
@@ -25,19 +25,26 @@ import org.apache.spark.sql.catalyst.expressions._
  */
 abstract class Predicate {
   def eval(r: InternalRow): Boolean
+
+  /**
+   * Initialize internal states given the current partition index.
+   * This is used by non-deterministic expressions to set initial states.
+   * The default implementation does nothing.
+   */
+  def initializeStatesForPartition(partitionIndex: Int): Unit = {}
```
rxin: to make this safer, i'd create an internal variable "isInitialized" similar to the one in nondeterministic expression, and assert in eval if isInitialized is false
.

rxin:
Reviewing this code makes me realize how painful it is when project/filter are just scala functions ... it'd be much easier to review if they have methods defined (e.g. eval, or execute) ...

rxin:
So my biggest question is whether you've changed all the places to call initialize where projection/predicate are used.

core/src/main/scala/org/apache/spark/rdd/RDD.scala
mengxr: There are 20+ probably valid use of `mapPartitionsInternal`. The main problem is that changing it to `mapPartitionsWithIndexInternal` doesn't really force people to initialize the partition.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/Expression.scala
mengxr: I don't want to overload the name `initialize`, which is a little vague, how about `initStates`? Again, the issue is that even with comments we cannot force users to initialize it.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/codegen/CodeGenerator.scala
mengxr: I'm a little worried about introducing more issues by moving `initMutableStates` out from the constructor. The current implementation at least maintains the existing behavior if we missed `initializeStatesForPartition` somewhere.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/Expression.scala
mengxr: will do

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/SparkPartitionID.scala
mengxr: I thought about this. But I don't think the current behavior is the expected behavior from users. This is the same issue as in `monotonically_increasing_id`.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/codegen/GeneratePredicate.scala
mengxr: I didn't test. Would doing that hurt the performance?

rxin: I don't think so, since it is in the interpreted path which is already very slow. Also in the normal case the condition will always be false, so CPU branch prediction should work its magic.

shearerp:
SPARK-14241 doesn't just occur with union and coalesce, it also occurs with filter and probably other operations. Hopefully this PR will address all of those situations. I strongly agree with the expected semantic in the original PR message by mengxr - this has bitten me on multiple occasions.

mengxr:
@rxin I updated the implementation to force initialization in Projection/Expression. This will fail many tests. I fixed all in `catalyst`, but not yet in `sql`. I want to propose the following change:

* add `Deterministic` extends `PartitionDependent` (always set initialized to true)
* remove `Nondeterministic`

Basically, we assume non-deterministic by default unless marked as deterministic. This will require updating all expressions but make the code less messy.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/Expression.scala
```diff
@@ -264,23 +264,43 @@ trait NonSQLExpression extends Expression {
   }
 }
 
+/**
+ * Trait for expressions that requires initialization based on the partition index prior to task
+ * execution on a partition.
+ */
+trait PartitionDependent {
```
rxin: as discussed offline this is StatefulExpression

mengxr: Actually, `Projection` also need the same trait. Shall we call it `PartitionedStateful`?

mengxr:
I reverted the changes I made to enforce `Projection.initialize`, which touched too many files and most of them doesn't really need to handle nondeterministic expressions. The current implementation at least throws a runtime error if a nondeterministic expression didn't get initialized, instead of returning incorrect result.

Renamed `initializeStatesForPartition` to `initialize`.

rxin:

sql/core/src/main/scala/org/apache/spark/sql/execution/basicPhysicalOperators.scala
```diff
@@ -205,10 +206,11 @@ case class FilterExec(condition: Expression, child: SparkPlan)
 
   protected override def doExecute(): RDD[InternalRow] = {
     val numOutputRows = longMetric("numOutputRows")
-    child.execute().mapPartitionsInternal { iter =>
+    child.execute().mapPartitionsWithIndexInternal { (index, iter) =>
       val predicate = newPredicate(condition, child.output)
+      predicate.initialize(0)
```
gatorsmile: Just wondering why `FilterExec` is not using `index` to initialize the conditions?

PR #15302, opened 2016-09-29
[SPARK-17732][SQL] ALTER TABLE DROP PARTITION should support comparators

dongjoon-hyun:
## What changes were proposed in this pull request?

This PR aims to support `comparators`, e.g. '<', '<=', '>', '>=', again in Apache Spark 2.0 for backward compatibility.

**Spark 1.6.2**

``` scala
scala> sql("CREATE TABLE sales(id INT) PARTITIONED BY (country STRING, quarter STRING)")
res0: org.apache.spark.sql.DataFrame = [result: string]

scala> sql("ALTER TABLE sales DROP PARTITION (country < 'KR')")
res1: org.apache.spark.sql.DataFrame = [result: string]
```

**Spark 2.0**

``` scala
scala> sql("CREATE TABLE sales(id INT) PARTITIONED BY (country STRING, quarter STRING)")
res0: org.apache.spark.sql.DataFrame = []

scala> sql("ALTER TABLE sales DROP PARTITION (country < 'KR')")
org.apache.spark.sql.catalyst.parser.ParseException:
mismatched input '<' expecting {')', ','}(line 1, pos 42)
```

After this PR, it's supported.
## How was this patch tested?

Pass the Jenkins test with a newly added testcase.

PR #15568, opened 2016-10-20
[SPARK-18028][SQL] simplify TableFileCatalog

cloud-fan:

Simplify/cleanup TableFileCatalog:
1. pass a `CatalogTable` instead of `databaseName` and `tableName` into `TableFileCatalog`, so that we don't need to fetch table metadata from metastore again
2. In `TableFileCatalog.filterPartitions0`, DO NOT set `PartitioningAwareFileCatalog.BASE_PATH_PARAM`. According to the [classdoc](https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/PartitioningAwareFileCatalog.scala#L189-L209), the default value of `basePath` already satisfies our need. What's more, if we set this parameter, we may break the case 2 which is metioned in the classdoc.
3. add `equals` and `hashCode` to `TableFileCatalog`
4. add `SessionCatalog.listPartitionsByFilter` which handles case sensitivity.
## How was this patch tested?

existing tests.

### DIFF
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/SessionCatalog.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/SessionCatalog.scala
@@ -756,6 +756,20 @@ class SessionCatalog(
   }
 
   /**
+   * List the metadata of partitions that belong to the specified table, assuming it exists, that
+   * satisfy the given partition-pruning predicate expressions.
+   */
+  def listPartitionsByFilter(
+      tableName: TableIdentifier,
+      predicates: Seq[Expression]): Seq[CatalogTablePartition] = {
+    val db = formatDatabaseName(tableName.database.getOrElse(getCurrentDatabase))
+    val table = formatTableName(tableName.table)
+    requireDbExists(db)
+    requireTableExists(TableIdentifier(table, Option(db)))
+    externalCatalog.listPartitionsByFilter(db, table, predicates)
+  }
+
+  /**
    * Verify if the input partition spec exactly matches the existing defined partition spec
    * The columns must be the same but the orders could be different.
    */
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/TableFileCatalog.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/TableFileCatalog.scala
@@ -77,6 +77,9 @@ class TableFileCatalog(
 
   override def inputFiles: Array[String] = filterPartitions(Nil).inputFiles
 
+  // `TableFileCatalog` may be a member of `HadoopFsRelation`, `HadoopFsRelation` may be a member
+  // of `LogicalRelation`, and `LogicalRelation` may be used as the cache key. So we need to
+  // implement `equals` and `hashCode` here, to make it work with cache lookup.
   override def equals(o: Any): Boolean = o match {
     case other: TableFileCatalog => this.table.identifier == other.table.identifier
     case _ => false
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/TableFileCatalog.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/TableFileCatalog.scala
@@ -20,36 +20,30 @@ package org.apache.spark.sql.execution.datasources
 import org.apache.hadoop.fs.Path
 
 import org.apache.spark.sql.SparkSession
+import org.apache.spark.sql.catalyst.catalog.CatalogTable
 import org.apache.spark.sql.catalyst.expressions._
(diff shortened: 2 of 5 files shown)


cloud-fan:
cc @ericl @mallman @yhuai

mallman:
@cloud-fan, I don't know what you mean in item 4 about `SessionCatalog.listPartitionsByFilter` handling case-sensitivity. What case-sensitivity issue are you referring to, and does this PR handle it differently?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/SessionCatalog.scala
```diff
@@ -756,6 +756,20 @@ class SessionCatalog(
   }
 
   /**
+   * List the metadata of partitions that belong to the specified table, assuming it exists, that
+   * satisfy the given partition-pruning predicate expressions.
+   */
+  def listPartitionsByFilter(
+      tableName: TableIdentifier,
+      predicates: Seq[Expression]): Seq[CatalogTablePartition] = {
```
dongjoon-hyun: Thank you for adding this, @cloud-fan ! I can use this in  https://github.com/apache/spark/pull/15302 .

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/TableFileCatalog.scala
```diff
@@ -102,6 +95,13 @@ class TableFileCatalog(
   }
 
   override def inputFiles: Array[String] = allPartitions.inputFiles
+
+  override def equals(o: Any): Boolean = o match {
```
ericl: What is this needed for?

ericl:
LGTM, but had a question on why we need to override equals().

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/TableFileCatalog.scala
cloud-fan: Under hive context, we will cache the `LogicalRelation` for every data source table(including converted from hive), which means every table will always have a `TableFileCatalog` of same instance.

However, it's not true in sql core. We will re-construct the `TableFileCatalog` and `LogicalRelation` everytime we look up a table. Thus we may encounter cache miss even if the table is cached, because `TableFileCatalog` of difference instances never equal to each other.

Although it's not a real problem now, I think it's reasonable to follow `ListFileCatalong` and add the `equals` and `hashCode`

cloud-fan:
@mallman , item 4 is a potential problem in the future. The current workflow is, we get the `MetastoreRelation` via `HiveMetastoreCatalog.lookupRelation`, which always lower case the database and table name. Then we construct `TableFileCatalog` and call `ExternalCatalog.listPartitionsByFilter` with the database and table name. So we won't have case senstivity problem here.

However, we may have a workflow which call `listPartitionsByFilter` directly with database and table name given by users. e.g. #15302. Then we need to care about case sensitivity.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/TableFileCatalog.scala
ericl: I see. Could we add a unit test for this and a comment here?

ericl:
lgtm, thanks for adding the test!

cloud-fan:

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/TableFileCatalog.scala
```diff
@@ -66,24 +60,32 @@ class TableFileCatalog(
...
+        table.identifier, filters)
+      val partitionSchema = table.partitionSchema
+      val partitions = selectedPartitions.map { p =>
+        PartitionPath(p.toRow(partitionSchema), p.storage.locationUri.get)
+      }
+      val partitionSpec = PartitionSpec(partitionSchema, partitions)
+      new PrunedTableFileCatalog(
+        sparkSession, new Path(baseLocation.get), fileStatusCache, partitionSpec)
+    } else {
+      new ListingFileCatalog(sparkSession, rootPaths, table.storage.properties, None)
```
ConeyLiu: Hi @cloud-fan, here I have a question why should we remove the `fileStatusCache` for `ListingFileCatalog`? Do we need to add it back?

cloud-fan: seems it was a mistake. Can you send a PR to add it? thanks!

ConeyLiu: OK, sure. Thanks for your answer.

PR #15569, opened 2016-10-20
[SPARK-18029][SQL] PruneFileSourcePartitions should not change the output of LogicalRelation

cloud-fan:

In `PruneFileSourcePartitions`, we will replace the `LogicalRelation` with a pruned one. However, this replacement may change the output of the `LogicalRelation` if it doesn't have `expectedOutputAttributes`. This PR fixes it.
## How was this patch tested?

the new `PruneFileSourcePartitionsSuite`

### DIFF
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/interface.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/interface.scala
@@ -102,8 +102,8 @@ case class CatalogTablePartition(
    * Given the partition schema, returns a row with that schema holding the partition values.
    */
   def toRow(partitionSchema: StructType): InternalRow = {
-    InternalRow.fromSeq(partitionSchema.map { case StructField(name, dataType, _, _) =>
-      Cast(Literal(spec(name)), dataType).eval()
+    InternalRow.fromSeq(partitionSchema.map { field =>
+      Cast(Literal(spec(field.name)), field.dataType).eval()
     })
   }
 }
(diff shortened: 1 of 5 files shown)


cloud-fan:
cc @ericl @mallman @yhuai

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/interface.scala
```diff
@@ -102,8 +102,8 @@ case class CatalogTablePartition(
    * Given the partition schema, returns a row with that schema holding the partition values.
    */
   def toRow(partitionSchema: StructType): InternalRow = {
-    InternalRow.fromSeq(partitionSchema.map { case StructField(name, dataType, _, _) =>
-      Cast(Literal(spec(name)), dataType).eval()
+    InternalRow.fromSeq(partitionSchema.map { field =>
```
mallman: Why are you proposing this change?

rxin: See https://github.com/databricks/scala-style-guide/pull/40

mallman: I see. Do we have a Spark style guide?

rxin: There is this one: https://cwiki.apache.org/confluence/display/SPARK/Spark+Code+Style+Guide

But it is fairly basic.

mallman:
LGTM.

cloud-fan:

PR #15571, opened 2016-10-20
[SPARK-18034] Upgrade to MiMa 0.1.11 to fix flakiness

JoshRosen:
We should upgrade to the latest release of MiMa (0.1.11) in order to include a fix for a bug which led to flakiness in the MiMa checks (https://github.com/typesafehub/migration-manager/issues/115).

### DIFF
--- a/project/MimaExcludes.scala
+++ b/project/MimaExcludes.scala
@@ -81,7 +81,12 @@ object MimaExcludes {
       // [SPARK-17338][SQL] add global temp view
       ProblemFilters.exclude[ReversedMissingMethodProblem]("org.apache.spark.sql.catalog.Catalog.dropGlobalTempView"),
       ProblemFilters.exclude[IncompatibleResultTypeProblem]("org.apache.spark.sql.catalog.Catalog.dropTempView"),
-      ProblemFilters.exclude[ReversedMissingMethodProblem]("org.apache.spark.sql.catalog.Catalog.dropTempView")
+      ProblemFilters.exclude[ReversedMissingMethodProblem]("org.apache.spark.sql.catalog.Catalog.dropTempView"),
+
+      // [SPARK-18034] Upgrade to MiMa 0.1.11 to fix flakiness.
+      ProblemFilters.exclude[InheritedNewAbstractMethodProblem]("org.apache.spark.ml.param.shared.HasAggregationDepth.aggregationDepth"),
+      ProblemFilters.exclude[InheritedNewAbstractMethodProblem]("org.apache.spark.ml.param.shared.HasAggregationDepth.getAggregationDepth"),
+      ProblemFilters.exclude[InheritedNewAbstractMethodProblem]("org.apache.spark.ml.param.shared.HasAggregationDepth.org$apache$spark$ml$param$shared$HasAggregationDepth$_setter_$aggregationDepth_=")
     )
   }
 
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -6,7 +6,7 @@ addSbtPlugin("net.virtual-void" % "sbt-dependency-graph" % "0.8.2")
 
 addSbtPlugin("org.scalastyle" %% "scalastyle-sbt-plugin" % "0.8.0")
 
-addSbtPlugin("com.typesafe" % "sbt-mima-plugin" % "0.1.9")
+addSbtPlugin("com.typesafe" % "sbt-mima-plugin" % "0.1.11")
 
 addSbtPlugin("com.alpinenow" % "junit_xml_listener" % "0.5.1")
 


JoshRosen:
Hmm, looks like MiMa gained some new checks in its latest release:

```
[error]  * abstract method aggregationDepth()org.apache.spark.ml.param.IntParam in trait org.apache.spark.ml.param.shared.HasAggregationDepth is inherited by class LogisticRegressionParams in current version.
[error]    filter with: ProblemFilters.exclude[InheritedNewAbstractMethodProblem]("org.apache.spark.ml.param.shared.HasAggregationDepth.aggregationDepth")
[error]  * abstract method getAggregationDepth()Int in trait org.apache.spark.ml.param.shared.HasAggregationDepth is inherited by class LogisticRegressionParams in current version.
[error]    filter with: ProblemFilters.exclude[InheritedNewAbstractMethodProblem]("org.apache.spark.ml.param.shared.HasAggregationDepth.getAggregationDepth")
[error]  * abstract synthetic method org$apache$spark$ml$param$shared$HasAggregationDepth$_setter_$aggregationDepth_=(org.apache.spark.ml.param.IntParam)Unit in trait org.apache.spark.ml.param.shared.HasAggregationDepth is inherited by class LogisticRegressionParams in current version.
[error]    filter with: ProblemFilters.exclude[InheritedNewAbstractMethodProblem]("org.apache.spark.ml.param.shared.HasAggregationDepth.org$apache$spark$ml$param$shared$HasAggregationDepth$_setter_$aggregationDepth_=")
[error]  * abstract method aggregationDepth()org.apache.spark.ml.param.IntParam in trait org.apache.spark.ml.param.shared.HasAggregationDepth is inherited by class LinearRegressionParams in current version.
[error]    filter with: ProblemFilters.exclude[InheritedNewAbstractMethodProblem]("org.apache.spark.ml.param.shared.HasAggregationDepth.aggregationDepth")
[error]  * abstract method getAggregationDepth()Int in trait org.apache.spark.ml.param.shared.HasAggregationDepth is inherited by class LinearRegressionParams in current version.
[error]    filter with: ProblemFilters.exclude[InheritedNewAbstractMethodProblem]("org.apache.spark.ml.param.shared.HasAggregationDepth.getAggregationDepth")
[error]  * abstract synthetic method org$apache$spark$ml$param$shared$HasAggregationDepth$_setter_$aggregationDepth_=(org.apache.spark.ml.param.IntParam)Unit in trait org.apache.spark.ml.param.shared.HasAggregationDepth is inherited by class LinearRegressionParams in current version.
[error]    filter with: ProblemFilters.exclude[InheritedNewAbstractMethodProblem]("org.apache.spark.ml.param.shared.HasAggregationDepth.org$apache$spark$ml$param$shared$HasAggregationDepth$_setter_$aggregationDepth_=")
[error]  * abstract method aggregationDepth()org.apache.spark.ml.param.IntParam in trait org.apache.spark.ml.param.shared.HasAggregationDepth is inherited by class AFTSurvivalRegressionParams in current version.
[error]    filter with: ProblemFilters.exclude[InheritedNewAbstractMethodProblem]("org.apache.spark.ml.param.shared.HasAggregationDepth.aggregationDepth")
[error]  * abstract method getAggregationDepth()Int in trait org.apache.spark.ml.param.shared.HasAggregationDepth is inherited by class AFTSurvivalRegressionParams in current version.
[error]    filter with: ProblemFilters.exclude[InheritedNewAbstractMethodProblem]("org.apache.spark.ml.param.shared.HasAggregationDepth.getAggregationDepth")
[error]  * abstract synthetic method org$apache$spark$ml$param$shared$HasAggregationDepth$_setter_$aggregationDepth_=(org.apache.spark.ml.param.IntParam)Unit in trait org.apache.spark.ml.param.shared.HasAggregationDepth is inherited by class AFTSurvivalRegressionParams in current version.
[error]    filter with: ProblemFilters.exclude[InheritedNewAbstractMethodProblem]("org.apache.spark.ml.param.shared.HasAggregationDepth.org$apache$spark$ml$param$shared$HasAggregationDepth$_setter_$aggregationDepth_=")
```

JoshRosen:
Yep, this new MiMa inspection was added in https://github.com/typesafehub/migration-manager/issues/119:

> When a new abstract class/trait with abstract members is added to the hierarchy of an existing class/trait, a binary incompatibility must be reported if the abstract members added by the new type did not exist in the hierarchy of the existing class/trait.

Looking at our particular error, it appears that this was introduced by https://github.com/apache/spark/commit/61ef74f2272faa7ce8f2badc7e00039908e3551f, which added a new `HasAggregationDepth` trait and then mixed it into other traits.

It looks like `HasAggregationDepth` actually defines `final` implementations for al of its methods:

```scala
/**
 * Trait for shared param aggregationDepth (default: 2).
 */
private[ml] trait HasAggregationDepth extends Params {

  /**
   * Param for suggested depth for treeAggregate (>= 2).
   * @group param
   */
  final val aggregationDepth: IntParam = new IntParam(this, "aggregationDepth", "suggested depth for treeAggregate (>= 2)", ParamValidators.gtEq(2))

  setDefault(aggregationDepth, 2)

  /** @group getParam */
  final def getAggregationDepth: Int = $(aggregationDepth)
}
```

I think that a trait with fields / implementations compiles down to both an interface and an abstract class in Scala 2.10 / 2.11, so I imagine that this isn't actually okay. This is confirmed by one of the test cases in MiMa: https://github.com/dotta/migration-manager/tree/54d54a8d1d7f0da8909c860e734bc1335ddae4c4/reporter/functional-tests/src/test/trait-extending-new-trait-with-concrete-method-nok

@dotta, to clarify the wording in your commit to MiMa, is it accurate to replace "abstract members" by just "members" and say that adding any members to a trait, abstract or concrete, can introduce a binary incompatibility if it wasn't already presence in subclasses's existing hierarchy?

jkbradley:
@JoshRosen I could imagine us introducing this same issue in the future with more Params in sharedParams.scala (like ```HasAggregationDepth```).  As long as it's fine for us add MimaExcludes (and are sure Java users won't experience issues), then this seems fine.

JoshRosen:
I'm not super familiar with MLlib, but are the existing `Params` subclasses intended to be subclassed by code which lives outside of Spark? If not, then I think we should be okay in terms of compatibility.

jkbradley:
They currently aren't, but a lot of users want them to be.  To make them available outside of Spark, we could create corresponding Java interfaces for them and expose those interfaces instead of the Scala traits.

JoshRosen:
Gotcha. For now, I've added MiMa excludes to fix the build and plan to merge this as soon as tests pass since I want to fix the flakiness in our builds. We can revisit removing the exclude and fixing this later if we want to.

PR #15573, opened 2016-10-20
[SPARK-18035] [SQL] Introduce performant and memory efficient APIs to create ArrayBasedMapData

tejasapatil:

Jira: https://issues.apache.org/jira/browse/SPARK-18035

In HiveInspectors, I saw that converting Java map to Spark's `ArrayBasedMapData` spent quite sometime in buffer copying : https://github.com/apache/spark/blob/master/sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveInspectors.scala#L658

The reason being `map.toSeq` allocates a new buffer and copies the map entries to it: https://github.com/scala/scala/blob/2.11.x/src/library/scala/collection/MapLike.scala#L323

This copy is not needed as we get rid of it once we extract the key and value arrays.

Here is the call trace:

```
org.apache.spark.sql.hive.HiveInspectors$$anonfun$unwrapperFor$41.apply(HiveInspectors.scala:664)
scala.collection.AbstractMap.toSeq(Map.scala:59)
scala.collection.MapLike$class.toSeq(MapLike.scala:323)
scala.collection.AbstractMap.toBuffer(Map.scala:59)
scala.collection.MapLike$class.toBuffer(MapLike.scala:326)
scala.collection.AbstractTraversable.copyToBuffer(Traversable.scala:104)
scala.collection.TraversableOnce$class.copyToBuffer(TraversableOnce.scala:275)
scala.collection.mutable.ArrayBuffer.$plus$plus$eq(ArrayBuffer.scala:48)
scala.collection.mutable.ArrayBuffer.$plus$plus$eq(ArrayBuffer.scala:104)
scala.collection.generic.Growable$class.$plus$plus$eq(Growable.scala:59)
scala.collection.AbstractIterable.foreach(Iterable.scala:54)
scala.collection.IterableLike$class.foreach(IterableLike.scala:72)
scala.collection.AbstractIterator.foreach(Iterator.scala:1336)
scala.collection.Iterator$class.foreach(Iterator.scala:893)
scala.collection.generic.Growable$$anonfun$$plus$plus$eq$1.apply(Growable.scala:59)
scala.collection.generic.Growable$$anonfun$$plus$plus$eq$1.apply(Growable.scala:59)
```

Also, earlier code was populating keys and values arrays separately by iterating twice. The PR avoids double iteration of the map and does it in one iteration.

EDIT: During code review, there were several more places in the code which were found to do similar thing. The PR dedupes those instances and introduces convenient APIs which are performant and memory efficient
## Performance gains

The number is subjective and depends on how many map columns are accessed in the query and average entries per map. For one the queries that I tried out, I saw 3% CPU savings (end-to-end) for the query.
## How was this patch tested?

This does not change the end result produced so relying on existing tests.

### DIFF
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/CatalystTypeConverters.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/CatalystTypeConverters.scala
@@ -415,10 +415,9 @@ object CatalystTypeConverters {
     case arr: Array[Any] => new GenericArrayData(arr.map(convertToCatalyst))
     case map: Map[_, _] =>
       ArrayBasedMapData(
-        map.iterator,
-        map.size,
-        (key) => convertToCatalyst(key),
-        (value) => convertToCatalyst(value))
+        map,
+        (key: Any) => convertToCatalyst(key),
+        (value: Any) => convertToCatalyst(value))
     case other => other
   }
 
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/CatalystTypeConverters.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/CatalystTypeConverters.scala
@@ -199,34 +199,14 @@ object CatalystTypeConverters {
     private[this] val keyConverter = getConverterForType(keyType)
     private[this] val valueConverter = getConverterForType(valueType)
 
-    override def toCatalystImpl(scalaValue: Any): MapData = scalaValue match {
-      case m: Map[_, _] =>
-        val length = m.size
-        val convertedKeys = new Array[Any](length)
-        val convertedValues = new Array[Any](length)
+    override def toCatalystImpl(scalaValue: Any): MapData = {
+      val keyFunction = (k: Any) => keyConverter.toCatalyst(k)
+      val valueFunction = (k: Any) => valueConverter.toCatalyst(k)
 
-        var i = 0
-        for ((key, value) <- m) {
-          convertedKeys(i) = keyConverter.toCatalyst(key)
-          convertedValues(i) = valueConverter.toCatalyst(value)
-          i += 1
-        }
-        ArrayBasedMapData(convertedKeys, convertedValues)
-
-      case jmap: JavaMap[_, _] =>
-        val length = jmap.size()
-        val convertedKeys = new Array[Any](length)
-        val convertedValues = new Array[Any](length)
-
-        var i = 0
-        val iter = jmap.entrySet.iterator
-        while (iter.hasNext) {
-          val entry = iter.next()
-          convertedKeys(i) = keyConverter.toCatalyst(entry.getKey)
-          convertedValues(i) = valueConverter.toCatalyst(entry.getValue)
-          i += 1
-        }
-        ArrayBasedMapData(convertedKeys, convertedValues)
+      scalaValue match {
+        case map: Map[_, _] => ArrayBasedMapData(map, keyFunction, valueFunction)
+        case javaMap: JavaMap[_, _] => ArrayBasedMapData(javaMap, keyFunction, valueFunction)
+      }
     }
 
     override def toScala(catalystValue: MapData): Map[Any, Any] = {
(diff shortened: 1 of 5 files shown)


tejasapatil:
Jenkins test this please

gatorsmile:
Can you de-duplicate the code changes? 

BTW, 3% savings? Can you show us the query and the plan? Thanks!

tejasapatil:
>> Can you de-duplicate the code changes?
done

>> Can you show us the query and the plan?
Its a production query so I would avoid putting that publically.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/ArrayBasedMapData.scala
```diff
@@ -35,6 +39,23 @@ object ArrayBasedMapData {
     ArrayBasedMapData(array.map(_._1), array.map(_._2))
   }
 
+  def apply(map: JMap[_, _],
+            keyUnwrapper: (Any) => Any,
+            valueUnwrapper: (Any) => Any): ArrayBasedMapData = {
```
gatorsmile: Nit: indent issues:
```Scala
def apply(
    map: JMap[_, _],
    keyUnwrapper: (Any) => Any,
    valueUnwrapper: (Any) => Any): ArrayBasedMapData = {
...  
}
```

gatorsmile:
Could you explain the pattern? For example, How many do such columns exist in your query? How large is your table? Are your query complex? Or just a simple SELECT query?

gatorsmile:
Just checked the code base. It sounds like multiple places are having the same logics. For example, `MapConverter` in `CatalystTypeConverters`. Could you do a quick check for all the caller who is using the [apply function of ArrayBasedMapData](https://github.com/tejasapatil/spark/blob/abc1d6f53efeedfaf9e35f064e5ff6d75c4b3774/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/ArrayBasedMapData.scala#L59).

tejasapatil:
@gatorsmile

> Could you explain the pattern? For example, How many do such columns exist in your query? How large is your table? Are your query complex? Or just a simple SELECT query?

It was < 10 TB data written out. Query was reading from 2 hive tables, performing FULL OUTER JOIN and writing data back to an output table. Output table was 1 map column, rest half a dozen numeric columns.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/ArrayBasedMapData.scala
```diff
@@ -35,6 +39,23 @@ object ArrayBasedMapData {
...
+    val keys: Array[Any] = new Array[Any](map.size())
+    val values: Array[Any] = new Array[Any](map.size())
+
+    var i: Int = 0
+    map.asScala.foreach(
```
tejasapatil: done

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/ArrayBasedMapData.scala
```diff
@@ -35,6 +39,23 @@ object ArrayBasedMapData {
     ArrayBasedMapData(array.map(_._1), array.map(_._2))
   }
 
+  def apply(map: JMap[_, _],
```
tejasapatil: done

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/ArrayBasedMapData.scala
tejasapatil: fixed

tejasapatil:
>> Could you do a quick check for all the callers who are using 

I have updated all places except three : 
- JacksonParser : can be abstracted out but the API will be too specific to this usage and ugly.
- ParquetRowConverter : this has coupling with some existing interfaces
- HiveInspectorSuite : the test already created dummy arrays so not applicable

rxin:
High level change LGTM.

@gatorsmile can you do the final review & merge?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/CatalystTypeConverters.scala
```diff
@@ -433,18 +413,12 @@ object CatalystTypeConverters {
...
-        convertedValues(i) = convertToCatalyst(value)
-        i += 1
-      }
-      ArrayBasedMapData(convertedKeys, convertedValues)
+    case map: Map[_, _] =>
+      ArrayBasedMapData(
+        map.iterator,
+        map.size,
+        (key) => convertToCatalyst(key),
+        (value) => convertToCatalyst(value))
```
gatorsmile: It just looks weird to use different apply functions in the same file. How about this?
```Scala
      ArrayBasedMapData(
        map, (key: Any) => convertToCatalyst(key), (value: Any) => convertToCatalyst(value))
```

tejasapatil: changed

gatorsmile:
LGTM pending test

gatorsmile:
Merging to master! Thanks!

PR #15575, opened 2016-10-21
[SPARK-18038] [SQL] Move output partitioning definition from UnaryNodeExec to its children

tejasapatil:

Jira : https://issues.apache.org/jira/browse/SPARK-18038

This was a suggestion by @rxin over one of the dev list discussion : http://apache-spark-developers-list.1001551.n3.nabble.com/Project-not-preserving-child-partitioning-td19417.html

His words:

> > It would be better (safer) to move the output partitioning definition into each of the operator and remove it from UnaryExecNode.

With this PR, following is the output partitioning and ordering for all the impls of `UnaryExecNode`.

| UnaryExecNode's impl | outputPartitioning | outputOrdering | comment |
| --- | --- | --- | --- |
| AppendColumnsExec | child's | Nil | child's ordering can be used |
| AppendColumnsWithObjectExec | child's | Nil | child's ordering can be used |
| BroadcastExchangeExec | BroadcastPartitioning | Nil | - |
| CoalesceExec | UnknownPartitioning | Nil | - |
| CollectLimitExec | SinglePartition | Nil | - |
| DebugExec | child's | Nil | child's ordering can be used |
| DeserializeToObjectExec | child's | Nil | child's ordering can be used |
| ExpandExec | UnknownPartitioning | Nil | - |
| FilterExec | child's | child's | - |
| FlatMapGroupsInRExec | child's | Nil | child's ordering can be used |
| GenerateExec | child's | Nil | need to dig more |
| GlobalLimitExec | child's | child's | - |
| HashAggregateExec | child's | Nil | - |
| InputAdapter | child's | child's | - |
| InsertIntoHiveTable | child's | Nil | terminal node, doesn't need partitioning |
| LocalLimitExec | child's | child's | - |
| MapElementsExec | child's | child's | - |
| MapGroupsExec | child's | Nil | child's ordering can be used |
| MapPartitionsExec | child's | Nil | child's ordering can be used |
| ProjectExec | child's | child's | - |
| SampleExec | child's | Nil | child's ordering can be used |
| ScriptTransformation | child's | Nil | child's ordering can be used |
| SerializeFromObjectExec | child's | Nil | child's ordering can be used |
| ShuffleExchange | custom | Nil | - |
| SortAggregateExec | child's | sort over grouped exprs | - |
| SortExec | child's | custom | - |
| StateStoreRestoreExec | child's | Nil | child's ordering can be used |
| StateStoreSaveExec | child's | Nil | child's ordering can be used |
| SubqueryExec | child's | child's | - |
| TakeOrderedAndProjectExec | SinglePartition | custom | - |
| WholeStageCodegenExec | child's | child's | - |
| WindowExec | child's | child's | - |
## How was this patch tested?

This does NOT change any existing functionality so relying on existing tests

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/SortExec.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/SortExec.scala
@@ -23,7 +23,7 @@ import org.apache.spark.rdd.RDD
 import org.apache.spark.sql.catalyst.InternalRow
 import org.apache.spark.sql.catalyst.expressions._
 import org.apache.spark.sql.catalyst.expressions.codegen.{CodegenContext, ExprCode}
-import org.apache.spark.sql.catalyst.plans.physical.{Distribution, OrderedDistribution, UnspecifiedDistribution}
+import org.apache.spark.sql.catalyst.plans.physical._
 import org.apache.spark.sql.execution.metric.SQLMetrics
 
 /**
@@ -44,6 +44,8 @@ case class SortExec(
   override def output: Seq[Attribute] = child.output
 
   override def outputOrdering: Seq[SortOrder] = sortOrder
+
+  override def outputPartitioning: Partitioning = child.outputPartitioning
 
   override def requiredChildDistribution: Seq[Distribution] =
     if (global) OrderedDistribution(sortOrder) :: Nil else UnspecifiedDistribution :: Nil
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/SortExec.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/SortExec.scala
@@ -45,6 +45,8 @@ case class SortExec(
 
   override def outputOrdering: Seq[SortOrder] = sortOrder
 
+  // sort performed is local within a given partition so will retain
+  // child operator's partitioning
   override def outputPartitioning: Partitioning = child.outputPartitioning
 
   override def requiredChildDistribution: Seq[Distribution] =
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/aggregate/HashAggregateExec.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/aggregate/HashAggregateExec.scala
@@ -62,6 +62,7 @@ case class HashAggregateExec(
     "aggTime" -> SQLMetrics.createTimingMetric(sparkContext, "aggregate time"))
 
   override def output: Seq[Attribute] = resultExpressions.map(_.toAttribute)
+  override def outputPartitioning: Partitioning = child.outputPartitioning
 
   override def producedAttributes: AttributeSet =
     AttributeSet(aggregateAttributes) ++
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/aggregate/HashAggregateExec.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/aggregate/HashAggregateExec.scala
@@ -62,6 +62,7 @@ case class HashAggregateExec(
     "aggTime" -> SQLMetrics.createTimingMetric(sparkContext, "aggregate time"))
 
   override def output: Seq[Attribute] = resultExpressions.map(_.toAttribute)
+
   override def outputPartitioning: Partitioning = child.outputPartitioning
 
   override def producedAttributes: AttributeSet =
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/limit.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/limit.scala
@@ -26,7 +26,6 @@ import org.apache.spark.sql.catalyst.plans.physical._
 import org.apache.spark.sql.execution.exchange.ShuffleExchange
 import org.apache.spark.util.Utils
 
-
 /**
  * Take the first `limit` elements and collect them to a single partition.
  *
@@ -96,6 +95,7 @@ trait BaseLimitExec extends UnaryExecNode with CodegenSupport {
  */
 case class LocalLimitExec(limit: Int, child: SparkPlan) extends BaseLimitExec {
   override def outputOrdering: Seq[SortOrder] = child.outputOrdering
+  override def outputPartitioning: Partitioning = child.outputPartitioning
 }
 
 /**
@@ -103,6 +103,7 @@ case class LocalLimitExec(limit: Int, child: SparkPlan) extends BaseLimitExec {
  */
 case class GlobalLimitExec(limit: Int, child: SparkPlan) extends BaseLimitExec {
   override def requiredChildDistribution: List[Distribution] = AllTuples :: Nil
+  override def outputPartitioning: Partitioning = child.outputPartitioning
 }
 
 /**
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/limit.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/limit.scala
@@ -53,8 +53,7 @@ case class CollectLimitExec(limit: Int, child: SparkPlan) extends UnaryExecNode 
 trait BaseLimitExec extends UnaryExecNode with CodegenSupport {
   val limit: Int
   override def output: Seq[Attribute] = child.output
-  override def outputOrdering: Seq[SortOrder] = child.outputOrdering
-  override def outputPartitioning: Partitioning = child.outputPartitioning
+
   protected override def doExecute(): RDD[InternalRow] = child.execute().mapPartitions { iter =>
     iter.take(limit)
   }
@@ -104,6 +103,7 @@ case class LocalLimitExec(limit: Int, child: SparkPlan) extends BaseLimitExec {
 case class GlobalLimitExec(limit: Int, child: SparkPlan) extends BaseLimitExec {
   override def requiredChildDistribution: List[Distribution] = AllTuples :: Nil
   override def outputPartitioning: Partitioning = child.outputPartitioning
+  override def outputOrdering: Seq[SortOrder] = child.outputOrdering
 }
 
 /**
@@ -122,8 +122,6 @@ case class TakeOrderedAndProjectExec(
   override def output: Seq[Attribute] = {
     projectList.map(_.toAttribute)
   }
-
-  override def outputPartitioning: Partitioning = SinglePartition
 
   override def executeCollect(): Array[InternalRow] = {
     val ord = new LazilyGeneratedOrdering(sortOrder, child.output)
@@ -160,6 +158,8 @@ case class TakeOrderedAndProjectExec(
   }
 
   override def outputOrdering: Seq[SortOrder] = sortOrder
+
+  override def outputPartitioning: Partitioning = SinglePartition
 
   override def simpleString: String = {
     val orderByString = Utils.truncatedString(sortOrder, "[", ",", "]")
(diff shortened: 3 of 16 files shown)


tejasapatil:
Jenkins test this please

sql/core/src/main/scala/org/apache/spark/sql/execution/SortExec.scala
```diff
@@ -45,6 +45,8 @@ case class SortExec(
 
   override def outputOrdering: Seq[SortOrder] = sortOrder
 
+  override def outputPartitioning: Partitioning = child.outputPartitioning
```
rxin: this depends on global, doesn't it?

cc @yhuai

sql/core/src/main/scala/org/apache/spark/sql/execution/aggregate/HashAggregateExec.scala
```diff
@@ -62,6 +62,7 @@ case class HashAggregateExec(
     "aggTime" -> SQLMetrics.createTimingMetric(sparkContext, "aggregate time"))
 
   override def output: Seq[Attribute] = resultExpressions.map(_.toAttribute)
+  override def outputPartitioning: Partitioning = child.outputPartitioning
```
rxin: add a blank line?

sql/core/src/main/scala/org/apache/spark/sql/execution/limit.scala
```diff
@@ -96,13 +95,15 @@ trait BaseLimitExec extends UnaryExecNode with CodegenSupport {
...
   override def outputOrdering: Seq[SortOrder] = child.outputOrdering
+  override def outputPartitioning: Partitioning = child.outputPartitioning
 }
 
 /**
  * Take the first `limit` elements of the child's single output partition.
  */
 case class GlobalLimitExec(limit: Int, child: SparkPlan) extends BaseLimitExec {
   override def requiredChildDistribution: List[Distribution] = AllTuples :: Nil
+  override def outputPartitioning: Partitioning = child.outputPartitioning
```
rxin: similar to sort -- i think this depends on what "child" is referring to.

yhuai: This should be `SinglePartition` since `requiredChildDistribution` is `AllTuples`.

yhuai: sorry. nvm. `child.outputPartitioning` is right.

sql/core/src/main/scala/org/apache/spark/sql/execution/window/WindowExec.scala
```diff
@@ -103,6 +103,8 @@ case class WindowExec(
 
   override def outputOrdering: Seq[SortOrder] = child.outputOrdering
 
+  override def outputPartitioning: Partitioning = child.outputPartitioning
```
yhuai: This is right. This operator does not shuffle data.

yhuai:
`outputPartitioning` of a node will only be changed if this node shuffles data. Right now, only `ShuffleExchange` shuffles data.

sql/core/src/main/scala/org/apache/spark/sql/execution/limit.scala
gatorsmile: Nit: also add an empty line here.

rxin:
@yhuai let me be more clear. It is actually fairly confusing to say "this only changes if the operator doesn't shuffle data". 

The thing is that we are relying on each operator's output partitioning to determine injecting exchanges. So is it safe to depend on "child", which is ambiguous in this context? Child could mean the current child before injecting exchange, or it could mean the injected exchange operator. Can the query planner incorrectly not injecting exchange because a plan node is deemed already having the correct output partitioning? Or is this always guaranteed to not happen due to certain ordering we enforce in planning.

viirya:
The query planner will inject exchange based on the required distribution for the child of `UnaryNodeExec`. Looks like all unary nodes have `child.outputPartitioning` as its output partition. I can't see an exception based on the fact that only `ShuffleExchange` will change `outputPartitioning`.

yhuai:
Our planner decides if to add an `ShuffleExchange` by consider `outputPartitioning` and `requiredDistribution` together. If the `outputPartitioning` of the child does not satisfy the `requiredDistribution` of the parent, we will add a `ShuffleExchange` operator.

When we say `child`, it is literally the child of a node. It does not matter if the child is a `ShuffleExchange` or not. For a node, when its `outputPartitioning` is `child. outputPartitioning`, this node does not shuffle data.

viirya:
Yeah. So I don't see any exception that an unary node, if it is not `ShuffleExchange`, can have an `outputPartitioning` other than `child.outputPartitioning`.

tejasapatil:
Agree with the planner behavior described in the last few comments (relevant code : https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/exchange/EnsureRequirements.scala#L165)

I have updated the description of the PR with a table which shows the output partitioning and ordering for each implementation of `UnaryExecNode`. It will help with the review in case we are missing something / setting something wrong.

`BroadcastExchangeExec `, `CoalesceExec `, `CollectLimitExec `, `ExpandExec`, `TakeOrderedAndProjectExec` and `ShuffleExchange` do not use child's output partitioning.

I felt that there are numerous places where child's output ordering could be used but the operators don't set it (thus using default `Nil`). I won't made those modifications in this PR as I want to only do the refac in this diff.

yhuai:
> I felt that there are numerous places where child's output ordering could be used but the operators don't set it

Can you list them at here?

tejasapatil:
@yhuai : please see the table in this PRs description. I have added a `comment` (last column) for each entry to point out those cases.

rxin:
ExpandExec should follow child's output partitioning and ordering, shouldn't it?

tejasapatil:
@rxin : Yes. There was a comment over there [0] but based on code it does not look like its moving rows across input partitions (cc @chenghao-intel who added that).

[0] : https://github.com/facebook/FB-Spark/blob/d26c42982c18da8fb1b21c9eb75aef9364d1b992/sql/core/src/main/scala/org/apache/spark/sql/execution/Expand.scala#L45

viirya:
`CoalesceExec` can also have `SinglePartition` as `outputPartitioning` too.

viirya:
But `ExpandExec`'s `GroupExpressions` could generate data with arbitrary partitioning, so even it doesn't move rows across input partitions, doesn't it still can change child' output partition?

rxin:
CoalesceExec is not SinglePartition. Coalesce can lead to an arbitrary number of partitions.

ExpandExec doesn't really change the partitioning either.

rxin:
@tejasapatil you were linking to a fb internal branch :)

viirya:
@rxin In the table in the description, `CoalesceExec` output `UnknownPartitioning`, actually it can be `SinglePartition` if what you do is `coalesce(1)`.

`ExpandExec` doesn't actually move rows across partitions as @tejasapatil pointed out. However, if its child has certain partition such as `HashPartition`, after `ExpandExec` it becomes a `UnknownPartitioning`. I am not sure if it does change the partitioning or not. From the view of output partition of the physical plan, it is changed indeed.

rxin:
@viirya of course if you say coalesce(1) it is a single partition -- any operator that changes partition to 1 partition is single partition.

For Expand isn't it just the same as Generate?

tejasapatil:
@viirya 

>> In the table in the description, CoalesceExec output UnknownPartitioning

Yes. Since partitions == 1 is a corner case, I did not put that in the table. If you look at the code, its doing the right thing: 
https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/basicPhysicalOperators.scala#L490

rxin:
The current thing LGTM. 

cc @yhuai do you have any other feedback?

tejasapatil:
@viirya

>> However, if its child has certain partition such as HashPartition, after ExpandExec it becomes a UnknownPartitioning

The notion of `Partitioning` in Spark is the distribution of rows across tasks. Even if the child's output has `HashPartitioning`, there is a 1:1 mapping among output partition of child operator and output partition of `ExpandExec`. So, applying `ExpandExec` does not alter the partitioning of child outputs.

viirya:
@tejasapatil I see there is 1:1 mapping among output partition of child operator and output partition of `ExpandExec`. 

For example we have an Expand applying on a data set like col: [1, 2, 3]. If the projections are col, col + 1, col + 2.

Assume the partition of the data set is HashPartition(col). We have three partitions:
p1: [1]
p2: [2]
p3: [3]

After the Expand, the data set becomes:

p1: [1, 2, 3]
p2: [2, 3, 4]
p3: [3, 4, 5]

Is it still valid for HashPartition(col)? Looks like it doesn't. I think It is why there is a comment on ExpandExec in the code position you links to.

BTW, in your table `ExpandExec`'s `outputPartitioning` is `UnknownPartitioning`, right? If it doesn't change child's partition, why we don't set it to child's outputPartitioning?

viirya:
@rxin yeah, I am curious why `ExpandExec` and `GenerateExec` have different `outputPartitioning`...

tejasapatil:
@viirya : As per my understanding, if the child operator emits `col`, after applying `ExpandExec`, the output is `col'`. The original child partitioning being over `col`, `ExpandExec` does not seem to alter that. 

The table above was to summarise the state of things before this PR. I did not change any semantics in this PR as its pure refactoring.

viirya:
@tejasapatil yeah, that is correct. however, I am wondering if we can say this `ExpandExec` have the same distribution of rows as its child...because it even doesn't have the `col`...

viirya:
In practice, setting the `outputPartitioning` of a physical plan like `ExpandExec` to `child.outputPartitioning` doesn't cause any real problem, even this physical plan doesn't keep the same row distribution of its child. That is because if the physical plan changes output, it will have different output attributes, e.g., `col` to `col'` as @tejasapatil pointed out.

If its parent plan requires a distribution, says `HashPartition`, this distribution will bound to the physical plan's output `col'`, instead of its child plan's `col`. So even the physical plan uses `child.outputPartitioning`, `EnsureRequirements ` will step in and inject an extra shuffle exchange of `HashPartition(col')` to satisfy the requirement.

It works like that as per my understanding. However it doesn't mean the physical plan's output partitioning is exactly as same as its child's, i.e., `HashPartition(col)`, because it doesn't have the output `col`. This part might be confusing to some people, so I think it might be better to explain it more. That is what I understood about this, if I am wrong please kindly point out.

gatorsmile:
LGTM, since the scope of this PR is just refactoring. 

Let me first post the existing code for `outputPartitioning` in `ExpandExec`:
```Scala
  // The GroupExpressions can output data with arbitrary partitioning, so set it
  // as UNKNOWN partitioning
  override def outputPartitioning: Partitioning = UnknownPartitioning(0)
```

It makes sense to set it either `UnknownPartitioning` or `child.outputPartitioning`. However, the above code is setting to a wrong number of partitions. We need to correct it no matter whether we are using the number or not.

rxin:
LGTM too. Unforunately my internet sucks (on a plane) and I can't merge this right now.

viirya:
yeah, LGTM, it doesn't change current outputPartitioning of operators.

hvanhovell:

jaceklaskowski:
I'm late with this, but just leaving it for future code reviewers...

I think the change took the most extreme path where even such simple `outputPartitioning` as the one in `WindowExec` (see below) is currently part of the extension of `UnaryExecNode` while I think it should stay with `UnaryExecNode` and be overridden when the output partitioning is different. It's a kind of code duplication.

```
  override def outputPartitioning: Partitioning = child.outputPartitioning
```

The above serves nothing but says _"I simply don't care what happens upstream"_.

PR #15560, opened 2016-10-20
[SPARKR] fix warnings

felixcheung:
## What changes were proposed in this pull request?

Fix for a bunch of test warnings that were added recently.
We need to investigate why warnings are not turning into errors.

```
Warnings -----------------------------------------------------------------------
1. createDataFrame uses files for large objects (@test_sparkSQL.R#215) - Use Sepal_Length instead of Sepal.Length  as column name

2. createDataFrame uses files for large objects (@test_sparkSQL.R#215) - Use Sepal_Width instead of Sepal.Width  as column name

3. createDataFrame uses files for large objects (@test_sparkSQL.R#215) - Use Petal_Length instead of Petal.Length  as column name

4. createDataFrame uses files for large objects (@test_sparkSQL.R#215) - Use Petal_Width instead of Petal.Width  as column name

Consider adding
  importFrom("utils", "object.size")
to your NAMESPACE file.
```
## How was this patch tested?

unit tests

PR #15576, opened 2016-10-21
[SPARK-17674][SPARKR] check for warning in test output

felixcheung:

testthat library we are using for testing R is redirecting warning (and disabling `options("warn" = 2)`), we need to have a way to detect any new warning and fail
## How was this patch tested?

manual testing, Jenkins


felixcheung:
Test failure is intentional, it's picking up the following warnings:

```
Warnings -----------------------------------------------------------------------
1. createDataFrame uses files for large objects (@test_sparkSQL.R#215) - Use Sepal_Length instead of Sepal.Length  as column name

2. createDataFrame uses files for large objects (@test_sparkSQL.R#215) - Use Sepal_Width instead of Sepal.Width  as column name

3. createDataFrame uses files for large objects (@test_sparkSQL.R#215) - Use Petal_Length instead of Petal.Length  as column name

4. createDataFrame uses files for large objects (@test_sparkSQL.R#215) - Use Petal_Width instead of Petal.Width  as column name
```

Which is fixed in PR #15560

felixcheung:
Rebased. This should pass now.

wangmiao1981:
Thanks for fixing this! I encountered this issue before.

felixcheung:
merged to master, thanks

PR #14529, opened 2016-08-08
[TRIVIAL][SQL] Match the name of OrcRelation companion object to OrcFileFormat

HyukjinKwon:
## What changes were proposed in this pull request?

This PR simply matches the name `OrcRelation` to `OrcFileFormat`.

Just to make sure, `ParquetFileFormat` has the object named `ParquetFileFormat` and the related ones were renamed in https://github.com/apache/spark/commit/361ebc282b2d09dc6dcf21419a53c5c617b1b6bd all together but it seems renaming`OrcFileFormat` was missed.
## How was this patch tested?

N/A

PR #15580, opened 2016-10-21
[SPARK-18042][SQL] OutputWriter should expose file path written

rxin:

This patch adds a new "path" method on OutputWriter that returns the path of the file written by the OutputWriter. This is part of the necessary work to consolidate structured streaming and batch write paths.

The batch write path has a nice feature that each data source can define the extension of the files, and allow Spark to specify the staging directory and the prefix for the files. However, in the streaming path we need to collect the list of files written, and there is no interface right now to do that.
## How was this patch tested?

N/A - there is no behavior change and this should be covered by existing tests.

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/OutputWriter.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/OutputWriter.scala
@@ -42,11 +42,12 @@ abstract class OutputWriterFactory extends Serializable {
    * @param fileNamePrefix Prefix of the file name. The returned OutputWriter must make sure this
    *                       prefix is used in the actual file name. For example, if the prefix is
    *                       "part-1-2-3", then the file name must start with "part_1_2_3" but can
-   *                       end in arbitrary extension.
+   *                       end in arbitrary extension that is deterministic given the configuration
+   *                       (i.e. the suffix extension should not depend on any task id, attempt id,
+   *                       or partition id).
    * @param dataSchema Schema of the rows to be written. Partition columns are not included in the
    *        schema if the relation being written is partitioned.
    * @param context The Hadoop MapReduce task context.
-   * @since 1.4.0
    */
   def newInstance(
       stagingDir: String,
@@ -62,7 +63,6 @@ abstract class OutputWriterFactory extends Serializable {
    * and not modify it (do not add subdirectories, extensions, etc.). All other
    * file-format-specific information needed to create the writer must be passed
    * through the [[OutputWriterFactory]] implementation.
-   * @since 2.0.0
    */
   def newWriter(path: String): OutputWriter = {
     throw new UnsupportedOperationException("newInstance with just path not supported")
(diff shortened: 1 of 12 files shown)


rxin:
cc @hvanhovell @cloud-fan 

and @ericl

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetOutputWriter.scala
```diff
@@ -80,15 +81,15 @@ private[parquet] class ParquetOutputWriterFactory(
...
+    conf.set(ParquetOutputFormat.COMPRESSION, parquetOptions.compressionCodecClassName)
     new SerializableConfiguration(conf)
   }
 
   /**
    * Returns a [[OutputWriter]] that writes data to the give path without using
    * [[OutputCommitter]].
    */
-  override def newWriter(path: String): OutputWriter = new OutputWriter {
+  override def newWriter(path1: String): OutputWriter = new OutputWriter {
```
cloud-fan: how about `_path`? `path1` looks weird...

cloud-fan: or maybe we can create a class for this `OutputWriter` here.

rxin:
cc @HyukjinKwon

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/OutputWriter.scala
```diff
@@ -77,19 +77,22 @@ abstract class OutputWriterFactory extends Serializable {
  * executor side.  This instance is used to persist rows to this single output file.
  */
 abstract class OutputWriter {
+
+  /**
+   * The path of the file to be written out. This path should include the staging directory and
+   * the file name prefix passed into the associated createOutputWriter function.
+   */
+  def path: String
```
ericl: `fullOutputPath`?

rxin:
Thanks - I'm going to merge this. I will address the doc and naming comment in the next pr in this series.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetOptions.scala
```diff
@@ -35,7 +35,7 @@ private[parquet] class ParquetOptions(
    * Compression codec to use. By default use the value specified in SQLConf.
    * Acceptable values are defined in [[shortParquetCompressionCodecNames]].
    */
-  val compressionCodec: String = {
+  val compressionCodecClassName: String = {
```
HyukjinKwon: @rxin This is a super minor but there are the same options to potentially rename in `OrcOptions`, `JsonOptions`, `CSVOptions` and `TextFileFormat` - [TextFileFormat.scala#L71](https://github.com/apache/spark/blob/3fbf5a58c236fc5d5fee39cb29e7f5c7e01c0ee7/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/text/TextFileFormat.scala#L71).

Also, I'd like to note, just in case, that the value in this is actually not the class name (it's something like `SNAPPY` and `LZO`) in case of `ParquetOptions` and `OrcOptions` whereas text-based ones are actual class names.

HyukjinKwon:
@rxin BTW, could I please ask to include https://github.com/apache/spark/pull/14529 in your future related PRs? I will close if you will.

rxin:
@HyukjinKwon sure.

PR #15583, opened 2016-10-21
[SPARK-18047][Core] Spark worker port should be greater than 1023

darionyaphet:

The port numbers in the range from 0 to 1023 are the well-known ports (system ports) .

They are widely used by system network services. Such as Telnet(23), Simple Mail Transfer Protocol(25) and Domain Name System(53).

Work port should avoid using this ports .
## How was this patch tested?

None


srowen:
I disagree with this change, per the JIRA

darionyaphet:
Could you please point out the reason why?

srowen:
(See JIRA)

HyukjinKwon:
+1 for not a problem if I understood this correctly.

PR #15587, opened 2016-10-21
[SPARK-18051][Spark Core] fix bug of custom PartitionCoalescer causing serialization exception

WeichenXu123:

add a require check in `CoalescedRDD` to make sure the passed in `partitionCoalescer` to be `serializable`.
and update the document for api `RDD.coalesce`
## How was this patch tested?

Manual.(test code in jira [SPARK-18051])


WeichenXu123:
@rxin How about this way: modify the parameter as `Option[PartitionerCoalescer with Serializable]`, so that if user forget to mark the implementation class as serializable, compiling time will print the error rather than the runtime throw some unreadable exceptions... what do you think about it ?

rxin:
Is this an API breaking change? 

How about we just add a require check in CoalesceRDD to verify the passed in PartitionCoalescer  is serializable?

WeichenXu123:
@rxin Good suggestion. code & document updated, thanks!

rxin:
Merging in master. Thanks!

PR #15591, opened 2016-10-21
[SPARK-17922] [SQL] ClassCastException ..GeneratedClass$GeneratedIterator cannot be cast to ...expressions.UnsafeProjection

amadav:

Janino compiler internally creates a byteclassloader (http://grepcode.com/file/repo1.maven.org/maven2/org.codehaus.janino/janino/2.5.15/org/codehaus/janino/ByteArrayClassLoader.java#ByteArrayClassLoader) to load the compiled generated class. But this class loader doesnot override load class to load the class from byte array for the generated class. Instead the call first goes to parent class loader and if somehow the classloader finds the old generatedClass( all the generated class names are same) it will incorrectly load the old generated class. This class loader will be used to intercept delegation to parent if the class has to be loaded by the current byte class loader. This will be set as the parent class loader for janino compiler in CodeGenerator.doCompile.

Special classloader to skip delegating to parent class loader when the class name is same as the generated class name. Because that class should be loaded by the current class loader 
## How was this patch tested?

manual tests

### DIFF
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/DelegateClassLoader.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/DelegateClassLoader.scala
@@ -1,0 +1,37 @@
+package org.apache.spark.sql.catalyst.util
+/*
+ * See : https://issues.apache.org/jira/browse/SPARK-17922. 
+ * Janino compiler internally creates a byteclassloader (http://grepcode.com/file/repo1.maven.org/maven2/org.codehaus.janino/janino/2.5.15/org/codehaus/janino/ByteArrayClassLoader.java#ByteArrayClassLoader)
+ * to load the compiled generated class. But this class loader doesnot override load class to load the class from byte array for the generated class.
+ * Instead the call first goes to parent class loader and if somehow the classloader finds the old generatedClass( all the generated class names are same)
+ * it will incorrectly load the old generated class. This class loader will be used to intercept delegation to parent if the class has to be loaded by the current byte class loader.
+ * This will be set as the parent class loader for janino compiler in CodeGenerator.doCompile
+ * Special classloader to skip delegating to parent class loader when the class name is same as the generated class name.
+ * Because that class should be loaded by the current class loader 
+ */
+class DelegateClassLoader(parent:ClassLoader, skipClass: String) extends ClassLoader(parent) {
+  override def findClass(name: String): Class[_] = {
+    if(checkClassName(name)) {
+      return null
+    }
+    super.findClass(name)
+  }
+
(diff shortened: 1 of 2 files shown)


HyukjinKwon:
@AbhiMadav I think we need a test to confirm this PR fixes that problem and avoid the regression in the future.

AbhiMadav:
@yhuai could you please review? It is seemingly not very easy to get a test setup for this case alone. Would appreciate your advice on the same.

AbhiMadav:
Can one of the admins verify this patch?

rsotnychenko:
@yhuai @davies maybe you can review this patch?

AbhiMadav:
Added a repro case in the JIRA https://issues.apache.org/jira/browse/SPARK-17922

Devian-ua:
@srowen @gatorsmile can you review this patch?

srowen:
I don't know this code well enough to review, but this looks like a hacky solution, to make a classloader ignore one class. I would not imagine this is the answer.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/DelegateClassLoader.scala
```diff
@@ -0,0 +1,37 @@
+package org.apache.spark.sql.catalyst.util
+/*
+ * See : https://issues.apache.org/jira/browse/SPARK-17922. 
+ * Janino compiler internally creates a byteclassloader (http://grepcode.com/file/repo1.maven.org/maven2/org.codehaus.janino/janino/2.5.15/org/codehaus/janino/ByteArrayClassLoader.java#ByteArrayClassLoader)
+ * to load the compiled generated class. But this class loader doesnot override load class to load the class from byte array for the generated class.
+ * Instead the call first goes to parent class loader and if somehow the classloader finds the old generatedClass( all the generated class names are same)
+ * it will incorrectly load the old generated class. This class loader will be used to intercept delegation to parent if the class has to be loaded by the current byte class loader.
+ * This will be set as the parent class loader for janino compiler in CodeGenerator.doCompile
+ * Special classloader to skip delegating to parent class loader when the class name is same as the generated class name.
+ * Because that class should be loaded by the current class loader
```
kiszk: What do you mean "the current class loader"? Application class loader? System class loader? Other class loaders?

kdhuria1: By current class loader here ,we mean the byteclassloader used by Janino compiler.

gatorsmile:
ok to test

gatorsmile:
@AbhiMadav add a test case?

HyukjinKwon:
Closing this PR for the author being inactive.

PR #15596, opened 2016-10-22
[SPARK-18089][SQL] Remove shuffle codes in CollectLimitExec

viirya:

Currently, `CollectLimitExec` is an operator used when the logical Limit is the last operator in a logical plan. In fact, the job of `CollectLimitExec` is not different to `GlobalLimitExec`. We can do little refactoring to `GlobalLimitExec` and replace `CollectLimitExec`.
## How was this patch tested?

Jenkins tests.


viirya:
cc @rxin @yhuai May you take a quick look to see if this direction is ok for you? Thanks!

sql/core/src/main/scala/org/apache/spark/sql/execution/limit.scala
```diff
@@ -83,9 +83,8 @@ trait BaseLimitExec extends UnaryExecNode with CodegenSupport {
     s"""
        | if ($countTerm < $limit) {
        |   $countTerm += 1;
+       |   if ($countTerm == $limit) $stopEarly = true;
```
viirya: Set stop early flag to true so we won't step into next element in iterator. Otherwise, we still go to get next element from the iterator. So if the limitation is 1, we will get 2 elements from the iterator.

Note: This won't cause real problem because we have an `if` to guard against processing element.

viirya:
BTW, we can see if there is an exchange added for `CollectLimit` from physical plan, e.g.,

    CollectLimit 1
    +- Exchange SinglePartition
      +- *LocalLimit 1
         +- *HashAggregate(keys=[str#227], functions=[count(1)], output=[str#227, count(1)#235L])
            +- Exchange hashpartitioning(str#227, 5)
               +- *HashAggregate(keys=[str#227], functions=[partial_count(1)], output=[str#227, count#241L])
                  +- *Project [str#227]
                     +- *BroadcastHashJoin [str#227], [str#233], Inner, BuildRight
                        :- *Project [_2#224 AS str#227]
                        :  +- *Filter isnotnull(_2#224)
                        :     +- LocalTableScan [_1#223, _2#224]
                        +- BroadcastExchange HashedRelationBroadcastMode(List(input[0, string, true]))
                           +- *Project [_2#224 AS str#233]
                              +- *Filter isnotnull(_2#224)
                                 +- LocalTableScan [_1#223, _2#224]

sql/core/src/main/scala/org/apache/spark/sql/execution/limit.scala
```diff
@@ -36,14 +36,14 @@ import org.apache.spark.util.Utils
...
-  override def executeCollect(): Array[InternalRow] = child.executeTake(limit)
-  private val serializer: Serializer = new UnsafeRowSerializer(child.output.size)
+  override def requiredChildDistribution: List[Distribution] = AllTuples :: Nil
+  override def executeCollect(): Array[InternalRow] = child match {
+    case e: Exchange => e.child.executeTake(limit)
+    case _ => child.executeTake(limit)
+  }
+
   protected override def doExecute(): RDD[InternalRow] = {
-    val locallyLimited = child.execute().mapPartitionsInternal(_.take(limit))
```
hvanhovell: We are removing an optimization here right? We can greatly reduce the number of shuffled records by applying the limit before anything gets shuffled.

viirya: One advantage is `LocalLimit` supports whole stage codegen. We can also easily get the idea of this optimization from physical plans.

hvanhovell:
@JoshRosen could you shed some light on why we are not using the regular `EnsureRequirements` based code path for `CollectLimitExec`?

sql/core/src/main/java/org/apache/spark/sql/execution/BufferedRowIterator.java
```diff
@@ -39,7 +39,7 @@
   protected int partitionIndex = -1;
 
   public boolean hasNext() throws IOException {
-    if (currentRows.isEmpty()) {
+    if (!shouldStop()) {
```
viirya: I revert this change as it could cause problem in releasing memory.

sql/core/src/main/scala/org/apache/spark/sql/execution/aggregate/HashAggregateExec.scala
```diff
@@ -661,6 +661,15 @@ case class HashAggregateExec(
          """.stripMargin
     }
 
+    ctx.addNewFunction("releaseResource", s"""
```
viirya: We do need to release memory by calling these functions. Because we stop iterating the data early by letting `shouldStop` return true, we won't call next `processNext` and memory leak will happen. So we wrap these calls in override `releaseResource` function.

sql/core/src/main/scala/org/apache/spark/sql/execution/LocalTableScanExec.scala
```diff
@@ -47,6 +50,22 @@ case class LocalTableScanExec(
 
   private lazy val rdd = sqlContext.sparkContext.parallelize(unsafeRows, numParallelism)
 
+  protected override def doProduce(ctx: CodegenContext): String = {
```
viirya: Let `LocalTableScanExec` support whole stage codegen.

Because `CollectLimitExec` now supports whole stage codegen, the test in `SQLMetricsSuite`:

    val df2 = spark.createDataset(Seq(1, 2, 3)).limit(2)
    df2.collect()
    val metrics2 = df2.queryExecution.executedPlan.collectLeaves().head.metrics
    assert(metrics2.contains("numOutputRows"))
    assert(metrics2("numOutputRows").value === 2)

will execute the `LocalTableScanExec` node to get its RDD. Then an InputAdapter will connect it to `CollectLimitExec`'s whole stage codegen node. So it will output all 3 rows in the local table.

Adding this whole stage code support seems straightforward. So I adds it here to pass the tests.

viirya:
hmmm, I don't image that it is needed for such related changes to remove shuffle codes in CollectLimitExec.

This is not a serious issue and if you think this change is too big for the purpose, please let me know. Thanks!

JoshRosen:
For a bit of background on `CollectLimitExec`, note that the intent of this operator is to be able to optimize the special case where a limit is the terminal operator of a plan: in this case, we don't need to perform a shuffle because we can have the driver run multiple jobs which scan an increasingly large portion of the RDD to get the limited items; in a nutshell, the goal is to allow a logic similar to the RDD `take()` action to stop early without having to compute all partitions of the limited RDD.

In typical operation, we don't necessarily expect `CollectLimitExec` to appear in the middle of a query plan, so `CollectLimitExec.execute()` should generally only be called in special cases such as calling `.rdd()` on a limited RDD then performing further operations on it. This is why I didn't use `EnsureRequirements` here: if we did, then we'd end up shuffling all limited partitions to a single non-driver partition, then limiting that and collecting to the driver, degrading performance in the case where a limit is the terminal operation in the query plan.

sql/core/src/main/scala/org/apache/spark/sql/execution/limit.scala
```diff
@@ -36,14 +36,15 @@ import org.apache.spark.util.Utils
 case class CollectLimitExec(limit: Int, child: SparkPlan) extends UnaryExecNode {
   override def output: Seq[Attribute] = child.output
   override def outputPartitioning: Partitioning = SinglePartition
-  override def executeCollect(): Array[InternalRow] = child.executeTake(limit)
-  private val serializer: Serializer = new UnsafeRowSerializer(child.output.size)
+  override def requiredChildDistribution: List[Distribution] = AllTuples :: Nil
+  override def executeCollect(): Array[InternalRow] = {
+    child.collect {
```
JoshRosen: I think that I understand the intent here: it looks like the idea is to let `EnsureRequirements` handle the shuffling, if necessary, then to strip off the shuffling and walk backwards in the DAG to optimize the `collect()` case. This is pretty clever.

One concern of mine, though, is that there seems to be an implicit assumption on the types of children that this operator can have and I think it would be a good idea to write them down and pattern-match more explicitly. For example, I'm not sure that this strategy is safe in case the child isn't an exchange because then you run the risk of simply dropping any operators that occur post-exchange. I understand that this case can't crop up in the types of plans that we currently generate but it would be good to future-proof by matching explicitly rather than having the less-constrained `collect` behavior here.

JoshRosen: Also, one other idea: if you're going to try this strategy, what do you think about putting this logic into `GlobalLimitExec` and removing the `CollectLimitExec` planning logic?

viirya:
retest this please.

viirya:
> In typical operation, we don't necessarily expect CollectLimitExec to appear in the middle of a query plan, so CollectLimitExec.execute() should generally only be called in special cases such as calling .rdd() on a limited RDD then performing further operations on it. This is why I didn't use EnsureRequirements here: if we did, then we'd end up shuffling all limited partitions to a single non-driver partition, then limiting that and collecting to the driver, degrading performance in the case where a limit is the terminal operation in the query plan.

Yeah, I think I understand the intent. So in the `executeCollect`, I do strip off the shuffling if needed.

sql/core/src/main/scala/org/apache/spark/sql/execution/WholeStageCodegenExec.scala
```diff
@@ -298,6 +298,12 @@ case class WholeStageCodegenExec(child: SparkPlan) extends UnaryExecNode with Co
   override def outputPartitioning: Partitioning = child.outputPartitioning
 
   override def outputOrdering: Seq[SortOrder] = child.outputOrdering
+  override def executeCollect(): Array[InternalRow] = child match {
+    // This happens when the user is collecting results back to the driver, we could skip
+    // the shuffling and scan increasingly the RDD to get the limited items.
+    case g: GlobalLimitExec => g.executeCollect()
```
rxin: this is really confusing ...

viirya: Still think this is confusing as you said. Removed it.

viirya:
@JoshRosen After few tries, I think to replace `CollectLimitExec` with `GlobalLimitExec` is not a good idea.

The main reason is whole stage codegen. Since `GlobalLimitExec` supports whole stage codegen, it will be wrapped in a `WholeStageCodegenExec`. So we will call `executeCollect()` on `WholeStageCodegenExec` wrapping `GlobalLimitExec` when we do `collect()` on `df.limit(1).collect()`, for example.

`WholeStageCodegenExec.executeCollect()` is `SparkPlan.executeCollect()` actually. So we will do shuffling and retrieve the results. It doesn't harm to anything, but fails few tests, as the Jenkins test results showed.

Of course we can change the tests to fit it. But I don't think it is necessary and good way to do.

Another workaround is to override `WholeStageCodegenExec.executeCollect()`. But as @rxin pointed out in previous comment, it is confusing.

So based on such facts, I think we better keep `CollectLimitExec` but just remove its shuffling code as I did in initial commit.

What do you think?

sql/core/src/main/scala/org/apache/spark/sql/execution/limit.scala
```diff
@@ -27,32 +27,14 @@ import org.apache.spark.sql.execution.exchange.ShuffleExchange
...
-
-/**
  * Helper trait which defines methods that are shared by both
  * [[LocalLimitExec]] and [[GlobalLimitExec]].
  */
 trait BaseLimitExec extends UnaryExecNode with CodegenSupport {
   val limit: Int
   override def output: Seq[Attribute] = child.output
+  override def executeCollect(): Array[InternalRow] = child.executeTake(limit)
+  override def executeTake(n: Int): Array[InternalRow] = child.executeTake(limit)
```
viirya: Thanks @pwoody! Agreed. But I am thinking not to replace `CollectLimitExec` with `GlobalLimitExec`. The reason is commented below. Let's wait for @JoshRosen's response. If we decide to keep `CollectLimitExec`, your change at #15614 can be applied then.

viirya:
I'd close this now.

PR #15598, opened 2016-10-22
[SPARK-18027][YARN] .sparkStaging not clean on RM ApplicationNotFoundException

srowen:

Cleanup YARN staging dir on all `KILLED`/`FAILED` paths in `monitorApplication`
## How was this patch tested?

Existing tests

### DIFF
--- a/yarn/src/main/scala/org/apache/spark/deploy/yarn/Client.scala
+++ b/yarn/src/main/scala/org/apache/spark/deploy/yarn/Client.scala
@@ -1063,7 +1063,7 @@ private[spark] class Client(
             return (YarnApplicationState.KILLED, FinalApplicationStatus.KILLED)
           case NonFatal(e) =>
             logError(s"Failed to contact YARN for application $appId.", e)
-            cleanupStagingDir(appId)
+            // Don't necessarily clean up staging dir because status is unknown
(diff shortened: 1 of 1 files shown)


yarn/src/main/scala/org/apache/spark/deploy/yarn/Client.scala
```diff
@@ -1059,9 +1059,11 @@ private[spark] class Client(
         } catch {
           case e: ApplicationNotFoundException =>
             logError(s"Application $appId not found.")
+            cleanupStagingDir(appId)
             return (YarnApplicationState.KILLED, FinalApplicationStatus.KILLED)
           case NonFatal(e) =>
```
jerryshao: One scenario related to this issue might be connection failure between Client and AM in yarn cluster mode, in such scenario it is not safe to remove staging dir, probably Spark application is still running and it is OK local client is failed early.

srowen: OK I think you answered by question on JIRA then. I'll remove this.

srowen:
Merged to  master. I'm not against merging to 2.0 but wasn't sure how essential this is as a fix.

PR #15599, opened 2016-10-22
[SPARK-18022][SQL] java.lang.NullPointerException instead of real exception when saving DF to MySQL

srowen:

On null next exception in JDBC, don't init it as cause or suppressed
## How was this patch tested?

Existing tests

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
@@ -607,7 +607,7 @@ object JdbcUtils extends Logging {
     } catch {
       case e: SQLException =>
         val cause = e.getNextException
-        if (e.getCause != cause) {
+        if (cause != null && e.getCause != cause) {
           if (e.getCause == null) {
             e.initCause(cause)
           } else {


gatorsmile:
If we do not have a test case, could you post the exception before the fix and the exception output after the fix? Thanks!

maver1ck:
I can try this fix on Monday.

srowen:
I think we don't know exactly what the real exception is. The NPE occurs while handling the exception.

srowen:
@maver1ck @gatorsmile are you OK with this change, at least? I feel like I'm missing the additional change you have in mind. I think the purpose here is just to avoid the NPE during exception handling.

gatorsmile:
@maver1ck Could you please try to run the fix? Thank you!

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
```diff
@@ -607,7 +607,7 @@ object JdbcUtils extends Logging {
     } catch {
       case e: SQLException =>
         val cause = e.getNextException
-        if (e.getCause != cause) {
+        if (cause != null && e.getCause != cause) {
```
viirya: This looks correct as `addSuppressed(null)` will throw NPE. However, it might be hard to create a test for it...

srowen:
Merged to master, and to 2.0 as a small clear fix. If there is a follow-on problem involving whatever exception this suppressed, I think that's separate.

shatestest:
@uzadude sir I am getting same error in spark-sql 2.3.1 version
https://stackoverflow.com/questions/53011256/how-to-solve-this-error-org-apache-spark-sql-catalyst-errors-packagetreenodeexc/53030016?noredirect=1#comment92965464_53030016

https://gist.github.com/shatestest/86ae9559c6114999e13a6eec3c80ec2b#file-driver-scala

is there anything wrong am doing here ...

uzadude:
> @uzadude sir I am getting same error in spark-sql 2.3.1 version
I don't think you meant to approach me..

gatorsmile:
@shatestest Your problem is different from the issue this PR tries to resolve. If you can provide a test case to reproduce it, feel free to open a JIRA

PR #15603, opened 2016-10-23
[SPARK-18204][WEBUI] Remove SparkUI.appUIAddress

jaceklaskowski:

Removing `appUIAddress` attribute since it is no longer in use.
## How was this patch tested?

Local build

### DIFF
--- a/core/src/main/scala/org/apache/spark/ui/SparkUI.scala
+++ b/core/src/main/scala/org/apache/spark/ui/SparkUI.scala
@@ -94,15 +94,8 @@ private[spark] class SparkUI private (
   /** Stop the server behind this web interface. Only valid after bind(). */
   override def stop() {
     super.stop()
-    logInfo(s"Stopped Spark web UI at $appUIAddress")
+    logInfo(s"Stopped Spark web UI at $webUrl")
   }
-
-  /**
-   * Return the application UI host:port. This does not include the scheme (http://).
-   */
-  private[spark] def appUIHostPort = publicHostName + ":" + boundPort
-
-  private[spark] def appUIAddress = webUrl
 
   def getSparkUI(appId: String): Option[SparkUI] = {
     if (appId == this.appId) Some(this) else None
--- a/core/src/main/scala/org/apache/spark/ui/SparkUI.scala
+++ b/core/src/main/scala/org/apache/spark/ui/SparkUI.scala
@@ -82,7 +82,7 @@ private[spark] class SparkUI private (
   initialize()
 
   def getSparkUser: String = {
-    environmentListener.systemProperties.toMap.get("user.name").getOrElse("<unknown>")
+    environmentListener.systemProperties.toMap.getOrElse("user.name", "<unknown>")
   }
 
   def getAppName: String = appName
@@ -94,7 +94,7 @@ private[spark] class SparkUI private (
   /** Stop the server behind this web interface. Only valid after bind(). */
   override def stop() {
     super.stop()
-    logInfo("Stopped Spark web UI at %s".format(appUIAddress))
+    logInfo(s"Stopped Spark web UI at $appUIAddress")
   }
 
   /**
@@ -102,7 +102,7 @@ private[spark] class SparkUI private (
    */
   private[spark] def appUIHostPort = publicHostName + ":" + boundPort
 
-  private[spark] def appUIAddress = s"http://$appUIHostPort"
+  private[spark] def appUIAddress = webUrl
 
   def getSparkUI(appId: String): Option[SparkUI] = {
     if (appId == this.appId) Some(this) else None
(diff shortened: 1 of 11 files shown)


jaceklaskowski:
The build has just proved that my thinking was correct. I think I'll propose few other (more agressive) changes to clean the code up a little bit more. I'd appreciate any comments (or acceptance).

core/src/main/scala/org/apache/spark/ui/SparkUI.scala
```diff
@@ -93,15 +91,15 @@ private[spark] class SparkUI private (
...
+    logInfo(s"Stopped Spark web UI at $appUIAddress")
   }
 
   /**
    * Return the application UI host:port. This does not include the scheme (http://).
    */
   private[spark] def appUIHostPort = publicHostName + ":" + boundPort
 
-  private[spark] def appUIAddress = s"http://$appUIHostPort"
+  private[spark] def appUIAddress = webUrl
```
srowen: I think these two methods can just be eliminated, because appUIAddress does seem to return exactly the same thing, and appUIHostPort is only referenced from tests.

core/src/main/scala/org/apache/spark/ui/SparkUI.scala
```diff
@@ -135,7 +133,7 @@ private[spark] class SparkUI private (
 private[spark] abstract class SparkUITab(parent: SparkUI, prefix: String)
   extends WebUITab(parent, prefix) {
 
-  def appName: String = parent.getAppName
+  def appName: String = parent.appName
```
srowen: This can just be removed then, because this becomes a no-op. And then getAppName in the superclass can be removed. It's a pointless getter.

srowen (changes requested):
Some of this, I wouldn't change randomly for its own sake. Removing the bit of dead code is reasonable if you take it a step further.

core/src/main/scala/org/apache/spark/ui/SparkUI.scala
jaceklaskowski: Yes! That's how I imagined the conversation would go. Thanks Sean!

jaceklaskowski: I might have missed why we could remove it. Mind refreshing my memory @srowen (sorry getting too old :))

srowen: Actually, leave this one. It does serve a purpose.

core/src/test/scala/org/apache/spark/ui/UISuite.scala
```diff
@@ -179,16 +179,15 @@ class UISuite extends SparkFunSuite {
   test("verify appUIAddress contains the scheme") {
```
srowen: You might also change occurrences of "appUIAddress" in places like this test name. Have a look with a search and replace.

srowen (changes requested):
Otherwise looks good yes.

core/src/test/scala/org/apache/spark/ui/UISuite.scala
jaceklaskowski: Thanks Sean! How do you manage to find so much time to help people like me?!?! You're certainly not a robot -- I do know it since I met you few times and can confirm ;-)

jaceklaskowski: What do you think about renaming `spark.driver.appUIAddress` to `spark.driver.webUrl`? Not necessarily important, but merely for the sake of consistency...

srowen (approved):

core/src/test/scala/org/apache/spark/ui/UISuite.scala
```diff
@@ -176,21 +176,20 @@ class UISuite extends SparkFunSuite {
...
 
-  test("verify appUIAddress contains the port") {
+  test("verify webUrl contains the port") {
     withSpark(newSparkContext()) { sc =>
       val ui = sc.ui.get
-      val splitUIAddress = ui.appUIAddress.split(':')
+      val splitUIAddress = ui.webUrl.split(':')
       val boundPort = ui.boundPort
-      assert(splitUIAddress(2).toInt == boundPort)
+      assert(splitUIAddress(3).toInt == boundPort)
```
srowen: I think this actually should not change. The test fails and the behavior should be identical anyway.

jaceklaskowski:
The test failure doesn't seem to be due to my change.

jaceklaskowski:
Anything else you'd change/remove/add, Mr @srowen?

srowen:
Since this turned into a bit more of a change than where it started, and because it removes a method (internal, defunct as it may be), and since this isn't a hurry, would you open a JIRA just to track it for the record?

PR #15326, opened 2016-10-02
[SPARK-17759] [CORE] Avoid adding duplicate schedulables

erenavsarogullari:
## What changes were proposed in this pull request?

If `spark.scheduler.allocation.file` has duplicate pools, all of them are created when `SparkContext` is initialized but just one of them is used and the other ones look **redundant**. This causes redundant pool creation and needs to be fixed. 

**Code to Reproduce** :

```
val conf = new SparkConf().setAppName("spark-fairscheduler").setMaster("local")
conf.set("spark.scheduler.mode", "FAIR")
conf.set("spark.scheduler.allocation.file", "src/main/resources/fairscheduler-duplicate-pools.xml")
val sc = new SparkContext(conf)
```

**fairscheduler-duplicate-pools.xml** :

The following sample just shows two **default** and **duplicate_pool1** but this also needs to be thought for **N** default and/or other duplicate pools.

```
<allocations>
    <pool name="default">
        <minShare>0</minShare>
        <weight>1</weight>
        <schedulingMode>FIFO</schedulingMode>
    </pool>
    <pool name="default">
        <minShare>0</minShare>
        <weight>1</weight>
        <schedulingMode>FIFO</schedulingMode>
    </pool>
    <pool name="duplicate_pool1">
        <minShare>1</minShare>
        <weight>1</weight>
        <schedulingMode>FAIR</schedulingMode>
    </pool>
    <pool name="duplicate_pool1">
        <minShare>2</minShare>
        <weight>2</weight>
        <schedulingMode>FAIR</schedulingMode>
    </pool>
</allocations>
```

**Debug Screenshot** :
The following screenshots show `Pool.schedulableQueue(ConcurrentLinkedQueue[Schedulable])` has **4** pools as 

> default, default, duplicate_pool1 and duplicate_pool1

 but `Pool.schedulableNameToSchedulable(ConcurrentHashMap[String, Schedulable])` has 

> default and duplicate_pool1

due to pool name as key so one of **default** and **duplicate_pool1** look **redundant** and live in `Pool.schedulableQueue`.

duplicate_pools
duplicate_pools2
## How was this patch tested?

Added new Unit Test case.

PR #15237, opened 2016-09-25
[SPARK-17663] [CORE] SchedulableBuilder should handle invalid data access via scheduler.allocation.file

erenavsarogullari:
## What changes were proposed in this pull request?

If `spark.scheduler.allocation.file` has invalid `minShare` or/and `weight` values, these cause :
- `NumberFormatException` due to `toInt` function
- `SparkContext` can not be initialized.
- It does not show meaningful error message to user.

In a nutshell, this functionality can be more robust by selecting one of the following flows :

**1-** Currently, if `schedulingMode` has an invalid value, a warning message is logged and default value is set as `FIFO`. Same pattern can be used for `minShare`(default: 0) and `weight`(default: 1) as well
**2-** Meaningful error message can be shown to the user for all invalid cases.

PR offers :
- `schedulingMode` handles just empty values. It also needs to be supported for **whitespace**, **non-uppercase**(fair, FaIr etc...) or `SchedulingMode.NONE` cases by setting default value(`FIFO`)
- `minShare` and `weight` handle just empty values. They also need to be supported for **non-integer** cases by setting default values.
- Some refactoring of `PoolSuite`.

**Code to Reproduce :**

```
val conf = new SparkConf().setAppName("spark-fairscheduler").setMaster("local")
conf.set("spark.scheduler.mode", "FAIR")
conf.set("spark.scheduler.allocation.file", "src/main/resources/fairscheduler-invalid-data.xml")
val sc = new SparkContext(conf)
```

**fairscheduler-invalid-data.xml :**

```
<allocations>
    <pool name="production">
        <schedulingMode>FIFO</schedulingMode>
        <weight>invalid_weight</weight>
        <minShare>2</minShare>
    </pool>
</allocations>
```

**Stacktrace :**

```
Exception in thread "main" java.lang.NumberFormatException: For input string: "invalid_weight"
    at java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
    at java.lang.Integer.parseInt(Integer.java:580)
    at java.lang.Integer.parseInt(Integer.java:615)
    at scala.collection.immutable.StringLike$class.toInt(StringLike.scala:272)
    at scala.collection.immutable.StringOps.toInt(StringOps.scala:29)
    at org.apache.spark.scheduler.FairSchedulableBuilder$$anonfun$org$apache$spark$scheduler$FairSchedulableBuilder$$buildFairSchedulerPool$1.apply(SchedulableBuilder.scala:127)
[...]

PR #15604, opened 2016-10-23
[SPARK-18066] [CORE] [TESTS] Add Pool usage policies test coverage for FIFO & FAIR Schedulers

erenavsarogullari:

The following FIFO & FAIR Schedulers Pool usage cases need to have unit test coverage :
- FIFO Scheduler just uses **root pool** so even if `spark.scheduler.pool` property is set, related pool is not created and `TaskSetManagers` are added to **root pool**.
- FAIR Scheduler uses `default pool` when `spark.scheduler.pool` property is not set. This can be happened when 
  - `Properties` object is **null**, 
  - `Properties` object is **empty**(`new Properties()`), 
  - **default pool** is set(`spark.scheduler.pool=default`).
- FAIR Scheduler creates a **new pool** with **default values** when `spark.scheduler.pool` property points a **non-existent** pool. This can be happened when **scheduler allocation file** is not set or it does not contain related pool.
## How was this patch tested?

New Unit tests are added.

### DIFF
--- a/core/src/main/scala/org/apache/spark/scheduler/SchedulableBuilder.scala
+++ b/core/src/main/scala/org/apache/spark/scheduler/SchedulableBuilder.scala
@@ -191,11 +191,11 @@ private[spark] class FairSchedulableBuilder(val rootPool: Pool, conf: SparkConf)
         parentPool = new Pool(poolName, DEFAULT_SCHEDULING_MODE,
           DEFAULT_MINIMUM_SHARE, DEFAULT_WEIGHT)
         rootPool.addSchedulable(parentPool)
-        logWarning(("An Unconfigured pool is found. It is built with default configuration. " +
-          "This can happen when the file that pools are read from isn't set, or when that file " +
-          "doesn't contain the pool name specified by spark.scheduler.pool. " +
-          "Created pool: %s, schedulingMode: %s, minShare: %d, weight: %d").format(
-          poolName, DEFAULT_SCHEDULING_MODE, DEFAULT_MINIMUM_SHARE, DEFAULT_WEIGHT))
+        logWarning(s"A job was submitted with scheduler pool $poolName, which has not been " +
+          "configured. This can happen when the file that pools are read from isn't set, or " +
+          s"when that file doesn't contain $poolName. Created $poolName with default " +
+          s"configuration (schedulingMode: $DEFAULT_SCHEDULING_MODE, " +
+          s"minShare: $DEFAULT_MINIMUM_SHARE, weight: $DEFAULT_WEIGHT)")
       }
     }
     parentPool.addSchedulable(manager)
--- a/core/src/main/scala/org/apache/spark/scheduler/SchedulableBuilder.scala
+++ b/core/src/main/scala/org/apache/spark/scheduler/SchedulableBuilder.scala
@@ -191,7 +191,7 @@ private[spark] class FairSchedulableBuilder(val rootPool: Pool, conf: SparkConf)
         parentPool = new Pool(poolName, DEFAULT_SCHEDULING_MODE,
           DEFAULT_MINIMUM_SHARE, DEFAULT_WEIGHT)
         rootPool.addSchedulable(parentPool)
-        logInfo(("An Unconfigured pool is found. It is built with default configuration. " +
+        logWarning(("An Unconfigured pool is found. It is built with default configuration. " +
           "This can happen when the file that pools are read from isn't set, or when that file " +
           "doesn't contain the pool name specified by spark.scheduler.pool. " +
           "Created pool: %s, schedulingMode: %s, minShare: %d, weight: %d").format(
--- a/core/src/main/scala/org/apache/spark/scheduler/SchedulableBuilder.scala
+++ b/core/src/main/scala/org/apache/spark/scheduler/SchedulableBuilder.scala
@@ -191,7 +191,10 @@ private[spark] class FairSchedulableBuilder(val rootPool: Pool, conf: SparkConf)
         parentPool = new Pool(poolName, DEFAULT_SCHEDULING_MODE,
           DEFAULT_MINIMUM_SHARE, DEFAULT_WEIGHT)
         rootPool.addSchedulable(parentPool)
-        logInfo("Created pool: %s, schedulingMode: %s, minShare: %d, weight: %d".format(
+        logInfo(("An Unconfigured pool is found. It is built with default configuration. " +
+          "This can happen when the file that pools are read from isn't set, or when that file " +
+          "doesn't contain the pool name specified by spark.scheduler.pool. " +
+          "Created pool: %s, schedulingMode: %s, minShare: %d, weight: %d").format(
           poolName, DEFAULT_SCHEDULING_MODE, DEFAULT_MINIMUM_SHARE, DEFAULT_WEIGHT))
       }
     }
--- a/core/src/test/scala/org/apache/spark/scheduler/PoolSuite.scala
+++ b/core/src/test/scala/org/apache/spark/scheduler/PoolSuite.scala
@@ -203,9 +203,9 @@ class PoolSuite extends SparkFunSuite with LocalSparkContext {
   }
 
   /**
-    * spark.scheduler.pool property should be ignored for the FIFO scheduler,
-    * because pools are only needed for fair scheduling.
-    */
+   * spark.scheduler.pool property should be ignored for the FIFO scheduler,
+   * because pools are only needed for fair scheduling.
+   */
   test("FIFO scheduler uses root pool and not spark.scheduler.pool property") {
     sc = new SparkContext("local", "PoolSuite")
     val taskScheduler = new TaskSchedulerImpl(sc)
--- a/core/src/test/scala/org/apache/spark/scheduler/PoolSuite.scala
+++ b/core/src/test/scala/org/apache/spark/scheduler/PoolSuite.scala
@@ -31,6 +31,7 @@ class PoolSuite extends SparkFunSuite with LocalSparkContext {
   val LOCAL = "local"
   val APP_NAME = "PoolSuite"
   val SCHEDULER_ALLOCATION_FILE_PROPERTY = "spark.scheduler.allocation.file"
+  val TEST_POOL = "testPool"
 
   def createTaskSetManager(stageId: Int, numTasks: Int, taskScheduler: TaskSchedulerImpl)
     : TaskSetManager = {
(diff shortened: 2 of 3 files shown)


erenavsarogullari:
cc @kayousterhout @markhamstra

erenavsarogullari:
Hi @kayousterhout @markhamstra @squito,
This PR aims to extend Pool usage policies unit test coverage and is also ready for review.
Thanks and all feedbacks are welcome in advance ;)

squito:
Jenkins, ok to test

core/src/test/scala/org/apache/spark/scheduler/PoolSuite.scala
```diff
@@ -178,4 +180,97 @@ class PoolSuite extends SparkFunSuite with LocalSparkContext {
...
+
+    // FAIR Scheduler uses default pool when spark.scheduler.pool property is not set
+    val taskSetManager1 = createTaskSetManager(stageId = 1, numTasks = 1, taskScheduler)
+
+    schedulableBuilder.addTaskSetManager(taskSetManager1, new Properties())
+
+    assert(defaultPool.schedulableQueue.size == 2)
+    assert(defaultPool.getSchedulableByName(taskSetManager1.name) === taskSetManager1)
+
+    // FAIR Scheduler uses default pool when spark.scheduler.pool property is set as default pool
```
squito: I'm all for more tests, but this final case here seems pretty unnecessary

core/src/test/scala/org/apache/spark/scheduler/PoolSuite.scala
```diff
@@ -178,4 +180,97 @@ class PoolSuite extends SparkFunSuite with LocalSparkContext {
...
+      .DEFAULT_POOL_NAME)
+
+    schedulableBuilder.addTaskSetManager(taskSetManager2, properties)
+
+    assert(defaultPool.schedulableQueue.size == 3)
+    assert(defaultPool.getSchedulableByName(taskSetManager2.name) === taskSetManager2)
+  }
+
+  test("SPARK-18066: FAIR Scheduler creates a new pool when spark.scheduler.pool property points " +
+    "non-existent") {
```
squito: minor nit: rename to "FAIR Scheduler creates a new pool when spark.scheduler.pool property points to a non-existent pool"

bigger question:
@markhamstra @kayousterhout Is this really the desired behavior?  Or is there a bug -- should it fail fast?  It looks like [this behavior was intentional](https://github.com/apache/spark/blob/b67b35f76b684c5176dc683e7491fd01b43f4467/core/src/main/scala/org/apache/spark/scheduler/SchedulableBuilder.scala#L143-L145).  Then again, that comment makes it sound like the user is going to directly modify the `weight` and `minShare` of the constructed pool, but though [they are `var`s](https://github.com/apache/spark/blob/b67b35f76b684c5176dc683e7491fd01b43f4467/core/src/main/scala/org/apache/spark/scheduler/Pool.scala#L40-L41), its `private [spark]` and there isn't anything else exposing it to the user.

I prefer fail-fast behavior, but it seems like fair scheduler configuration prefers to assume defaults when there is misconfiguration.  If you think this the right behavior, than there is probably some minor cleanup to do in Pool & SchedulableBuilder to clarify.

core/src/test/scala/org/apache/spark/scheduler/PoolSuite.scala
```diff
@@ -178,4 +180,97 @@ class PoolSuite extends SparkFunSuite with LocalSparkContext {
     scheduleTaskAndVerifyId(2, rootPool, 6)
     scheduleTaskAndVerifyId(3, rootPool, 2)
   }
+
+  test("SPARK-18066: FIFO Scheduler just uses root pool") {
```
squito: super nit: the jira doesn't really say anything other than "add these tests", so not much point in listing it here, just name test "FIFO Scheduler just uses root pool".

squito (commented):
Thanks for this @erenavsarogullari , great to have more tests.  I have super trivial suggested changes, but one bigger question for other committers I'd like to wait a bit to hear back on.

core/src/test/scala/org/apache/spark/scheduler/PoolSuite.scala
kayousterhout: +1. Also how about "FIFO scheduler uses root pool and not spark.scheduler.pool" and then adding a comment at the top of the test that says something like "spark.scheduler.pool should be ignored for the FIFO scheduler, because pools are only needed for fair scheduling"? Took me a minute to figure out what was going on here.

core/src/test/scala/org/apache/spark/scheduler/PoolSuite.scala
```diff
@@ -178,4 +180,97 @@ class PoolSuite extends SparkFunSuite with LocalSparkContext {
...
+    val schedulableBuilder = new FIFOSchedulableBuilder(rootPool)
+
+    val taskSetManager0 = createTaskSetManager(stageId = 0, numTasks = 1, taskScheduler)
+    val taskSetManager1 = createTaskSetManager(stageId = 1, numTasks = 1, taskScheduler)
+
+    val properties = new Properties()
+    properties.setProperty("spark.scheduler.pool", TEST_POOL)
+
+    // FIFOSchedulableBuilder just uses rootPool so even if properties are set, related pool
+    // (testPool) is not created and TaskSetManagers are added to rootPool
```
kayousterhout: Can you re-phrase this as what *should* happen, which makes these comments easier to read? e.g. 

"When task sets are submitted, they should be added to the root pool, and no additional pools should be created (even though there's a configured default pool)."

core/src/test/scala/org/apache/spark/scheduler/PoolSuite.scala
```diff
@@ -178,4 +180,97 @@ class PoolSuite extends SparkFunSuite with LocalSparkContext {
...
+    schedulableBuilder.addTaskSetManager(taskSetManager1, properties)
+
+    assert(rootPool.getSchedulableByName(TEST_POOL) == null)
+    assert(rootPool.schedulableQueue.size == 2)
+    assert(rootPool.getSchedulableByName(taskSetManager0.name) === taskSetManager0)
+    assert(rootPool.getSchedulableByName(taskSetManager1.name) === taskSetManager1)
+  }
+
+  test("SPARK-18066: FAIR Scheduler uses default pool when spark.scheduler.pool property is not " +
+    "set") {
```
kayousterhout: nit: +2 spaces here

squito:
hmm, so I'm thinking about this again, along with your other pr https://github.com/apache/spark/pull/15237

and it got me thinking -- there also aren't any tests for a mix of FIFO and FAIR (at least, I don't see it in "Nested Pool Test" in PoolSuite).  While you are doing this, would you like to add some tests for that as well?  (to be perfectly honest, I need to spend a little bit of time thinking about what that behavior should be anyway, but I will have to do that a little later.)

core/src/test/scala/org/apache/spark/scheduler/PoolSuite.scala
```diff
@@ -178,4 +180,97 @@ class PoolSuite extends SparkFunSuite with LocalSparkContext {
...
+    sc = new SparkContext("local", "PoolSuite")
+    val taskScheduler = new TaskSchedulerImpl(sc)
+
+    val rootPool = new Pool("", SchedulingMode.FAIR, initMinShare = 0, initWeight = 0)
+    val schedulableBuilder = new FairSchedulableBuilder(rootPool, sc.conf)
+    schedulableBuilder.buildPools()
+
+    // FAIR Scheduler uses default pool when pool properties are null
```
kayousterhout: super nit with commenting, but can you eliminate the blank lines on 219 and 221 (which don't help readability) and change this comment to something like

"Submit a new task set manager with pool properties set to null. This should result in the task set manager getting added to the default pool."

(it's helpful in test comments to be clear about what the test is doing itself, versus the expected behavior it's verifying)

core/src/test/scala/org/apache/spark/scheduler/PoolSuite.scala
```diff
@@ -178,4 +180,97 @@ class PoolSuite extends SparkFunSuite with LocalSparkContext {
...
+    val taskSetManager0 = createTaskSetManager(stageId = 0, numTasks = 1, taskScheduler)
+
+    schedulableBuilder.addTaskSetManager(taskSetManager0, null)
+
+    val defaultPool = rootPool.getSchedulableByName(schedulableBuilder.DEFAULT_POOL_NAME)
+    assert(defaultPool != null)
+    assert(defaultPool.schedulableQueue.size == 1)
+    assert(defaultPool.getSchedulableByName(taskSetManager0.name) === taskSetManager0)
+
+    // FAIR Scheduler uses default pool when spark.scheduler.pool property is not set
```
kayousterhout: Similar to the above, can you eliminate the blank lines here, and change this to "When a task set manager is submitted with spark.scheduler.pool unset, it should be added to the default pool (as above)."

core/src/test/scala/org/apache/spark/scheduler/PoolSuite.scala
kayousterhout: +1

core/src/test/scala/org/apache/spark/scheduler/PoolSuite.scala
kayousterhout: I tracked this down and looks like it was changed [back in 2013](https://github.com/apache/spark/commit/5892393140eb024a32585b6d5b51146ddde8f63a#diff-30ccb25ca46ef204d2446b03ae4f1117R103). @xiajunluan do you remember the motivation for this?

I'm torn because in general I prefer failing fast (e.g., this is better in the case where someone has a small typo they might not notice, and is intending to use a configured pool).  On the other hand, I'm pretty sure I've abused this behavior in the past to avoid configuring pools when I want a bunch of equally weighted pools.  If others are doing that, we shouldn't change the behavior (although I think that, in that case, we should add a big warning / future-deprecation message).

core/src/test/scala/org/apache/spark/scheduler/PoolSuite.scala
```diff
@@ -178,4 +180,97 @@ class PoolSuite extends SparkFunSuite with LocalSparkContext {
...
+    assert(rootPool.getSchedulableByName(TEST_POOL) == null)
+
+    val taskSetManager = createTaskSetManager(stageId = 0, numTasks = 1, taskScheduler)
+
+    val properties = new Properties()
+    properties.setProperty(schedulableBuilder.FAIR_SCHEDULER_PROPERTIES, TEST_POOL)
+
+    // FAIR Scheduler creates a new pool with default values when spark.scheduler.pool property
+    // points non-existent pool. This can be happened when scheduler allocation file is not set or
+    // it does not contain related pool
```
kayousterhout: can you clean up this comment a bit too?

"The fair scheduler should create a new pool with default values when spark.scheduler.pool points to a pool that doesn't exist yet (this can happen when the file that pools are read from isn't set, or when that file doesn't contain the pool name specified by spark.scheduler.pool)."

kayousterhout (commented):
Thanks for adding these tests!

core/src/test/scala/org/apache/spark/scheduler/PoolSuite.scala
squito: yeah I prefer failing fast for the same reason.  But I guess we can't actually change this, since there is a valid use case for this behavior.  We don't really know if others are using this, and I don't think we should change it under them in the next release.

In any case, I still think we should also change `Pool.weight` and `Pool.minShare` to `val`s.

kayousterhout:
@erenavsarogullari what's the status with this PR?  I know you have at least one other fair scheduler PR outstanding, and it makes reviews easier to minimize the number of outstanding things (and helps avoid merge conflicts).

erenavsarogullari:
@kayousterhout thanks for query this PR. I will be submitting this and the other one( #15326) once #16813 is merged.

kayousterhout:
Ok thanks for the update!

kayousterhout:
Ok thanks for the update!

erenavsarogullari:
Jenkins, retest this please

core/src/test/scala/org/apache/spark/scheduler/PoolSuite.scala
erenavsarogullari: In this case, at least, i think we can inform the user by emphasizing that the related pool does not exist and it is created with **default settings**. 
Our current logging just points pool creation. WDYT?  

Also Pool `var weight / minShare` have been addressed ;)

erenavsarogullari:
Hi @kayousterhout and @squito,
Firstly, many thanks for the review. All comments are addressed. This is ready to re-review ;)

erenavsarogullari:
build failure looks unrelated (due to timeout at KafkaSourceSuite level)

Jenkins, retest this please

erenavsarogullari:
Jenkins, retest this please

erenavsarogullari:
Hi @kayousterhout and @squito,
This PR is ready to re-review / merge if it is possible.
All feedbacks are welcome and thanks in advance ;)

erenavsarogullari:
Hi @kayousterhout and @squito,
This PR is ready to re-review.
All feedbacks are welcome and thanks in advance ;)

core/src/test/scala/org/apache/spark/scheduler/PoolSuite.scala
```diff
@@ -201,12 +202,96 @@ class PoolSuite extends SparkFunSuite with LocalSparkContext {
...
+
+    assert(rootPool.getSchedulableByName(TEST_POOL) === null)
+
+    val taskSetManager = createTaskSetManager(stageId = 0, numTasks = 1, taskScheduler)
+
+    val properties = new Properties()
+    properties.setProperty(schedulableBuilder.FAIR_SCHEDULER_PROPERTIES, TEST_POOL)
+
+    // The fair scheduler should create a new pool with default values when spark.scheduler.pool
+    // points to a pool that doesn't exist yet (this can happen when the file that pools are read
```
kayousterhout: JOOC does this result in a warning being printed to the user?  (I know you can't test for that -- but it does seem like useful behavior to have)

kayousterhout (changes requested):
One last small commeent

core/src/test/scala/org/apache/spark/scheduler/PoolSuite.scala
erenavsarogullari: Currently, this behavior is **not** highlighted via logs and we just points the pool creation as 
`Created pool: ..., schedulingMode: ..., minShare: ..., weight: ...` so i agree it can be useful for the user and it is addressed with the latest commits.

erenavsarogullari:
Hi @kayousterhout and @squito,
Last comments are addressed so this PR is ready to re-review / merge if it is possible.
Thanks in advance ;)

erenavsarogullari:
The following unit tests look failed due to timeout and _irrelevant_ so triggered again.

**1-** org.apache.spark.storage.BlockManagerProactiveReplicationSuite.proactive block replication - 5 replicas - 4 block manager deletions

**Error Message:** org.scalatest.exceptions.TestFailedDueToTimeoutException: The code passed to eventually never returned normally. Attempted 493 times over 5.007521253999999 seconds. Last failure message: 4 did not equal 5.

**2-** org.apache.spark.deploy.SparkSubmitSuite.includes jars passed in through --packages

**Error Message:** org.scalatest.exceptions.TestFailedDueToTimeoutException: The code passed to failAfter did not complete within 60 seconds.

erenavsarogullari:
Jenkins, retest this please

kayousterhout:
@erenavsarogullari please file a JIRA when you see test failures instead of ignoring them.  I updated https://issues.apache.org/jira/browse/SPARK-19803 for the first failure, but please file a JIRA for the second one.

kayousterhout:
Jenkins retest this please

core/src/main/scala/org/apache/spark/scheduler/SchedulableBuilder.scala
```diff
@@ -191,7 +191,10 @@ private[spark] class FairSchedulableBuilder(val rootPool: Pool, conf: SparkConf)
         parentPool = new Pool(poolName, DEFAULT_SCHEDULING_MODE,
           DEFAULT_MINIMUM_SHARE, DEFAULT_WEIGHT)
         rootPool.addSchedulable(parentPool)
-        logInfo("Created pool: %s, schedulingMode: %s, minShare: %d, weight: %d".format(
+        logWarning(("An Unconfigured pool is found. It is built with default configuration. " +
```
kayousterhout: It's not totally clear what the first part of this error message means.  How about changing the message to:

```
logWarning(s"A job was submitted with scheduler pool $poolName, which has not been configured." +
    "This can happen when the file that pools are read from isn't set, or when that file " +
    s"doesn't contain $poolName.  Created $poolName with default configuration (" +
    s"schedulingMode: $DEFAULT_SCHEDULING_MODE, minShare: $DEFAULT_MINIMUM_SHARE, " +
    s"weight: $DEFAULT_WEIGHT)")
```

core/src/test/scala/org/apache/spark/scheduler/PoolSuite.scala
kayousterhout: Thanks for updating this!

kayousterhout (commented):
This looks great -- one last comment update and then I can merge this!

core/src/main/scala/org/apache/spark/scheduler/SchedulableBuilder.scala
erenavsarogullari: Sure, it is updated ;)

erenavsarogullari:
Many thanks @kayousterhout for reviewing this again. Last comment has also been addressed.

Also the following Jira has been created for previous second UT failure.
https://issues.apache.org/jira/browse/SPARK-19964

kayousterhout:
Thanks for your work on this @erenavsarogullari. I've merged this into master.

PR #15606, opened 2016-10-24
[SPARK-18070][SQL] binary operator should not consider nullability when comparing input types

cloud-fan:

Binary operator requires its inputs to be of same type, but it should not consider nullability, e.g. `EqualTo` should be able to compare an element-nullable array and an element-non-nullable array.
## How was this patch tested?

a regression test in `DataFrameSuite`


cloud-fan:
cc @gatorsmile @yhuai

gatorsmile:
This bug only affects the complex type, (`ArrayType`, `StructType`, and `MapType`)?

cloud-fan:
@gatorsmile yup, the `DataType.sameType` has no override

gatorsmile:
LGTM

gatorsmile:
retest this please

yhuai:
LGTM. Merging to master and branch 2.0.

PR #15608, opened 2016-10-24
[SPARK-17838][SparkR] Check named arguments for options and use formatted R friendly message from JVM exception message

HyukjinKwon:

This PR proposes to
- improve the R-friendly error messages rather than raw JVM exception one.
  
  As `read.json`, `read.text`, `read.orc`, `read.parquet` and `read.jdbc` are executed in the same  path with `read.df`, and `write.json`, `write.text`, `write.orc`, `write.parquet` and `write.jdbc` shares the same path with `write.df`, it seems it is safe to call `handledCallJMethod` to handle 
  JVM messages.
-  prevent `zero-length variable name` and prints the ignored options as an warning message.

**Before**

``` r
> read.json("path", a = 1, 2, 3, "a")
Error in env[[name]] <- value :
  zero-length variable name
```

``` r
> read.json("arbitrary_path")
Error in invokeJava(isStatic = FALSE, objId$id, methodName, ...) :
  org.apache.spark.sql.AnalysisException: Path does not exist: file:/...;
  at org.apache.spark.sql.execution.datasources.DataSource$$anonfun$12.apply(DataSource.scala:398)
  ...

> read.orc("arbitrary_path")
Error in invokeJava(isStatic = FALSE, objId$id, methodName, ...) :
  org.apache.spark.sql.AnalysisException: Path does not exist: file:/...;
  at org.apache.spark.sql.execution.datasources.DataSource$$anonfun$12.apply(DataSource.scala:398)
  ...

> read.text("arbitrary_path")
Error in invokeJava(isStatic = FALSE, objId$id, methodName, ...) :
  org.apache.spark.sql.AnalysisException: Path does not exist: file:/...;
  at org.apache.spark.sql.execution.datasources.DataSource$$anonfun$12.apply(DataSource.scala:398)
  ...

> read.parquet("arbitrary_path")
Error in invokeJava(isStatic = FALSE, objId$id, methodName, ...) :
  org.apache.spark.sql.AnalysisException: Path does not exist: file:/...;
  at org.apache.spark.sql.execution.datasources.DataSource$$anonfun$12.apply(DataSource.scala:398)
  ...
```

``` r
> write.json(df, "existing_path")
Error in invokeJava(isStatic = FALSE, objId$id, methodName, ...) :
  org.apache.spark.sql.AnalysisException: path file:/... already exists.;
  at org.apache.spark.sql.execution.datasources.InsertIntoHadoopFsRelationCommand.run(InsertIntoHadoopFsRelationCommand.scala:68)

> write.orc(df, "existing_path")
Error in invokeJava(isStatic = FALSE, objId$id, methodName, ...) :
  org.apache.spark.sql.AnalysisException: path file:/... already exists.;
  at org.apache.spark.sql.execution.datasources.InsertIntoHadoopFsRelationCommand.run(InsertIntoHadoopFsRelationCommand.scala:68)

> write.text(df, "existing_path")
Error in invokeJava(isStatic = FALSE, objId$id, methodName, ...) :
  org.apache.spark.sql.AnalysisException: path file:/... already exists.;
  at org.apache.spark.sql.execution.datasources.InsertIntoHadoopFsRelationCommand.run(InsertIntoHadoopFsRelationCommand.scala:68)

> write.parquet(df, "existing_path")
Error in invokeJava(isStatic = FALSE, objId$id, methodName, ...) :
  org.apache.spark.sql.AnalysisException: path file:/... already exists.;
  at org.apache.spark.sql.execution.datasources.InsertIntoHadoopFsRelationCommand.run(InsertIntoHadoopFsRelationCommand.scala:68)
```

**After**

``` r
read.json("arbitrary_path", a = 1, 2, 3, "a")
Unnamed arguments ignored: 2, 3, a.
```

``` r
> read.json("arbitrary_path")
Error in json : analysis error - Path does not exist: file:/...

> read.orc("arbitrary_path")
Error in orc : analysis error - Path does not exist: file:/...

> read.text("arbitrary_path")
Error in text : analysis error - Path does not exist: file:/...

> read.parquet("arbitrary_path")
Error in parquet : analysis error - Path does not exist: file:/...
```

``` r
> write.json(df, "existing_path")
Error in json : analysis error - path file:/... already exists.;

> write.orc(df, "existing_path")
Error in orc : analysis error - path file:/... already exists.;

> write.text(df, "existing_path")
Error in text : analysis error - path file:/... already exists.;

> write.parquet(df, "existing_path")
Error in parquet : analysis error - path file:/... already exists.;
```
## How was this patch tested?

Unit tests in `test_utils.R` and `test_sparkSQL.R`.

### DIFF
--- a/R/pkg/R/utils.R
+++ b/R/pkg/R/utils.R
@@ -338,19 +338,20 @@ varargsToEnv <- function(...) {
 # into string.
 varargsToStrEnv <- function(...) {
   pairs <- list(...)
-  env <- new.env()
   nameList <- names(pairs)
+  env <- new.env()
   ignoredNames <- list()
-  i <- 1
 
   if (is.null(nameList)) {
     # When all arguments are not named, names(..) returns NULL.
     ignoredNames <- pairs
   } else {
-    for (name in names(pairs)) {
+    for (i in seq_along(pairs)) {
+      name <- nameList[i]
+      value <- pairs[i]
       if (identical(name, "")) {
         # When some of arguments are not named, name is "".
-        ignoredNames <- append(ignoredNames, pairs[i])
+        ignoredNames <- append(ignoredNames, value)
       } else {
         value <- pairs[[name]]
         if (!(is.logical(value) || is.numeric(value) || is.character(value) || is.null(value))) {
@@ -365,12 +366,11 @@ varargsToStrEnv <- function(...) {
           env[[name]] <- as.character(value)
         }
       }
-      i <- i + 1
     }
   }
 
   if (length(ignoredNames) != 0) {
-    warning(paste0("Non-named arguments ignored: ", paste(ignoredNames, collapse = ", "), "."),
+    warning(paste0("Unnamed arguments ignored: ", paste(ignoredNames, collapse = ", "), "."),
             call. = FALSE)
   }
   env
--- a/R/pkg/R/utils.R
+++ b/R/pkg/R/utils.R
@@ -339,19 +339,39 @@ varargsToEnv <- function(...) {
 varargsToStrEnv <- function(...) {
   pairs <- list(...)
   env <- new.env()
-  for (name in names(pairs)) {
-    value <- pairs[[name]]
-    if (!(is.logical(value) || is.numeric(value) || is.character(value) || is.null(value))) {
-      stop(paste0("Unsupported type for ", name, " : ", class(value),
-           ". Supported types are logical, numeric, character and NULL."))
+  nameList <- names(pairs)
+  ignoredNames <- list()
+  i <- 1
+
+  if (is.null(nameList)) {
+    # When all arguments are not named, names(..) returns NULL.
+    ignoredNames <- pairs
+  } else {
+    for (name in names(pairs)) {
+      if (identical(name, "")) {
+        # When some of arguments are not named, name is "".
+        ignoredNames <- append(ignoredNames, pairs[i])
+      } else {
+        value <- pairs[[name]]
+        if (!(is.logical(value) || is.numeric(value) || is.character(value) || is.null(value))) {
+          stop(paste0("Unsupported type for ", name, " : ", class(value),
+               ". Supported types are logical, numeric, character and NULL."), call. = FALSE)
+        }
+        if (is.logical(value)) {
+          env[[name]] <- tolower(as.character(value))
+        } else if (is.null(value)) {
+          env[[name]] <- value
+        } else {
+          env[[name]] <- as.character(value)
+        }
+      }
+      i <- i + 1
     }
-    if (is.logical(value)) {
-      env[[name]] <- tolower(as.character(value))
-    } else if (is.null(value)) {
-      env[[name]] <- value
-    } else {
-      env[[name]] <- as.character(value)
-    }
+  }
+
+  if (length(ignoredNames) != 0) {
+    warning(paste0("Non-named arguments ignored: ", paste(ignoredNames, collapse = ", "), "."),
+            call. = FALSE)
   }
   env
 }
--- a/R/pkg/R/DataFrame.R
+++ b/R/pkg/R/DataFrame.R
@@ -761,7 +761,7 @@ setMethod("write.json",
           function(x, path, mode = "error", ...) {
             write <- callJMethod(x@sdf, "write")
             write <- setWriteOptions(write, mode = mode, ...)
-            invisible(callJMethod(write, "json", path))
+            invisible(handledCallJMethod(write, "json", path))
           })
 
 #' Save the contents of SparkDataFrame as an ORC file, preserving the schema.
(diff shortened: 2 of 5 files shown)


HyukjinKwon:
cc @felixcheung I recall we talked about this before. I first wanted to handle all the argument type checking but I just decided to do what I am pretty sure of (I remember we were concern of sweeping). Could you please take a look?

R/pkg/R/utils.R
```diff
@@ -339,20 +339,40 @@ varargsToEnv <- function(...) {
...
-    }
-    if (is.logical(value)) {
-      env[[name]] <- tolower(as.character(value))
-    } else if (is.null(value)) {
-      env[[name]] <- value
-    } else {
-      env[[name]] <- as.character(value)
+  nameList <- names(pairs)
+  ignoredNames <- list()
+  i <- 1
```
felixcheung: why is `i` needed here?
could you elaborate on changes in this file? why do we need them?

felixcheung:
Thanks @HyukjinKwon This is definitely good checks to have. Calls to read.* and write.* are not easily checked for parameters (file paths are hard) so to me it is better to leave the checking to JVM and just handle the exception better.

I wouldn't close this JIRA though but it is covering a broader case - those we could check parameters before passing to JVM? Would you like to keep this JIRA open after this PR, or open a separate JIRA for this specific class of handling?

One case I'm not sure about, is what Reader/Writer does with unnamed parameter - should we optionally fail here on the R side, instead of just ignoring them (in R or JVM)?

R/pkg/R/utils.R
HyukjinKwon: Oh, sure. Actually, it took me a while to find out a clean way but it ended up with introducing another variable here...

```r
value <- pairs[[name]]
```

is failed when `name` is empty string (when they are non-named arguments), producing the error as below:

```r
Error in env[[name]] <- value :
  zero-length variable name
```

I wanted to access to that problematic values to print in 

```r
ignoredNames <- append(ignoredNames, pairs[i])
...
warning(paste0("Non-named arguments ignored: ", paste(ignoredNames, collapse = ", "), "."),
         call. = FALSE)
```

To cut it short, I introduced that variable to access to the value when `name` is empty. Maybe there should be another cleaver way but I couldn't come up with it.

HyukjinKwon:
I am fine with leaving the JIRA open. I can definitely try to open followups. Otherwise, I also can convert the JIRA as a sub-task after introducing a parent JIRA.

I will follow your lead. (If you close the JIRA then I will open another to make this as a sub-task. If you don't I will just try to make a followup in the future).

> One case I'm not sure about, is what Reader/Writer does with unnamed parameter - should we optionally fail here on the R side, instead of just ignoring them (in R or JVM)?

I was worried of this part too and It took me a while to build up my argument for this... My argument to convince myself was that currently we don't make this failed when unused arbitrary options are given (e.g. `option(`abc`,  1`)), so, it'd be okay to not make this failed. but of course this is not a strong opinion.

felixcheung:
re: JIRA - I don't mind one way or the other - both proposals sound good to me.

> currently we don't make this failed when unused arbitrary options are given (e.g. option("abc", "1")), so, it'd be okay to not make this failed. but of course this is not a strong opinion.

a valid point. does it work properly when the Spark properties are set into a Java Properties class?
https://docs.oracle.com/javase/7/docs/api/java/util/Properties.html#setProperty(java.lang.String,%20java.lang.String)
It seems if the key is `""` or `null` then it would overwrite?

R/pkg/R/utils.R
HyukjinKwon: That looks nicer. Let me try that!

HyukjinKwon:
I guess you meant when the options, for example, `DataFrameReader.extraOptions` turns into `Properties` when we call JDBC APIs. In this case, `Properties` always has a higher precedence and will overwrite `DataFrameReader.extraOptions`[1] and exclude Spark internal options[2].

[1]https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/DataFrameReader.scala#L235
[2]https://github.com/apache/spark/blob/0c0ad436ad909364915b910867d08262c62bc95d/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JDBCOptions.scala#L43

(BTW, it seems throwing `NullPointException` when we call `setProperty` and the key is `null`)

HyukjinKwon:
@felixcheung I just made it as a single for loop (slightly different with the suggested one though..). Could you please check it again?

felixcheung:
LGTM. I ran through a few other cases and I think the omitted names are handled properly with this.
This should go to master then (handledCallJMethod is not in branch-2.0)

HyukjinKwon:
Yup, it seems not even `varargsToStrEnv`.

felixcheung:
merged to master.

PR #15609, opened 2016-10-24
[SPARK-18048][SQL] To make behaviour of If consistent, in case of true expression and false expression are of compatible data types.

priyankagargnitk:

This change adds a type conversion from false value datatype to true value’s datatype.

Earlier, the expression If(Literal.create(true, BooleanType), Literal.create(identity(1), DateType), Literal.create(identity(2L), TimestampType))  was throwing Exception while the expression If(Literal.create(true, BooleanType), Literal.create(identity(1L), TimestampType), Literal.create(identity(2), DateType)) was working fine. So , if we interchange the true and false expressions, behaviour of the IF expression changes.
## How was this patch tested?

Added test case in ConditionalExpressionSuite.scala

jira entry for detail: https://issues.apache.org/jira/browse/SPARK-18048

### DIFF
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/conditionalExpressions.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/conditionalExpressions.scala
@@ -70,7 +70,7 @@ case class If(predicate: Expression, trueValue: Expression, falseValue: Expressi
       } else {
         ${falseEval.code}
         ${ev.isNull} = ${falseEval.isNull};
-        ${ev.value} = ${falseEval.value};
+        ${ev.value} = (${ctx.javaType(dataType)}) ${falseEval.value};
       }""")
   }
 
--- a/sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/ConditionalExpressionSuite.scala
+++ b/sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/ConditionalExpressionSuite.scala
@@ -66,6 +66,19 @@ class ConditionalExpressionSuite extends SparkFunSuite with ExpressionEvalHelper
     DataTypeTestUtils.propertyCheckSupported.foreach { dt =>
       checkConsistencyBetweenInterpretedAndCodegen(If, BooleanType, dt, dt)
     }
+
+    checkEvaluation(
+      If(Literal.create(true, BooleanType),
+        Literal.create(identity(1L), TimestampType),
+        Literal.create(identity(2), DateType)),
+      identity(1L))
+
+    checkEvaluation(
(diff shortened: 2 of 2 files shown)


HyukjinKwon:
Hm, does this cause a actual problem? I thought this is covered in type-widening.

priyankagargnitk:
actually, its not covered in type widening... So, if any of my expression is calling if expression ( in nested) .. its failing because of this issue.

HyukjinKwon:
@priyankagargnitk Could I please ask some codes which I can reproduce this issue? It seems fine the sql below:

```sql
spark-sql> SELECT if(true, cast('2014-01-01' as date), cast('2014-12-12' as timestamp));
2014-01-01 00:00:00
```

I just checked and it seems this is covered in [TypeCoercion.scala#L596-L601](https://github.com/apache/spark/blob/39e2bad6a866d27c3ca594d15e574a1da3ee84cc/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/TypeCoercion.scala#L596-L601). It seems the cases here are unreachable to end users. 

Also, it seems it can't just cast it. For example, if we add a test with `IntegerType` and `StringType` here, i guess it'd be failed due to the cast.

priyankagargnitk:
Actually our use case is little different, we are not invoking it with the Select queries.. We are using JS to let user type expressions and then we have created out own layer on top of spark that converts this JS to Spark expressions.. So if user writes an expression that returns date in true branch and timestamp in false branch.. Then the IF expression fails... because it internally getting mapped to If(Literal.create(true, BooleanType), Literal.create(identity(1), DateType), Literal.create(identity(2L), TimestampType)) , Which fails.

HyukjinKwon:
Then, wouldn't the responsibility of type coercion be your own layer? I guess this PR virtually mean every expression handled in `TypeCoercion` should be fixed too. For example, I guess 

```scala
Greatest(Seq(Literal(10, IntegerType), Literal("a", StringType))).eval()
```

wouldn't work too.

hvanhovell:
@priyankagargnitk this should be covered in type widening. The `trueValue` and `falseValue` expressions should have the same type by the time we reach the execution phase. The `IfCoercion` rule should have injected casts if the types were not compatible. Could you please take a look there?

hvanhovell:
@priyankagargnitk the use case you are describing is not supported, and I don't think we should merge this. You can just call type coercion rules on your expression tree, to get the properly types tree.

priyankagargnitk:
Hmm... I think that makes sense.. i am trying a couple of more things... Thanks

HyukjinKwon:
@priyankagargnitk Could maybe we then close this?

priyankagargnitk:
Sure.. I am closing it... If still face any issues... will raise another.. Thanks for the help.

PR #15612, opened 2016-10-24
[SPARK-18078] Add zipPartitionsWithPreferredLocation function for better data-locality zipPartition operation

WeichenXu123:

add `ZipPartitionsRDDUtils.zipPartitionsWithPreferredLocation` helper function.
The zipPartitionsWithPreferredLocation helper function is similar to RDD.zipPartitions, but
it can be used to specify the zipPartitions task preferred locations to be consistent with
the fisrt zipped RDD. If the fisrt zipped RDD do not have preferred location, it will fallback to the default `zipPartition` preferred location strategy.

This option can be used in such scenario:
A large RDD zipped with other much smaller RDDs,  we can set the first zipped RDD to be the large RDD to improve data locality.
### Performance test

I test this improvement on **spark-tfocus** DMatrix multiplying DVector operator **LinopMatrixAdjoint**
https://github.com/databricks/spark-tfocs/blob/master/src/main/scala/org/apache/spark/mllib/optimization/tfocs/fs/dvector/vector/LinopMatrixAdjoint.scala

I generate the `DMatrix` RDD, using 100 partitions and each partition contains about 100MB data. and the `DVector` also 100 partitions, each partitions about 1MB data, and I start the job using 20 executors on 20 nodes, each executor contains 4 cores. and I set the `spark.locality.wait` to be 3000ms, the result show that using `ZipPartitionsRDDUtils.zipPartitionsWithPreferredLocation`, it brings about 20% ~ 30% performance improvement.
## How was this patch tested?

`ZipPartitionsRDDUtilsSuite` added.

### DIFF
--- a/core/src/main/scala/org/apache/spark/rdd/ZipPartitionsRDDUtils.scala
+++ b/core/src/main/scala/org/apache/spark/rdd/ZipPartitionsRDDUtils.scala
@@ -19,69 +19,105 @@ package org.apache.spark.rdd
 
 import scala.reflect.ClassTag
 
+/**
+ * The `ZipPartitionsRDDUtils` contains a series of `zipPartitionsWithPreferredLocation` helper
+ * functions which are similar to `RDD.zipPartitions`, but it can be used to specify the
+ * zipPartitions task preferred locations to be consistent with the fisrt zipped RDD.
+ *
+ * If the fisrt zipped RDD do not have preferred location, it will fallback to the default
+ * `zipPartition` preferred location strategy.
+ */
 object ZipPartitionsRDDUtils {
 
   /**
-   * The zipPartitionsWithPreferredLocation helper function is similar to RDD.zipPartitions, but
-   * it can be used to specify the zipPartitions task preferred locations to be consistent with
-   * the fisrt zipped RDD. If the fisrt zipped RDD do not have preferred location,
-   * it will fallback to the default `zipPartition` preferred location strategy.
-   * This helper function can be used when one large RDD zipped with other small RDDs, we can set
-   * the first zipped RDD (the `majorRdd` parameter) to be the large RDD to improve data locality.
+   * This helper function can be used when one large RDD zipped with another one small RDD, we
+   * can set first zipped RDD (the `majorRdd` parameter) to be the large RDD to improve data
+   * locality.
    *
-   * `preservesPartitioner` indicates whether the input function preserves the partitioner, which
+   * `preservesPartitioning` indicates whether the input function preserves the partitioner, which
    * should be `false` unless this is a pair RDD and the input function doesn't modify the keys.
    */
   def zipPartitionsWithPreferredLocation[A: ClassTag, B: ClassTag, V: ClassTag]
-      (majorRdd: RDD[A], rdd2: RDD[B], preservesPartitioner: Boolean)
+      (majorRdd: RDD[A], rdd2: RDD[B], preservesPartitioning: Boolean)
       (f: (Iterator[A], Iterator[B]) => Iterator[V]): RDD[V] = {
     val sc = majorRdd.sparkContext
     majorRdd.withScope {
-      new ZippedPartitionsRDD2(sc, sc.clean(f), majorRdd, rdd2, preservesPartitioner,
+      new ZippedPartitionsRDD2(sc, sc.clean(f), majorRdd, rdd2, preservesPartitioning,
         useFirstParentPreferredLocations = true)
     }
   }
 
+  /**
+   * This helper function can be used when one large RDD zipped with another one small RDD, we
+   * can set first zipped RDD (the `majorRdd` parameter) to be the large RDD to improve data
+   * locality.
+   */
   def zipPartitionsWithPreferredLocation[A: ClassTag, B: ClassTag, V: ClassTag]
       (majorRdd: RDD[A], rdd2: RDD[B])
       (f: (Iterator[A], Iterator[B]) => Iterator[V]): RDD[V] = {
-    zipPartitionsWithPreferredLocation(majorRdd, rdd2, preservesPartitioner = false)(f)
+    zipPartitionsWithPreferredLocation(majorRdd, rdd2, preservesPartitioning = false)(f)
   }
 
+  /**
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/apache/spark/rdd/RDD.scala
```diff
@@ -848,39 +849,44 @@ abstract class RDD[T: ClassTag](
    * of elements in each partition.
    */
   def zipPartitions[B: ClassTag, V: ClassTag]
-      (rdd2: RDD[B], preservesPartitioning: Boolean)
+      (rdd2: RDD[B], preservesPartitioning: Boolean, preferredLocationRDD: RDD[_])
```
mridulm: Make preferredLocationRDD and Option.
Also can you please add some documentation about it ?
Like what happens when it is specified, what happens when it is not ?

core/src/main/scala/org/apache/spark/rdd/ZippedPartitionsRDD.scala
```diff
@@ -58,10 +63,14 @@ private[spark] abstract class ZippedPartitionsBaseRDD[V: ClassTag](
         s"Can't zip RDDs with unequal numbers of partitions: ${rdds.map(_.partitions.length)}")
     }
     Array.tabulate[Partition](numParts) { i =>
-      val prefs = rdds.map(rdd => rdd.preferredLocations(rdd.partitions(i)))
-      // Check whether there are any hosts that match all RDDs; otherwise return the union
-      val exactMatchLocations = prefs.reduce((x, y) => x.intersect(y))
-      val locs = if (!exactMatchLocations.isEmpty) exactMatchLocations else prefs.flatten.distinct
+      val locs = if (preferredLocationRDD != null) {
+        preferredLocationRDD.preferredLocations(preferredLocationRDD.partitions(i))
```
mridulm: What happens when there are no preferred location for a partition in preferredLocationRDD ?
Do we want to fallback to earlier computation ? (which becomes a distinct union I guess)

WeichenXu123: Good suggestion, it need a fallback handling, thanks!

WeichenXu123:
@rxin Hope for your advice, thanks!

rxin:
This pollutes the RDD API too much. I'd do the low level changes (ZipPartitionsRDD) and perhaps create some utility (e.g. ZipPartitionsRDDUtil) and just have the methods there.

WeichenXu123:
cc @rxin @mridulm Several updates, thanks!

WeichenXu123:
Jenkins, test this please.

core/src/main/scala/org/apache/spark/util/ZipPartitionsRDDUtils.scala
```diff
@@ -0,0 +1,87 @@
...
+ */
+
+package org.apache.spark.util
+
+import scala.reflect.ClassTag
+
+import org.apache.spark.rdd.{RDD, ZippedPartitionsRDD2, ZippedPartitionsRDD3,
+  ZippedPartitionsRDD4}
+
+object ZipPartitionsRDDUtils {
```
rxin: put this in rdd package rather than util?

core/src/main/scala/org/apache/spark/rdd/ZippedPartitionsRDD.scala
```diff
@@ -45,7 +45,8 @@ private[spark] class ZippedPartitionsPartition(
 private[spark] abstract class ZippedPartitionsBaseRDD[V: ClassTag](
     sc: SparkContext,
     var rdds: Seq[RDD[_]],
-    preservesPartitioning: Boolean = false)
+    preservesPartitioning: Boolean = false,
```
rxin: can you rename this preservesPartitioner, and add some classdoc to explain the parameters?

mridulm:
@rxin I am not sure I like this better than previous version.
We have traded an RDD reference for a boolean, and enforced constrain that the 'major' RDD must be the first to get zipped against (as opposed to any of the rdd's).

WeichenXu123:
@mridulm yeah, I'am thinking about this, A util class is to avoid polluting the RDD class as rxin like, and the reason I change parameter `preferredLocationsRDD` into `useFirstParentPreferredLocations` is because,  the way `preferredLocationsRDD` we need to dispose the reference to the `preferredLocationsRDD` when the zipPartitionRDD is cleaned, it makes code fussy. Would restricting the majorRdd to be the first RDD bring some side-effect ? I think it is equivalent to the previous version only current version may need user to swap the parameter orders.

WeichenXu123:
@rxin Move into `rdd` package and add doc for `preservesPartitioner`. The code base have many other places using `preservesPartitioning` I don't update them for now, we can let other person to complete this var renaming work.
@mridulm pleasure to hear you for finding more possible improvements.

WeichenXu123:
Jenkins, test this please.

core/src/main/scala/org/apache/spark/rdd/ZipPartitionsRDDUtils.scala
```diff
@@ -0,0 +1,87 @@
...
+      (majorRdd: RDD[A], rdd2: RDD[B], preservesPartitioner: Boolean)
+      (f: (Iterator[A], Iterator[B]) => Iterator[V]): RDD[V] = {
+    val sc = majorRdd.sparkContext
+    majorRdd.withScope {
+      new ZippedPartitionsRDD2(sc, sc.clean(f), majorRdd, rdd2, preservesPartitioner,
+        useFirstParentPreferredLocations = true)
+    }
+  }
+
+  def zipPartitionsWithPreferredLocation[A: ClassTag, B: ClassTag, V: ClassTag]
```
rxin: we need documentation for all the methods

WeichenXu123:
@rxin 
Documents updated for every methods.
and still keep var name `preservesPartitioning` to keep consistent with other places in code base.
thanks!

WeichenXu123:
Jenkins, test this please.

jiangxb1987:
Do we still want this? @WeichenXu123

PR #15614, opened 2016-10-24
[SPARK-18079] [SQL] CollectLimitExec.executeToIterator should perform per-partition limits

pwoody:

This change adds a partition local limit to the executeToIterator method.
## How was this patch tested?

Added a test to SQLQuerySuite to ensure that only the limited amount is read from the partition.


pwoody:
@JoshRosen since you had made the analogous change to executeCollect.

I wonder if it also makes sense to simply push the LocalLimitExec into the planning within SparkStrategies so that we don't need to deal with this individually in CollectLimitExec. Happy to do either approach!

JoshRosen:
Jenkins, this is okay to test.

This looks fine to me. There's a larger ongoing discussion in #15596 which relates to the planning of these limit operations; let's see if further optimizations are subsumed by that change.

pwoody:
Yep, in that PR the LocalLimitExec is added into planning straight up. That would reduce this to simply adding the child.executeToIterator.take(limit) as an override.

pwoody:
@JoshRosen unfortunately the other PR got closed. thoughts on this independently?

pwoody:
@JoshRosen holler if you want me to make any other changes here.

PR #15266, opened 2016-09-27
[SPARK-17693] [SQL] Fixed Insert Failure To Data Source Tables when the Schema has the Comment Field

gatorsmile:
### What changes were proposed in this pull request?

``` SQL
CREATE TABLE tab1(col1 int COMMENT 'a', col2 int) USING parquet
INSERT INTO TABLE tab1 SELECT 1, 2
```

The insert attempt will fail if the target table has a column with comments. The error is strange to the external users:

```
assertion failed: No plan for InsertIntoTable Relation[col1#15,col2#16] parquet, false, false
+- Project [1 AS col1#19, 2 AS col2#20]
   +- OneRowRelation$
```

This PR is to fix the above bug by ignoring the metadata when comparing the schema between the table and the query.
### How was this patch tested?

Added a test case

PR #15615, opened 2016-10-24
[SPARK-17693] [SQL] Fixed Insert Failure To Data Source Tables when the Schema has the Comment Field

gatorsmile:

``` SQL
CREATE TABLE tab1(col1 int COMMENT 'a', col2 int) USING parquet
INSERT INTO TABLE tab1 SELECT 1, 2
```

The insert attempt will fail if the target table has a column with comments. The error is strange to the external users:

```
assertion failed: No plan for InsertIntoTable Relation[col1#15,col2#16] parquet, false, false
+- Project [1 AS col1#19, 2 AS col2#20]
   +- OneRowRelation$
```

This PR is to fix the above bug by checking the metadata when comparing the schema between the table and the query. If not matched, we also copy the metadata. This is an alternative to https://github.com/apache/spark/pull/15266

Added a test case


cloud-fan:
LGTM

gatorsmile:
Will merge it when the test can pass. Thanks!

gatorsmile:
Merging to master!

Will backport it to 2.0 later. Thanks!

PR #15347, opened 2016-10-04
[SPARK-16827] Stop reporting spill metrics as shuffle metrics

bchocho:
## What changes were proposed in this pull request?

Fix a bug where spill metrics were being reported as shuffle metrics. Eventually these spill metrics should be reported (SPARK-3577), but separate from shuffle metrics. The fix itself basically reverts the line to what it was in 1.6.
## How was this patch tested?

Tested on a job that was reporting shuffle writes even for the final stage, when no shuffle writes should take place. After the change the job no longer shows these writes. 

Before:
screen shot 2016-10-03 at 6 39 59 pm

After:
screen shot 2016-10-03 at 11 44 44 pm

PR #15616, opened 2016-10-24
[SPARK-16827][Shuffle] add disk spill bytes to UnsafeExternalSorter

dreamworks007:

disk spill bytes metrics was disabled due to apache/spark#15347. This PR adds that metrics.
## How was this patch tested?

Integration test with the following query:
val sqlString  = """
SELECT  a.target_id AS userid
         FROM  table_a a
           JOIN table_b b
         ON    a.ds = '2016-07-15'
          AND  b.ds = '2016-07-15'
          AND  a.source_id = b.id"""
hc.sql(sqlString).count()

in 1.6, in stage 2 , it does not have spill bytes metrics reporting, and in 2.0 (with my change) , in stage 2, it has the spill bytes metrics reported
Spark 1.6 - 
spark_1 6

Spark 2.0 - 
spark_2 0


rxin:
cc @joshrosen

dreamworks007:
ping

dreamworks007:
@joshrosen - could you take a look ?

dreamworks007:
ping ... @JoshRosen

dreamworks007:
ping ..

vanzin:
@rxin you closed the related bug, does that mean this patch is not needed or does the bug need to be reopened?

JoshRosen:
Is there any way to add a test for this?

PR #15320, opened 2016-10-01
[SPARK-14914][CORE][SQL] Skip/fix some test cases on Windows due to limitation of Windows

HyukjinKwon:
## What changes were proposed in this pull request?

This PR proposes to fix/skip some tests failed on Windows. This PR takes over https://github.com/apache/spark/pull/12696.

**Before**
- **SparkSubmitSuite**
  
  ```
  [info] - launch simple application with spark-submit *** FAILED *** (202 milliseconds)
  [info]   java.io.IOException: Cannot run program "./bin/spark-submit" (in directory "C:\projects\spark"): CreateProcess error=2, The system cannot find the file specifie
  
  [info] - includes jars passed in through --jars *** FAILED *** (1 second, 625 milliseconds)
  [info]   java.io.IOException: Cannot run program "./bin/spark-submit" (in directory "C:\projects\spark"): CreateProcess error=2, The system cannot find the file specified
  ```
- **DiskStoreSuite**
  
  ```
  [info] - reads of memory-mapped and non memory-mapped files are equivalent *** FAILED *** (1 second, 78 milliseconds)
  [info]   diskStoreMapped.remove(blockId) was false (DiskStoreSuite.scala:41)
  ```

**After**
- **SparkSubmitSuite**
  
  ```
  [info] - launch simple application with spark-submit (578 milliseconds)
  [info] - includes jars passed in through --jars (1 second, 875 milliseconds)
  ```
- **DiskStoreSuite**
  
  ```
  [info] DiskStoreSuite:
  [info] - reads of memory-mapped and non memory-mapped files are equivalent !!! CANCELED !!! (766 milliseconds
  ```

For `CreateTableAsSelectSuite` and `FsHistoryProviderSuite`, I could not reproduce as the Java version seems higher than the one that has the bugs about `setReadable(..)` and `setWritable(...)` but as they are bugs reported clearly, it'd be sensible to skip those. We should revert the changes for both back as soon as we drop the support of Java 7.
## How was this patch tested?

Manually tested via AppVeyor.

Closes #12696

PR #15618, opened 2016-10-25
[SPARK-14914][CORE] Fix Resource not closed after using, mostly for unit tests

HyukjinKwon:

Close `FileStreams`, `ZipFiles` etc to release the resources after using. Not closing the resources will cause IO Exception to be raised while deleting temp files.
## How was this patch tested?

Existing tests


srowen (approved):

HyukjinKwon:
retest this please

HyukjinKwon:
retest this please

HyukjinKwon:
Let me take a look into this deeper of some tests constantly fail.

mridulm:
There are a bunch of methods in Utils which actually nicely apply to this PR.
* Utils.tryWithSafeFinally
* Utils.tryWithResource, etc

They also handle exception suppression, etc.

core/src/main/scala/org/apache/spark/rdd/ReliableCheckpointRDD.scala
```diff
@@ -239,7 +239,14 @@ private[spark] object ReliableCheckpointRDD extends Logging {
       val fs = partitionerFilePath.getFileSystem(sc.hadoopConfiguration)
       val fileInputStream = fs.open(partitionerFilePath, bufferSize)
       val serializer = SparkEnv.get.serializer.newInstance()
-      val deserializeStream = serializer.deserializeStream(fileInputStream)
+      val deserializeStream = try {
+        serializer.deserializeStream(fileInputStream)
+      } catch {
+        case e : Throwable =>
+          fileInputStream.close()
```
felixcheung: shouldn't this be in a try {} finally {} instead?

HyukjinKwon:
Thanks @mridulm and @felixcheung. Let me try to address your comments.

HyukjinKwon:
cc @fuzhouch too (we talk about this via emails).

core/src/main/scala/org/apache/spark/rdd/ReliableCheckpointRDD.scala
srowen: Not here because the stream needs to be open afterwards. I had a similar discussion on the original pr.

HyukjinKwon:
I just took a look for `Utils.tryWithSafeFinally` and `Utils.tryWithResource`. It seems both `Utils.tryWithSafeFinally` and `Utils.tryWithResource closes it in `finally` if I understand this correctly. Could I just leave as it is?

mridulm:
@HyukjinKwon So the idea is that you acquire resources required and dont need to track it by wrapping them in Utils.tryWithResource (similar to memory management in jvm).

As an example: 
main/scala/org/apache/spark/rdd/ReliableCheckpointRDD.scala change will simply acquire the fileInputStream in the try and release it in the finally automatically - without needing to manage it via catch/rethrow, etc (ex: what if close() throws exception ?).

Even core/src/test/scala/org/apache/spark/FileSuite.scala, core/src/test/scala/org/apache/spark/deploy/history/FsHistoryProviderSuite.scala, etc change can be modelled the same way.
You get the idea :-)

This is essentially analogous to try-with-resources in java.
Which is not to say it applies every where ofcourse : drawback is that unlike in java, you need to explicitly specify the finally action, which can be pita imo compared to java's idiom.

Since you are anyway going through the pain of making all these changes to fix up code, might be a good idea to change it such that future tests will follow the same pattern.
Thoughts ?

srowen:
It makes sense to use the Utils methods where possible, sure.

HyukjinKwon:
Thank you for a good reference and explanation @mridulm . I will try to handle it. (It seems I misunderstood the comments.)

core/src/main/scala/org/apache/spark/rdd/ReliableCheckpointRDD.scala
HyukjinKwon: Ah, did you mean something like this?

```scala
val partitioner = Utils.tryWithSafeFinally[Partitioner] {
  val deserializeStream = serializer.deserializeStream(fileInputStream)
  Utils.tryWithSafeFinally[Partitioner] {
    deserializeStream.readObject[Partitioner]
  } {
    deserializeStream.close()
  }
} {
  fileInputStream.close()
}
```

srowen: Yes I think that's a good idea. Do you need the generic types on the "try" methods, even?

HyukjinKwon: Ah-ha, sure. Let me then try. Thanks for bearing with me.

felixcheung:
I don't mean putting the finally here on this line since we do use it in the success case.

But try {} catch e { handleStuff(); throw e; }
is just try {} finally {}. I think there are several ways we could structure it that could work.

But I agree perhaps the difference is not significant at this point.

HyukjinKwon:
Could I please ask to take a look if the additional changes make sense?

srowen (approved):

core/src/test/scala/org/apache/spark/scheduler/TaskResultGetterSuite.scala
```diff
@@ -209,7 +209,8 @@ class TaskResultGetterSuite extends SparkFunSuite with BeforeAndAfter with Local
 
       assert(expectedFailure.findFirstMatchIn(exceptionMessage).isDefined)
       assert(unknownFailure.findFirstMatchIn(exceptionMessage).isEmpty)
-    } finally {
+    } {
+      loader.close()
       Thread.currentThread.setContextClassLoader(originalClassLoader)
```
HyukjinKwon: Sure, makes sense.

streaming/src/main/scala/org/apache/spark/streaming/scheduler/ReceiverTracker.scala
```diff
@@ -194,10 +194,13 @@ class ReceiverTracker(ssc: StreamingContext, skipReceiverLaunch: Boolean = false
...
       ssc.env.rpcEnv.stop(endpoint)
       endpoint = null
-      receivedBlockTracker.stop()
       logInfo("ReceiverTracker stopped")
       trackerState = Stopped
     }
+
+    // note that the output writer is created at construction time, we have to close
+    // them even if it hasn't been started.
+    receivedBlockTracker.stop()
```
HyukjinKwon: Thanks for pointing this out. I just manually tested them by modifying the existing tests in `ReceiverTrackerSuite` and `ReceivedBlockTrackerSuite` and it seems fine with multiple invocations. Should we need a test maybe?

HyukjinKwon:
retest this please

HyukjinKwon:
retest this please

HyukjinKwon:
retest this please

streaming/src/main/scala/org/apache/spark/streaming/scheduler/ReceiverTracker.scala
mridulm: I am actually not sure, since I am not very familiar with streaming : wanted to get it clarified.
@tdas, @zsxwing any comments ?

srowen: That's a good point. Another way to handle this is to move the initialization of `receivedBlockTracker` into `start`, which makes more sense actually. If that's getting risk for this PR, I'd just punt on this change and not move `stop()`

HyukjinKwon:
I tried to initialize it in `start()`. Could you take a look  at https://github.com/apache/spark/pull/15618/commits/49cb4e7f259ba0a236b0a977e69719cfc165c265 ?
(If it does not looks nice, I will definitely revert the change back)

streaming/src/main/scala/org/apache/spark/streaming/scheduler/ReceiverTracker.scala
```diff
@@ -202,18 +206,21 @@ class ReceiverTracker(ssc: StreamingContext, skipReceiverLaunch: Boolean = false
 
   /** Allocate all unallocated blocks to the given batch. */
   def allocateBlocksToBatch(batchTime: Time): Unit = {
+    require(receivedBlockTracker != null, "ReceiverTracker should be started first.")
```
srowen: Per comments above, I don't know if you need to assert about the state here. I suppose you'll already get an error if this isn't true, but, this would indicate something else already went wrong.

streaming/src/main/scala/org/apache/spark/streaming/scheduler/ReceiverTracker.scala
```diff
@@ -154,6 +147,17 @@ class ReceiverTracker(ssc: StreamingContext, skipReceiverLaunch: Boolean = false
       throw new SparkException("ReceiverTracker already started")
     }
 
+    if (receivedBlockTracker == null) {
```
srowen: It looks like this is meant to be restartable, in that it will allow a stopped tracker to call `start()` again. If that's the case, then `receivedBlockTracker` needs to be nulled in `stop()`, and then this check isn't needed I think because it should really be null. In fact if you like you can require that it's null, and that endpoint is null, in `start()`?

But then ... I think this method needs to throw an exception also if `isTrackerStopping` at the start. It won't observe the intermediate state because of syncrhonization, but it could be left in this state if stopping failed.

I am wary of going too far down a path of fixing or changing the corner cases here. I do agree that there's a legitimate mismatch between where the tracker is initialized and stopped, and I think the changes you made and that possibly the ones I mention here together improve the situation without complicating it further.

srowen (changes requested):

streaming/src/main/scala/org/apache/spark/streaming/scheduler/ReceiverTracker.scala
HyukjinKwon: Thank you so much for detailed explanation. Will try to follow this.

HyukjinKwon:
retest this please

HyukjinKwon:
Ll

HyukjinKwon:
Oh wait, it seems the failed test is related. Will take a look.

HyukjinKwon:
Hm.. @srowen @tdas and @mridulm . I just got rid of the changes in `ReceiverTracker.scala` here for now. It seems we can't make this `null` in `stop()` and create an instance in `start()` because this instance is being accessed after gracefully stopped. It seems the blocks should be able to be retrieved after stopped.

One way I am thinking is, initiate this `ReceivedBlockTracker` instance at `start()` in `ReceiverTracker` and add null checks in each method using `ReceivedBlockTracker` instance in `ReceiverTracker`. However, I am not sure whether it is worth adding such checks and that this is right or cleanest way as I am not familiar with this streaming part.

Could I please ask anyone who is familiar with this to make a followup if it is fine?

streaming/src/test/scala/org/apache/spark/streaming/MapWithStateSuite.scala
```diff
@@ -38,19 +38,16 @@ class MapWithStateSuite extends SparkFunSuite
...
-    StreamingContext.getActive().foreach { _.stop(stopSparkContext = false) }
-    if (checkpointDir != null) {
-      Utils.deleteRecursively(checkpointDir)
-    }
+    StreamingContext.getActive().foreach(_.stop(stopSparkContext = false))
   }
 
   override def beforeAll(): Unit = {
     super.beforeAll()
+    checkpointDir = Utils.createTempDir("checkpoint")
```
srowen: This might happen to be OK, but now the dir is not deleted between tests. Is that going to be OK?

srowen (approved):
OK perhaps except for one last question.

streaming/src/test/scala/org/apache/spark/streaming/MapWithStateSuite.scala
HyukjinKwon: Actually, it seems the original codes are already fine in here.. I took a look and ran several tests and it seems now I may understand why @taoli91 tried to fix the codes like this.

It seems somehow the state in `ReceiverTracker` went wrong on Windows and ended up without closing `checkpointDir`. It seems the original test was failed[1] due to the issue in `ReceiverTracker`. So, I think he tried to only create/delete the folder once[2] and ensure closing in `ReceiverTracker`.

I tested this with manually stopping  `ReceivedBlockTracker` regardless of the state (the original proposal) and it seems fine without the changes in here, `MapWithStateSuite.scala`[3]. Of course,  it is fine with this change[4]. 


[1]https://ci.appveyor.com/project/spark-test/spark/build/56-F88EDDAF-E576-4787-9530-A4185FC46B1E
[2]https://ci.appveyor.com/project/spark-test/spark/build/57-test-MapWithStateSuite
[3]https://ci.appveyor.com/project/spark-test/spark/build/58-test-MapWithStateSuite
[4]https://ci.appveyor.com/project/spark-test/spark/build/59-test-MapWithStateSuite

HyukjinKwon: I believe this change is not valid. I will get rid of this. Thank you for pointing this out.

HyukjinKwon: BTW, the reason why it was failed on Windows without ensuring stopping `ReceivedBlockTracker` seems, the directory, `checkpointDir`, is being opened so it fails to delete the directory.

srowen (approved):

mridulm:
LGTM, merging it into master.
Thx @HyukjinKwon

srowen:
I'm going to merge to 2.1 as well to match https://github.com/apache/spark/pull/15320

srowen:
Scratch that; it doesn't merge cleanly. Let's leave it.

PR #12394, opened 2016-04-14
[SPARK-14634][ML] Add BisectingKMeansSummary

zhengruifeng:
## What changes were proposed in this pull request?

Add BisectingKMeansSummary
## How was this patch tested?

unit test

PR #15619, opened 2016-10-25
[SPARK-14634][ML][FOLLOWUP] Delete superfluous line in BisectingKMeans 

zhengruifeng:

As commented by @jkbradley in https://github.com/apache/spark/pull/12394, `model.setSummary(summary)` is superfluous
## How was this patch tested?

existing tests

### DIFF
--- a/mllib/src/main/scala/org/apache/spark/ml/clustering/BisectingKMeans.scala
+++ b/mllib/src/main/scala/org/apache/spark/ml/clustering/BisectingKMeans.scala
@@ -264,9 +264,9 @@ class BisectingKMeans @Since("2.0.0") (
     val model = copyValues(new BisectingKMeansModel(uid, parentModel).setParent(this))
     val summary = new BisectingKMeansSummary(
       model.transform(dataset), $(predictionCol), $(featuresCol), $(k))
-    val m = model.setSummary(summary)
-    instr.logSuccess(m)
-    m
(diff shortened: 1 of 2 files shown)


mllib/src/main/scala/org/apache/spark/ml/clustering/BisectingKMeans.scala
```diff
@@ -264,7 +264,6 @@ class BisectingKMeans @Since("2.0.0") (
     val model = copyValues(new BisectingKMeansModel(uid, parentModel).setParent(this))
     val summary = new BisectingKMeansSummary(
       model.transform(dataset), $(predictionCol), $(featuresCol), $(k))
-    model.setSummary(summary)
```
srowen: While we're here, isn't the rest of this method simpler as

```
model.setSummary(summary)
instr.logSuccess(model)
model
```

zhengruifeng: @srowen  Using another val `m` is to keep in line with `KMeans`.
But after refering to more algos (GMM/LoR/LiR), I think both `KMeans` and `BisectingKMeans` can be changed according to your comments. Thanks.

srowen (approved):

yanboliang:
LGTM, merged into master. Thanks!

PR #13235, opened 2016-05-20
[SPARK-15327] [SQL] fix split expression in whole stage codegen

davies:
## What changes were proposed in this pull request?

Right now, we will split the code for expressions into multiple functions when it exceed 64k, which requires that the the expressions are using Row object, but this is not true for whole-state codegen, it will fail to compile after splitted.

This PR will not split the code in whole-stage codegen.
## How was this patch tested?

Added regression tests.

PR #15620, opened 2016-10-25
[SPARK-18091] [SQL] Deep if expressions cause Generated SpecificUnsafeProjection code to exceed JVM code size limit

kapilsingh5050:

Fix for SPARK-18091 which is a bug related to large if expressions causing generated SpecificUnsafeProjection code to exceed JVM code size limit. 

This PR changes if expression's code generation to place its predicate, true value and false value expressions' generated code in separate methods in context so as to never generate too long combined code.
## How was this patch tested?

Added a unit test and also tested manually with the application (having transformations similar to the unit test) which caused the issue to be identified in the first place.

### DIFF
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/conditionalExpressions.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/conditionalExpressions.scala
@@ -67,7 +67,10 @@ case class If(predicate: Expression, trueValue: Expression, falseValue: Expressi
     // place generated code of condition, true value and false value in separate methods if
     // their code combined is large
     val combinedLength = condEval.code.length + trueEval.code.length + falseEval.code.length
-    val generatedCode = if (combinedLength > 16 * 1000) {
+    val generatedCode = if (combinedLength > 1024 &&
+      // Split these expressions only if they are created from a row object
+      (ctx.INPUT_ROW != null && ctx.currentVars == null)) {
+
       val (condFuncName, condGlobalIsNull, condGlobalValue) =
         createAndAddFunction(ctx, condEval, predicate, "evalIfCondExpr")
       val (trueFuncName, trueGlobalIsNull, trueGlobalValue) =
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/conditionalExpressions.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/conditionalExpressions.scala
@@ -72,11 +72,11 @@ case class If(predicate: Expression, trueValue: Expression, falseValue: Expressi
       (ctx.INPUT_ROW != null && ctx.currentVars == null)) {
 
       val (condFuncName, condGlobalIsNull, condGlobalValue) =
-        createAndAddFunction(ctx, condEval, predicate, "evalIfCondExpr")
+        createAndAddFunction(ctx, condEval, predicate.dataType, "evalIfCondExpr")
       val (trueFuncName, trueGlobalIsNull, trueGlobalValue) =
-        createAndAddFunction(ctx, trueEval, trueValue, "evalIfTrueExpr")
+        createAndAddFunction(ctx, trueEval, trueValue.dataType, "evalIfTrueExpr")
       val (falseFuncName, falseGlobalIsNull, falseGlobalValue) =
-        createAndAddFunction(ctx, falseEval, falseValue, "evalIfFalseExpr")
+        createAndAddFunction(ctx, falseEval, falseValue.dataType, "evalIfFalseExpr")
       s"""
         $condFuncName(${ctx.INPUT_ROW});
         boolean ${ev.isNull} = false;
@@ -112,13 +112,15 @@ case class If(predicate: Expression, trueValue: Expression, falseValue: Expressi
     ev.copy(code = generatedCode)
   }
 
-  private def createAndAddFunction(ctx: CodegenContext, ev: ExprCode, expr: Expression,
+  private def createAndAddFunction(ctx: CodegenContext,
+                                   ev: ExprCode,
+                                   dataType: DataType,
                                    baseFuncName: String): (String, String, String) = {
     val globalIsNull = ctx.freshName("isNull")
     ctx.addMutableState("boolean", globalIsNull, s"$globalIsNull = false;")
     val globalValue = ctx.freshName("value")
-    ctx.addMutableState(ctx.javaType(expr.dataType), globalValue,
-      s"$globalValue = ${ctx.defaultValue(expr.dataType)};")
+    ctx.addMutableState(ctx.javaType(dataType), globalValue,
+      s"$globalValue = ${ctx.defaultValue(dataType)};")
     val funcName = ctx.freshName(baseFuncName)
     val funcBody =
       s"""
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/conditionalExpressions.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/conditionalExpressions.scala
@@ -112,10 +112,11 @@ case class If(predicate: Expression, trueValue: Expression, falseValue: Expressi
     ev.copy(code = generatedCode)
   }
 
-  private def createAndAddFunction(ctx: CodegenContext,
-                                   ev: ExprCode,
-                                   dataType: DataType,
-                                   baseFuncName: String): (String, String, String) = {
+  private def createAndAddFunction(
+      ctx: CodegenContext,
+      ev: ExprCode,
+      dataType: DataType,
+      baseFuncName: String): (String, String, String) = {
     val globalIsNull = ctx.freshName("isNull")
     ctx.addMutableState("boolean", globalIsNull, s"$globalIsNull = false;")
     val globalValue = ctx.freshName("value")
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/conditionalExpressions.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/conditionalExpressions.scala
@@ -64,19 +64,69 @@ case class If(predicate: Expression, trueValue: Expression, falseValue: Expressi
     val trueEval = trueValue.genCode(ctx)
     val falseEval = falseValue.genCode(ctx)
 
-    ev.copy(code = s"""
-      ${condEval.code}
-      boolean ${ev.isNull} = false;
-      ${ctx.javaType(dataType)} ${ev.value} = ${ctx.defaultValue(dataType)};
-      if (!${condEval.isNull} && ${condEval.value}) {
-        ${trueEval.code}
-        ${ev.isNull} = ${trueEval.isNull};
-        ${ev.value} = ${trueEval.value};
-      } else {
-        ${falseEval.code}
-        ${ev.isNull} = ${falseEval.isNull};
-        ${ev.value} = ${falseEval.value};
-      }""")
+    // place generated code of condition, true value and false value in separate methods if
+    // their code combined is large
+    val combinedLength = condEval.code.length + trueEval.code.length + falseEval.code.length
+    val generatedCode = if (combinedLength > 16 * 1000) {
+      val (condFuncName, condGlobalIsNull, condGlobalValue) =
+        createAndAddFunction(ctx, condEval, predicate, "evalIfCondExpr")
+      val (trueFuncName, trueGlobalIsNull, trueGlobalValue) =
+        createAndAddFunction(ctx, trueEval, trueValue, "evalIfTrueExpr")
+      val (falseFuncName, falseGlobalIsNull, falseGlobalValue) =
+        createAndAddFunction(ctx, falseEval, falseValue, "evalIfFalseExpr")
+      s"""
+        $condFuncName(${ctx.INPUT_ROW});
+        boolean ${ev.isNull} = false;
+        ${ctx.javaType(dataType)} ${ev.value} = ${ctx.defaultValue(dataType)};
+        if (!$condGlobalIsNull && $condGlobalValue) {
+          $trueFuncName(${ctx.INPUT_ROW});
+          ${ev.isNull} = $trueGlobalIsNull;
+          ${ev.value} = $trueGlobalValue;
+        } else {
+          $falseFuncName(${ctx.INPUT_ROW});
+          ${ev.isNull} = $falseGlobalIsNull;
+          ${ev.value} = $falseGlobalValue;
+        }
+      """
+    }
+    else {
+      s"""
+        ${condEval.code}
+        boolean ${ev.isNull} = false;
+        ${ctx.javaType(dataType)} ${ev.value} = ${ctx.defaultValue(dataType)};
+        if (!${condEval.isNull} && ${condEval.value}) {
+          ${trueEval.code}
+          ${ev.isNull} = ${trueEval.isNull};
+          ${ev.value} = ${trueEval.value};
+        } else {
+          ${falseEval.code}
+          ${ev.isNull} = ${falseEval.isNull};
+          ${ev.value} = ${falseEval.value};
+        }
+      """
+    }
+
+    ev.copy(code = generatedCode)
+  }
+
+  private def createAndAddFunction(ctx: CodegenContext, ev: ExprCode, expr: Expression,
+                                   baseFuncName: String): (String, String, String) = {
+    val globalIsNull = ctx.freshName("isNull")
+    ctx.addMutableState("boolean", globalIsNull, s"$globalIsNull = false;")
+    val globalValue = ctx.freshName("value")
+    ctx.addMutableState(ctx.javaType(expr.dataType), globalValue,
+      s"$globalValue = ${ctx.defaultValue(expr.dataType)};")
+    val funcName = ctx.freshName(baseFuncName)
+    val funcBody =
+      s"""
+         |private void $funcName(InternalRow ${ctx.INPUT_ROW}) {
+         |  ${ev.code.trim}
+         |  $globalIsNull = ${ev.isNull};
+         |  $globalValue = ${ev.value};
+         |}
+         """.stripMargin
+    ctx.addNewFunction(funcName, funcBody)
+    (funcName, globalIsNull, globalValue)
   }
 
   override def toString: String = s"if ($predicate) $trueValue else $falseValue"
--- a/sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/CodeGenerationSuite.scala
+++ b/sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/CodeGenerationSuite.scala
@@ -97,6 +97,38 @@ class CodeGenerationSuite extends SparkFunSuite with ExpressionEvalHelper {
     assert(actual(0) == cases)
   }
 
+  test("SPARK-18091: split large if expressions into blocks due to JVM code size limit") {
+    val row = create_row("afafFAFFsqcategory2dadDADcategory8sasasadscategory24", 0)
+    val inputStr = 'a.string.at(0)
+    val inputIdx = 'a.int.at(1)
+
+    val length = 10
+    val valuesToCompareTo = for (i <- 1 to (length + 1)) yield ("category" + i)
+
+    val initCondition = EqualTo(RegExpExtract(inputStr, Literal("category1"), inputIdx),
+      valuesToCompareTo(0))
+    var res: Expression = If(initCondition, Literal("category1"), Literal("NULL"))
+    var cummulativeCondition: Expression = Not(initCondition)
+    for (index <- 1 to length) {
+      val valueExtractedFromInput = RegExpExtract(inputStr,
+        Literal("category" + (index + 1).toString), inputIdx)
+      val currComparee = If(cummulativeCondition, valueExtractedFromInput, Literal("NULL"))
+      val currCondition = EqualTo(currComparee, valuesToCompareTo(index))
+      val combinedCond = And(cummulativeCondition, currCondition)
+      res = If(combinedCond, If(combinedCond, valueExtractedFromInput, Literal("NULL")), res)
+      cummulativeCondition = And(Not(currCondition), cummulativeCondition)
+    }
+
+    val expressions = Seq(res)
+    val plan = GenerateUnsafeProjection.generate(expressions, true)
+    val actual = plan(row).toSeq(expressions.map(_.dataType))
+    val expected = Seq(UTF8String.fromString("category2"))
+
+    if (!checkResult(actual, expected)) {
+      fail(s"Incorrect Evaluation: expressions: $expressions, actual: $actual, expected: $expected")
+    }
+  }
+
   test("SPARK-14793: split wide array creation into blocks due to JVM code size limit") {
     val length = 5000
     val expressions = Seq(CreateArray(List.fill(length)(EqualTo(Literal(1), Literal(1)))))
--- a/sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/CodeGenerationSuite.scala
+++ b/sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/CodeGenerationSuite.scala
@@ -98,31 +98,20 @@ class CodeGenerationSuite extends SparkFunSuite with ExpressionEvalHelper {
   }
 
   test("SPARK-18091: split large if expressions into blocks due to JVM code size limit") {
-    val row = create_row("afafFAFFsqcategory2dadDADcategory8sasasadscategory24", 0)
-    val inputStr = 'a.string.at(0)
-    val inputIdx = 'a.int.at(1)
+    val inStr = "StringForTesting"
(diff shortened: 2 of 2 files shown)


kapilsingh5050:
@rxin @liancheng @sameeragarwal Please review

cloud-fan:
ok to test.

kapilsingh5050:
I ran the tests on my machine and they passed. @cloud-fan Can you please trigger a re-run of test build?

cloud-fan:
retest this please

kapilsingh5050:
@cloud-fan the same tests which failed earlier are failing again. Could you issue retest request one more time?

cloud-fan:
retest this please

viirya:
Looks like it is not flaky test. The test failure is due to compilation error of janino: A local variable can not be accessed in the generated function.

kapilsingh5050:
I did not want to run tests using ./dev/run-tests on my machine because it'll take hours. I ran the mllib module tests using "mvn test" which worked fine on my machine. What could be the difference between such a run on my machine and the Jenkins Test Build? Please guide me on further steps.

cloud-fan:
can you try sbt? The jenkins run tests using sbt instead of maven

viirya:
I've tested one of the failed tests locally with sbt. It failed.

kapilsingh5050:
Let me try running the unit tests with sbt

kapilsingh5050:
I have been able to reproduce the failures using sbt. Investigating the cause.

rxin:
@kapilsingh5050 any update?

kapilsingh5050:
I'm not sure whether I understand the problem completely. It seems WholeStage CodeGen wants the complete stage code in a single function. The fix I made violates this because it creates separate methods for codegen of If expression's condition, if block and then block (of course, when the total code is beyond a threshold).

Now, if my assumption about WholeStage CodeGen is correct then how does the split expressions logic work? Or maybe the problem is specific to initialisation of input attributes of a stage.

kapilsingh5050:
The unit tests passed after incorporating related changes from SPARK-15327 and SPARK-17115

kapilsingh5050:
@viirya @cloud-fan @rxin Please review and merge

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/conditionalExpressions.scala
```diff
@@ -64,19 +64,72 @@ case class If(predicate: Expression, trueValue: Expression, falseValue: Expressi
...
-        ${ev.value} = ${trueEval.value};
-      } else {
-        ${falseEval.code}
-        ${ev.isNull} = ${falseEval.isNull};
-        ${ev.value} = ${falseEval.value};
-      }""")
+    // place generated code of condition, true value and false value in separate methods if
+    // their code combined is large
+    val combinedLength = condEval.code.length + trueEval.code.length + falseEval.code.length
+    val generatedCode = if (combinedLength > 1024 &&
```
cloud-fan: The limitation is 64k, we don't need to be so conservative

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/conditionalExpressions.scala
```diff
@@ -64,19 +64,72 @@ case class If(predicate: Expression, trueValue: Expression, falseValue: Expressi
...
+          ${ev.isNull} = ${falseEval.isNull};
+          ${ev.value} = ${falseEval.value};
+        }
+      """
+    }
+
+    ev.copy(code = generatedCode)
+  }
+
+  private def createAndAddFunction(ctx: CodegenContext, ev: ExprCode, expr: Expression,
```
cloud-fan: please follow the existing code style in Spark, i.e.
```
def xxx(
  param1: xxx,
  param2: xxx): xxx
```

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/conditionalExpressions.scala
```diff
@@ -64,19 +64,72 @@ case class If(predicate: Expression, trueValue: Expression, falseValue: Expressi
...
+                                   baseFuncName: String): (String, String, String) = {
+    val globalIsNull = ctx.freshName("isNull")
+    ctx.addMutableState("boolean", globalIsNull, s"$globalIsNull = false;")
+    val globalValue = ctx.freshName("value")
+    ctx.addMutableState(ctx.javaType(expr.dataType), globalValue,
```
cloud-fan: looks like we don't need to pass in the `expr`, but only its data type

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/conditionalExpressions.scala
```diff
@@ -64,19 +64,72 @@ case class If(predicate: Expression, trueValue: Expression, falseValue: Expressi
...
-        ${falseEval.code}
-        ${ev.isNull} = ${falseEval.isNull};
-        ${ev.value} = ${falseEval.value};
-      }""")
+    // place generated code of condition, true value and false value in separate methods if
+    // their code combined is large
+    val combinedLength = condEval.code.length + trueEval.code.length + falseEval.code.length
+    val generatedCode = if (combinedLength > 1024 &&
+      // Split these expressions only if they are created from a row object
+      (ctx.INPUT_ROW != null && ctx.currentVars == null)) {
```
viirya: So if condition and true/false expressions are bound to `currentVars`, we still exceed JVM code size limit?

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/CodeGenerationSuite.scala
```diff
@@ -97,6 +97,38 @@ class CodeGenerationSuite extends SparkFunSuite with ExpressionEvalHelper {
     assert(actual(0) == cases)
   }
 
+  test("SPARK-18091: split large if expressions into blocks due to JVM code size limit") {
+    val row = create_row("afafFAFFsqcategory2dadDADcategory8sasasadscategory24", 0)
```
cloud-fan: why is this test so complicated? This test suite contains some similar tests for large code size bugs, they are all simple.

ueshin: +1

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/conditionalExpressions.scala
cloud-fan: I'm curious about how other similar patches handle whole stage codegen, any ideas?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/conditionalExpressions.scala
cloud-fan: ah i see

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/conditionalExpressions.scala
kapilsingh5050: @viirya I'm not sure but looking at the conversation on following change:
https://github.com/apache/spark/pull/13235
it looks like that

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/conditionalExpressions.scala
kapilsingh5050: Yep, I'll fix it

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/CodeGenerationSuite.scala
kapilsingh5050: This sort of simulates the scenario I faced which surfaced this issue. I tried a couple of other simple ways but those couldn't reproduce the issue. Since if expression's generated code will be large when it's constituents (i.e. predicate, true value, false value) will be large I thought I'll go with a nested recursive testcase consisting of if expressions

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/conditionalExpressions.scala
viirya: We will pass needed local variables to the separated functions.

viirya: I am not sure if we want to do it here. It will add a bit complexity. If for simplicity, I think we can not to support it now.

kapilsingh5050: I think the conversation on the change I mentioned also went like there will be some refactoring required in whole stage code generation to support that

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/conditionalExpressions.scala
```diff
@@ -64,19 +64,74 @@ case class If(predicate: Expression, trueValue: Expression, falseValue: Expressi
...
+          ${ev.isNull} = ${falseEval.isNull};
+          ${ev.value} = ${falseEval.value};
+        }
+      """
+    }
+
+    ev.copy(code = generatedCode)
+  }
+
+  private def createAndAddFunction(ctx: CodegenContext,
```
cloud-fan: the code style is still wrong, see https://github.com/apache/spark/pull/15620#discussion_r90185562

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/CodeGenerationSuite.scala
```diff
@@ -97,6 +97,38 @@ class CodeGenerationSuite extends SparkFunSuite with ExpressionEvalHelper {
...
+    val inputIdx = 'a.int.at(1)
+
+    val length = 10
+    val valuesToCompareTo = for (i <- 1 to (length + 1)) yield ("category" + i)
+
+    val initCondition = EqualTo(RegExpExtract(inputStr, Literal("category1"), inputIdx),
+      valuesToCompareTo(0))
+    var res: Expression = If(initCondition, Literal("category1"), Literal("NULL"))
+    var cummulativeCondition: Expression = Not(initCondition)
+    for (index <- 1 to length) {
```
cloud-fan: can we test something like this?
```
var stringExpr = stringAttr
for (_ <- 1 to 10) {
  stringExpr = Decode(Encoder(stringExpr, "utf-8"), "utf-8")
}
val ifExpr = If(booleanAttr, stringExpr, stringExpr)
```

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/conditionalExpressions.scala
kapilsingh5050: apologies, I'll fix it

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/CodeGenerationSuite.scala
kapilsingh5050: let me try this

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/CodeGenerationSuite.scala
viirya: I think you can construct a very long condition expression instead of the current approach. It is hard to know what this `If` expression looks like for now.

viirya: An example looks like the following test case. I've tested it. It failed in current master branch due to the 64kb issue.

    test("split large if expression") {
      val conditionExpr = (1 to 1000).map { _ =>
        And(GreaterThan(new Rand(), Literal(0.5)), LessThan(new Rand(), Literal(0.8)))
      }.reduce(And)

      val expressions = Seq(If(conditionExpr, Literal(true), Literal(false)))
      val plan = GenerateUnsafeProjection.generate(expressions, true)
      // Rand needs initialization.
      plan.initialize(1)
      val actual = plan(new GenericInternalRow(10)).toSeq(expressions.map(_.dataType))
      val expected = Seq(false)
      if (!checkResult(actual, expected)) {
        fail(s"Incorrect Evaluation: expressions: $expressions, actual: $actual, expected: $expected")
      }
    }

kapilsingh5050: I think I didn't make this clear. So the fix I've made in If expression is actually required in all of the expressions so that all the expressions break their generated code into methods whenever it grows beyond an appropriate threshold. Since the fix in this pull request is only in If expression, a working testcase with the fix will be one which consists (recursively) of If expressions. Because otherwise the test will need to be tuned so that the size of If expression's code is greater than threshold but of it's predicate, trueValue and falseValue doesn't grow beyond threshold

viirya: > Since the fix in this pull request is only in If expression, a working testcase with the fix will be one which consists (recursively) of If expressions.

I can't see why this is reasonable. You just need to construct enough big generated codes of condition + true expression + false expression.

kapilsingh5050:
I didn't find any failure/error message in the PR Builder job logs apart from:

    main()
KeyError: -9
[error] running /home/jenkins/workspace/SparkPullRequestBuilder@3/build/sbt -Phadoop-2.3 -Phive-thriftserver -Phive -Dtest.exclude.tags=org.apache.spark.tags.ExtendedHiveTest,org.apache.spark.tags.ExtendedYarnTest hive-thriftserver/test mllib/test hive/test examples/test sql/test sql-kafka-0-10/test catalyst/test ; process was terminated by signal 9
Build step 'Execute shell' marked build as failure

Could it be because of OOM? Or should we retry in case this is a transient issue.

cloud-fan:
retest this please

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/CodeGenerationSuite.scala
```diff
@@ -97,6 +97,27 @@ class CodeGenerationSuite extends SparkFunSuite with ExpressionEvalHelper {
...
+    val inputStrAttr = 'a.string.at(0)
+
+    var strExpr: Expression = inputStrAttr
+    for (_ <- 1 to 13) {
+      strExpr = If(EqualTo(Decode(Encode(strExpr, "utf-8"), "utf-8"), inputStrAttr),
+        strExpr, strExpr)
+    }
+
+    val expressions = Seq(strExpr)
+    val plan = GenerateUnsafeProjection.generate(expressions, true)
```
cloud-fan: nit: other similar tests use `GenerateMutableProjection` to test the codegen, any specific reason we have to use `GenerateUnsafeProjection` here?

viirya:
This looks good now.

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/CodeGenerationSuite.scala
kapilsingh5050: The issue I had encountered involved GenerateUnsafeProjection and I was trying to reproduce the same in the unit test so I went with it. I can't think of any other reason. Any specific reason you guys use GenerateMutableProjection?

kapilsingh5050:
@viirya @cloud-fan Are we good to merge? Please let me know if there is any other action on my end.

viirya:
LGTM and cc @cloud-fan for another check.

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/CodeGenerationSuite.scala
cloud-fan: no specific reason too.

cloud-fan:

kapilsingh5050:
@cloud-fan Can we merge this change to 1.6 and 2.0 too?

cloud-fan:
merged to 2.0. Can you send a new PR to backport this to 1.6? There are a lot of code changes beween 1.6 and 2.1, it's safer to open a PR and run all tests.

kapilsingh5050:
@cloud-fan created https://github.com/apache/spark/pull/16146 for backporting to 1.6

srowen:
@cloud-fan @kapilsingh5050 it's weird, but it looks like after this change, all of the Maven-based 2.0 Jenkins jobs now time out. It looks consistent: https://amplab.cs.berkeley.edu/jenkins/view/Spark%20QA%20Test%20(Dashboard)/

They all end like this:

```
CodeGenerationSuite:
- multithreaded eval
- metrics are recorded on compile
- SPARK-8443: split wide projections into blocks due to JVM code size limit
- SPARK-13242: case-when expression with large number of branches (or cases)
Exception in thread "ScalaTest-dispatcher" java.lang.OutOfMemoryError: Java heap space
	at java.util.Arrays.copyOf(Arrays.java:3332)
	at java.lang.AbstractStringBuilder.expandCapacity(AbstractStringBuilder.java:137)
	at java.lang.AbstractStringBuilder.ensureCapacityInternal(AbstractStringBuilder.java:121)
	at java.lang.AbstractStringBuilder.append(AbstractStringBuilder.java:421)
	at java.lang.StringBuilder.append(StringBuilder.java:136)
	at scala.collection.mutable.StringBuilder.append(StringBuilder.scala:200)
	at scala.collection.TraversableOnce$$anonfun$addString$1.apply(TraversableOnce.scala:364)
	at scala.collection.IndexedSeqOptimized$class.foreach(IndexedSeqOptimized.scala:33)
	at scala.collection.mutable.ArrayOps$ofRef.foreach(ArrayOps.scala:186)
	at scala.collection.TraversableOnce$class.addString(TraversableOnce.scala:357)
	at scala.collection.mutable.ArrayOps$ofRef.addString(ArrayOps.scala:186)
	at scala.collection.TraversableOnce$class.mkString(TraversableOnce.scala:323)
	at scala.collection.mutable.ArrayOps$ofRef.mkString(ArrayOps.scala:186)
	at scala.collection.TraversableOnce$class.mkString(TraversableOnce.scala:325)
	at scala.collection.mutable.ArrayOps$ofRef.mkString(ArrayOps.scala:186)
	at org.scalatest.tools.Fragment.toPossiblyColoredText(Fragment.scala:28)
	at org.scalatest.tools.PrintReporter.printPossiblyInColor(PrintReporter.scala:136)
	at org.scalatest.tools.StringReporter$$anonfun$apply$1.apply(StringReporter.scala:180)
	at org.scalatest.tools.StringReporter$$anonfun$apply$1.apply(StringReporter.scala:180)
	at org.scalatest.tools.StringReporter.apply(StringReporter.scala:180)
	at org.scalatest.tools.PrintReporter.apply(PrintReporter.scala:141)
	at org.scalatest.DispatchReporter$Propagator$$anonfun$run$1.apply(DispatchReporter.scala:240)
	at org.scalatest.DispatchReporter$Propagator$$anonfun$run$1.apply(DispatchReporter.scala:239)
	at org.scalatest.DispatchReporter$Propagator.run(DispatchReporter.scala:239)
Build timed out (after 210 minutes). Marking the build as aborted.
Build was aborted
Archiving artifacts
Recording test results
Finished: ABORTED
```

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/CodeGenerationSuite.scala
```diff
@@ -97,6 +97,27 @@ class CodeGenerationSuite extends SparkFunSuite with ExpressionEvalHelper {
...
   }
 
+  test("SPARK-18091: split large if expressions into blocks due to JVM code size limit") {
+    val inStr = "StringForTesting"
+    val row = create_row(inStr)
+    val inputStrAttr = 'a.string.at(0)
+
+    var strExpr: Expression = inputStrAttr
+    for (_ <- 1 to 13) {
+      strExpr = If(EqualTo(Decode(Encode(strExpr, "utf-8"), "utf-8"), inputStrAttr),
```
cloud-fan: cc @srowen I think this is the root cause. This test is an overkill, although this PR fixed the code size limitation problem, this test may still hit constants pool size limitation, which is a known limiation and has not been fixed yet. It seems that maven and sbt have different JVM settings when run test, so the problem only exists at maven side.

I'm going to submit a PR to simplify this test a bit.

a-roberts: Very interested in this, we know with 5 iterations instead of a 13 we don't get the problem (which makes sense as we'd only be generating so much code, not hitting the 64k constant pool size limit). I'm testing with IBM's SDK for Java so curious if it manifests itself in a different way for us or if we have a problem to fix on our end.

I have log files exceeding 2 GB from the test printing the generated code on failure. 

If we add prints for the strExpr we see something like

The problem is
```
CodeGenerationSuite:
- multithreaded eval
- metrics are recorded on compile
- SPARK-8443: split wide projections into blocks due to JVM code size limit
- SPARK-13242: case-when expression with large number of branches (or cases)
- SPARK-18091: split large if expressions into blocks due to JVM code size limit *** FAILED ***
  java.util.concurrent.ExecutionException: java.lang.Exception: failed to compile: org.codehaus.janino.JaninoRuntimeException: Constant pool for class org.apache.spark.sql.catalyst.expressions.GeneratedClass$SpecificUnsafeProjection has grown past JVM limit of 0xFFFF
/* 001 */ public java.lang.Object generate(Object[] references) {
/* 002 */   return new SpecificUnsafeProjection(references);
/* 003 */ }
/* 004 */
/* 005 */ class SpecificUnsafeProjection extends org.apache.spark.sql.catalyst.expressions.UnsafeProjection {
/* 006 */
/* 007 */   private Object[] references;
/* 008 */   private boolean isNull42;
/* 009 */   private boolean value42;
/* 010 */   private boolean isNull43;
/* 011 */   private UTF8String value43;
/* 012 */   private boolean isNull44;
/* 013 */   private UTF8String value44;
/* 014 */   private boolean isNull58;
```

With my prints:
```
debug, row: [StringForTesting]
input string attr: input[0, string, true]

in the loop
strExpr is: if ((decode(encode(input[0, string, true], utf-8), utf-8) = input[0, string, true])) input[0, string, true] else input[0, string, true]

in the loop
strExpr is: if ((decode(encode(if ((decode(encode(input[0, string, true], utf-8), utf-8) = input[0, string, true])) input[0, string, true] else input[0, string, true], utf-8), utf-8) = input[0, string, true])) if ((decode(encode(input[0, string, true], utf-8), utf-8) = input[0, string, true])) input[0, string, true] else input[0, string, true] else if ((decode(encode(input[0, string, true], utf-8), utf-8) = input[0, string, true])) input[0, string, true] else input[0, string, true]

in the loop
strExpr is: if ((decode(encode(if ((decode(encode(if ((decode(encode(input[0, string, true], utf-8), utf-8) = input[0, string, true])) input[0, string, true] else input[0, string, true], utf-8), utf-8) = input[0, string, true])) if ((decode(encode(input[0, string, true], utf-8), utf-8) = input[0, string, true])) input[0, string, true] else input[0, string, true] else if ((decode(encode(input[0, string, true], utf-8), utf-8) = input[0, string, true])) input[0, string, true] else input[0, string, true], utf-8), utf-8) = input[0, string, true])) if ((decode(encode(if ((decode(encode(input[0, string, true], utf-8), utf-8) = input[0, string, true])) input[0, string, true] else input[0, string, true], utf-8), utf-8) = input[0, string, true])) if ((decode(encode(input[0, string, true], utf-8), utf-8) = input[0, string, true])) input[0, string, true] else input[0, string, true] else if ((decode(encode(input[0, string, true], utf-8), utf-8) = input[0, string, true])) input[0, string, true] else input[0, string, true] else if ((decode(encode(if ((decode(encode(input[0, string, true], utf-8), utf-8) = input[0, string, true])) input[0, string, true] else input[0, string, true], utf-8), utf-8) = input[0, string, true])) if ((decode(encode(input[0, string, true], utf-8), utf-8) = input[0, string, true])) input[0, string, true] else input[0, string, true] else if ((decode(encode(input[0, string, true], utf-8), utf-8) = input[0, string, true])) input[0, string, true] else input[0, string, true]
etc
```

Is this the same issue we're referring to? I've also seen timeouts and our Jenkins farm has 5h limits, I see the problem against branch-2.0, branch-2.1, master, but didn't see it for Spark 2.1.0 RC1.

viirya: @a-roberts I think this is merged into branch-2.1 after RC1 coming out.

cloud-fan:
the test is good now https://amplab.cs.berkeley.edu/jenkins/view/Spark%20QA%20Test%20(Dashboard)/job/spark-branch-2.0-test-maven-hadoop-2.2/

PR #15621, opened 2016-10-25
[SPARK-17748][FOLLOW-UP][ML] Reorg variables of WeightedLeastSquares.

yanboliang:

This is follow-up work of #15394.
Reorg some variables of `WeightedLeastSquares` and fix one minor issue of `WeightedLeastSquaresSuite`.
## How was this patch tested?

Existing tests.

### DIFF
--- a/mllib/src/main/scala/org/apache/spark/ml/optim/WeightedLeastSquares.scala
+++ b/mllib/src/main/scala/org/apache/spark/ml/optim/WeightedLeastSquares.scala
@@ -137,48 +137,57 @@ private[ml] class WeightedLeastSquares(
     val bbBar = summary.bbBar / (bStd * bStd)
 
     val aStd = summary.aStd
+    val aStdValues = aStd.values
+
     val aBar = {
       val _aBar = summary.aBar
+      val _aBarValues = _aBar.values
       var i = 0
       // scale aBar to standardized space in-place
       while (i < numFeatures) {
-        if (aStd(i) == 0.0) {
-          _aBar.values(i) = 0.0
+        if (aStdValues(i) == 0.0) {
+          _aBarValues(i) = 0.0
         } else {
-          _aBar.values(i) /= aStd(i)
+          _aBarValues(i) /= aStdValues(i)
         }
         i += 1
       }
       _aBar
     }
+    val aBarValues = aBar.values
+
     val abBar = {
       val _abBar = summary.abBar
+      val _abBarValues = _abBar.values
       var i = 0
       // scale abBar to standardized space in-place
       while (i < numFeatures) {
-        if (aStd(i) == 0.0) {
-          _abBar.values(i) = 0.0
+        if (aStdValues(i) == 0.0) {
+          _abBarValues(i) = 0.0
         } else {
-          _abBar.values(i) /= (aStd(i) * bStd)
+          _abBarValues(i) /= (aStdValues(i) * bStd)
         }
         i += 1
       }
       _abBar
     }
+    val abBarValues = abBar.values
+
     val aaBar = {
       val _aaBar = summary.aaBar
+      val _aaBarValues = _aaBar.values
       var j = 0
       var p = 0
       // scale aaBar to standardized space in-place
       while (j < numFeatures) {
-        val aStdJ = aStd.values(j)
+        val aStdJ = aStdValues(j)
         var i = 0
         while (i <= j) {
-          val aStdI = aStd.values(i)
+          val aStdI = aStdValues(i)
           if (aStdJ == 0.0 || aStdI == 0.0) {
-            _aaBar.values(p) = 0.0
+            _aaBarValues(p) = 0.0
           } else {
-            _aaBar.values(p) /= (aStdI * aStdJ)
+            _aaBarValues(p) /= (aStdI * aStdJ)
           }
           p += 1
           i += 1
@@ -187,6 +196,7 @@ private[ml] class WeightedLeastSquares(
       }
       _aaBar
     }
+    val aaBarValues = aaBar.values
 
     val effectiveRegParam = regParam / bStd
     val effectiveL1RegParam = elasticNetParam * effectiveRegParam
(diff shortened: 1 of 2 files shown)


yanboliang:
@sethah I wish you would not mind, I made some reorg for variables of ```WeightedLeastSquares```. Since we use ```bBar, bbBar, aBar, abBar, aaBar``` in standardized space always, I put the standardization code and variable declaration together. Then if developers would like to know the detail of each variable, they only need to refer one place which I think is more clear to be understand. Meanwhile, I improved some naming and fix the minor issue in ```WeightedLeastSquaresSuite```. I wish you like this change and open to hear your thoughts. If you think it's not necessary, I can also revert that part. Thanks.

mllib/src/main/scala/org/apache/spark/ml/optim/WeightedLeastSquares.scala
```diff
@@ -280,7 +289,7 @@ private[ml] class WeightedLeastSquares(
       solution.objectiveHistory.getOrElse(Array(0D)))
   }
 
-  /** Construct A^T^ A from summary statistics. */
+  /** Construct A^T^ A (append bias if necessary). */
```
yanboliang: Summary statistics are in original space, here we construct A^T^ A from standardized space.

mllib/src/main/scala/org/apache/spark/ml/optim/WeightedLeastSquares.scala
```diff
@@ -137,61 +133,68 @@ private[ml] class WeightedLeastSquares(
...
-    var j = 0
-    while (j < numFeatures) {
-      if (aStd(j) == 0.0) {
-        aBarValues(j) = 0.0
-      } else {
-        aBarValues(j) /= aStd(j)
-      }
-      j += 1
-    }
+    val bBar = summary.bBar / bStd
```
sethah: I prefer the names `bBarStd` and `bbBarStd` here, as I think they're more descriptive. But it is not a strong preference so I'll leave it up to you.

mllib/src/main/scala/org/apache/spark/ml/optim/WeightedLeastSquares.scala
```diff
@@ -137,61 +133,68 @@ private[ml] class WeightedLeastSquares(
...
+    val abBar = {
+      val _abBar = summary.abBar
       var i = 0
-      while (i <= j) {
-        val aStdI = aStdValues(i)
-        if (aStdJ == 0.0 || aStdI == 0.0) {
-          aaBarValues(p) = 0.0
+      // scale abBar to standardized space in-place
+      while (i < numFeatures) {
+        if (aStd(i) == 0.0) {
```
sethah: similar thing here, you're calling `aStd.apply(i)` which in turn calls an internal getter for `aStd.values` which then indexes the array.

mllib/src/main/scala/org/apache/spark/ml/optim/WeightedLeastSquares.scala
yanboliang: Since we use ```bBar, bbBar, aBar, abBar, aaBar``` in standardized space always, so I did not append ```Std``` as suffix for all variables. If we only add suffix for ```bBar``` and ```bbBar```, developers may misinterpret that other variables are not in standardized space.

mllib/src/main/scala/org/apache/spark/ml/optim/WeightedLeastSquares.scala
```diff
@@ -137,61 +133,68 @@ private[ml] class WeightedLeastSquares(
...
-      } else {
-        abBarValues(j) /= (aStdValues(j) * bStd)
+    val aStd = summary.aStd
+    val aBar = {
+      val _aBar = summary.aBar
+      var i = 0
+      // scale aBar to standardized space in-place
+      while (i < numFeatures) {
+        if (aStd(i) == 0.0) {
+          _aBar.values(i) = 0.0
```
yanboliang: Yeah, I total agree. I thought the impact of this issue is trivial, but making implicit pointer definitely should be better. I made update for all variables. Thanks.

sethah: You're probably right. The only thing it might have a real impact is `aaBar`, but since we use while loops for speed, it seems better to also use this optimization as well.

mllib/src/main/scala/org/apache/spark/ml/optim/WeightedLeastSquares.scala
```diff
@@ -137,65 +133,82 @@ private[ml] class WeightedLeastSquares(
...
+          _aBarValues(i) = 0.0
+        } else {
+          _aBarValues(i) /= aStdValues(i)
+        }
+        i += 1
       }
-      j += 1
+      _aBar
     }
+    val aBarValues = aBar.values
```
sethah: Why don't we just make the above be `val aBarValues = { ... }` returning `_aBarValues` from the code block instead of `_abar`. Since we make a pointer to `aBar` but then only use it to get its values.

sethah (commented):
One minor thing, otherwise LGTM.

mllib/src/main/scala/org/apache/spark/ml/optim/WeightedLeastSquares.scala
yanboliang: I'd like to keep their types consistent at interface level, and the following two scenarios:
```Scala
def solve(
      bBar: Double,
      bbBar: Double,
      abBar: DenseVector,
      aaBar: DenseVector,
      aBar: Array): NormalEquationSolution
```
and
```Scala
def solve(
      bBar: Double,
      bbBar: Double,
      abBar: Array,
      aaBar: Array,
      aBar: DenseVector): NormalEquationSolution
```
are not good. So I left them as it is. Is this OK for you?

sethah: You don't have to change the `solve` signature to intermix arrays and vectors since we just pass `aa` and `ab`. Still, this was just a suggestion to avoid a few lines of code, so let's just leave it as is.

yanboliang: Ok, I got what you mean. I'll leave it as is for now. Thanks.

yanboliang:
I'll merge it into master. Thanks for reviewing! @sethah

PR #15623, opened 2016-10-25
[SPARK-18093][SQL] Fix default value test in SQLConfSuite to work rega…

markgrover:
…rdless of warehouse dir's existence

Appending a trailing slash, if there already isn't one for the 
sake comparison of the two paths. It doesn't take away from
the essence of the check, but removes any potential mismatch
due to lack of trailing slash.
## How was this patch tested?

Ran unit tests and they passed.

### DIFF
--- a/sql/core/src/test/scala/org/apache/spark/sql/internal/SQLConfSuite.scala
+++ b/sql/core/src/test/scala/org/apache/spark/sql/internal/SQLConfSuite.scala
@@ -215,12 +215,18 @@ class SQLConfSuite extends QueryTest with SharedSQLContext {
   }
 
   test("default value of WAREHOUSE_PATH") {
+    def appendTrailingSlashIfNecessary(path: String): String = {
+      if (!path.endsWith("/")) path + "/" else path
+    }
+
     val original = spark.conf.get(SQLConf.WAREHOUSE_PATH)
     try {
       // to get the default value, always unset it
       spark.conf.unset(SQLConf.WAREHOUSE_PATH.key)
-      assert(new Path(Utils.resolveURI("spark-warehouse")).toString ===
-        spark.sessionState.conf.warehousePath + "/")
+      // JVM adds a trailing slash if the directory exists and leaves it as-is, if it doesn't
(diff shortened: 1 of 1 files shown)


srowen (approved):
OK by me. It could go the other way too with `def removeTrailingSlash(path: String) = if (path.endsWith("/")) path.init else path`

markgrover:
Thanks @srowen, indeed, it could be stripping the trailing slash too. I am ok either way too. Let me know if you have strong preference to the other one, otherwise, this should be ok.

Thanks again!

sql/core/src/test/scala/org/apache/spark/sql/internal/SQLConfSuite.scala
```diff
@@ -215,12 +215,15 @@ class SQLConfSuite extends QueryTest with SharedSQLContext {
   }
 
   test("default value of WAREHOUSE_PATH") {
+
```
vanzin: nit: shouldn't add this line but I'll clean it during merge.

vanzin:
LGTM will merge once tests pass.

sql/core/src/test/scala/org/apache/spark/sql/internal/SQLConfSuite.scala
```diff
@@ -215,12 +215,18 @@ class SQLConfSuite extends QueryTest with SharedSQLContext {
...
     val original = spark.conf.get(SQLConf.WAREHOUSE_PATH)
     try {
       // to get the default value, always unset it
       spark.conf.unset(SQLConf.WAREHOUSE_PATH.key)
-      assert(new Path(Utils.resolveURI("spark-warehouse")).toString ===
-        spark.sessionState.conf.warehousePath + "/")
+      // JVM adds a trailing slash if the directory exists and leaves it as-is, if it doesn't
+      // In our comparison, add trailing slash to both sides if necessary, to account for both cases
+      assert(appendTrailingSlashIfNecessary(new Path(Utils.resolveURI("spark-warehouse"))
```
srowen: Right forgot that one!

vanzin:
Merging to master / 2.0.

PR #15626, opened 2016-10-25
SPARK-17829 [SQL] Stable format for offset log

ghost:

Currently we use java serialization for the WAL that stores the offsets contained in each batch. This has two main issues:
It can break across spark releases (though this is not the only thing preventing us from upgrading a running query)
It is unnecessarily opaque to the user.
I'd propose we require offsets to provide a user readable serialization and use that instead. JSON is probably a good option.
## How was this patch tested?

Tests were added for KafkaSourceOffset in [KafkaSourceOffsetSuite](external/kafka-0-10-sql/src/test/scala/org/apache/spark/sql/kafka010/KafkaSourceOffsetSuite.scala) and for LongOffset in [OffsetSuite](sql/core/src/test/scala/org/apache/spark/sql/streaming/OffsetSuite.scala)

@zsxwing @marmbrus


zsxwing:
jenkins, add to whitelist

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/OffsetSeq.scala
```diff
@@ -22,7 +22,10 @@ package org.apache.spark.sql.execution.streaming
  * [[Source]]s that are present in a streaming query. This is similar to simplified, single-instance
  * vector clock that must progress linearly forward.
  */
-case class CompositeOffset(offsets: Seq[Option[Offset]]) extends Offset {
+case class OffsetSeq(offsets: Seq[Option[Offset]]) {
```
tdas: Why was this renamed?

external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/KafkaSourceOffset.scala
```diff
@@ -27,13 +28,21 @@ import org.apache.spark.sql.execution.streaming.Offset
  */
 private[kafka010]
 case class KafkaSourceOffset(partitionToOffsets: Map[TopicPartition, Long]) extends Offset {
+  import org.json4s.jackson.Serialization.{write}
```
tdas: Actually its easier to read the code later if you import Serialization, and then use Serialization.write. Same for read.

external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/KafkaSourceOffset.scala
```diff
@@ -51,4 +60,8 @@ private[kafka010] object KafkaSourceOffset {
   def apply(offsetTuples: (String, Int, Long)*): KafkaSourceOffset = {
     KafkaSourceOffset(offsetTuples.map { case(t, p, o) => (new TopicPartition(t, p), o) }.toMap)
   }
+
+  def apply(serialized: Offset): KafkaSourceOffset = {
```
tdas: If I get this right, this can be called only on SerializedOffset. In that case, best to type the param as SerializedOffset.

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/Offset.scala
```diff
@@ -23,4 +23,16 @@ package org.apache.spark.sql.execution.streaming
  * ordering of two [[Offset]] instances.  We do assume that if two offsets are `equal` then no
  * new data has arrived.
  */
-trait Offset extends Serializable {}
+abstract class Offset {
+  def json: String
+}
+
+/** Used when loading */
+class SerializedOffset(override val json: String) extends Offset
```
tdas: this can be a `case class` in which case the apply below wont be needed.

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/Offset.scala
```diff
@@ -23,4 +23,16 @@ package org.apache.spark.sql.execution.streaming
  * ordering of two [[Offset]] instances.  We do assume that if two offsets are `equal` then no
  * new data has arrived.
  */
-trait Offset extends Serializable {}
+abstract class Offset {
+  def json: String
+}
+
+/** Used when loading */
```
tdas: Improve the docs to something like "A JSON-serialized representation of an Offset that is used for saving offsets to the offset log"

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/LongOffset.scala
```diff
@@ -22,8 +22,18 @@ package org.apache.spark.sql.execution.streaming
...
+
   def +(increment: Long): LongOffset = new LongOffset(offset + increment)
   def -(decrement: Long): LongOffset = new LongOffset(offset - decrement)
 
   override def toString: String = s"#$offset"
 }
+
+object LongOffset {
+
+  def apply(serialized: Offset) : LongOffset = new LongOffset(serialized.json.toLong)
```
tdas: same comment as with SerializedOffset.apply, make the param SerializedOffset instead of Offset

tdas (changes requested):
Overall, this is a good start, but I think there one critical piece missing (or maybe i am not understanding the PR correctly).

The OffsetSeq is finally written to the log by the file HDFSMetadataLog. Currently, HDFSMetadataLog serializes the offset using Java serialization, and writes the bytes to the file. The goal of this PR is to ultimately not use java serialization, and rather use JSON text representation in the file. In this PR I dont see any changes to the HDFSMetadatalog, so the serialization format hasnt changed. All its doing is instead of Java-serializing CompositeOffset, its serializing OffsetSeq. 

I strongly suggest writing a unit test where you take a OffsetSeq, and use HDFSMetadataLog to write it a file. And see the contents of the file is a valid json or not.

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/OffsetSeq.scala
tcondie: I renamed to avoid confusing the class as being a type of Offset. Does that make sense or should I take an alternative approach?

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/LongOffset.scala
```diff
@@ -22,8 +22,18 @@ package org.apache.spark.sql.execution.streaming
...
+
+}
```
tcondie: Can you please clarify best practices w.r.t. extra lines? I didn't see anything in the code style guide.

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/Offset.scala
tcondie: Got it.

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/Offset.scala
tcondie: Makes sense.

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/OffsetSeq.scala
tdas: Aah right. Of course. I didnt notice that. That makes sense.

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/LongOffset.scala
tdas: Not a strict guideline i guess, its matter of the taste of the reviewer quite a bit. For me i guess where there is no need to visually introduce a gap, there is no need for an empty line. For example: between `def apply` and `}` there no need to introduce a separation.

koeninger:
So it looks like this is the json format being used for kafka offsets:

[{"_1":{"hash":0,"partition":0,"topic":"t"},"_2":2},{"_1":{"hash":0,"partition":1,"topic":"t"},"_2":2}]

My main concerns are that is unnecessarily verbose, and is completely different from the format being used to specify per-topicpartition offsets in org.apache.spark.sql.kafka010.JsonUtils

I see no reason why those formats should be different - a user should be able to take offsets directly from the checkpoint and use them to start a job, for instance.

tcondie:
@koeninger Thanks for pointing this out. I have revised to serialization procedure to use org.apache.spark.sql.kafka010.JsonUtils

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/CompactibleFileStreamLog.scala
```diff
@@ -37,14 +39,19 @@ import org.apache.spark.sql.SparkSession
  * compact log files every 10 batches by default into a big file. When
  * doing a compaction, it will read all old log files and merge them with the new batch.
  */
-abstract class CompactibleFileStreamLog[T: ClassTag](
+abstract class CompactibleFileStreamLog[T <: AnyRef : ClassTag](
```
tdas: Why is this AnyRef needed?

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/HDFSMetadataLog.scala
```diff
@@ -44,9 +45,14 @@ import org.apache.spark.util.UninterruptibleThread
...
  * files in a directory always shows the latest files.
  */
-class HDFSMetadataLog[T: ClassTag](sparkSession: SparkSession, path: String)
+class HDFSMetadataLog[T <: AnyRef : ClassTag](sparkSession: SparkSession, path: String)
   extends MetadataLog[T] with Logging {
 
+  private implicit val formats = Serialization.formats(NoTypeHints)
+
+  /** Needed to serialize type T into JSON */
+  private implicit val manifest = Manifest.classType[T](implicitly[ClassTag[T]].runtimeClass)
```
tdas: Why is this needed now? This wasnt needed earlier when we were using json serialization in FilestreamSinkLog, then why is the manifest needed now?

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/Offset.scala
```diff
@@ -23,4 +23,16 @@ package org.apache.spark.sql.execution.streaming
...
+  /**
+   * A JSON-serialized representation of an Offset that is
+   * used for saving offsets to the offset log.
+   *
+   * @return JSON string encoding
+   */
+  def json: String
+}
+
+/** Used when loading */
```
tdas: Needs better docs. I think i had commented that earlier.

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/OffsetSeqLog.scala
```diff
@@ -0,0 +1,62 @@
...
+    val version = lines.next()
+    if (version != offsetLogVersion) {
+      throw new IllegalStateException(s"Unknown log version: ${version}")
+    }
+    OffsetSeq.fill(lines.map(offset => SerializedOffset(offset)).toArray: _*)
+  }
+
+  override protected def serialize(metadata: OffsetSeq, out: OutputStream): Unit = {
+    // called inside a try-finally where the underlying stream is closed in the caller
+
```
tdas: nit: extra line

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/OffsetSeqLog.scala
```diff
@@ -0,0 +1,62 @@
...
+import org.json4s.jackson.Serialization
+
+import org.apache.spark.sql.SparkSession
+
+class OffsetSeqLog(offsetLogVersion: String, sparkSession: SparkSession, path: String)
+  extends HDFSMetadataLog[OffsetSeq](sparkSession, path) {
+
+  override protected def deserialize(in: InputStream): OffsetSeq = {
+    // called inside a try-finally where the underlying stream is closed in the caller
+
```
tdas: nit: extra line.

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamExecution.scala
```diff
@@ -713,6 +713,8 @@ class StreamExecution(
 }
 
 object StreamExecution {
+  val OFFSET_LOG_VERSION = "v1"
```
tdas: Why is this in StreamExecution, and not in OffsetSeqLog object?

sql/core/src/main/scala/org/apache/spark/sql/streaming/StreamingQueryException.scala
```diff
@@ -36,8 +36,8 @@ class StreamingQueryException private[sql](
     @transient val query: StreamingQuery,
     val message: String,
     val cause: Throwable,
-    val startOffset: Option[Offset] = None,
-    val endOffset: Option[Offset] = None)
+    val startOffset: Option[OffsetSeq] = None,
```
tdas: This is changing public API. We should think this over. 
@marmbrus

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/HDFSMetadataLog.scala
tdas: I got it why this is needed.

sql/core/src/main/scala/org/apache/spark/sql/streaming/StreamingQueryException.scala
tdas: Actually, we made a mistake in 2.0. We dont expose Offset in public API but this exposes Offset. This is wrong

Idea: is we replace startOffset with `startOffsetsDesc: String` (similar to `offsetDesc` in `SourceStatus`)

sql/core/src/test/scala/org/apache/spark/sql/execution/streaming/OffsetSeqLogSuite.scala
```diff
@@ -0,0 +1,65 @@
...
+import org.json4s.jackson.Serialization
+
+import org.apache.spark.SparkFunSuite
+import org.apache.spark.sql.test.SharedSQLContext
+
+class OffsetSeqLogSuite extends SparkFunSuite with SharedSQLContext {
+
+  private implicit val formats = Serialization.formats(NoTypeHints)
+
+  testWithUninterruptibleThread("OffsetSeqLogSuite: basic") {
```
tdas: Since there is only one test "OffsetSeqLogSuite:" in the name is not needed. Rather lets name it "serialization - deserialization" as that is what this is testing.

sql/core/src/test/scala/org/apache/spark/sql/execution/streaming/OffsetSeqLogSuite.scala
```diff
@@ -0,0 +1,65 @@
...
+      assert(metadataLog.get(None, Some(1)) ===
+             Array(0 -> batch0Serialized, 1 -> batch1Serialized))
+
+      // Adding the same batch does nothing
+      metadataLog.add(1, OffsetSeq.fill(LongOffset(3)))
+      assert(metadataLog.get(0) === Some(batch0Serialized))
+      assert(metadataLog.get(1) === Some(batch1Serialized))
+      assert(metadataLog.getLatest() === Some(1 -> batch1Serialized))
+      assert(metadataLog.get(None, Some(1)) ===
+             Array(0 -> batch0Serialized, 1 -> batch1Serialized))
```
tdas: incorrect continuation indent, should be 2 not eight.

sql/core/src/test/scala/org/apache/spark/sql/execution/streaming/OffsetSeqLogSuite.scala
```diff
@@ -0,0 +1,65 @@
...
+      assert(metadataLog.add(0, batch0))
+      assert(metadataLog.getLatest() === Some(0 -> batch0Serialized))
+      assert(metadataLog.get(0) === Some(batch0Serialized))
+
+      assert(metadataLog.add(1, batch1))
+      assert(metadataLog.get(0) === Some(batch0Serialized))
+      assert(metadataLog.get(1) === Some(batch1Serialized))
+      assert(metadataLog.getLatest() === Some(1 -> batch1Serialized))
+      assert(metadataLog.get(None, Some(1)) ===
+             Array(0 -> batch0Serialized, 1 -> batch1Serialized))
```
tdas: incorrect continuation indent, should be 2 not eight.

sql/core/src/test/scala/org/apache/spark/sql/execution/streaming/OffsetSeqLogSuite.scala
```diff
@@ -0,0 +1,65 @@
...
+    withTempDir { temp =>
+      val dir = new File(temp, "dir") // use non-existent directory to test whether log make the dir
+      val metadataLog = new OffsetSeqLog("v1", spark, dir.getAbsolutePath)
+      val batch0 = OffsetSeq.fill(LongOffset(0))
+      val batch1 = OffsetSeq.fill(LongOffset(1), LongOffset(2))
+
+      val batch0Serialized = OffsetSeq.fill(batch0.offsets.map(_.map(o =>
+        SerializedOffset(o.json))).flatten: _*)
+
+      val batch1Serialized = OffsetSeq.fill(batch1.offsets.map(_.map(o =>
```
tdas: simpler to read if you use `flatmap` instead of map...map...flatten.

sql/core/src/test/scala/org/apache/spark/sql/execution/streaming/OffsetSeqLogSuite.scala
```diff
@@ -0,0 +1,65 @@
...
+class OffsetSeqLogSuite extends SparkFunSuite with SharedSQLContext {
+
+  private implicit val formats = Serialization.formats(NoTypeHints)
+
+  testWithUninterruptibleThread("OffsetSeqLogSuite: basic") {
+    withTempDir { temp =>
+      val dir = new File(temp, "dir") // use non-existent directory to test whether log make the dir
+      val metadataLog = new OffsetSeqLog("v1", spark, dir.getAbsolutePath)
+      val batch0 = OffsetSeq.fill(LongOffset(0))
+      val batch1 = OffsetSeq.fill(LongOffset(1), LongOffset(2))
```
tdas: would like to see this test with mixed offset types. how about define a custom offset and use that.

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/HDFSMetadataLog.scala
tdas: Can you added "Needed by Jackson to serialize/deserialize... "

external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/KafkaSourceOffset.scala
```diff
@@ -51,4 +54,11 @@ private[kafka010] object KafkaSourceOffset {
   def apply(offsetTuples: (String, Int, Long)*): KafkaSourceOffset = {
     KafkaSourceOffset(offsetTuples.map { case(t, p, o) => (new TopicPartition(t, p), o) }.toMap)
   }
+
+  /**
+   * Returns [[KafkaSourceOffset]] from a JSON [[SerializedOffset]]
+   */
+  def apply(serialized: SerializedOffset): KafkaSourceOffset = {
```
tdas: This does not seem to used any where. I think the code in KafkaSourceOffset.getPartitionOffsets needs to change because it does not take into account that the param offset can be a SerializedOffset.

tdas (changes requested):
This is look good, pending some public API issues. In addition, can you also add to the scala docs of Source.getBatch that the user should not assume start and end offset to a particular specific type, and the safe way read the offset details is to use the Offset.json to parse the details.

external/kafka-0-10-sql/src/test/scala/org/apache/spark/sql/kafka010/KafkaSourceOffsetSuite.scala
```diff
@@ -36,4 +37,9 @@ class KafkaSourceOffsetSuite extends OffsetSuite {
   compare(
     one = KafkaSourceOffset(("t", 0, 1L)),
     two = KafkaSourceOffset(("t", 0, 2L), ("t", 1, 1L)))
+
+  compare(
```
tcondie: I have extended the KafkaSourceOffsetSuite to include more thorough evaluations of json serialization/deserialization, including to/from OffsetSeqLog.

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/CompactibleFileStreamLog.scala
tcondie: It's required by json4s that T be an AnyRef.

sql/core/src/test/scala/org/apache/spark/sql/execution/streaming/OffsetSeqLogSuite.scala
tcondie: I added a StringOffset type. Let me know if you'd like to see something more complex.

sql/core/src/test/scala/org/apache/spark/sql/execution/streaming/OffsetSeqLogSuite.scala
tcondie: Got it.

sql/core/src/test/scala/org/apache/spark/sql/execution/streaming/OffsetSeqLogSuite.scala
tcondie: Got it.

marmbrus:
Thanks for working on this!  Could you include examples of the various logs, since we are committing to this specific JSON.

external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/KafkaSourceOffset.scala
tcondie: Great catch, thanks! I also added a test for this case.

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamExecution.scala
tcondie: I removed this from StreamExecution. I was initially copying the approach in CompactibleStreamLog, but realize that the notion of a version need not be parameterized here. I'm still retaining the notion of a version, which no resides in the OffsetSeqLog companion object, as per your suggestion.

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/OffsetSeqLog.scala
tcondie: right.

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/Offset.scala
tcondie: Added.

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/HDFSMetadataLog.scala
tcondie: Done.

tcondie:
All log are now uniformly marshall entries to JSON. Each log has a particular way of organizing its JSON entries:
HDFSMetadataLog: 
  Each Batch ID is a separate file.
  All log entries within a batch id are serialized in one line.

OffsetSeqLog:
  Each Batch ID is a separate file. 
  All log entries (i.e., Offsets) are separated by a newline character i.e., one offset entry per line.

CompactibleFileStreamLog, FileStreamSinkLog, and FileStreamSourceLog
  No changes to log file format i.e., log entries are separated by a new line character. 
  Log entry -> JSON serialization is now done in CompactibleFileStreamLog directly.

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/Offset.scala
```diff
@@ -23,4 +23,36 @@ package org.apache.spark.sql.execution.streaming
  * ordering of two [[Offset]] instances.  We do assume that if two offsets are `equal` then no
  * new data has arrived.
  */
-trait Offset extends Serializable {}
+abstract class Offset {
```
zsxwing: nit: could you document that we only compare the json string, please?

zsxwing: We should also document in `Source.getBatch` that `start` and `end` may be semantics same but has different json representations and the Source should handle this case.

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/HDFSMetadataLog.scala
```diff
@@ -88,14 +92,13 @@ class HDFSMetadataLog[T: ClassTag](sparkSession: SparkSession, path: String)
...
-    val outStream = serializer.serializeStream(out)
-    outStream.writeObject(metadata)
+    Serialization.write(metadata, out)
   }
 
   protected def deserialize(in: InputStream): T = {
     // called inside a try-finally where the underlying stream is closed in the caller
-    val inStream = serializer.deserializeStream(in)
-    inStream.readObject[T]()
+    val reader = new InputStreamReader(in)
```
zsxwing: nit: should specify the charset like `new InputStreamReader(in, StandardCharsets.UTF_8)`

zsxwing (changes requested):
Looks pretty good. Just some minor issues.

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/FileStreamSource.scala
```diff
@@ -131,8 +131,8 @@ class FileStreamSource(
    * Returns the data that is between the offsets (`start`, `end`].
    */
   override def getBatch(start: Option[Offset], end: Offset): DataFrame = {
```
tdas: given this change that start and end Offset will not the custom Offset class defined by the source, i think it is important to document this in the Source.getBatch class that the user should not make that assumption.

zsxwing:
retest this please

zsxwing:
LGTM. Pending tests.

zsxwing:

PR #15628, opened 2016-10-25
[SPARK-17471][ML] Add compressed method to ML matrices

sethah:

This patch adds a `compressed` method to ML `Matrix` class, which returns the minimal storage representation of the matrix - either sparse or dense. Because the space occupied by a sparse matrix is dependent upon its layout (i.e. column major or row major), this method must consider both cases. It may also be useful to force the layout to be column or row major beforehand, so an overload is added which takes in a `columnMajor: Boolean` parameter.

The compressed implementation relies upon two new abstract methods `toDense(columnMajor: Boolean)` and `toSparse(columnMajor: Boolean)`, similar to the compressed method implemented in the `Vector` class. These methods also allow the layout of the resulting matrix to be specified via the `columnMajor` parameter. More detail on the new methods is given below.
## How was this patch tested?

Added many new unit tests
## New methods (summary, not exhaustive list)

**Matrix trait**
- `private[ml] def toDenseMatrix(columnMajor: Boolean): DenseMatrix` (abstract) - converts the matrix (either sparse or dense) to dense format
- `private[ml] def toSparseMatrix(columnMajor: Boolean): SparseMatrix` (abstract) -  converts the matrix (either sparse or dense) to sparse format
- `def toDense: DenseMatrix = toDense(true)`  - converts the matrix (either sparse or dense) to dense format in column major layout
- `def toSparse: SparseMatrix = toSparse(true)` -  converts the matrix (either sparse or dense) to sparse format in column major layout
- `def compressed: Matrix` - finds the minimum space representation of this matrix, considering both column and row major layouts, and converts it
- `def compressed(columnMajor: Boolean): Matrix` - finds the minimum space representation of this matrix considering only column OR row major, and converts it

**DenseMatrix class**
- `private[ml] def toDenseMatrix(columnMajor: Boolean): DenseMatrix` - converts the dense matrix to a dense matrix, optionally changing the layout (data is NOT duplicated if the layouts are the same)
- `private[ml] def toSparseMatrix(columnMajor: Boolean): SparseMatrix` - converts the dense matrix to sparse matrix, using the specified layout

**SparseMatrix class**
- `private[ml] def toDenseMatrix(columnMajor: Boolean): DenseMatrix` - converts the sparse matrix to a dense matrix, using the specified layout
- `private[ml] def toSparseMatrix(columnMajors: Boolean): SparseMatrix` - converts the sparse matrix to sparse matrix. If the sparse matrix contains any explicit zeros, they are removed. If the layout requested does not match the current layout, data is copied to a new representation. If the layouts match and no explicit zeros exist, the current matrix is returned.


sethah:
cc @dbtsai @yanboliang 

**Note for reviewers**: The driving need behind this change is that we now have multiclass logistic regression which represents coefficients as a matrix. When we use L1 regularization it can be more efficient to store the coefficient matrix as sparse since there can be a lot of zero elements. We will incorporate this compression in a follow-up PR

sethah:
I'm not sure I understand why there are MiMa failures. It should not be a problem that the new methods only exist in the current version AFAIK.

mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala
```diff
@@ -153,6 +153,86 @@ sealed trait Matrix extends Serializable {
...
+  @Since("2.1.0")
+  def toSparse: SparseMatrix = toSparseMatrix(columnMajor = true)
+
+  /**
+   * Converts this matrix to a dense matrix.
+   *
+   * @param columnMajor Whether the values of the resulting dense matrix should be in column major
+   *                    or row major order. If `false`, resulting matrix will be row major.
+   */
+  private [ml] def toDenseMatrix(columnMajor: Boolean): DenseMatrix
```
sethah: minor: space

mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala
```diff
@@ -153,6 +153,86 @@ sealed trait Matrix extends Serializable {
...
   @Since("2.0.0")
   def numActives: Int
+
+  /**
+   * Converts this matrix to a sparse matrix.
+   *
+   * @param columnMajor Whether the values of the resulting sparse matrix should be in column major
+   *                    or row major order. If `false`, resulting matrix will be row major.
+   */
+  private[ml] def toSparseMatrix(columnMajor: Boolean): SparseMatrix
```
sethah: A bit of explanation: I made this a private method with a different name because `toDense: DenseMatrix` and `toSparse: SparseMatrix` should be implemented in the trait, not in the subclasses. But we can't just put them here and use overloading, because we will get ambiguous reference compile errors. So, we implement them here and make this private with a different name to avoid this. I appreciate feedback on this approach - it feels a bit awkward.

WeichenXu123:
@sethah I think this is a good idea, I am going to add the same method to Vector, since high dimension data training on LIR/LOR with L1 reg, compressed coeffs storage will make the algo broadcast coeffs vector faster, how do you think about it ?

sethah:
What are you going to add? `compressed` already exists on vectors...

WeichenXu123:
@sethah Oh..I miss it..

mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala
```diff
@@ -1076,4 +1240,15 @@ object Matrices {
       SparseMatrix.fromCOO(numRows, numCols, entries)
     }
   }
+
+  private[ml] def getSparseSize(numActives: Long, numPtrs: Long): Long = {
+    // 8 * values.length + 4 * rowIndices.length + 4 * colPtrs.length + 8 + 8 + 1
+    12L * numActives + 4L * numPtrs + 17L
```
MLnick: The comment says 8 * values while this is 12? Seems like a mistype?

sethah: I wondered how confusing this comment might be. Since `values.length == rowIndices.length == numActives`:

`8 * values.length + 4 * rowIndices.length = 8 * numActives + 4 * numActives = 12 * numActives`

The comment is meant to show where each number comes from and the implementation is meant to just be a condensed computation. But please let me know if you think it's too confusing.

MLnick: ah right - no that's fine.

sethah:
ping @dbtsai :)

sethah:
pinging other potential reviewers @jkbradley @srowen @MLnick I think this is an important patch for multiclass logistic regression.

sethah:
ping @imatiach-msft @dbtsai

sethah:
re-ping @dbtsai @MLnick @yanboliang I still think this is an important patch :D

mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala
sethah: Should we just leave `toDenseMatrix` private and have `toDense` always use colMajor = true? I think that's ok to do for now.

mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala
```diff
@@ -153,6 +153,86 @@ sealed trait Matrix extends Serializable {
    */
```
sethah: I made the change. Not sure if we should do this in a separate PR though.

mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala
```diff
@@ -153,6 +154,97 @@ sealed trait Matrix extends Serializable {
...
+    }
+  }
+
+  /**
+   * Returns a matrix in dense column major, dense row major, sparse row major, or sparse column
+   * major format, whichever uses less storage. When dense representation is optimal, it maintains
+   * the current layout order.
+   */
+  @Since("2.2.0")
+  def compressed: Matrix = {
```
sethah: It's only a problem if we override/implement it in a subclass. Since it's contained wholly in the trait, it will be fine. I think this is ok to leave, though we could make it final? Also we could make three methods: `compressed`, `compressedCSC`, `compressedCSR`. I think the latter is a good solution, thoughts?

mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala
```diff
@@ -153,6 +154,97 @@ sealed trait Matrix extends Serializable {
...
   @Since("2.0.0")
   def numActives: Int
+
+  /**
+   * Converts this matrix to a sparse matrix.
+   *
+   * @param colMajor Whether the values of the resulting sparse matrix should be in column major
+   *                    or row major order. If `false`, resulting matrix will be row major.
+   */
+  private[ml] def toSparseMatrix(colMajor: Boolean): SparseMatrix
```
sethah: You mean `@inline private[ml] def ...` ? Do we expect this to be called often enough for that to make a difference?

mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala
```diff
@@ -153,6 +154,97 @@ sealed trait Matrix extends Serializable {
...
+   *                    or row major order. If `false`, resulting matrix will be row major.
+   */
+  private[ml] def toDenseMatrix(colMajor: Boolean): DenseMatrix
+
+  /**
+   * Converts this matrix to a dense matrix in column major order.
+   */
+  @Since("2.2.0")
+  def toDense: DenseMatrix = toDenseMatrix(colMajor = true)
+
```
sethah: Added.

sethah:
@dbtsai Let me know your thoughts on the comments I left. Thanks for the review!

sethah:
Jenkins test this please.

sethah:
ping @dbtsai! Code freeze is upcoming :)

mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala
dbtsai: Fair enough.

mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala
dbtsai: +1 on the later one.

mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala
```diff
@@ -161,6 +162,109 @@ sealed trait Matrix extends Serializable {
...
+   * @param colMajor Whether the values of the resulting sparse matrix should be in column major
+   *                    or row major order. If `false`, resulting matrix will be row major.
+   */
+  private[ml] def toSparseMatrix(colMajor: Boolean): SparseMatrix
+
+  /**
+   * Converts this matrix to a sparse matrix in column major order.
+   */
+  @Since("2.2.0")
+  def toCSC: SparseMatrix = toSparseMatrix(colMajor = true)
```
dbtsai: I'm not good at naming, but since we use `toDenseRowMajor` for dense vector, should we use `toSparseColumnMajor`? Almost many packages are using `toCSC`, but I think we can make them consistent. Just my 2 cents.

mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala
```diff
@@ -161,6 +162,109 @@ sealed trait Matrix extends Serializable {
...
+
+  /**
+   * Converts this matrix to a sparse matrix in row major order.
+   */
+  @Since("2.2.0")
+  def toCSR: SparseMatrix = toSparseMatrix(colMajor = false)
```
dbtsai: Same question.

mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala
```diff
@@ -161,6 +162,109 @@ sealed trait Matrix extends Serializable {
...
+  def toDenseColMajor: DenseMatrix = toDenseMatrix(colMajor = true)
+
+  /**
+   * Returns a matrix in either dense or sparse format, whichever uses less storage.
+   *
+   * @param colMajor Whether the values of the resulting matrix should be in column major
+   *                    or row major order. If `false`, resulting matrix will be row major.
+   */
+  @Since("2.2.0")
+  def compressed(colMajor: Boolean): Matrix = {
```
dbtsai: Let's make it private, and follow the previous style. Should add `compressedRowMajor`, `compressedColumnMajor` since it can be dense matrix in certain situations.

mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala
```diff
@@ -291,31 +395,60 @@ class DenseMatrix @Since("2.0.0") (
...
+  }
+
+  /**
+   * Generate a `DenseMatrix` from this `DenseMatrix`.
+   *
+   * @param colMajor Whether the resulting `DenseMatrix` values will be in column major order.
+   */
+  private[ml] override def toDenseMatrix(colMajor: Boolean): DenseMatrix = {
+    if (!(isTransposed ^ colMajor)) {
+      val newValues = new Array[Double](numCols * numRows)
```
dbtsai: Simpler to use `foreachActive`? With it, both `toDenseMatrix` can have the same implementation for sparse and dense in trait.

mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala
```diff
@@ -587,18 +722,69 @@ class SparseMatrix @Since("2.0.0") (
...
+   *
+   * @param colMajor Whether or not the resulting `SparseMatrix` values are in column major
+   *                    order.
    */
-  @Since("2.0.0")
-  def toDense: DenseMatrix = {
-    new DenseMatrix(numRows, numCols, toArray)
+  private[ml] override def toSparseMatrix(colMajor: Boolean): SparseMatrix = {
+    if (!(colMajor ^ isTransposed)) {
+      // breeze transpose rearranges values in column major and removes explicit zeros
```
dbtsai: I think it's hacky to use breeze's transpose behavior to remove zeros in sparse matrices. Can we have our own implementation given we're potentially remove breeze?

mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala
```diff
@@ -587,18 +722,69 @@ class SparseMatrix @Since("2.0.0") (
...
+      if (!isTransposed) {
+        // it is row major and we want col major
+        val breezeTransposed = asBreeze.asInstanceOf[BSM[Double]].t
+        Matrices.fromBreeze(breezeTransposed).transpose.asInstanceOf[SparseMatrix]
+      } else {
+        // it is col major and we want row major
+        val breezeTransposed = asBreeze.asInstanceOf[BSM[Double]]
+        Matrices.fromBreeze(breezeTransposed).asInstanceOf[SparseMatrix]
+      }
+    } else {
```
dbtsai: Can we document here that it's when the layout of this and colMajor is different? Easier read than `(colMajor ^ isTranspose)` condition here. Even more readable to use pattern matching with exact boolean on both variables.

mllib-local/src/test/scala/org/apache/spark/ml/linalg/MatricesSuite.scala
```diff
@@ -160,22 +160,385 @@ class MatricesSuite extends SparkMLFunSuite {
...
+     */
+    val dm1 = new DenseMatrix(2, 3, Array(4.0, -1.0, 2.0, 7.0, -8.0, 4.0))
+    val dm2 = new DenseMatrix(2, 3, Array(5.0, -9.0, 4.0, 1.0, -3.0, -8.0), isTransposed = true)
+
+    val dm3 = dm1.toDense
+    assert(dm3 === dm1)
+    assert(!dm3.isTransposed)
+    assert(dm3.values.equals(dm1.values))
+
+    val dm4 = dm1.toDenseMatrix(false)
```
dbtsai: I would like to make `toDenseMatrix` as private, and we test against `toDenseRowMajor` which is more explicit.

mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala
```diff
@@ -161,6 +162,110 @@ sealed trait Matrix extends Serializable {
...
+   * Converts this matrix to a sparse matrix in row major order.
+   */
+  @Since("2.2.0")
+  def toCSR: SparseMatrix = toSparseMatrix(colMajor = false)
+
+  /**
+   * Converts this matrix to a sparse matrix in column major order.
+   */
+  @Since("2.2.0")
+  def toSparse: SparseMatrix = toSparseMatrix(colMajor = true)
```
dbtsai: I'm debating that should we keep the same ordering of layout when we call `toSparse` or `toDense`?

mllib-local/src/test/scala/org/apache/spark/ml/linalg/MatricesSuite.scala
```diff
@@ -160,22 +160,385 @@ class MatricesSuite extends SparkMLFunSuite {
...
+  test("dense to dense") {
+    /*
+      dm1 =  4.0 2.0 -8.0
+            -1.0 7.0  4.0
+
+      dm2 = 5.0 -9.0  4.0
+            1.0 -3.0 -8.0
+     */
+    val dm1 = new DenseMatrix(2, 3, Array(4.0, -1.0, 2.0, 7.0, -8.0, 4.0))
+    val dm2 = new DenseMatrix(2, 3, Array(5.0, -9.0, 4.0, 1.0, -3.0, -8.0), isTransposed = true)
```
dbtsai: Why not just make `dm2` dm1.transposed, but explicitly assign the value?  Thus, you don't need to type the value in the array for the comparison.

mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala
```diff
@@ -291,31 +396,60 @@ class DenseMatrix @Since("2.0.0") (
...
+
+  /**
+   * Generate a `DenseMatrix` from this `DenseMatrix`.
+   *
+   * @param colMajor Whether the resulting `DenseMatrix` values will be in column major order.
+   */
+  private[ml] override def toDenseMatrix(colMajor: Boolean): DenseMatrix = {
+    if (!(isTransposed ^ colMajor)) {
+      val newValues = new Array[Double](numCols * numRows)
+      var j = 0
```
kiszk: Can we move `if (isTransposed) {` out of a loop like the following since it is a loop invariant and we want to remove div/mod operations?
```
if (isTransposed) {
  // it is row major and we want column major
  var j = 0
  var col = 0
  while (col < numCols) {
    var row = 0
    while (row < numRows) {
      ...
    }
    col += 1
  }
} else {
  // it is column major and we want row major
  var j = 0
  var row = 0
  while (row < numRows) {
    var col = 0
    while (col < numCols) {
      ...
    }
    rows += 1
  }
}
```

mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala
sethah: I'm not sure I have a preference. I don't mind leaving them as CSC and CSR.

mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala
sethah: I don't think we can put this in the trait, since when it is called on dense matrix we would like to return `this` in some cases (when no layout change is needed). But, yes, I think it simpler to use `toArray` which calls `foreachActive`. Thanks!

mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala
dbtsai: BTW, it's nice to have return type in public method. Can you add `Unit` as return type?

mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala
dbtsai: The following could work, and we only need one implementation in trait. Thanks.

```scala
trait Matrix {

  var isTransposed: Boolean = true
  var numCols: Int = 0
  var numRows: Int = 0

  def foreachActive(f: (Int, Int, Double) => Unit): Unit

  def toDenseMatrix(colMajor: Boolean): Matrix = {
    this match {
      case _: DenseMatrix if this.isTransposed != colMajor =>
        this
      case _: SparseMatrix | _: DenseMatrix if this.isTransposed == colMajor =>
        val newValues = new Array[Double](numCols * numRows)

        this.foreachActive { case (row, col, value) =>
            // filling the newValues
        }
        new DenseMatrix(numRows, numCols, newValues, isTransposed = !colMajor)
      case _ =>
        throw new IllegalArgumentException("")
    }
  }
}

class DenseMatrix extends Matrix {

  def foreachActive(f: (Int, Int, Double) => Unit): Unit = {

  }

}

class SparseMatrix extends Matrix {
  def foreachActive(f: (Int, Int, Double) => Unit): Unit = {

  }
}
```

mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala
dbtsai: After thinking about it again, let's have it as `toSparseColumnMajor` to make the apis consistent with the dense ones if you don't mind?

mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala
sethah: Hm, I don't think this solution is better. The entire point of abstract methods is to allow subclasses to implement a method differently. Since we need different implementations depending on the subclass, we should just implement them in the subclasses. We can do this with the following:

````scala
trait Matrix {
  def toDenseMatrix(colMajor: Boolean): Matrix
}
class DenseMatrix extends Matrix {
    private[ml] override def toDenseMatrix(colMajor: Boolean): DenseMatrix = {
    if (isTransposed && colMajor) {
      new DenseMatrix(numRows, numCols, toArray, isTransposed = false)
    } else if (!isTransposed && !colMajor) {
      new DenseMatrix(numRows, numCols, transpose.toArray, isTransposed = true)
    } else {
      this
    }
  }
}
class SparseMatrix extends Matrix {
  private[ml] override def toDenseMatrix(colMajor: Boolean): DenseMatrix = {
    if (colMajor) new DenseMatrix(numRows, numCols, toArray)
    else new DenseMatrix(numRows, numCols, this.transpose.toArray, isTransposed = true)
  }
}
````

Which is less verbose than the previous code. I'm going to put that in the next commit. Let me know what you think.

File: mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala (hunk 12 above)
Comment (sethah) [inline mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala]: I pretty much agree with you, but this is non-trivial code if we want to do it efficiently. Breeze has a pretty well-optimized implementation to do this. I would leave it as a follow up JIRA, or do it when/if we ever remove the Breeze dependency. Or do you think this is a blocker for this PR?

File: mllib-local/src/test/scala/org/apache/spark/ml/linalg/MatricesSuite.scala (hunk 16 above)
Comment (sethah) [inline mllib-local/src/test/scala/org/apache/spark/ml/linalg/MatricesSuite.scala]: I'm not sure I understand your meaning here. These are made to be two entirely different matrices anyway.

File: mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala (hunk 15 above)
Comment (sethah) [inline mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala]: Well, this would change the behavior of external facing code. Before if I called `toSparse` on a row major matrix, I'd get a column major matrix. If we maintain the layout, then I'd now get something different (column major). Otherwise, I'd agree it is best to maintain the layout.

File: mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala (hunk 17 above)
Comment (sethah) [inline mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala]: See above discussion, I am going to change this to use `toArray`.

File: mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala (hunk 12 above)
Comment (dbtsai) [inline mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala]: This is not a blocker.

File: mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala (hunk 11 above)
Comment (dbtsai) [inline mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala]: This looks great to me!

File: mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala (hunk 17 above)
Comment (kiszk) [inline mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala]: It is fine with me since `isTransposed` is checked beforehand.

Comment (kiszk) [inline mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala]: I have one question regarding performance.
To use [`toArray`](https://github.com/sethah/spark/blob/4746ec0d97c002241be344494a6d2ddee3a7c2d5/mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala#L49-L55) introduce allocation of a temporary array and data copy.
Can we avoid this allocation and copy by passing an original array and access function to `new DenseMatrix`?

Comment (sethah) [inline mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala]: I don't see how there is an unnecessary copy. `toArray` copies the elements of the current matrix to a new Array, then uses that as the backing array of a new `DenseMatrix`. We cannot modify the original matrix values.

**Comment by sethah** [comment]
Jenkins retest this please.

**Comment by sethah** [comment]
Jenkins retest this please.

File: mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala (hunk 15 above)
Comment (dbtsai) [inline mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala]: But I thought this is a new api being added, so we can make it to maintain the same layout.

File: mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala [hunk 18]
```diff
@@ -161,6 +162,118 @@ sealed trait Matrix extends Serializable {
...
+  @Since("2.2.0")
+  def compressed: Matrix = {
+    val cscSize = getSparseSizeInBytes(colMajor = true)
+    val csrSize = getSparseSizeInBytes(colMajor = false)
+    if (getDenseSizeInBytes < math.min(cscSize, csrSize)) {
+      // dense matrix size is the same for column major and row major, so maintain current layout
+      toDenseMatrix(!isTransposed)
+    } else {
+      if (cscSize <= csrSize) {
+        toSparseMatrix(colMajor = true)
```
dbtsai: `toSparseColumnMajor`

mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala
```diff
@@ -161,6 +162,118 @@ sealed trait Matrix extends Serializable {
...
+      } else {
+        toSparseMatrix(colMajor = false)
```
dbtsai: `toSparseRowMajor`

mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala
```diff
@@ -161,6 +162,118 @@ sealed trait Matrix extends Serializable {
...
+   * major format, whichever uses less storage. When dense representation is optimal, it maintains
+   * the current layout order.
+   */
+  @Since("2.2.0")
+  def compressed: Matrix = {
+    val cscSize = getSparseSizeInBytes(colMajor = true)
+    val csrSize = getSparseSizeInBytes(colMajor = false)
+    if (getDenseSizeInBytes < math.min(cscSize, csrSize)) {
+      // dense matrix size is the same for column major and row major, so maintain current layout
+      toDenseMatrix(!isTransposed)
```
dbtsai: Call `toDense` if we decide to make `toDense` and `toSparse` outputting the same layout.

mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala
```diff
@@ -291,31 +404,49 @@ class DenseMatrix @Since("2.0.0") (
...
-      colPtrs(j) = nnz
+      new SparseMatrix(numRows, numCols, colPtrs, rowIndices.result(), spVals.result())
+    }
+  }
+
+  /**
+   * Generate a `DenseMatrix` from this `DenseMatrix`.
+   *
+   * @param colMajor Whether the resulting `DenseMatrix` values will be in column major order.
+   */
```
dbtsai: Minor, can we have 
```scala
protected def isColMajor = !isTransposed

protected def isRowMajor = isTransposed
```

So the code can be understood easier?

mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala
```diff
@@ -291,31 +404,49 @@ class DenseMatrix @Since("2.0.0") (
...
+  private[ml] override def toDenseMatrix(colMajor: Boolean): DenseMatrix = {
+    if (isTransposed && colMajor) {
+      new DenseMatrix(numRows, numCols, toArray, isTransposed = false)
+    } else if (!isTransposed && !colMajor) {
+      new DenseMatrix(numRows, numCols, transpose.toArray, isTransposed = true)
+    } else {
+      this
     }
-    new SparseMatrix(numRows, numCols, colPtrs, rowIndices.result(), spVals.result())
   }
```
dbtsai: Could we override the `toArray` in DenseMatrix so when `this` is column major, we just return `this.values`? Otherwise, it's very expensive to create a new array.

mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala
sethah: Done

mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala
sethah: Done

mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala
sethah: Done

mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala
sethah: I have to say I'm a bit surprised - I thought that's what `toArray` already did! Yeah, that seems like a good change, but I'd prefer to do it in another pr because we need to make sure that this doesn't adversely affect other places that use `toArray` as well as adding unit tests. If that sounds ok, I'll make a JIRA for it.

mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala
sethah: Ok, I added these methods. I updated the test suites to use them instead of `isTransposed`.

mllib-local/src/test/scala/org/apache/spark/ml/linalg/MatricesSuite.scala
```diff
@@ -160,22 +160,395 @@ class MatricesSuite extends SparkMLFunSuite {
...
+
+    val dm3 = dm1.toDense
+    assert(dm3 === dm1)
+    assert(!dm3.isTransposed)
+    assert(dm3.values.equals(dm1.values))
```
dbtsai: `val dm4 = dm1.toDenseRowMajor` and `val dm7 = dm1.toDenseRowMajor` are the same.

mllib-local/src/test/scala/org/apache/spark/ml/linalg/MatricesSuite.scala
```diff
@@ -160,22 +160,395 @@ class MatricesSuite extends SparkMLFunSuite {
...
+    val sm1 = new SparseMatrix(2, 3, Array(0, 0, 2, 3), Array(0, 1, 0), Array(4.0, 2.0, 5.0))
+    val sm2 = new SparseMatrix(2, 3, Array(0, 2, 3), Array(1, 2, 1), Array(4.0, 5.0, 2.0),
+      isTransposed = true)
+    val sm3 = new SparseMatrix(2, 3, Array(0, 0, 2, 4), Array(0, 1, 0, 1),
+      Array(4.0, 2.0, 5.0, 0.0))
+    val sm4 = new SparseMatrix(2, 3, Array(0, 2, 4), Array(1, 2, 1, 2),
+      Array(4.0, 5.0, 2.0, 0.0), isTransposed = true)
+    val smZeros = new SparseMatrix(2, 3, Array(0, 2, 4, 6), Array(0, 1, 0, 1, 0, 1),
+      Array(0.0, 0.0, 0.0, 0.0, 0.0, 0.0))
+
```
dbtsai: Group the tests by methods

mllib-local/src/test/scala/org/apache/spark/ml/linalg/MatricesSuite.scala
```diff
@@ -160,22 +160,395 @@ class MatricesSuite extends SparkMLFunSuite {
...
+    assert(dm5 === sm3)
+    assert(!dm5.isTransposed)
+    assert(dm5.values === Array.fill(6)(0.0))
+
+    val dm6 = sm2.toDenseColMajor
+    assert(dm6 === sm2)
+    assert(!dm6.isTransposed)
+    assert(dm6.values === Array(0.0, 0.0, 4.0, 2.0, 5.0, 0.0))
+
+    val dm7 = sm2.toDenseRowMajor
```
dbtsai: This is duplicated to `val dm4 = sm2.toDenseRowMajor`. And some of them don't have test of `toDenseColMajor`

mllib-local/src/test/scala/org/apache/spark/ml/linalg/MatricesSuite.scala
```diff
@@ -160,22 +160,395 @@ class MatricesSuite extends SparkMLFunSuite {
...
+            0.0 0.0 0.0 0.0
+     */
+    // this should compress to a sparse matrix
+    val dm1 = new DenseMatrix(3, 4, Array.fill(2)(1.0) ++ Array.fill(10)(0.0))
+
+    // optimal compression layout is row major since numRows < numCols
+    val cm1 = dm1.compressed.asInstanceOf[SparseMatrix]
+    assert(cm1 === dm1)
+    assert(cm1.isTransposed)
+    assert(cm1.getSizeInBytes <= dm1.getSizeInBytes)
```
dbtsai: Why not `<` instead of `<=`?

mllib-local/src/test/scala/org/apache/spark/ml/linalg/MatricesSuite.scala
```diff
@@ -160,22 +160,395 @@ class MatricesSuite extends SparkMLFunSuite {
...
+
+    // force compressed column major
+    val cm2 = dm1.compressedColMajor.asInstanceOf[SparseMatrix]
+    assert(cm2 === dm1)
+    assert(!cm2.isTransposed)
+    assert(cm2.getSizeInBytes <= dm1.getSizeInBytes)
```
dbtsai: ditto

dbtsai: Also check compressedRowMajor.

mllib-local/src/test/scala/org/apache/spark/ml/linalg/MatricesSuite.scala
```diff
@@ -160,22 +160,395 @@ class MatricesSuite extends SparkMLFunSuite {
...
+
+    // optimal compression layout for transpose is column major
+    val dm2 = dm1.transpose
+    val cm3 = dm2.compressed.asInstanceOf[SparseMatrix]
+    assert(cm3 === dm2)
+    assert(!cm3.isTransposed)
+    assert(cm3.getSizeInBytes <= dm2.getSizeInBytes)
```
dbtsai: ditto.

mllib-local/src/test/scala/org/apache/spark/ml/linalg/MatricesSuite.scala
```diff
@@ -160,22 +160,395 @@ class MatricesSuite extends SparkMLFunSuite {
...
+     */
+    // this should compress to a dense matrix
+    val dm3 = new DenseMatrix(3, 4, Array.fill(7)(1.0) ++ Array.fill(5)(0.0))
+    val dm4 = new DenseMatrix(3, 4, Array.fill(7)(1.0) ++ Array.fill(5)(0.0), isTransposed = true)
+
+    val cm4 = dm3.compressed.asInstanceOf[DenseMatrix]
+    assert(cm4 === dm3)
+    assert(!cm4.isTransposed)
+    assert(cm4.values.equals(dm3.values))
+    assert(cm4.getSizeInBytes <= dm3.getSizeInBytes)
```
dbtsai: Should be `=` here.

mllib-local/src/test/scala/org/apache/spark/ml/linalg/MatricesSuite.scala
```diff
@@ -160,22 +160,395 @@ class MatricesSuite extends SparkMLFunSuite {
...
+
+    // force compressed row major
+    val cm5 = dm3.compressedRowMajor.asInstanceOf[DenseMatrix]
+    assert(cm5 === dm3)
+    assert(cm5.isTransposed)
+    assert(cm5.getSizeInBytes <= dm3.getSizeInBytes)
```
dbtsai: Should be `=` here.

mllib-local/src/test/scala/org/apache/spark/ml/linalg/MatricesSuite.scala
```diff
@@ -160,22 +160,395 @@ class MatricesSuite extends SparkMLFunSuite {
...
+
+    val cm6 = dm4.compressed.asInstanceOf[DenseMatrix]
+    assert(cm6 === dm4)
+    assert(cm6.isTransposed)
+    assert(cm6.values.equals(dm4.values))
+    assert(cm6.getSizeInBytes <= dm4.getSizeInBytes)
```
dbtsai: ditto.

mllib-local/src/test/scala/org/apache/spark/ml/linalg/MatricesSuite.scala
```diff
@@ -160,22 +160,395 @@ class MatricesSuite extends SparkMLFunSuite {
...
+    // these should compress to sparse matrices
+    val sm1 = new SparseMatrix(4, 2, Array(0, 1, 2), Array(2, 0), Array(-4.0, -1.0))
+    val sm2 = sm1.transpose
+
+    val cm1 = sm1.compressed.asInstanceOf[SparseMatrix]
+    // optimal is column major
+    assert(cm1 === sm1)
+    assert(!cm1.isTransposed)
+    assert(cm1.values.equals(sm1.values))
+    assert(cm1.getSizeInBytes <= sm1.getSizeInBytes)
```
dbtsai: ditto. remove `=`

mllib-local/src/test/scala/org/apache/spark/ml/linalg/MatricesSuite.scala
```diff
@@ -160,22 +160,395 @@ class MatricesSuite extends SparkMLFunSuite {
...
+
+    val cm2 = sm1.compressedRowMajor.asInstanceOf[SparseMatrix]
+    assert(cm2 === sm1)
+    assert(cm2.isTransposed)
+    // forced to be row major, so we have increased the size
+    assert(cm2.getSizeInBytes > sm1.getSizeInBytes)
+    assert(cm2.getSizeInBytes <= sm1.toDense.getSizeInBytes)
```
dbtsai: `=` is needed?

mllib-local/src/test/scala/org/apache/spark/ml/linalg/MatricesSuite.scala
```diff
@@ -160,22 +160,395 @@ class MatricesSuite extends SparkMLFunSuite {
...
+    assert(cm3 === sm2)
+    assert(cm3.isTransposed)
+    assert(cm3.values.equals(sm2.values))
+    assert(cm3.getSizeInBytes <= sm2.getSizeInBytes)
+
+    val cm8 = sm2.compressedColMajor.asInstanceOf[SparseMatrix]
+    assert(cm8 === sm2)
+    assert(!cm8.isTransposed)
+    assert(cm8.getSizeInBytes > sm2.getSizeInBytes)
+    assert(cm8.getSizeInBytes <= sm2.toDense.getSizeInBytes)
```
dbtsai: For completion, could you test `sm1. compressedColumnMajor` and `sm2. compressedRowMajor`? Thanks.

mllib-local/src/test/scala/org/apache/spark/ml/linalg/MatricesSuite.scala
```diff
@@ -160,22 +160,395 @@ class MatricesSuite extends SparkMLFunSuite {
...
+     */
+    // this should compress to a dense matrix
+    val sm3 = new SparseMatrix(3, 2, Array(0, 2, 5), Array(1, 2, 0, 1, 2),
+      Array(2.0, -4.0, -1.0, 3.0, 9.0))
+
+    // dense is optimal, and maintains column major
+    val cm4 = sm3.compressed.asInstanceOf[DenseMatrix]
+    assert(cm4 === sm3)
+    assert(!cm4.isTransposed)
+    assert(cm4.getSizeInBytes <= sm3.getSizeInBytes)
```
dbtsai: remove `=`?

mllib-local/src/test/scala/org/apache/spark/ml/linalg/MatricesSuite.scala
```diff
@@ -160,22 +160,395 @@ class MatricesSuite extends SparkMLFunSuite {
...
+
+    val cm5 = sm3.compressedRowMajor.asInstanceOf[DenseMatrix]
+    assert(cm5 === sm3)
+    assert(cm5.isTransposed)
+    assert(cm5.getSizeInBytes <= sm3.getSizeInBytes)
```
dbtsai: remove `=`? and `sm3.compressedColumnMajor`?

mllib-local/src/test/scala/org/apache/spark/ml/linalg/MatricesSuite.scala
sethah: I think it's ok to leave it here since they could potentially be equal.

sethah:
@dbtsai Thanks for the good suggestions. I rearranged the test suites, removed redundancies, and filled in some gaps. Things got a bit jumbled when changing some of the methods around. I also added the `this` prefix to some of the method calls as you suggest.

mllib-local/src/test/scala/org/apache/spark/ml/linalg/MatricesSuite.scala
dbtsai: If =, dense should be preferred since it will be faster. As a result, we should only check <.

dbtsai (commented):
Minor

mllib-local/src/test/scala/org/apache/spark/ml/linalg/MatricesSuite.scala
sethah: Ok, before we broke ties with sparse, so I changed it to choose dense, and added unit tests. Also removed the = here.

mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala
```diff
@@ -1079,4 +1267,15 @@ object Matrices {
       SparseMatrix.fromCOO(numRows, numCols, entries)
     }
   }
+
+  private[ml] def getSparseSize(numActives: Long, numPtrs: Long): Long = {
+    // 8 * values.length + 4 * rowIndices.length + 4 * colPtrs.length + 12 + 12 + 12 + 1
+    12L * numActives + 4L * numPtrs + 37L
```
dbtsai: `(java.lang.Double.BYTES + java.lang.Integer.BYTES) * numActives + java.lang.Integer.BYTES * numPtrs + 37L`

dbtsai (approved):
LGTM. Couple small comments.

mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala
dbtsai: Nice that we can get `37L` using java apis to ensure the portability.

sethah:
Thanks for finally cooperating, Jenkins!

dbtsai:
Thanks @sethah and Jenkins! Merged into master.

sethah:
Thanks for all the time reviewing!

mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala
kiszk: I see. `DenseMatrix.update()` modifies the contexts of `value`.

mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala
```diff
@@ -148,7 +154,8 @@ sealed trait Matrix extends Serializable {
    *          and column indices respectively with the type `Int`, and the final parameter is the
    *          corresponding value in the matrix with type `Double`.
    */
-  private[spark] def foreachActive(f: (Int, Int, Double) => Unit)
+  @Since("2.2.0")
+  def foreachActive(f: (Int, Int, Double) => Unit): Unit
```
jkbradley: @sethah @dbtsai Hi all, just saw this during QA.  This method is not very Java-friendly.  I'm OK with adding it as long as we document the fact that it's not Java-friendly.  We could also consider adding a Java-friendly version, perhaps using https://spark.apache.org/docs/latest/api/java/org/apache/spark/api/java/function/Function2.html

dbtsai: I think we should do the same for `foreachActive` in Vector which was already a public api long time ago.

jkbradley: That's a good point.  I guess we can leave it until someone complains & add a Java-friendly one as needed.

PR #15630, opened 2016-10-25
[SPARK-18104][DOC]Don't build KafkaSource doc

zsxwing:

Don't need to build doc for KafkaSource because the user should use the data source APIs to use KafkaSource. All KafkaSource APIs are internal.
## How was this patch tested?

Verified manually.


zsxwing:
cc @tdas

zsxwing:
Discussed with @tdas offline. I'm going to merge to master and 2.0.

PR #15632, opened 2016-10-25
[SPARK-18105] fix buffer overflow in LZ4

davies:

In the LZ4BlockInputStream, there is a mistake that it use the originalLen as compressedLen, in the case that the compressed data is longer than original data, then compressedBuffer could overflow (some InputStream may does not check that), which cause it failed to decompress the buffer.

File a issue in uptream: https://github.com/jpountz/lz4-java/issues/89
## How was this patch tested?

Can't generate some dataset to reproduce this bug, will test it manually.

### DIFF
--- a/core/src/main/java/org/apache/spark/io/LZ4BlockInputStream.java
+++ b/core/src/main/java/org/apache/spark/io/LZ4BlockInputStream.java
@@ -197,7 +197,7 @@ public final class LZ4BlockInputStream extends FilterInputStream {
         readFully(buffer, originalLen);
         break;
       case COMPRESSION_METHOD_LZ4:
-        if (compressedBuffer.length < originalLen) {
+        if (compressedBuffer.length < compressedLen) {
           compressedBuffer = new byte[Math.max(compressedLen, compressedBuffer.length * 3 / 2)];
         }
         readFully(compressedBuffer, compressedLen);


mridulm:
Ah, nice catch !
LGTM, can you merge it when build it done ? Thx.

srowen:
Out of curiosity

brkyvz:
@srowen I think it's a matter of how fast upstream publishes a new version, and we can make sure that everything works

tejasapatil:
+1 for @srowen's comment. There are no test cases for this class in Spark. Since LZ4 is used by default, changing such a core class without any test case sounds risky.

mridulm:
@srowen The last update to the upstream library was 1.5+ years ago. Is it still maintained ?
You are right, pushing it there and getting it released it might be good - if possible.

core/src/main/java/org/apache/spark/io/LZ4BlockInputStream.java
```diff
@@ -197,7 +197,7 @@ private void refill() throws IOException {
         readFully(buffer, originalLen);
         break;
       case COMPRESSION_METHOD_LZ4:
-        if (compressedBuffer.length < originalLen) {
+        if (compressedBuffer.length < compressedLen) {
```
viirya: Does this possibly happen?

I go to check https://github.com/jpountz/lz4-java/blob/master/src/java/net/jpountz/lz4/LZ4BlockOutputStream.java#L192

If the compressed lengh is more the original length, it will choose `COMPRESSION_METHOD_RAW` as compress method, instead of `COMPRESSION_METHOD_LZ4`. In other words, the compressed length is never more then original length under `COMPRESSION_METHOD_LZ4`.

davies:
Thanks to @viirya, this is a false alarm, I will continue to looking for the real cause.

@srowen The reason we have a fork here is that the upstream does not support concated stream, we send a patch for it a year ago but that is still open. We have some tests to cover concated stream on Spark side (not a unit test for LZ4).

mridulm:
@davies that check looks wrong - behaviorally we might not be hitting it
due to other preconditions enforced ; worrying to see this

PR #15633, opened 2016-10-26
[SPARK-18087] [SQL] Optimize insert to not require REPAIR TABLE

ericl:

When inserting into datasource tables with partitions managed by the hive metastore, we need to notify the metastore of newly added partitions. Previously this was implemented via `msck repair table`, but this is more expensive than needed.

This optimizes the insertion path to add only the updated partitions.
## How was this patch tested?

Existing tests (I verified manually that tests fail if the repair operation is omitted).

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSourceStrategy.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSourceStrategy.scala
@@ -26,6 +26,7 @@ import org.apache.spark.sql.catalyst.{CatalystConf, CatalystTypeConverters, Inte
 import org.apache.spark.sql.catalyst.CatalystTypeConverters.convertToScala
 import org.apache.spark.sql.catalyst.analysis._
 import org.apache.spark.sql.catalyst.catalog.{CatalogTable, SimpleCatalogRelation}
+import org.apache.spark.sql.catalyst.catalog.CatalogTypes.TablePartitionSpec
 import org.apache.spark.sql.catalyst.expressions
 import org.apache.spark.sql.catalyst.expressions._
 import org.apache.spark.sql.catalyst.planning.PhysicalOperation
@@ -34,7 +35,7 @@ import org.apache.spark.sql.catalyst.plans.logical.{LogicalPlan, Project, Union}
 import org.apache.spark.sql.catalyst.plans.physical.{HashPartitioning, UnknownPartitioning}
 import org.apache.spark.sql.catalyst.rules.Rule
 import org.apache.spark.sql.execution.{RowDataSourceScanExec, SparkPlan}
-import org.apache.spark.sql.execution.command.{AlterTableRecoverPartitionsCommand, DDLUtils, ExecutedCommandExec}
+import org.apache.spark.sql.execution.command.{AlterTableAddPartitionCommand, DDLUtils, ExecutedCommandExec}
 import org.apache.spark.sql.sources._
 import org.apache.spark.sql.types._
 import org.apache.spark.unsafe.types.UTF8String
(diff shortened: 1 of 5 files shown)


ericl:
TODO: uncomment `//          metastoreUpdater.run(t.sparkSession)` once tests are verified to fail.

ericl:
cc @cloud-fan @davies

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/WriteOutput.scala
```diff
@@ -386,13 +390,18 @@ object WriteOutput extends Logging {
           logDebug(s"Writing partition: $currentKey")
 
           currentWriter = newOutputWriter(currentKey, getPartitionString)
+          val partitionStr = getPartitionString(currentKey).getString(0)
```
davies: partitionStr  => partitionPath ?

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/WriteOutput.scala
```diff
@@ -375,6 +378,7 @@ object WriteOutput extends Logging {
 
       // If anything below fails, we should abort the task.
       var currentKey: UnsafeRow = null
+      var updatedPartitions: List[String] = Nil
```
davies: In case of bucketing, There are multiple files (writer) per partition, so partitionPath will have duplicated value, should we use Set[String] here?

ericl: Done

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/WriteOutput.scala
ericl: Done

cloud-fan:
should we also fix `DataFrameWriter.saveAsTable`?

ericl:
I think that one is ok since we have to scan the full table anyways. If it becomes a performance issue we can also add this optimization.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSourceStrategy.scala
```diff
@@ -179,24 +180,30 @@ case class DataSourceAnalysis(conf: CatalystConf) extends Rule[LogicalPlan] {
           "Cannot overwrite a path that is also being read from.")
       }
 
+      def refreshPartitionsCallback(updatedPartitions: Seq[TablePartitionSpec]): Unit = {
+        if (l.catalogTable.isDefined &&
```
cloud-fan: shall we move this if out of the function? e.g.
```
val refreshPartitionsCallback = if (...) {
  ...
} else {
  _ => ()
}
```

ericl: imo that is a little harder to read, since you have two anonymous function declarations instead of one.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSourceStrategy.scala
```diff
@@ -179,24 +180,30 @@ case class DataSourceAnalysis(conf: CatalystConf) extends Rule[LogicalPlan] {
...
+            l.catalogTable.get.partitionColumnNames.nonEmpty &&
+            l.catalogTable.get.partitionProviderIsHive) {
+          val metastoreUpdater = AlterTableAddPartitionCommand(
```
ericl: I'd rather keep it, since the fetch overhead is pretty small

cloud-fan:
LGTM, pending jenkins

rxin:
Merging in master.

rxin:
I took a look at this again tonight, in the context of consolidating this with Hive. I think doing it through callback is actually not ideal, as callbacks are harder to trace. In 2.2 we should make this an explicit action, rather than callbacks.

PR #15636, opened 2016-10-26
[SPARK-18109][ML] Add instrumentation to GMM

zhengruifeng:

Add instrumentation to GMM
## How was this patch tested?

Test in spark-shell

### DIFF
--- a/mllib/src/main/scala/org/apache/spark/ml/clustering/GaussianMixture.scala
+++ b/mllib/src/main/scala/org/apache/spark/ml/clustering/GaussianMixture.scala
@@ -323,6 +323,9 @@ class GaussianMixture @Since("2.0.0") (
       case Row(point: Vector) => OldVectors.fromML(point)
     }
 
+    val instr = Instrumentation.create(this, rdd)
+    instr.logParams(featuresCol, predictionCol, probabilityCol, k, maxIter, seed, tol)
+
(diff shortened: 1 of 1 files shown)


zhengruifeng:
@jkbradley @yanboliang Could you please make a review in your spare time?

mllib/src/main/scala/org/apache/spark/ml/clustering/GaussianMixture.scala
```diff
@@ -337,6 +340,8 @@ class GaussianMixture @Since("2.0.0") (
     val summary = new GaussianMixtureSummary(model.transform(dataset),
       $(predictionCol), $(probabilityCol), $(featuresCol), $(k))
     model.setSummary(summary)
+    instr.logSuccess(model)
```
jkbradley: It would be good to log numFeatures at this point (since it's readily available from the model).

jkbradley:
Other than that, this looks fine.

mllib/src/main/scala/org/apache/spark/ml/clustering/GaussianMixture.scala
zhengruifeng: OK, I will add numFeatures here.

yanboliang:
LGTM, merged into master. Thanks!

PR #15637, opened 2016-10-26
[SPARK-18000] [SQL] Aggregation function for computing bins (distinct value, count) pairs for equi-width histograms

wzhfy:

This function computes the count for each distinct non-null value of a column.
It returns:
1. null if the table is empty or all values of the column are null.
2. (distinct non-null value, frequency) pairs if the number of distinct non-null values is less than or equal to the specified threshold.
3. an empty result if the number of distinct non-null values exceeds that threshold.

## How was this patch tested?

add test cases


rxin:
cc @srinathshankar

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/HistogramEndpoints.scala
```diff
@@ -0,0 +1,465 @@
...
+    val doubleValue = dataType match {
+      case n: NumericType =>
+        n.numeric.toDouble(value.asInstanceOf[n.InternalType])
+      case d: DateType =>
+        value.asInstanceOf[Int].toDouble
+      case t: TimestampType =>
+        value.asInstanceOf[Long].toDouble
+    }
+    // update percentileDigest
+    percentileDigest.add(doubleValue)
```
srinathshankar: I take it you're updating the percentile in all cases because you don;t want to merge maps with percentiledigests ?

srinathshankar (changes requested):
Given that you're always computing a percentile, have you considered just splitting out the "exact map" aggregate from the percentile aggregate ?
That is, define only an 
`MapsAggregate(col, numBins)` that returns an exact map if there are less than numBins distinct values for col, and null otherwise.
Then for string types you do
`select MapsAggregate(stringcol, numbins) from table`
For numerics you do
`select MapsAggregate(numericcol, numbins), approx_quantile(numericcol, array(...)) from table`

wzhfy:
@srinathshankar Yeah, I think your proposal is better. Thanks for the advice! I'll update this pr asap.

wzhfy:
retest this please

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/MapAggregate.scala
```diff
@@ -0,0 +1,324 @@
...
+ * specified maximum number of bins. Otherwise, it returns an empty map.
+ *
+ * @param child child expression that can produce column value with `child.eval(inputRow)`
+ * @param numBinsExpression The maximum number of bins.
+ */
+@ExpressionDescription(
+  usage =
+    """
+      _FUNC_(col, numBins) - Returns bins - (distinct value, frequency) pairs of equi-width
+      histogram when the number of distinct values is less than or equal to the specified
```
gatorsmile: Here, we need to document `distinct non-null values`. This function excludes NULL.

However, I am not sure whether this is expected. NULL is also a value, right?

sql/core/src/test/scala/org/apache/spark/sql/MapAggregateQuerySuite.scala
```diff
@@ -0,0 +1,124 @@
...
+        ints(i).toLong,
+        doubles(i).toFloat,
+        doubles(i),
+        Decimal(doubles(i)),
+        Date.valueOf(dates(i)),
+        Timestamp.valueOf(timestamps(i)),
+        strings(i))
+    }
+    withTempView(table) {
+      data.toDF("c1", "c2", "c3", "c4", "c5", "c6", "c7", "c8", "c9", "c10")
```
gatorsmile: Can you use `def createDataFrame(rowRDD: RDD[Row], schema: StructType)`, instead of using implicit schema inference? 

We need an explicit schema definition when we verifying different data types.

gatorsmile: In this example, below is the printed schema

```
+--------+--------------+-------+
|col_name|     data_type|comment|
+--------+--------------+-------+
|      c1|       tinyint|   null|
|      c2|      smallint|   null|
|      c3|           int|   null|
|      c4|        bigint|   null|
|      c5|         float|   null|
|      c6|        double|   null|
|      c7|decimal(38,18)|   null|
|      c8|          date|   null|
|      c9|     timestamp|   null|
|     c10|        string|   null|
+--------+--------------+-------+
```

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/MapAggregate.scala
```diff
@@ -0,0 +1,324 @@
...
+  }
+
+  override def children: Seq[Expression] = Seq(child, numBinsExpression)
+
+  override def prettyName: String = "map_aggregate"
+}
+
+trait MapDigest {
+  // Mark this MapDigest invalid when the size of the hashmap (ndv of the column) exceeds numBins
+  var invalid: Boolean = false
```
gatorsmile: `invalid` is confusing. How about `isInvalid`?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/MapAggregate.scala
```diff
@@ -0,0 +1,324 @@
...
+    numBinsExpression: Expression,
+    override val mutableAggBufferOffset: Int,
+    override val inputAggBufferOffset: Int) extends TypedImperativeAggregate[MapDigest] {
+
+  def this(child: Expression, numBinsExpression: Expression) = {
+    this(child, numBinsExpression, 0, 0)
+  }
+
+  // Mark as lazy so that numBinsExpression is not evaluated during tree transformation.
+  private lazy val numBins: Int = numBinsExpression.eval().asInstanceOf[Int]
```
gatorsmile: `asInstanceOf[Int]` is risky. When `eval()` returns null, it will be casted to 0.

gatorsmile: How about this change?
```Scala
  private lazy val numBins: Int = numBinsExpression.eval() match {
    case o: Int => o
    case x => throw new AnalysisException(
      s"The maximum number of bins provided must be a foldable integer expression: $x")
  }
```

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/MapAggregate.scala
wzhfy: We only keep null-null values in histogram, number of nulls is already stored in ColumnStat.

gatorsmile: I see. Thanks!

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/MapAggregate.scala
wzhfy: yeah, that's better.

sql/core/src/test/scala/org/apache/spark/sql/MapAggregateQuerySuite.scala
wzhfy: ok

sql/core/src/test/scala/org/apache/spark/sql/MapAggregateQuerySuite.scala
```diff
@@ -0,0 +1,144 @@
...
+      StructField("c1", ByteType),
+      StructField("c2", ShortType),
+      StructField("c3", IntegerType),
+      StructField("c4", LongType),
+      StructField("c5", FloatType),
+      StructField("c6", DoubleType),
+      StructField("c7", DecimalType(10, 5)),
+      StructField("c8", DateType),
+      StructField("c9", TimestampType),
+      StructField("c10", StringType)))
```
gatorsmile: Here, it still misses `BinaryType` and `BooleanType`

gatorsmile:
How about the complex types? `Array`, `Map` and `Struct`? It sounds like the test cases do not cover these test cases. Thanks!

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/MapAggregate.scala
```diff
@@ -0,0 +1,332 @@
...
+ * @param child child expression that can produce column value with `child.eval(inputRow)`
+ * @param numBinsExpression The maximum number of bins.
+ */
+@ExpressionDescription(
+  usage =
+    """
+      _FUNC_(col, numBins) - Returns 1. null if no non-null value exists.
+      2. (distinct non-null value, frequency) pairs of equi-width histogram when the number of
+      distinct non-null values is less than or equal to the specified maximum number of bins.
+      3. an empty map otherwise.
```
gatorsmile: Describe the general description of this function at first, and then explains the return values?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/MapAggregate.scala
```diff
@@ -0,0 +1,332 @@
...
+
+  def this(child: Expression, numBinsExpression: Expression) = {
+    this(child, numBinsExpression, 0, 0)
+  }
+
+  // Mark as lazy so that numBinsExpression is not evaluated during tree transformation.
+  private lazy val numBins: Int = numBinsExpression.eval().asInstanceOf[Int]
+
+  override def inputTypes: Seq[AbstractDataType] = {
+    Seq(TypeCollection(NumericType, TimestampType, DateType, StringType), IntegerType)
```
gatorsmile: Any reason why `Boolean` and `Binary` types are not supported?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/MapAggregate.scala
gatorsmile: Try to use `null` as the value of numBins. You will understand what I mean.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/MapAggregate.scala
```diff
@@ -0,0 +1,332 @@
...
+  }
+
+  override def update(buffer: MapDigest, input: InternalRow): Unit = {
+    if (buffer.isInvalid) {
+      return
+    }
+    val evaluated = child.eval(input)
+    if (evaluated != null) {
+      buffer.update(child.dataType, evaluated, numBins)
+    }
```
gatorsmile: A general comment about the impl. Here, I think we should avoid `return` if possible. For example, we can re-write it like
```Scala
    if (!buffer.isInvalid) {
      val evaluated = child.eval(input)
      if (evaluated != null) {
        buffer.update(child.dataType, evaluated, numBins)
      }
    }
```

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/MapAggregate.scala
wzhfy: Boolean data only has two values, and they are already in `numTrues` and `numFalses` in `ColumnStat`.
Binary data usually represents big objects, and usually it won't be used in filter conditions or join conditions, so we don't compute its ndv and histogram.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/MapAggregate.scala
wzhfy: ok

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/MapAggregate.scala
wzhfy: Oh, I see, thanks!

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/MapAggregate.scala
```diff
@@ -0,0 +1,332 @@
...
+case class NumericMapDigest(
+    bins: mutable.HashMap[Double, Long] = mutable.HashMap.empty) extends MapDigest {
+  def this(bins: mutable.HashMap[Double, Long], isInvalid: Boolean) = {
+    this(bins)
+    this.isInvalid = isInvalid
+  }
+
+  override def update(dataType: DataType, value: Any, numBins: Int): Unit = {
+    // Use Double to represent endpoints (in histograms) for simplicity
+    val doubleValue = dataType match {
```
srinathshankar: Couldn't you leave this type promotion to the engine ?
That is, for numeric types instead of calling
`mapaggregate(numericcol, bins)`
call
`mapaggregate(cast(numericcol as double), bins)`

That way you only have to deal with doubles in your aggregate.

srinathshankar (commented):
There's  repeated code in serialization and deserialization between the StringMapDigest and NumericMapDigest objects. It seems like the only real  difference between these classes is the serde of the keys. Couldn't you just extract that out ? As a suggestion, maybe something like this

```
trait MapDigest {
 def merge(otherDigest: MapDigest, numBins: Int): Unit
  def update(dataType: DataType, value: Any, numBins: Int): Unit
  def clear(): Unit
  def serialize(): Array[Byte]
}

abstract class MapDigestGeneric[T](val bins: mutable.HashMap[T, Long], var isInvalid: Boolean = false) extends MapDigest {
  // Implementation of above methods with appropriate casts
 // Also deserialize/fromBytes
  // Overridden by String and DoubleDigest
  def keyLen(key: T): Int
  def putKey(key: T, buffer: ByteBuffer): Unit
  def getKey(buffer: ByteBuffer): T
}

class StringMapDigest extends MapDigestBase[UTF8String] {
}
// and 
class DoubleMapDigest extends MapDigestBase[Double] {
}
```

This might help cut down the repetition a but.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/MapAggregate.scala
gatorsmile: In the use cases of CBO, YES. However, this function becomes a general one. It could be also used/called by external users. Then, it might not make sense for this limit.

gatorsmile:
BTW, please follow the PR https://github.com/apache/spark/issues/15677 to rewrite your function description. Add an example for the new function if possible.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/MapAggregate.scala
ron8hu: In today's software development environment, speed is an important factor.  One of the agile development methodology principles is to avoid over design.  The over design can slow down the development and divert an engineer's attention on the unimportant cases.  Since we develop MapAggregate.scala for CBO, let's focus on CBO's need.  There is no need to develop NDV and histogram for binary data type.  For Boolean data type, we already have number of trues and number of falses defined in ColumnStat.  I think we are doing fine here.

ron8hu: In today's software development environment, speed is an important factor.  One of the agile development methodology principles is to avoid over design.  The over design can slow down the development and divert an engineer's attention on the unimportant cases.  Since we develop MapAggregate.scala for CBO, let's focus on CBO's need.  There is no need to develop NDV and histogram for binary data type.  For Boolean data type, we already have number of trues and number of falses defined in ColumnStat.  I think we are doing fine here.

ron8hu: In today's software development environment, speed is an important factor.  One of the agile development methodology principles is to avoid over design.  The over design can slow down the development and divert an engineer's attention on the unimportant cases.  Since we develop MapAggregate.scala for CBO, let's focus on CBO's need.  There is no need to develop NDV and histogram for binary data type.  For Boolean data type, we already have number of trues and number of falses defined in ColumnStat.  I think wxhfy's code is doing fine here.

ron8hu:
test

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/MapAggregate.scala
ron8hu: In today's software development environment, speed is an important factor.  One of the agile development methodology principles is to avoid over design.  The over design can slow down the development and divert an engineer's attention on the unimportant cases.  Since we develop MapAggregate.scala for CBO, let's focus on CBO's need.  There is no need to develop NDV and histogram for binary data type.  For Boolean data type, we already have number of trues and number of falses defined in ColumnStat.  I think wxhfy's code is doing fine here.

ron8hu: In today's software development environment, speed is an important factor.  One of the agile development methodology principles is to avoid over design.  The over design can slow down the development and divert an engineer's attention on the unimportant cases.  Since we develop MapAggregate.scala for CBO, let's focus on CBO's need.  There is no need to develop NDV and histogram for binary data type.  For Boolean data type, we already have number of trues and number of falses defined in ColumnStat.  I think wxhfy's code is doing fine here.

ron8hu: In today's software development environment, speed is an important factor.  One of the agile development methodology principles is to avoid over design.  The over design can slow down the development and divert an engineer's attention on the unimportant cases.  Since we develop MapAggregate.scala for CBO, let's focus on CBO's need.  There is no need to develop NDV and histogram for binary data type.  For Boolean data type, we already have number of trues and number of falses defined in ColumnStat.  I think wxhfy's code is doing fine here.

ron8hu: I posted same comments multiple times due to a browser glitch.  I just cleaned it up.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/MapAggregate.scala
wzhfy: Date and Timestamp will be casted to UTF8String instead of Double, right?

rxin:
BTW why is this function called map_aggregate?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/MapAggregate.scala
```diff
@@ -0,0 +1,310 @@
...
+  override def inputTypes: Seq[AbstractDataType] = {
+    Seq(TypeCollection(NumericType, TimestampType, DateType, StringType), IntegerType)
+  }
+
+  override def checkInputDataTypes(): TypeCheckResult = {
+    val defaultCheck = super.checkInputDataTypes()
+    if (defaultCheck.isFailure) {
+      defaultCheck
+    } else if (!numBinsExpression.foldable) {
+      TypeCheckFailure("The maximum number of bins provided must be a constant literal")
```
rxin: "must be a literal or constant foldable"

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/MapAggregate.scala
```diff
@@ -0,0 +1,310 @@
...
+
+/**
+ * The MapAggregate function Computes frequency for each distinct non-null value of a column.
+ * It returns: 1. null if the table is empty or all values of the column are null.
+ * 2. (distinct non-null value, frequency) pairs if the number of distinct non-null values is
+ * less than or equal to the specified threshold.
+ * 3. an empty result if the number of distinct non-null values exceeds that threshold.
+ *
+ * @param child child expression that can produce column value with `child.eval(inputRow)`
+ * @param numBinsExpression The maximum number of pairs.
```
rxin: can you improve the documentation for numBins? I don't really understand without reading the code or design doc what you mean by "max number of pairs"

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/MapAggregate.scala
```diff
@@ -0,0 +1,310 @@
...
+    }
+  }
+
+  override def children: Seq[Expression] = Seq(child, numBinsExpression)
+
+  override def prettyName: String = "map_aggregate"
+}
+
+trait MapDigest {
+  // Mark this MapDigest invalid when the size of the hashmap (ndv of the column) exceeds numBins
```
rxin: note - a basic thing: please use `/**  ...  */` for commenting member variables or methods, rather than `//`

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/MapAggregate.scala
```diff
@@ -0,0 +1,310 @@
...
+      case _ => MapType(DoubleType, LongType)
+    }
+  }
+
+  override def children: Seq[Expression] = Seq(child, numBinsExpression)
+
+  override def prettyName: String = "map_aggregate"
+}
+
+trait MapDigest {
```
rxin: add some classdoc explaining what this is. 

also why is this called "Map"?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/MapAggregate.scala
```diff
@@ -0,0 +1,310 @@
...
+  // Mark this MapDigest invalid when the size of the hashmap (ndv of the column) exceeds numBins
+  var isInvalid: Boolean
+  def update(dataType: DataType, value: Any, numBins: Int): Unit
+  def merge(otherDigest: MapDigest, numBins: Int): Unit
+  def clear(): Unit
+  def serialize(): Array[Byte]
+}
+
+abstract class MapDigestBase[T] extends MapDigest {
+  val bins: mutable.HashMap[T, Long]
```
rxin: use java.util.HashMap for better performance

rxin: also you need to document what T is, and what the Long is.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/MapAggregate.scala
```diff
@@ -0,0 +1,310 @@
...
+        } else {
+          bins.put(key, value)
+        }
+      }
+    }
+  }
+
+  override def clear(): Unit = bins.clear()
+
+  override def serialize(): Array[Byte] = {
```
rxin: should we just serialize using Kryo rather than writing our own serializer here?

rxin: maybe it's fine, but if you are writing our own binary protocol, I'd document it here.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/MapAggregate.scala
```diff
@@ -0,0 +1,310 @@
...
+  def clear(): Unit
+  def serialize(): Array[Byte]
+}
+
+abstract class MapDigestBase[T] extends MapDigest {
+  val bins: mutable.HashMap[T, Long]
+
+  // Update bins and clear it when its size exceeds numBins.
+  def updateMap(value: T, numBins: Int): Unit = {
+    mergeMap(mutable.HashMap(value -> 1L), numBins)
```
rxin: i think you are doing this backwards by calling mergeMap in updateMap. If anything, it should be the other way around, i.e. mergeMap has a loop that calls updateMap. That way you will not need to create a whole new hash map for each record, or loop over some hash map of one entry for every record.

rxin: once you do that you can also just put mergeMap into merge.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/MapAggregate.scala
rxin: actually if you just use kryo maybe you don't need any of the keyLen, putKey, getKey functions and it will look a lot simpler.

rxin:
@wzhfy I'm pretty confused by this. I go ahead and read the JIRA -- it is about computing histograms, but the current aggregate function just computes counts for each value, up until a certain number of distinct elements. Where are we doing the binning?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/MapAggregate.scala
```diff
@@ -0,0 +1,310 @@
...
+ * 3. an empty result if the number of distinct non-null values exceeds that threshold.
+ *
+ * @param child child expression that can produce column value with `child.eval(inputRow)`
+ * @param numBinsExpression The maximum number of pairs.
+ */
+@ExpressionDescription(
+  usage = """
+    _FUNC_(col, numBins) - Computes frequency for each distinct non-null value of column `col`.
+      It returns: 1. null if the table is empty or all values of column `col` are null.
+      2. (distinct non-null value, frequency) pairs if the number of distinct non-null values
```
rxin: it's really not frequency, but just count?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/MapAggregate.scala
rxin: Yea I think it's fine for now to not support those.

wzhfy:
@rxin Originally this pr computed endpoints for equi-width (distinct values) or equi-height histograms (percentiles), then @srinathshankar gave a comment [here](https://github.com/apache/spark/pull/15637#pullrequestreview-6319075) and I think it was reasonable, so I adopted it. Now this pr computes distinct values and their counts for building equi-width histograms, and when ndv > threshold, it returns an empty result. I just updated description of the corresponding Jira.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/MapAggregate.scala
wzhfy: I think why we are writing our own serializer here (and also in `ApproximatePercentile`) is because `SerializerInstance` is not thread-safe, right?

wzhfy: I think why we are writing our own serializer here (and also in `ApproximatePercentile`) is because `SerializerInstance` is not thread-safe, right?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/MapAggregate.scala
ron8hu: Frequency usually means the rate something occurs.  In statistics, frequency means the number of items occurring in a given category.  Here frequency actually is equivalent to count.  We use frequency and count interchangeably.

rxin:
Thanks - that makes sense. I got really confused by "numBins".

Sorry for throwing a monkey wrench now: should we just use count min sketch and use that for string columns? For numeric columns it seems to me just using histograms will be sufficient.

rxin:
BTW we do have a count min sketch implementation already in Spark, and we can turn it into an aggregate function pretty easily.

wzhfy:
@rxin By "histograms" for numeric columns, do you mean equi-height histogram? The main purpose of this pr is to construct equi-width histogram without prior knowledge of ndv, so that we can compute it with other stats in the same scan. And equi-width histogram is important for both string and numeric(including timestamp and date) types, because it provides more accurate estimation than equi-height histogram.

wzhfy:
@rxin About the name "map_aggregate", actually it was suggested by srinath, do you have a better name?

wzhfy:
@rxin Moreover, with this pr we can compute accurate equi-width histogram, while count min sketch only gets an estimated result? I think it's better to have an accurate result given that both methods need a table scan.

rxin:
How would this pr help you with equi-width histogram? This function just gives you the frequency count for keys when the cardinality is low.

wzhfy:
Yes, in our design, equi-width histogram is a seq of single valued bins, which is used for columns with low cardinality, so that we can get accurate estimation. When cardinality is high, equi-height histogram is better than equi-width histogram (with some fixed width), especially in data skew scenario, where the skewed value would be hidden in a equi-width bin, while in equi-height histogram, it would occupies many contiguous bins.

rxin:
Why not use count min sketch then? You would get more signal (with some error) for a much larger range of values.

wzhfy:
For equi-height hsitogram, we need extra info like ndv's in each bin. Does count min sketch also have this information? I had a discussion with Herman and Tim about histogram construction before, we want to compute all stats including histograms in a single table scan. They suggested we modify ApproximatePercentile to get equi-height histograms, i.e. endpoints of bins and the ndv's within them. And this pr is only for getting equi-width histograms. You can check [this JIRA](https://issues.apache.org/jira/browse/SPARK-17074) for the discussion.

wzhfy:
Oh, do you mean using count min sketch for equi-width histogram? Actually I don't know about the sketch, I need to look into it to see if it's easy to use as an agg function and also support float/double/decimal types. But I do think this pr takes much less effort to implement what we want :)

rxin:
It supports any data types, uses less space and can provide probabilistic frequencies for a large number of distinct values, and it's already implemented in Spark. We just need to add a wrapper for it to turn it into an aggregate function.

wzhfy:
OK, I'll try to use count min sketch. BTW, seems it only supports Integral and String types for now?

rxin:
It supports arbitrary objects

```
  /**
   * Increments {@code item}'s count by one.
   */
  public abstract void add(Object item);
```

wzhfy:
Yes, but in its impl, it actually only supports above types.
```
  @Override
  public void add(Object item, long count) {
    if (item instanceof String) {
      addString((String) item, count);
    } else {
      addLong(Utils.integralToLong(item), count);
    }
  }
```

rxin:
Yea good point - the underlying implementation really only needs a hash code, it would be trivial to support all types. But even easier, I think you can just compute the hash (using sql expression) and pass it in to support all data types.

wzhfy:
In that way, we can only get the hashed value instead of real value of the column, right? So I think we still need to implement a hash code for fractional types.

rxin:
We already have those don't we? sparks own hash expresssion.

rxin:
To be clear, I'm suggesting

```
select count_min_sketch(hash(col1)), count_min_sketch(hash(col2)), ... from table
```

the count_min_sketch aggregate function only accepts a long value, and returns a byte array (binary)

wzhfy:
I mean count_min_sketch(hash(col1)) returns result of hash(col1) rather than col1, right? but we need the result of col1.

wzhfy:
Please correct me if i'm wrong about count_min_sketch, i don't have laptop with me now and will look at it when i get back.

wzhfy:
@rxin After looking at the `CountMinSketch`, I have a question:
It supports point queries, i.e., given a value, return its frequency. Can we get all distinct values  from it? What we want is (value, frequency) pairs. If we only support point queries, how can we estimate selectivity of predicates like col < 100 and col > 50?

rxin:
Weird - github didn't email me notifications.

Why do you care about all (value, freq) pairs? All you want should just be some summary statistics that can tell you given a value, what the estimated freq is, isn't it?

wzhfy:
With (value, freq) pairs, i.e. equi-width histogram, given a predicate col<100 and col>50, we can know exactly how many records are in that range, which is the estimated rowcount after the filter. And we can also know the exact ndv after the filter, which is important for following estimation, e.g., in estimation of output row count of agg(group by).

wzhfy:
Point query of countMinSketch can only help us on estimation for predicates like col=1.

rxin:
So I don't get it - I thought you were only using the value, freq pairs (i.e. equi-width histogram with width = 1 - can we stop calling it equi-width histogram?) for string types. Why would the question you are asking matter for string types?

wzhfy:
Since the JIRA is closed, I'm closing this PR too.

PR #15639, opened 2016-10-26
[SPARK-18126][Spark-Core]:getIteratorZipWithIndex accepts negative value as index

wangmiao1981:

(Please fill in changes proposed in this fix)

`Utils.getIteratorZipWithIndex` was added to deal with number of records > 2147483647 in one partition.

method `getIteratorZipWithIndex` accepts `startIndex` < 0, which leads to negative index.

This PR just adds a defensive check on `startIndex` to make sure it is >= 0.
## How was this patch tested?

Add a new unit test.

### DIFF
--- a/core/src/main/scala/org/apache/spark/util/Utils.scala
+++ b/core/src/main/scala/org/apache/spark/util/Utils.scala
@@ -1765,6 +1765,7 @@ private[spark] object Utils extends Logging {
    */
   def getIteratorZipWithIndex[T](iterator: Iterator[T], startIndex: Long): Iterator[(T, Long)] = {
     new Iterator[(T, Long)] {
+      require(startIndex > 0, "startIndex should be > 0.")
       var index: Long = startIndex - 1L
       def hasNext: Boolean = iterator.hasNext
       def next(): (T, Long) = {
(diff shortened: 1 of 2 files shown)


wangmiao1981:
@WeichenXu123 Can you take a look? Thanks!

core/src/main/scala/org/apache/spark/util/Utils.scala
```diff
@@ -1765,6 +1765,7 @@ private[spark] object Utils extends Logging {
    */
   def getIteratorZipWithIndex[T](iterator: Iterator[T], startIndex: Long): Iterator[(T, Long)] = {
     new Iterator[(T, Long)] {
+      require(startIndex > 0, "startIndex should be > 0.")
```
WeichenXu123: It seems to be `startIndex >= 0` ?

wangmiao1981: In the following line, you do `var index: Long = startIndex - 1L`. If it is == 0, then this line is -1L.

WeichenXu123: yeah, this case inital value = -1, but fisrt generated index is 0, because there is a `index += 1` clause running first.

wangmiao1981: OK. I will update the check to >= 0.

WeichenXu123:
LGTM. thanks~

srowen (approved):

rxin:
Can you create a JIRA ticket and add it to the title?

wangmiao1981:
Sure. I am doing it now. I thought it was simple. Thanks!

wangmiao1981:
@rxin Done. I created a jira and updated the title accordingly.

rxin:
Thanks - merging in master.

PR #15640, opened 2016-10-26
[SPARK-18106][SQL] ANALYZE TABLE should raise a ParseException for invalid option

dongjoon-hyun:

Currently, `ANALYZE TABLE` command accepts `identifier` for option `NOSCAN`. This PR raises a ParseException for unknown option.

**Before**

``` scala
scala> sql("create table test(a int)")

scala> sql("analyze table test compute statistics blah")
res1: org.apache.spark.sql.DataFrame = []
```

**After**

``` scala
scala> sql("create table test(a int)")

scala> sql("analyze table test compute statistics blah")
Expected `NOSCAN` instead of `blah`(line 1, pos 0)
```
## How was this patch tested?

Pass the Jenkins test with a new test case.

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/SparkSqlParser.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/SparkSqlParser.scala
@@ -98,9 +98,10 @@ class SparkSqlAstBuilder(conf: SQLConf) extends AstBuilder {
    * }}}
    */
   override def visitAnalyze(ctx: AnalyzeContext): LogicalPlan = withOrigin(ctx) {
-    if (ctx.partitionSpec == null &&
-      ctx.identifier != null &&
-      ctx.identifier.getText.toLowerCase == "noscan") {
+    if (ctx.partitionSpec == null && ctx.identifier != null) {
+      if (ctx.identifier.getText.toLowerCase != "noscan") {
+        throw new ParseException(s"Expected `NOSCAN` instead of `${ctx.identifier.getText}`", ctx)
+      }
       AnalyzeTableCommand(visitTableIdentifier(ctx.tableIdentifier))
     } else if (ctx.identifierSeq() == null) {
       AnalyzeTableCommand(visitTableIdentifier(ctx.tableIdentifier), noscan = false)
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/SparkSqlParser.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/SparkSqlParser.scala
@@ -98,10 +98,10 @@ class SparkSqlAstBuilder(conf: SQLConf) extends AstBuilder {
    * }}}
    */
   override def visitAnalyze(ctx: AnalyzeContext): LogicalPlan = withOrigin(ctx) {
+    if (ctx.identifier != null && ctx.identifier.getText.toLowerCase != "noscan") {
+      throw new ParseException(s"Expected `NOSCAN` instead of `${ctx.identifier.getText}`", ctx)
+    }
     if (ctx.partitionSpec == null && ctx.identifier != null) {
-      if (ctx.identifier.getText.toLowerCase != "noscan") {
-        throw new ParseException(s"Expected `NOSCAN` instead of `${ctx.identifier.getText}`", ctx)
-      }
       AnalyzeTableCommand(visitTableIdentifier(ctx.tableIdentifier))
     } else if (ctx.identifierSeq() == null) {
       AnalyzeTableCommand(visitTableIdentifier(ctx.tableIdentifier), noscan = false)
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/SparkSqlParser.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/SparkSqlParser.scala
@@ -98,12 +98,12 @@ class SparkSqlAstBuilder(conf: SQLConf) extends AstBuilder {
    * }}}
    */
   override def visitAnalyze(ctx: AnalyzeContext): LogicalPlan = withOrigin(ctx) {
+    if (ctx.partitionSpec != null) {
+      logWarning(s"Partition specification is ignored: ${ctx.partitionSpec.getText}")
+    }
     if (ctx.identifier != null) {
       if (ctx.identifier.getText.toLowerCase != "noscan") {
         throw new ParseException(s"Expected `NOSCAN` instead of `${ctx.identifier.getText}`", ctx)
-      }
-      if (ctx.partitionSpec != null) {
-        logWarning(s"Partition specification is ignored: ${ctx.partitionSpec.getText}")
       }
       AnalyzeTableCommand(visitTableIdentifier(ctx.tableIdentifier))
     } else if (ctx.identifierSeq() == null) {
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/SparkSqlParser.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/SparkSqlParser.scala
@@ -98,10 +98,13 @@ class SparkSqlAstBuilder(conf: SQLConf) extends AstBuilder {
    * }}}
    */
   override def visitAnalyze(ctx: AnalyzeContext): LogicalPlan = withOrigin(ctx) {
-    if (ctx.identifier != null && ctx.identifier.getText.toLowerCase != "noscan") {
-      throw new ParseException(s"Expected `NOSCAN` instead of `${ctx.identifier.getText}`", ctx)
-    }
-    if (ctx.partitionSpec == null && ctx.identifier != null) {
+    if (ctx.identifier != null) {
+      if (ctx.identifier.getText.toLowerCase != "noscan") {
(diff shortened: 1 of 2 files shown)


sql/core/src/main/scala/org/apache/spark/sql/execution/SparkSqlParser.scala
```diff
@@ -98,9 +98,10 @@ class SparkSqlAstBuilder(conf: SQLConf) extends AstBuilder {
    * }}}
    */
   override def visitAnalyze(ctx: AnalyzeContext): LogicalPlan = withOrigin(ctx) {
-    if (ctx.partitionSpec == null &&
-      ctx.identifier != null &&
-      ctx.identifier.getText.toLowerCase == "noscan") {
+    if (ctx.partitionSpec == null && ctx.identifier != null) {
```
srinathshankar: What if partition spec is not null ?
What happens with something like 
```
ANALYZE TABLE mytable PARTITION (a) garbage
```
(Could you add a test for that ?)
Maybe 
```
if (ctx.identifier != null && ctx.identifier.getText.toLowerCase != "noscan") {
  throw new ParseException(s"Expected `NOSCAN` instead of `${ctx.identifier.getText}`", ctx)
}
```
could be moved to the top ?

sql/core/src/test/scala/org/apache/spark/sql/execution/SparkSqlParserSuite.scala
```diff
@@ -220,4 +220,12 @@ class SparkSqlParserSuite extends PlanTest {
 
     intercept("explain describe tables x", "Unsupported SQL statement")
   }
+
+  test("SPARK-18106 analyze table") {
```
srinathshankar: There are also parse tests for AnalyzeTable in sql/hive/src/test/scala/org/apache/spark/sql/hive/StatisticsSuite.scala
Let's have these in the same place

srinathshankar (changes requested):
Thanks for the quick PR

sql/core/src/main/scala/org/apache/spark/sql/execution/SparkSqlParser.scala
dongjoon-hyun: Thank you for review and I'll handle that too.

sql/core/src/test/scala/org/apache/spark/sql/execution/SparkSqlParserSuite.scala
dongjoon-hyun: Sure.

dongjoon-hyun: Ur, @srinathshankar .

I understand the reason why you put this there, so I looked at the StatisticsSuite.scala in both `hive` module and `sql` module.
But, we will not compare the value for this test case. If it's a parsing only grammar testcase, I prefer to put this in `core`.

How do you think about that?

dongjoon-hyun: If you want, I'll remove the normal cases which raises no exceptions.

sql/core/src/main/scala/org/apache/spark/sql/execution/SparkSqlParser.scala
dongjoon-hyun: Test case is added.

dongjoon-hyun:
Hi, @srinathshankar .
For the test suite, I'll update again if you give more opinion. So far, I didn't update that because I'm not sure.

dongjoon-hyun:
Hi, @hvanhovell .
Could you review this PR when you have some time?

sql/core/src/test/scala/org/apache/spark/sql/execution/SparkSqlParserSuite.scala
srinathshankar: Then maybe you can move those parse tests here ? All I'm suggesting is that the parse tests all be together.

sql/core/src/test/scala/org/apache/spark/sql/execution/SparkSqlParserSuite.scala
```diff
@@ -220,4 +220,13 @@ class SparkSqlParserSuite extends PlanTest {
...
+    assertEqual("analyze table t compute statistics",
+      AnalyzeTableCommand(TableIdentifier("t"), noscan = false))
+    assertEqual("analyze table t compute statistics noscan",
+      AnalyzeTableCommand(TableIdentifier("t"), noscan = true))
+    intercept("analyze table t compute statistics xxxx", "Expected `NOSCAN` instead of `xxxx`")
+    intercept("analyze table t partition (a) compute statistics xxxx")
```
srinathshankar: Nit: `, "Expected `NOSCAN` instead of `xxxx`"` here as well.

dongjoon-hyun: Thank you. But, the parse test should be rewritten. Is it okay? Those testcase uses `assertAnalyzeCommand` using SparkSession and looks like the following.
```scala
    def assertAnalyzeCommand(analyzeCommand: String, c: Class[_]) {
      val parsed = spark.sessionState.sqlParser.parsePlan(analyzeCommand)
      val operators = parsed.collect {
        case a: AnalyzeTableCommand => a
        case o => o
      }

      assert(operators.size === 1)
      if (operators(0).getClass() != c) {
        fail(
          s"""$analyzeCommand expected command: $c, but got ${operators(0)}
             |parsed command:
             |$parsed
           """.stripMargin)
      }
    }

    assertAnalyzeCommand(
      "ANALYZE TABLE Table1 COMPUTE STATISTICS",
      classOf[AnalyzeTableCommand])
```

sql/core/src/test/scala/org/apache/spark/sql/execution/SparkSqlParserSuite.scala
dongjoon-hyun: Thank you. But, the parse test should be rewritten. Is it okay? Those testcase uses `assertAnalyzeCommand` using SparkSession and looks like the following.
```scala
    def assertAnalyzeCommand(analyzeCommand: String, c: Class[_]) {
      val parsed = spark.sessionState.sqlParser.parsePlan(analyzeCommand)
      val operators = parsed.collect {
        case a: AnalyzeTableCommand => a
        case o => o
      }

      assert(operators.size === 1)
      if (operators(0).getClass() != c) {
        fail(
          s"""$analyzeCommand expected command: $c, but got ${operators(0)}
             |parsed command:
             |$parsed
           """.stripMargin)
      }
    }

    assertAnalyzeCommand(
      "ANALYZE TABLE Table1 COMPUTE STATISTICS",
      classOf[AnalyzeTableCommand])
```

srinathshankar: In my opinion it's fine to rewrite and simplify. If you could do that, that would be great.

dongjoon-hyun: Yep. I have no objection for that. Actually, I love to do that.
But, I'd like to wait for some directional advice from committer.

sql/core/src/main/scala/org/apache/spark/sql/execution/SparkSqlParser.scala
```diff
@@ -98,9 +98,10 @@ class SparkSqlAstBuilder(conf: SQLConf) extends AstBuilder {
    * }}}
    */
   override def visitAnalyze(ctx: AnalyzeContext): LogicalPlan = withOrigin(ctx) {
-    if (ctx.partitionSpec == null &&
-      ctx.identifier != null &&
-      ctx.identifier.getText.toLowerCase == "noscan") {
+    if (ctx.identifier != null && ctx.identifier.getText.toLowerCase != "noscan") {
```
hvanhovell: I also think that the treatment of the `partionSpec` is quite funky. We scan the table as soon as a user defines a spec. Could you remove the null check; maybe it is better to just log a warning message and do what the user specified.

sql/core/src/test/scala/org/apache/spark/sql/execution/SparkSqlParserSuite.scala
hvanhovell: Lets keep the parsing unit test (this file) and the analyze table integration test separate for now.

hvanhovell (changes requested):
This looks ok. I left a few comments.

sql/core/src/test/scala/org/apache/spark/sql/execution/SparkSqlParserSuite.scala
dongjoon-hyun: Thank you for the guide!

dongjoon-hyun:
The only test failure seems to be irrelevant. Let's see the final test result which is still running.
```
[info] - SPARK-10562: partition by column with mixed case name *** FAILED *** (687 milliseconds)
```

hvanhovell:

dongjoon-hyun:
Thank you, @hvanhovell ! Also, Thank you, @srinathshankar .

PR #15644, opened 2016-10-26
[SPARK-18117][CORE] Add test for TaskSetBlacklist

squito:

This adds tests to verify the interaction between TaskSetBlacklist and
TaskSchedulerImpl.  TaskSetBlacklist was introduced by SPARK-17675 but
it neglected to add these tests.

This change does not fix any bugs -- it is just for increasing test
coverage.
## How was this patch tested?

Jenkins

### DIFF
--- a/core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
+++ b/core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
@@ -317,19 +317,15 @@ class TaskSchedulerImplSuite extends SparkFunSuite with LocalSparkContext with B
     assert(!failedTaskSet)
   }
 
-  test("scheduled tasks obey task and stage blacklists") {
-    taskScheduler = setupSchedulerWithMockTaskSetBlacklist()
-    (0 to 2).foreach { stageId =>
+  /**
+   * Create tasksets, a mock blacklist, and a set of offers commmon to some tests on taskset
+   * blacklists.
+   */
+  private def taskSetBlacklistSetup(): IndexedSeq[WorkerOffer] = {
+    (0 to 2).foreach {stageId =>
       val taskSet = FakeTask.createTaskSet(numTasks = 2, stageId = stageId, stageAttemptId = 0)
       taskScheduler.submitTasks(taskSet)
     }
-
-    val offers = IndexedSeq(
-      new WorkerOffer("executor0", "host0", 1),
-      new WorkerOffer("executor1", "host1", 1),
-      new WorkerOffer("executor2", "host1", 1),
-      new WorkerOffer("executor3", "host2", 10)
-    )
 
     // Setup our mock blacklist:
     // * stage 0 is blacklisted on node "host1"
@@ -348,8 +344,17 @@ class TaskSchedulerImplSuite extends SparkFunSuite with LocalSparkContext with B
       .thenReturn(true)
     when(stageToMockTaskSetBlacklist(0).isExecutorBlacklistedForTask("executor0", 0))
       .thenReturn(true)
+    IndexedSeq(
+      new WorkerOffer("executor0", "host0", 1),
+      new WorkerOffer("executor1", "host1", 1),
+      new WorkerOffer("executor2", "host1", 1),
+      new WorkerOffer("executor3", "host2", 10)
+    )
+  }
 
-    val firstTaskAttempts = taskScheduler.resourceOffers(offers).flatten
+  test("scheduled tasks obey task and stage blacklists") {
+    taskScheduler = setupSchedulerWithMockTaskSetBlacklist()
+    val firstTaskAttempts = taskScheduler.resourceOffers(taskSetBlacklistSetup()).flatten
     // Whenever we schedule a task, we must consult the node and executor blacklist.  (The test
     // doesn't check exactly what checks are made the offers get shuffled.)
     (0 to 2).foreach { stageId =>
@@ -433,6 +438,30 @@ class TaskSchedulerImplSuite extends SparkFunSuite with LocalSparkContext with B
         }
       }
       assert(tsm.isZombie)
+    }
+  }
+
+  test("executor and node blacklisting prevents per-task checks") {
+    // When an executor or node is blacklisted, we want to make sure that we don't try scheduling
+    // each pending task, one by one, to discover they are all blacklisted.  This is important for
+    // performance -- if we did check each task one-by-one, then responding to a resource offer
+    // (which is usually O(1)-ish) would become O(numPendingTasks), which would slow down
+    // scheduler throughput and slow down scheduling even on healthy executors.
+    // Here, we check a proxy for the runtime -- we make sure the scheduling is short-circuited
+    // at the node or executor blacklist, so we never check the per-task blacklist.
+
+    taskScheduler = setupSchedulerWithMockTaskSetBlacklist()
+    val firstTaskAttempts = taskScheduler.resourceOffers(taskSetBlacklistSetup()).flatten
+    for {
+      exec <- Seq("executor1", "executor2")
+      part <- 0 to 1
+    } {
+      // stage 0 is blacklisted on the host for these executors
+      verify(stageToMockTaskSetBlacklist(0), never).isExecutorBlacklistedForTask(exec, part)
+    }
+    // stage 0 is blacklisted for executor3
+    (0 to 1).foreach { part =>
+      verify(stageToMockTaskSetBlacklist(1), never).isExecutorBlacklistedForTask("executor3", part)
     }
   }
 
--- a/core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
+++ b/core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
@@ -80,7 +80,7 @@ class TaskSchedulerImplSuite extends SparkFunSuite with LocalSparkContext with B
     setupHelper()
   }
 
-  def setupSchedulerWithMockTsm(): TaskSchedulerImpl = {
+  def setupSchedulerWithMockTaskSetBlacklist(): TaskSchedulerImpl = {
     val conf = new SparkConf().setMaster("local").setAppName("TaskSchedulerImplSuite")
     conf.set(config.BLACKLIST_ENABLED, true)
     sc = new SparkContext(conf)
@@ -88,6 +88,7 @@ class TaskSchedulerImplSuite extends SparkFunSuite with LocalSparkContext with B
       new TaskSchedulerImpl(sc, sc.conf.getInt("spark.task.maxFailures", 4)) {
         override def createTaskSetManager(taskSet: TaskSet, maxFailures: Int): TaskSetManager = {
           val tsm = super.createTaskSetManager(taskSet, maxFailures)
+          // we need to create a spied tsm just so we can set the TaskSetBlacklist
           val tsmSpy = spy(tsm)
           val taskSetBlacklist = mock[TaskSetBlacklist]
           when(tsmSpy.taskSetBlacklistHelperOpt).thenReturn(Some(taskSetBlacklist))
@@ -317,7 +318,7 @@ class TaskSchedulerImplSuite extends SparkFunSuite with LocalSparkContext with B
   }
 
   test("scheduled tasks obey task and stage blacklists") {
-    taskScheduler = setupSchedulerWithMockTsm()
+    taskScheduler = setupSchedulerWithMockTaskSetBlacklist()
     (0 to 2).foreach { stageId =>
       val taskSet = FakeTask.createTaskSet(numTasks = 2, stageId = stageId, stageAttemptId = 0)
       taskScheduler.submitTasks(taskSet)
@@ -333,8 +334,8 @@ class TaskSchedulerImplSuite extends SparkFunSuite with LocalSparkContext with B
     // Setup our mock blacklist:
     // * stage 0 is blacklisted on node "host1"
     // * stage 1 is blacklisted on executor "executor3"
-    // * stage 0, part 0 is blacklisted on executor 0
-    // Setup some defaults, then override them with particulars.
+    // * stage 0, partition 0 is blacklisted on executor 0
+    // Setup some defaults (nothing is blacklisted), then override them with particulars.
     // (Later stubs take precedence over earlier ones.)
     stageToMockTaskSetBlacklist.values.foreach { taskSetBlacklist =>
       when(taskSetBlacklist.isNodeBlacklistedForTaskSet(anyString())).thenReturn(false)
@@ -349,22 +350,30 @@ class TaskSchedulerImplSuite extends SparkFunSuite with LocalSparkContext with B
       .thenReturn(true)
 
     val firstTaskAttempts = taskScheduler.resourceOffers(offers).flatten
+    // Whenever we schedule a task, we must consult the node and executor blacklist.  (The test
+    // doesn't check exactly what checks are made the offers get shuffled.)
     (0 to 2).foreach { stageId =>
       verify(stageToMockTaskSetBlacklist(stageId), atLeast(1))
         .isNodeBlacklistedForTaskSet(anyString())
+      verify(stageToMockTaskSetBlacklist(stageId), atLeast(1))
+        .isExecutorBlacklistedForTaskSet(anyString())
     }
+
+    // When an executor or node is blacklisted, we want to make sure that we don't try scheduling
+    // each pending task, one by one, to discover they are all blacklisted.  This is important for
+    // performance -- if we did check each task one-by-one, then responding to a resource offer
+    // (which is usually O(1)-ish) would become O(numPendingTasks), which would slow down
+    // scheduler throughput and slow down scheduling even on healthy executors.
+    // Here, we check a proxy for the runtime -- we make sure the scheduling is short-circuited
+    // at the node or executor blacklist, so we never check the per-task blacklist.
     for {
       exec <- Seq("executor1", "executor2")
       part <- 0 to 1
     } {
-      // The node blacklist should ensure we never check the task blacklist.  This is important
-      // for performance, otherwise we end up changing an O(1) operation into a
-      // O(numPendingTasks) one.
+      // stage 0 is blacklisted on the host for these executors
       verify(stageToMockTaskSetBlacklist(0), never).isExecutorBlacklistedForTask(exec, part)
     }
-
-    // Similarly, the executor blacklist for an entire stage should prevent us from ever checking
-    // the blacklist for specific parts in a stage.
+    // stage 0 is blacklisted for executor3
     (0 to 1).foreach { part =>
       verify(stageToMockTaskSetBlacklist(1), never).isExecutorBlacklistedForTask("executor3", part)
     }
@@ -375,8 +384,8 @@ class TaskSchedulerImplSuite extends SparkFunSuite with LocalSparkContext with B
       firstTaskAttempts.filter{_.name.contains(s"stage $stageId")}
     }
     tasksForStage(0).foreach { task =>
-      // exec 1 & 2 blacklisted for node
-      // exec 0 blacklisted just for part 0
+      // executors 1 & 2 blacklisted for node
+      // executor 0 blacklisted just for partition 0
       if (task.index == 0) {
         assert(task.executorId === "executor3")
       } else {
@@ -384,12 +393,14 @@ class TaskSchedulerImplSuite extends SparkFunSuite with LocalSparkContext with B
       }
     }
     tasksForStage(1).foreach { task =>
-      // exec 3 blacklisted
+      // executor 3 blacklisted
       assert("executor3" != task.executorId)
     }
     // no restrictions on stage 2
 
-    // Have all tasksets finish (stages 0 & 1 successfully, 2 unsuccessfully).
+    // Finally, just make sure that we can still complete tasks as usual with blacklisting
+    // in effect.  Finish each of the tasksets -- taskset 0 & 1 complete successfully, taskset 2
+    // fails.
     (0 to 2).foreach { stageId =>
       val tasks = tasksForStage(stageId)
       val tsm = taskScheduler.taskSetManagerForAttempt(stageId, 0).get
@@ -421,6 +432,7 @@ class TaskSchedulerImplSuite extends SparkFunSuite with LocalSparkContext with B
           tsm.handleSuccessfulTask(task.taskId, result)
         }
       }
+      assert(tsm.isZombie)
     }
   }
 
@@ -555,7 +567,7 @@ class TaskSchedulerImplSuite extends SparkFunSuite with LocalSparkContext with B
     "or killed tasks") {
     // Setup a taskset, and fail some tasks for a fetch failure, preemption, denied commit,
     // and killed task.
-    taskScheduler = setupSchedulerWithMockTsm()
+    taskScheduler = setupSchedulerWithMockTaskSetBlacklist()
     val stage0 = FakeTask.createTaskSet(numTasks = 4, stageId = 0, stageAttemptId = 0)
     taskScheduler.submitTasks(stage0)
     val taskDescs = taskScheduler.resourceOffers(
--- a/core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
+++ b/core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
@@ -27,7 +27,6 @@ import org.scalatest.mock.MockitoSugar
 import org.apache.spark._
 import org.apache.spark.internal.config
 import org.apache.spark.internal.Logging
-import org.apache.spark.storage.BlockManagerId
 
 class FakeSchedulerBackend extends SchedulerBackend {
   def start() {}
@@ -412,6 +411,8 @@ class TaskSchedulerImplSuite extends SparkFunSuite with LocalSparkContext with B
   /**
    * Helper for performance tests.  Takes the explicitly blacklisted nodes and executors; verifies
    * that the blacklists are used efficiently to ensure scheduling is not O(numPendingTasks).
+   * Creates 1 offer on executor[1-3].  Executor1 & 2 are on host1, executor3 is on host2.  Passed
+   * in nodes and executors should be on that list.
    */
   private def blacklistPerformanceCheck(
       nodeBlacklist: Seq[String],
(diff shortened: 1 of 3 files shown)


squito:
@kayousterhout here are the tests that should have been in SPARK-17675.

core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
```diff
@@ -282,6 +316,114 @@ class TaskSchedulerImplSuite extends SparkFunSuite with LocalSparkContext with B
...
+      new WorkerOffer("executor0", "host0", 1),
+      new WorkerOffer("executor1", "host1", 1),
+      new WorkerOffer("executor2", "host1", 1),
+      new WorkerOffer("executor3", "host2", 10)
+    )
+
+    // Setup our mock blacklist:
+    // * stage 0 is blacklisted on node "host1"
+    // * stage 1 is blacklisted on executor "executor3"
+    // * stage 0, part 0 is blacklisted on executor 0
```
kayousterhout: can you write out partition here? (took me a minute...)

core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
```diff
@@ -282,6 +316,114 @@ class TaskSchedulerImplSuite extends SparkFunSuite with LocalSparkContext with B
...
+    // Setup some defaults, then override them with particulars.
```
kayousterhout: can you add maybe "(nothing is blacklisted)" after "some defaults"?

core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
```diff
@@ -282,6 +316,114 @@ class TaskSchedulerImplSuite extends SparkFunSuite with LocalSparkContext with B
...
+    when(stageToMockTaskSetBlacklist(0).isNodeBlacklistedForTaskSet("host1")).thenReturn(true)
+    when(stageToMockTaskSetBlacklist(1).isExecutorBlacklistedForTaskSet("executor3"))
+      .thenReturn(true)
+    when(stageToMockTaskSetBlacklist(0).isExecutorBlacklistedForTask("executor0", 0))
+      .thenReturn(true)
+
+    val firstTaskAttempts = taskScheduler.resourceOffers(offers).flatten
+    (0 to 2).foreach { stageId =>
+      verify(stageToMockTaskSetBlacklist(stageId), atLeast(1))
+        .isNodeBlacklistedForTaskSet(anyString())
```
kayousterhout: What's happening here? Just checking that TaskSchedulerImpl invoked isNodeBlacklistedForTaskSet?  I'd be inclined to just eliminate this because it seems too white-box (and therefore brittle to changing code), and doesn't have the important performance implications of the tests below -- but if you want to keep it, can you add a comment saying what's going on?

core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
```diff
@@ -282,6 +316,114 @@ class TaskSchedulerImplSuite extends SparkFunSuite with LocalSparkContext with B
...
+    }
+    for {
+      exec <- Seq("executor1", "executor2")
+      part <- 0 to 1
+    } {
+      // The node blacklist should ensure we never check the task blacklist.  This is important
+      // for performance, otherwise we end up changing an O(1) operation into a
+      // O(numPendingTasks) one.
```
kayousterhout: Why is this important for performance?  Won't this call still happen in the usual case, where no nodes are blacklisted? (and why does it become an O(numPendingTasks) operation?

core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
```diff
@@ -282,6 +316,114 @@ class TaskSchedulerImplSuite extends SparkFunSuite with LocalSparkContext with B
...
+      verify(stageToMockTaskSetBlacklist(1), never).isExecutorBlacklistedForTask("executor3", part)
+    }
+
+    // We should schedule all tasks.
+    assert(firstTaskAttempts.size === 6)
+    def tasksForStage(stageId: Int): Seq[TaskDescription] = {
+      firstTaskAttempts.filter{_.name.contains(s"stage $stageId")}
+    }
+    tasksForStage(0).foreach { task =>
+      // exec 1 & 2 blacklisted for node
```
kayousterhout: Can you write this out? ("Executors 1 and 2 shouldn't be used because the node they're on is blacklisted")

core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
```diff
@@ -282,6 +316,114 @@ class TaskSchedulerImplSuite extends SparkFunSuite with LocalSparkContext with B
...
+      // exec 0 blacklisted just for part 0
```
kayousterhout: partition

core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
```diff
@@ -282,6 +316,114 @@ class TaskSchedulerImplSuite extends SparkFunSuite with LocalSparkContext with B
...
+    // Have all tasksets finish (stages 0 & 1 successfully, 2 unsuccessfully).
+    (0 to 2).foreach { stageId =>
+      val tasks = tasksForStage(stageId)
+      val tsm = taskScheduler.taskSetManagerForAttempt(stageId, 0).get
+      val valueSer = SparkEnv.get.serializer.newInstance()
+      if (stageId == 2) {
+        // Just need to make one task fail 4 times.
+        var task = tasks(0)
+        val taskIndex = task.index
+        (0 until 4).foreach { attempt =>
```
kayousterhout: It looks like this part of the test is making sure no further attempts are scheduled after the 4th failure.  Is this necessary for the blacklisting functionality / different than what's already covered by other tests?

core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
```diff
@@ -66,6 +77,29 @@ class TaskSchedulerImplSuite extends SparkFunSuite with LocalSparkContext with B
     }
     sc = new SparkContext(conf)
     taskScheduler = new TaskSchedulerImpl(sc)
+    setupHelper()
+  }
+
+  def setupSchedulerWithMockTsm(): TaskSchedulerImpl = {
```
kayousterhout: Also, where is the spied TSM actually used? (can you add a comment about this) From a quick skim I could only find the blacklist methods being checked.

core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
```diff
@@ -282,6 +316,114 @@ class TaskSchedulerImplSuite extends SparkFunSuite with LocalSparkContext with B
     assert(!failedTaskSet)
   }
 
+  test("scheduled tasks obey task and stage blacklists") {
```
kayousterhout: Is it possible to split the performance tests (that certain successful blacklist calls cause others not to be called) into a separate test from the correctness tests? This test is giant and as a result I'm worried about the complexity of debugging future failures.

kayousterhout (commented):
Thanks for adding these tests!! The earlier one is scary huge -- I added some comments about this to potentially remove or move into a separate test to make it easier to read and debug in the future.

core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
squito: yeah the important part here is actually the mock blacklist, not the spied tsm -- I think that name was leftover from long ago.  I've updated.

core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
squito: sure.  to be honest, I'm not sure how much it helps -- pulling out the common setup makes it a bit harder to trace what is going on, IMO.  but I also don't feel strongly.  Its the last commit so you can see it both ways:

https://github.com/apache/spark/pull/15644/commits/6a026341de96c67218d18421a69a780a04994112

core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
squito: I've added a comment to explain a bit (and actually I expanded the check to include more).  I take your point about this being a little brittle, but I think to some extent that is unavoidable at this level of testing.

The code now really expects that every time a task is scheduled, teh node & executor blacklist get checked, which is the point of this check.  That is implicitly checked by other assertions below, but I worry that those are a narrow set of cases.  If there is some later changes to blacklisting, it would be good if this breaks and forces the developer to think about the main condition they are violating, rather than just seeing specific failure cases which maybe they shoe-horn a way around.  If this check gets removed later its fine, but it should be thought through.

core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
squito: I've updated the comment (and re-organized slightly) so hopefully this is more clear.  The point is to avoid calling `dequeueTask()` O(numPendingTasks) times inside the scheduler (that's one of the major problems w/ the old blacklisting).

core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
squito: The truth is, the whole block at the end is really more useful when I add in application-level blacklisting -- then its followed by checks to make sure that the app-level blacklist is updated appropriately.

I decided to leave it in here since its still seems somewhat useful to make sure that normal scheduling (as covered by other tests) still works the same with a bit of blacklisting in effect.  That does exist in some other places too, and this is also just covering a little bit, so its not a huge improvement.  happy to go either way.

core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
```diff
@@ -282,6 +317,154 @@ class TaskSchedulerImplSuite extends SparkFunSuite with LocalSparkContext with B
...
+        .isExecutorBlacklistedForTaskSet(anyString())
+    }
+
+    // When an executor or node is blacklisted, we want to make sure that we don't try scheduling
+    // each pending task, one by one, to discover they are all blacklisted.  This is important for
+    // performance -- if we did check each task one-by-one, then responding to a resource offer
+    // (which is usually O(1)-ish) would become O(numPendingTasks), which would slow down
+    // scheduler throughput and slow down scheduling even on healthy executors.
+    // Here, we check a proxy for the runtime -- we make sure the scheduling is short-circuited
+    // at the node or executor blacklist, so we never check the per-task blacklist.
```
kayousterhout: Also, can this be cut, since it's tested in the next test below? Thinking of ways to make this more of a unit test, since it's testing a ton of stuff right now

core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
```diff
@@ -282,6 +317,154 @@ class TaskSchedulerImplSuite extends SparkFunSuite with LocalSparkContext with B
...
+
+    taskScheduler = setupSchedulerWithMockTaskSetBlacklist()
+    val firstTaskAttempts = taskScheduler.resourceOffers(taskSetBlacklistSetup()).flatten
+    for {
+      exec <- Seq("executor1", "executor2")
+      part <- 0 to 1
+    } {
+      // stage 0 is blacklisted on the host for these executors
```
kayousterhout: can you elaborate here -- add something like ", so the host-level blacklist for the stage should keep the task-level blacklisting from every being checked"? Or alternately just eliminate the partial comment, since it's a little redundant with the longer method-level comment

core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
```diff
@@ -282,6 +317,154 @@ class TaskSchedulerImplSuite extends SparkFunSuite with LocalSparkContext with B
...
+      verify(stageToMockTaskSetBlacklist(0), never).isExecutorBlacklistedForTask(exec, part)
+    }
+    // stage 0 is blacklisted for executor3
```
kayousterhout: ditto to the above

core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
```diff
@@ -282,6 +317,154 @@ class TaskSchedulerImplSuite extends SparkFunSuite with LocalSparkContext with B
...
+    (0 to 1).foreach { part =>
+      verify(stageToMockTaskSetBlacklist(1), never).isExecutorBlacklistedForTask("executor3", part)
+    }
+
+    // We should schedule all tasks.
+    assert(firstTaskAttempts.size === 6)
```
kayousterhout: maybe move this to right after firstTaskAttempts is assigned?  Seems better to do this check right away, since if it fails, many other things will fail, but it will be less obvious what's wrong.

core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
kayousterhout: Ok makes sense.

core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
squito: ugh, i totally goofed my update here, sorry.  Yes, this should be cut here.

But for the equivalent part in the performance test:

We can't just check that `isExecutorBlacklistForTask` is never called, because it actually should be called for the tasks that are scheduled.  (I think its worthwhile that this test actually schedules tasks on some tasksets, to avoid any corner cases where nothing is scheduled.)  It does need to be specific for the stage / executor combos.

But you have a good point -- we really need to check that *none* of the methods are called O(numPendingTasks) times.  Eg., we want to make sure that we don't loop through each task in the taskset, and *inside* the loop check the node & executor blacklist for the entire taskset.

this will complicate this test (well, the performance one, not this one) somewhat, but I think its necessary given the issue you're pointing out.

squito:
actually, hang on ... I realize the performance test still isn't right.  I need to make sure that some of the blacklisted resources don't get scheduled to any tasks, since that is how you'd end up looping through all the pending tasks.  otherwise this might pass just b/c all the resources get scheduled, even if things changed to be O(numPendingTasks)

tgravescs:
Jenkins, test this please

squito:
@kayousterhout I've updated again, and did another pass for a little more cleanup.  I think it should be OK now.  sorry the earlier update left out some things.

core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
```diff
@@ -282,6 +317,188 @@ class TaskSchedulerImplSuite extends SparkFunSuite with LocalSparkContext with B
...
+      // If we had a node-blacklist, then we could tighten the next check to *never*.  But, it also
+      // doesn't particular matter if we the executor check happens in addition.
+      verify(stageToMockTaskSetBlacklist(0), atMost(maxBlacklistChecks))
+        .isExecutorBlacklistedForTaskSet(exec)
+      verify(stageToMockTaskSetBlacklist(0), never)
+        .isExecutorBlacklistedForTask(meq(exec), anyInt())
+    }
+  }
+
+  test("(node, taskset) blacklist prevents per-task checks") {
```
kayousterhout: How about "Blacklisted node for entire task set prevents per-task blacklist checks"? (found the existing version non-obvious in what it meant). And for below maybe "Blacklisted executor for entire task set prevents per-task blacklist checks"?

core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
```diff
@@ -282,6 +317,188 @@ class TaskSchedulerImplSuite extends SparkFunSuite with LocalSparkContext with B
...
+        }
+      }
+      assert(tsm.isZombie)
+    }
+  }
+
+  /**
+   * Helper for performance tests.  Takes the explicitly blacklisted nodes and executors; verifies
+   * that the blacklists are used efficiently to ensure scheduling is not O(numPendingTasks).
+   */
```
kayousterhout: can you add the setup of the test (executor1 and executor2 on host1, executor3 on host2) to this comment (and say that the passed in nodes and executors should be in that list)

core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
```diff
@@ -282,6 +317,188 @@ class TaskSchedulerImplSuite extends SparkFunSuite with LocalSparkContext with B
...
+      execBlacklist.contains(offer.executorId)
+    }.map(_.host).toSet.toSeq
+    val nodesToCheck = nodeBlacklist ++ nodesForBlacklistedExecutors
+    val execsForBlacklistedNodes = offers.filter { offer =>
+      nodeBlacklist.contains(offer.host)
+    }.map(_.executorId).toSeq
+    val executorsToCheck = execBlacklist ++ execsForBlacklistedNodes
+
+    // Schedule a taskset, do a bit of basic sanity checking (that our test is operating the way its
+    // supposed to).
```
kayousterhout: Can you update this comment to say what sanity check is happening here? (it's not obvious from "that our test is operating the way its supposed to"). Also what's the 3 - executorsToCheck.size magic below?

core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
```diff
@@ -282,6 +317,188 @@ class TaskSchedulerImplSuite extends SparkFunSuite with LocalSparkContext with B
...
+      when(stageToMockTaskSetBlacklist(0).isExecutorBlacklistedForTaskSet(exec)).thenReturn(true)
+    }
+    val nodesForBlacklistedExecutors = offers.filter { offer =>
+      execBlacklist.contains(offer.executorId)
+    }.map(_.host).toSet.toSeq
+    val nodesToCheck = nodeBlacklist ++ nodesForBlacklistedExecutors
+    val execsForBlacklistedNodes = offers.filter { offer =>
+      nodeBlacklist.contains(offer.host)
+    }.map(_.executorId).toSeq
+    val executorsToCheck = execBlacklist ++ execsForBlacklistedNodes
```
kayousterhout: This test is pretty complicated, and hard to reason about as a result.  Can you simplify fit by making a resource offer for just one executor/node, which is the executor/node that's blacklisted?  Then the list of executors / nodes to check is small (just 1), and you know exactly the number of times isNodeBlacklistedForTaskSet should be called (# locality levels) because there aren't any ordering issues?

I'm proposing this because it took me a while to read/understand this test, which makes me concerned about maintainability.

core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
```diff
@@ -282,6 +317,188 @@ class TaskSchedulerImplSuite extends SparkFunSuite with LocalSparkContext with B
...
+    val offers = IndexedSeq(
+      new WorkerOffer("executor1", "host1", 1),
+      new WorkerOffer("executor2", "host1", 1),
+      new WorkerOffer("executor3", "host2", 1)
+    )
+    // We should check the node & exec blacklists, but only O(numOffers), not O(numPendingTasks)
+    // times.  Its O(numOffers), not exactly numOffers, because we offer resources multiple
+    // times.  An upper bound on the worst case is -- we check it once for every core at each
+    // locality level.  (We could tighten the bound a bit but that should be a good enough check.)
+    val numCoresOnAllOffers = offers.map(_.cores).sum
```
kayousterhout: Why do you need to consider the number of cores? My understanding is that resourceOfferSingleTaskSet will only be called repeatedly (i.e., once for each core) when tasks are successfully launched, but because this test is checking for blacklisted executors / nodes, resourceOfferSingleTaskSet won't launch a task on the blacklisted nodes.

core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
```diff
@@ -408,4 +625,30 @@ class TaskSchedulerImplSuite extends SparkFunSuite with LocalSparkContext with B
     assert(thirdTaskDescs.size === 0)
     assert(taskScheduler.getExecutorsAliveOnHost("host1") === Some(Set("executor1", "executor3")))
   }
+
+  test("don't update blacklist for shuffle-fetch failures, preemption, denied commits, " +
```
kayousterhout: Should this test be in TaskSetManagerSuite.scala? It seems like it's really testing the TSM (which is where updateBlacklistForFailedTask gets called) and not the TaskScheduler.

kayousterhout (commented):
These tests remain a little painful to read... hopeful this can improve with a few more readability changes!

core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
kayousterhout: Also I know in an earlier comment you said you wanted some successful tasks here too.  I'm in favor of eliminating that because all of these tests have become super integration-test-y rather than unit-test-y, and I think having smaller unit tests is helpful for debugging future issues (and you do have some other more integration-y tests for the blacklist).

core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
squito: In this particular case, we might be able to tighten the bound.  But I was also trying to right this in a way that it didn't involve too many magic constants, if somebody modified it in the future.  Imagine there were 500 cores offered, but after shuffling the offers around, the blacklisted once was first in the list of offers.  Then when you call [`resourceOfferSingleTaskSet`](https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala#L257), you'd need to get through the blacklisted resource before scheduling something.  But since you *do* launch something on the offers later in the shuffled list, you'd [call it again](https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala#L330), and every iteration, you repeat that blacklist check.

this could be improved a lot by either:
(a) pulling the filtering from the taskset blacklist further out of those nested loops, out to [`taskScheduler.resourceOffer`](https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala#L328)
or (b) push that entire `do...while` down into `TaskSetManager`, so that the TaskSetManager is responsible for more assigning more than one task at the time.
(the same logic either way, just a question of code organization.)

Though `O(numOffers)` sounds bad, I decided it was OK because most of the time this is called with one offer (the resource which just freed up from a completed task).  And when you do have a ton of offers, eg. when you start a stage, well scheduling is overall `O(numOffers)` anyhow.

Thinking about this again, pretty sure the bound can actually be `numCores + numLocalityLevels` -- each iteration either schedules something, or it terminates that locality level.

core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
squito: I understand wanting focused tests, but don't you think that skips over the complexity of scheduling when you have some good resources and some bad ones?  eg., the discussion on how many checks are made to the blacklist above?  of course I could add those cases, but that would leave this complexity in place so wouldn't alleviate your concerns.

I can take a stab at updating comments a bit, which might help a bit, but I dunno if it will alleviate your concern.

kayousterhout: For the performance tests though, we're not really testing the performance when tasks *do* get scheduled (i.e., for non-blacklisted locations) right?  If the isBlacklisted calls were getting made too frequently for tasks scheduled on non-blacklisted machines, then it would mean that for some reason we were skipping over perfectly good tasks (and not scheduling them, even though they weren't blacklisted).  That would seem to indicate either (a) a correctness bug with blacklisting or (b) a correctness bug with the other scheduling -- but not a performance bug.  For that reason, for the performance tests, I think it's better to isolate *just* the blacklist case.  Thoughts?

core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
kayousterhout: Ah got it.  I buy the numCores + numLocalityLevels bound.  Overall the bound seems fine; I was just trying to simplify the test logic.

core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
squito: executorsToCheck is all executors that are effectively blacklisted, explicitly or implicitly via the node blacklist.  I'll update the name and add a comment.

core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
```diff
@@ -437,62 +438,76 @@ class TaskSchedulerImplSuite extends SparkFunSuite with LocalSparkContext with B
...
     val execsForBlacklistedNodes = offers.filter { offer =>
       nodeBlacklist.contains(offer.host)
     }.map(_.executorId).toSeq
-    val executorsToCheck = execBlacklist ++ execsForBlacklistedNodes
+    val executorsWithAnyBlacklisting = execBlacklist ++ execsForBlacklistedNodes
 
-    // Schedule a taskset, do a bit of basic sanity checking (that our test is operating the way its
-    // supposed to).
+    // Schedule a taskset, and make sure our test setup is correct -- we are able to schedule
+    // a task on all executors that aren't blacklisted (even the ones implicitly blacklisted by the
```
kayousterhout: does the thing in parens belong here? (seems like no tasks should go on executors implicitly blacklisted)

core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
```diff
@@ -437,62 +438,76 @@ class TaskSchedulerImplSuite extends SparkFunSuite with LocalSparkContext with B
...
+    // node blacklist).
     val firstTaskAttempts = taskScheduler.resourceOffers(offers).flatten
-    assert(firstTaskAttempts.size === 3 - executorsToCheck.size)
-    assert(firstTaskAttempts.size < offers.size)
+    assert(firstTaskAttempts.size === 3 - executorsWithAnyBlacklisting.size)
```
kayousterhout: oh I finally understand the 3 here -- can you do "offers.size" instead? That seems clearer to me (Even though I realize an offer could have multiple cores -- so that only works since cores = 1 for all offers)

core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
```diff
@@ -437,62 +438,76 @@ class TaskSchedulerImplSuite extends SparkFunSuite with LocalSparkContext with B
...
         .isNodeBlacklistedForTask(meq(node), anyInt())
     }
-    executorsToCheck.foreach { exec =>
-      // If we had a node-blacklist, then we could tighten the next check to *never*.  But, it also
-      // doesn't particular matter if we the executor check happens in addition.
+    executorsWithAnyBlacklisting.foreach { exec =>
+      // We should be checking our executor blacklist, but it should be within the bound defined
+      // above.  Its possible that this will be significantly fewer calls, maybe even 0, if there
+      // is also a node-blacklist which takes effect first.  But this assert is all we need to
+      // avoid an O(numPendingTask) slowdown.
```
kayousterhout: Reading this comment and thinking about the randomness made me wonder if this test should be run in a loop where it gets run a bunch of times, to try to force more of the random orderings of offers? (this is assuming you don't like the suggestion to simplify the test to just have the blacklisted executors / nodes)

core/src/test/scala/org/apache/spark/scheduler/TaskSetManagerSuite.scala
```diff
@@ -992,6 +994,46 @@ class TaskSetManagerSuite extends SparkFunSuite with LocalSparkContext with Logg
...
+    when(tsmSpy.taskSetBlacklistHelperOpt).thenReturn(Some(blacklist))
+
+    // make some offers to our taskset, to get tasks we will fail
+    val taskDescs = Seq(
+      "exec1" -> "host1",
+      "exec2" -> "host1"
+    ).flatMap { case (exec, host) =>
+      // offer each executor twice (simulating 2 cores per executor)
+      (0 until 2).flatMap{ _ => tsmSpy.resourceOffer(exec, host, TaskLocality.ANY)}
+    }
```
kayousterhout: add check that this has length 4? (so there are no surprises later)

core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
squito: ok I think I see what you mean now.  I still feel like it would be really easy to end up with different behavior when all the offers were blacklisted vs. only some of them.  To put it another way, you are talking about what happens to the non-blacklisted resources; I'm concerned about what happens to the blacklisted resources, but it could be different in cases where everything is blacklisted, or when only some things are blacklisted.

Anyway, I've made an attempt at making it more clear, though not sure how much it helps.  If you still feel like it would *really* be better to change to have everything blacklisted, I will make that change.

core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
squito: just worded poorly on my part -- I will change parenthetical to "(whether that executors is a explicitly blacklisted, or implicitly blacklisted via the node blacklist)"

core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
squito: ah sorry I misunderstood the confusion earlier.  Yes, that makes sense -- `executorsWithAnyBlacklisting.size` also is expecting 1 core per executor (doesn't seem worth doing a more complex diff of offers and then summing cores ...)

core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
squito: sure, say run them 100 times each?  on my laptop its about 50 milliseconds per iteration.

but just to be clear -- the comment right here has more to do with the interaction between node blacklisting and executor blacklisting, not the randomness of offer shuffling.  The point here is that if an executor is covered by both blacklists, then the executor blacklist won't actually get called at all.  It'll check the node blacklist, and then bail out.  that is even with the worst-case shuffling of the offers.

core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
```diff
@@ -282,6 +316,211 @@ class TaskSchedulerImplSuite extends SparkFunSuite with LocalSparkContext with B
...
+    }
+  }
+
+  /**
+   * Helper for performance tests.  Takes the explicitly blacklisted nodes and executors; verifies
+   * that the blacklists are used efficiently to ensure scheduling is not O(numPendingTasks).
+   * Creates 1 offer on executor[1-3].  Executor1 & 2 are on host1, executor3 is on host2.  Passed
+   * in nodes and executors should be on that list.
+   */
+  private def testBlacklistPerformance(
```
kayousterhout: I should have thought of this last night (sorry) -- but shouldn't this one be in TaskSetManagerSuite.scala too?

core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
```diff
@@ -282,6 +316,211 @@ class TaskSchedulerImplSuite extends SparkFunSuite with LocalSparkContext with B
...
+      testName: String,
+      nodeBlacklist: Seq[String],
+      execBlacklist: Seq[String]): Unit = {
+    // Because scheduling involves shuffling the order of offers around, we run this test a few
+    // times to cover more possibilities.  There are only 3 offers, which means 6 permutations,
+    // so 10 iterations is pretty good.
+    (0 until 10).foreach { testItr =>
+      test(s"$testName: iteration $testItr") {
```
kayousterhout: why name each iteration like this? seems like it clogs up the output for 10 tests that are basically the same (or is this convention used elsewhere? in which case fine to leave it)

core/src/test/scala/org/apache/spark/scheduler/TaskSetManagerSuite.scala
```diff
@@ -992,6 +994,47 @@ class TaskSetManagerSuite extends SparkFunSuite with LocalSparkContext with Logg
     assert(manager3.name === "TaskSet_1.1")
   }
 
+  test("don't update blacklist for shuffle-fetch failures, preemption, denied commits, " +
+    "or killed tasks") {
```
kayousterhout: super nit: I think this should be 2 more spaces indented

kayousterhout (commented):
Commenting is awesome now! A few last small things.

core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
squito: In this case, the main reason is to get `beforeEach` and `afterEach` called between iterations.  We could it ourselves, but seems better to let it happen automatically?  If you are just referring to giving each iteration a unique name, that is because its a requirement of scalatest.

I'm also in the habit of doing this because its useful when trying to flush out flaky tests.  If you put the loop *inside* the test, then the whole test dies at the first exception, so its harder to figure out how often the test fails (I'd rather have it do a full 1k iterations and count how many failed, rather than have it fail at iteration 235).  Also the unit test logs are a little more verbose, but in a good way -- each iteration is clearly labeled, so its a lot easier to jump to the failed iteration and compare to a good iteration.  But that is more for doing it while developing, not necessarily something to check-in.  I dunno if there is a convention here.

The downside is there are 10 entries in the test report, which is perhaps a little more confusing, but its doesn't really seem so bad.  I suppose one problem is that flaky test reports might not work since each time it'll fail on a different iteration.

I don't have strong feelings on this -- I'll push one final change which just pushes the loop inside the test, so its easy for you to choose.

core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
squito: hmm, I don't think so, but I think this comes back to our earlier discussions about whether this is should be more unit-y and less integration-y.  Probably repeating myself here, but while I'm for more narrowly scoped unit tests when possible, I think there is also a need for higher level tests sometimes, and this is one of those cases.

IMO, its important that this test the back-and-forth between TaskScheduler and TaskSetManager,  from `resourceOfferSingleTaskSet` (and even `resourceOffers`) down to `taskSet.resourceOffer`.  We're really trying to make sure we control the performance of the way those work together.  That interaction makes the test more complicated, but that is also why I want to test it.

I don't think this is most important thing in the world if you feel really strongly about it, so just lemme know and we can move on to other things.

squito:
The final commit (https://github.com/apache/spark/pull/15644/commits/4a91e5fc0a64cfe740530fbeddf3722d2006bd10)  just changes whether the iterations are inside or outside of the test so you can easily go either way.

core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
kayousterhout: Ah got it that makes sense re:before / after.  Either way seems fine here -- so leave whichever you prefer!

core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
kayousterhout: Ok sounds good!

core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
squito: I'm honestly a little torn b/c of the flaky test reporting, but I think really I'd rather have the the before /after stuff work normally so it doesn't get messed up later on.

kayousterhout:
LGTM

squito:
thanks @kayousterhout , merged to master

PR #15652, opened 2016-10-27
[SPARK-16987] [None] Add spark-default.conf property to define https port for spark history server

chie8842:
I created a pull request, but there is a problem.
Port should be separated for each particular process such as history server and spark ui.
But user can configure spark.ssl.port with this code and when configure it, each process will try to create connections with same port and it doesn't work good.
Because of structure of existing function, it has impact to other functions to solve the problem.
So there are three answer about it.
1. Use this fix code so user can configure a specific secure port, but spark.ssl.port doesn't work good.
2. Don't change and close issue.
3. We should be able to configure a specific secure port completely, so drastically change many functions.
Please comments.

### DIFF
--- a/core/src/main/scala/org/apache/spark/ui/JettyUtils.scala
+++ b/core/src/main/scala/org/apache/spark/ui/JettyUtils.scala
@@ -307,15 +307,18 @@ private[spark] object JettyUtils extends Logging {
       connectors += httpConnector
 
       sslOptions.createJettySslContextFactory().foreach { factory =>
-        // If the new port wraps around, do not try a privileged port.
 
-        require(sslOptions.port == 0 || (1024 <= sslOptions.port && sslOptions.port < 65536))
+        require(sslOptions.port == 0 || (1024 <= sslOptions.port && sslOptions.port < 65536),
+          "securePort should be between 1024 and 65535 (inclusive)," +
+            " or 0 for determined automatically.")
 
         val securePort =
           if (currentPort != 0) {
-            if (1024 < sslOptions.port && sslOptions.port < 65536) {
+            // If the new port wraps around, do not try a privileged port.
+            if (1024 <= sslOptions.port && sslOptions.port < 65536) {
               sslOptions.port
             } else {
+              // If the new port wraps around, do not try a privilege port
               (currentPort + 400 - 1024) % (65536 - 1024) + 1024
             }
           } else {
--- a/core/src/main/scala/org/apache/spark/ui/JettyUtils.scala
+++ b/core/src/main/scala/org/apache/spark/ui/JettyUtils.scala
@@ -284,7 +284,7 @@ private[spark] object JettyUtils extends Logging {
     }
 
     // Bind to the given port, or throw a java.net.BindException if the port is occupied
-    def connect(currentPort: Int): (Server, Int) = {
+    def connect(currentPort: Int, securePort: Int = sslOptions.port): (Server, Int) = {
       val pool = new QueuedThreadPool
       if (serverName.nonEmpty) {
         pool.setName(serverName)
--- a/core/src/main/scala/org/apache/spark/ui/JettyUtils.scala
+++ b/core/src/main/scala/org/apache/spark/ui/JettyUtils.scala
@@ -308,21 +308,40 @@ private[spark] object JettyUtils extends Logging {
 
       sslOptions.createJettySslContextFactory().foreach { factory =>
         // If the new port wraps around, do not try a privileged port.
-        val securePort =
-          if (currentPort != 0) {
-            (currentPort + 400 - 1024) % (65536 - 1024) + 1024
-          } else {
-            0
-          }
+
+        require(sslOptions.port == 0 || (1024 <= sslOptions.port && sslOptions.port < 65536))
+        val maxRetries = Utils.portMaxRetries(conf)
+
         val scheme = "https"
         // Create a connector on port securePort to listen for HTTPS requests
-        val connector = new ServerConnector(server, factory)
-        connector.setPort(securePort)
-
-        connectors += connector
-
-        // redirect the HTTP requests to HTTPS port
-        collection.addHandler(createRedirectHttpsHandler(securePort, scheme))
+        var offset = 0
+        var connected = false
+        while (offset < maxRetries && connected == false) {
+          val securePort =
+            if (currentPort != 0) {
+              if (1024 < sslOptions.port && sslOptions.port < 65536) {
+                sslOptions.port + offset
+              } else {
+                (currentPort + 400 - 1024) % (65536 - 1024) + 1024
+              }
+            } else {
+              0
+            }
+          val connector = new ServerConnector(server, factory)
+          connector.setPort(securePort)
+          connectors += connector
+          try{
+            // redirect the HTTP requests to HTTPS port
+            collection.addHandler(createRedirectHttpsHandler(securePort, scheme))
+            connected = true
+          } catch {
+            case e: Exception =>
+              if (offset >= maxRetries) {
+                throw e
+              }
+          }
+          offset += 1
+        }
       }
 
       gzipHandlers.foreach(collection.addHandler)
--- a/core/src/main/scala/org/apache/spark/ui/JettyUtils.scala
+++ b/core/src/main/scala/org/apache/spark/ui/JettyUtils.scala
@@ -314,12 +314,12 @@ private[spark] object JettyUtils extends Logging {
 
         val securePort =
           if (currentPort != 0) {
-            // If the new port wraps around, do not try a privileged port.
-            if (1024 <= sslOptions.port && sslOptions.port < 65536) {
-              sslOptions.port
-            } else {
-              // If the new port wraps around, do not try a privilege port
+            if (sslOptions.port == 0) {
+              // If the new port wraps around, do not try a privileged port
               (currentPort + 400 - 1024) % (65536 - 1024) + 1024
+            } else {
+              // use sslOptions.port value as securePort
+              sslOptions.port
             }
           } else {
             0
--- a/core/src/main/scala/org/apache/spark/ui/JettyUtils.scala
+++ b/core/src/main/scala/org/apache/spark/ui/JettyUtils.scala
@@ -310,38 +310,27 @@ private[spark] object JettyUtils extends Logging {
         // If the new port wraps around, do not try a privileged port.
 
         require(sslOptions.port == 0 || (1024 <= sslOptions.port && sslOptions.port < 65536))
-        val maxRetries = Utils.portMaxRetries(conf)
 
-        val scheme = "https"
-        // Create a connector on port securePort to listen for HTTPS requests
-        var offset = 0
-        var connected = false
-        while (offset < maxRetries && connected == false) {
-          val securePort =
-            if (currentPort != 0) {
-              if (1024 < sslOptions.port && sslOptions.port < 65536) {
-                sslOptions.port + offset
-              } else {
-                (currentPort + 400 - 1024) % (65536 - 1024) + 1024
-              }
+        val securePort =
+          if (currentPort != 0) {
+            if (1024 < sslOptions.port && sslOptions.port < 65536) {
+              sslOptions.port
             } else {
-              0
+              (currentPort + 400 - 1024) % (65536 - 1024) + 1024
             }
-          val connector = new ServerConnector(server, factory)
-          connector.setPort(securePort)
-          connectors += connector
-          try{
-            // redirect the HTTP requests to HTTPS port
-            collection.addHandler(createRedirectHttpsHandler(securePort, scheme))
-            connected = true
-          } catch {
-            case e: Exception =>
-              if (offset >= maxRetries) {
-                throw e
-              }
+          } else {
+            0
           }
-          offset += 1
-        }
+        val scheme = "https"
+        // Create a connector on port securePort to listen for HTTPS requestswork.
+
+        val connector = new ServerConnector(server, factory)
+        connector.setPort(securePort)
+
+        connectors += connector
+
+        // redirect the HTTP requests to HTTPS port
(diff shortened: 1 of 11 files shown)


core/src/main/scala/org/apache/spark/ui/WebUI.scala
```diff
@@ -147,7 +147,10 @@ private[spark] abstract class WebUI(
   }
 
   /** Return the url of web interface. Only valid after bind(). */
-  def webUrl: String = s"http://$publicHostName:$boundPort"
+  def webUrl: String = {
```
srowen: These seem unrelated; aren't these in master already? can you rebase?

docs/configuration.md
```diff
@@ -1664,6 +1664,15 @@ Apart from these, the following properties are also available, and may be useful
         </td>
     </tr>
     <tr>
+        <td><code>spark.ssl.<particular protocol>.port</code></td>
+        <td>0</td>
+        <td>
+            A port number hen connecting with SSL. Default value 0 means to be determined automatically.
```
srowen: There are a few typos here, and I don't think you need to repeat the point about multiple services. I'd suggest changing the prop name above to `spark.ssl.port` and then:

Port number to listen on for SSL connections. Default value of 0 means the port will be determined automatically.

core/src/main/scala/org/apache/spark/ui/JettyUtils.scala
```diff
@@ -308,14 +308,22 @@ private[spark] object JettyUtils extends Logging {
...
+              sslOptions.port
+            } else {
+              (currentPort + 400 - 1024) % (65536 - 1024) + 1024
+            }
           } else {
             0
           }
         val scheme = "https"
-        // Create a connector on port securePort to listen for HTTPS requests
+        // Create a connector on port securePort to listen for HTTPS requestswork.
```
srowen: I think this was an unintended change

srowen (changes requested):

core/src/main/scala/org/apache/spark/ui/WebUI.scala
hayashidac: Sorry, I rebased this change.

docs/configuration.md
hayashidac: Thank you. I fixed the description.

core/src/main/scala/org/apache/spark/ui/JettyUtils.scala
hayashidac: Sorry, I rebased this change.

hayashidac:
Thank you very much. I fixed for your comment. Please check again.

docs/configuration.md
```diff
@@ -1664,6 +1664,14 @@ Apart from these, the following properties are also available, and may be useful
         </td>
     </tr>
     <tr>
+        <td><code>spark.ssl.<particular protocol>.port</code></td>
```
tejasapatil: What does `<particular protocol>` mean here ? Is this config name parameterized ?

srowen: Agree; see https://github.com/apache/spark/pull/15652#discussion_r85307232  This should just refer to `spark.ssl.port` because the rest is documented above.

docs/configuration.md
```diff
@@ -1598,7 +1598,7 @@ Apart from these, the following properties are also available, and may be useful
         <td>false</td>
         <td>
             <p>Whether to enable SSL connections on all supported protocols.</p>
-
+
```
srowen: If you can, remove this change

docs/configuration.md
hayashidac: It has a reason for this.
If an user set spark.ssl.port, the port of history server, SparkUI, MasterWebUI, and WorkerWebUI will collapse as I mentioned at the beginning of this PR.
It is because WebUI.scala is inheried by HistoryServer.scala, SparkUI, MasterWebUI, and WorkerWebUI.
How do you think about it?


> I created a pull request, but there is a problem.
> Port should be separated for each particular process such as history server and spark 
> ui.
> But user can configure spark.ssl.port with this code and when configure it, each 
> process will try to create connections with same port and it doesn't work good.
> Because of structure of existing function, it has impact to other functions to solve the
> problem.
> So there are three answer about it.
>
> Use this fix code so user can configure a specific secure port, but spark.ssl.port doesn't > work good.
> Don't change and close issue.
> We should be able to configure a specific secure port completely, so drastically change > many functions.
> Please comments.

srowen: Yes, setting `spark.ssl.port` sets the SSL port for all services. That's how it's documented to work above, and you can override with `spark.ui.ssl.port`, etc. I'm just saying that this is already documented, and to be consistent with other props, the config key in this section can be `spark.ssl.port` -- or did I misunderstand?

srowen: Yes that's what I'm referring to. I would still use `spark.ssl.port` as the key for this section for consistency. It's OK to maybe add some text reminding the reader that this port can and probably should be set separately for each service.

hayashidac: OK, I understood. I refered about attention, and changed the key to spark.ssl.port.

core/src/main/scala/org/apache/spark/ui/JettyUtils.scala
```diff
@@ -308,14 +308,22 @@ private[spark] object JettyUtils extends Logging {
 
       sslOptions.createJettySslContextFactory().foreach { factory =>
         // If the new port wraps around, do not try a privileged port.
+
+        require(sslOptions.port == 0 || (1024 <= sslOptions.port && sslOptions.port < 65536))
+
         val securePort =
           if (currentPort != 0) {
-            (currentPort + 400 - 1024) % (65536 - 1024) + 1024
+            if (1024 < sslOptions.port && sslOptions.port < 65536) {
```
srowen: This condition should only happen when the port is 0 right? given the `require` above. It would be clearer to check against 0 only.

docs/configuration.md
```diff
@@ -1664,6 +1664,15 @@ Apart from these, the following properties are also available, and may be useful
         </td>
     </tr>
     <tr>
+        <td><code>spark.ssl.port</code></td>
+        <td>0</td>
+        <td>
+            Port number to listen on for SSL connections. 
+            Default value of 0 means the port will be determined automatically.
+            Attention that the port should be separated for each particular protocols.
```
srowen: I think it would be clearer to write something like "Note that the SSL port for individual services can be overridden by setting values like `spark.ssl.YYY.port` as described above."

srowen (changes requested):

core/src/main/scala/org/apache/spark/ui/JettyUtils.scala
```diff
@@ -308,14 +308,22 @@ private[spark] object JettyUtils extends Logging {
...
+              sslOptions.port
+            } else {
+              (currentPort + 400 - 1024) % (65536 - 1024) + 1024
```
tejasapatil: Can you add a comment in the code explaining the math done here ? It will help readability of the code.

docs/configuration.md
tejasapatil: In the code above you are ensuring that the SSL port is always between 1024 and 65536. You could document that here as "allowed range of values".

hayashidac:
I reflected some comments.
And I handled about trivial issues in ml library. Is it OK to contain them in this PR?

mllib/src/main/scala/org/apache/spark/ml/feature/VectorIndexer.scala
```diff
@@ -66,7 +66,7 @@ private[ml] trait VectorIndexerParams extends Params with HasInputCol with HasOu
  *     - This helps process a dataset of unknown vectors into a dataset with some continuous
  *       features and some categorical features. The choice between continuous and categorical
  *       is based upon a maxCategories parameter.
- *     - Set maxCategories to the maximum number of categorical any categorical feature should have.
+ *     - Set maxCategories to the maximum number of categories which categorical feature should have.
```
srowen: This doesn't look related? it should also be more like "number of categories that any categorical ..."

mllib/src/main/scala/org/apache/spark/ml/tree/treeParams.scala
```diff
@@ -85,8 +85,10 @@ private[ml] trait DecisionTreeParams extends PredictorParams
    * (default = 256 MB)
    * @group expertParam
    */
-  final val maxMemoryInMB: IntParam = new IntParam(this, "maxMemoryInMB",
-    "Maximum memory in MB allocated to histogram aggregation.",
+  final val maxMemoryInMB: IntParam = new IntParam(this, "maxMemoryInMB", "Maximum memory in MB" +
```
srowen: Likewise, looks unrelated

docs/configuration.md
```diff
@@ -1664,6 +1664,16 @@ Apart from these, the following properties are also available, and may be useful
         </td>
     </tr>
     <tr>
+        <td><code>spark.ssl.port</code></td>
+        <td>0</td>
+        <td>
+            Port number to listen on for SSL connections.
+            The SSL port should be between 1024 and 65535 (inclusive). 
+            Default value of 0 means the port will be determined automatically.
+            Attention that the port should be separated for each particular protocols.
```
srowen: I don't think it's clear what the last sentence means. I'd either remove it or say something like "the port can be specified for services individually, with properties like `spark.ssl.YYYY.port`, as described above."

core/src/main/scala/org/apache/spark/ui/JettyUtils.scala
```diff
@@ -307,15 +307,26 @@ private[spark] object JettyUtils extends Logging {
...
-        // If the new port wraps around, do not try a privileged port.
+
+        require(sslOptions.port == 0 || (1024 <= sslOptions.port && sslOptions.port < 65536),
+          "securePort should be between 1024 and 65535 (inclusive)," +
+            " or 0 for determined automatically.")
+
         val securePort =
           if (currentPort != 0) {
-            (currentPort + 400 - 1024) % (65536 - 1024) + 1024
+            // If the new port wraps around, do not try a privileged port.
```
srowen: This comment is duplicated and not correct here

core/src/main/scala/org/apache/spark/ui/JettyUtils.scala
```diff
@@ -307,15 +307,26 @@ private[spark] object JettyUtils extends Logging {
...
+            if (1024 <= sslOptions.port && sslOptions.port < 65536) {
```
srowen: To me it seems easier to flip the if condition, and just check for the port being 0, rather than repeat this condition.

srowen (changes requested):

mllib/src/main/scala/org/apache/spark/ml/tree/treeParams.scala
hayashidac: It isn't related to SPARK-16987 as you say. But I think the documentation is not sufficient, and I'd like to fix it. I think it's very trivial so we can handle at this PR.
Is it better to create another issue at jira for this?
And treeParams.scala's comment is same.

srowen:
OK, I wouldn't mind getting @vanzin to glance at this as it involves SSL.

(Can you also replace [None] with [WEBUI] in the title?)

Looks fairly straightforward to me.

core/src/main/scala/org/apache/spark/ui/JettyUtils.scala
```diff
@@ -307,15 +307,26 @@ private[spark] object JettyUtils extends Logging {
...
-        // If the new port wraps around, do not try a privileged port.
+
+        require(sslOptions.port == 0 || (1024 <= sslOptions.port && sslOptions.port < 65536),
+          "securePort should be between 1024 and 65535 (inclusive)," +
+            " or 0 for determined automatically.")
+
         val securePort =
           if (currentPort != 0) {
-            (currentPort + 400 - 1024) % (65536 - 1024) + 1024
+            if (sslOptions.port == 0) {
```
hayashidac: Thank you for comment. Yes, as you say,this logic will occur some conflicts.
But as I refered at the top of PR comment, `startServiceOnPort` is called from many unrelated methods.
So I conscious that this PR  affects many unrelated components.

vanzin: It is not unrelated. As I described, you code will break things in the normal web UI if you set that option.

hayashidac: I know that some methods which call `startServiceOnPort` is related, and for example, `NettyBlockTransferService.init` or `RestSubmissionServer.this` is unrelated, I think.
But I'm trying to fix according to your comment. I'm in progress so please wait...

core/src/main/scala/org/apache/spark/util/Utils.scala
```diff
@@ -2142,9 +2142,10 @@ private[spark] object Utils extends Logging {
    */
   def startServiceOnPort[T](
       startPort: Int,
-      startService: Int => (T, Int),
+      startService: (Int, Int) => (T, Int),
       conf: SparkConf,
-      serviceName: String = ""): (T, Int) = {
+      serviceName: String = "",
+      securePort: Int = 0): (T, Int) = {
```
vanzin: I don't like what you did here. As you yourself explained, not all consumers of this API have the concept of separate secure and non-secure ports. And as I explained in my previous comments, it's possible to fix the problem you introduced without changing this API.

vanzin:
Aside from me not liking the API change you're introducing, you need to add a unit test to make sure the feature is working. I suggest taking the scenario I described in a previous comment (https://github.com/apache/spark/pull/15652#discussion_r88137680) and turning it into a unit test.

HyukjinKwon:
Hi @chie8842, is it still active?

chie8842:
@HyukjinKwon I'm sorry, I haven't reply for a long time.
I'll check it.

vanzin:
I added this with #16625, this PR is not needed anymore. Please close.

PR #15653, opened 2016-10-27
[SPARK-18130][SQL] Don't add inferred redundant isnotnull condition from constraints

viirya:

In `Optimizer`, we will infer additional conditions from `Filter`'s constraints which include `IsNotNull`. Some inferred `IsNotNull`s are useful because they can be short circuit for other predicates which refer the same attributes. But for an `IsNotNull` condition, we also infer other `IsNotNull` from it. Those `IsNotNull`s are redundant. We don't need to include them.

E.g., for a condition like `IsNotNull(a + Rand())`, an additional `IsNotNull(a)` will be added into `Filter`'s condition. So we need to evaluate `IsNotNull(a + Rand()) && IsNotNull(a)`. Actually we don't need `IsNotNull(a)` because it is guaranteed by the original `IsNotNull(a + Rand())`.
## How was this patch tested?

Jenkins tests.


viirya:
This is not a serious issue. And it doesn't affect correctness. I would left this to @cloud-fan or @hvanhovell decide if we want to fix it or not.

viirya:
Close it for now since no interest from committers.

PR #15658, opened 2016-10-27
[SPARK-13770] [Documentation][ML] Document the ML feature Interaction

chie8842:
I created Scala and Java example and added documentation.


docs/ml-features.md
```diff
@@ -729,6 +729,28 @@ for more details on the API.
 </div>
 </div>
 
+## Interaction
+
+`Implements` is a `Transformer` which implements interaction transform.
```
srowen: "Interaction" is a Transformer ...

I think this examples is valuable but it actually says nothing about what the transformer does. I know it outputs the various interaction features but we should define and give an example here and/or in the comments of the examples.

hayashidac: Thank you for your comment. I added example and changed description.

srowen:
Ping @hayashidac I'd be happy to work with you to get a more complete example into the docs for this operation. It could use them, yes.

hayashidac:
@srowen sorry, I have some other task and will work on this tomorrow.

hayashidac:
I changed input column format. Please check again.

srowen:
Yeah I see what you've done, but the example still doesn't suggest what the output is. I think it's OK if the example has to be run to show the output, but, I think the documentation needs to exhibit a more representative input and output so that people understand what it does without having to run the example.

hayashidac:
I reflected the contents of the indications. Please confirm it.

docs/ml-features.md
```diff
@@ -729,6 +729,58 @@ for more details on the API.
 </div>
 </div>
 
+## Interaction
+
+`Interaction` is a `Transformer` which takes a vector/double columns, and generate a single vector column that contains multiplication results of all combination of each vector/double values.
```
srowen: Two small changes: "... which takes vector or double-valued columns, and generates a ..." 
"... contains the product of all combinations of one value from each input column"

docs/ml-features.md
```diff
@@ -729,6 +729,58 @@ for more details on the API.
...
+
+For example, if you have two vector type columns each of which contains three double type values as input columns, then you'll get a vector with 9 double type values as the output column.
```
srowen: Suggest: "... each of which has 3 dimensions ..."  "... then you'll get a 9-dimensional vector ..."

srowen (changes requested):
Looks good with a few edits to the text.

docs/ml-features.md
hayashidac: Thank you for your check. I fixed the document for your comment.

srowen (approved):

srowen:
Merged to master/2.1

PR #15664, opened 2016-10-27
[SPARK-18123][SQL] Use db column names instead of RDD column ones during JDBC Writing

dongjoon-hyun:

Apache Spark supports the following cases **by quoting RDD column names** while saving through JDBC.
- Allow reserved keyword as a column name, e.g., 'order'.
- Allow mixed-case colume names like the following, e.g., `[a: int, A: int]`.
  
  ``` scala
  scala> val df = sql("select 1 a, 1 A")
  df: org.apache.spark.sql.DataFrame = [a: int, A: int]
  ...
  scala> df.write.mode("overwrite").format("jdbc").options(option).save()
  scala> df.write.mode("append").format("jdbc").options(option).save()
  ```

This PR aims to use **database column names** instead of RDD column ones in order to support the following additionally.
Note that this case succeeds with `MySQL`, but fails on `Postgres`/`Oracle` before.

``` scala
val df1 = sql("select 1 a")
val df2 = sql("select 1 A")
...
df1.write.mode("overwrite").format("jdbc").options(option).save()
df2.write.mode("append").format("jdbc").options(option).save()
```
## How was this patch tested?

Pass the Jenkins test with a new testcase.

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcRelationProvider.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcRelationProvider.scala
@@ -60,7 +60,7 @@ class JdbcRelationProvider extends CreatableRelationProvider
 
     val conn = JdbcUtils.createConnectionFactory(jdbcOptions)()
     try {
-      val tableSchema = JdbcUtils.getSchema(conn, url, table)
+      val tableSchema = JdbcUtils.getSchemaOption(conn, url, table)
       val tableExists = tableSchema.isDefined
       val isCaseSensitive = sqlContext.conf.caseSensitiveAnalysis
       if (tableExists) {
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcRelationProvider.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcRelationProvider.scala
@@ -62,26 +62,23 @@ class JdbcRelationProvider extends CreatableRelationProvider
     try {
       val tableSchema = JdbcUtils.getSchema(conn, url, table)
       val tableExists = tableSchema.isDefined
-      val caseSensitive = sqlContext.conf.caseSensitiveAnalysis
+      val isCaseSensitive = sqlContext.conf.caseSensitiveAnalysis
       if (tableExists) {
         mode match {
           case SaveMode.Overwrite =>
-            val normalizedSchema = if (isTruncate && isCascadingTruncateTable(url) == Some(false)) {
+            if (isTruncate && isCascadingTruncateTable(url) == Some(false)) {
               // In this case, we should truncate table and then load.
               truncateTable(conn, table)
-              JdbcUtils.normalizeSchema(df.schema, tableSchema.get, caseSensitive)
+              saveTable(df, url, table, tableSchema.get, isCaseSensitive, jdbcOptions)
             } else {
               // Otherwise, do not truncate the table, instead drop and recreate it
               dropTable(conn, table)
               createTable(df.schema, url, table, createTableOptions, conn)
-              df.schema
+              saveTable(df, url, table, df.schema, isCaseSensitive, jdbcOptions)
             }
-            saveTable(df, url, table, normalizedSchema, jdbcOptions)
 
           case SaveMode.Append =>
-            val normalizedSchema =
-              JdbcUtils.normalizeSchema(df.schema, tableSchema.get, caseSensitive)
-            saveTable(df, url, table, normalizedSchema, jdbcOptions)
+            saveTable(df, url, table, tableSchema.get, isCaseSensitive, jdbcOptions)
 
           case SaveMode.ErrorIfExists =>
             throw new AnalysisException(
@@ -94,7 +91,7 @@ class JdbcRelationProvider extends CreatableRelationProvider
         }
       } else {
         createTable(df.schema, url, table, createTableOptions, conn)
-        saveTable(df, url, table, df.schema, jdbcOptions)
+        saveTable(df, url, table, df.schema, isCaseSensitive, jdbcOptions)
       }
     } finally {
       conn.close()
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcRelationProvider.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcRelationProvider.scala
@@ -60,23 +60,27 @@ class JdbcRelationProvider extends CreatableRelationProvider
 
     val conn = JdbcUtils.createConnectionFactory(jdbcOptions)()
     try {
-      val tableExists = JdbcUtils.tableExists(conn, url, table)
+      val tableSchema = JdbcUtils.getSchema(conn, url, table)
+      val tableExists = tableSchema.isDefined
+      val caseSensitive = sqlContext.conf.caseSensitiveAnalysis
       if (tableExists) {
         mode match {
           case SaveMode.Overwrite =>
-            if (isTruncate && isCascadingTruncateTable(url) == Some(false)) {
+            val savingSchema = if (isTruncate && isCascadingTruncateTable(url) == Some(false)) {
               // In this case, we should truncate table and then load.
               truncateTable(conn, table)
-              saveTable(df, url, table, jdbcOptions)
+              JdbcUtils.getSavingSchema(df.schema, tableSchema.get, caseSensitive)
             } else {
               // Otherwise, do not truncate the table, instead drop and recreate it
               dropTable(conn, table)
               createTable(df.schema, url, table, createTableOptions, conn)
-              saveTable(df, url, table, jdbcOptions)
+              df.schema
             }
+            saveTable(df, url, table, savingSchema, jdbcOptions)
 
           case SaveMode.Append =>
-            saveTable(df, url, table, jdbcOptions)
+            val savingSchema = JdbcUtils.getSavingSchema(df.schema, tableSchema.get, caseSensitive)
+            saveTable(df, url, table, savingSchema, jdbcOptions)
 
           case SaveMode.ErrorIfExists =>
             throw new AnalysisException(
@@ -89,7 +93,7 @@ class JdbcRelationProvider extends CreatableRelationProvider
         }
       } else {
         createTable(df.schema, url, table, createTableOptions, conn)
-        saveTable(df, url, table, jdbcOptions)
+        saveTable(df, url, table, df.schema, jdbcOptions)
       }
     } finally {
       conn.close()
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcRelationProvider.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcRelationProvider.scala
@@ -57,13 +57,13 @@ class JdbcRelationProvider extends CreatableRelationProvider
     val table = jdbcOptions.table
     val createTableOptions = jdbcOptions.createTableOptions
     val isTruncate = jdbcOptions.isTruncate
+    val isCaseSensitive = sqlContext.conf.caseSensitiveAnalysis
 
     val conn = JdbcUtils.createConnectionFactory(jdbcOptions)()
     try {
-      val tableSchema = JdbcUtils.getSchemaOption(conn, url, table)
-      val tableExists = tableSchema.isDefined
-      val isCaseSensitive = sqlContext.conf.caseSensitiveAnalysis
+      val tableExists = JdbcUtils.tableExists(conn, url, table)
       if (tableExists) {
+        val tableSchema = JdbcUtils.getSchemaOption(conn, url, table)
         mode match {
           case SaveMode.Overwrite =>
             if (isTruncate && isCascadingTruncateTable(url) == Some(false)) {
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcRelationProvider.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcRelationProvider.scala
@@ -66,21 +66,22 @@ class JdbcRelationProvider extends CreatableRelationProvider
       if (tableExists) {
         mode match {
           case SaveMode.Overwrite =>
-            val savingSchema = if (isTruncate && isCascadingTruncateTable(url) == Some(false)) {
+            val normalizedSchema = if (isTruncate && isCascadingTruncateTable(url) == Some(false)) {
               // In this case, we should truncate table and then load.
               truncateTable(conn, table)
-              JdbcUtils.getSavingSchema(df.schema, tableSchema.get, caseSensitive)
+              JdbcUtils.normalizeSchema(df.schema, tableSchema.get, caseSensitive)
             } else {
               // Otherwise, do not truncate the table, instead drop and recreate it
               dropTable(conn, table)
               createTable(df.schema, url, table, createTableOptions, conn)
               df.schema
             }
-            saveTable(df, url, table, savingSchema, jdbcOptions)
+            saveTable(df, url, table, normalizedSchema, jdbcOptions)
 
           case SaveMode.Append =>
-            val savingSchema = JdbcUtils.getSavingSchema(df.schema, tableSchema.get, caseSensitive)
-            saveTable(df, url, table, savingSchema, jdbcOptions)
+            val normalizedSchema =
+              JdbcUtils.normalizeSchema(df.schema, tableSchema.get, caseSensitive)
+            saveTable(df, url, table, normalizedSchema, jdbcOptions)
 
           case SaveMode.ErrorIfExists =>
             throw new AnalysisException(
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcRelationProvider.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcRelationProvider.scala
@@ -63,22 +63,23 @@ class JdbcRelationProvider extends CreatableRelationProvider
     try {
       val tableExists = JdbcUtils.tableExists(conn, url, table)
       if (tableExists) {
-        val tableSchema = JdbcUtils.getSchemaOption(conn, url, table)
         mode match {
           case SaveMode.Overwrite =>
             if (isTruncate && isCascadingTruncateTable(url) == Some(false)) {
               // In this case, we should truncate table and then load.
               truncateTable(conn, table)
-              saveTable(df, url, table, tableSchema.get, isCaseSensitive, jdbcOptions)
+              val tableSchema = JdbcUtils.getSchemaOption(conn, url, table)
+              saveTable(df, url, table, tableSchema, isCaseSensitive, jdbcOptions)
             } else {
               // Otherwise, do not truncate the table, instead drop and recreate it
               dropTable(conn, table)
               createTable(df.schema, url, table, createTableOptions, conn)
-              saveTable(df, url, table, df.schema, isCaseSensitive, jdbcOptions)
+              saveTable(df, url, table, Some(df.schema), isCaseSensitive, jdbcOptions)
             }
 
           case SaveMode.Append =>
-            saveTable(df, url, table, tableSchema.get, isCaseSensitive, jdbcOptions)
+            val tableSchema = JdbcUtils.getSchemaOption(conn, url, table)
+            saveTable(df, url, table, tableSchema, isCaseSensitive, jdbcOptions)
 
           case SaveMode.ErrorIfExists =>
             throw new AnalysisException(
@@ -91,7 +92,7 @@ class JdbcRelationProvider extends CreatableRelationProvider
         }
       } else {
         createTable(df.schema, url, table, createTableOptions, conn)
-        saveTable(df, url, table, df.schema, isCaseSensitive, jdbcOptions)
+        saveTable(df, url, table, Some(df.schema), isCaseSensitive, jdbcOptions)
       }
     } finally {
       conn.close()
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
@@ -123,7 +123,7 @@ object JdbcUtils extends Logging {
         if (nameMap.isDefinedAt(x.name)) {
           dialect.quoteIdentifier(x.name)
         } else if (lowercaseNameMap.isDefinedAt(x.name.toLowerCase)) {
-          dialect.quoteIdentifier(nameMap(x.name.toLowerCase))
+          dialect.quoteIdentifier(lowercaseNameMap(x.name.toLowerCase))
         } else {
           throw new SQLException(s"""Column "${x.name}" not found""")
         }
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
@@ -20,13 +20,13 @@ package org.apache.spark.sql.execution.datasources.jdbc
 import java.sql.{Connection, Driver, DriverManager, PreparedStatement, ResultSet, ResultSetMetaData, SQLException}
 
 import scala.collection.JavaConverters._
-import scala.util.{Failure, Success, Try}
+import scala.util.Try
 import scala.util.control.NonFatal
 
 import org.apache.spark.TaskContext
 import org.apache.spark.executor.InputMetrics
 import org.apache.spark.internal.Logging
-import org.apache.spark.sql.{DataFrame, Row}
+import org.apache.spark.sql.{AnalysisException, DataFrame, Row}
 import org.apache.spark.sql.catalyst.InternalRow
 import org.apache.spark.sql.catalyst.encoders.RowEncoder
 import org.apache.spark.sql.catalyst.expressions.SpecificInternalRow
@@ -216,25 +216,28 @@ object JdbcUtils extends Logging {
   def getSchema(conn: Connection, url: String, table: String): Option[StructType] = {
     val dialect = JdbcDialects.get(url)
 
-    Try {
+    try {
       val statement = conn.prepareStatement(dialect.getSchemaQuery(table))
       try {
-        getSchema(statement.executeQuery(), dialect)
+        Some(getSchema(statement.executeQuery(), dialect))
+      } catch {
+        case _: SQLException => None
       } finally {
         statement.close()
       }
-    } match {
-      case Success(v) =>
-        Some(v)
-      case Failure(e) =>
-        None
+    } catch {
+      case _: SQLException => None
     }
   }
 
   /**
-   * Returns the saving schema using rddSchema's sequence and tableSchema's name.
+   * Returns a schema using rddSchema's column sequence and tableSchema's column names.
+   *
+   * When appending data into some case-sensitive DBMSs like PostgreSQL/Oracle, we need to respect
+   * the existing case-sensitive column names instead of RDD column names for user convenience.
+   * See SPARK-18123 for more details.
    */
-  def getSavingSchema(
+  def normalizeSchema(
       rddSchema: StructType,
       tableSchema: StructType,
       caseSensitive: Boolean): StructType = {
@@ -247,10 +250,10 @@ object JdbcUtils extends Logging {
         // identical names
         schema = schema.add(nameMap(f.name))
       } else if (!caseSensitive && lowercaseNameMap.isDefinedAt(f.name.toLowerCase)) {
-        // case-insensitive identical names
+        // identical names in a case-insensitive way
         schema = schema.add(lowercaseNameMap(f.name.toLowerCase))
       } else {
-        throw new org.apache.spark.SparkException(s"""Column "${f.name}" not found""")
+        throw new AnalysisException(s"""Column "${f.name}" not found""")
       }
     }
     schema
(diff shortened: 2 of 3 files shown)


sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
```diff
@@ -112,7 +112,25 @@ object JdbcUtils extends Logging {
    */
   def insertStatement(conn: Connection, table: String, rddSchema: StructType, dialect: JdbcDialect)
       : PreparedStatement = {
```
dongjoon-hyun: Actually, this is an approach similar to `normalizePartitionSpec` in `PartitioningUtils.scala`.

https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/PartitioningUtils.scala#L248-L251

dongjoon-hyun:
Retest this please.

dongjoon-hyun:
Two failures seems to be irrelevant. 
```
[info] - randomized aggregation test - [typed, with partial + safe] - with grouping keys - with non-empty input *** FAILED *** (1 second, 367 milliseconds)
```

dongjoon-hyun:
Retest this please.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
```diff
@@ -112,7 +112,25 @@ object JdbcUtils extends Logging {
...
+    var columns: String = ""
+    try {
+      val tableSchema = getSchema(tableSchemaQuery.executeQuery(), dialect)
+      val nameMap = tableSchema.fields.map(f => f.name -> f.name).toMap
+      val lowercaseNameMap = tableSchema.fields.map(f => f.name.toLowerCase -> f.name).toMap
+      columns = rddSchema.fields.map { x =>
+        if (nameMap.isDefinedAt(x.name)) {
+          dialect.quoteIdentifier(x.name)
+        } else if (lowercaseNameMap.isDefinedAt(x.name.toLowerCase)) {
+          dialect.quoteIdentifier(nameMap(x.name.toLowerCase))
```
viirya: nameMap -> lowercaseNameMap.

dongjoon-hyun:
Thank you for review, @viirya .
I'll update like that.

gatorsmile:
This is a bug fix, right? Will review this tomorrow.

dongjoon-hyun:
Yes, right! Thank you, @gatorsmile !

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
```diff
@@ -112,7 +112,25 @@ object JdbcUtils extends Logging {
    */
   def insertStatement(conn: Connection, table: String, rddSchema: StructType, dialect: JdbcDialect)
       : PreparedStatement = {
-    val columns = rddSchema.fields.map(x => dialect.quoteIdentifier(x.name)).mkString(",")
+    // Use database column names instead of RDD schema column names
+    val tableSchemaQuery = conn.prepareStatement(dialect.getSchemaQuery(table))
```
viirya: I think the table schema won't change during inserting all data. You ask table schema for every insert statement now. Can we do this once in caller side (i.e., `savePartition`) and reuse the schema then?

dongjoon-hyun: Thank you for review, @viirya . I'll try to update like that.

gatorsmile: We can get the table schema [when we checking whether the table exists](https://github.com/apache/spark/blob/fb07bbe575aabe68422fd3a31865101fb7fa1722/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcRelationProvider.scala#L63).

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
```diff
@@ -112,7 +112,25 @@ object JdbcUtils extends Logging {
...
+      val nameMap = tableSchema.fields.map(f => f.name -> f.name).toMap
+      val lowercaseNameMap = tableSchema.fields.map(f => f.name.toLowerCase -> f.name).toMap
+      columns = rddSchema.fields.map { x =>
+        if (nameMap.isDefinedAt(x.name)) {
+          dialect.quoteIdentifier(x.name)
+        } else if (lowercaseNameMap.isDefinedAt(x.name.toLowerCase)) {
+          dialect.quoteIdentifier(lowercaseNameMap(x.name.toLowerCase))
+        } else {
+          throw new SQLException(s"""Column "${x.name}" not found""")
+        }
```
gatorsmile: The name resolution should be still controlled by `spark.sql.caseSensitive`, right?

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
```diff
@@ -112,7 +112,25 @@ object JdbcUtils extends Logging {
...
+      }.mkString(",")
+    } finally {
+      tableSchemaQuery.close()
+    }
     val placeholders = rddSchema.fields.map(_ => "?").mkString(",")
     val sql = s"INSERT INTO $table ($columns) VALUES ($placeholders)"
     conn.prepareStatement(sql)
```
gatorsmile: Can we build the INSERT SQL statement in `saveTable` based on the schema? No need to prepare the generated statement in `saveTable`.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
dongjoon-hyun: Thank you for review, @gatorsmile .
Yes. That looks great! We can use `getSchemaQuery` instead of `tableExist`.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
dongjoon-hyun: Yep. I'll fix that.

dongjoon-hyun:
The PR is updated to 
- get table schema once in `createRelation`
- respect `spark.sql.caseSensitive`

For `insertStatement`, I thought it seems to be better to keep the current structure to isolate `INSERT` PreparedStatement generation code like `dialect.quoteIdentifier`.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
```diff
@@ -211,6 +211,52 @@ object JdbcUtils extends Logging {
...
+        getSchema(statement.executeQuery(), dialect)
+      } finally {
+        statement.close()
+      }
+    } match {
+      case Success(v) =>
+        Some(v)
+      case Failure(e) =>
+        None
+    }
```
gatorsmile: Please do not use Try/Success/Failure. https://github.com/databricks/scala-style-guide#exception-handling-try-vs-try

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcRelationProvider.scala
```diff
@@ -60,23 +60,27 @@ class JdbcRelationProvider extends CreatableRelationProvider
...
+      val caseSensitive = sqlContext.conf.caseSensitiveAnalysis
       if (tableExists) {
         mode match {
           case SaveMode.Overwrite =>
-            if (isTruncate && isCascadingTruncateTable(url) == Some(false)) {
+            val savingSchema = if (isTruncate && isCascadingTruncateTable(url) == Some(false)) {
               // In this case, we should truncate table and then load.
               truncateTable(conn, table)
-              saveTable(df, url, table, jdbcOptions)
+              JdbcUtils.getSavingSchema(df.schema, tableSchema.get, caseSensitive)
```
gatorsmile: Based on your current implementation, the `tableSchema` could be None, right?

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
```diff
@@ -211,6 +211,52 @@ object JdbcUtils extends Logging {
...
+  }
+
+  /**
+   * Returns the saving schema using rddSchema's sequence and tableSchema's name.
```
gatorsmile: `saving schema` is not right. We need a better name here. 

`rddSchema's sequence and tableSchema's name` -> `rddSchema's column sequence and tableSchema's column names`

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
```diff
@@ -211,6 +211,52 @@ object JdbcUtils extends Logging {
...
+    val nameMap = tableSchema.fields.map(f => f.name -> f).toMap
+    val lowercaseNameMap = tableSchema.fields.map(f => f.name.toLowerCase -> f).toMap
+
+    var schema = new StructType()
+    rddSchema.fields.foreach { f =>
+      if (nameMap.isDefinedAt(f.name)) {
+        // identical names
+        schema = schema.add(nameMap(f.name))
+      } else if (!caseSensitive && lowercaseNameMap.isDefinedAt(f.name.toLowerCase)) {
+        // case-insensitive identical names
```
gatorsmile: Need to improve the comments. Actually, we return case sensitive column names.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
```diff
@@ -211,6 +211,52 @@ object JdbcUtils extends Logging {
...
+        schema = schema.add(lowercaseNameMap(f.name.toLowerCase))
+      } else {
+        throw new org.apache.spark.SparkException(s"""Column "${f.name}" not found""")
```
gatorsmile: `org.apache.spark.SparkException ` -> `AnalysisException`

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
gatorsmile: Here, we need to explain why we need to use the column sequences in `rddSchema` and why we need to use the column names in `tableSchema`

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcRelationProvider.scala
```diff
@@ -60,23 +60,27 @@ class JdbcRelationProvider extends CreatableRelationProvider
...
             } else {
               // Otherwise, do not truncate the table, instead drop and recreate it
               dropTable(conn, table)
               createTable(df.schema, url, table, createTableOptions, conn)
-              saveTable(df, url, table, jdbcOptions)
+              df.schema
             }
+            saveTable(df, url, table, savingSchema, jdbcOptions)
```
gatorsmile: How about passing the table schema and resolve the schema inside `saveTable`? It might simplify the codes

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
dongjoon-hyun: I see. Thank you for review! I wrote that by keeping the same logic in [tableExists](https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala#L66-L80
) because I thought the guideline is about using return `Try` as a return value before this PR. 

Sure, I'll remove the usage of `Try/Success/Failure`.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
dongjoon-hyun: Yep. I'll add more correct details here.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
dongjoon-hyun: My bad. I meant `case-insensitively identical names`. I'll revise this.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
dongjoon-hyun: Sure!

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcRelationProvider.scala
dongjoon-hyun: That could be. But, to do that, `JdbcUtil.saveTable` need to understand `SaveMode`, too. Is it okay?
Currently, `JdbcUtil` only provides somewhat primitive APIs.

dongjoon-hyun:
The failure is not related to this PR.
```
[info] - fatal errors from a source should be sent to the user *** FAILED *** (84 milliseconds)
[info]   org.apache.spark.sql.streaming.StreamingQueryException: Query [id = 5ae8eeb2-81f1-44df-b8ad-55b780512ee5, runId = 3a5bfebc-e3a3-496e-987b-e0e789f1fdcb] terminated with exception: null
```

gatorsmile:
You can move normalization of schemas into `saveTable` without passing save mode. See my private branch: https://github.com/apache/spark/compare/master...gatorsmile:pr-15664Changed?expand=1

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
```diff
@@ -211,6 +211,55 @@ object JdbcUtils extends Logging {
...
+   * Returns a schema using rddSchema's column sequence and tableSchema's column names.
+   *
+   * When appending data into some case-sensitive DBMSs like PostgreSQL/Oracle, we need to respect
+   * the existing case-sensitive column names instead of RDD column names for user convenience.
+   * See SPARK-18123 for more details.
+   */
+  def normalizeSchema(
+      rddSchema: StructType,
+      tableSchema: StructType,
+      caseSensitive: Boolean): StructType = {
```
gatorsmile: `isCaseSensitive`

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
```diff
@@ -211,6 +211,55 @@ object JdbcUtils extends Logging {
   }
 
   /**
+   * Returns the schema if the table already exists in the JDBC database.
+   */
+  def getSchema(conn: Connection, url: String, table: String): Option[StructType] = {
+    val dialect = JdbcDialects.get(url)
+
+    try {
+      val statement = conn.prepareStatement(dialect.getSchemaQuery(table))
```
gatorsmile: What is the reason why this should be included in the try block?

dongjoon-hyun:
Wow. Thank you for the code.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
dongjoon-hyun: Yes. It's due to `PreparedStatement prepareStatement(String sql) throws SQLException`. I want to make it sure to return `None` instead of raising SQLException at this line.

dongjoon-hyun: Is it better not to hide that here? If then, I'll remove that `try`.

gatorsmile: JDBC data sources might detect whether the table exists when preparing the SQL statement. Thus, please keep it. Thanks

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
```diff
@@ -568,10 +617,10 @@ object JdbcUtils extends Logging {
         conn.setAutoCommit(false) // Everything in the same db transaction.
         conn.setTransactionIsolation(finalIsolationLevel)
       }
-      val stmt = insertStatement(conn, table, rddSchema, dialect)
-      val setters: Array[JDBCValueSetter] = rddSchema.fields.map(_.dataType)
+      val stmt = insertStatement(conn, table, schema, dialect)
+      val setters: Array[JDBCValueSetter] = schema.fields.map(_.dataType)
         .map(makeSetter(conn, dialect, _)).toArray
```
gatorsmile: Could you just simplify it? 
```Scala
val setters = schema.fields.map(f => makeSetter(conn, dialect, f.dataType))
```

dongjoon-hyun: Sure!

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
gatorsmile: After rethinking about it, the result of this function `normalizeSchema` is a mixed of two schemas (source/rdd schema + target/table schema). We should not mix them together.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
```diff
@@ -568,10 +617,9 @@ object JdbcUtils extends Logging {
         conn.setAutoCommit(false) // Everything in the same db transaction.
         conn.setTransactionIsolation(finalIsolationLevel)
       }
-      val stmt = insertStatement(conn, table, rddSchema, dialect)
-      val setters: Array[JDBCValueSetter] = rddSchema.fields.map(_.dataType)
-        .map(makeSetter(conn, dialect, _)).toArray
-      val numFields = rddSchema.fields.length
+      val stmt = insertStatement(conn, table, schema, dialect)
```
gatorsmile: `insertStatement` should use the target/table schema; the other places in this function should use the source/rdd schema.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
dongjoon-hyun: Then, do you want to create `rddSchema` StructField with tableSchema's column name (if not matched)?

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
gatorsmile: See the comment in https://github.com/apache/spark/pull/15664#discussion_r94094625

dongjoon-hyun: I see.

dongjoon-hyun: Ur, `target/table schema` has a different column order than `rdd` order. Why do we need to use `table` schema? Actually, we need table column names only.

gatorsmile: We need to think about what is a better API interface of `savePartition`. Let me submit another commit and show you my rough idea.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
```diff
@@ -211,6 +211,55 @@ object JdbcUtils extends Logging {
...
   /**
+   * Returns the schema if the table already exists in the JDBC database.
+   */
+  def getSchema(conn: Connection, url: String, table: String): Option[StructType] = {
+    val dialect = JdbcDialects.get(url)
+
+    try {
+      val statement = conn.prepareStatement(dialect.getSchemaQuery(table))
+      try {
+        Some(getSchema(statement.executeQuery(), dialect))
```
viirya: `getSchema` will throw an exception when the schema contains an unsupported type. Now we use it to check if the table exists. Does it change current behavior? E.g., the insertion working before now fails.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
gatorsmile: This is my rough idea: https://github.com/apache/spark/compare/master...gatorsmile:pr-15664Changed1

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
```diff
@@ -211,6 +211,55 @@ object JdbcUtils extends Logging {
...
+    var schema = new StructType()
+    rddSchema.fields.foreach { f =>
+      if (nameMap.isDefinedAt(f.name)) {
+        // identical names
+        schema = schema.add(nameMap(f.name))
+      } else if (!caseSensitive && lowercaseNameMap.isDefinedAt(f.name.toLowerCase)) {
+        // identical names in a case-insensitive way
+        schema = schema.add(lowercaseNameMap(f.name.toLowerCase))
+      } else {
+        throw new AnalysisException(s"""Column "${f.name}" not found""")
```
viirya: The error message looks ambiguous. Maybe `Column "${f.name}" in RDD not found in table schema`?

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
gatorsmile: For unsupported types, it will throw an `SQLException`, right?

viirya: yes.

gatorsmile: Then, we can keep the existing way. See https://github.com/apache/spark/compare/master...gatorsmile:pr-15664Changed1

dongjoon-hyun:
Thank you for deep review again, @viirya and @gatorsmile .
I'll update the PR soon.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
dongjoon-hyun: Yep. That looks much better than the current! Thank you.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
dongjoon-hyun: It may affect some other test cases, but I'll try!

dongjoon-hyun: Oh, it's already in @gatorsmile 's patch.
```
        throw new AnalysisException(s"""Column "${col.name}" not found in schema $tableSchema""")
```

dongjoon-hyun:
Hi, @gatorsmile .
I have no objections on your patch! So, I merged that into this PR.

dongjoon-hyun:
The only failure is irrelevant to this PR.
```
[info] StreamSuite:
[info] - fatal errors from a source should be sent to the user *** FAILED *** (101 milliseconds)
[info]   org.apache.spark.sql.streaming.StreamingQueryException: Query [id = b3869a67-68d5-4fc1-a3e2-d5879a49166c, runId = cdb70502-63b9-439a-9ece-bdbdcad5c2b4] terminated with exception: null
[info]   at org.apache.spark.sql.execution.streaming.StreamExecution.org$apache$spark$sql$execution$streaming$StreamExecution$$runBatches(StreamExecution.scala:296)
[info]   at org.apache.spark.sql.execution.streaming.StreamExecution$$anon$1.run(StreamExecution.scala:186)
[info]   Cause: org.apache.spark.sql.streaming.StreamSuite$$anonfun$12$$anon$2:
```

dongjoon-hyun:
Retest this please

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcRelationProvider.scala
```diff
@@ -57,26 +57,28 @@ class JdbcRelationProvider extends CreatableRelationProvider
...
       val tableExists = JdbcUtils.tableExists(conn, url, table)
       if (tableExists) {
+        val tableSchema = JdbcUtils.getSchemaOption(conn, url, table)
         mode match {
           case SaveMode.Overwrite =>
             if (isTruncate && isCascadingTruncateTable(url) == Some(false)) {
               // In this case, we should truncate table and then load.
               truncateTable(conn, table)
-              saveTable(df, url, table, jdbcOptions)
+              saveTable(df, url, table, tableSchema.get, isCaseSensitive, jdbcOptions)
```
gatorsmile: We need to pass `tableSchema`. It is not safe to do `tableSchema.get`. : )

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
```diff
@@ -108,14 +108,32 @@ object JdbcUtils extends Logging {
...
-   * Returns a PreparedStatement that inserts a row into table via conn.
+   * Returns an Insert SQL statement for inserting a row into the target table via JDBC conn.
    */
-  def insertStatement(conn: Connection, table: String, rddSchema: StructType, dialect: JdbcDialect)
-      : PreparedStatement = {
-    val columns = rddSchema.fields.map(x => dialect.quoteIdentifier(x.name)).mkString(",")
+  def getInsertStatement(
+      table: String,
+      rddSchema: StructType,
+      tableSchema: StructType,
```
gatorsmile: If `tableSchema` is None, we should follow the original way. Please add the extra logics. Thanks!

dongjoon-hyun: Right. I see. I'll update like that.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcRelationProvider.scala
dongjoon-hyun: Sure! Now, there is no pre-check.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcRelationProvider.scala
```diff
@@ -67,16 +68,18 @@ class JdbcRelationProvider extends CreatableRelationProvider
             if (isTruncate && isCascadingTruncateTable(url) == Some(false)) {
               // In this case, we should truncate table and then load.
               truncateTable(conn, table)
-              saveTable(df, url, table, jdbcOptions)
+              val tableSchema = JdbcUtils.getSchemaOption(conn, url, table)
```
dongjoon-hyun: I moved this into `case` statements.
Since `JdbcUtils.tableExists` is used, `getSchemaOption` can be skipped for the other `SaveMode`.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
```diff
@@ -108,14 +108,36 @@ object JdbcUtils extends Logging {
...
-      : PreparedStatement = {
-    val columns = rddSchema.fields.map(x => dialect.quoteIdentifier(x.name)).mkString(",")
+  def getInsertStatement(
+      table: String,
+      rddSchema: StructType,
+      tableSchema: Option[StructType],
+      isCaseSensitive: Boolean,
+      dialect: JdbcDialect): String = {
+    val columns = if (tableSchema.isEmpty) {
+      rddSchema.fields.map(x => dialect.quoteIdentifier(x.name)).mkString(",")
```
dongjoon-hyun: The legacy behavior is used when `tableSchema` is `None`.

dongjoon-hyun:
The only failure is irrelevant to this PR.
```
[info] StreamSuite:
[info] - fatal errors from a source should be sent to the user *** FAILED *** (84 milliseconds)
[info]   org.apache.spark.sql.streaming.StreamingQueryException: Query [id = d522aafe-085e-43e4-b796-037695dec113, runId = fc2f679a-907a-4445-8042-9198649bb55d] terminated with exception: null
[info]   at org.apache.spark.sql.execution.streaming.StreamExecution.org$apache$spark$sql$execution$streaming$StreamExecution$$runBatches(StreamExecution.scala:296)
[info]   at org.apache.spark.sql.execution.streaming.StreamExecution$$anon$1.run(StreamExecution.scala:186)
[info]   Cause: org.apache.spark.sql.streaming.StreamSuite$$anonfun$12$$anon$2:
```

dongjoon-hyun:
Retest this please.

gatorsmile:
LGTM

gatorsmile:
Merging to master. Thanks!

dongjoon-hyun:
Thank you, @gatorsmile .
Happy New Year! :)

PR #15665, opened 2016-10-27
[SPARK-18146] [SQL] Avoid using Union to chain together create table and repair partition commands

ericl:

The behavior of union is not well defined here. It is safer to explicitly execute these commands in order. The other use of `Union` in this way will be removed by https://github.com/apache/spark/pull/15633
## How was this patch tested?

Existing tests.

cc @yhuai @cloud-fan

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/DataFrameWriter.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/DataFrameWriter.scala
@@ -388,7 +388,7 @@ final class DataFrameWriter[T] private[sql](ds: Dataset[T]) {
           partitionColumnNames = partitioningColumns.getOrElse(Nil),
           bucketSpec = getBucketSpec
         )
-        val result = df.sparkSession.sessionState.executePlan(
+        df.sparkSession.sessionState.executePlan(
           CreateTable(tableDesc, mode, Some(df.logicalPlan))).toRdd
         if (tableDesc.partitionColumnNames.nonEmpty &&
             df.sparkSession.sqlContext.conf.manageFilesourcePartitions) {
@@ -396,7 +396,6 @@ final class DataFrameWriter[T] private[sql](ds: Dataset[T]) {
           df.sparkSession.sessionState.executePlan(
             AlterTableRecoverPartitionsCommand(tableDesc.identifier)).toRdd
         }
-        result
     }
   }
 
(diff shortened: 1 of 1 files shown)


sql/core/src/main/scala/org/apache/spark/sql/DataFrameWriter.scala
```diff
@@ -388,16 +388,15 @@ final class DataFrameWriter[T] private[sql](ds: Dataset[T]) {
           partitionColumnNames = partitioningColumns.getOrElse(Nil),
           bucketSpec = getBucketSpec
         )
-        val createCmd = CreateTable(tableDesc, mode, Some(df.logicalPlan))
-        val cmd = if (tableDesc.partitionColumnNames.nonEmpty &&
+        val result = df.sparkSession.sessionState.executePlan(
+          CreateTable(tableDesc, mode, Some(df.logicalPlan)))
```
yhuai: Is the execution of this command triggered at here?

yhuai: Looks like you want to call `result.toRdd` at here to trigger the execution of a command? Then inside the loop to use toRdd again.

cloud-fan: yea, we don't need the `result` variable.

sql/core/src/main/scala/org/apache/spark/sql/DataFrameWriter.scala
```diff
@@ -254,7 +254,8 @@ final class DataFrameWriter[T] private[sql](ds: Dataset[T]) {
       )
     }
 
-    df.sparkSession.sessionState.executePlan(
+    df.sparkSession.sessionState.
```
rxin: what's going on here? a dot ?

ericl: Some github text editing gone wrong... fixed it

sql/core/src/main/scala/org/apache/spark/sql/DataFrameWriter.scala
ericl: Fixed

rxin:
cc @hvanhovell @cloud-fan did you have anything to do with this in the past?

ericl:
This was added in the recent pr for datasource partition management: https://github.com/apache/spark/pull/15515

cloud-fan:
LGTM, merging to master!

PR #15667, opened 2016-10-28
[SPARK-18107][SQL] Insert overwrite statement runs much slower in spark-sql than it does in hive-client

viirya:

As reported on the jira, insert overwrite statement runs much slower in Spark, compared with hive-client.

It seems there is a patch [HIVE-11940](https://github.com/apache/hive/commit/ba21806b77287e237e1aa68fa169d2a81e07346d) which largely improves insert overwrite performance on Hive. HIVE-11940 is patched after Hive 2.0.0.

Because Spark SQL uses older Hive library, we can not benefit from such improvement.

The reporter verified that there is also a big performance gap between Hive 1.2.1 (520.037 secs) and Hive 2.0.1 (35.975 secs) on insert overwrite execution.

Instead of upgrading to Hive 2.0 in Spark SQL, which might not be a trivial task, this patch provides an approach to delete the partition before asking Hive to load data files into the partition.

Note: The case reported on the jira is insert overwrite to partition. Since `Hive.loadTable` also uses the function to replace files, insert overwrite to table should has the same issue. We can take the same approach to delete the table first. I will upgrade this to include this.
## How was this patch tested?

Jenkins tests.

There are existing tests using insert overwrite statement. Those tests should be passed. I added a new test to specially test insert overwrite into partition.

For performance issue, as I don't have Hive 2.0 environment, this needs the reporter to verify it. Please refer to the jira.

### DIFF
--- a/sql/hive/src/main/scala/org/apache/spark/sql/hive/execution/InsertIntoHiveTable.scala
+++ b/sql/hive/src/main/scala/org/apache/spark/sql/hive/execution/InsertIntoHiveTable.scala
@@ -37,7 +37,7 @@ import org.apache.spark.sql.catalyst.InternalRow
 import org.apache.spark.sql.catalyst.expressions.Attribute
 import org.apache.spark.sql.catalyst.plans.physical.Partitioning
 import org.apache.spark.sql.execution.{SparkPlan, UnaryExecNode}
-import org.apache.spark.sql.execution.command.AlterTableAddPartitionCommand
+import org.apache.spark.sql.execution.command.{AlterTableAddPartitionCommand, AlterTableDropPartitionCommand}
 import org.apache.spark.sql.hive._
 import org.apache.spark.sql.hive.HiveShim.{ShimFileSinkDesc => FileSinkDesc}
 import org.apache.spark.SparkException
@@ -266,24 +266,21 @@ case class InsertIntoHiveTable(
           // version and we may not want to catch up new Hive version every time. We delete the
           // Hive partition first and then load data file into the Hive partition.
           if (oldPart.nonEmpty && overwrite) {
-            externalCatalog.dropPartitions(
-              table.catalogTable.database,
-              table.catalogTable.identifier.table,
-              parts = Seq(partitionSpec),
-              ignoreIfNotExists = true,
-              purge = false)
-            // Although `externalCatalog.loadPartition` will create the partition,
-            // the old partition may use custom properties such as location, so we need to create
-            // it manually.
-            val partitionSpecAndLocation = Seq((partitionSpec, oldPart.get.storage.locationUri))
-            AlterTableAddPartitionCommand(
-              tableName = table.catalogTable.identifier,
-              partitionSpecsAndLocs = partitionSpecAndLocation,
-              ifNotExists = true
-            ).run(sqlContext.sparkSession)
-
-            // Don't let Hive do overwrite operation since it is slower.
-            doOverwrite = false
+            oldPart.get.storage.locationUri.map { uri =>
+              val partitionPath = new Path(uri)
+              val fs = partitionPath.getFileSystem(hadoopConf)
+              if (fs.exists(partitionPath)) {
+                val pathPermission = fs.getFileStatus(partitionPath).getPermission()
+                if (!fs.delete(partitionPath, true)) {
+                  throw new RuntimeException(
+                    "Cannot remove partition directory '" + partitionPath.toString)
+                } else {
+                  fs.mkdirs(partitionPath, pathPermission)
+                }
+                // Don't let Hive do overwrite operation since it is slower.
+                doOverwrite = false
+              }
+            }
           }
 
           // inheritTableSpecs is set to true. It should be set to false for an IMPORT query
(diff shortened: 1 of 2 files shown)


rxin:
cc @ericl

viirya:
I checked the failed test. Special characters in the partition path cause the failure, e..g.,

    alter table ppr_test add partition (ds = '12:4');
    alter table ppr_test add partition (ds = '12%4');
    alter table ppr_test add partition (ds = '12*4');

ericl:
How does the performance look like before / after this patch?

viirya:
@ericl I don't have the Hive environment to compare. We need to wait for issue reporter to verify that.

snodawn:
Sorry, I'm later, for preparing the environment for testing cost me a lot of time. I have tested the performance before and after the patch. But it seems to improve a few after patching, where it costs 531 seconds before patching, and costs 518 seconds after patching.  So I think I need to do more testing to find out the problem.

viirya:
@snodawn Interesting...I will try to find out it too.

viirya:
@snodawn Can you try it again? I've updated this.

snodawn:
@viirya I have tested the new patch, which performs better than expected. Before patching, it costs about 500~600 seconds, but now it just cost me about 16 seconds to run the same statement. But it still runs slow when I run such sql:

insert overwrite table login4game partition(pt,dt)    select distinct account_name,role_id,server,'1476979200' as recdate, 'mix' as platform, 'mix' as pid, 'mix' as dev, pt, dt  from tbllog_login  where pt='mix_en' and  dt='2016-10-21';

It's the dynamic  partition in hive, where we needn't to specify the partition value when inserting.  I test it in hive 2.0.1, it costs 47.822 seconds, but in hive 1.2.1, it costs 574.33 seconds,  as the same with what it does in spark, which is 526.44 seconds.

viirya:
@snodawn I just updated this with a new commit. Do you use this new patch to test?

Yeah, current fixing doesn't consider dynamic partition. I would like to see if we can improve static partition. I will look into dynamic partition later.

snodawn:
The execution logs in spark show me that, it does the same thing as what it does before I add the patch, which may be the reason why it runs so slow when running dynamic insert overwrite statement.

viirya:
@snodawn Current fixing does not do anything for dynamic partition. So we can expect this.

viirya:
@snodawn You just said the new patch runs better. Do you use the latest patch I just updated in about 1 hr ago? Thanks.

snodawn:
Ok, I see. I haven't tested the newest code, I would try it later.

viirya:
@snodawn Thanks. I expect it should perform as good as you tested. But there are few tests failed I fix in newer patch.

snodawn:
@viirya  I have tested the newest patch.  It performs good in running the same sql as I ran before.

viirya:
@snodawn Thanks! I will address dynamic partition in next commit.

sql/hive/src/main/scala/org/apache/spark/sql/hive/execution/InsertIntoHiveTable.scala
```diff
@@ -257,7 +258,31 @@ case class InsertIntoHiveTable(
...
+            oldPart.get.storage.locationUri.map { uri =>
+              val partitionPath = new Path(uri)
+              val fs = partitionPath.getFileSystem(hadoopConf)
+              if (fs.exists(partitionPath)) {
+                val pathPermission = fs.getFileStatus(partitionPath).getPermission()
+                if (!fs.delete(partitionPath, true)) {
+                  throw new RuntimeException(
+                    "Cannot remove partition directory '" + partitionPath.toString)
+                } else {
+                  fs.mkdirs(partitionPath, pathPermission)
```
ericl: Is the mkdir necessary?

sql/hive/src/main/scala/org/apache/spark/sql/hive/execution/InsertIntoHiveTable.scala
```diff
@@ -257,7 +258,31 @@ case class InsertIntoHiveTable(
             table.catalogTable.identifier.table,
             partitionSpec)
 
+        var doOverwrite = overwrite
```
ericl: nit: `doHiveOverwrite`?

sql/hive/src/main/scala/org/apache/spark/sql/hive/execution/InsertIntoHiveTable.scala
viirya: I was thinking Hive will complain if the dir is not existing. But looks like it won't. Let me remove this and see if tests can passed.

sql/hive/src/main/scala/org/apache/spark/sql/hive/execution/InsertIntoHiveTable.scala
viirya: ok. updated.

ericl:
lgtm if tests pass

viirya:
@ericl Dynamic partition would be more complicated. Should we do it in this or in follow-up?

ericl:
Let's do it in a follow-up.

snodawn:
@viirya @ericl Thanks a lot. The new patch is also passed in my test. And I sincerely hope dynamic partition could be completed in the future, thanks.

rxin:
Merging in master. Thanks.

viirya:
@rxin @ericl Thanks!

@snodawn I will address dynamic partition in a follow-up pr.

PR #15668, opened 2016-10-28
[SPARK-18137][SQL]Fix RewriteDistinctAggregates UnresolvedException when a UDAF has a foldable TypeCheck

windpiger:

In RewriteDistinctAggregates rewrite funtion,after the UDAF's childs are mapped to AttributeRefference, If the UDAF(such as ApproximatePercentile) has a foldable TypeCheck for the input, It will failed because the AttributeRefference is not foldable,then the UDAF is not resolved, and then nullify on the unresolved object will throw a Exception.

In this PR, only map Unfoldable child to AttributeRefference, this can avoid the UDAF's foldable TypeCheck. and then only Expand Unfoldable child, there is no need to Expand a static value(foldable value).

**Before sql result**

> select percentile_approxy(key,0.99999),count(distinct key),sume(distinc key) from src limit 1
> org.apache.spark.sql.catalyst.analysis.UnresolvedException: Invalid call to dataType on unresolved object, tree: 'percentile_approx(CAST(src.`key` AS DOUBLE), CAST(0.99999BD AS DOUBLE), 10000)
> at org.apache.spark.sql.catalyst.analysis.UnresolvedAttribute.dataType(unresolved.scala:92)
>     at org.apache.spark.sql.catalyst.optimizer.RewriteDistinctAggregates$.org$apache$spark$sql$catalyst$optimizer$RewriteDistinctAggregates$$nullify(RewriteDistinctAggregates.scala:261)

**After sql result**

> select percentile_approxy(key,0.99999),count(distinct key),sume(distinc key) from src limit 1
> [498.0,309,79136]
## How was this patch tested?

Add a test case in HiveUDFSuit.

### DIFF
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/RewriteDistinctAggregates.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/RewriteDistinctAggregates.scala
@@ -115,8 +115,7 @@ object RewriteDistinctAggregates extends Rule[LogicalPlan] {
     }
 
     // Extract distinct aggregate expressions.
-    val distinctAggGroups = aggExpressions.filter(_.isDistinct).groupBy {
-      e =>
+    val distinctAggGroups = aggExpressions.filter(_.isDistinct).groupBy { e =>
         val unfoldableChildren = e.aggregateFunction.children.filter(!_.foldable).toSet
         if (unfoldableChildren.nonEmpty) {
           // Only expand the unfoldable children
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/RewriteDistinctAggregates.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/RewriteDistinctAggregates.scala
@@ -115,11 +115,12 @@ object RewriteDistinctAggregates extends Rule[LogicalPlan] {
     }
 
     // Extract distinct aggregate expressions.
-    val distinctAggGroups = aggExpressions.filter(_.isDistinct).groupBy{
+    val distinctAggGroups = aggExpressions.filter(_.isDistinct).groupBy {
       e =>
-        if (e.aggregateFunction.children.exists(!_.foldable)) {
+        val unfoldableChildren = e.aggregateFunction.children.filter(!_.foldable).toSet
+        if (unfoldableChildren.nonEmpty) {
           // Only expand the unfoldable children
-          e.aggregateFunction.children.filter(!_.foldable).toSet
+          unfoldableChildren
         } else {
           // If aggregateFunction's children are all foldable
           // we must expand at least one of the children (here we take the first child),
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/RewriteDistinctAggregates.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/RewriteDistinctAggregates.scala
@@ -142,11 +142,19 @@ object RewriteDistinctAggregates extends Rule[LogicalPlan] {
 
       // Setup unique distinct aggregate children.
       val distinctAggChildren = distinctAggGroups.keySet.flatten.toSeq.distinct
-      val distinctAggChildAttrMap = distinctAggChildren.map(expressionAttributePair)
-      val distinctAggChildAttrs = distinctAggChildAttrMap.map(_._2)
+      val distinctAggChildFoldable = distinctAggChildren.filter(_.foldable)
+      // 1.only unfoldable child should be expand
+      // 2.if foldable child mapped to AttributeRefference using expressionAttributePair,
+      // the udaf function(such as ApproximatePercentile)
+      // which has a foldable TypeCheck will failed,because AttributeRefference is unfoldable
+      val distinctAggChildUnFoldableAttrMap = distinctAggChildren
+        .filter(!_.foldable).map(expressionAttributePair)
+
+      val distinctAggChildrenUnFoldableAttrs = distinctAggChildUnFoldableAttrMap.map(_._2)
 
       // Setup expand & aggregate operators for distinct aggregate expressions.
-      val distinctAggChildAttrLookup = distinctAggChildAttrMap.toMap
+      val distinctAggChildAttrLookup = (distinctAggChildUnFoldableAttrMap
+        ++ distinctAggChildFoldable.map(c => c -> c)).toMap
       val distinctAggOperatorMap = distinctAggGroups.toSeq.zipWithIndex.map {
         case ((group, expressions), i) =>
           val id = Literal(i + 1)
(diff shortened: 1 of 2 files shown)


cloud-fan:
OK to test

cloud-fan:
ok to test

hvanhovell:
@windpiger I have taken a quick look at your PR. I think your approach has merit, however I was wondering if it is easier not to extract/move the foldable expressions at all?

cloud-fan:
+1 on @hvanhovell 's idea, literals don't need to be pre-executed.

windpiger:
@hvanhovell @cloud-fan  Do you mean that just change distinctAggChildren/regularAggChildren to represent the unfoldable expressions, and then the followings code could be unchanged? 

...
val distinctAggChildren = distinctAggGroups.keySet.flatten.toSeq.distinct.**filter(!_.foldable)**
...

val regularAggChildren = regularAggExprs.flatMap(\_.aggregateFunction.children).distinct.**filter(!_.foldable)**

cloud-fan:
isn't it just a two-line change? https://github.com/apache/spark/compare/master...cloud-fan:showcase?expand=1

I haven't run all tests yet, but I think it should work

cloud-fan:
It looks to me that after this PR, we still have literals in expand, can you call `df.explain(true)` to double check the query plan?

and how about this? https://github.com/apache/spark/compare/master...cloud-fan:showcase?expand=1   It's a little simpler and do avoid puting literals in expand.

windpiger:
@cloud-fan yes,It is simpler,I check the query plan, my test case 
`sql("SELECTpercentile_approx(key, 0.99999)" +
       ", count(distinct key),sum(distinct key) FROM src LIMIT 1")`
explain is ok as follows:
`+- *Expand [List(null, null, 0, cast(key#29 as double)), List(key#29, null, 1, null), List(null, cast(key#29 as bigint), 2, null)], [src.`key`#38, CAST(src.`key` AS BIGINT)#39L, gid#37, CAST(src.`key` AS DOUBLE)#40]
                           +- HiveTableScan [key#29], MetastoreRelation default, src`

But, I miss one case that the distinct on a constant like:
`sql("SELECTpercentile_approx(key, 0.99999)" +
       ", count(distinct key),sum(distinct key),count(distinct 1) FROM src LIMIT 1")`
explain with a literal expand as follows:
`+- *Expand [List(null, null, null, 0, cast(key#69 as double)), List(1, null, null, 1, null), List(null, key#69, null, 2, null), List(null, null, cast(key#69 as bigint), 3, null)], [src.`key`#80, CAST(src.`key` AS BIGINT)#81L, gid#79, CAST(src.`key` AS DOUBLE)#82]
                           +- HiveTableScan [key#69], MetastoreRelation default, src`

windpiger:
@cloud-fan  I rewrite the expand logic:
1) If the aggFunction has unfoldable children,it will only expand the unfoldable children;
2) If the aggFunction only has foldable children, it will only expand the first children, generally the aggFunction should not run foldable TypeChecker for the first child(this is the way to avoid foldable typechecker for the first child).

for example:
select percentile_approx(2,0.99999),sum(distinct 1),count(distinct 1,2,3,1) from src limit 1

explained:
`
== Physical Plan ==
CollectLimit 1
+- *HashAggregate(keys=[], functions=[first(if ((gid#177 = 0)) percentile_approx(CAST(2 AS DOUBLE), CAST(0.99999BD AS DOUBLE), 10000)#180 else null, true), sum(if ((gid#177 = 1)) CAST(1 AS BIGINT)#178L else null), count(if ((gid#177 = 2)) 1#179 else null, 2, 3, if ((gid#177 = 2)) 1#179 else null)])
   +- Exchange SinglePartition
      +- *HashAggregate(keys=[], functions=[partial_first(if ((gid#177 = 0)) percentile_approx(CAST(2 AS DOUBLE), CAST(0.99999BD AS DOUBLE), 10000)#180 else null, true), partial_sum(if ((gid#177 = 1)) CAST(1 AS BIGINT)#178L else null), partial_count(if ((gid#177 = 2)) 1#179 else null, 2, 3, if ((gid#177 = 2)) 1#179 else null)])
         +- SortAggregate(key=[CAST(1 AS BIGINT)#178L, 1#179, gid#177], functions=[percentile_approx(2.0, 0.99999, 10000, 0, 0)])
            +- *Sort [CAST(1 AS BIGINT)#178L ASC NULLS FIRST, 1#179 ASC NULLS FIRST, gid#177 ASC NULLS FIRST], false, 0
               +- Exchange hashpartitioning(CAST(1 AS BIGINT)#178L, 1#179, gid#177, 200)
                  +- SortAggregate(key=[CAST(1 AS BIGINT)#178L, 1#179, gid#177], functions=[partial_percentile_approx(2.0, 0.99999, 10000, 0, 0)])
                     +- *Sort [CAST(1 AS BIGINT)#178L ASC NULLS FIRST, 1#179 ASC NULLS FIRST, gid#177 ASC NULLS FIRST], false, 0
                        +- *Expand [List(null, null, 0), List(1, null, 1), List(null, 1, 2)], [CAST(1 AS BIGINT)#178L, 1#179, gid#177]
                           +- HiveTableScan MetastoreRelation default, src
`
and the result is:
`
+--------------------------------------------------------------------+---------------+--------------------------+
|percentile_approx(CAST(2 AS DOUBLE), CAST(0.99999 AS DOUBLE), 10000)|sum(DISTINCT 1)|count(DISTINCT 1, 2, 3, 1)|
+--------------------------------------------------------------------+---------------+--------------------------+
|                                                                 2.0|              1|                         1|
+--------------------------------------------------------------------+---------------+--------------------------+
`

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/RewriteDistinctAggregates.scala
```diff
@@ -115,9 +115,21 @@ object RewriteDistinctAggregates extends Rule[LogicalPlan] {
...
+          // Only expand the unfoldable children
+          e.aggregateFunction.children.filter(!_.foldable).toSet
+        } else {
+          // If aggregateFunction's children are all foldable
+          // we must expand at least one of the children (here we take the first child),
+          // or If we don't, we will get the wrong result, for example:
+          // count(distinct 1) will be explained to count(1) after the rewrite function.
+          // Generally, the distinct aggregateFunction should not run
+          // foldable TypeCheck for the first child.
+          e.aggregateFunction.children.take(1).toSet
```
hvanhovell: This is a good catch. It would be great if we could git rid of this by constant folding (not needed in this PR). Another way of getting rid of this, would be by creating a separate processing group for these distincts.

hvanhovell (changes requested):
This is shaping up nicely. I left a few minor comments.

windpiger:
retest this please

windpiger:
@cloud-fan @hvanhovell could you please help to retest this？

cloud-fan:
retest this please

cloud-fan:
add to whitelist

cloud-fan:
LGTM, pending jenkins

hvanhovell:

hvanhovell:
@windpiger do you have a JIRA username? So I can credit you on the JIRA.

windpiger:
@hvanhovell  username is Song Jun

PR #15669, opened 2016-10-28
[SPARK-18160][CORE][YARN] spark.files & spark.jars should not be passed to driver in yarn mode

zjffdu:

spark.files is still passed to driver in yarn mode, so SparkContext will still handle it which cause the error in the jira desc.

## How was this patch tested?

Tested manually in a 5 node cluster. As this issue only happens in multiple node cluster, so I didn't write test for it.

### DIFF
--- a/core/src/main/scala/org/apache/spark/SparkContext.scala
+++ b/core/src/main/scala/org/apache/spark/SparkContext.scala
@@ -1716,29 +1716,12 @@ class SparkContext(config: SparkConf) extends Logging {
         key = uri.getScheme match {
           // A JAR file which exists only on the driver node
           case null | "file" =>
-            if (master == "yarn" && deployMode == "cluster") {
-              // In order for this to work in yarn cluster mode the user must specify the
-              // --addJars option to the client to upload the file into the distributed cache
-              // of the AM to make it show up in the current working directory.
-              val fileName = new Path(uri.getPath).getName()
-              try {
-                env.rpcEnv.fileServer.addJar(new File(fileName))
-              } catch {
-                case e: Exception =>
-                  // For now just log an error but allow to go through so spark examples work.
-                  // The spark examples don't really need the jar distributed since its also
-                  // the app jar.
-                  logError("Error adding jar (" + e + "), was the --addJars option used?")
-                  null
-              }
-            } else {
-              try {
-                env.rpcEnv.fileServer.addJar(new File(uri.getPath))
-              } catch {
-                case exc: FileNotFoundException =>
-                  logError(s"Jar not found at $path")
-                  null
-              }
+            try {
+              env.rpcEnv.fileServer.addJar(new File(uri.getPath))
+            } catch {
+              case exc: FileNotFoundException =>
+                logError(s"Jar not found at $path")
+                null
             }
           // A JAR file which exists locally on every worker node
           case "local" =>
--- a/core/src/main/scala/org/apache/spark/SparkContext.scala
+++ b/core/src/main/scala/org/apache/spark/SparkContext.scala
@@ -1440,16 +1440,14 @@ class SparkContext(config: SparkConf) extends Logging {
     val scheme = new URI(schemeCorrectedPath).getScheme
     if (!Array("http", "https", "ftp").contains(scheme)) {
       val fs = hadoopPath.getFileSystem(hadoopConfiguration)
-      if (fs.exists(hadoopPath)) {
-        val isDir = fs.getFileStatus(hadoopPath).isDirectory
-        if (!isLocal && scheme == "file" && isDir) {
-          throw new SparkException(s"addFile does not support local directories when not running " +
-            "local mode.")
-        }
-        if (!recursive && isDir) {
-          throw new SparkException(s"Added file $hadoopPath is a directory and recursive is not " +
-            "turned on.")
-        }
+      val isDir = fs.getFileStatus(hadoopPath).isDirectory
+      if (!isLocal && scheme == "file" && isDir) {
+        throw new SparkException(s"addFile does not support local directories when not running " +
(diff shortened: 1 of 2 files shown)


zjffdu:
@vanzin @JoshRosen @yanboliang @brkyvz Please help review.

tgravescs:
I thought we had fixed this in 2.0 to make it consistent that anytime --files is specified with yarn it goes through distributed cache.  It shouldn't be using the spark addFiles method to distribute.  Perhaps that broke?  @vanzin  do you remember

jerryshao:
Quite strange `--files` in yarn-client will fall into `SparkContext`? AFAIK it will be handled by yarn#client to add to distributed cache, it will not involve SparkContext's addFiles.

zjffdu:
@jerryshao spark.files is always passed to driver so SparkContext.addFile is called in yarn-cluster. 
https://github.com/apache/spark/blob/7bf8a4049866b2ec7fdf0406b1ad0c3a12488645/core/src/main/scala/org/apache/spark/deploy/SparkSubmit.scala#L609
```
    // Load any properties specified through --conf and the default properties file
    for ((k, v) <- args.sparkProperties) {
      sysProps.getOrElseUpdate(k, v)
    }
```

Seems the issue is that spark.files don't need to be passed to driver in yarn-cluster mode. In that case it can be fixed in SparkSubmit.scala, another thing is that I notice some suspicious code in SparkContext.addJar, Is the following code still needed ? https://github.com/apache/spark/blob/39e2bad6a866d27c3ca594d15e574a1da3ee84cc/core/src/main/scala/org/apache/spark/SparkContext.scala#L1710
```
 if (master == "yarn" && deployMode == "cluster") {
              // In order for this to work in yarn cluster mode the user must specify the
              // --addJars option to the client to upload the file into the distributed cache
              // of the AM to make it show up in the current working directory.
              val fileName = new Path(uri.getPath).getName()
              try {
                env.rpcEnv.fileServer.addJar(new File(fileName))
              } catch {
                case e: Exception =>
                  // For now just log an error but allow to go through so spark examples work.
                  // The spark examples don't really need the jar distributed since its also
                  // the app jar.
                  logError("Error adding jar (" + e + "), was the --addJars option used?")
                  null
              }
            } else {
```

jerryshao:
From the code what I can is that `spark.files` will not be set in yarn mode.

```scala
      OptionAssigner(args.files, LOCAL | STANDALONE | MESOS, ALL_DEPLOY_MODES,
        sysProp = "spark.files"),
```

zjffdu:
spark.files would still be passed to driver even in yarn-cluster if you check the following code. 
https://github.com/apache/spark/blob/7bf8a4049866b2ec7fdf0406b1ad0c3a12488645/core/src/main/scala/org/apache/spark/deploy/SparkSubmit.scala#L609
```
    // Load any properties specified through --conf and the default properties file
    for ((k, v) <- args.sparkProperties) {
      sysProps.getOrElseUpdate(k, v)
    }
```

tgravescs:
yeah so I think the fix here should be that we don't pass it through or don't handle it in sparkContext.  All the yarn stuff should go through the distributed cache.

mridulm:
I agree with Tom, for yarn cluster mode we should rely on distributed cache

vanzin:
If I understand the issue here, the problem is not `--files` but setting `spark.files` through `--conf` (or the config file). The former is translated into `spark.yarn.dist.files` by `SparkSubmit` but it seems the latter isn't, and to me that's the bug. So basically what Tom and Mridul have said.

I think the correct fix would be in either `SparkSubmit.scala` or YARN's `Client.scala` - probably the latter to keep the YARN-specific stuff in one place.

zjffdu:
that's correct, it is due to `spark.files`, jira has been updated.  Will update the PR soon.

core/src/main/scala/org/apache/spark/SparkContext.scala
```diff
@@ -1716,29 +1716,12 @@ class SparkContext(config: SparkConf) extends Logging {
...
-                  null
-              }
-            } else {
-              try {
-                env.rpcEnv.fileServer.addJar(new File(uri.getPath))
-              } catch {
-                case exc: FileNotFoundException =>
-                  logError(s"Jar not found at $path")
-                  null
-              }
```
zjffdu: These are obsoleted code IMO, so I remove them.

jerryshao:
I think this issue also existed in yarn client mode, files will be added twice (once with Spark's file server, another with distributed cache), because both file server and yarn#client could find this file, then there's no exception. But this will be failed in yarn cluster mode since spark file server could not find out file with specified path.

zjffdu:
That's correct, this PR will also fix the yarn-client case. PR title is updated.

vanzin:
One minor comment update, otherwise LGTM.

jerryshao:
I'm wondering is it enough to address issue in yarn client mode, since we will have to properties "spark.files" and "spark.yarn.dist.files", the former will be handled by SparkContext and the latter will be handled by yarn#client, while yarn#client is created after SparkContext is launched, so is it too late to remove "spark.conf" in yarn#client, also yarn#client only honors "spark.yarn.dist.files", this will still add files again.

zjffdu:
hmm, notice spark.files is still passed to SparkContext in yarn-client mode, seems I need to do that in SparkSubmit

vanzin:
I think it would be better to treat that as a separate issue. spark.files has always worked like this in client mode; the behavior being changed here fixes an actual bug in cluster mode.

I'd consider using the cache for spark.files in client mode a separate enhancement.

vanzin:
Merging to master / 2.0.

vanzin:
Looks like there's a branch-2.1 now, so I merge into that branch too.

PR #15671, opened 2016-10-28
[SPARK-18206][ML]Add instrumentation for MLP,NB,LDA,AFT,GLM,Isotonic,LiR

zhengruifeng:

add instrumentation for MLP,NB,LDA,AFT,GLM,Isotonic,LiR
## How was this patch tested?

local test in spark-shell


mllib/src/main/scala/org/apache/spark/ml/classification/DecisionTreeClassifier.scala
```diff
@@ -96,11 +96,13 @@ class DecisionTreeClassifier @Since("1.4.0") (
 
     val instr = Instrumentation.create(this, oldDataset)
     instr.logParams(params: _*)
+    instr.logNumClasses(numClasses)
```
sethah: These are already logged inside of `RandomForest.run`

zhengruifeng:
@sethah Thanks. I have revert those changes in Tree-Algos.

zhengruifeng:
@jkbradley @yanboliang Could you please make a reivew?

mllib/src/main/scala/org/apache/spark/ml/clustering/LDA.scala
```diff
@@ -896,6 +896,10 @@ class LDA @Since("1.6.0") (
   @Since("2.0.0")
   override def fit(dataset: Dataset[_]): LDAModel = {
     transformSchema(dataset.schema, logging = true)
+
+    val instr = Instrumentation.create(this, dataset)
+    instr.logParams(params : _*)
```
sethah: Likewise here we probably should not log `docConcentration`.

mllib/src/main/scala/org/apache/spark/ml/clustering/LDA.scala
```diff
@@ -913,7 +917,10 @@ class LDA @Since("1.6.0") (
       case m: OldDistributedLDAModel =>
         new DistributedLDAModel(uid, m.vocabSize, m, dataset.sparkSession, None)
     }
-    copyValues(newModel).setParent(this)
+
+    val model = copyValues(newModel).setParent(this)
+    instr.logSuccess(newModel)
```
sethah: minor: `instr.logSuccess(model)`

sethah:
Minor: would you mind changing the title to "Add instrumentation logs to _ML_ training algorithms". I initially thought this PR was adding them to the old MLlib API, so I think it's a bit more clear.

mllib/src/main/scala/org/apache/spark/ml/regression/AFTSurvivalRegression.scala
```diff
@@ -226,6 +226,10 @@ class AFTSurvivalRegression @Since("1.6.0") (@Since("1.6.0") override val uid: S
     val featuresStd = featuresSummarizer.variance.toArray.map(math.sqrt)
     val numFeatures = featuresStd.size
 
+    val instr = Instrumentation.create(this, dataset)
+    instr.logParams(params : _*)
```
sethah: maybe we should not log `quantileProbabilities`? I am not familiar with this algorithm so I'm not sure if these can ever be too large.

mllib/src/main/scala/org/apache/spark/ml/regression/AFTSurvivalRegression.scala
```diff
@@ -276,6 +280,7 @@ class AFTSurvivalRegression @Since("1.6.0") (@Since("1.6.0") override val uid: S
     val intercept = parameters(1)
     val scale = math.exp(parameters(0))
     val model = new AFTSurvivalRegressionModel(uid, coefficients, intercept, scale)
+    instr.logSuccess(model)
```
sethah: minor: move the `logSuccess` call after the `copyValues`. That way if we ever do something in `logSuccess` like logging parameters of the model, the call will reflect the copied parameters.

mllib/src/main/scala/org/apache/spark/ml/regression/GeneralizedLinearRegression.scala
```diff
@@ -284,6 +288,7 @@ class GeneralizedLinearRegression @Since("2.0.0") (@Since("2.0.0") override val
         .setParent(this))
     val trainingSummary = new GeneralizedLinearRegressionTrainingSummary(dataset, model,
       irlsModel.diagInvAtWA.toArray, irlsModel.numIterations, getSolver)
+    instr.logSuccess(model)
     model.setSummary(trainingSummary)
```
sethah: minor: move `logSuccess` after setSummary

mllib/src/main/scala/org/apache/spark/ml/regression/GeneralizedLinearRegression.scala
```diff
@@ -284,6 +288,7 @@ class GeneralizedLinearRegression @Since("2.0.0") (@Since("2.0.0") override val
         .setParent(this))
     val trainingSummary = new GeneralizedLinearRegressionTrainingSummary(dataset, model,
       irlsModel.diagInvAtWA.toArray, irlsModel.numIterations, getSolver)
+    instr.logSuccess(model)
```
sethah: We need to log success if we enter the `WeightedLeastSquares` branch above as well. Maybe we can refactor the code to use `if else` instead of an `if` with a `return`. That way we only need to log success in one place.

mllib/src/main/scala/org/apache/spark/ml/regression/LinearRegression.scala
```diff
@@ -392,6 +399,8 @@ class LinearRegression @Since("1.3.0") (@Since("1.3.0") override val uid: String
       model,
       Array(0D),
       objectiveHistory)
+
+    instr.logSuccess(model)
```
sethah: minor: move logSuccess after setSummary, here and elsewhere.

mllib/src/main/scala/org/apache/spark/ml/regression/AFTSurvivalRegression.scala
sethah: Also, would you mind changing the doc in `Instrumentation.logSuccess` from:

"Logs the successful completion of the training session and the value of the learned model." 

to

"Logs the successful completion of the training session."

Since we currently don't do anything other than log a string inside `logSuccess`.

sethah (commented):
Left mostly minor comments. Let's create a JIRA for other meta algorithms like `CrossValidator`, `TrainValidationSplit`, and `OneVSRest`. Thanks for doing this!

mllib/src/main/scala/org/apache/spark/ml/classification/MultilayerPerceptronClassifier.scala
```diff
@@ -234,8 +234,14 @@ class MultilayerPerceptronClassifier @Since("1.5.0") (
    * @return Fitted model
    */
   override protected def train(dataset: Dataset[_]): MultilayerPerceptronClassificationModel = {
+    val instr = Instrumentation.create(this, dataset)
+    instr.logParams(params : _*)
```
zhengruifeng: ok, I will update it here, and other algos which support a initalModel

zhengruifeng:
@sethah Thanks for your reviewing. I have made changes according to your comments. And I will create JIRAs for meta algos.

jkbradley:
I just made a subtask for the algs covered here.  Can you please link this PR to SPARK-18206 in the title instead of the umbrella task?

sethah:
So, if we consistently use `instr.logParams(params: _*)` (even in cases where it's acceptable) then we run the risk of adding some param in the future that could "overload" the logs (like initialModel). However, if we manually select the appropriate params to log, then we risk adding some other param in the future which we do want to log, but it never gets added. Both could be problematic.

For now, I think I lean towards manually selecting which params to log rather than logging all params. If we add more params later we will have to remember to add them to the logging. What are others' thoughts?

jkbradley:
Good point.  I agree with you about logging selected Params only.

zhengruifeng:
@sethah @jkbradley OK, I will link this pr to the subtask and change algos which use `instr.logParams(params: _*)` to explicitly select some params.

sethah:
IMO, the way we're doing this logging right now is unsustainable. It requires too much manual work. We can leave this discussion for a different JIRA, but what we could do is modify the `Instrumentation` class to just truncate the param value string after a certain number of characters. Then, we could even modify `Predictor.fit` to create Instrumentation and log all params. The only thing we'd have to do in the individual algorithms is log anything else - algorithm specific - that we want to add. I haven't tested this yet. @jkbradley @zhengruifeng  what do you think?

mllib/src/main/scala/org/apache/spark/ml/regression/GeneralizedLinearRegression.scala
```diff
@@ -270,7 +275,9 @@ class GeneralizedLinearRegression @Since("2.0.0") (@Since("2.0.0") override val
           .setParent(this))
       val trainingSummary = new GeneralizedLinearRegressionTrainingSummary(dataset, model,
         wlsModel.diagInvAtWA.toArray, 1, getSolver)
-      return model.setSummary(trainingSummary)
+      model.setSummary(trainingSummary)
+      instr.logSuccess(model)
+      return model
```
sethah: would you mind rewriting this if statement to:

````scala
val model = if (familyObj == Gaussian && linkObj == Identity) {
  ...
} else {
  ...
}
instr.logSuccess(model)
model
````

Then we can avoid the code duplication?

File: mllib/src/main/scala/org/apache/spark/ml/regression/AFTSurvivalRegression.scala [hunk 11]
```diff
@@ -227,7 +227,8 @@ class AFTSurvivalRegression @Since("1.6.0") (@Since("1.6.0") override val uid: S
     val numFeatures = featuresStd.size
 
     val instr = Instrumentation.create(this, dataset)
-    instr.logParams(params : _*)
+    instr.logParams(labelCol, featuresCol, censorCol, predictionCol, quantilesCol,
```
sethah: add `aggregationDepth`

mllib/src/main/scala/org/apache/spark/ml/regression/LinearRegression.scala
```diff
@@ -196,7 +196,8 @@ class LinearRegression @Since("1.3.0") (@Since("1.3.0") override val uid: String
     }
 
     val instr = Instrumentation.create(this, dataset)
-    instr.logParams(params : _*)
+    instr.logParams(labelCol, featuresCol, weightCol, predictionCol, solver, tol,
```
sethah: add `aggregationDepth`

sethah (commented):
A few minor comments otherwise LGTM. Still, we definitely need to find a way to make this more sustainable.

mllib/src/main/scala/org/apache/spark/ml/regression/GeneralizedLinearRegression.scala
zhengruifeng: OK, I will change this.

mllib/src/main/scala/org/apache/spark/ml/regression/AFTSurvivalRegression.scala
zhengruifeng: OK, Thanks

zhengruifeng:
@sethah I have make changes according to the comments. Thanks for your reviewing.

zhengruifeng:
@sethah I agree that manually listing traceable params is prone to mistake. I think we can log all params expect some params which are labeled `dont-log` in the individual algorithms. Or we can create a new methods `def logParams(params: Seq[Param], except: Seq[Param])`.

sethah:
I created [SPARK-18253](https://issues.apache.org/jira/browse/SPARK-18253) to track it. We may have to get to it after 2.1 QA period.

jkbradley:
I don't want to truncate Param strings because it would create invalid JSON in case people want to try to catch and parse the logs.  I like the idea of allowing exceptions and possibly adding unit tests to ensure the logs do not blow up.

sethah:
@jkbradley Thanks for bringing that up. I'm ok with alternate solutions provided they don't require someone to remember to manually add or manually except a new param, and that we can ensure that we aren't logging params that could blow up the logs. We can discuss it on the JIRA, thanks!

mllib/src/main/scala/org/apache/spark/ml/classification/GBTClassifier.scala
```diff
@@ -139,7 +139,10 @@ class GBTClassifier @Since("1.4.0") (
     val boostingStrategy = super.getOldBoostingStrategy(categoricalFeatures, OldAlgo.Classification)
 
     val instr = Instrumentation.create(this, oldDataset)
-    instr.logParams(params: _*)
+    instr.logParams(labelCol, featuresCol, predictionCol, impurity, lossType,
+                    maxDepth, maxBins, maxIter, maxMemoryInMB, minInfoGain,
```
jkbradley: style: indent 2 spaces, not to parenthesis

zhengruifeng:
@jkbradley  Updated

zhengruifeng:
ping @sethah @jkbradley

jkbradley:
Taking a look now!

mllib/src/main/scala/org/apache/spark/ml/regression/LinearRegression.scala
```diff
@@ -225,7 +230,7 @@ class LinearRegression @Since("1.3.0") (@Since("1.3.0") override val uid: String
         summaryModel,
         model.diagInvAtWA.toArray,
         model.objectiveHistory)
-
+      instr.logSuccess(lrModel)
```
jkbradley: Could you please move this to happen after setSummary?  Same for the others places below.

(The summary isn't used now, but I can imagine us wanting to log info about it when possible...in the future.)

mllib/src/main/scala/org/apache/spark/ml/regression/AFTSurvivalRegression.scala
```diff
@@ -227,6 +227,11 @@ class AFTSurvivalRegression @Since("1.6.0") (@Since("1.6.0") override val uid: S
     val featuresStd = featuresSummarizer.variance.toArray.map(math.sqrt)
     val numFeatures = featuresStd.size
 
+    val instr = Instrumentation.create(this, dataset)
+    instr.logParams(labelCol, featuresCol, censorCol, predictionCol, quantilesCol,
```
jkbradley: also log "quantileProbabilities"

mllib/src/main/scala/org/apache/spark/ml/clustering/LDA.scala
```diff
@@ -905,7 +911,10 @@ class LDA @Since("1.6.0") (
       case m: OldDistributedLDAModel =>
         new DistributedLDAModel(uid, m.vocabSize, m, dataset.sparkSession, None)
     }
-    copyValues(newModel).setParent(this)
+
+    val model = copyValues(newModel).setParent(this)
```
jkbradley: Also log numFeatures (model.vocabSize)

mllib/src/main/scala/org/apache/spark/ml/clustering/LDA.scala
```diff
@@ -888,6 +888,12 @@ class LDA @Since("1.6.0") (
   @Since("2.0.0")
   override def fit(dataset: Dataset[_]): LDAModel = {
     transformSchema(dataset.schema, logging = true)
+
+    val instr = Instrumentation.create(this, dataset)
+    instr.logParams(featuresCol, topicDistributionCol, k, maxIter, subsamplingRate,
+      checkpointInterval, keepLastCheckpoint, optimizeDocConcentration, topicConcentration,
```
jkbradley: missing "docConcentration"

jkbradley (commented):
This looks almost ready.  Thanks!

mllib/src/main/scala/org/apache/spark/ml/clustering/LDA.scala
zhengruifeng: I leave it out by intention. Because `docConcentration` may be a double arrary of size `k`, it maybe too large to log.

mllib/src/main/scala/org/apache/spark/ml/regression/AFTSurvivalRegression.scala
zhengruifeng: Ok, I will add it

mllib/src/main/scala/org/apache/spark/ml/regression/LinearRegression.scala
zhengruifeng: Ok, I will make sure `instr.logSuccess` happen after `model.setSummary`

mllib/src/main/scala/org/apache/spark/ml/clustering/LDA.scala
zhengruifeng: ok, I'll add it

zhengruifeng:
@jkbradley Update according to your comments, including adding `quantileProbabilities` and `docConcentration`.

zhengruifeng:
ping @jkbradley

jkbradley:
Thanks for the updates!

For docConcentration and quantileProbabilities, I agree it could be problematic if these are too large.  How about:
* We don't log docConcentration since that could easily be large and since it is less likely to cause errors which users will interpret as bugs.
* We log quantileProbabilities.size since a large array could cause problems.

zhengruifeng:
@jkbradley Updated. Thanks for reviewing!

zhengruifeng:
re-ping @jkbradley

jkbradley:
LGTM pending one more run of the tests
Thanks a lot @zhengruifeng !

Jenkins test this please

jkbradley:
Merging with master
Thanks!

PR #15672, opened 2016-10-28
[SPARK-18148][SQL] Misleading Error Message for Aggregation Without Window/GroupBy

jiangxb1987:

Aggregation Without Window/GroupBy expressions will fail in `checkAnalysis`, the error message is a bit misleading, we should generate a more specific error message for this case.

For example,

```
spark.read.load("/some-data")
  .withColumn("date_dt", to_date($"date"))
  .withColumn("year", year($"date_dt"))
  .withColumn("week", weekofyear($"date_dt"))
  .withColumn("user_count", count($"userId"))
  .withColumn("daily_max_in_week", max($"user_count").over(weeklyWindow))
)
```

creates the following output:

```
org.apache.spark.sql.AnalysisException: expression '`randomColumn`' is neither present in the group by, nor is it an aggregate function. Add to group by or wrap in first() (or first_value) if you don't care which value you get.;
```

In the error message above, `randomColumn` doesn't appear in the query(acturally it's added by function `withColumn`), so the message is not enough for the user to address the problem.
## How was this patch tested?

Manually test

Before:

```
scala> spark.sql("select col, count(col) from tbl")
org.apache.spark.sql.AnalysisException: expression 'tbl.`col`' is neither present in the group by, nor is it an aggregate function. Add to group by or wrap in first() (or first_value) if you don't care which value you get.;;
```

After:

```
scala> spark.sql("select col, count(col) from tbl")
org.apache.spark.sql.AnalysisException: grouping expressions sequence is empty, and 'tbl.`col`' is not an aggregate function. Wrap '(count(col#231L) AS count(col)#239L)' in windowing function(s) or wrap 'tbl.`col`' in first() (or first_value) if you don't care which value you get.;;
```

Also add new test sqls in `group-by.sql`.


rxin:
Is this confusing msg reproducible using SQL queries (not dataframe APIs)? If yes, perhaps we can add a test in SQLQueryTestSuite.

jiangxb1987:
@rxin I've added new test sqls in `group-by.sql`, which covers the both error message generated by `checkAnalysis`. I also moved the test cases for `Aggregate` operator from `SQLQuerySuite` to `group-by.sql`.

jiangxb1987:
The failed test case looks not related to our change here. It didn't fail in my local envirement.

jiangxb1987:
retest this please

sql/core/src/test/resources/sql-tests/inputs/group-by.sql
```diff
@@ -15,3 +15,26 @@ select 'foo', approx_count_distinct(int_col) from myview where int_col == 0 grou
...
+-- Test data.
+CREATE OR REPLACE TEMPORARY VIEW testData AS SELECT * FROM VALUES
+(1, 1), (1, 2), (2, 1), (2, 2), (3, 1), (3, 2), (null, 1), (3, null), (null, null)
+AS testData(a, b);
+
+-- Aggregate with empty GroupBy expressions.
+SELECT a, COUNT(b) FROM testData;
+SELECT COUNT(a), COUNT(b) FROM testData;
+
+-- Aggregate with non-empty GroupBy expressions.
```
rxin: these are already tested earlier ain't they?

jiangxb1987:
@rxin I've updated test sqls to use one dataset. Thank you!

rxin:
Thanks - merging in master.

PR #15673, opened 2016-10-28
[SPARK-17992][SQL] Return all partitions from HiveShim when Hive throws a metastore exception when attempting to fetch partitions by filter

mallman:
(Link to Jira issue: https://issues.apache.org/jira/browse/SPARK-17992)

We recently added table partition pruning for partitioned Hive tables converted to using `TableFileCatalog`. When the Hive configuration option `hive.metastore.try.direct.sql` is set to `false`, Hive will throw an exception for unsupported filter expressions. For example, attempting to filter on an integer partition column will throw a `org.apache.hadoop.hive.metastore.api.MetaException`.

I discovered this behavior because VideoAmp uses the CDH version of Hive with a Postgresql metastore DB. In this configuration, CDH sets `hive.metastore.try.direct.sql` to `false` by default, and queries that filter on a non-string partition column will fail.

Rather than throw an exception in query planning, this patch catches this exception, logs a warning and returns all table partitions instead. Clients of this method are already expected to handle the possibility that the filters will not be honored.
## How was this patch tested?

A unit test was added.

### DIFF
--- a/sql/hive/src/main/scala/org/apache/spark/sql/hive/client/HiveShim.scala
+++ b/sql/hive/src/main/scala/org/apache/spark/sql/hive/client/HiveShim.scala
@@ -29,7 +29,7 @@ import scala.util.control.NonFatal
 
 import org.apache.hadoop.fs.{FileSystem, Path}
 import org.apache.hadoop.hive.conf.HiveConf
-import org.apache.hadoop.hive.metastore.api.{Function => HiveFunction, FunctionType, NoSuchObjectException, PrincipalType, ResourceType, ResourceUri}
+import org.apache.hadoop.hive.metastore.api.{Function => HiveFunction, FunctionType, MetaException, PrincipalType, ResourceType, ResourceUri}
 import org.apache.hadoop.hive.ql.Driver
 import org.apache.hadoop.hive.ql.metadata.{Hive, HiveException, Partition, Table}
 import org.apache.hadoop.hive.ql.plan.AddPartitionDesc
@@ -587,16 +587,16 @@ private[client] class Shim_v0_13 extends Shim_v0_12 {
       } else {
         logDebug(s"Hive metastore filter is '$filter'.")
         try {
+          // Hive may throw an exception when calling this method in some circumstances, such as
+          // when filtering on a non-string partition column when the hive config key
+          // hive.metastore.try.direct.sql is false
           getPartitionsByFilterMethod.invoke(hive, table, filter)
             .asInstanceOf[JArrayList[Partition]]
         } catch {
-          case e: InvocationTargetException =>
-            // SPARK-18167 retry to investigate the flaky test. This should be reverted before
-            // the release is cut.
-            val retry = Try(getPartitionsByFilterMethod.invoke(hive, table, filter))
-            logError("getPartitionsByFilter failed, retry success = " + retry.isSuccess)
-            logError("all partitions: " + getAllPartitions(hive, table))
-            throw e
+          case ex: InvocationTargetException if ex.getCause.isInstanceOf[MetaException] =>
+            logWarning("Caught MetaException attempting to get partitions by filter from Hive", ex)
+            // Return all partitions. HiveShim clients are expected to handle this possibility
+            getAllPartitionsMethod.invoke(hive, table).asInstanceOf[JSet[Partition]]
         }
       }
 
--- a/sql/hive/src/main/scala/org/apache/spark/sql/hive/client/HiveShim.scala
+++ b/sql/hive/src/main/scala/org/apache/spark/sql/hive/client/HiveShim.scala
@@ -601,15 +601,16 @@ private[client] class Shim_v0_13 extends Shim_v0_12 {
               !tryDirectSql =>
             logWarning("Caught Hive MetaException attempting to get partition metadata by " +
               "filter from Hive. Falling back to fetching all partition metadata, which will " +
-              "degrade performance. Consider modifying your Hive metastore configuration to " +
-              s"set ${tryDirectSqlConfVar.varname} to true.", ex)
+              "degrade performance. Modifying your Hive metastore configuration to set " +
+              s"${tryDirectSqlConfVar.varname} to true may resolve this problem.", ex)
             // HiveShim clients are expected to handle a superset of the requested partitions
             getAllPartitionsMethod.invoke(hive, table).asInstanceOf[JSet[Partition]]
           case ex: InvocationTargetException if ex.getCause.isInstanceOf[MetaException] &&
               tryDirectSql =>
             throw new RuntimeException("Caught Hive MetaException attempting to get partition " +
-              "metadata by filter from Hive. Set the Spark configuration setting " +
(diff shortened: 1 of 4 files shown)


sql/hive/src/main/scala/org/apache/spark/sql/hive/client/HiveShim.scala
```diff
@@ -585,7 +585,18 @@ private[client] class Shim_v0_13 extends Shim_v0_12 {
...
-        getPartitionsByFilterMethod.invoke(hive, table, filter).asInstanceOf[JArrayList[Partition]]
+        try {
+          // Hive may throw an exception when calling this method in some circumstances, such as
+          // when filtering on a non-string partition column when the hive config key
+          // hive.metastore.try.direct.sql is false
+          getPartitionsByFilterMethod.invoke(hive, table, filter)
+            .asInstanceOf[JArrayList[Partition]]
+        } catch {
+          case ex: InvocationTargetException if ex.getCause.isInstanceOf[MetaException] =>
+            logWarning("Caught MetaException attempting to get partitions by filter from Hive", ex)
```
rxin: change the msg to say we are falling back to fetch all partitions' medatadata?

rxin:
cc @ericl

ericl:
Could we enable this fallback only when the conf is set to false? Otherwise, it might mask legitimate bugs.

I also wonder if some of our flaky tests around this issue are due to the conf being leaked by some suites...

mallman:
> Could we enable this fallback only when the conf is set to false? Otherwise, it might mask legitimate bugs.

Certainly, but my intent with this PR is to prevent a (painful and confusing) regression for some Hive users of Spark 2.1 which can occur, because Spark 2.1 enables our new partition pruning implementation by default. I mentioned one case where this will happen, but we can't be sure this is the only case. If we make the conditions under which we use a fallback too narrow, we are assuming that other configurations of Hive are compatible with partition pruning outside of the specific conditions we check. I think that's a bit too risky. In fact, before submitting this PR I had written the catch block to catch and fall back for all types of `Exception`. What I ended up with here is a middle ground.

mallman:
The current merge conflict is from d2d438d1d549628a0183e468ed11d6e85b5d6061, which touches the same code. I'll wait for that to be settled before rebasing.

ericl:
For large tables, the degraded performance should be considered a bug as well.

How about this.
1) If direct sql is disabled, log a warning about degraded performance with this flag and fall back to fetching all partitions.
2) If direct sql is enabled, crash with a message suggesting to disable filesource partition management and report a bug.

That way, we will know if there are cases where metastore pruning fails with direct sql enabled.

mallman:
@ericl I've pushed a commit with the changes you recommended.

mallman:
It looks like all the unit tests passed, however one of the forked test java processes exited with nonzero status for some unknown reason.

sql/hive/src/main/scala/org/apache/spark/sql/hive/client/HiveShim.scala
```diff
@@ -585,7 +586,31 @@ private[client] class Shim_v0_13 extends Shim_v0_12 {
...
+            logWarning("Caught Hive MetaException attempting to get partition metadata by " +
+              "filter from Hive. Falling back to fetching all partition metadata, which will " +
+              "degrade performance. Consider modifying your Hive metastore configuration to " +
+              s"set ${tryDirectSqlConfVar.varname} to true.", ex)
+            // HiveShim clients are expected to handle a superset of the requested partitions
+            getAllPartitionsMethod.invoke(hive, table).asInstanceOf[JSet[Partition]]
+          case ex: InvocationTargetException if ex.getCause.isInstanceOf[MetaException] &&
+              tryDirectSql =>
+            throw new RuntimeException("Caught Hive MetaException attempting to get partition " +
+              "metadata by filter from Hive. Set the Spark configuration setting " +
```
mallman: Good point.

mallman: I made some revisions. LMK what you think.

ericl:
This looks good to me. cc @cloud-fan

cloud-fan:
LGTM

rxin:
@mallman can you bring this up-to-date?

mallman:
@rxin I believe https://issues.apache.org/jira/browse/SPARK-18168 will need to be resolved before I can rebase this PR.

ericl:
@mallman shall we go ahead and revert that in this PR? It didn't help with debugging the flaky test much.

mallman:
@ericl I can do that, yes. I'm current tied down. I will push a new commit later today or tonight.

mallman:
Rebased.

rxin:
Merging in master. Thanks.

mallman:
Happy to help.

sql/hive/src/main/scala/org/apache/spark/sql/hive/client/HiveShim.scala
```diff
@@ -586,17 +587,31 @@ private[client] class Shim_v0_13 extends Shim_v0_12 {
...
+              !tryDirectSql =>
+            logWarning("Caught Hive MetaException attempting to get partition metadata by " +
+              "filter from Hive. Falling back to fetching all partition metadata, which will " +
+              "degrade performance. Modifying your Hive metastore configuration to set " +
+              s"${tryDirectSqlConfVar.varname} to true may resolve this problem.", ex)
+            // HiveShim clients are expected to handle a superset of the requested partitions
+            getAllPartitionsMethod.invoke(hive, table).asInstanceOf[JSet[Partition]]
+          case ex: InvocationTargetException if ex.getCause.isInstanceOf[MetaException] &&
+              tryDirectSql =>
+            throw new RuntimeException("Caught Hive MetaException attempting to get partition " +
```
rezasafi: @mallman sorry to disturb you here, but what is the reason that when direct sql isn't set only a warning is logged?and why when direct sql is set a runtime exception is being raised instead of just a warning like no direct sql case?

mallman: Hi @rezasafi

I believe the reasoning is if the user has disabled direct sql, we will try to fetch the partitions for the requested partition predicate anyway. However, since we don't expect that call to succeed, we just log a warning and fallback to the legacy behavior.

On the other hand, if the user has enabled direct sql, then we expect the call to Hive to succeed. If it fails, we consider that an error and throw an exception.

I hope that helps clarify things.

rezasafi: Thank you very much for the explanation @mallman. I appreciate it.

kmanamcheri: @mallman Your assumption is incorrect. If Hive on direct sql fails, it will retry with ORM. So in this case, I am able to reproduce a issue with postgres where direct sql fails and if it retries with ORM, spark fails! Hive has fallback behavior for direct sql. 

Filed SPARK-25561

PR #15674, opened 2016-10-28
[SPARK-18164][SQL]ForeachSink should fail the Spark job if `process` throws exception

zsxwing:

Fixed the issue that ForeachSink didn't rethrow the exception.
## How was this patch tested?

The fixed unit test.

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/ForeachSink.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/ForeachSink.scala
@@ -77,6 +77,7 @@ class ForeachSink[T : Encoder](writer: ForeachWriter[T]) extends Sink with Seria
           case e: Throwable =>
             isFailed = true
             writer.close(e)
+            throw e
         }
         if (!isFailed) {
           writer.close(null)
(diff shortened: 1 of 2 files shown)


zsxwing:
cc @tdas

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/ForeachSink.scala
```diff
@@ -77,6 +77,7 @@ class ForeachSink[T : Encoder](writer: ForeachWriter[T]) extends Sink with Seria
           case e: Throwable =>
             isFailed = true
             writer.close(e)
+            throw e
         }
         if (!isFailed) {
```
tdas: I think there is a better of write this now using `try...catch...finally`. Because the condition here will always be true

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/ForeachSink.scala
```diff
@@ -68,19 +68,16 @@ class ForeachSink[T : Encoder](writer: ForeachWriter[T]) extends Sink with Seria
       }
     datasetWithIncrementalExecution.foreachPartition { iter =>
       if (writer.open(TaskContext.getPartitionId(), batchId)) {
-        var isFailed = false
         try {
           while (iter.hasNext) {
             writer.process(iter.next())
           }
         } catch {
           case e: Throwable =>
```
zsxwing: cannot access `e` inside finally.

tdas: oops. my bad.

tdas:
LGTM pending tests.

zsxwing:
Merging to master and 2.0. Thanks, @tdas

PR #15675, opened 2016-10-28
[SPARK-18144][SQL] logging StreamingQueryListener$QueryStartedEvent

CodingCat:

The PR fixes the bug that the QueryStartedEvent is not logged

the postToAll() in the original code is actually calling StreamingQueryListenerBus.postToAll() which has no listener at all....we shall post by sparkListenerBus.postToAll(s) and this.postToAll() to trigger local listeners as well as the listeners registered in LiveListenerBus

@zsxwing 
## How was this patch tested?

The following snapshot shows that QueryStartedEvent has been logged correctly

image

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamingQueryListenerBus.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamingQueryListenerBus.scala
@@ -39,12 +39,7 @@ class StreamingQueryListenerBus(sparkListenerBus: LiveListenerBus)
    * be dispatched to all StreamingQueryListener in the thread of the Spark listener bus.
    */
   def post(event: StreamingQueryListener.Event) {
-    event match {
-      case s: QueryStartedEvent =>
-        postToAll(s)
-      case _ =>
-        sparkListenerBus.post(event)
-    }
+    sparkListenerBus.post(event)
   }
 
   override def onOtherEvent(event: SparkListenerEvent): Unit = {
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamingQueryListenerBus.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamingQueryListenerBus.scala
@@ -52,7 +52,14 @@ class StreamingQueryListenerBus(sparkListenerBus: LiveListenerBus)
   override def onOtherEvent(event: SparkListenerEvent): Unit = {
     event match {
       case e: StreamingQueryListener.Event =>
-        postToAll(e)
+        // SPARK-18144: we broadcast QueryStartedEvent to all listeners attached to this bus
+        // synchronously and to listeners attached to LiveListenerBus asynchronously. Therefore,
+        // we need to ignore QueryStartedEvent if this method is called within SparkListenerBus
+        // thread
+        if (Thread.currentThread().getName != "SparkListenerBus" ||
+          !e.isInstanceOf[QueryStartedEvent]) {
+          postToAll(e)
+        }
       case _ =>
     }
   }
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamingQueryListenerBus.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamingQueryListenerBus.scala
@@ -53,11 +53,10 @@ class StreamingQueryListenerBus(sparkListenerBus: LiveListenerBus)
     event match {
       case e: StreamingQueryListener.Event =>
         // SPARK-18144: we broadcast QueryStartedEvent to all listeners attached to this bus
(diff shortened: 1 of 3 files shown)


sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamingQueryListenerBus.scala
```diff
@@ -39,12 +39,7 @@ class StreamingQueryListenerBus(sparkListenerBus: LiveListenerBus)
    * be dispatched to all StreamingQueryListener in the thread of the Spark listener bus.
    */
   def post(event: StreamingQueryListener.Event) {
-    event match {
```
CodingCat: actually I didn't see any reason to distinguish these two types of event when posting

zsxwing:
It's intentional. Just to provide a synchronous callback for the `start` event... It's documented in `StreamingQueryListener`. This issue now is not really a big issue since no place use it now.

zsxwing:
Ideally, it should not use the listener api for a synchronous callback. But right now I don't have any good API design for this one.

CodingCat:
Sorry, I misunderstood the real reason of the bug

The real reason is actually very simple...the original postToAll() is actually calling `StreamingQueryListenerBus.postToAll()` which has no listener at all....we shall post by `sparkListenerBus.postToAll(s)` and `this.postToAll()` to trigger local listeners as well as the listeners registered in LiveListenerBus

CodingCat:
@zsxwing mind looking at it again?

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamingQueryListenerBus.scala
```diff
@@ -41,7 +41,9 @@ class StreamingQueryListenerBus(sparkListenerBus: LiveListenerBus)
   def post(event: StreamingQueryListener.Event) {
     event match {
       case s: QueryStartedEvent =>
-        postToAll(s)
+        sparkListenerBus.postToAll(s)
```
zsxwing: This may break the existing spark listeners because they are not thread-safe.

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamingQueryListenerBus.scala
```diff
@@ -41,6 +41,8 @@ class StreamingQueryListenerBus(sparkListenerBus: LiveListenerBus)
   def post(event: StreamingQueryListener.Event) {
     event match {
       case s: QueryStartedEvent =>
+        sparkListenerBus.post(s)
```
zsxwing: `QueryStartedEvent` will be sent twice to `StreamingQueryListener`. There is a thread local variable in `LiveListenerBus`. You can use it to ignore duplicated  `QueryStartedEvent` if it's posted in the listener thread.

zsxwing: Since this is hacky, it's better to have some tests to make sure we won't break things in future when removing the hacky codes.

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamingQueryListenerBus.scala
```diff
@@ -50,7 +52,14 @@ class StreamingQueryListenerBus(sparkListenerBus: LiveListenerBus)
   override def onOtherEvent(event: SparkListenerEvent): Unit = {
     event match {
       case e: StreamingQueryListener.Event =>
-        postToAll(e)
+        // SPARK-18144: we broadcast QueryStartedEvent to all listeners attached to this bus
+        // synchronously and to listeners attached to LiveListenerBus asynchronously. Therefore,
+        // we need to ignore QueryStartedEvent if this method is called within SparkListenerBus
+        // thread
+        if (Thread.currentThread().getName != "SparkListenerBus" ||
```
zsxwing: nit: please use `!LiveListenerBus.withinListenerThread.value` instead.

sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamingQuerySuite.scala
```diff
@@ -290,7 +290,10 @@ class StreamingQuerySuite extends StreamTest with BeforeAndAfter with Logging {
     // A StreamingQueryListener that gets the query status after the first completed trigger
     val listener = new StreamingQueryListener {
       @volatile var firstStatus: StreamingQueryStatus = null
-      override def onQueryStarted(queryStarted: QueryStartedEvent): Unit = { }
```
zsxwing: nit: please add `@volatile`

zsxwing (changes requested):
Looks good. Just some nits

zsxwing:
LGTM. Thanks! Merging to master, 2.1 and 2.0.

CodingCat:
thanks

PR #15676, opened 2016-10-28
[SPARK-18167] [SQL] Add debug code for SQLQuerySuite flakiness when metastore partition pruning is enabled

ericl:

org.apache.spark.sql.hive.execution.SQLQuerySuite is flaking when hive partition pruning is enabled.
Based on the stack traces, it seems to be an old issue where Hive fails to cast a numeric partition column ("Invalid character string format for type DECIMAL"). There are two possibilities here: either we are somehow corrupting the partition table to have non-decimal values in that column, or there is a transient issue with Derby.

This PR logs the result of the retry when this exception is encountered, so we can confirm what is going on.
## How was this patch tested?

n/a

cc @yhuai

### DIFF
--- a/sql/hive/src/main/scala/org/apache/spark/sql/hive/client/HiveShim.scala
+++ b/sql/hive/src/main/scala/org/apache/spark/sql/hive/client/HiveShim.scala
@@ -593,7 +593,9 @@ private[client] class Shim_v0_13 extends Shim_v0_12 {
           case e: InvocationTargetException =>
             // SPARK-18167 retry to investigate the flaky test
             val retry = Try(getPartitionsByFilterMethod.invoke(hive, table, filter))
+            val full = Try(getAllPartitionsMethod.invoke(hive, table))
             logError("getPartitionsByFilter failed, retry success = " + retry.isSuccess, e)
+            logError("getPartitionsByFilter failed, full fetch success = " + full.isSuccess, e)
             throw e
         }
       }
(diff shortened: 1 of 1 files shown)


sql/hive/src/main/scala/org/apache/spark/sql/hive/client/HiveShim.scala
```diff
@@ -585,7 +586,19 @@ private[client] class Shim_v0_13 extends Shim_v0_12 {
...
+          getPartitionsByFilterMethod.invoke(hive, table, filter)
+            .asInstanceOf[JArrayList[Partition]]
+        } catch {
+          case e: InvocationTargetException =>
+            // SPARK-18167 retry to investigate the flaky test. This should be reverted before
+            // the release is cut.
+            val retry = Try(getPartitionsByFilterMethod.invoke(hive, table, filter))
+            val full = Try(getAllPartitionsMethod.invoke(hive, table))
+            logError("getPartitionsByFilter failed, retry success = " + retry.isSuccess)
+            logError("getPartitionsByFilter failed, full fetch success = " + full.isSuccess)
```
yhuai: I am wondering if we should also log all of partition specs?

rxin:
Merging in first so we can get more signal.

sql/hive/src/main/scala/org/apache/spark/sql/hive/client/HiveShim.scala
ericl: Yeah that's good idea, we should definitely do it if turns out the retry fails

PR #15681, opened 2016-10-29
[SPARK-18176][Streaming][Kafka] Kafka010 .createRDD() scala API should expect scala Map

lw-lin:

Thoughout `external/kafka-010`, Java APIs are expecting `java.util.Map`s and Scala APIs are expecting `scala.collection.Map`s, with the exception of `KafkaUtils.createRDD()` Scala API expecting a `java.util.Map`.

This patch has added another createRDD overload which takes a `scala.collection.Map`.

## How was this patch tested?

Existing test cases & added new basic test case.

### DIFF
--- a/external/kafka-0-10/src/main/scala/org/apache/spark/streaming/kafka010/DirectKafkaInputDStream.scala
+++ b/external/kafka-0-10/src/main/scala/org/apache/spark/streaming/kafka010/DirectKafkaInputDStream.scala
@@ -44,6 +44,11 @@ import org.apache.spark.streaming.scheduler.rate.RateEstimator
  * per second that each '''partition''' will accept.
  * @param locationStrategy In most cases, pass in [[PreferConsistent]],
  *   see [[LocationStrategy]] for more details.
+ * @param executorKafkaParams Kafka
+ * <a href="http://kafka.apache.org/documentation.html#newconsumerconfigs">
+ * configuration parameters</a>.
+ *   Requires  "bootstrap.servers" to be set with Kafka broker(s),
+ *   NOT zookeeper servers, specified in host1:port1,host2:port2 form.
  * @param consumerStrategy In most cases, pass in [[Subscribe]],
  *   see [[ConsumerStrategy]] for more details
  * @tparam K type of Kafka message key
--- a/external/kafka-0-10/src/main/scala/org/apache/spark/streaming/kafka010/DirectKafkaInputDStream.scala
+++ b/external/kafka-0-10/src/main/scala/org/apache/spark/streaming/kafka010/DirectKafkaInputDStream.scala
@@ -44,11 +44,6 @@ import org.apache.spark.streaming.scheduler.rate.RateEstimator
  * per second that each '''partition''' will accept.
  * @param locationStrategy In most cases, pass in [[PreferConsistent]],
  *   see [[LocationStrategy]] for more details.
- * @param executorKafkaParams Kafka
- * <a href="http://kafka.apache.org/documentation.html#newconsumerconfigs">
- * configuration parameters</a>.
- *   Requires  "bootstrap.servers" to be set with Kafka broker(s),
- *   NOT zookeeper servers, specified in host1:port1,host2:port2 form.
  * @param consumerStrategy In most cases, pass in [[Subscribe]],
  *   see [[ConsumerStrategy]] for more details
  * @tparam K type of Kafka message key
--- a/external/kafka-0-10/src/main/scala/org/apache/spark/streaming/kafka010/KafkaUtils.scala
+++ b/external/kafka-0-10/src/main/scala/org/apache/spark/streaming/kafka010/KafkaUtils.scala
@@ -27,6 +27,7 @@ import org.apache.kafka.common.TopicPartition
 import org.apache.spark.SparkContext
 import org.apache.spark.annotation.Experimental
 import org.apache.spark.api.java.{ JavaRDD, JavaSparkContext }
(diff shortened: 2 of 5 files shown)


external/kafka-0-10/src/main/scala/org/apache/spark/streaming/kafka010/DirectKafkaInputDStream.scala
```diff
@@ -44,11 +44,6 @@ import org.apache.spark.streaming.scheduler.rate.RateEstimator
  * per second that each '''partition''' will accept.
  * @param locationStrategy In most cases, pass in [[PreferConsistent]],
  *   see [[LocationStrategy]] for more details.
- * @param executorKafkaParams Kafka
- * <a href="http://kafka.apache.org/documentation.html#newconsumerconfigs">
- * configuration parameters</a>.
- *   Requires  "bootstrap.servers" to be set with Kafka broker(s),
- *   NOT zookeeper servers, specified in host1:port1,host2:port2 form.
```
lw-lin: we don't have a `executorKafkaParams` param at all; same below

external/kafka-0-10/src/main/scala/org/apache/spark/streaming/kafka010/KafkaUtils.scala
```diff
@@ -56,32 +57,18 @@ object KafkaUtils extends Logging {
   @Experimental
   def createRDD[K, V](
       sc: SparkContext,
-      kafkaParams: ju.Map[String, Object],
+      kafkaParams: collection.Map[String, Object],
```
lw-lin: this is the change

external/kafka-0-10/src/main/scala/org/apache/spark/streaming/kafka010/KafkaUtils.scala
```diff
@@ -56,32 +57,18 @@ object KafkaUtils extends Logging {
...
-          "If you want to prefer brokers, you must provide a mapping using PreferFixed " +
-          "A single KafkaRDD does not have a driver consumer and cannot look up brokers for you.")
-      case PreferConsistent => ju.Collections.emptyMap[TopicPartition, String]()
-      case PreferFixed(hostMap) => hostMap
-    }
-    val kp = new ju.HashMap[String, Object](kafkaParams)
-    fixKafkaParams(kp)
-    val osr = offsetRanges.clone()
-
-    new KafkaRDD[K, V](sc, kp, osr, preferredHosts, true)
```
lw-lin: all is exacted into a `private createRDDInternal()` method untouched

external/kafka-0-10/src/main/scala/org/apache/spark/streaming/kafka010/KafkaUtils.scala
lw-lin: this is public API change but I think it would be OK

lw-lin:
@koeninger could you also take a look at this, thanks!

koeninger:
Public API changes even on an experimental module aren't a minor thing, open a Jira.

It's also better not to lump unrelated changes together.

lw-lin:
@koeninger well received and thanks! JIRA opened and unrelated changes removed.

Let's see what Jenkins would say.

koeninger:
I think this should just be another createRDD overload that takes a scala map.  The minor additional maintenance overhead of that method as opposed to change the existing one isn't worth breaking clients.  I'm all in favor of changes to experimental apis when they are necessary, but this isn't necessary.

That also makes most of the other changes in this PR unnecessary - you don't need to touch mima, you don't need to refactor to createRddInternal, etc etc.

lw-lin:
Reverted all changes and added another createRDD overload ! :-)

koeninger:
LGTM, thanks.

If you want to open a separate PR to cleanup the private doc issues you noticed, go for it, shouldn't need another Jira imho if it isn't changing code

lw-lin:
Thank you @koeninger !

Please let me cc @srowen who's been around to also take a look~

srowen:
Should we then deprecate the existing method this is 'replacing', if it's not going to be removed? the message I get is that the existing one shouldn't really be used. Otherwise it's hard to understand why both exist when looking at the API.

lw-lin:
Deprecating the existing one would mean we still need to introduce some `createRDDInternal`, and let the deprecated one call it; then we can just remove the deprecated one some time in the future?

I don't have strong preferences here :)

srowen:
We would want to change the code so that nothing calls the deprecated method including tests. I think you can let the deprecated one call the new one? When to remove it is a question for the future, yes.

koeninger:
I don't think there's a reason to deprecate it.  ju.Map is the lowest
common denominator for kafka params, it's used by the underlying consumer,
and it's what the ConsumerStrategy interface expects to return.  I would
definitely prefer to use it in my Scala jobs that were generating a
sequence of RDDs.

srowen:
I see, so it's not an 'oversight'. If you're saying you prefer a Java Map in this API, is there much value in an overload for a Scala Map? the conversion is just one method call.

koeninger:
I'm agnostic on the value of adding the overload, if @lw-lin thinks it's more convenient for users.  There are considerably fewer overloads as it stands than the old 0.8 version of KafkaUtils, so I don't think we're getting too messy/crowded yet.

srowen:
@lw-lin are you persuaded or still feel it's fairly valuable for users? I guess I'm neutral too.

lw-lin:
I'm closing this for now; can re-open if other people should demand this.  thanks!

PR #15682, opened 2016-10-29
[SPARK-18169][SQL] Suppress warnings when dropping views on a dropped table

dongjoon-hyun:

Apache Spark 2.x shows an **AnalysisException** warning message when dropping a **view** on a dropped table. This does not happen on dropping **temporary views**. Also, Spark 1.6.x does not show warnings. We had better suppress this to be more consistent in Spark 2.x and with Spark 1.6.x.

``` scala
scala> sql("create table t(a int)")

scala> sql("create view v as select * from t")

scala> sql("create temporary view tv as select * from t")

scala> sql("drop table t")

scala> sql("drop view tv")

scala> sql("drop view v")
16/10/29 15:50:03 WARN DropTableCommand: org.apache.spark.sql.AnalysisException: Table or view not found: `default`.`t`; line 1 pos 91;
'SubqueryAlias v, `default`.`v`
+- 'Project ['gen_attr_0 AS a#19]
   +- 'SubqueryAlias t
      +- 'Project ['gen_attr_0]
         +- 'SubqueryAlias gen_subquery_0
            +- 'Project ['a AS gen_attr_0#18]
               +- 'UnresolvedRelation `default`.`t`

org.apache.spark.sql.AnalysisException: Table or view not found: `default`.`t`; line 1 pos 91;
'SubqueryAlias v, `default`.`v`
+- 'Project ['gen_attr_0 AS a#19]
   +- 'SubqueryAlias t
      +- 'Project ['gen_attr_0]
         +- 'SubqueryAlias gen_subquery_0
            +- 'Project ['a AS gen_attr_0#18]
               +- 'UnresolvedRelation `default`.`t`
...
res5: org.apache.spark.sql.DataFrame = []
```

Note that this is different case of dropping non-exist views. For those cases, Spark raises **NoSuchTableException**.
## How was this patch tested?

Manual because this is a suppression of warning message.

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
@@ -202,6 +202,7 @@ case class DropTableCommand(
       sparkSession.sharedState.cacheManager.uncacheQuery(
         sparkSession.table(tableName.quotedString))
     } catch {
+      case ae: AnalysisException if ae.getMessage.contains("Table or view not found") =>
       case NonFatal(e) => log.warn(e.toString, e)
     }
     catalog.refreshTable(tableName)


dongjoon-hyun:
Retest this please.

dongjoon-hyun:
Hmm. `randomized aggregation test` fails consistently.
```
- randomized aggregation test - [typed, with distinct] - with grouping keys - with non-empty input
```

dongjoon-hyun:
Retest this please

dongjoon-hyun:
Retest this please.

dongjoon-hyun:
Hi, @gatorsmile .
How do you think about this PR? I can close this if you don't feel strongly.

sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
```diff
@@ -202,6 +202,7 @@ case class DropTableCommand(
       sparkSession.sharedState.cacheManager.uncacheQuery(
         sparkSession.table(tableName.quotedString))
     } catch {
+      case ae: AnalysisException if ae.getMessage.contains("Table or view not found") =>
```
gatorsmile: Is it `NoSuchTableException`?

dongjoon-hyun: Thank you for review, @gatorsmile .
It's originally `NoSuchTableException`, but `Analyzer` changed into this at [here](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala#L516-L517).
So, the catch expression becomes ugly.

gatorsmile:
Based on my understanding, when we try to drop a view, we [looking up and analyzing the view](https://github.com/dongjoon-hyun/spark/blob/fef9981ac140112c05f40c093b2174d1584caaf9/sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala#L203) and then issue an exception if any involved table has been dropped. We might have a bug here.

That means, we will not call `uncacheQuery`, if any involved table is dropped before the view. It is a bug, right?

dongjoon-hyun:
Sorry, but `uncacheQuery` is related to this PR?

dongjoon-hyun:
Maybe, do you mean #15896 ?

dongjoon-hyun:
Ah, you mean that code you gave the pointer.

dongjoon-hyun:
Yep. `uncachedQuery` is not called here. Yep. I got it. It's a bug.

gatorsmile:
This case is more complicated. Please try to check whether there is a bug. If yes, please fix it in this PR. `Suppress warnings` might be easily fixed when you fix this bug.

dongjoon-hyun:
Yep. I investigate the behavior here.

gatorsmile:
Thank you! 

I think the PR https://github.com/apache/spark/pull/15896 is fixing a related issue in `override def uncacheTable(tableName: String)`. The fixed behavior need to be consistent with each other. 

BTW, `TruncateTableCommand` also has the same issue.

dongjoon-hyun:
I see. I'll include that command here, too.

gatorsmile:
Please follw the ongoing design decision changes in Spark-18465

dongjoon-hyun:
Sure!

dongjoon-hyun:
Hi, @gatorsmile .
Is there any decision for the following?
> The fixed behavior needs to be consistent with each other.

dongjoon-hyun:
If the issue (and bugs) about *uncacheQuery* and *uncacheTable* is postpone, I'd like to focus on the warning messages in this PR. Is it okay? We can revisit that cache bugs for Apache Spark 2.2.

gatorsmile:
We should not suppress the warnings. Instead, we should improve the error message to show the unache attempt failed, right?

hvanhovell:
@dongjoon-hyun @gatorsmile I totally missed this one. Shall we try to get  this in 2.1?

gatorsmile:
The warning message is valid. We might need to issue a better exception here. 

To resolve the root cause, we might need adding a new API for uncaching the view in which a relevant table has been dropped? I am not confident how to do it in a perfect way. Let me open a JIRA now and cc both of you in the discussion.

dongjoon-hyun:
Thank you for coming back, @gatorsmile and @hvanhovell !

gatorsmile:
A JIRA is created: https://issues.apache.org/jira/browse/SPARK-18549

dongjoon-hyun:
Ah, In fact, that issue includes this. May I close this PR and the issue SPARK-18169 ?

gatorsmile:
Sure, maybe close it now. If we are unable to contain it in Spark 2.1, we can continue this work to issue a better error message.

dongjoon-hyun:
I guess SPARK-18169 will cover the better error message too. That issue will need to revisit all occurrence of `uncache` and related exception handling. Thank you, @gatorsmile and @hvanhovell . I close this in favor of SPARK-18549 .

PR #15683, opened 2016-10-30
[SPARK-18166][MLlib] Fix Poisson GLM bug due to wrong requirement of response values

actuaryzhang:

The current implementation of Poisson GLM seems to allow only positive values. This is incorrect since the support of Poisson includes the origin. The bug is easily fixed by changing the test of the Poisson variable from  'require(y **>** 0.0' to  'require(y **>=** 0.0'. 

@mengxr  @srowen

### DIFF
--- a/mllib/src/test/scala/org/apache/spark/ml/regression/GeneralizedLinearRegressionSuite.scala
+++ b/mllib/src/test/scala/org/apache/spark/ml/regression/GeneralizedLinearRegressionSuite.scala
@@ -477,7 +477,6 @@ class GeneralizedLinearRegressionSuite
        }
        [1]  0.4272661 -0.1565423
        [1] -3.6911354  0.6214301  0.1295814
-
      */
     val expected = Seq(
       Vectors.dense(0.0, 0.4272661, -0.1565423),
@@ -495,21 +494,6 @@ class GeneralizedLinearRegressionSuite
       val actual = Vectors.dense(model.intercept, model.coefficients(0), model.coefficients(1))
       assert(actual ~= expected(idx) absTol 1e-4, "Model mismatch: GLM with poisson family, " +
         s"$link link and fitIntercept = $fitIntercept (with zero values).")
-
-      val familyLink = new FamilyAndLink(Poisson, Link.fromName(link))
-      model.transform(dataset).select("features", "prediction", "linkPrediction").collect()
-        .foreach {
-          case Row(features: DenseVector, prediction1: Double, linkPrediction1: Double) =>
-            val eta = BLAS.dot(features, model.coefficients) + model.intercept
-            val prediction2 = familyLink.fitted(eta)
-            val linkPrediction2 = eta
-            assert(prediction1 ~= prediction2 relTol 1E-5, "Prediction mismatch: GLM with " +
-              s"poisson family, $link link and fitIntercept = $fitIntercept. (with zero values)")
-            assert(linkPrediction1 ~= linkPrediction2 relTol 1E-5, "Link Prediction mismatch: " +
-              s"GLM with poisson family, $link link and fitIntercept = $fitIntercept " +
-              "(with zero values).")
-        }
-
       idx += 1
     }
   }
--- a/mllib/src/test/scala/org/apache/spark/ml/regression/GeneralizedLinearRegressionSuite.scala
+++ b/mllib/src/test/scala/org/apache/spark/ml/regression/GeneralizedLinearRegressionSuite.scala
@@ -44,6 +44,7 @@ class GeneralizedLinearRegressionSuite
   @transient var datasetGaussianInverse: DataFrame = _
   @transient var datasetBinomial: DataFrame = _
   @transient var datasetPoissonLog: DataFrame = _
+  @transient var datasetPoissonLogWithZero: DataFrame = _
   @transient var datasetPoissonIdentity: DataFrame = _
   @transient var datasetPoissonSqrt: DataFrame = _
   @transient var datasetGammaInverse: DataFrame = _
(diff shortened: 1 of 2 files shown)


srowen:
Jenkins test this please

srowen (approved):
This seems OK to me. I was wondering why the existing tests didn't exercise this case. I think it's because it generates data for Poisson regression where the target variable is real-valued, not an integer. So it occasionally generates pretty small values but never 0. In a lot of of real-world contexts I'd of course expect to regress integer counts where 0 is quite possible.

If you're up for it, feel free to write some targeted simple unit tests of the initialize() method for the various families to make sure their support is correct, but in any event I am OK merging this.

sethah:
+1 for unit tests.

srowen:
@actuaryzhang could you add some simple tests for the initialize() methods of the various families?

actuaryzhang:
@srowen Will add the unit test over the weekend.

srowen:
@actuaryzhang would love to get this in as a fix for 2.1. Are you able to? or should I finish it up (and give you credit on JIRA)?

thunterdb:
+1 for trying to get it into 2.1 (modulo tests)

actuaryzhang:
@srowen @thunterdb 

I just updated the unit test for poisson GLM (only for the log link). The simulated data are now forced to take values of zero. Existing data generation is not really simulating from the Poisson distribution. The addition of `mu` and `noiseLevel` produces values on the real line. 
`val label = mu + noiseLevel * (generator.nextValue() - mean)`

All tests passed on my end. Please review and let me know if there is any question. Thanks.

mllib/src/test/scala/org/apache/spark/ml/regression/GeneralizedLinearRegressionSuite.scala
```diff
@@ -83,10 +83,11 @@ class GeneralizedLinearRegressionSuite
       testData.toDF()
     }
 
+    // force some labels to be exactly zero
     datasetPoissonLog = generateGeneralizedLinearRegressionInput(
-      intercept = 0.25, coefficients = Array(0.22, 0.06), xMean = Array(2.9, 10.5),
+      intercept = -1.5, coefficients = Array(0.22, 0.06), xMean = Array(2.9, 10.5),
       xVariance = Array(0.7, 1.2), nPoints = 10000, seed, noiseLevel = 0.01,
-      family = "poisson", link = "log").toDF()
+      family = "poisson", link = "log").map{x => LabeledPoint(if (x.label < 0.7) 0.0 else x.label, x.features)}.toDF()
```
sethah: I'd suggest creating a test case explicitly to check that Poisson accepts 0 values. Then we can just use some very small dataset (5 points or similar) which contains a zero label. 

The way it is now, if someone were to change the intercept to some other large number in the future, then we wouldn't have any zero labels in the data and we wouldn't exercise that test case. 

Also, you can run the style checker via `./dev/lint-scala` to check that the patch meets style guidelines. This line is too long, currently.

mllib/src/test/scala/org/apache/spark/ml/regression/GeneralizedLinearRegressionSuite.scala
```diff
@@ -453,6 +454,8 @@ class GeneralizedLinearRegressionSuite
     }
   }
 
+
```
sethah: we should remove these two lines.

actuaryzhang:
@sethah  Thanks for the review and comments. I now created a separate unit test. It also passed the style test. 

I accidentally merged master into a branch... and don't know how to revert it back. Would someone please fix that? Thanks.

srowen:
Looks good; I was actually just thinking of testing the family itself rather than a whole new test case for GLM, but this gets the job done.

mllib/src/test/scala/org/apache/spark/ml/regression/GeneralizedLinearRegressionSuite.scala
```diff
@@ -453,6 +464,56 @@ class GeneralizedLinearRegressionSuite
...
+    for (fitIntercept <- Seq(false, true)) {
+      val trainer = new GeneralizedLinearRegression().setFamily("poisson").setLink(link)
+        .setFitIntercept(fitIntercept).setLinkPredictionCol("linkPrediction")
+      val model = trainer.fit(dataset)
+      val actual = Vectors.dense(model.intercept, model.coefficients(0), model.coefficients(1))
+      assert(actual ~= expected(idx) absTol 1e-4, "Model mismatch: GLM with poisson family, " +
+        s"$link link and fitIntercept = $fitIntercept (with zero values).")
+
+      val familyLink = new FamilyAndLink(Poisson, Link.fromName(link))
+      model.transform(dataset).select("features", "prediction", "linkPrediction").collect()
```
sethah: I don't see any reason to test the predictions here. Checking the coefficients should be enough, since we test for correctness in other places.

sethah:
@actuaryzhang Thanks a lot for correcting this! I just had a small comment to make the additional test shorter.

actuaryzhang:
@sethah Thanks for your review and suggestion. I have made a new commit reflecting your comments. 

@srowen Thanks for all the suggestions. When do you think this change could be merged back?

srowen (approved):

mllib/src/test/scala/org/apache/spark/ml/regression/GeneralizedLinearRegressionSuite.scala
```diff
@@ -88,6 +89,12 @@ class GeneralizedLinearRegressionSuite
       xVariance = Array(0.7, 1.2), nPoints = 10000, seed, noiseLevel = 0.01,
       family = "poisson", link = "log").toDF()
 
+    datasetPoissonLogWithZero = generateGeneralizedLinearRegressionInput(
+      intercept = -1.5, coefficients = Array(0.22, 0.06), xMean = Array(2.9, 10.5),
+      xVariance = Array(0.7, 1.2), nPoints = 100, seed, noiseLevel = 0.01,
+      family = "poisson", link = "log")
+      .map{x => LabeledPoint(if (x.label < 0.7) 0.0 else x.label, x.features)}.toDF()
```
sethah: Well, I still prefer the explicit data but that seems ok. I'm still curious why the test fails with the above dataset, but I don't have the bandwidth to investigate right now.

srowen:
Merged to master/2.1. @actuaryzhang it doesn't pick cleanly into 2.0 but if you want to open a new PR for 2.0 I'll merge that too.

PR #15688, opened 2016-10-30
[SPARK-18173][SQL] data source tables should support truncating partition

cloud-fan:

Previously `TRUNCATE TABLE ... PARTITION` will always truncate the whole table for data source tables, this PR fixes it and improve `InMemoryCatalog` to make this command work with it.
## How was this patch tested?

existing tests

### DIFF
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/InMemoryCatalog.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/InMemoryCatalog.scala
@@ -493,11 +493,26 @@ class InMemoryCatalog(
       table: String,
       partialSpec: Option[TablePartitionSpec] = None): Seq[CatalogTablePartition] = synchronized {
     requireTableExists(db, table)
-    if (partialSpec.nonEmpty) {
-      throw new UnsupportedOperationException(
-        "listPartition with partial partition spec is not implemented")
+
+    partialSpec match {
+      case None => catalog(db).tables(table).partitions.values.toSeq
+      case Some(partial) =>
+        catalog(db).tables(table).partitions.toSeq.collect {
+          case (spec, partition) if isPartialPartitionSpec(partial, spec) => partition
+        }
     }
-    catalog(db).tables(table).partitions.values.toSeq
+  }
+
+  /**
+   * Returns true if `spec1` is a partial partition spec w.r.t. `spec2`, e.g. PARTITION (a=1) is a
+   * partial partition spec w.r.t. PARTITION (a=1,b=2).
+   */
+  private def isPartialPartitionSpec(
+      spec1: TablePartitionSpec,
+      spec2: TablePartitionSpec): Boolean = {
+    spec1.forall {
+      case (partitionColumn, value) => spec2(partitionColumn) == value
+    }
   }
 
   override def listPartitionsByFilter(
--- a/sql/core/src/test/scala/org/apache/spark/sql/execution/command/DDLSuite.scala
+++ b/sql/core/src/test/scala/org/apache/spark/sql/execution/command/DDLSuite.scala
@@ -1678,6 +1678,12 @@ class DDLSuite extends QueryTest with SharedSQLContext with BeforeAndAfterEach {
       // Hive's behaviour or stick with our existing behaviour later.
       sql("TRUNCATE TABLE partTable PARTITION (width=100, length=100)")
       assert(spark.table("partTable").count() == data.count())
+
+      // throw exception if the column in partition spec is not a partition column.
+      val e = intercept[AnalysisException] {
+        sql("TRUNCATE TABLE partTable PARTITION (unknown=1)")
+      }
+      assert(e.message.contains("unknown is not a valid partition column"))
     }
   }
 
--- a/sql/core/src/test/scala/org/apache/spark/sql/execution/command/DDLSuite.scala
+++ b/sql/core/src/test/scala/org/apache/spark/sql/execution/command/DDLSuite.scala
@@ -1628,29 +1628,56 @@ class DDLSuite extends QueryTest with SharedSQLContext with BeforeAndAfterEach {
 
   test("truncate table - datasource table") {
     import testImplicits._
-    val data = (1 to 10).map { i => (i, i) }.toDF("width", "length")
 
+    val data = (1 to 10).map { i => (i, i) }.toDF("width", "length")
     // Test both a Hive compatible and incompatible code path.
     Seq("json", "parquet").foreach { format =>
       withTable("rectangles") {
         data.write.format(format).saveAsTable("rectangles")
         assume(spark.table("rectangles").collect().nonEmpty,
           "bad test; table was empty to begin with")
+
         sql("TRUNCATE TABLE rectangles")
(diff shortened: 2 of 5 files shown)


cloud-fan:
cc @ericl @yhuai @gatorsmile

sql/core/src/test/scala/org/apache/spark/sql/execution/command/DDLSuite.scala
```diff
@@ -1636,15 +1635,23 @@ class DDLSuite extends QueryTest with SharedSQLContext with BeforeAndAfterEach {
     }
 
     withTable("rectangles", "rectangles2") {
```
jaceklaskowski: Mind changing `rectangles2` to `rectPartitioned` or similar? Think the test would get simpler to read (later).

ericl:
This lgtm

gatorsmile:
If we try to truncate a partition that does not exist, we just do nothing?

cloud-fan:
retest this please

cloud-fan:
@gatorsmile that's a good point. I checked with hive, the behaviour is:

1. if the given partition spec specifies all partition columns, throw exception if the partition not exist
2. if the given partition spec is partial, do nothing if the partition not exist

I have updated this PR according to this

cloud-fan:
retest this please

cloud-fan:
retest this please

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/InMemoryCatalog.scala
```diff
@@ -471,11 +471,30 @@ class InMemoryCatalog(
...
+          })
+        } else {
+          catalog(db).tables(table).partitions.toSeq.collect {
+            case (spec, partition) if isPartialPartitionSpec(partial, spec) => partition
+          }
+        }
+    }
+  }
+
+  private def isPartialPartitionSpec(
```
rxin: what does partial mean here?

rxin: we should document it.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/InMemoryCatalog.scala
```diff
@@ -471,11 +471,30 @@ class InMemoryCatalog(
...
+      partialSpec: TablePartitionSpec,
+      completedSpec: TablePartitionSpec): Boolean = {
+    partialSpec.forall {
+      case (partitionColumn, value) => completedSpec(partitionColumn) == value
```
gatorsmile: If the specified column names are not partitioning columns, we will get a strange error:
```
key not found: height
java.util.NoSuchElementException: key not found: height
```

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/InMemoryCatalog.scala
```diff
@@ -471,11 +471,30 @@ class InMemoryCatalog(
...
       partialSpec: Option[TablePartitionSpec] = None): Seq[CatalogTablePartition] = synchronized {
     requireTableExists(db, table)
-    if (partialSpec.nonEmpty) {
-      throw new UnsupportedOperationException(
-        "listPartition with partial partition spec is not implemented")
+
+    partialSpec match {
+      case None => catalog(db).tables(table).partitions.values.toSeq
+      case Some(partial) =>
+        if (partial.size == getTable(db, table).partitionColumnNames.length) {
```
gatorsmile: Comparing the size is not enough. We also need to check whether all the columns are partitioning columns.

gatorsmile:
Could we add all the positive and negative cases in `ExternalCatalogSuite`, like the other APIs. Then, we can easily know whether the behaviors of `InMemoryCatalog` and `HiveExternalCatalog` are consistent.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/InMemoryCatalog.scala
cloud-fan: I don't think we will/should hit this in the `ExternalCatalog` level, as most of partition related methods assumes the column names in partition spec are partition columns, e.g. `loadPartition`, `loadDynamicPartitions`, `dropPartitions`, etc.

If we do hit this, we should fix it at the caller side.

gatorsmile: The above message was got from the statement
```
sql("TRUNCATE TABLE partTable PARTITION (height =1)")
```

cloud-fan:
retest this please

sql/core/src/test/scala/org/apache/spark/sql/execution/command/DDLSuite.scala
```diff
@@ -1628,29 +1628,56 @@ class DDLSuite extends QueryTest with SharedSQLContext with BeforeAndAfterEach {
...
+
+    withTable("partTable") {
+      data.write.partitionBy("width", "length").saveAsTable("partTable")
+      // do nothing if no partition is matched for the given partial partition spec
+      sql("TRUNCATE TABLE partTable PARTITION (width=100)")
+      assert(spark.table("partTable").count() == data.count())
+
+      // do nothing if no partition is matched for the given non-partial partition spec
+      // TODO: This behaviour is different from Hive, we should decide whether we need to follow
+      // Hive's behaviour or stick with our existing behaviour later.
```
rxin: what's our and hive's behavior?

rxin:
cc @gatorsmile

sql/core/src/test/scala/org/apache/spark/sql/execution/command/DDLSuite.scala
cloud-fan: our behaviour: do nothing if no partition is matched, no matter whether the given partition spec is partial or not.

Hive's behaviour: when no partition is matched, do nothing if the given partition spec is partial, throw exception if the given partition spec is non-partial.

gatorsmile: Could we document the behavior difference in the code comments, if we decide to stick with our existing behavior?

gatorsmile:
Regarding the current code changes, LGTM

Like what @cloud-fan said in the code comment, we need to decide whether we should follow Hive.

sql/core/src/test/scala/org/apache/spark/sql/execution/command/DDLSuite.scala
rxin: I actually think Hive's behavior makes sense here. If I'm giving you an exact match, you should warn me if there is an issue.

rxin: The alternative, of course, is to provide "truncate if exists", which doesn't throw exceptions if that's desired.

rxin:

rxin:
@cloud-fan can you create a follow-up pr to switch over to Hive's behavior?

PR #7202, opened 2015-07-02
[SPARK-8801][SQL] Support TypeCollection in ExpectsInputTypes

rxin:
This patch adds a new TypeCollection AbstractDataType that can be used by expressions to specify more than one expected input types.

PR #15690, opened 2016-10-30
[SPARK-18174] [SQL] [DO-NOT-MERGE] Avoid Implicit Type Cast in Arguments of Expressions Extending String2StringExpression

gatorsmile:

For the expressions that extend `String2StringExpression` (including, `lower`, `upper`, `ltrim`, `rtrim`, `trim` and `reverse`), `Analyzer` should not implicitly cast the arguments to string. Instead, if users input the some data types instead of string, we should issue an exception for this misuse.

Added a test case in the `StringFunctionsSuite`

### DIFF
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/stringExpressions.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/stringExpressions.scala
@@ -230,7 +230,7 @@ case class Elt(children: Seq[Expression])
 }
 
 
-trait String2StringExpression extends ImplicitCastInputTypes {
+trait String2StringExpression extends ExpectsInputTypes {
   self: UnaryExpression =>
 
   def convert(v: UTF8String): UTF8String
--- a/sql/core/src/test/scala/org/apache/spark/sql/StringFunctionsSuite.scala
+++ b/sql/core/src/test/scala/org/apache/spark/sql/StringFunctionsSuite.scala
@@ -449,6 +449,24 @@ class StringFunctionsSuite extends QueryTest with SharedSQLContext {
       df2.selectExpr("str_to_map(a)"),
       Seq(Row(Map("a" -> "1", "b" -> "2", "c" -> "3")))
     )
+  }
 
+  test("String2StringExpression - no implicit type casting") {
+    def checkTypeErrorMsg(funcName: String): Unit = {
+      val df = Seq(345).toDF("a")
+      val expectedMsg = "argument 1 requires string type, however, '`a`' is of int type"
+      // non-string type of arguments is unacceptable
+      val e = intercept[AnalysisException] {
+        df.selectExpr(s"$funcName(a)")
+      }.getMessage
+      assert(e.contains(expectedMsg))
+    }
+
+    checkTypeErrorMsg("reverse")
+    checkTypeErrorMsg("lower")
+    checkTypeErrorMsg("upper")
+    checkTypeErrorMsg("ltrim")
+    checkTypeErrorMsg("rtrim")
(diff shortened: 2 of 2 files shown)


gatorsmile:
@rxin @cloud-fan Could you check whether this PR makes sense? Without this PR, these string functions can consume all the data types. Actually, almost all the string functions are facing the same issue.

If they really want to use these string functions, should they do explicit type casting instead? This is a design decision we need to make. Either looks OK to me. Thanks!

tejasapatil:
If we want to go with the model of being super strict about users doing proper cast, shouldn't all UDFs / functions avoid doing implicit cast ? Else it will be confusing for users as sometimes CAST is not needed but in other places it is.

Downside being that this model (and also this PR) will affect existing jobs. If we leave things as they are now, would it lead to wrong results ? I am trying to get sense for impact of living without this change.

HyukjinKwon:
This is a rapid change that breaks the compatibility in existing codes. I guess we should check other databases as well in this case.

HyukjinKwon:
cc @hvanhovell too. I guess we talked about the similar issue in another PR.

gatorsmile:
@tejasapatil I see your point. Implicit casting is powerful. Actually, different languages have different preferences. I checked the related change history of Spark SQL. I think we prefer to implicit casting in the Spark SQL design. Especially after we introduce `TypeCollection` in the https://github.com/apache/spark/pull/7202, the order of types defined in `TypeCollection` also matters. 

Question to @tejasapatil : have you hit the behavior differences between Hive and Spark SQL caused by the implicit type casting? That means, although the function signatures between Hive and Spark SQL are the same, the results might be different due to the casted types are different.

gatorsmile:
@HyukjinKwon This PR is to collect the feedbacks from the others about implicit casting. Now, it seems like we should follow the general design principle. If we just put the target type as the function argument types, users easily hit the inconsistency between documents and source codes. 

Different databases have different behaviors. Actually, if you check the programing languages, the behaviors are also different.

gatorsmile:
Found another side effect caused by implicit casting. When we having `CAST` in the filter conditions, we are unable to push down these filters into the data sources. Performance could be very bad if users do not properly specify the data types of these literals.

For example, when the data type of `a` is integer, the following filter can be pushed down to the parquet:
```
spark.read.parquet(path).filter("a < 100")
```
However, users will not see the filter push down if the literal is not an integer.
```
spark.read.parquet(path).filter("a < 100.0")
```

tejasapatil:
>> have you hit the behavior differences between Hive and Spark SQL caused by the implicit type casting?

So far I have not seen correctness issue due to casting. I can imagine casting leading to loss of precision but I would say that its sloppy work from user side to let engine choose the precision while they really wanted accuracy. Query failure (during analysis phase) can happen if someone tries to run an existing Spark SQL query (or run a hive query) with this change and did not have strictly cast.

For a new pipeline being migrated from Hive => Spark, the user cost is to analyse the query and make necessary change. Having query backward compatible with Hive is _preferable_ because one can switch across engines based on choice without changing the query.

For existing pipelines, there is a greater cost as in someone will have to look at all existing jobs and see that they won't fail. If they do, make the change and ensure that the results are same. This part doesn't scale well.

I am quite neutral to have this change. Its basically a choice. My only opinion is towards having consistency wrt usage of implicit casting : do it everywhere possible OR don't do it at all.

gatorsmile:
Thank you for your inputs! @tejasapatil 

We should not break the existing behaviors. Thus, I close this PR. Thanks!

rxin:
Why not have a strict mode flag?

gatorsmile:
@rxin A good idea! Then, users can turn it off! Will submit a separate PR for it. Thanks!

PR #15694, opened 2016-10-31
[SPARK-18179][SQL] Throws analysis exception with a proper message for unsupported argument types in reflect/java_method function

HyukjinKwon:

This PR proposes throwing an `AnalysisException` with a proper message rather than `NoSuchElementException` with the message ` key not found: TimestampType` when unsupported types are given to `reflect` and `java_method` functions.

```scala
spark.range(1).selectExpr("reflect('java.lang.String', 'valueOf', cast('1990-01-01' as timestamp))")
```

produces

**Before**

```
java.util.NoSuchElementException: key not found: TimestampType
  at scala.collection.MapLike$class.default(MapLike.scala:228)
  at scala.collection.AbstractMap.default(Map.scala:59)
  at scala.collection.MapLike$class.apply(MapLike.scala:141)
  at scala.collection.AbstractMap.apply(Map.scala:59)
  at org.apache.spark.sql.catalyst.expressions.CallMethodViaReflection$$anonfun$findMethod$1$$anonfun$apply$1.apply(CallMethodViaReflection.scala:159)
...
```

**After**

```
cannot resolve 'reflect('java.lang.String', 'valueOf', CAST('1990-01-01' AS TIMESTAMP))' due to data type mismatch: arguments from the third require boolean, byte, short, integer, long, float, double or string expressions; line 1 pos 0;
'Project [unresolvedalias(reflect(java.lang.String, valueOf, cast(1990-01-01 as timestamp)), Some(<function1>))]
+- Range (0, 1, step=1, splits=Some(2))
...
```

Added message is,

```
arguments from the third require boolean, byte, short, integer, long, float, double or string expressions
```

## How was this patch tested?

Tests added in `CallMethodViaReflection`.

### DIFF
--- a/sql/core/src/test/scala/org/apache/spark/sql/MiscFunctionsSuite.scala
+++ b/sql/core/src/test/scala/org/apache/spark/sql/MiscFunctionsSuite.scala
@@ -35,7 +35,7 @@ class MiscFunctionsSuite extends QueryTest with SharedSQLContext {
       Row("m1one", "m1one"))
   }
 
-  test("reflect and java_method throws an analysis exception for unsupported types") {
+  test("reflect and java_method throw an analysis exception for unsupported types") {
     val df = Seq((new Timestamp(1), Decimal(10))).toDF("a", "b")
     val className = ReflectClass.getClass.getName.stripSuffix("$")
     val messageOne = intercept[AnalysisException] {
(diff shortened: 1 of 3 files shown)


sql/core/src/test/scala/org/apache/spark/sql/MiscFunctionsSuite.scala
```diff
@@ -31,6 +34,47 @@ class MiscFunctionsSuite extends QueryTest with SharedSQLContext {
...
+      df.selectExpr(
+        s"java_method('$className', 'method1', a, b)").collect()
+    }.getMessage
+
+    assert(messageTwo.contains(
+      "arguments from the third require boolean, byte, short, " +
+        "integer, long, float, double or string expressions"))
+  }
+
+  test("reflect and java_method throw an analysis exception for non-existing method/class") {
```
HyukjinKwon: Oh, yes, it is. Will clean up including the comment below.

HyukjinKwon:
Would there be other things I should take care of maybe?

rxin:
Merging in master/branch-2.1.

PR #15698, opened 2016-10-31
[SPARK-18182] Expose ReplayListenerBus.read() overload which takes string iterator

JoshRosen:
The `ReplayListenerBus.read()` method is used when implementing a custom `ApplicationHistoryProvider`. The current interface only exposes a `read()` method which takes an `InputStream` and performs stream-to-lines conversion itself, but it would also be useful to expose an overloaded method which accepts an iterator of strings, thereby enabling events to be provided from non-`InputStream` sources.


andrewor14:
LGTM retest this please

rxin:
Merging in master.

PR #15699, opened 2016-10-31
[SPARK-18030][Tests]Fix flaky FileStreamSourceSuite by not deleting the files

zsxwing:

The test `when schema inference is turned on, should read partition data` should not delete files because the source maybe is listing files. This PR just removes the delete actions since they are not necessary.

## How was this patch tested?

Jenkins


zsxwing:
cc @marmbrus

marmbrus:
LGTM

zsxwing:

PR #15701, opened 2016-10-31
[SPARK-18167] [SQL] Also log all partitions when the SQLQuerySuite test flakes

ericl:

One possibility for this test flaking is that we have corrupted the partition schema somehow in the tests, which causes the cast to decimal to fail in the call. This should at least show us the actual partition values.

## How was this patch tested?

Run it locally, it prints out something like `ArrayBuffer(test(partcol=0), test(partcol=1), test(partcol=2), test(partcol=3), test(partcol=4))`.


ericl:
@yhuai

yhuai:
lgtm

yhuai:
merging to master. Thanks

PR #15702, opened 2016-10-31
[SPARK-18124] Observed delay based Event Time Watermarks

marmbrus:
This PR adds a new method `withWatermark` to the `Dataset` API, which can be used specify an _event time watermark_.  An event time watermark allows the streaming engine to reason about the point in time after which we no longer expect to see late data.  This PR also has augmented `StreamExecution` to use this watermark for several purposes:
  - To know when a given time window aggregation is finalized and thus results can be emitted when using output modes that do not allow updates (e.g. `Append` mode).
  - To minimize the amount of state that we need to keep for on-going aggregations, by evicting state for groups that are no longer expected to change.  Although, we do still maintain all state if the query requires (i.e. if the event time is not present in the `groupBy` or when running in `Complete` mode).

An example that emits windowed counts of records, waiting up to 5 minutes for late data to arrive.
```scala
df.withWatermark("eventTime", "5 minutes")
  .groupBy(window($"eventTime", "1 minute") as 'window)
  .count()
  .writeStream
  .format("console")
  .mode("append") // In append mode, we only output finalized aggregations.
  .start()
```

### Calculating the watermark.
The current event time is computed by looking at the `MAX(eventTime)` seen this epoch across all of the partitions in the query minus some user defined _delayThreshold_.  An additional constraint is that the watermark must increase monotonically.

Note that since we must coordinate this value across partitions occasionally, the actual watermark used is only guaranteed to be at least `delay` behind the actual event time.  In some cases we may still process records that arrive more than delay late.

This mechanism was chosen for the initial implementation over processing time for two reasons:   
  - it is robust to downtime that could affect processing delay
  - it does not require syncing of time or timezones between the producer and the processing engine.

### Other notable implementation details
 - A new trigger metric `eventTimeWatermark` outputs the current value of the watermark.
 - We mark the event time column in the `Attribute` metadata using the key `spark.watermarkDelay`.  This allows downstream operations to know which column holds the event time.  Operations like `window` propagate this metadata.
 - `explain()` marks the watermark with a suffix of `-T${delayMs}` to ease debugging of how this information is propagated.
 - Currently, we don't filter out late records, but instead rely on the state store to avoid emitting records that are both added and filtered in the same epoch.

### Remaining in this PR
 - [ ] The test for recovery is currently failing as we don't record the watermark used in the offset log.  We will need to do so to ensure determinism, but this is deferred until #15626 is merged.

### Other follow-ups
There are some natural additional features that we should consider for future work:
 - Ability to write records that arrive too late to some external store in case any out-of-band remediation is required.
 - `Update` mode so you can get partial results before a group is evicted.
 - Other mechanisms for calculating the watermark.  In particular a watermark based on quantiles would be more robust to outliers.


sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/EventTimeWatermarkExec.scala
```diff
@@ -0,0 +1,93 @@
...
+ * `watermark` after which we assume we will no longer see late records for a particular time
+ * period.
+ */
+case class EventTimeWatermarkExec(
+    eventTime: Attribute,
+    delay: CalendarInterval,
+    child: SparkPlan) extends SparkPlan {
+
+  // TODO: Use Spark SQL Metrics?
+  val maxEventTime = new MaxLong
```
marmbrus: @zsxwing am I doing this right?

sql/core/src/main/scala/org/apache/spark/sql/Dataset.scala
```diff
@@ -536,6 +535,37 @@ class Dataset[T] private[sql](
...
+   * @param eventTime the name of the column that contains the event time of the row.
+   * @param delay the minimum delay to wait to data to arrive late, relative to the latest record
+   *              that has been processed.
+   *
+   * @group streaming
+   * @since 2.1.0
+   */
+  @Experimental
+  @InterfaceStability.Evolving
+  def withWatermark(eventTime: String, delay: String): Dataset[T] = withTypedPlan {
```
rxin: you'd need one that takes in a column wouldn't you?

marmbrus:
@ekl - flaky test...  Should we turn it off for now?

retest this please

ericl:
I'm still trying to find a failure that includes https://github.com/apache/spark/pull/15701/files. Until then it's hard to debug.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/namedExpressions.scala
```diff
@@ -292,11 +294,22 @@ case class AttributeReference(
...
+  }
+
   override protected final def otherCopyArgs: Seq[AnyRef] = {
     exprId :: qualifier :: isGenerated :: Nil
   }
 
-  override def toString: String = s"$name#${exprId.id}$typeSuffix"
+  /** Used to signal the column used to calculate an eventTime watermark (e.g. a#1-T{delayMs}) */
+  private def delaySuffix = if (metadata.contains(EventTimeWatermark.delayKey)) {
+    s"-T${metadata.getLong(EventTimeWatermark.delayKey)}"
```
brkyvz: is this in milliseconds or microseconds like `timestamp` type?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/EventTimeWatermark.scala
```diff
@@ -0,0 +1,51 @@
...
+    eventTime: Attribute,
+    delay: CalendarInterval,
+    child: LogicalPlan) extends LogicalPlan {
+
+  // Update the metadata on the eventTime column to include the desired delay.
+  override val output: Seq[Attribute] = child.output.map { a =>
+    if (a semanticEquals eventTime) {
+      val updatedMetadata = new MetadataBuilder()
+        .withMetadata(a.metadata)
+        .putLong(EventTimeWatermark.delayKey, delay.milliseconds)
```
brkyvz: I'm a bit confused. Normally Spark SQL uses microsecond precision for TimestampType. When it converts it to LongType, it uses second precision. Here we're using milliseconds. Wouldn't that be super confusing to reason about?

brkyvz:
A very dumb question (I apologize), there is nothing stopping a user to actually use processing time as watermarks with this API either. One can easily do `df.withColumn("timestamp", current_timestamp()).withWatermark("timestamp")`. I like that we're suggesting users to use `eventTime` for stability, but we're not actually constraining them, right?

My biggest confusion here, that I couldn't find documented was the Type of the watermark column. Does it need to be timestamp type or can it be LongType?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/EventTimeWatermark.scala
marmbrus: I switched it to using `CalendarInterval` to make it clearer what units were being used where.  I chose milliseconds because it seemed like the right granularity.  Microseconds are too short for the global coordination required and seconds lack granularity.  It should be easy to change, and I'm open to that if there's consensus this is too confusing though.

marmbrus:
Not a dumb question!  You can certainly use processing time if those are the semantics you require.  I do think there is a little bit of work we need to do to ensure determinism for these functions.  Specifically, `now()` in SQL is supposed to return the moment query execution began.  For streaming we'll need to record this and make sure that we give the same time even in the case of failures (based on the batch id with the current execution model).

Good point on the documentation.  The thing you are missing is that it must be used in a window function, which does require `TimestampType`.  I can see how to make this more clear.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/EventTimeWatermark.scala
marmbrus: Updating the key to include `Ms`

sql/core/src/main/scala/org/apache/spark/sql/Dataset.scala
```diff
@@ -536,6 +535,41 @@ class Dataset[T] private[sql](
...
 
   /**
+   * :: Experimental ::
+   * Defines an event time watermark for this [[Dataset]]. A watermark tracks a point in time
+   * before which we assume no more late data is going to arrive.
+   *
+   * Spark will use this watermark for several purposes:
+   *  - To know when a given time window aggregation can be finalized and thus can be emitted when
+   *    using output modes that do not allow updates.
+   *  - To minimize the amount of state that we need to keep for on-going aggregations.
```
koeninger: For append, this sounds like the intention is emit only once watermark has passed, and drop state.
But for other output modes, it's not clear from reading this what the effect of the watermark on emission and dropping state is.

sql/core/src/main/scala/org/apache/spark/sql/Dataset.scala
```diff
@@ -536,6 +535,41 @@ class Dataset[T] private[sql](
...
+   *  The current event time is computed by looking at the `MAX(eventTime)` seen in an epoch across
+   *  all of the partitions in the query minus a user specified `delayThreshold`.  Due to the cost
+   *  of coordinating this value across partitions, the actual watermark used is only guaranteed
+   *  to be at least `delayThreshold` behind the actual event time.  In some cases we may still
+   *  process records that arrive more than `delayThreshold` late.
+   *
+   * @param eventTime the name of the column that contains the event time of the row.
+   * @param delayThreshold the minimum delay to wait to data to arrive late, relative to the latest
+   *                       record that has been processed in the form of an interval
+   *                       (e.g. "1 minute" or "5 hours").
```
koeninger: Should this make it clear what the minimum useful granularity is (ms)?

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StatefulAggregate.scala
```diff
@@ -104,85 +110,105 @@ case class StateStoreSaveExec(
...
+            store.commit()
+
+            numTotalStateRows += store.numKeys()
+            store.updates().filter(_.isInstanceOf[ValueRemoved]).map { removed =>
+              numOutputRows += 1
+              removed.value.asInstanceOf[InternalRow]
+            }
+
+          // Update and output modified rows from the StateStore.
+          case Some(Update) =>
```
koeninger: I'm not clear on why the semantics of Update mean that watermarks shouldn't be used to remove state.

koeninger:
Given the concerns Ofir raised about a single far future event screwing up monotonic event time, do you want to document that problem even if there isn't an enforced filter for it?

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StatefulAggregate.scala
koeninger: To put it the other way, do the docs in this PR tell you as a user that for any output method other than Append, you are potentially keeping unlimited aggregate state in memory, regardless of whether you set a watermark?

koeninger: Yes, I think it's a good idea to explicitly say for each output mode whether watermarks affect emit and evict.  Just so I'm clear, the intention is

Append: affects emit, affects evict
Update: doesn't affect emit, affects evict
Complete: doesn't affect emit, no eviction

Is that right?

sql/core/src/main/scala/org/apache/spark/sql/Dataset.scala
```diff
@@ -536,6 +535,41 @@ class Dataset[T] private[sql](
...
+   * :: Experimental ::
+   * Defines an event time watermark for this [[Dataset]]. A watermark tracks a point in time
+   * before which we assume no more late data is going to arrive.
+   *
+   * Spark will use this watermark for several purposes:
+   *  - To know when a given time window aggregation can be finalized and thus can be emitted when
+   *    using output modes that do not allow updates.
+   *  - To minimize the amount of state that we need to keep for on-going aggregations.
+   *
+   *  The current event time is computed by looking at the `MAX(eventTime)` seen in an epoch across
```
marmbrus: Changed to watermark.  For epoch, I really just mean "during some period of time where we decide too coordinate across the partitions".  This happens at batch boundaries now, but that is not part of the contract we are promising.  I just removed that word to avoid confusion.

sql/core/src/main/scala/org/apache/spark/sql/Dataset.scala
marmbrus: That seems like more of an implementation detail, rather than a contract of the API.  The real contract is stated above as `the actual watermark used is only guaranteed to be at least 'delayThreshold' behind the actual event time`.  There aren't really any bounds we can promise without knowing more about the query (even ms).

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StatefulAggregate.scala
marmbrus: That is correct.

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/EventTimeWatermarkExec.scala
```diff
@@ -0,0 +1,93 @@
...
+import org.apache.spark.rdd.RDD
+import org.apache.spark.sql.catalyst.InternalRow
+import org.apache.spark.sql.catalyst.expressions.{Attribute, UnsafeProjection}
+import org.apache.spark.sql.catalyst.plans.logical.EventTimeWatermark
+import org.apache.spark.sql.execution.SparkPlan
+import org.apache.spark.sql.types.MetadataBuilder
+import org.apache.spark.unsafe.types.CalendarInterval
+import org.apache.spark.util.AccumulatorV2
+
+class MaxLong(protected var currentValue: Long = 0)
```
zsxwing: nit: `protected` -> `private`

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/EventTimeWatermarkExec.scala
```diff
@@ -0,0 +1,93 @@
...
+  extends AccumulatorV2[Long, Long]
+  with Serializable {
```
zsxwing: nit: not needed. `AccumulatorV2` is already `Serializable`.

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/state/StateStore.scala
```diff
@@ -99,13 +99,16 @@ trait StateStoreProvider {
...
+  def key: UnsafeRow
+  def value: UnsafeRow
+}
 
 case class ValueAdded(key: UnsafeRow, value: UnsafeRow) extends StoreUpdate
 
 case class ValueUpdated(key: UnsafeRow, value: UnsafeRow) extends StoreUpdate
 
-case class KeyRemoved(key: UnsafeRow) extends StoreUpdate
+case class ValueRemoved(key: UnsafeRow, value: UnsafeRow) extends StoreUpdate
```
zsxwing: Any special reason to change this? It seems weird that adding an unused field `value`.

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamExecution.scala
```diff
@@ -437,6 +441,22 @@ class StreamExecution(
...
+      case e: EventTimeWatermarkExec =>
+        logTrace(s"Maximum observed eventTime: ${e.maxEventTime.value}")
+        (e.maxEventTime.value / 1000) - e.delay.milliseconds()
+    }.headOption.foreach { newWatermark =>
+      if (newWatermark > currentEventTimeWatermark) {
+        logInfo(s"Updating eventTime watermark to: $newWatermark ms")
+        streamMetrics.reportTriggerDetail(EVENT_TIME_WATERMARK, newWatermark)
+        currentEventTimeWatermark = newWatermark
+      } else {
+        logTrace(s"Event time didn't move: $newWatermark < $currentEventTimeWatermark")
```
zsxwing: We need to call `streamMetrics.reportTriggerDetail(EVENT_TIME_WATERMARK, newWatermark)` here. Otherwise, the trigger details won't have `EVENT_TIME_WATERMARK` for this batch.

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamExecution.scala
```diff
@@ -437,6 +441,22 @@ class StreamExecution(
...
+    // Update the eventTime watermark if we find one in the plan.
+    // TODO: Does this need to be an AttributeMap?
+    lastExecution.executedPlan.collect {
+      case e: EventTimeWatermarkExec =>
+        logTrace(s"Maximum observed eventTime: ${e.maxEventTime.value}")
+        (e.maxEventTime.value / 1000) - e.delay.milliseconds()
+    }.headOption.foreach { newWatermark =>
+      if (newWatermark > currentEventTimeWatermark) {
+        logInfo(s"Updating eventTime watermark to: $newWatermark ms")
+        streamMetrics.reportTriggerDetail(EVENT_TIME_WATERMARK, newWatermark)
```
zsxwing: Is it fine to just set `EVENT_TIME_WATERMARK` to `0` if the first batch doesn't have any data (E.g., the filter specified by the user drops all data)?

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/EventTimeWatermarkExec.scala
zsxwing: nit: Could you document that this one only support positive longs?

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/EventTimeWatermarkExec.scala
```diff
@@ -0,0 +1,93 @@
...
+
+  // TODO: Use Spark SQL Metrics?
+  val maxEventTime = new MaxLong
+  sparkContext.register(maxEventTime)
+
+  override protected def doExecute(): RDD[InternalRow] = {
+    child.execute().mapPartitions { iter =>
+      val getEventTime = UnsafeProjection.create(eventTime :: Nil, child.output)
+      iter.map { row =>
+        maxEventTime.add(getEventTime(row).getLong(0))
```
zsxwing: Just a small question: which place will check the `eventTime` type? I guess `getLong` just throws an exception if the format is wrong. Can we fail it before starting the spark job?

sql/core/src/test/scala/org/apache/spark/sql/streaming/WatermarkSuite.scala
```diff
@@ -0,0 +1,181 @@
...
+    testStream(windowedAggregation)(
+      AddData(inputData, 10, 11, 12),
+      CheckAnswer(),
+      AddData(inputData, 25),     // Advance watermark to 15 seconds
+      CheckAnswer(),
+      AddData(inputData, 25),     // Evict items less than previous watermark.
+      CheckAnswer((10, 3)),
+      AddData(inputData, 10),     // 10 is later than 15 second watermark
+      CheckAnswer((10, 3)),
+      AddData(inputData, 25),     // 10 is later than 15 second watermark
```
zsxwing: nit: the comment is wrong

zsxwing (commented):
Looks good overall. My comments can be addressed later.

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/state/StateStore.scala
marmbrus: It is used.  We need the value to emit the result upon eviction.

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamExecution.scala
zsxwing: I suggest just fixing it since it's pretty easy. Just `if (newWatermark == 0) "-" else newWatermark.toString`

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/EventTimeWatermarkExec.scala
marmbrus: Added to checkAnalysis.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/CheckAnalysis.scala
```diff
@@ -155,6 +155,16 @@ trait CheckAnalysis extends PredicateHelper {
         }
 
         operator match {
+          case etw: EventTimeWatermark =>
+            etw.eventTime.dataType match {
+              case s: StructType
+                if s.find(_.name == "start").map(_.dataType).contains(TimestampType) =>
```
zsxwing: nit: `Option.contains` is not in Scala 2.10.

marmbrus: really? lame...

marmbrus: Oh... it should also check the end of the window, not the start...

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/CheckAnalysis.scala
```diff
@@ -155,6 +155,16 @@ trait CheckAnalysis extends PredicateHelper {
         }
 
         operator match {
+          case etw: EventTimeWatermark =>
+            etw.eventTime.dataType match {
+              case s: StructType
```
tdas: Which high level case is caught by this condition?

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/EventTimeWatermarkExec.scala
```diff
@@ -0,0 +1,93 @@
...
+  override def isZero: Boolean = value == 0
+  override def value: Long = currentValue
+  override def copy(): AccumulatorV2[Long, Long] = new MaxLong(currentValue)
+
+  override def reset(): Unit = {
+    currentValue = 0
+  }
+
+  override def add(v: Long): Unit = {
+    if (value < v) { currentValue = v }
```
tdas: nit: less confusing to read  if `if (currentValue < v) { currentValue = v }`. 
In fact why not used math.max?

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/EventTimeWatermarkExec.scala
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  override def merge(other: AccumulatorV2[Long, Long]): Unit = {
+    if (currentValue < other.value) {
```
tdas: nit: same as above, why not use math.max

tdas (commented):
Major feedback - Python API for `withWatermark()`?
Other than its looking good

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/CheckAnalysis.scala
marmbrus: The result of a window operation.

tdas:
LGTM, pending tests.

marmbrus:
jenkins, test this please

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StatefulAggregate.scala
amitsela: Generally, updates should be able to take into account late arrivals (in respect to Watermark) and allow to act upon a user defined strategy, such as: `update for each following element`.

marmbrus:
jenkins test this please

tdas:
I am merging this to master and 2.1

PR #15703, opened 2016-10-31
[SPARK-18186] Migrate HiveUDAFFunction to TypedImperativeAggregate for partial aggregation support

liancheng:

While being evaluated in Spark SQL, Hive UDAFs don't support partial aggregation. This PR migrates `HiveUDAFFunction`s to `TypedImperativeAggregate`, which already provides partial aggregation support for aggregate functions that may use arbitrary Java objects as aggregation states.

The following snippet shows the effect of this PR:

```scala
import org.apache.hadoop.hive.ql.udf.generic.GenericUDAFMax
sql(s"CREATE FUNCTION hive_max AS '${classOf[GenericUDAFMax].getName}'")

spark.range(100).createOrReplaceTempView("t")

// A query using both Spark SQL native `max` and Hive `max`
sql(s"SELECT max(id), hive_max(id) FROM t").explain()
```

Before this PR:

```
== Physical Plan ==
SortAggregate(key=[], functions=[max(id#1L), default.hive_max(default.hive_max, HiveFunctionWrapper(org.apache.hadoop.hive.ql.udf.generic.GenericUDAFMax,org.apache.hadoop.hive.ql.udf.generic.GenericUDAFMax@7475f57e), id#1L, false, 0, 0)])
+- Exchange SinglePartition
   +- *Range (0, 100, step=1, splits=Some(1))
```

After this PR:

```
== Physical Plan ==
SortAggregate(key=[], functions=[max(id#1L), default.hive_max(default.hive_max, HiveFunctionWrapper(org.apache.hadoop.hive.ql.udf.generic.GenericUDAFMax,org.apache.hadoop.hive.ql.udf.generic.GenericUDAFMax@5e18a6a7), id#1L, false, 0, 0)])
+- Exchange SinglePartition
   +- SortAggregate(key=[], functions=[partial_max(id#1L), partial_default.hive_max(default.hive_max, HiveFunctionWrapper(org.apache.hadoop.hive.ql.udf.generic.GenericUDAFMax,org.apache.hadoop.hive.ql.udf.generic.GenericUDAFMax@5e18a6a7), id#1L, false, 0, 0)])
      +- *Range (0, 100, step=1, splits=Some(1))
```

The tricky part of the PR is mostly about updating and passing around aggregation states of `HiveUDAFFunction`s since the aggregation state of a Hive UDAF may appear in three different forms. Let's take a look at the testing `MockUDAF` added in this PR as an example. This UDAF computes the count of non-null values together with the count of nulls of a given column. Its aggregation state may appear as the following forms at different time:

1. A `MockUDAFBuffer`, which is a concrete subclass of `GenericUDAFEvaluator.AggregationBuffer`

   The form used by Hive UDAF API. This form is required by the following scenarios:

   - Calling `GenericUDAFEvaluator.iterate()` to update an existing aggregation state with new input values.
   - Calling `GenericUDAFEvaluator.terminate()` to get the final aggregated value from an existing aggregation state.
   - Calling `GenericUDAFEvaluator.merge()` to merge other aggregation states into an existing aggregation state.

     The existing aggregation state to be updated must be in this form.

   Conversions:

   - To form 2:

     `GenericUDAFEvaluator.terminatePartial()`

   - To form 3:

     Convert to form 2 first, and then to 3.

2. An `Object[]` array containing two `java.lang.Long` values.

   The form used to interact with Hive's `ObjectInspector`s. This form is required by the following scenarios:

   - Calling `GenericUDAFEvaluator.terminatePartial()` to convert an existing aggregation state in form 1 to form 2.
   - Calling `GenericUDAFEvaluator.merge()` to merge other aggregation states into an existing aggregation state.

     The input aggregation state must be in this form.

   Conversions:

   - To form 1:

     No direct method. Have to create an empty `AggregationBuffer` and merge it into the empty buffer.

   - To form 3:

     `unwrapperFor()`/`unwrap()` method of `HiveInspectors`

3. The byte array that holds data of an `UnsafeRow` with two `LongType` fields.

   The form used by Spark SQL to shuffle partial aggregation results. This form is required because `TypedImperativeAggregate` always asks its subclasses to serialize their aggregation states into a byte array.

   Conversions:

   - To form 1:

     Convert to form 2 first, and then to 1.

   - To form 2:

     `wrapperFor()`/`wrap()` method of `HiveInspectors`

Here're some micro-benchmark results produced by the most recent master and this PR branch.

Master:

```
Java HotSpot(TM) 64-Bit Server VM 1.8.0_92-b14 on Mac OS X 10.10.5
Intel(R) Core(TM) i7-4960HQ CPU @ 2.60GHz

hive udaf vs spark af:                   Best/Avg Time(ms)    Rate(M/s)   Per Row(ns)   Relative
w/o groupBy                                    339 /  372          3.1         323.2       1.0X
w/ groupBy                                     503 /  529          2.1         479.7       0.7X
```

This PR:

```
Java HotSpot(TM) 64-Bit Server VM 1.8.0_92-b14 on Mac OS X 10.10.5
Intel(R) Core(TM) i7-4960HQ CPU @ 2.60GHz

hive udaf vs spark af:                   Best/Avg Time(ms)    Rate(M/s)   Per Row(ns)   Relative
w/o groupBy                                    116 /  126          9.0         110.8       1.0X
w/ groupBy                                     151 /  159          6.9         144.0       0.8X
```

Benchmark code snippet:

```scala
  test("Hive UDAF benchmark") {
    val N = 1 << 20

    sparkSession.sql(s"CREATE TEMPORARY FUNCTION hive_max AS '${classOf[GenericUDAFMax].getName}'")

    val benchmark = new Benchmark(
      name = "hive udaf vs spark af",
      valuesPerIteration = N,
      minNumIters = 5,
      warmupTime = 5.seconds,
      minTime = 5.seconds,
      outputPerIteration = true
    )

    benchmark.addCase("w/o groupBy") { _ =>
      sparkSession.range(N).agg("id" -> "hive_max").collect()
    }

    benchmark.addCase("w/ groupBy") { _ =>
      sparkSession.range(N).groupBy($"id" % 10).agg("id" -> "hive_max").collect()
    }

    benchmark.run()

    sparkSession.sql(s"DROP TEMPORARY FUNCTION IF EXISTS hive_max")
  }
```

## How was this patch tested?

New test suite `HiveUDAFSuite` is added.

### DIFF
--- a/sql/hive/src/main/scala/org/apache/spark/sql/hive/hiveUDFs.scala
+++ b/sql/hive/src/main/scala/org/apache/spark/sql/hive/hiveUDFs.scala
@@ -302,43 +302,61 @@ private[hive] case class HiveUDAFFunction(
       funcWrapper.createFunction[AbstractGenericUDAFResolver]()
     }
 
+  // Hive `ObjectInspector`s for all child expressions (input parameters of the function).
   @transient
   private lazy val inputInspectors = children.map(toInspector).toArray
 
+  // Spark SQL data types of input parameters.
   @transient
-  val parameterInfo = new SimpleGenericUDAFParameterInfo(inputInspectors, false, false)
+  private lazy val inputDataTypes: Array[DataType] = children.map(_.dataType).toArray
 
+  // The UDAF evaluator used to consume raw input rows and produce partial aggregation results.
   @transient
-  private lazy val partial1ModeEvaluator = resolver.getEvaluator(parameterInfo)
+  private lazy val partial1ModeEvaluator = {
+    val parameterInfo = new SimpleGenericUDAFParameterInfo(inputInspectors, false, false)
+    resolver.getEvaluator(parameterInfo)
+  }
 
+  // Hive `ObjectInspector` used to inspect partial aggregation results.
   @transient
   private val partialResultInspector = partial1ModeEvaluator.init(
     GenericUDAFEvaluator.Mode.PARTIAL1,
     inputInspectors
   )
 
+  // The UDAF evaluator used to merge partial aggregation results.
   @transient
   private lazy val partial2ModeEvaluator = {
+    val parameterInfo = new SimpleGenericUDAFParameterInfo(inputInspectors, false, false)
     val evaluator = resolver.getEvaluator(parameterInfo)
     evaluator.init(GenericUDAFEvaluator.Mode.PARTIAL2, Array(partialResultInspector))
     evaluator
   }
 
+  // Spark SQL data type of partial aggregation results
   @transient
-  private lazy val finalModeEvaluator = resolver.getEvaluator(parameterInfo)
+  private lazy val partialResultDataType = inspectorToDataType(partialResultInspector)
 
+  // The UDAF evaluator used to compute the final result from a partial aggregation result objects.
+  @transient
+  private lazy val finalModeEvaluator = {
+    val parameterInfo = new SimpleGenericUDAFParameterInfo(inputInspectors, false, false)
+    resolver.getEvaluator(parameterInfo)
+  }
+
+  // Hive `ObjectInspector` used to inspect the final aggregation result object.
   @transient
   private val returnInspector = finalModeEvaluator.init(
     GenericUDAFEvaluator.Mode.FINAL,
     Array(partialResultInspector)
   )
 
+  // Wrapper functions used to wrap Spark SQL input arguments into Hive specific format.
   @transient
-  private lazy val wrappers = children.map(x => wrapperFor(toInspector(x), x.dataType)).toArray
+  private lazy val inputWrappers = children.map(x => wrapperFor(toInspector(x), x.dataType)).toArray
 
-  @transient
-  private lazy val partialResultDataType = inspectorToDataType(partialResultInspector)
-
+  // Unwrapper function used to unwrap final aggregation result objects returned by Hive UDAFs into
+  // Spark SQL specific format.
   @transient
   private lazy val resultUnwrapper = unwrapperFor(returnInspector)
 
@@ -346,10 +364,7 @@ private[hive] case class HiveUDAFFunction(
   private lazy val cached: Array[AnyRef] = new Array[AnyRef](children.length)
 
   @transient
-  private lazy val inputDataTypes: Array[DataType] = children.map(_.dataType).toArray
-
-  @transient
-  private lazy val bufferSerDe: AggregationBufferSerDe = new AggregationBufferSerDe
+  private lazy val aggBufferSerDe: AggregationBufferSerDe = new AggregationBufferSerDe
 
   // We rely on Hive to check the input data types, so use `AnyDataType` here to bypass our
   // catalyst type checking framework.
@@ -376,7 +391,7 @@ private[hive] case class HiveUDAFFunction(
 
   override def update(buffer: AggregationBuffer, input: InternalRow): Unit = {
     partial1ModeEvaluator.iterate(
-      buffer, wrap(inputProjection(input), wrappers, cached, inputDataTypes))
+      buffer, wrap(inputProjection(input), inputWrappers, cached, inputDataTypes))
   }
 
   override def merge(buffer: AggregationBuffer, input: AggregationBuffer): Unit = {
@@ -388,13 +403,14 @@ private[hive] case class HiveUDAFFunction(
   }
 
   override def serialize(buffer: AggregationBuffer): Array[Byte] = {
-    bufferSerDe.serialize(buffer)
+    aggBufferSerDe.serialize(buffer)
   }
 
   override def deserialize(bytes: Array[Byte]): AggregationBuffer = {
-    bufferSerDe.deserialize(bytes)
+    aggBufferSerDe.deserialize(bytes)
   }
 
+  // Helper class used to de/serialize Hive UDAF `AggregationBuffer` objects
   private class AggregationBufferSerDe {
     private val partialResultUnwrapper = unwrapperFor(partialResultInspector)
 
--- a/sql/hive/src/main/scala/org/apache/spark/sql/hive/hiveUDFs.scala
+++ b/sql/hive/src/main/scala/org/apache/spark/sql/hive/hiveUDFs.scala
@@ -37,7 +37,6 @@ import org.apache.spark.sql.catalyst.expressions._
 import org.apache.spark.sql.catalyst.expressions.aggregate._
 import org.apache.spark.sql.catalyst.expressions.codegen.CodegenFallback
 import org.apache.spark.sql.hive.HiveShim._
-import org.apache.spark.sql.hive.HiveUDAFFunction.AggregationBufferSerDe
 import org.apache.spark.sql.types._
 
 
@@ -304,33 +303,41 @@ private[hive] case class HiveUDAFFunction(
     }
 
   @transient
-  private lazy val inspectors = children.map(toInspector).toArray
+  private lazy val inputInspectors = children.map(toInspector).toArray
 
   @transient
-  private lazy val function = {
-    val parameterInfo = new SimpleGenericUDAFParameterInfo(inspectors, false, false)
-    resolver.getEvaluator(parameterInfo)
+  val parameterInfo = new SimpleGenericUDAFParameterInfo(inputInspectors, false, false)
+
+  @transient
+  private lazy val partial1ModeEvaluator = resolver.getEvaluator(parameterInfo)
+
+  @transient
+  private val partialResultInspector = partial1ModeEvaluator.init(
+    GenericUDAFEvaluator.Mode.PARTIAL1,
+    inputInspectors
+  )
+
+  @transient
+  private lazy val partial2ModeEvaluator = {
+    val evaluator = resolver.getEvaluator(parameterInfo)
+    evaluator.init(GenericUDAFEvaluator.Mode.PARTIAL2, Array(partialResultInspector))
+    evaluator
   }
+
+  @transient
+  private lazy val finalModeEvaluator = resolver.getEvaluator(parameterInfo)
+
+  @transient
+  private val returnInspector = finalModeEvaluator.init(
+    GenericUDAFEvaluator.Mode.FINAL,
+    Array(partialResultInspector)
+  )
 
   @transient
   private lazy val wrappers = children.map(x => wrapperFor(toInspector(x), x.dataType)).toArray
 
   @transient
-  private val returnInspector =
-    function.init(GenericUDAFEvaluator.Mode.COMPLETE, inspectors)
-
-  @transient
-  private val partialResultInspector =
-    function.init(GenericUDAFEvaluator.Mode.PARTIAL1, inspectors)
-
-  @transient
   private lazy val partialResultDataType = inspectorToDataType(partialResultInspector)
-
-  @transient
-  private lazy val partialResultUnwrapper = unwrapperFor(partialResultInspector)
-
-  @transient
-  private lazy val partialResultWrapper = wrapperFor(partialResultInspector, partialResultDataType)
 
   @transient
   private lazy val resultUnwrapper = unwrapperFor(returnInspector)
@@ -342,9 +349,7 @@ private[hive] case class HiveUDAFFunction(
   private lazy val inputDataTypes: Array[DataType] = children.map(_.dataType).toArray
 
   @transient
-  private lazy val bufferSerDe: AggregationBufferSerDe =
-    new AggregationBufferSerDe(
-      function, partialResultDataType, partialResultUnwrapper, partialResultWrapper)
+  private lazy val bufferSerDe: AggregationBufferSerDe = new AggregationBufferSerDe
 
   // We rely on Hive to check the input data types, so use `AnyDataType` here to bypass our
   // catalyst type checking framework.
@@ -363,20 +368,24 @@ private[hive] case class HiveUDAFFunction(
     s"$name($distinct${children.map(_.sql).mkString(", ")})"
   }
 
-  override def createAggregationBuffer(): AggregationBuffer = function.getNewAggregationBuffer
+  override def createAggregationBuffer(): AggregationBuffer =
+    partial1ModeEvaluator.getNewAggregationBuffer
 
   @transient
   private lazy val inputProjection = new InterpretedProjection(children)
 
   override def update(buffer: AggregationBuffer, input: InternalRow): Unit = {
-    function.iterate(buffer, wrap(inputProjection(input), wrappers, cached, inputDataTypes))
+    partial1ModeEvaluator.iterate(
+      buffer, wrap(inputProjection(input), wrappers, cached, inputDataTypes))
   }
 
   override def merge(buffer: AggregationBuffer, input: AggregationBuffer): Unit = {
-    function.merge(buffer, function.terminatePartial(input))
+    partial2ModeEvaluator.merge(buffer, partial1ModeEvaluator.terminatePartial(input))
   }
 
-  override def eval(buffer: AggregationBuffer): Any = resultUnwrapper(function.evaluate(buffer))
+  override def eval(buffer: AggregationBuffer): Any = {
+    resultUnwrapper(finalModeEvaluator.terminate(buffer))
+  }
 
   override def serialize(buffer: AggregationBuffer): Array[Byte] = {
     bufferSerDe.serialize(buffer)
@@ -385,15 +394,11 @@ private[hive] case class HiveUDAFFunction(
   override def deserialize(bytes: Array[Byte]): AggregationBuffer = {
     bufferSerDe.deserialize(bytes)
   }
-}
 
-private[hive] object HiveUDAFFunction {
-  class AggregationBufferSerDe(
-      function: GenericUDAFEvaluator,
-      partialResultDataType: DataType,
-      partialResultUnwrapper: Any => Any,
-      partialResultWrapper: Any => Any)
-    extends HiveInspectors {
+  private class AggregationBufferSerDe {
+    private val partialResultUnwrapper = unwrapperFor(partialResultInspector)
+
+    private val partialResultWrapper = wrapperFor(partialResultInspector, partialResultDataType)
 
     private val projection = UnsafeProjection.create(Array(partialResultDataType))
 
@@ -403,7 +408,7 @@ private[hive] object HiveUDAFFunction {
       // `GenericUDAFEvaluator.terminatePartial()` converts an `AggregationBuffer` into an object
       // that can be inspected by the `ObjectInspector` returned by `GenericUDAFEvaluator.init()`.
       // Then we can unwrap it to a Spark SQL value.
-      mutableRow.update(0, partialResultUnwrapper(function.terminatePartial(buffer)))
+      mutableRow.update(0, partialResultUnwrapper(partial1ModeEvaluator.terminatePartial(buffer)))
       val unsafeRow = projection(mutableRow)
       val bytes = ByteBuffer.allocate(unsafeRow.getSizeInBytes)
       unsafeRow.writeTo(bytes)
@@ -415,11 +420,11 @@ private[hive] object HiveUDAFFunction {
       // returned by `GenericUDAFEvaluator.terminatePartial()` back to an `AggregationBuffer`. The
       // workaround here is creating an initial `AggregationBuffer` first and then merge the
       // deserialized object into the buffer.
-      val buffer = function.getNewAggregationBuffer
+      val buffer = partial2ModeEvaluator.getNewAggregationBuffer
       val unsafeRow = new UnsafeRow(1)
       unsafeRow.pointTo(bytes, bytes.length)
       val partialResult = unsafeRow.get(0, partialResultDataType)
-      function.merge(buffer, partialResultWrapper(partialResult))
+      partial2ModeEvaluator.merge(buffer, partialResultWrapper(partialResult))
       buffer
     }
   }
--- a/sql/hive/src/main/scala/org/apache/spark/sql/hive/hiveUDFs.scala
+++ b/sql/hive/src/main/scala/org/apache/spark/sql/hive/hiveUDFs.scala
@@ -23,9 +23,9 @@ import scala.collection.JavaConverters._
 import scala.collection.mutable.ArrayBuffer
 
 import org.apache.hadoop.hive.ql.exec._
-import org.apache.hadoop.hive.ql.udf.generic.GenericUDAFEvaluator.AggregationBuffer
 import org.apache.hadoop.hive.ql.udf.{UDFType => HiveUDFType}
 import org.apache.hadoop.hive.ql.udf.generic._
+import org.apache.hadoop.hive.ql.udf.generic.GenericUDAFEvaluator.AggregationBuffer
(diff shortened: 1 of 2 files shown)


liancheng:
Will add more details in the PR description soon.

rxin:
cc @tejasapatil

liancheng:
I can't reproduce those test failures when executing failed test cases individually. Seems that it's related to execution order. Still investigating.

sql/hive/src/main/scala/org/apache/spark/sql/hive/hiveUDFs.scala
```diff
@@ -293,69 +307,57 @@ private[hive] case class HiveUDAFFunction(
...
-  private lazy val wrappers = children.map(x => wrapperFor(toInspector(x), x.dataType)).toArray
+  private lazy val returnInspector =
+    function.init(GenericUDAFEvaluator.Mode.COMPLETE, inspectors)
 
   @transient
-  private lazy val returnInspector = functionAndInspector._2
+  private lazy val partialResultInspector =
+    function.init(GenericUDAFEvaluator.Mode.PARTIAL1, inspectors)
+
+  // The following two lines initializes `function: GenericUDAFEvaluator` eagerly. These two fields
```
tejasapatil: This feels little weird. If you want eager init and not serialize, why not remove `lazy` and just keep it `@transient` ?

tejasapatil:
This will surely improve performance for the UDAFs where the data shrinks (eg. `max` as you pointed out). I am not sure if it would be better for UDAFs like `GenericUDAFCollectSet`, `GenericUDAFCollectList` where the aggregation does not shrink the data (it might be worse because of conversion cost ?).

It would be good to see some perf numbers to compare (native Spark UDFs) vs (before this change) vs (after the change).

liancheng:
@tejasapatil For `collect_set` and `collect_list`, we'll simply migrate them to `TypedImperativeAggregate` and so that they become Spark native aggregate functions. We can also handle other built-in Hive UDAFs that share the same issue in this approach.

In the long run, this PR is mostly useful for running legacy UDAFs developed by the users rather than Hive built-in ones since we'd prefer to migrate all the built-in ones to Spark at last.

liancheng:
I found that I'm handling bridged UDAFs properly, which caused a few test failures. Working on it.

liancheng:
@tejasapatil Another point that I'd like to add is that even if the performance for a single UDAF like `GenericUDAFCollectList` regresses, you still have performance gains if such UDAFs are used together with other aggregate functions within a single query. Before this PR, all the aggregate functions have to be evaluated without partial aggregation support in this case.

tejasapatil:
There is no doubt that this PR is not going to make things better. I'm already sold :)

liancheng:
It turned out that I didn't initialize Hive UDAF evaluators properly.

Quoted from commit message of my previous commit:

> Hive UDAFs are sensitive to aggregation mode, and must be initialized with proper modes before being used. Basically, it means that you can't use an evaluator initialized with mode `PARTIAL1` to merge two aggregation states (although it still works for aggregate functions whose partial result type is the same as the final result type).

liancheng:
OK, now it's ready for review and merge.

cc @yhuai @JoshRosen @cloud-fan

sql/hive/src/main/scala/org/apache/spark/sql/hive/hiveUDFs.scala
```diff
@@ -289,73 +302,77 @@ private[hive] case class HiveUDAFFunction(
...
+  private lazy val partialResultDataType = inspectorToDataType(partialResultInspector)
 
+  // The UDAF evaluator used to compute the final result from a partial aggregation result objects.
   @transient
-  private[this] var buffer: GenericUDAFEvaluator.AggregationBuffer = _
-
-  override def eval(input: InternalRow): Any = unwrapper(function.evaluate(buffer))
+  private lazy val finalModeEvaluator = {
+    val parameterInfo = new SimpleGenericUDAFParameterInfo(inputInspectors, false, false)
+    resolver.getEvaluator(parameterInfo)
```
cloud-fan: These 2 lines are duplicated many times, should we abstract them to a method?

sql/hive/src/main/scala/org/apache/spark/sql/hive/hiveUDFs.scala
```diff
@@ -365,4 +382,66 @@ private[hive] case class HiveUDAFFunction(
...
+    aggBufferSerDe.deserialize(bytes)
+  }
+
+  // Helper class used to de/serialize Hive UDAF `AggregationBuffer` objects
+  private class AggregationBufferSerDe {
+    private val partialResultUnwrapper = unwrapperFor(partialResultInspector)
+
+    private val partialResultWrapper = wrapperFor(partialResultInspector, partialResultDataType)
+
+    private val projection = UnsafeProjection.create(Array(partialResultDataType))
```
cloud-fan: Can you try to run hive udaf in spark shell? IIUC, we can't create unsafe projection inside UDAF

liancheng: It does work as expected:

```
scala> sql("CREATE TEMPORARY FUNCTION hive_max AS 'org.apache.hadoop.hive.ql.udf.generic.GenericUDAFMax'")

scala> spark.range(100).createOrReplaceTempView("t")

scala> sql("SELECT hive_max(id) FROM t").explain()
== Physical Plan ==
SortAggregate(key=[], functions=[hive_max(hive_max, HiveFunctionWrapper(org.apache.hadoop.hive.ql.udf.generic.GenericUDAFMax,org.apache.hadoop.hive.ql.udf.generic.GenericUDAFMax@144792d5), id#1L, false, 0, 0)])
+- Exchange SinglePartition
   +- SortAggregate(key=[], functions=[partial_hive_max(hive_max, HiveFunctionWrapper(org.apache.hadoop.hive.ql.udf.generic.GenericUDAFMax,org.apache.hadoop.hive.ql.udf.generic.GenericUDAFMax@144792d5), id#1L, false, 0, 0)])
      +- *Range (0, 100, step=1, splits=Some(8))

scala> sql("SELECT hive_max(id) FROM t").show()
+-------------+
|hive_max( id)|
+-------------+
|           99|
+-------------+
```

Why do you think `UnsafeProjection` can't be used here?

cloud-fan: I tried to use unsafe projection in `percentile_approx` before, but failed in spark shell, maybe it's a different problem, nvm.

liancheng:
The last build failure was because of a logical conflict between this PR and the master branch. Resolving it.

sql/hive/src/main/scala/org/apache/spark/sql/hive/hiveUDFs.scala
```diff
@@ -289,73 +302,75 @@ private[hive] case class HiveUDAFFunction(
...
+
+  private def newEvaluator(): GenericUDAFEvaluator = {
+    val parameterInfo = new SimpleGenericUDAFParameterInfo(inputInspectors, false, false)
+    resolver.getEvaluator(parameterInfo)
   }
 
+  // The UDAF evaluator used to consume raw input rows and produce partial aggregation results.
   @transient
-  private lazy val function = functionAndInspector._1
+  private lazy val partial1ModeEvaluator = newEvaluator()
```
yhuai: Do we need to make it a lazy val since partialResultInspector is uses it right below?

sql/hive/src/main/scala/org/apache/spark/sql/hive/hiveUDFs.scala
```diff
@@ -289,73 +302,75 @@ private[hive] case class HiveUDAFFunction(
...
+  private lazy val aggBufferSerDe: AggregationBufferSerDe = new AggregationBufferSerDe
 
   // We rely on Hive to check the input data types, so use `AnyDataType` here to bypass our
   // catalyst type checking framework.
   override def inputTypes: Seq[AbstractDataType] = children.map(_ => AnyDataType)
 
   override def nullable: Boolean = true
 
-  override def supportsPartial: Boolean = false
+  override def supportsPartial: Boolean = true
```
yhuai: Is there any Hive UDAF that does not support partial aggregation?

sql/hive/src/main/scala/org/apache/spark/sql/hive/hiveUDFs.scala
```diff
@@ -365,4 +380,66 @@ private[hive] case class HiveUDAFFunction(
     val distinct = if (isDistinct) "DISTINCT " else " "
     s"$name($distinct${children.map(_.sql).mkString(", ")})"
   }
+
+  override def createAggregationBuffer(): AggregationBuffer =
+    partial1ModeEvaluator.getNewAggregationBuffer
+
+  @transient
+  private lazy val inputProjection = new InterpretedProjection(children)
```
yhuai: Why `InterpretedProjection`?

sql/hive/src/main/scala/org/apache/spark/sql/hive/hiveUDFs.scala
```diff
@@ -365,4 +380,66 @@ private[hive] case class HiveUDAFFunction(
...
+    resultUnwrapper(finalModeEvaluator.terminate(buffer))
+  }
+
+  override def serialize(buffer: AggregationBuffer): Array[Byte] = {
+    aggBufferSerDe.serialize(buffer)
+  }
+
+  override def deserialize(bytes: Array[Byte]): AggregationBuffer = {
+    aggBufferSerDe.deserialize(bytes)
+  }
```
yhuai: Let's add docs to explain what these functions are doing.

sql/hive/src/main/scala/org/apache/spark/sql/hive/hiveUDFs.scala
```diff
@@ -365,4 +380,66 @@ private[hive] case class HiveUDAFFunction(
...
+  @transient
+  private lazy val inputProjection = new InterpretedProjection(children)
+
+  override def update(buffer: AggregationBuffer, input: InternalRow): Unit = {
+    partial1ModeEvaluator.iterate(
+      buffer, wrap(inputProjection(input), inputWrappers, cached, inputDataTypes))
+  }
+
+  override def merge(buffer: AggregationBuffer, input: AggregationBuffer): Unit = {
+    partial2ModeEvaluator.merge(buffer, partial1ModeEvaluator.terminatePartial(input))
```
yhuai: Let's explain what we are trying to do using `partial1ModeEvaluator.terminatePartial(input)`.

sql/hive/src/main/scala/org/apache/spark/sql/hive/hiveUDFs.scala
```diff
@@ -365,4 +380,66 @@ private[hive] case class HiveUDAFFunction(
...
+  private class AggregationBufferSerDe {
+    private val partialResultUnwrapper = unwrapperFor(partialResultInspector)
+
+    private val partialResultWrapper = wrapperFor(partialResultInspector, partialResultDataType)
+
+    private val projection = UnsafeProjection.create(Array(partialResultDataType))
+
+    private val mutableRow = new GenericInternalRow(1)
+
+    def serialize(buffer: AggregationBuffer): Array[Byte] = {
```
yhuai: doc?

sql/hive/src/test/scala/org/apache/spark/sql/hive/execution/HiveUDAFSuite.scala
```diff
@@ -0,0 +1,152 @@
...
+    assert(aggs.length == 2)
+
+    checkAnswer(df, Seq(
+      Row(2),
+      Row(3)
+    ))
+  }
+
+  test("customized Hive UDAF") {
+    val df = sql("SELECT mock(value) FROM t GROUP BY key % 2")
```
yhuai: How about we also keep the key?

sql/hive/src/test/scala/org/apache/spark/sql/hive/execution/HiveUDAFSuite.scala
```diff
@@ -0,0 +1,152 @@
...
+    ).toDF("key", "value").repartition(2).createOrReplaceTempView("t")
+  }
+
+  protected override def afterAll(): Unit = {
+    sql(s"DROP TEMPORARY FUNCTION IF EXISTS mock")
+    sql(s"DROP TEMPORARY FUNCTION IF EXISTS hive_max")
+  }
+
+  test("built-in Hive UDAF") {
+    val df = sql("SELECT hive_max(key) FROM t GROUP BY key % 2")
```
yhuai: Also keep the key?

yhuai (changes requested):

sql/hive/src/main/scala/org/apache/spark/sql/hive/hiveUDFs.scala
```diff
@@ -263,8 +265,19 @@ private[hive] case class HiveGenericUDTF(
...
 /**
- * Currently we don't support partial aggregation for queries using Hive UDAF, which may hurt
- * performance a lot.
+ * While being evaluated by Spark SQL, the aggregation state of a Hive UDAF may be in the following
+ * three formats:
+ *
+ *  1. a Spark SQL value, or
+ *  2. an instance of some concrete `GenericUDAFEvaluator.AggregationBuffer` class, or
+ *  3. a Java object that can be inspected using the `ObjectInspector` returned by the
+ *     `GenericUDAFEvaluator.init()` method.
```
yhuai: Besides of explaining what are these three formats, let's also explain when we will use each of them.

yhuai: (we can just put the pr description to here)

yhuai: (is the doc below enough?)

sql/hive/src/main/scala/org/apache/spark/sql/hive/hiveUDFs.scala
yhuai: Should we always call init to make the code consistent?

sql/hive/src/main/scala/org/apache/spark/sql/hive/hiveUDFs.scala
```diff
@@ -289,73 +302,75 @@ private[hive] case class HiveUDAFFunction(
       funcWrapper.createFunction[AbstractGenericUDAFResolver]()
     }
 
+  // Hive `ObjectInspector`s for all child expressions (input parameters of the function).
   @transient
-  private lazy val inspectors = children.map(toInspector).toArray
+  private lazy val inputInspectors = children.map(toInspector).toArray
```
yhuai: Let's add docs to explain when these internal vals are used (like which vals are needed for a given mode).

sql/hive/src/main/scala/org/apache/spark/sql/hive/hiveUDFs.scala
```diff
@@ -289,73 +302,75 @@ private[hive] case class HiveUDAFFunction(
...
+    val parameterInfo = new SimpleGenericUDAFParameterInfo(inputInspectors, false, false)
+    resolver.getEvaluator(parameterInfo)
   }
 
+  // The UDAF evaluator used to consume raw input rows and produce partial aggregation results.
   @transient
-  private lazy val function = functionAndInspector._1
+  private lazy val partial1ModeEvaluator = newEvaluator()
 
+  // Hive `ObjectInspector` used to inspect partial aggregation results.
```
yhuai: Partial aggregation result is aggregation buffer, right?

sql/hive/src/main/scala/org/apache/spark/sql/hive/hiveUDFs.scala
```diff
@@ -289,73 +302,75 @@ private[hive] case class HiveUDAFFunction(
...
-
-  override val aggBufferAttributes: Seq[AttributeReference] = Nil
+  @transient
+  private lazy val cached: Array[AnyRef] = new Array[AnyRef](children.length)
 
-  // Note: although this simply copies aggBufferAttributes, this common code can not be placed
-  // in the superclass because that will lead to initialization ordering issues.
-  override val inputAggBufferAttributes: Seq[AttributeReference] = Nil
+  @transient
+  private lazy val aggBufferSerDe: AggregationBufferSerDe = new AggregationBufferSerDe
```
yhuai: doc

sql/hive/src/main/scala/org/apache/spark/sql/hive/hiveUDFs.scala
```diff
@@ -365,4 +380,66 @@ private[hive] case class HiveUDAFFunction(
...
+
+    def serialize(buffer: AggregationBuffer): Array[Byte] = {
+      // `GenericUDAFEvaluator.terminatePartial()` converts an `AggregationBuffer` into an object
+      // that can be inspected by the `ObjectInspector` returned by `GenericUDAFEvaluator.init()`.
+      // Then we can unwrap it to a Spark SQL value.
+      mutableRow.update(0, partialResultUnwrapper(partial1ModeEvaluator.terminatePartial(buffer)))
+      val unsafeRow = projection(mutableRow)
+      val bytes = ByteBuffer.allocate(unsafeRow.getSizeInBytes)
+      unsafeRow.writeTo(bytes)
+      bytes.array()
```
yhuai: Should we just use `unsafeRow.getBytes`?

sql/hive/src/main/scala/org/apache/spark/sql/hive/hiveUDFs.scala
liancheng: Thanks, added.

sql/hive/src/main/scala/org/apache/spark/sql/hive/hiveUDFs.scala
liancheng: Yes. It has to be transient and lazy so that it's also available on executor side since Hive UDAF evaluators are not serializable.

sql/hive/src/main/scala/org/apache/spark/sql/hive/hiveUDFs.scala
liancheng: Yea. They are those objects returned by `terminatePartial()`, which is the inspectable version of Hive `AggregationBuffer`.

sql/hive/src/main/scala/org/apache/spark/sql/hive/hiveUDFs.scala
liancheng: Copied from the original code. Replaced with `UnsafeProjection` instead. Thanks.

sql/hive/src/main/scala/org/apache/spark/sql/hive/hiveUDFs.scala
liancheng: Comment added.

sql/hive/src/main/scala/org/apache/spark/sql/hive/hiveUDFs.scala
liancheng: Done.

sql/hive/src/main/scala/org/apache/spark/sql/hive/hiveUDFs.scala
```diff
@@ -365,4 +380,66 @@ private[hive] case class HiveUDAFFunction(
...
+  override def serialize(buffer: AggregationBuffer): Array[Byte] = {
+    aggBufferSerDe.serialize(buffer)
+  }
+
+  override def deserialize(bytes: Array[Byte]): AggregationBuffer = {
+    aggBufferSerDe.deserialize(bytes)
+  }
+
+  // Helper class used to de/serialize Hive UDAF `AggregationBuffer` objects
+  private class AggregationBufferSerDe {
```
liancheng: We can, but it doesn't seem to be necessary. Make it a nested class also simplifies implementation since it has access to fields of the outer class.

sql/hive/src/main/scala/org/apache/spark/sql/hive/hiveUDFs.scala
liancheng: Done.

sql/hive/src/main/scala/org/apache/spark/sql/hive/hiveUDFs.scala
liancheng: Aren't they equivalent in this case? `UnsafeRow.getBytes` also performs some more checks that are not necessary here.

sql/hive/src/test/scala/org/apache/spark/sql/hive/execution/HiveUDAFSuite.scala
liancheng: Done. Thanks.

sql/hive/src/main/scala/org/apache/spark/sql/hive/hiveUDFs.scala
yhuai: ok. I think in general we should avoid of using this pattern. If we have to use it now, let's explain it in the comment.

yhuai:
Code changes looks good to me. Let's also do a benchmark to sanity check our implementation.

yhuai:
LGTM. Merging to master.

liancheng:
Thanks everyone for the review!

sql/hive/src/main/scala/org/apache/spark/sql/hive/hiveUDFs.scala
cloud-fan: but you also create an unnecessary `ByteBuffer`... as they are equivalent, doesn't `unsafeRow.getBytes` simpler?

cloud-fan: Actually they are different. If the buffer type is fixed length, then the `unsafeRow` is just a fixed-length bytes array, and `UnsafeRow.getBytes` will just return that array, instead of copying the memory.

sql/hive/src/main/scala/org/apache/spark/sql/hive/hiveUDFs.scala
```diff
@@ -365,4 +394,74 @@ private[hive] case class HiveUDAFFunction(
...
+    partial1ModeEvaluator.iterate(
+      buffer, wrap(inputProjection(input), inputWrappers, cached, inputDataTypes))
+  }
+
+  override def merge(buffer: AggregationBuffer, input: AggregationBuffer): Unit = {
+    // The 2nd argument of the Hive `GenericUDAFEvaluator.merge()` method is an input aggregation
+    // buffer in the 3rd format mentioned in the ScalaDoc of this class. Originally, Hive converts
+    // this `AggregationBuffer`s into this format before shuffling partial aggregation results, and
+    // calls `GenericUDAFEvaluator.terminatePartial()` to do the conversion.
+    partial2ModeEvaluator.merge(buffer, partial1ModeEvaluator.terminatePartial(input))
```
pgandhi999: If we follow the code flow from interfaces.scala, we see that the results of aggregation buffer mode in PARTIAL2 is merged with the aggregation buffer in PARTIAL1. I am new to Spark and Hive, so just wanted to know the reason behind the above behaviour. If there are any docs suggesting this, do let me know. Thank you.

PR #15704, opened 2016-10-31
[SPARK-17732][SQL] ALTER TABLE DROP PARTITION should support comparators

dongjoon-hyun:

This PR aims to support `comparators`, e.g. '<', '<=', '>', '>=', again in Apache Spark 2.0 for backward compatibility.

**Spark 1.6**

``` scala
scala> sql("CREATE TABLE sales(id INT) PARTITIONED BY (country STRING, quarter STRING)")
res0: org.apache.spark.sql.DataFrame = [result: string]

scala> sql("ALTER TABLE sales DROP PARTITION (country < 'KR')")
res1: org.apache.spark.sql.DataFrame = [result: string]
```

**Spark 2.0**

``` scala
scala> sql("CREATE TABLE sales(id INT) PARTITIONED BY (country STRING, quarter STRING)")

scala> sql("ALTER TABLE sales DROP PARTITION (country < 'KR')")
mismatched input '<' expecting {')', ','}(line 1, pos 42)
```

After this PR, it's supported.

## How was this patch tested?

Pass the Jenkins test with a newly added testcase.

### DIFF
--- a/sql/catalyst/src/main/antlr4/org/apache/spark/sql/catalyst/parser/SqlBase.g4
+++ b/sql/catalyst/src/main/antlr4/org/apache/spark/sql/catalyst/parser/SqlBase.g4
@@ -239,11 +239,7 @@ partitionSpecLocation
     ;
 
 partitionSpec
-    : PARTITION '(' partitionVal (',' partitionVal)* ')'
-    ;
-
-partitionVal
-    : expression
+    : PARTITION '(' expression (',' expression)* ')'
     ;
 
 describeFuncName
--- a/sql/catalyst/src/main/antlr4/org/apache/spark/sql/catalyst/parser/SqlBase.g4
+++ b/sql/catalyst/src/main/antlr4/org/apache/spark/sql/catalyst/parser/SqlBase.g4
@@ -243,7 +243,7 @@ partitionSpec
     ;
 
 partitionVal
-    : identifier (EQ constant)?
+    : identifier (comparisonOperator constant)?
     ;
 
 describeFuncName
--- a/sql/catalyst/src/main/antlr4/org/apache/spark/sql/catalyst/parser/SqlBase.g4
+++ b/sql/catalyst/src/main/antlr4/org/apache/spark/sql/catalyst/parser/SqlBase.g4
@@ -243,7 +243,7 @@ partitionSpec
     ;
 
 partitionVal
-    : identifier (comparisonOperator constant)?
+    : expression
     ;
 
 describeFuncName
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/parser/AstBuilder.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/parser/AstBuilder.scala
@@ -216,8 +216,6 @@ class AstBuilder extends SqlBaseBaseVisitor[AnyRef] with Logging {
     val parts = ctx.partitionVal.asScala.map { pVal =>
       val name = pVal.identifier.getText
       val operator = Option(pVal.comparisonOperator).map(_.getText)
-      validate(operator.nonEmpty && !operator.get.equals("<=>"),
-        "'<=>' operator is not allowed in partition specification.", ctx)
       if (operator.isDefined) {
         val left = AttributeReference(name, DataTypes.StringType)()
         val right = expression(pVal.constant)
@@ -225,6 +223,9 @@ class AstBuilder extends SqlBaseBaseVisitor[AnyRef] with Logging {
         operator.getSymbol.getType match {
           case SqlBaseParser.EQ =>
             EqualTo(left, right)
+          case SqlBaseParser.NSEQ =>
+            throw new ParseException(
+              "'<=>' operator is not allowed in partition specification.", ctx)
           case SqlBaseParser.NEQ | SqlBaseParser.NEQJ =>
             Not(EqualTo(left, right))
           case SqlBaseParser.LT =>
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/parser/AstBuilder.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/parser/AstBuilder.scala
@@ -198,7 +198,7 @@ class AstBuilder extends SqlBaseBaseVisitor[AnyRef] with Logging {
       val name = pVal.identifier.getText
       val operator = Option(pVal.comparisonOperator).map(_.getText)
       validate(operator.isEmpty || operator.get.equals("="),
-        "Only '=' partition specification is allowed.", ctx)
+        "Only '=' operator is allowed for this partition specification.", ctx)
       val value = Option(pVal.constant).map(visitStringConstant)
       name -> value
     }
@@ -216,6 +216,8 @@ class AstBuilder extends SqlBaseBaseVisitor[AnyRef] with Logging {
     val parts = ctx.partitionVal.asScala.map { pVal =>
       val name = pVal.identifier.getText
       val operator = Option(pVal.comparisonOperator).map(_.getText)
+      validate(operator.nonEmpty && !operator.get.equals("<=>"),
+        "'<=>' operator is not allowed in partition specification.", ctx)
       if (operator.isDefined) {
         val left = AttributeReference(name, DataTypes.StringType)()
         val right = expression(pVal.constant)
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/parser/AstBuilder.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/parser/AstBuilder.scala
@@ -196,6 +196,9 @@ class AstBuilder extends SqlBaseBaseVisitor[AnyRef] with Logging {
       ctx: PartitionSpecContext): Map[String, Option[String]] = withOrigin(ctx) {
     val parts = ctx.partitionVal.asScala.map { pVal =>
       val name = pVal.identifier.getText
+      val operator = Option(pVal.comparisonOperator).map(_.getText)
+      validate(operator.isEmpty || operator.get.equals("="),
+        "Only '=' partition specification is allowed.", ctx)
       val value = Option(pVal.constant).map(visitStringConstant)
       name -> value
     }
@@ -204,6 +207,38 @@ class AstBuilder extends SqlBaseBaseVisitor[AnyRef] with Logging {
     // partition columns will be done in analyzer.
     checkDuplicateKeys(parts, ctx)
     parts.toMap
+  }
+
+  /**
+   * Create a partition filter specification.
+   */
+  def visitPartitionFilterSpec(ctx: PartitionSpecContext): Expression = withOrigin(ctx) {
+    val parts = ctx.partitionVal.asScala.map { pVal =>
+      val name = pVal.identifier.getText.toLowerCase
+      val operator = Option(pVal.comparisonOperator).map(_.getText)
+      if (operator.isDefined) {
+        val left = AttributeReference(name, DataTypes.StringType)()
+        val right = expression(pVal.constant)
+        val operator = pVal.comparisonOperator().getChild(0).asInstanceOf[TerminalNode]
+        operator.getSymbol.getType match {
+          case SqlBaseParser.EQ =>
+            EqualTo(left, right)
+          case SqlBaseParser.NEQ | SqlBaseParser.NEQJ =>
+            Not(EqualTo(left, right))
+          case SqlBaseParser.LT =>
+            LessThan(left, right)
+          case SqlBaseParser.LTE =>
+            LessThanOrEqual(left, right)
+          case SqlBaseParser.GT =>
+            GreaterThan(left, right)
+          case SqlBaseParser.GTE =>
+            GreaterThanOrEqual(left, right)
+        }
+      } else {
+        throw new ParseException("Invalid partition filter specification", ctx)
+      }
+    }
+    parts.reduceLeft(And)
   }
 
   /**
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/parser/AstBuilder.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/parser/AstBuilder.scala
@@ -194,7 +194,7 @@ class AstBuilder extends SqlBaseBaseVisitor[AnyRef] with Logging {
    */
   override def visitPartitionSpec(
       ctx: PartitionSpecContext): Map[String, Option[String]] = withOrigin(ctx) {
-    val parts = ctx.partitionVal.asScala.map { pVal =>
+    val parts = ctx.expression.asScala.map { pVal =>
       expression(pVal) match {
         case UnresolvedAttribute(name :: Nil) =>
           name -> None
@@ -215,7 +215,7 @@ class AstBuilder extends SqlBaseBaseVisitor[AnyRef] with Logging {
    * Create a partition filter specification.
    */
   def visitPartitionFilterSpec(ctx: PartitionSpecContext): Expression = withOrigin(ctx) {
-    val parts = ctx.partitionVal.asScala.map { pVal =>
+    val parts = ctx.expression.asScala.map { pVal =>
       expression(pVal) match {
         case EqualNullSafe(_, _) =>
           throw new ParseException("'<=>' operator is not allowed in partition specification.", ctx)
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/parser/AstBuilder.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/parser/AstBuilder.scala
@@ -214,7 +214,7 @@ class AstBuilder extends SqlBaseBaseVisitor[AnyRef] with Logging {
    */
   def visitPartitionFilterSpec(ctx: PartitionSpecContext): Expression = withOrigin(ctx) {
     val parts = ctx.partitionVal.asScala.map { pVal =>
-      val name = pVal.identifier.getText.toLowerCase
+      val name = pVal.identifier.getText
       val operator = Option(pVal.comparisonOperator).map(_.getText)
       if (operator.isDefined) {
         val left = AttributeReference(name, DataTypes.StringType)()
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/parser/AstBuilder.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/parser/AstBuilder.scala
@@ -195,12 +195,14 @@ class AstBuilder extends SqlBaseBaseVisitor[AnyRef] with Logging {
   override def visitPartitionSpec(
       ctx: PartitionSpecContext): Map[String, Option[String]] = withOrigin(ctx) {
     val parts = ctx.partitionVal.asScala.map { pVal =>
-      val name = pVal.identifier.getText
-      val operator = Option(pVal.comparisonOperator).map(_.getText)
-      validate(operator.isEmpty || operator.get.equals("="),
-        "Only '=' operator is allowed for this partition specification.", ctx)
-      val value = Option(pVal.constant).map(visitStringConstant)
-      name -> value
+      expression(pVal) match {
+        case UnresolvedAttribute(name :: Nil) =>
+          name -> None
+        case cmp @ EqualTo(UnresolvedAttribute(name :: Nil), constant: Literal) =>
+          name -> Option(constant.toString)
+        case _ =>
+          throw new ParseException("Invalid partition filter specification", ctx)
+      }
     }
     // Before calling `toMap`, we check duplicated keys to avoid silently ignore partition values
     // in partition spec like PARTITION(a='1', b='2', a='3'). The real semantical check for
@@ -214,31 +216,13 @@ class AstBuilder extends SqlBaseBaseVisitor[AnyRef] with Logging {
    */
   def visitPartitionFilterSpec(ctx: PartitionSpecContext): Expression = withOrigin(ctx) {
     val parts = ctx.partitionVal.asScala.map { pVal =>
-      val name = pVal.identifier.getText
-      val operator = Option(pVal.comparisonOperator).map(_.getText)
-      if (operator.isDefined) {
-        val left = AttributeReference(name, DataTypes.StringType)()
-        val right = expression(pVal.constant)
-        val operator = pVal.comparisonOperator().getChild(0).asInstanceOf[TerminalNode]
-        operator.getSymbol.getType match {
-          case SqlBaseParser.EQ =>
-            EqualTo(left, right)
-          case SqlBaseParser.NSEQ =>
-            throw new ParseException(
-              "'<=>' operator is not allowed in partition specification.", ctx)
-          case SqlBaseParser.NEQ | SqlBaseParser.NEQJ =>
-            Not(EqualTo(left, right))
-          case SqlBaseParser.LT =>
-            LessThan(left, right)
-          case SqlBaseParser.LTE =>
-            LessThanOrEqual(left, right)
-          case SqlBaseParser.GT =>
-            GreaterThan(left, right)
-          case SqlBaseParser.GTE =>
-            GreaterThanOrEqual(left, right)
-        }
-      } else {
-        throw new ParseException("Invalid partition filter specification", ctx)
+      expression(pVal) match {
+        case EqualNullSafe(_, _) =>
+          throw new ParseException("'<=>' operator is not allowed in partition specification.", ctx)
+        case cmp @ BinaryComparison(UnresolvedAttribute(name :: Nil), constant: Literal) =>
+          cmp.withNewChildren(Seq(AttributeReference(name, StringType)(), constant))
+        case _ =>
+          throw new ParseException("Invalid partition filter specification", ctx)
       }
     }
     parts.reduceLeft(And)
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
@@ -445,22 +445,19 @@ case class AlterTableDropPartitionCommand(
     }
 
     if (specs.exists(isRangeComparison)) {
-      val partitionSet = scala.collection.mutable.Set.empty[CatalogTablePartition]
-      specs.foreach { spec =>
-        val partitions = catalog.listPartitionsByFilter(table.identifier, Seq(spec))
-        if (partitions.nonEmpty) {
-          partitionSet ++= partitions
-        } else if (!ifExists) {
+      val partitionSet = specs.flatMap { spec =>
+        val partitions = catalog.listPartitionsByFilter(table.identifier, Seq(spec)).map(_.spec)
+        if (partitions.isEmpty && !ifExists) {
           throw new AnalysisException(s"There is no partition for ${spec.sql}")
         }
-      }
-      catalog.dropPartitions(table.identifier, partitionSet.map(_.spec).toSeq,
-        ignoreIfNotExists = ifExists, purge = purge)
+        partitions
+      }.distinct
+      catalog.dropPartitions(
+        table.identifier, partitionSet, ignoreIfNotExists = ifExists, purge = purge)
     } else {
       val normalizedSpecs = specs.map { expr =>
         val spec = splitConjunctivePredicates(expr).map {
-          case BinaryComparison(left, right) =>
-            left.asInstanceOf[AttributeReference].name -> right.toString
+          case BinaryComparison(AttributeReference(name, _, _, _), right) => name -> right.toString
         }.toMap
         PartitioningUtils.normalizePartitionSpec(
           spec,
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
@@ -31,7 +31,8 @@ import org.apache.spark.sql.catalyst.TableIdentifier
 import org.apache.spark.sql.catalyst.analysis.Resolver
 import org.apache.spark.sql.catalyst.catalog._
 import org.apache.spark.sql.catalyst.catalog.CatalogTypes.TablePartitionSpec
-import org.apache.spark.sql.catalyst.expressions.{Attribute, AttributeReference}
+import org.apache.spark.sql.catalyst.expressions.{Attribute, AttributeReference, BinaryComparison,
+  EqualTo, Expression, PredicateHelper}
 import org.apache.spark.sql.execution.datasources.{CaseInsensitiveMap, PartitioningUtils}
 import org.apache.spark.sql.types._
 import org.apache.spark.util.SerializableConfiguration
@@ -418,10 +419,14 @@ case class AlterTableRenamePartitionCommand(
  */
 case class AlterTableDropPartitionCommand(
     tableName: TableIdentifier,
-    specs: Seq[TablePartitionSpec],
+    specs: Seq[Expression],
     ifExists: Boolean,
     purge: Boolean)
-  extends RunnableCommand {
+  extends RunnableCommand with PredicateHelper {
+
+  private def hasComplexExpr(expr: Expression): Boolean = {
+    expr.find(e => e.isInstanceOf[BinaryComparison] && !e.isInstanceOf[EqualTo]).isDefined
+  }
 
   override def run(sparkSession: SparkSession): Seq[Row] = {
     val catalog = sparkSession.sessionState.catalog
@@ -429,16 +434,26 @@ case class AlterTableDropPartitionCommand(
     DDLUtils.verifyAlterTableType(catalog, table, isView = false)
     DDLUtils.verifyPartitionProviderIsHive(sparkSession, table, "ALTER TABLE DROP PARTITION")
 
-    val normalizedSpecs = specs.map { spec =>
-      PartitioningUtils.normalizePartitionSpec(
-        spec,
-        table.partitionColumnNames,
-        table.identifier.quotedString,
-        sparkSession.sessionState.conf.resolver)
+    if (specs.exists(hasComplexExpr)) {
+      catalog.dropPartitionsByFilter(
+        table.identifier, specs, ignoreIfNotExists = ifExists, purge = purge)
+    } else {
+      val partitionSpec = specs.map { x =>
+        splitConjunctivePredicates(x)
+          .map(_.asInstanceOf[BinaryComparison])
+          .map(p => p.left.asInstanceOf[AttributeReference].name -> p.right.toString)
+          .toMap
+      }
+      val normalizedSpecs = partitionSpec.map { spec =>
+        PartitioningUtils.normalizePartitionSpec(
+          spec,
+          table.partitionColumnNames,
+          table.identifier.quotedString,
+          sparkSession.sessionState.conf.resolver)
+      }
+      catalog.dropPartitions(
+        table.identifier, normalizedSpecs, ignoreIfNotExists = ifExists, purge = purge)
     }
-
-    catalog.dropPartitions(
-      table.identifier, normalizedSpecs, ignoreIfNotExists = ifExists, purge = purge)
     Seq.empty[Row]
   }
 
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
@@ -424,7 +424,7 @@ case class AlterTableDropPartitionCommand(
     purge: Boolean)
   extends RunnableCommand with PredicateHelper {
 
-  private def hasNonEqualToComparison(expr: Expression): Boolean = {
+  private def isRangeComparison(expr: Expression): Boolean = {
     expr.find(e => e.isInstanceOf[BinaryComparison] && !e.isInstanceOf[EqualTo]).isDefined
   }
 
@@ -444,7 +444,7 @@ case class AlterTableDropPartitionCommand(
       }
     }
 
-    if (specs.exists(hasNonEqualToComparison)) {
+    if (specs.exists(isRangeComparison)) {
       specs.foreach { spec =>
         val partitions = catalog.listPartitionsByFilter(table.identifier, Seq(spec))
         if (partitions.nonEmpty) {
@@ -456,10 +456,10 @@ case class AlterTableDropPartitionCommand(
       }
     } else {
       val normalizedSpecs = specs.map { expr =>
-        val spec = splitConjunctivePredicates(expr)
-          .map(_.asInstanceOf[BinaryComparison])
-          .map(p => p.left.asInstanceOf[AttributeReference].name -> p.right.toString)
-          .toMap
+        val spec = splitConjunctivePredicates(expr).map {
+          case BinaryComparison(left, right) =>
+            left.asInstanceOf[AttributeReference].name -> right.toString
+        }.toMap
         PartitioningUtils.normalizePartitionSpec(
           spec,
           table.partitionColumnNames,
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
@@ -31,8 +31,8 @@ import org.apache.spark.sql.catalyst.TableIdentifier
 import org.apache.spark.sql.catalyst.analysis.Resolver
 import org.apache.spark.sql.catalyst.catalog._
 import org.apache.spark.sql.catalyst.catalog.CatalogTypes.TablePartitionSpec
-import org.apache.spark.sql.catalyst.expressions.{Attribute, AttributeReference, BinaryComparison,
-  EqualTo, Expression, PredicateHelper}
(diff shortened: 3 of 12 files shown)


sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveExternalCatalog.scala
```diff
@@ -631,6 +631,42 @@ private[spark] class HiveExternalCatalog(conf: SparkConf, hadoopConf: Configurat
     client.dropPartitions(db, table, parts.map(lowerCasePartitionSpec), ignoreIfNotExists, purge)
   }
 
+  override def dropPartitionsByFilter(
```
dongjoon-hyun: Thank you for review, @viirya and @hvanhovell .
Sure, no problem. I just thought we need to have this in `ExternalCatalog` before `Catalog Federation ( SPARK-15777 )`.
I will remove those stuff.

sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
```diff
@@ -418,27 +419,53 @@ case class AlterTableRenamePartitionCommand(
...
+      case BinaryComparison(AttributeReference(key, _, _, _), _) =>
+        table.partitionColumnNames.find(resolver(_, key)).getOrElse {
+          throw new AnalysisException(
+            s"$key is not a valid partition column in table ${table.identifier.quotedString}.")
+        }
     }
 
-    catalog.dropPartitions(
-      table.identifier, normalizedSpecs, ignoreIfNotExists = ifExists, purge = purge)
+    if (specs.exists(hasComplexExpr)) {
```
dongjoon-hyun: `listPartitionsByFilter` is not supported in `InMemoryCatalog`. So, we should use this only when it is needed.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/parser/AstBuilder.scala
```diff
@@ -204,6 +207,38 @@ class AstBuilder extends SqlBaseBaseVisitor[AnyRef] with Logging {
...
+          case SqlBaseParser.NEQ | SqlBaseParser.NEQJ =>
+            Not(EqualTo(left, right))
+          case SqlBaseParser.LT =>
+            LessThan(left, right)
+          case SqlBaseParser.LTE =>
+            LessThanOrEqual(left, right)
+          case SqlBaseParser.GT =>
+            GreaterThan(left, right)
+          case SqlBaseParser.GTE =>
+            GreaterThanOrEqual(left, right)
```
viirya: Failed to match `SqlBaseParser.NSEQ` might cause runtime error.

sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
```diff
@@ -418,27 +419,53 @@ case class AlterTableRenamePartitionCommand(
...
 case class AlterTableDropPartitionCommand(
     tableName: TableIdentifier,
-    specs: Seq[TablePartitionSpec],
+    specs: Seq[Expression],
     ifExists: Boolean,
     purge: Boolean)
-  extends RunnableCommand {
+  extends RunnableCommand with PredicateHelper {
+
+  private def hasComplexExpr(expr: Expression): Boolean = {
```
viirya: The function name looks confusing. Actually they are not more complex operators, are they?

sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
```diff
@@ -418,27 +419,53 @@ case class AlterTableRenamePartitionCommand(
...
     DDLUtils.verifyAlterTableType(catalog, table, isView = false)
     DDLUtils.verifyPartitionProviderIsHive(sparkSession, table, "ALTER TABLE DROP PARTITION")
 
-    val normalizedSpecs = specs.map { spec =>
-      PartitioningUtils.normalizePartitionSpec(
-        spec,
-        table.partitionColumnNames,
-        table.identifier.quotedString,
-        sparkSession.sessionState.conf.resolver)
+    specs.flatMap(splitConjunctivePredicates).map {
```
viirya: I think we don't need to do `splitConjunctivePredicates`. Just iterates each attribute in every spec expression's `references` and do the following resolving check, should be enough.

sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
```diff
@@ -418,27 +419,53 @@ case class AlterTableRenamePartitionCommand(
...
 
-    catalog.dropPartitions(
-      table.identifier, normalizedSpecs, ignoreIfNotExists = ifExists, purge = purge)
+    if (specs.exists(hasComplexExpr)) {
+      val partitions = catalog.listPartitionsByFilter(table.identifier, specs)
+      if (partitions.nonEmpty) {
+        catalog.dropPartitions(
+          table.identifier, partitions.map(_.spec), ignoreIfNotExists = ifExists, purge = purge)
+      } else if (!ifExists) {
+        throw new AnalysisException(specs.toString)
```
viirya: This might be not clear enough. Add a short error message?

sql/hive/src/test/scala/org/apache/spark/sql/hive/execution/HiveDDLSuite.scala
```diff
@@ -226,6 +227,63 @@ class HiveDDLSuite
...
+
+      sql("ALTER TABLE sales DROP PARTITION (quarter <= '2')")
+      checkAnswer(sql("SHOW PARTITIONS sales"),
+        Row("country=KR/quarter=3") ::
+        Row("country=KR/quarter=4") ::
+        Row("country=US/quarter=3") ::
+        Row("country=US/quarter=4") :: Nil)
+
+      sql("ALTER TABLE sales DROP PARTITION (country='KR', quarter='4')")
+      sql("ALTER TABLE sales DROP PARTITION (country='US', quarter='3')")
```
viirya: Let's add a test for dropping multiple partition specs?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/parser/AstBuilder.scala
```diff
@@ -193,6 +193,9 @@ class AstBuilder extends SqlBaseBaseVisitor[AnyRef] with Logging {
       ctx: PartitionSpecContext): Map[String, Option[String]] = withOrigin(ctx) {
     val parts = ctx.partitionVal.asScala.map { pVal =>
       val name = pVal.identifier.getText
+      val operator = Option(pVal.comparisonOperator).map(_.getText)
+      validate(operator.isEmpty || operator.get.equals("="),
+        "Only '=' partition specification is allowed.", ctx)
```
viirya: Rephrase this error message? Looks a bit confusing.

dongjoon-hyun:
Thank you for review, @viirya . I'll fix them tomorrow~

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/parser/AstBuilder.scala
dongjoon-hyun: Yep. Validation is added.

sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
dongjoon-hyun: Maybe, `hasNonEqualToComparison` is better?

sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
dongjoon-hyun: Yep. That's much better.

sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
dongjoon-hyun: Sure!

sql/hive/src/test/scala/org/apache/spark/sql/hive/execution/HiveDDLSuite.scala
dongjoon-hyun: Line 251 ?

viirya: I mean something like PARTITON (quarter <= '2'), PARTITION (quarter >= '4').

viirya:
LGTM except one comment for the test.

cc @hvanhovell for second look.

dongjoon-hyun:
Thank you so much, @viirya . I added the testcase and fixed related bug again.

sql/hive/src/test/scala/org/apache/spark/sql/hive/execution/HiveDDLSuite.scala
```diff
@@ -226,6 +227,79 @@ class HiveDDLSuite
...
+
+      val m2 = intercept[AnalysisException] {
+        sql("ALTER TABLE sales DROP PARTITION (unknown < 'KR')")
+      }.getMessage
+      assert(m2.contains("unknown is not a valid partition column in table"))
+
+      val m3 = intercept[AnalysisException] {
+        sql("ALTER TABLE sales DROP PARTITION (unknown <=> 'KR')")
+      }.getMessage
+      assert(m3.contains("'<=>' operator is not allowed in partition specification"))
```
gatorsmile: One more negative case? How about `unknown <=> upper('KR')`?

sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
```diff
@@ -418,27 +419,56 @@ case class AlterTableRenamePartitionCommand(
...
-    catalog.dropPartitions(
-      table.identifier, normalizedSpecs, ignoreIfNotExists = ifExists, purge = purge)
+    if (specs.exists(hasNonEqualToComparison)) {
+      specs.foreach { spec =>
+        val partitions = catalog.listPartitionsByFilter(table.identifier, Seq(spec))
+        if (partitions.nonEmpty) {
+          catalog.dropPartitions(
+            table.identifier, partitions.map(_.spec), ignoreIfNotExists = ifExists, purge = purge)
+        } else if (!ifExists) {
+          throw new AnalysisException(s"There is no partition for ${spec}.")
```
gatorsmile: Nit: `${spec}` -> `$spec`

dongjoon-hyun:
Thank you, @gatorsmile ! I'll fix soon.

sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
```diff
@@ -29,9 +29,10 @@ import org.apache.hadoop.mapred.{FileInputFormat, JobConf}
 import org.apache.spark.sql.{AnalysisException, Row, SparkSession}
 import org.apache.spark.sql.catalyst.TableIdentifier
 import org.apache.spark.sql.catalyst.analysis.Resolver
-import org.apache.spark.sql.catalyst.catalog.{CatalogDatabase, CatalogTable, CatalogTablePartition, CatalogTableType, SessionCatalog}
+import org.apache.spark.sql.catalyst.catalog._
 import org.apache.spark.sql.catalyst.catalog.CatalogTypes.TablePartitionSpec
-import org.apache.spark.sql.catalyst.expressions.{Attribute, AttributeReference}
+import org.apache.spark.sql.catalyst.expressions.{Attribute, AttributeReference, BinaryComparison,
+  EqualTo, Expression, PredicateHelper}
```
gatorsmile:
```Scala
import org.apache.spark.sql.catalyst.expressions.{Attribute, AttributeReference, BinaryComparison}
import org.apache.spark.sql.catalyst.expressions.{EqualTo, Expression, PredicateHelper}
```

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/parser/AstBuilder.scala
```diff
@@ -204,6 +207,40 @@ class AstBuilder extends SqlBaseBaseVisitor[AnyRef] with Logging {
...
 
   /**
+   * Create a partition filter specification.
+   */
+  def visitPartitionFilterSpec(ctx: PartitionSpecContext): Expression = withOrigin(ctx) {
+    val parts = ctx.partitionVal.asScala.map { pVal =>
+      val name = pVal.identifier.getText
+      val operator = Option(pVal.comparisonOperator).map(_.getText)
+      validate(operator.nonEmpty && !operator.get.equals("<=>"),
+        "'<=>' operator is not allowed in partition specification.", ctx)
```
dongjoon-hyun: Yep. Moved.

sql/hive/src/test/scala/org/apache/spark/sql/hive/execution/HiveDDLSuite.scala
dongjoon-hyun: The current master behavior looks the following. I added new case with the same behavior.
```
scala> sql("ALTER TABLE sales DROP PARTITION (unknown = upper('KR'))").show
extraneous input '(' expecting STRING(line 1, pos 49)
```

sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
dongjoon-hyun: Yep.

sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
```diff
@@ -418,27 +419,56 @@ case class AlterTableRenamePartitionCommand(
...
     tableName: TableIdentifier,
-    specs: Seq[TablePartitionSpec],
+    specs: Seq[Expression],
     ifExists: Boolean,
     purge: Boolean)
-  extends RunnableCommand {
+  extends RunnableCommand with PredicateHelper {
+
+  private def hasNonEqualToComparison(expr: Expression): Boolean = {
+    expr.find(e => e.isInstanceOf[BinaryComparison] && !e.isInstanceOf[EqualTo]).isDefined
```
gatorsmile: If we just read the function name, the logics should be like:
```Scala
expr.find(e => !e.isInstanceOf[EqualTo]).isDefined
```
Right?

dongjoon-hyun: There is `And` expression.

sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
dongjoon-hyun: Yep.

sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
gatorsmile: Sorry, I deleted my previous comment. I just realized it.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/parser/AstBuilder.scala
gatorsmile: Actually, this is wrong. 
```Scala
sql("ALTER TABLE sales DROP PARTITION (country < 'KR', quarter)")
```
The above statement also matches this case, right?

gatorsmile: Please add the above test case into the negative test cases. Thanks!

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/parser/AstBuilder.scala
```diff
@@ -193,6 +193,9 @@ class AstBuilder extends SqlBaseBaseVisitor[AnyRef] with Logging {
       ctx: PartitionSpecContext): Map[String, Option[String]] = withOrigin(ctx) {
     val parts = ctx.partitionVal.asScala.map { pVal =>
       val name = pVal.identifier.getText
+      val operator = Option(pVal.comparisonOperator).map(_.getText)
+      validate(operator.isEmpty || operator.get.equals("="),
+        "Only '=' operator is allowed for this partition specification.", ctx)
```
gatorsmile: Please improve the message to reflect the first case `operator.isEmpty`.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/parser/AstBuilder.scala
dongjoon-hyun: `quater` has no operator. So, it will not reach here due to `if (operator.isDefined) {`.
To make it sure, I'll add the testcase.

dongjoon-hyun:
I addressed all comments. For the upcoming comments, I'll handle them tomorrow. Thank you so much always, @gatorsmile .

sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
```diff
@@ -418,27 +419,56 @@ case class AlterTableRenamePartitionCommand(
...
+            table.identifier, partitions.map(_.spec), ignoreIfNotExists = ifExists, purge = purge)
+        } else if (!ifExists) {
+          throw new AnalysisException(s"There is no partition for $spec.")
+        }
+      }
+    } else {
+      val normalizedSpecs = specs.map { expr =>
+        val spec = splitConjunctivePredicates(expr)
+          .map(_.asInstanceOf[BinaryComparison])
+          .map(p => p.left.asInstanceOf[AttributeReference].name -> p.right.toString)
```
gatorsmile: `asInstanceOf` looks risky. It depends on the generation of `specs`

gatorsmile:
Tonight, I did not finish the review, but I have a general question about this PR:
```
ALTER TABLE table DROP [IF EXISTS] PARTITION spec1[, PARTITION spec2, ...]
```
It sounds like this PR is not handling `IF EXISTS` when users use non equalTo operators. Is this intentional? What is the behavior of Hive?

sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
```diff
@@ -418,27 +419,56 @@ case class AlterTableRenamePartitionCommand(
...
 
-    catalog.dropPartitions(
-      table.identifier, normalizedSpecs, ignoreIfNotExists = ifExists, purge = purge)
+    if (specs.exists(hasNonEqualToComparison)) {
+      specs.foreach { spec =>
+        val partitions = catalog.listPartitionsByFilter(table.identifier, Seq(spec))
+        if (partitions.nonEmpty) {
+          catalog.dropPartitions(
+            table.identifier, partitions.map(_.spec), ignoreIfNotExists = ifExists, purge = purge)
+        } else if (!ifExists) {
```
viirya: @gatorsmile Do you mean this?

gatorsmile: uh, we are handling `ifExists` here. It sounds like no test case to cover this case.

sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
```diff
@@ -418,27 +419,56 @@ case class AlterTableRenamePartitionCommand(
...
+          throw new AnalysisException(s"There is no partition for $spec.")
```
gatorsmile: Please check the Hive's error message and maybe we can improve this

viirya:
Described in hive manual: https://cwiki.apache.org/confluence/display/Hive/LanguageManual+DDL#LanguageManualDDL-DropPartitions

In Hive 0.7.0 or later, DROP returns an error if the partition doesn't exist, unless IF EXISTS is specified or the configuration variable hive.exec.drop.ignorenonexistent is set to true.

sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
viirya: https://github.com/apache/hive/blob/345353c0ea5d3ddda9f6d89cbf8cd0e92726fcb6/ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java#L2288

I think it should be "Partition or table doesn't exist."

gatorsmile:
Let us think this more carefully. In the following case, in which the partition specs have the overlapping range
```Scala
sql("ALTER TABLE sales DROP PARTITION (quarter <= 3), PARTITION (quarter <= 2)")
```

Assuming there is only one partition `quarter = 1`, will Hive issue an exception when handling the second spect `PARTITION (quarter <= 2)` after completing the first partition. Is the handling of `IF EXISTS` before the actual dropping?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/parser/AstBuilder.scala
viirya: Ah. I missed this. Move the check of "<=>" to match should avoid this...

viirya:
No. The first spec will drop the partition and the second spec will be failed.

gatorsmile:
When the partition specs have `equal to` only, our implementation first checks all before starting the actual drop. See the [codes](https://github.com/apache/spark/blob/ccb11543048dccd4cc590a8db1df1d9d5847d112/sql/hive/src/main/scala/org/apache/spark/sql/hive/client/HiveClientImpl.scala#L454-L489). 

In addition, when we hit any failure after starting to drop the partitions, we log the following error message:
```Scala
logError(
  s"""
     |======================
     |Attempt to drop the partition specs in table '$table' database '$db':
     |${specs.mkString("\n")}
     |In this attempt, the following partitions have been dropped successfully:
     |${droppedParts.mkString("\n")}
     |The remaining partitions have not been dropped:
     |${remainingParts.mkString("\n")}
     |======================
   """.stripMargin)
```

I think we should make the behaviors of DROP PARTITION consistent, no matter whether there exists any non-equal-to comparator.

viirya:
Making sense. I think we can collect all matched partitions for all given specs. And do if exists check during the collecting. And then drop the distinct partitions.

sql/catalyst/src/main/antlr4/org/apache/spark/sql/catalyst/parser/SqlBase.g4
```diff
@@ -243,7 +243,7 @@ partitionSpec
     ;
 
 partitionVal
-    : identifier (EQ constant)?
+    : identifier (comparisonOperator constant)?
```
hvanhovell: Just make this an expression and check if expressions makes sense in the `AstBuilder`.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/parser/AstBuilder.scala
```diff
@@ -204,6 +207,41 @@ class AstBuilder extends SqlBaseBaseVisitor[AnyRef] with Logging {
   }
 
   /**
+   * Create a partition filter specification.
+   */
+  def visitPartitionFilterSpec(ctx: PartitionSpecContext): Expression = withOrigin(ctx) {
```
hvanhovell: Modify this. Let the AST builder make the expression(s) and validate them here. You can also introduce the AttributeReference here. For example:
```scala
val parts = ctx.partitionVal.asScala.map { pVal =>
  expression(pVal) match {
    case cmp @ BinaryComparison(UnresolvedAttribute(name), constant: Literal) =>
      cmp.withNewChildren(Seq(AttributeReference(name, StringType)(), constant))
    case _ => throw new ParseException(..., ctx)
  }
}

```

sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
```diff
@@ -418,27 +419,56 @@ case class AlterTableRenamePartitionCommand(
...
+        }
+      }
+    } else {
+      val normalizedSpecs = specs.map { expr =>
+        val spec = splitConjunctivePredicates(expr)
+          .map(_.asInstanceOf[BinaryComparison])
```
hvanhovell: Just use a pattern match here!

sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
```diff
@@ -418,27 +419,56 @@ case class AlterTableRenamePartitionCommand(
...
     DDLUtils.verifyAlterTableType(catalog, table, isView = false)
     DDLUtils.verifyPartitionProviderIsHive(sparkSession, table, "ALTER TABLE DROP PARTITION")
 
-    val normalizedSpecs = specs.map { spec =>
-      PartitioningUtils.normalizePartitionSpec(
-        spec,
-        table.partitionColumnNames,
-        table.identifier.quotedString,
-        sparkSession.sessionState.conf.resolver)
+    specs.foreach { expr =>
```
hvanhovell: > #### rant:
> this code here to me is evidence that we should actually analyze these commands. The analyzer should throw out invalid commands, instead of doing it all during run().

dongjoon-hyun:
Thank you, @hvanhovell . I'll revise the PR.

sql/catalyst/src/main/antlr4/org/apache/spark/sql/catalyst/parser/SqlBase.g4
dongjoon-hyun: You mean that we change this as an expression and validate differently case by case.

For example,
- `ADD PARTITION` should validate `column=value`.
- `DROP PARTITION` should validate `column comparisonOperator constant` except NSEQ.

Did I understand correctly?

sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
```diff
@@ -418,27 +419,56 @@ case class AlterTableRenamePartitionCommand(
...
 case class AlterTableDropPartitionCommand(
     tableName: TableIdentifier,
-    specs: Seq[TablePartitionSpec],
+    specs: Seq[Expression],
     ifExists: Boolean,
     purge: Boolean)
-  extends RunnableCommand {
+  extends RunnableCommand with PredicateHelper {
+
+  private def hasNonEqualToComparison(expr: Expression): Boolean = {
```
hvanhovell: isRangeComparison?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/parser/AstBuilder.scala
dongjoon-hyun: Yep. I will add that. Also, add similar one to `visitPartitionSpec`, too.

sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
dongjoon-hyun: Yep.

sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
dongjoon-hyun: Perfect!

sql/catalyst/src/main/antlr4/org/apache/spark/sql/catalyst/parser/SqlBase.g4
hvanhovell: @dongjoon-hyun that is an interesting point. We might need to take a step back here. The only thing I want to prevent from happening is that we are going to recreate an expression hierarchy, and not reuse what is already there. That being said, things must remain concise. So if it takes a large amount of time or code, then we should reconsider.

In this case I would like to try the expression approach first if you don't mind.

dongjoon-hyun:
Thank you, @hvanhovell .
Now, this PR becomes much concise due to your advice.

viirya:
The changes to parsing looks good to me.

dongjoon-hyun:
Thank you, @viirya ! I also feel like that.

sql/hive/src/test/scala/org/apache/spark/sql/hive/execution/HiveDDLSuite.scala
viirya: We should add a test case like PARTITON (quarter <= '4'), PARTITION (quarter <= '2') to see what will happen since after the first partition spec is removed the second one may be failed.

dongjoon-hyun: Sure!

dongjoon-hyun: To add that, we should make another testcases because the remaining partitions are not enough to test that.

sql/hive/src/test/scala/org/apache/spark/sql/hive/execution/HiveDDLSuite.scala
```diff
@@ -225,6 +226,102 @@ class HiveDDLSuite
...
+
+      sql("ALTER TABLE sales DROP PARTITION (quarter <= 2), PARTITION (quarter >= '4')")
+      checkAnswer(sql("SHOW PARTITIONS sales"),
+        Row("country=KR/quarter=3") :: Nil)
+
+      val m = intercept[AnalysisException] {
+        sql("ALTER TABLE sales DROP PARTITION (quarter <= 4), PARTITION (quarter <= '2')")
+      }.getMessage
+      // `PARTITION (quarter <= '2')` should raises exceptions because `PARTITION (quarter <= 4)`
+      // already removes all partitions.
```
viirya: As we have discussed before, this behavior may not the same as ALTER TABLE DROP PARTITION with only equal to spec. Should we make them consistent?

dongjoon-hyun: Oh, I see your point. I removed the previous comments.

dongjoon-hyun: There are two cases, so I confused them.

1. If there is only one partition `quarter=3`, `PARTITION (quater <= '2')` should not be allowed here.
2. There is only one partition `quarter=1', `PARTITION (quarter <= 4)` will succeed and `PARTITION (quater <= '2')` will raise exception on runtime.

You mean Case 1, right?

dongjoon-hyun: To prevent Case 1, we need to run `listPartitionsByFilter` twice.

viirya: I mean case 2. Current approach will drop the first spec successfully and then the second spec will fail. I am not sure if this behavior is consistent with Hive or not.

For the case of dropping partitions with only equal to spec, the current behavior is collecting distinct matching partitions. So the overlapping specs will not be a problem.

viirya: Case 1 is also interesting. Because the first spec is valid but the second is not. Should we run spec 1 and throw exception for spec 2? Or disallow all of them?

dongjoon-hyun: For Case 1, we had better prevent query execution. I think that is consistent with equal only spec.

dongjoon-hyun: I'll update PR like the above.

dongjoon-hyun:
Thank you, @viirya . I added a prevention logic to prevent that case. Now the PR works more like Hive.

Hi, @hvanhovell .
Could you review when you have sometime?

sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
```diff
@@ -418,27 +419,66 @@ case class AlterTableRenamePartitionCommand(
...
+      }
     }
 
-    catalog.dropPartitions(
-      table.identifier, normalizedSpecs, ignoreIfNotExists = ifExists, purge = purge)
+    if (specs.exists(isRangeComparison)) {
+      if (!ifExists) {
+        // Prevent query execution if one of partition specs is invalid.
+        specs.foreach { spec =>
+          val partitions = catalog.listPartitionsByFilter(table.identifier, Seq(spec))
```
viirya: Can we not list partitions twice? It might be time consuming task. We can keep listed partitions and drop it in later block.

sql/hive/src/test/scala/org/apache/spark/sql/hive/execution/HiveDDLSuite.scala
```diff
@@ -225,6 +226,111 @@ class HiveDDLSuite
...
+        Row("country=US/quarter=4") :: Nil)
+
+      sql("ALTER TABLE sales DROP PARTITION (quarter <= 2), PARTITION (quarter >= '4')")
+      checkAnswer(sql("SHOW PARTITIONS sales"),
+        Row("country=KR/quarter=3") :: Nil)
+
+      val m = intercept[AnalysisException] {
+        sql("ALTER TABLE sales DROP PARTITION (quarter <= 4), PARTITION (quarter <= '3')")
+      }.getMessage
+      // `PARTITION (quarter <= '2')` should raises exceptions because `PARTITION (quarter <= 4)`
```
viirya: Incorrect comment: `PARTITION (quarter <= '2')` -> `PARTITION (quarter <= '3')`.

sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
dongjoon-hyun: I tried first like that, but in that case, the listed partitions have overlapping partitions. As a result, we should be updated the remaining list at every single dropping execution.

sql/hive/src/test/scala/org/apache/spark/sql/hive/execution/HiveDDLSuite.scala
dongjoon-hyun: Thanks. I fixed that.

sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
dongjoon-hyun: Hi, @viirya .
I thought about that again.
There is a third one: we can make a single set of partition at the first `listPartitionsByFilter` and remove the set by single `dropPartitions` call.
This minimizes the number of calls underlying `listPartitionsByFilter` and `dropPartitions`.

Then, the behavior will be different from Hive's sequential dropping. But, the behavior will be more natural as SQL and more robust. How do you think about that?

viirya: I think it is good. Actually the partitions dropped in the end should be the same. The difference is only if an exception is thrown or not, right?

dongjoon-hyun: Yep, correct! Thank you so much, @viirya . Then, I'll update the PR like that.

viirya:
LGTM and let's see if @hvanhovell has more comments.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSourceStrategy.scala
```diff
@@ -215,8 +215,14 @@ case class DataSourceAnalysis(conf: CatalystConf) extends Rule[LogicalPlan] {
           if (overwrite.enabled) {
             val deletedPartitions = initialMatchingPartitions.toSet -- updatedPartitions
             if (deletedPartitions.nonEmpty) {
+              import org.apache.spark.sql.catalyst.expressions._
+              val expressions = deletedPartitions.map { specs =>
+                specs.map { case (key, value) =>
+                  EqualTo(AttributeReference(key, StringType)(), Literal.create(value, StringType))
+                }.reduceLeft(org.apache.spark.sql.catalyst.expressions.And)
```
hvanhovell: just `And`?

sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
```diff
@@ -418,27 +419,58 @@ case class AlterTableRenamePartitionCommand(
...
+            s"in table ${table.identifier.quotedString}.")
+        }
+      }
     }
 
-    catalog.dropPartitions(
-      table.identifier, normalizedSpecs, ignoreIfNotExists = ifExists, purge = purge)
+    if (specs.exists(isRangeComparison)) {
+      val partitionSet = scala.collection.mutable.Set.empty[CatalogTablePartition]
+      specs.foreach { spec =>
```
hvanhovell: use `flatMap` and `distinct`?

sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
```diff
@@ -418,27 +419,58 @@ case class AlterTableRenamePartitionCommand(
...
+        } else if (!ifExists) {
+          throw new AnalysisException(s"There is no partition for ${spec.sql}")
+        }
+      }
+      catalog.dropPartitions(table.identifier, partitionSet.map(_.spec).toSeq,
+        ignoreIfNotExists = ifExists, purge = purge)
+    } else {
+      val normalizedSpecs = specs.map { expr =>
+        val spec = splitConjunctivePredicates(expr).map {
+          case BinaryComparison(left, right) =>
```
hvanhovell: Use pattern match on left?

sql/catalyst/src/main/antlr4/org/apache/spark/sql/catalyst/parser/SqlBase.g4
```diff
@@ -243,7 +243,7 @@ partitionSpec
     ;
 
 partitionVal
-    : identifier (EQ constant)?
+    : expression
```
hvanhovell: You could also remove the `partitionVal` rule

sql/hive/src/test/scala/org/apache/spark/sql/hive/execution/HiveDDLSuite.scala
```diff
@@ -225,6 +226,108 @@ class HiveDDLSuite
...
+        Row("country=KR/quarter=3") ::
+        Row("country=US/quarter=2") ::
+        Row("country=US/quarter=4") :: Nil)
+
+      sql("ALTER TABLE sales DROP PARTITION (quarter <= 2), PARTITION (quarter >= '4')")
+      checkAnswer(sql("SHOW PARTITIONS sales"),
+        Row("country=KR/quarter=3") :: Nil)
+
+      // According to the declarative partition spec definitions, this drops the union of target
+      // partitions without exceptions. Hive raises exceptions because it handle them sequentially.
```
hvanhovell: NIT: handle -> handles

hvanhovell (changes requested):
@dongjoon-hyun Looks good. I left a few minor comment. We are almost there.

dongjoon-hyun:
Thank you for review, again. I'll fix them soon.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSourceStrategy.scala
dongjoon-hyun: Yep.

sql/hive/src/test/scala/org/apache/spark/sql/hive/execution/HiveDDLSuite.scala
dongjoon-hyun: Sure.

sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
dongjoon-hyun: Thanks.

sql/catalyst/src/main/antlr4/org/apache/spark/sql/catalyst/parser/SqlBase.g4
dongjoon-hyun: It's removed now.

sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
dongjoon-hyun: Sure. Done!

hvanhovell:
LGTM - pending jenkins

dongjoon-hyun:
Thank you, @hvanhovell !

dongjoon-hyun:
Hi, @hvanhovell or @gatorsmile .
Could you merge this PR?

hvanhovell:
Merging to master. Thanks!

dongjoon-hyun:
Thank you so much, @hvanhovell , @gatorsmile, @viirya !

dongjoon-hyun:
Oh, @hvanhovell.
Can make a backport for branch-2.1 which will release this month?

hvanhovell:
@dongjoon-hyun I have cherry picked it into branch 2.1. Please note that I will revert this as soon as it causes any problems, and then will push this to 2.2.

dongjoon-hyun:
I see. I agree. Thank you so much for cherry-picking.

dongjoon-hyun:
I feel that we are really so close to 2.1. :)

hvanhovell:
I am reverting this from 2.1. See https://issues.apache.org/jira/browse/SPARK-18515 for more information.

PR #15705, opened 2016-10-31
[SPARK-18183] [SPARK-18184] Fix INSERT [INTO|OVERWRITE] TABLE ... PARTITION for Datasource tables

ericl:

There are a couple issues with the current 2.1 behavior when inserting into Datasource tables with partitions managed by Hive.

(1) OVERWRITE TABLE ... PARTITION will actually overwrite the entire table instead of just the specified partition.
(2) INSERT|OVERWRITE does not work with partitions that have custom locations.

This PR fixes both of these issues for Datasource tables managed by Hive. The behavior for legacy tables or when `manageFilesourcePartitions = false` is unchanged.

There is one other issue in that INSERT OVERWRITE with dynamic partitions will overwrite the entire table instead of just the updated partitions, but this behavior is pretty complicated to implement for Datasource tables. We should address that in a future release.

## How was this patch tested?

Unit tests.


ericl:
cc @cloud-fan @yhuai

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/basicLogicalOperators.scala
```diff
@@ -345,18 +346,32 @@ case class BroadcastHint(child: LogicalPlan) extends UnaryNode {
   override lazy val statistics: Statistics = super.statistics.copy(isBroadcastable = true)
 }
 
+/**
+ * Options for writing new data into a table.
+ *
+ * @param enabled whether to overwrite existing data in the table.
```
cloud-fan: it's pretty confusing we call it `enabled`, can we just use `overwrite`?

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/CatalogFileIndex.scala
```diff
@@ -67,7 +67,10 @@ class CatalogFileIndex(
       val selectedPartitions = sparkSession.sessionState.catalog.listPartitionsByFilter(
         table.identifier, filters)
       val partitions = selectedPartitions.map { p =>
-        PartitionPath(p.toRow(partitionSchema), p.storage.locationUri.get)
+        val path = new Path(p.storage.locationUri.get)
+        val fs = path.getFileSystem(hadoopConf)
+        PartitionPath(
+          p.toRow(partitionSchema), path.makeQualified(fs.getUri, fs.getWorkingDirectory))
```
cloud-fan: why this change? Doesn't `new Path` qualify the path string?

ericl: Apparently not. The unit test actually fails if you do that, since the path seems to be missing the `file:` prefix.

cloud-fan: But we will qualify it before writing to it at [here](https://github.com/apache/spark/pull/15705/files#diff-f9858c2d9d1a3c3e48753ef675bc865aR69), doesn't it work?

ericl: The issue is that the user can store arbitrary string paths with `ALTER TABLE PARTITIONS SET LOCATION`. Therefore, we must manually qualify the locations that come from the catalog or else they might not necessarily match up with the paths read from the filesystem.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSourceStrategy.scala
```diff
@@ -173,12 +175,22 @@ case class DataSourceAnalysis(conf: CatalystConf) extends Rule[LogicalPlan] {
...
 
-      val mode = if (overwrite) SaveMode.Overwrite else SaveMode.Append
-      if (overwrite && inputPaths.contains(outputPath)) {
+      val mode = if (overwrite.enabled) SaveMode.Overwrite else SaveMode.Append
+      if (overwrite.enabled && inputPaths.contains(outputPath)) {
         throw new AnalysisException(
           "Cannot overwrite a path that is also being read from.")
       }
 
+      val overwritePartitionPath = if (overwrite.specificPartition.isDefined &&
```
ericl: Yeah, that seems a little cleaner.

ericl:
I think you don't have to since this is just the test suite.

cloud-fan:
LGTM, merging to master!

PR #15286, opened 2016-09-28
[SPARK-17710][HOTFIX] Fix ClassCircularityError in ReplSuite tests in Maven build: use 'Class.forName' instead of 'Utils.classForName'

weiqingy:
## What changes were proposed in this pull request?

Fix ClassCircularityError in ReplSuite tests when Spark is built by Maven build.
## How was this patch tested?

(1)

```
build/mvn -DskipTests -Phadoop-2.3 -Pyarn -Phive -Phive-thriftserver -Pkinesis-asl -Pmesos clean package
```

Then test:

```
build/mvn -Dtest=none -DwildcardSuites=org.apache.spark.repl.ReplSuite test
```

ReplSuite tests passed

(2)
Manual Tests against some Spark applications in Yarn client mode and Yarn cluster mode. Need to check if spark caller contexts are written into HDFS hdfs-audit.log and Yarn RM audit log successfully.

PR #15706, opened 2016-11-01
[SPARK-18189] [SQL] Fix serialization issue in KeyValueGroupedDataset

seyfe:
Likewise [DataSet.scala](https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/Dataset.scala#L156) KeyValueGroupedDataset should mark the queryExecution as transient. 

As mentioned in the Jira ticket, without @transient we saw serialization issues like 

```
Caused by: java.io.NotSerializableException: org.apache.spark.sql.execution.QueryExecution
Serialization stack:
        - object not serializable (class: org.apache.spark.sql.execution.QueryExecution, value: ==
```

## How was this patch tested?

Run the query which is specified in the Jira ticket before and after:
```
val a = spark.createDataFrame(sc.parallelize(Seq((1,2),(3,4)))).as[(Int,Int)]
val grouped = a.groupByKey(
{x:(Int,Int)=>x._1}
)
val mappedGroups = grouped.mapGroups((k,x)=>
{(k,1)}
)
val yyy = sc.broadcast(1)
val last = mappedGroups.rdd.map(xx=>
{ val simpley = yyy.value 1 }
)
```

### DIFF
--- a/repl/scala-2.11/src/test/scala/org/apache/spark/repl/ReplSuite.scala
+++ b/repl/scala-2.11/src/test/scala/org/apache/spark/repl/ReplSuite.scala
@@ -473,4 +473,21 @@ class ReplSuite extends SparkFunSuite {
     assertDoesNotContain("AssertionError", output)
     assertDoesNotContain("Exception", output)
   }
+
+  test("SPARK-18189: Fix serialization issue in KeyValueGroupedDataset") {
+    val resultValue = 12345
+    val output = runInterpreter("local",
+      s"""
+         |val keyValueGrouped = Seq((1, 2), (3, 4)).toDS().groupByKey(_._1)
+         |val mapGroups = keyValueGrouped.mapGroups((k, v) => (k, 1))
+         |val broadcasted = sc.broadcast($resultValue)
+         |
+         |// Using broadcast triggers serialization issue in KeyValueGroupedDataset
+         |val dataset = mapGroups.map(_ => broadcasted.value)
+         |dataset.collect()
+      """.stripMargin)
+    assertDoesNotContain("error:", output)
+    assertDoesNotContain("Exception", output)
+    assertContains(s": Array[Int] = Array($resultValue, $resultValue)", output)
+  }
 }
--- a/sql/core/src/main/scala/org/apache/spark/sql/KeyValueGroupedDataset.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/KeyValueGroupedDataset.scala
@@ -40,7 +40,7 @@ import org.apache.spark.sql.expressions.ReduceAggregator
 class KeyValueGroupedDataset[K, V] private[sql](
     kEncoder: Encoder[K],
     vEncoder: Encoder[V],
-    val queryExecution: QueryExecution,
+    @transient val queryExecution: QueryExecution,
     private val dataAttributes: Seq[Attribute],
     private val groupingAttributes: Seq[Attribute]) extends Serializable {
 


hvanhovell:
Ok to test

rxin:
This guy @transient must get a lot of emails ...

rxin:
Can you change the title to "[SQL]" rather than "[core]"?

rxin:
LGTM pending Jenkins.

rxin:
Actually is there a test case you can create for this? Make sure the test case would fail when the patch is not applied.

seyfe:
Hi @rxin ,

I added a unittest in [DatasetSuite](https://github.com/apache/spark/blob/master/sql/core/src/test/scala/org/apache/spark/sql/DatasetSuite.scala) but I wasn't able to repro it. Bug only reproduce in spark-shell. That's why I moved the unittest to [ReplSuite](https://github.com/apache/spark/blob/master/repl/scala-2.11/src/test/scala/org/apache/spark/repl/ReplSuite.scala).

I checked the history and seems like we are adding unittests to only scala-2.11 now. If that's not the case, I can create same one in [ReplSuite scala-210](https://github.com/apache/spark/blob/master/repl/scala-2.10/src/test/scala/org/apache/spark/repl/ReplSuite.scala)

Please let me know.

Thanks

rxin:
Thanks! LGTM pending Jenkins.

rxin:
Merging in master/branch-2.0.
 Thanks!

srowen:
I think this might be causing the master/2.0 builds (but only Maven, weirdly) to fail. This includes the 2.0.2 RC2, which I built with Maven. See for example:

https://amplab.cs.berkeley.edu/jenkins/view/Spark%20QA%20Test%20(Dashboard)/job/spark-master-test-maven-hadoop-2.2/2200/
https://amplab.cs.berkeley.edu/jenkins/view/Spark%20QA%20Test%20(Dashboard)/job/spark-branch-2.0-test-maven-hadoop-2.7/855/

rxin:
@srowen that's weird - do you have time to dig into this a bit?

srowen:
Yeah will dig in, may not have an answer until tomorrow. If you're looking to push ahead 2.0.2 I'd just revert in branch-2.0 for now.

srowen:
@seyfe so my guess is that, not surprisingly, this is some pretty arcane interaction between several of a) codegen, b) Netty's class hierarchy, c) SBT vs Maven classloading and d) the spark-shell's classloading peculiarities.

I strongly suspect it's an issue with the test of course, not the underlying change. I wonder if we can just write a different test that doesn't trip this.

Is it necessary to test this in the REPL? can we just write the code it tests as a test directly?

seyfe:
Hi @srowen. I initially put the test in DatasetSuite but the test was passing with and without my change. Since, I was able to repro only in spark-shell I moved the test to REPL suite.

I didn't run the test in branch-2.0 only run in master. Let me check if we can fix it so, it would work on branch-2.0 as well.

seyfe:
I run the test in branch-2.0 locally and it works fine for me.
Seems like related with https://github.com/apache/spark/pull/15286. I need to dig in more.

srowen:
I think it may only involve the same exception but I am not sure it is the same. 

I don't like removing tests of course but I think I conclude that the issue you intend to fix is fixed. It is the test that has a problem. We do want to make the change but need a test that will pass. If there is no obvious workaround for what looks to me like a fairly obscure class loader issue, then we could make this test not use the REPL.

It would still test something useful and catch some potential regressions. It would not evidently catch this one though I suppose it is unlikely this exact issue would manifest again.  

I could live with a simpler test to make this work and get the change into 2.0.2 because otherwise it has to be reverted for now.

seyfe:
@srowen, that is fair. Let me move this test from Repl into DatasetSuite for branch-2.0 and send a PR.

seyfe:
Hi @srowen ,

Here is the PR:
https://github.com/apache/spark/pull/15774

PR #15696, opened 2016-10-31
[SPARK-18024][SQL] Introduce an internal commit protocol API

rxin:
## What changes were proposed in this pull request?
This patch introduces an internal commit protocol API that is used by the batch data source to do write commits. It currently has only one implementation that uses Hadoop MapReduce's OutputCommitter API. In the future, this commit API can be used to unify streaming and batch commits.

## How was this patch tested?
Should be covered by existing write tests.

PR #15707, opened 2016-11-01
[SPARK-18024][SQL] Introduce an internal commit protocol API

rxin:
This patch introduces an internal commit protocol API that is used by the batch data source to do write commits. It currently has only one implementation that uses Hadoop MapReduce's OutputCommitter API. In the future, this commit API can be used to unify streaming and batch commits.

## How was this patch tested?
Should be covered by existing write tests.


rxin:
This is the same as https://github.com/apache/spark/pull/15696

but rebased with https://github.com/apache/spark/pull/15633

ericl:
This lgtm, modulo the comments in #15696

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/WriteOutput.scala
```diff
@@ -133,7 +133,7 @@ object WriteOutput extends Logging {
               sparkAttemptNumber = taskContext.attemptNumber(),
               committer,
               iterator = iter)
-          }).flatten.distinct
+          })
```
ericl: Move the distinct to updatedPartitions?

sql/core/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala
```diff
@@ -518,6 +519,12 @@ object SQLConf {
       .booleanConf
       .createWithDefault(true)
 
+  val STREAMING_FILE_COMMIT_PROTOCOL_CLASS =
+    SQLConfigBuilder("spark.sql.streaming.commitProtocolClass")
+        .internal()
```
ericl: nit: two spaces

ericl:
This LGTM, just a minor comment

rxin:
Looks like the test failed due to a flaky test, but other than that everything else was fine. I'm going to merge this optimistically.

PR #15708, opened 2016-11-01
[SPARK-18167] [SQL] Retry when the SQLQuerySuite test flakes

ericl:

This will re-run the flaky test a few times after it fails. This will help determine if it's due to nondeterministic test setup, or because of some environment issue (e.g. leaked config from another test).

cc @yhuai


rxin:
Merging in master.

ericl:
Ok interesting, it seems the test repeatedly fails. This probably means it is a suite ordering issue.

HyukjinKwon:
Hi @ericl, I just happened to come here to track down the flaky tests. I just wonder if it'd make sense to just ignore this test maybe?

PR #15709, opened 2016-11-01
[SPARK-18190][Build][SparkR] Fix R version to not the latest in AppVeyor

HyukjinKwon:

Currently, Spark supports the test on Windows via AppVeyor but not it seems failing to download R 3.3.1 after R 3.3.2 is released. It downloads given R version after checking if that is the latest or not via http://rversions.r-pkg.org/r-release because the URL.

For example, the latest one has the URL as below:

https://cran.r-project.org/bin/windows/base/R-3.3.1-win.exe

and the old one has the URL as below.

https://cran.r-project.org/bin/windows/base/old/3.3.0/R-3.3.0-win.exe

The problem is, it seems (_according to my observation, please check the main page_[1])  the versions of R on Windows are not always synced with the latest versions. It seems R 3.3.2 for Windows does not exist.

So, it seems currently we can't download R 3.3.1 from the old repository[2] . So, currently, AppVeyor tries to find R 3.3.1 in the old repository[2] as 3.3.2 is released but it does not exist because it seems R 3.3.2 for Windows is not there and R 3.3.1 is still the latest.

Therefore, this PR proposes to lower the R version into 3.3.0 rather than rely on the latest as  SparkR supports R 3.1+ if I recall correctly.

[1]https://cran.r-project.org
[2]https://cran.r-project.org/bin/windows/base/old

## How was this patch tested?

Manually via AppVeyor CI - https://ci.appveyor.com/project/spark-test/spark/build/43-test

### DIFF
--- a/dev/appveyor-install-dependencies.ps1
+++ b/dev/appveyor-install-dependencies.ps1
@@ -113,7 +113,7 @@ $env:HADOOP_HOME = "$hadoopPath/winutils-master/hadoop-$hadoopVer"
 Pop-Location
 
 # ========================== R
-$rVer = "3.3.1"
+$rVer = "3.3.0"
 $rToolsVer = "3.4.0"
 
 InstallR


HyukjinKwon:
Build started: [SparkR] `ALL` [PR-15709](https://ci.appveyor.com/project/spark-test/spark/branch/B0651ECA-1AB2-4452-89B7-A1BF7652113A)
Diff: https://github.com/apache/spark/compare/master...spark-test:B0651ECA-1AB2-4452-89B7-A1BF7652113A

HyukjinKwon:
cc @felixcheung, @shivaram, @srowen and @wangmiao1981 (who I believe met this issue first).

HyukjinKwon:
Maybe, I might create a JIRA to describe the known issues with AppVeyor. (does it make sense?)

HyukjinKwon:
(cc @nchammas too who I also guess met this issue first. I would like to let him know.)

shivaram:
This approach is fine - The other thing we could do is just use the latest stable version as described in https://cloud.r-project.org/bin/windows/base/ - If you see the link at the bottom it says ```
Note to webmasters: A stable link which will redirect to the current Windows binary release is
https://cloud.r-project.org/bin/windows/base/release.htm
```

HyukjinKwon:
I see. That seems nice too but I hope the version is fixed if this one is also fine too..

I am worried if the version is suddenly bumped up and some tests are failed (at least, I was a bit upset when all other AppVeyor wend suddenly failed). Maybe, I can try make the script take the version `latest` which uses that link.

I remember I met a incompatibility case between 3.1.x and 3.3.x when I use `env` (it seems it is immutable in 3.3.x but not in 3.1.x) which showed a failure mark. It is good to find such cases ahead but it might be better if it does not affect other PRs.

(Not a strong opinion but I just wanted to what I thought).

shivaram:
Yeah I'm fine with fixing a version - but as you said its sometimes helpful to find if we have some problems with a newly released R version. But yeah we should have a stable download URL whichever way we choose.

HyukjinKwon:
Oh wait. Please do not merge this.

HyukjinKwon:
Oh, it seems 3.3.1 is just right now released for Windows as well. It seems fine now - https://ci.appveyor.com/project/spark-test/spark/build/45-test1122
I will close this if the tests pass fine. 

Maybe, we should take care of using the latest in the future. Thank you for quick responses @shivaram !

HyukjinKwon:
It seems fine now. Closing this.

PR #15711, opened 2016-11-01
[SPARK-18192] Support all file formats in structured streaming

rxin:
This patch adds support for all file formats in structured streaming sinks. This is actually a very small change thanks to all the previous refactoring done using the new internal commit protocol API.

## How was this patch tested?
Updated FileStreamSinkSuite to add test cases for json, text, and parquet.


marmbrus:
LGTM pending tests

rxin:
@marmbrus I ended up adding test cases since they were trivial to add (and an existing test case was asserting that text would not be supported).

rxin:
Merging in master/branch-2.1

PR #15714, opened 2016-11-01
[SPARK-18197] [CORE] Optimise AppendOnlyMap implementation

a-roberts:
This improvement works by using the fastest comparison test first and we observed a 1% throughput performance improvement on PageRank (HiBench large profile) with this change.

We used tprof and before the change in AppendOnlyMap.changeValue (where the optimisation occurs) this method was being used for 8053 profiling ticks representing 0.72% of the overall application time.

After this change we observed this method only occurring for 2786 ticks and for 0.25% of the overall time.

## How was this patch tested?
Existing unit tests and for performance we used HiBench large, profiling with tprof and IBM Healthcenter.

### DIFF
--- a/core/src/main/scala/org/apache/spark/util/collection/AppendOnlyMap.scala
+++ b/core/src/main/scala/org/apache/spark/util/collection/AppendOnlyMap.scala
@@ -140,15 +140,15 @@ class AppendOnlyMap[K, V](initialCapacity: Int = 64)
     var i = 1
     while (true) {
       val curKey = data(2 * pos)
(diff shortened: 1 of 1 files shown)


srowen (approved):
Seems fine if you observe this is a win, empirically.

rxin:
Can you put the details in the pull request description as well? It becomes part of the commit log.

srowen:
Just so I'm clear on whether this is ready -- the details in question are the ones already added to the description?

rxin:
I think he updated it already.

Merging in master/branch-2.1.

PR #15715, opened 2016-11-01
[SPARK-18198][Doc][Streaming] Highlight code snippets

lw-lin:

This patch uses `{% highlight lang %}...{% endhighlight %}` to highlight code snippets in the `Structured Streaming Kafka010 integration doc` and the `Spark Streaming Kafka010 integration doc`.

This patch consists of two commits:
- the first commit fixes only the leading spaces -- this is large
- the second commit adds the highlight instructions -- this is much simpler and easier to review

## How was this patch tested?

SKIP_API=1 jekyll build

## Screenshots

**Before**

snip20161101_3

**After**

snip20161101_1

### DIFF
--- a/docs/streaming-kafka-0-10-integration.md
+++ b/docs/streaming-kafka-0-10-integration.md
@@ -17,69 +17,69 @@ For Scala/Java applications using SBT/Maven project definitions, link your strea
 
 <div class="codetabs">
 <div data-lang="scala" markdown="1">
-	import org.apache.kafka.clients.consumer.ConsumerRecord
-	import org.apache.kafka.common.serialization.StringDeserializer
-	import org.apache.spark.streaming.kafka010._
-	import org.apache.spark.streaming.kafka010.LocationStrategies.PreferConsistent
-	import org.apache.spark.streaming.kafka010.ConsumerStrategies.Subscribe
+import org.apache.kafka.clients.consumer.ConsumerRecord
+import org.apache.kafka.common.serialization.StringDeserializer
+import org.apache.spark.streaming.kafka010._
+import org.apache.spark.streaming.kafka010.LocationStrategies.PreferConsistent
+import org.apache.spark.streaming.kafka010.ConsumerStrategies.Subscribe
 
-	val kafkaParams = Map[String, Object](
-	  "bootstrap.servers" -> "localhost:9092,anotherhost:9092",
-	  "key.deserializer" -> classOf[StringDeserializer],
-	  "value.deserializer" -> classOf[StringDeserializer],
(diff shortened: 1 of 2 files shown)


lw-lin:
Fix the leading spaces is needed, because without the fixing the code snippet would contain leading spaces as well(see the pic below), which is quite inconsistent with the other programming guides.

snip20161101_5

lw-lin:
@koeninger @srowen it'd be great if you could take a look at this too :)

srowen:
That is nicer. Are there any other code snippets across the code base that need this treatment? seems like this is always the right thing to do.

lw-lin:
@srowen thanks.

I'm afraid the streaming `flume integration doc`, `kinesis integration doc`, as well as `kafka08 integration doc` also need code highlights, but this `{% hightlight %}` syntax seem to be conflict with the indentation syntax (e.g. https://github.com/apache/spark/blame/branch-2.0/docs/streaming-flume-integration.md#L40-L73). So that's why the `flume`, `kinesis`, `kafka08` docs don't have code highlights in the first place. `structured streaming kafka010`, `spark streaming kafka010`doc, on the other hand, don't have such conflicts -- so we're adding code highlights here.

And to my acknowledgement, no other code base need this treatment.

srowen:
What about unindenting those blocks? I don't see why they have to be in the first place. They're HTML tags.

lw-lin:
snip20161101_7

Please see the screenshot above -- this is `flume doc` -- I got this by unindenting the second block. Then:

- the second block's indent does not look as pretty as the first's and the third's
- the third block's numbering just goes wrong -- it should have been `3.`

srowen:
What about keeping the indent of the div tags but un-indenting only the code? it might look weird in markdown but does it look nice in the rendered output?

If it's any more trouble than that, forget it.

lw-lin:
Good question!

Actually I had tried keeping the indent of the div tags but un-indenting only the code before I open this pr -- and it looked like this:

snip20161102_8

That was when I gave up trying more on the `flume`, `kinesis`, `kafka08` docs. :-D

srowen:
I guess we could restructure the headings to not have these numbered, indented sections to begin with, but that's a separate task I think.

lw-lin:
Sounds good!

koeninger:
LGTM

HyukjinKwon:
+1

srowen:
Merged to master

lw-lin:
@srowen thanks -- could we also merge this into branch-2.1?

srowen:
Oops, I see, we've cut branch 2.1. I must have missed that one too along with other dev@ lists going to spam. Yes I will go back and make sure a few I've just merged to master are in 2.1.

PR #15716, opened 2016-11-01
[SPARK-18201][ML] Add toDense and toSparse into Matrix trait, like Vector design

WeichenXu123:

add toDense and toSparse into Matrix trait, like Vector design
so that when we have a Matrix object `matrix: Matrix`, maybe dense or sparse,
we can call `matrix.toDense` to get DenseMatrix
and call `matrix.toSparse` to get SparseMatrix

## How was this patch tested?

N/A

### DIFF
--- a/mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala
+++ b/mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala
@@ -153,6 +153,18 @@ sealed trait Matrix extends Serializable {
    */
   @Since("2.0.0")
   def numActives: Int
+
+  /**
+    * Converts this matrix to a sparse matrix.
+    */
+  @Since("2.1.0")
+  def toSparse: SparseMatrix
+
+  /**
+    * Converts this matrix to a dense matrix.
+    */
+  @Since("2.1.0")
+  def toDense: DenseMatrix
 }
 
 /**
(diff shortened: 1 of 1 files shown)


mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala
```diff
@@ -588,6 +603,9 @@ class SparseMatrix @Since("2.0.0") (
     new DenseMatrix(numRows, numCols, toArray)
   }
 
+  @Since("2.1.0")
+  def toSparse: SparseMatrix = this
```
hhbyyh: `override` here and before `toDense`?

mllib-local/src/main/scala/org/apache/spark/ml/linalg/Matrices.scala
```diff
@@ -153,6 +153,18 @@ sealed trait Matrix extends Serializable {
    */
   @Since("2.0.0")
   def numActives: Int
+
+  /**
+    * Converts this matrix to a sparse matrix.
+    */
```
hhbyyh: comment style

sethah:
These changes are all covered in https://github.com/apache/spark/pull/15628. Would you mind helping to review that instead?

WeichenXu123:
@sethah All right, I will take a look at it later, thanks~

PR #15717, opened 2016-11-01
[SPARK-17910][SQL] Allow users to update the comment of a column

jiangxb1987:

Right now, once a user set the comment of a column with create table command, he/she cannot update the comment. It will be useful to provide a public interface (e.g. SQL) to do that.

This PR implements the following SQL statement:
```
ALTER TABLE table [PARTITION partition_spec]
CHANGE [COLUMN] column_old_name column_new_name column_dataType
[COMMENT column_comment]
[FIRST | AFTER column_name];
```

For further expansion, we could support alter `name`/`dataType`/`index` of a column too.

## How was this patch tested?

Add new test cases in `ExternalCatalogSuite` and `SessionCatalogSuite`.
Add sql file test for `ALTER TABLE CHANGE COLUMN` statement.


sql/catalyst/src/main/scala/org/apache/spark/sql/types/StructField.scala
```diff
@@ -74,4 +74,11 @@ case class StructField(
   def getComment(): Option[String] = {
     if (metadata.contains("comment")) Option(metadata.getString("comment")) else None
   }
+
+  /**
+   * Return the full description of the StructField.
+   */
+  def getDesc(): String = {
```
rxin: is this really needed?

rxin:
Looks like there is an actual test failure.

sql/catalyst/src/main/scala/org/apache/spark/sql/types/StructField.scala
jiangxb1987: Yes because we have to print the full description of the StructField, but `StructField.toString` don't output `metadata` field.

sql/catalyst/src/main/antlr4/org/apache/spark/sql/catalyst/parser/SqlBase.g4
```diff
@@ -93,6 +93,8 @@ statement
         SET TBLPROPERTIES tablePropertyList                            #setTableProperties
     | ALTER (TABLE | VIEW) tableIdentifier
         UNSET TBLPROPERTIES (IF EXISTS)? tablePropertyList             #unsetTableProperties
+    | ALTER (TABLE | VIEW) tableIdentifier
+        CHANGE COLUMN? expandColTypeList                               #changeColumns
```
jiangxb1987: Don't support change partitions for now, will implement it in another PR.

sql/core/src/test/resources/sql-tests/inputs/change-column.sql
```diff
@@ -0,0 +1,50 @@
+-- Create the origin table
+CREATE TABLE test_change(a Int, b String, c Int);
+CREATE VIEW test_view(a, b, c) AS
+SELECT * FROM VALUES (1, "one", 11), (null, "two", 22) AS testData(a, b, c);
```
jiangxb1987: We have to make the column `a` nullable here, otherwise it can't be coverted to `StructField(a,IntegerType,true,{comment: newComment})`

jiangxb1987 (commented):
This is ready for review now. Perhaps we could easily add full support for HIVE-style `ALTER TABLE CHANGE COLUMN` statement after this PR have been merged.

jiangxb1987:
retest this please. - This build error is strange because it successfully built on my local envirement.

gatorsmile:
Any update on this PR?

jiangxb1987:
@gatorsmile I have been busy on other stuffs recently so no progress has been made on this PR, I will try to find some time to make this PR pass all unit cases this weekend. Thank you for looking at this!

sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
```diff
@@ -273,6 +273,68 @@ case class AlterTableUnsetPropertiesCommand(
 
 }
 
+
+/**
+ * A command to change the columns for a table, only support change column comment for now.
```
gatorsmile: Nit: `support changing column comments`

sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
```diff
@@ -273,6 +273,68 @@ case class AlterTableUnsetPropertiesCommand(
...
+ * This function creates a [[AlterTableChangeColumnsCommand]] logical plan.
```
gatorsmile: Remove this line.

sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
```diff
@@ -273,6 +273,68 @@ case class AlterTableUnsetPropertiesCommand(
...
+ *
+ * The syntax of using this command in SQL is:
+ * {{{
+ *   ALTER (TABLE | VIEW) table_identifier
+ *   CHANGE (COLUMN) column_name column_name column_dataType column_comment
+ *   (FIRST | AFTER column_name);
```
gatorsmile: Above is not clear. Please check how Hive documents it. 
```
ALTER TABLE table_name [PARTITION partition_spec] CHANGE [COLUMN] col_old_name col_new_name column_type
  [COMMENT col_comment] [FIRST|AFTER column_name] [CASCADE|RESTRICT];
```

gatorsmile:
A basic question. Like Hive, we should not support `ALTER VIEW CHANGE COLUMN`, right?

jiangxb1987:
@gatorsmile Thank you for clarify, will review the `ALTER TABLE CHANGE COLUMN` statement today. BTW do you think we should add `[CASCADE|RESTRICT]` in this PR?

jiangxb1987:
@gatorsmile Would you please have a look at this when you have time? Thank you!

gatorsmile:
I prefer to doing `[CASCADE|RESTRICT]` in a separate PR. However, we still need to add a test case to verify whether the behavior is following the default `RESTRICT`.

gatorsmile:
Will review this PR tonight. Thanks!

gatorsmile:
Another question about this PR: does it support data source tables?

jiangxb1987:
It should support data source tables, acturally perhaps I should add test cases in `DDLSuite` for that - yes I'll add that this evening. Thank you!

jiangxb1987:
@gatorsmile I've added tests ensure that it support data source tables, please check when you have time, thank you!

gatorsmile:
Will review this PR tomorrow.

gatorsmile:
How about the partition columns of partitioned Hive serde tables?

sql/core/src/test/scala/org/apache/spark/sql/execution/command/DDLCommandSuite.scala
```diff
@@ -694,19 +694,6 @@ class DDLCommandSuite extends PlanTest {
     assertUnsupported("ALTER TABLE table_name SKEWED BY (key) ON (1,5,6) STORED AS DIRECTORIES")
   }
 
-  test("alter table: change column name/type/position/comment (not allowed)") {
```
gatorsmile: The test suite `DDLCommandSuite` is used for testing the parser. Could you added unit test cases for the new parser rules?

sql/catalyst/src/main/antlr4/org/apache/spark/sql/catalyst/parser/SqlBase.g4
```diff
@@ -93,6 +93,8 @@ statement
         SET TBLPROPERTIES tablePropertyList                            #setTableProperties
     | ALTER (TABLE | VIEW) tableIdentifier
         UNSET TBLPROPERTIES (IF EXISTS)? tablePropertyList             #unsetTableProperties
+    | ALTER TABLE tableIdentifier
+        CHANGE COLUMN? expandColTypeList                               #changeColumns
```
gatorsmile: Here, you do not have `col_old_name` and `col_new_name`.

Please follow Hive DDL syntax. 
```
ALTER TABLE table_name [PARTITION partition_spec] CHANGE [COLUMN] col_old_name col_new_name column_type
  [COMMENT col_comment] [FIRST|AFTER column_name] [CASCADE|RESTRICT];
```

gatorsmile:
Let us first resolve the paser/syntax-related issues.

sql/catalyst/src/main/antlr4/org/apache/spark/sql/catalyst/parser/SqlBase.g4
jiangxb1987: I think we are following the HIVE DDL syntax by the following definition:
```
colType
    : identifier dataType (COMMENT STRING)?
    ;

expandColTypeList
    : expandColType (',' expandColType)*
    ;

expandColType
    : identifier colType colPosition?
    ;
```

sql/core/src/test/scala/org/apache/spark/sql/execution/command/DDLCommandSuite.scala
jiangxb1987: added.

sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
```diff
@@ -273,6 +273,59 @@ case class AlterTableUnsetPropertiesCommand(
...
+    val table = catalog.getTableMetadata(tableName)
+    DDLUtils.verifyAlterTableType(catalog, table, false)
+    // Assert we only alter the column's comment, other fields e.g. name/dataType remain
+    // unchanged.
+    columns.foreach { pair =>
+      val originColumn = table.schema.collectFirst {
+        case field if field.name == pair._1 => field
+      }
+      val newColumn = pair._2
+      val unchanged = originColumn.map(equalIgnoreComment(_, newColumn)).getOrElse(true)
```
gatorsmile:
```Scala
val unchanged = originColumn.forall(equalIgnoreComment(_, newColumn))
```

sql/catalyst/src/main/scala/org/apache/spark/sql/types/StructField.scala
gatorsmile: `case class StructField` is external and marked as stable.

`getDesc ` is only used in the error message. We should avoid adding this.

sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
```diff
@@ -273,6 +273,59 @@ case class AlterTableUnsetPropertiesCommand(
...
+      if (!unchanged) {
+        throw new AnalysisException(
+          s"ALTER TABLE CHANGE COLUMN don't support change column " +
+            s"'${originColumn.get.getDesc}' to '${newColumn.getDesc}'")
```
gatorsmile: Please update this without using `getDesc`

sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
```diff
@@ -273,6 +273,59 @@ case class AlterTableUnsetPropertiesCommand(
...
+ */
+case class AlterTableChangeColumnsCommand(
+    tableName: TableIdentifier,
+    columns: Map[String, StructField]) extends RunnableCommand {
+
+  // TODO: support change column name/dataType/metadata/position.
+  override def run(sparkSession: SparkSession): Seq[Row] = {
+    val catalog = sparkSession.sessionState.catalog
+    val table = catalog.getTableMetadata(tableName)
+    DDLUtils.verifyAlterTableType(catalog, table, false)
```
gatorsmile:
```Scala
DDLUtils.verifyAlterTableType(catalog, table, isView = false)
```

gatorsmile:
What happened if no any change is made? What is the behavior?

gatorsmile:
@yhuai @rxin Are we still targeting this to 2.1?

sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
```diff
@@ -273,6 +273,59 @@ case class AlterTableUnsetPropertiesCommand(
...
+    // Assert we only alter the column's comment, other fields e.g. name/dataType remain
+    // unchanged.
+    columns.foreach { pair =>
```
gatorsmile: If you change `pair` to `case (oldColumnName, newColumn)`, the following code can be cleaned.

gatorsmile:
@yhuai @rxin Are we still targeting this to 2.1?

rxin:
I would not.

gatorsmile:
@rxin I see. 

@jiangxb1987 Could we revisit this PR after we release Spark 2.1? Now, the major focus is on the bug fixing.

jiangxb1987:
@gatorsmile Sure I'm fine with that. Thanks!

gatorsmile:
Thanks!

jiangxb1987:
retest this please.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/ExternalCatalog.scala
```diff
@@ -217,6 +218,29 @@ abstract class ExternalCatalog {
...
+  /**
+   * Alter one or more column(s)'s comment.
+   *
+   * Note: If the underlying implementation does not support altering a certain field,
+   * this becomes a no-op.
+   * @param db database name
+   * @param table table name
+   * @param columnComments map from column names to comments
+   */
+  def alterColumnComments(db: String, table: String, columnComments: Map[String, String]): Unit
```
cloud-fan: can we just reuse `alterTable`?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/ExternalCatalog.scala
```diff
@@ -217,6 +218,29 @@ abstract class ExternalCatalog {
...
+
+  /**
+   * List the metadata of all columns that are defined in the specified table schema, assuming it
+   * exists.
+   * @param db database name
+   * @param table table name
+   */
+  def listColumns(db: String, table: String): Seq[StructField]
```
cloud-fan: does it worth to add a new API in `ExternalCatalog`?

cloud-fan:
I don't think we will support changing column type/name/position/metadata in the near future, shall we simplify the syntax for changing column comment? cc @yhuai @hvanhovell

jiangxb1987:
@cloud-fan Could you explain more about the reason why we are not planning to support changing columns? It seems that implementing these statements are not very hard if we decide to follow the HIVE way(that only modifies the schema but not the data). And I believe these are really useful. Thanks!

cloud-fan:
hmmm, Hive only change the metadata not data? When will users need to do so?

jiangxb1987:
> The column change command will only modify Hive's metadata, and will not modify data. Users should make sure the actual data layout of the table/partition conforms with the metadata definition.

Hive requires the user to ensure that, after the column change, the metadata is still valid on current data set. This is useful in certain conditions, e.g. A column may be created with its dataType as `Int` but some times later the column number grows bigger and should be changed to a Long column.

gatorsmile:
https://issues.apache.org/jira/browse/SPARK-18727 is a related JIRA. If we allow users to manually change the schema, do we need to follow what Hive does now?  ALTER TABLE ADD/REPLACE/CHANGE COLUMN? 

BTW, to support these DDL, it might also impact the new view management.

rxin:
Those are super useful to add in the future.

cloud-fan:
ok then the parser rule looks good to me, my only concern is the new APIs in `ExternalCatalog`, I don't think they are necessary, @jiangxb1987 what's the motivation you added them?

jiangxb1987:
@cloud-fan @gatorsmile Thank you for comments! Perhaps we should focus on adding the `ALTER TABLE CHANGE COLUMN` statement, and only support change column comments in this PR. The remaining issues could be addressed in other followup PRs in the future.
The view management is not fully designed now, so it's a bit hard to have detailed discussion on that topic, but I'm sure it will be a really important topic.
The functions `listColumns` could be removed, since it's only used in catalog related test cases, we may move the function to `CatalogTestUtils`.

cloud-fan:
Why can't we remove `alterTableComment`? It seems that the implementations are same, so we can make it a util function to map alter column comment action to `alterTable` call.

sql/core/src/main/scala/org/apache/spark/sql/execution/SparkSqlParser.scala
```diff
@@ -877,6 +877,35 @@ class SparkSqlAstBuilder(conf: SQLConf) extends AstBuilder {
...
+      operationNotAllowed(s"ALTER TABLE PARTITION partition_spec CHANGE COLUMN, can not change " +
+        "the columns in partition spec", ctx)
+    }
+
+    val tableName = visitTableIdentifier(ctx.tableIdentifier)
+
+    val columns = ctx.expandColTypeList.expandColType.asScala.map { col =>
+      if (col.colPosition != null) {
+        operationNotAllowed(
+          s"ALTER TABLE [PARTITION] CHANGE COLUMN ... FIRST | AFTER otherCol", ctx)
```
gatorsmile: Nit: remove the string interpolation

sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
```diff
@@ -274,6 +274,63 @@ case class AlterTableUnsetPropertiesCommand(
...
+    // Currently only support changing column comment, throw a Exception if other fields e.g.
+    // name/dataType are changed.
+    columns.foreach { case (oldColumnName, newColumn) =>
+      val originColumn = table.schema.collectFirst {
+        case field if field.name == oldColumnName => field
+      }
+      val unchanged = originColumn.forall(equalIgnoreComment(_, newColumn))
+      if (!unchanged) {
+        throw new AnalysisException(
+          s"ALTER TABLE CHANGE COLUMN is not supported for changing column " +
```
gatorsmile: Nit: remove the string interpolation

sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
```diff
@@ -274,6 +274,63 @@ case class AlterTableUnsetPropertiesCommand(
...
+case class AlterTableChangeColumnsCommand(
+    tableName: TableIdentifier,
+    columns: Map[String, StructField]) extends RunnableCommand {
+
+  // TODO: support change column name/dataType/metadata/position.
+  override def run(sparkSession: SparkSession): Seq[Row] = {
+    val catalog = sparkSession.sessionState.catalog
+    val table = catalog.getTableMetadata(tableName)
+    DDLUtils.verifyAlterTableType(catalog, table, isView = false)
+    // Currently only support changing column comment, throw a Exception if other fields e.g.
```
gatorsmile: `a Exception` -> `an exception`

gatorsmile: `changing column comment` -> `changing the column comment`

sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
```diff
@@ -274,6 +274,63 @@ case class AlterTableUnsetPropertiesCommand(
...
+        throw new AnalysisException(
+          s"ALTER TABLE CHANGE COLUMN is not supported for changing column " +
+            s"'${getDesc(originColumn.get)}' to '${getDesc(newColumn)}'")
+      }
+    }
+
+    val newSchema = table.schema.fields.map { field =>
+      // If `columns` contains field name, update the field to the new column, else respect the
+      // field.
+      columns.get(field.name).getOrElse(field)
```
gatorsmile: This way will lose the `metadata` field of the column. Let me show you the way how we add metadata to the column
```Scala
      val df = Seq((1, "abc"), (2, "hello")).toDF("a", "b")
      val md = new MetadataBuilder().putString("key", "value").build()
      val dfWithmeta = df.select('a, 'b.as("b", md))
      dfWithmeta.write.saveAsTable("test1")
```

sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
```diff
@@ -274,6 +274,80 @@ case class AlterTableUnsetPropertiesCommand(
...
+  // Compare a [[StructField]] to another, return true if they have the same column
+  // name(by resolver) and dataType.
+  private def columnEqual(
+      field: StructField, other: StructField, resolver: Resolver): Boolean = {
+    resolver(field.name, other.name) && field.dataType == other.dataType
+  }
+
+  // Genereate the full description of a StructField.
+  private def getDesc(field: StructField): String = {
+    s"StructField(${field.name},${field.dataType},${field.nullable},${field.metadata})"
```
cloud-fan: it looks weird to mention `StructField` in a `CHANGE COLUMN` command

jiangxb1987: How should we change the description then?

sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
```diff
@@ -274,6 +274,80 @@ case class AlterTableUnsetPropertiesCommand(
...
+  // Find the origin column from schema by column name, throw a Exception if the column
+  // reference is invalid.
+  private def findColumn(schema: StructType, name: String, resolver: Resolver): StructField = {
+    schema.fields.collectFirst {
+      case field if resolver(field.name, name) => field
+    }.getOrElse(throw new AnalysisException(
+      s"Invalid column reference '$name', table schema is '${schema}'"))
+  }
+
+  // Add the comment to a column, if comment is empty, return the original column.
```
cloud-fan: if the comment is empty, shall we remove the existing comment? what's the behavior of hive?

jiangxb1987: Hive won't change the original comment if the COMMENT statement is empty. We keep the same with this behavior.

sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
```diff
@@ -274,6 +274,80 @@ case class AlterTableUnsetPropertiesCommand(
...
+          "ALTER TABLE CHANGE COLUMN is not supported for changing column " +
+            s"'${getDesc(originColumn)}' to '${getDesc(newField)}'")
+      }
+      // Create a new column from the origin column with new comment.
+      val newColumn = addComment(originColumn, newField.getComment)
+      // Create the map from origin column to changed column
+      originColumn -> newColumn
+    }
+
+    val newSchema = table.schema.fields.map(field => columnsMap.getOrElse(field, field))
```
cloud-fan: how about:
```
val newSchema = table.schema.fields.map { field =>
  val newColumn = columns.find { case (oldName, _) => resolver(oldName, field.name) }
  newColumn.map { case (_, newField) =>
    if (!columnEqual(field, newField, resolver)) {
      ......
    }
    addComment(field, newField.getComment)
  }.getOrElse(field)
}
```

sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
cloud-fan: as we only check name and data type, how about `ALTER TABLE CHANGE COLUMN is not supported for changing column abc with type int to xyz with type long`?

sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
```diff
@@ -274,6 +274,80 @@ case class AlterTableUnsetPropertiesCommand(
...
+      // Throw a Exception if the column name/dataType is changed.
+      if (!columnEqual(originColumn, newField, resolver)) {
+        throw new AnalysisException(
+          "ALTER TABLE CHANGE COLUMN is not supported for changing column " +
+            s"'${getDesc(originColumn)}' to '${getDesc(newField)}'")
+      }
+      // Create a new column from the origin column with new comment.
+      val newColumn = addComment(originColumn, newField.getComment)
+      // Create the map from origin column to changed column
+      originColumn -> newColumn
```
cloud-fan: how about use `originColumn.name` as map key?

jiangxb1987: +1

sql/core/src/test/resources/sql-tests/inputs/change-column.sql
```diff
@@ -0,0 +1,53 @@
+-- Create the origin table
+CREATE TABLE test_change(a Int, b String, c Int);
```
cloud-fan: nit: in SQL statement, we should upper case the type string, i.e. `STRING`, `INT`, etc.

hvanhovell:
retest this please

sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
```diff
@@ -274,6 +274,76 @@ case class AlterTableUnsetPropertiesCommand(
...
+    val catalog = sparkSession.sessionState.catalog
+    val table = catalog.getTableMetadata(tableName)
+    val resolver = sparkSession.sessionState.conf.resolver
+    DDLUtils.verifyAlterTableType(catalog, table, isView = false)
+
+    // Create a map that converts the origin column to the new column with changed comment, throw
+    // a Exception if the column reference is invalid or the column name/dataType is changed.
+    val columnsMap = columns.map { case (oldName: String, newField: StructField) =>
+      // Find the origin column from schema by column name.
+      val originColumn = findColumn(table.schema, oldName, resolver)
```
gatorsmile: How about renaming `findColumn` to `findColumnByName`?

gatorsmile:
Could you please add a test case for verifying the metadata field of the column will not be lost after adding a comment? Thanks!

sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
```diff
@@ -274,6 +274,76 @@ case class AlterTableUnsetPropertiesCommand(
...
+      // Throw a Exception if the column name/dataType is changed.
```
gatorsmile: Nit: please correct `a Exception` to `an exception` in all the comments you added. Thanks!

sql/core/src/test/resources/sql-tests/inputs/change-column.sql
```diff
@@ -0,0 +1,53 @@
...
+
+-- Change a invalid column
+ALTER TABLE test_change CHANGE invalid_col invalid_col INT;
+DESC test_change;
+
+-- Change column name/dataType/position/comment together (not supported yet)
+ALTER TABLE test_change CHANGE a a1 STRING COMMENT 'this is column a1' AFTER b;
+DESC test_change;
+
+-- Case sensitive
```
gatorsmile: Please update the comment. This test case is to check the behavior when `spark.sql.caseSensitive` is off.

sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
jiangxb1987: I understand that `a` should be changed to `an`, but should we also change `Exception` to `exception`? Thanks!

gatorsmile: It is not very common to use `Exception` here. Or you can change it to -> `an AnalysisException`.

gatorsmile:
Could you please also add a test case when `spark.sql.caseSensitive` is on?

sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
jiangxb1987: @gatorsmile That's great! Thank you!

jiangxb1987:
@gatorsmile Thank you for your suggestions! I'll update the code this evening!

sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
gatorsmile: To drop a comment, I think users can do something like
```SQL
ALTER TABLE test_change CHANGE a A INT COMMENT ''
```

sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
```diff
@@ -274,6 +274,76 @@ case class AlterTableUnsetPropertiesCommand(
...
+      if (!columnEqual(originColumn, newField, resolver)) {
+        throw new AnalysisException(
+          "ALTER TABLE CHANGE COLUMN is not supported for changing column " +
+            s"'${originColumn.name}' with type '${originColumn.dataType}' to " +
+            s"'${newField.name}' with type '${newField.dataType}'")
+      }
+      // Create a new column from the origin column with new comment.
```
gatorsmile: Nit: `new comment` -> `the new comment`

jiangxb1987: Sorry for my poor English... :'(

cloud-fan:
LGTM, @gatorsmile for final sign-off

sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
```diff
@@ -274,6 +274,78 @@ case class AlterTableUnsetPropertiesCommand(
...
+
+/**
+ * A command to change the columns for a table, only support changing the comments of non-partition
+ * columns for now.
+ *
+ * The syntax of using this command in SQL is:
+ * {{{
+ *   ALTER TABLE table_identifier
+ *   CHANGE [COLUMN] column_old_name column_new_name column_dataType [COMMENT column_comment]
+ *   [FIRST | AFTER column_name];
```
gatorsmile: This is the right Hive syntax, but `SparkSqlParser.scala` is using the different one.

sql/core/src/main/scala/org/apache/spark/sql/execution/SparkSqlParser.scala
```diff
@@ -877,6 +877,35 @@ class SparkSqlAstBuilder(conf: SQLConf) extends AstBuilder {
   }
 
   /**
+   * Create a [[AlterTableChangeColumnsCommand]] command.
+   *
+   * For example:
+   * {{{
+   *   ALTER TABLE table [PARTITION partition_spec]
+   *   CHANGE [COLUMN] `col` `col` dataType [COMMENT "comment"] [FIRST | AFTER `otherCol`]
+   *   [, `col2` `col2` dataType [COMMENT "comment"] [FIRST | AFTER `otherCol`], ...]
```
gatorsmile: What is the reason do we allow users to modify multiple columns in the same DDL? This is different from what Hive supports. Should we do it? cc @cloud-fan 

In addition, based on the existing syntax, if we really want to support multiple columns, we should change the keywords to `CHANGE COLUMNS`.

gatorsmile: When we supporting multiple columns syntax, we also need to consider the edge case. For example, duplicate column names.

gatorsmile:
Sorry for the last minute comment. I did not realize it until I manually run these test cases in Hive.

sql/core/src/main/scala/org/apache/spark/sql/execution/SparkSqlParser.scala
cloud-fan: let's follow hive's syntax

jiangxb1987:
@gatorsmile @cloud-fan In fact that's mainly my fault, I should have checked that the syntax is the same as that in HIVE. I'm working on this now. Thank you!

gatorsmile:
LGTM

gatorsmile:
@jiangxb1987 Could you update the PR description to reflect the new syntax? Thanks!

jiangxb1987:
@gatorsmile I've updated the PR description, thanks!

rxin:
Merging in master.

skliarpawlo:
@jiangxb1987 any ideas why this still doesn't work for me in spark 2.2.0?
Sample code:
```
spark.sql("""
    alter table test_table
    change column metric metric string comment "metic doc"
""")
```
The same sql query executed from hive works as expected, but from spark it has no effect

Thanks

jiangxb1987:
Can you also show the result of DESC TABLE before and after the command?

jiangxb1987:
Please file a JIRA issue if you think there is unexpected behavior.

skliarpawlo:
@jiangxb1987 thanks for so quick response! I filed the issue:
https://issues.apache.org/jira/browse/SPARK-23525

PR #15720, opened 2016-11-01
[SPARK-18167] Disable flaky SQLQuerySuite test

ericl:
We now know it's a persistent environmental issue that is causing this test to sometimes fail. One hypothesis is that some configuration is leaked from another suite, and depending on suite ordering this can cause this test to fail.

I am planning on mining the jenkins logs to try to narrow down which suite could be causing this. For now, disable the test.


rxin:
Merging in master.

PR #15721, opened 2016-11-01
[SPARK-17772][ML][TEST] Add test functions for ML sample weights

sethah:

More and more ML algos are accepting sample weights, and they have been tested rather heterogeneously and with code duplication. This patch adds extensible helper methods to `MLTestingUtils` that can be reused by various algorithms accepting sample weights. Up to now, there seems to be a few tests that have been implemented commonly:

* Check that oversampling is the same as giving the instances sample weights proportional to the number of samples
* Check that outliers with tiny sample weights do not affect the algorithm's performance

This patch adds an additional test:

* Check that algorithms are invariant to constant scaling of the sample weights. i.e. uniform sample weights with `w_i = 1.0` is effectively the same as uniform sample weights with `w_i = 10000` or `w_i = 0.0001`

The instances of these tests occurred in LinearRegression, NaiveBayes, and LogisticRegression. Those tests have been removed/modified to use the new helper methods. These helper functions will be of use when [SPARK-9478](https://issues.apache.org/jira/browse/SPARK-9478) is implemented. 

## How was this patch tested?

This patch only involves modifying test suites.

## Other notes

Both IsotonicRegression and GeneralizedLinearRegression also extend `HasWeightCol`. I did not modify these test suites because it will make this patch easier to review, and because they did not duplicate the same tests as the three suites that were modified. If we want to change them later, we can create a JIRA for it now, but it's open for debate.

### DIFF
--- a/mllib/src/test/scala/org/apache/spark/ml/classification/LogisticRegressionSuite.scala
+++ b/mllib/src/test/scala/org/apache/spark/ml/classification/LogisticRegressionSuite.scala
@@ -1811,27 +1811,21 @@ class LogisticRegressionSuite
   }
 
   test("logistic regression with sample weights") {
-    val sqlContext = spark.sqlContext
-    import sqlContext.implicits._
-    val numFeatures = 5
-    val numPoints = 50
     def modelEquals(m1: LogisticRegressionModel, m2: LogisticRegressionModel): Unit = {
       assert(m1.coefficientMatrix ~== m2.coefficientMatrix absTol 0.01)
     }
     val testParams = Seq(
       ("binomial", smallBinaryDataset, 2),
-      ("multinomial", smallMultinomialDataset, 4)
+      ("multinomial", smallMultinomialDataset, 3)
     )
     testParams.foreach { case (family, dataset, numClasses) =>
       val estimator = new LogisticRegression().setFamily(family)
       MLTestingUtils.testArbitrarilyScaledWeights[LogisticRegressionModel, LogisticRegression](
         dataset.as[LabeledPoint], estimator, modelEquals)
       MLTestingUtils.testOutliersWithSmallWeights[LogisticRegressionModel, LogisticRegression](
-        spark, estimator, Map.empty[Int, Int], numPoints, numClasses, numFeatures, modelEquals,
-        seed)
-      MLTestingUtils.testOversamplingVsWeighting[LogisticRegressionModel, LogisticRegression](spark,
-        estimator, Map.empty[Int, Int], numPoints, numClasses, numFeatures,
-        modelEquals, seed)
+        dataset.withColumn("weight", lit(1.0)).as[Instance], estimator, numClasses, modelEquals)
+      MLTestingUtils.testOversamplingVsWeighting[LogisticRegressionModel, LogisticRegression](
+        dataset.toDF(), estimator, modelEquals, seed)
     }
   }
 
--- a/mllib/src/test/scala/org/apache/spark/ml/classification/LogisticRegressionSuite.scala
+++ b/mllib/src/test/scala/org/apache/spark/ml/classification/LogisticRegressionSuite.scala
@@ -1812,7 +1812,8 @@ class LogisticRegressionSuite
 
   test("logistic regression with sample weights") {
     def modelEquals(m1: LogisticRegressionModel, m2: LogisticRegressionModel): Unit = {
-      assert(m1.coefficientMatrix ~== m2.coefficientMatrix absTol 0.01)
+      assert(m1.coefficientMatrix ~== m2.coefficientMatrix absTol 0.05)
+      assert(m1.interceptVector ~== m2.interceptVector absTol 0.05)
     }
     val testParams = Seq(
       ("binomial", smallBinaryDataset, 2),
@@ -1823,9 +1824,9 @@ class LogisticRegressionSuite
       MLTestingUtils.testArbitrarilyScaledWeights[LogisticRegressionModel, LogisticRegression](
         dataset.as[LabeledPoint], estimator, modelEquals)
       MLTestingUtils.testOutliersWithSmallWeights[LogisticRegressionModel, LogisticRegression](
-        dataset.withColumn("weight", lit(1.0)).as[Instance], estimator, numClasses, modelEquals)
+        dataset.as[LabeledPoint], estimator, numClasses, modelEquals)
       MLTestingUtils.testOversamplingVsWeighting[LogisticRegressionModel, LogisticRegression](
-        dataset.toDF(), estimator, modelEquals, seed)
+        dataset.as[LabeledPoint], estimator, modelEquals, seed)
     }
   }
 
--- a/mllib/src/test/scala/org/apache/spark/ml/classification/LogisticRegressionSuite.scala
+++ b/mllib/src/test/scala/org/apache/spark/ml/classification/LogisticRegressionSuite.scala
@@ -1810,52 +1810,29 @@ class LogisticRegressionSuite
 
   }
 
-  test("binary logistic regression with weighted data") {
-    val numClasses = 2
-    val numPoints = 40
-    val outlierData = MLTestingUtils.genClassificationInstancesWithWeightedOutliers(spark,
-      numClasses, numPoints)
-    val testData = Array.tabulate[LabeledPoint](numClasses) { i =>
-      LabeledPoint(i.toDouble, Vectors.dense(i.toDouble))
-    }.toSeq.toDF()
-    val lr = new LogisticRegression().setFamily("binomial").setWeightCol("weight")
-    val model = lr.fit(outlierData)
-    val results = model.transform(testData).select("label", "prediction").collect()
-
-    // check that the predictions are the one to one mapping
-    results.foreach { case Row(label: Double, pred: Double) =>
-      assert(label === pred)
+  test("logistic regression with sample weights") {
+    val sqlContext = spark.sqlContext
+    import sqlContext.implicits._
+    val numFeatures = 5
+    val numPoints = 50
+    def modelEquals(m1: LogisticRegressionModel, m2: LogisticRegressionModel): Unit = {
+      assert(m1.coefficientMatrix ~== m2.coefficientMatrix absTol 0.01)
     }
-    val (overSampledData, weightedData) =
-      MLTestingUtils.genEquivalentOversampledAndWeightedInstances(outlierData, "label", "features",
-        42L)
-    val weightedModel = lr.fit(weightedData)
-    val overSampledModel = lr.setWeightCol("").fit(overSampledData)
-    assert(weightedModel.coefficientMatrix ~== overSampledModel.coefficientMatrix relTol 0.01)
-  }
-
-  test("multinomial logistic regression with weighted data") {
-    val numClasses = 5
-    val numPoints = 40
-    val outlierData = MLTestingUtils.genClassificationInstancesWithWeightedOutliers(spark,
-      numClasses, numPoints)
-    val testData = Array.tabulate[LabeledPoint](numClasses) { i =>
-      LabeledPoint(i.toDouble, Vectors.dense(i.toDouble))
-    }.toSeq.toDF()
-    val mlr = new LogisticRegression().setFamily("multinomial").setWeightCol("weight")
-    val model = mlr.fit(outlierData)
-    val results = model.transform(testData).select("label", "prediction").collect()
-
-    // check that the predictions are the one to one mapping
-    results.foreach { case Row(label: Double, pred: Double) =>
-      assert(label === pred)
+    val testParams = Seq(
+      ("binomial", smallBinaryDataset, 2),
+      ("multinomial", smallMultinomialDataset, 4)
+    )
+    testParams.foreach { case (family, dataset, numClasses) =>
+      val estimator = new LogisticRegression().setFamily(family)
+      MLTestingUtils.testArbitrarilyScaledWeights[LogisticRegressionModel, LogisticRegression](
+        dataset.as[LabeledPoint], estimator, modelEquals)
+      MLTestingUtils.testOutliersWithSmallWeights[LogisticRegressionModel, LogisticRegression](
+        spark, estimator, Map.empty[Int, Int], numPoints, numClasses, numFeatures, modelEquals,
+        seed)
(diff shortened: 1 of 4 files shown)


sethah:
This issue also brings up a more general point of how sample weights should be tested. It seems there are some common rules that all algorithms that incorporate sample weights are thought to follow (mentioned above), but there are also algorithm specific details in some cases. I am of the opinion that we ought to incorporate some mixture of the two - common sample weight tests and algorithm specific tests. This is the approach followed in linear/logistic/generalized linear regression which all compare weighted data sets to R's output. Other algorithms that can't be compared directly with R will make use of the common helper functions. Random forests and decision trees will also heavily incorporate these into testing.

I do appreciate others' thoughts on this issue.

sethah:
ping @jkbradley @MLnick @yanboliang

yanboliang:
Sorry for late response, I like this change, will make a pass tomorrow. Thanks.

mllib/src/test/scala/org/apache/spark/ml/util/MLTestingUtils.scala
```diff
@@ -224,4 +208,139 @@ object MLTestingUtils extends SparkFunSuite {
...
+        categoricalFeaturesInfo: Map[Int, Int],
+        numPoints: Int,
+        numClasses: Int,
+        numFeatures: Int,
+        modelEquals: (M, M) => Unit,
+        seed: Long): Unit = {
+    import spark.implicits._
+    val df = generateNoisyData(numPoints, numClasses, numFeatures, categoricalFeaturesInfo,
+      seed).toDF()
+    val outlierFunction = getRandomLinearPredictionFunction(numFeatures, numClasses, seed - 1)
```
yanboliang: I'm more prefer to implement ```outlierFunction``` as a simple function such as:
* ```class -> numClass - class - 1``` for classification.
* ```label -> -label``` for regression.

which should be more intuitional and easy to be understand by developers/contributors.

mllib/src/test/scala/org/apache/spark/ml/util/MLTestingUtils.scala
```diff
@@ -224,4 +208,139 @@ object MLTestingUtils extends SparkFunSuite {
...
+   */
+  def testArbitrarilyScaledWeights[M <: Model[M], E <: Estimator[M]](
+      data: Dataset[LabeledPoint],
+      estimator: E with HasWeightCol with HasLabelCol with HasFeaturesCol,
+      modelEquals: (M, M) => Unit): Unit = {
+    estimator
+      .set(estimator.labelCol, "label")
+      .set(estimator.featuresCol, "features")
+      .set(estimator.weightCol, "weight")
+    val models = Seq(0.001, 1.0, 1000.0).map { w =>
```
yanboliang: I think ```Seq(1.0, 1000.0)``` should be enough.

yanboliang (commented):
@sethah I made a pass and left some comments. Thanks for working on it.

mllib/src/test/scala/org/apache/spark/ml/util/MLTestingUtils.scala
sethah: I disagree. 1.0 and 1000.0 are both integers, and I have already had experience with algorithms not properly handling fractional weights before. I think here we cover the case of (tiny, unit, and large) weights as well as (fractional, integer) weights.

yanboliang: Yeah, I got your concerns and I remember it happened at #16149, make sense.

mllib/src/test/scala/org/apache/spark/ml/util/MLTestingUtils.scala
```diff
@@ -224,4 +208,139 @@ object MLTestingUtils extends SparkFunSuite {
...
+   *
+   * @param numPoints The number of data points to generate.
+   * @param numClasses The number of classes the outcome can take on. 0 for continuous labels.
+   * @param numFeatures The number of features in the data.
+   * @param categoricalFeaturesInfo Map of (featureIndex -> numCategories) for categorical features.
+   * @param seed Random seed.
+   * @param noiseLevel A number in [0.0, 1.0] indicating how much noise to add to the label.
+   * @return Generated sequence of noisy instances.
+   */
+  def generateNoisyData(
```
sethah: Fair point. Actually, the noise is not strictly necessary for this patch in the other cases. I can use the existing datasets (for the most part). I removed this generator and passed the test data to the testing util methods.

mllib/src/test/scala/org/apache/spark/ml/util/MLTestingUtils.scala
```diff
@@ -224,4 +208,139 @@ object MLTestingUtils extends SparkFunSuite {
...
+      Instance(noisyLabel, 1.0, features)
+    }
+  }
+
+  /**
+   * Helper function for testing sample weights. Tests that oversampling each point is equivalent
+   * to assigning a sample weight proportional to the number of samples for each point.
+   */
+  def testOversamplingVsWeighting[M <: Model[M], E <: Estimator[M]](
+        spark: SparkSession,
```
sethah: done

mllib/src/test/scala/org/apache/spark/ml/util/MLTestingUtils.scala
```diff
@@ -224,4 +208,139 @@ object MLTestingUtils extends SparkFunSuite {
...
+        estimator: E with HasWeightCol with HasLabelCol with HasFeaturesCol,
+        categoricalFeaturesInfo: Map[Int, Int],
+        numPoints: Int,
+        numClasses: Int,
+        numFeatures: Int,
+        modelEquals: (M, M) => Unit,
+        seed: Long): Unit = {
+    import spark.implicits._
+    val df = generateNoisyData(numPoints, numClasses, numFeatures, categoricalFeaturesInfo,
```
sethah: done

mllib/src/test/scala/org/apache/spark/ml/util/MLTestingUtils.scala
sethah: Yeah, this works too and is simpler. Thanks!

sethah:
@yanboliang Thanks for reviewing, I addressed your comments.

mllib/src/test/scala/org/apache/spark/ml/regression/LinearRegressionSuite.scala
```diff
@@ -47,6 +49,11 @@ class LinearRegressionSuite
     datasetWithDenseFeature = sc.parallelize(LinearDataGenerator.generateLinearInput(
       intercept = 6.3, weights = Array(4.7, 7.2), xMean = Array(0.9, -1.3),
       xVariance = Array(0.7, 1.2), nPoints = 10000, seed, eps = 0.1), 2).map(_.asML).toDF()
+
+    weightedDatasetWithDenseFeature = sc.parallelize(LinearDataGenerator.generateLinearInput(
```
sethah: I added this small dataset with a higher noise value for weighted testing. It's necessary because when we test oversampling vs weighting, we need the noise to be high enough that the model learns incorrect coefficients when the weights are not applied. The coefficients used to generate each point are the same, but some points are emphasized more with weights. This dataset needs to be small enough and have enough noise that it doesn't still learn the true coefficients when the weights are not applied, if that makes sense.

yanboliang: ```datasetWithStrongNoise```? I think ```weighted*``` is really misleading.

mllib/src/test/scala/org/apache/spark/ml/classification/NaiveBayesSuite.scala
```diff
@@ -157,50 +162,26 @@ class NaiveBayesSuite extends SparkFunSuite with MLlibTestSparkContext with Defa
...
+      assert(m1.pi ~== m2.pi relTol 0.01)
+      assert(m1.theta ~== m2.theta relTol 0.01)
+    }
+    val testParams = Seq(
+      ("bernoulli", bernoulliDataset),
+      ("multinomial", dataset)
+    )
+    testParams.foreach { case (family, dataset) =>
+      // NaiveBayes is sensitive to constant scaling of the weights unless smoothing is set to 0
+      val estimator = new NaiveBayes().setSmoothing(0.0).setModelType(family)
```
yanboliang: I think it's not practical to set smoothing as 0.0, so it's better to test NB with none zero smoothing value. If it does not applicable to ```testArbitrarilyScaledWeights```, we can omit it. Or generating two estimators whose smoothing values are respectively 0.0 and 1.0 for different test functions.

mllib/src/test/scala/org/apache/spark/ml/classification/NaiveBayesSuite.scala
```diff
@@ -25,31 +25,36 @@ import breeze.stats.distributions.{Multinomial => BrzMultinomial}
...
 
-    val pi = Array(0.5, 0.1, 0.4).map(math.log)
+    val pi = Array(0.3, 0.3, 0.4).map(math.log)
     val theta = Array(
-      Array(0.70, 0.10, 0.10, 0.10), // label 0
-      Array(0.10, 0.70, 0.10, 0.10), // label 1
-      Array(0.10, 0.10, 0.70, 0.10)  // label 2
+      Array(0.30, 0.30, 0.30, 0.30), // label 0
+      Array(0.30, 0.30, 0.30, 0.30), // label 1
+      Array(0.40, 0.40, 0.40, 0.40)  // label 2
```
yanboliang: Could you let me know why you change this?

yanboliang (commented):
Only some minor comments, otherwise, looks good. Thanks.

mllib/src/test/scala/org/apache/spark/ml/classification/LogisticRegressionSuite.scala
```diff
@@ -1810,52 +1810,23 @@ class LogisticRegressionSuite
...
-    val lr = new LogisticRegression().setFamily("binomial").setWeightCol("weight")
-    val model = lr.fit(outlierData)
-    val results = model.transform(testData).select("label", "prediction").collect()
-
-    // check that the predictions are the one to one mapping
-    results.foreach { case Row(label: Double, pred: Double) =>
-      assert(label === pred)
+  test("logistic regression with sample weights") {
+    def modelEquals(m1: LogisticRegressionModel, m2: LogisticRegressionModel): Unit = {
+      assert(m1.coefficientMatrix ~== m2.coefficientMatrix absTol 0.01)
```
sethah: Done, thanks.

mllib/src/test/scala/org/apache/spark/ml/classification/NaiveBayesSuite.scala
sethah: Ya this is changed so that when we set smoothing to zero for the weighted tests, we don't get some theta values of infinity.

mllib/src/test/scala/org/apache/spark/ml/classification/NaiveBayesSuite.scala
sethah: I think the test with smoothing as 0.0 is a nice check on the weighting algorithm for Naive Bayes, so I prefer to keep it. I made separate smoothing/no smoothing estimators.

mllib/src/test/scala/org/apache/spark/ml/regression/LinearRegressionSuite.scala
sethah: Done.

mllib/src/test/scala/org/apache/spark/ml/util/MLTestingUtils.scala
```diff
@@ -18,15 +18,15 @@
...
 
 import org.apache.spark.SparkFunSuite
-import org.apache.spark.ml.{Estimator, Model}
-import org.apache.spark.ml.attribute.NominalAttribute
+import org.apache.spark.ml._
 import org.apache.spark.ml.evaluation.Evaluator
-import org.apache.spark.ml.feature.Instance
-import org.apache.spark.ml.linalg.{Vector, Vectors}
+import org.apache.spark.ml.feature.{Instance, LabeledPoint}
+import org.apache.spark.ml.linalg.{BLAS, DenseMatrix, DenseVector, Vector, Vectors}
```
sethah: done

mllib/src/test/scala/org/apache/spark/ml/util/MLTestingUtils.scala
```diff
@@ -182,34 +182,18 @@ object MLTestingUtils extends SparkFunSuite {
...
-      NominalAttribute.defaultAttr.withName("label").withNumValues(numClasses).toMetadata()
-    spark.createDataFrame(data).select(col("label").as("label", labelMeta), col("weight"),
-      col("features"))
-  }
-
+  /**
+   * Given a dataframe, generate two output dataframes: one having the original rows oversampled
+   * an integer number of times, and one having the original rows but with a column of weights
+   * proportional to the number of oversampled instances in the oversampled dataframe.
+   */
```
sethah: done

mllib/src/test/scala/org/apache/spark/ml/util/MLTestingUtils.scala
```diff
@@ -182,34 +182,18 @@ object MLTestingUtils extends SparkFunSuite {
...
   def genEquivalentOversampledAndWeightedInstances(
```
sethah: I made them all take `Dataset[LabeledPoint]`. Good suggestion.

mllib/src/test/scala/org/apache/spark/ml/util/MLTestingUtils.scala
```diff
@@ -224,4 +208,59 @@ object MLTestingUtils extends SparkFunSuite {
...
+    modelEquals(weightedModel, overSampledModel)
+  }
+
+  /**
+   * Helper function for testing sample weights. Tests that injecting a large number of outliers
+   * with very small sample weights does not affect fitting. The predictor should learn the true
+   * model despite the outliers.
+   */
+  def testOutliersWithSmallWeights[M <: Model[M], E <: Estimator[M]](
+      ds: Dataset[Instance],
```
sethah: done

yanboliang:
LGTM, merged into master. Thanks.

sethah:
Thanks @yanboliang!

PR #15722, opened 2016-11-01
[SPARK-18208] [Shuffle] Executor OOM due to a growing LongArray in BytesToBytesMap

jiexiong:

BytesToBytesMap currently does not release the in-memory storage (the longArray variable) after it spills to disk. This is typically not a problem during aggregation because the longArray should be much smaller than the pages, and because we grow the longArray at a conservative rate.

However this can lead to an OOM when an already running task is allocated more than its fair share, this can happen because of a scheduling delay. In this case the longArray can grow beyond the fair share of memory for the task. This becomes problematic when the task spills and the long array is not freed, that causes subsequent memory allocation requests to be denied by the memory manager resulting in an OOM.

This PR fixes this issuing by freeing the longArray when the BytesToBytesMap spills.

## How was this patch tested?

Existing tests and tested on realworld workloads.


core/src/main/java/org/apache/spark/unsafe/map/BytesToBytesMap.java
```diff
@@ -903,11 +906,12 @@ public void reset() {
     numKeys = 0;
     numValues = 0;
     longArray.zeroOut();
-
+    freeArray(longArray);
```
mridulm: You could also remove the zeroOut() since that is not relevant anymore

davies:
@jiexiong I don't think this is a memory leak, BytesToBytesMap does not release all memory for each spilling based on the assumption that the memory will be acquired back soon. What's the query that make you think this is a leak?

davies:
OK to test.

jiexiong:
Here is the query:
Here is the query:

INSERT OVERWRITE TABLE lookalike_trainer_campaign_conv_users_with_country_shadow
PARTITION(ds='2016-10-19')
SELECT c.source_id,
c.country,
c.user_id,
c.conversion_time
FROM (
SELECT b.source_id,
b.country,
b.user_id,
b.conversion_time,
FB_NUMBER_ROWS(b.country, b.source_id) as rank
FROM (
SELECT source_id,
country,
user_id,
MAX(conversion_time) / 1000 AS conversion_time
FROM (
SELECT v.campaigngroup_id,
v.campaign_id,
v.adgroup_id,
v.user_id,
Y.country,
v.last_conversion_time AS conversion_time
FROM dim_all_users_fast:bi Y
JOIN lookalike_trainer_campaign_conv_raw v
ON  v.user_id = Y.userid
WHERE v.ds='2016-10-19'
AND Y.ds = '2016-10-19'
AND Y.country IS NOT NULL
) a LATERAL VIEW
EXPLODE(ARRAY(campaigngroup_id, campaign_id, adgroup_id)) s
AS source_id
GROUP BY country, source_id, user_id
DISTRIBUTE by country, source_id
SORT BY country, source_id, conversion_time DESC
) b
) c  WHERE rank <= 60000

Before the fix, it would fail from OOM error. After the fix, the OOM error went away.

core/src/main/java/org/apache/spark/unsafe/map/BytesToBytesMap.java
jiexiong: Yeah, I will remove it.

jiexiong:
This is production query. Sorry, I could not share it. It is doing a join between two big tables.

davies:
@jiexiong Do you have a theory why this will cause OOM? To me, the current code will use more memory than needed but less allocation, why it will cause OOM?

jiexiong:
Here is my understanding: after spilling, it would call reset() to release the memory. In the reset() function, it deletes all the memory pages, but it did not release any memory from longArray(). So the longArray will keep growing. By freeing up the array and allocate only the initial size to it, we can ensure the longArray would not grow indefinitely.

mridulm:
I agree with @jiexiong's analysis : @davies to support the scenario you mentioned, we should not be free'ing the pages.

rxin:
@jiexiong can you update the pr description to include more information about the OOM and remove the facebook internal links? Thanks.

sitalkedia:
@davies - We fixed a similar issue with `UnsafeExternalSorter` in SPARK-14363. 
Basically following  scenario is leading to OOM - Lets say we have total 4G of memory available that is shared across 4 tasks, so fair share of each task is around 1G. At some point of time few tasks finish and before the scheduler can schedule more task on the executor, current running tasks grab all the memory from the memory manager. The `LongArray` in that situation grows beyond the fair share of memory for those tasks. Later when scheduler schedules more task on this executors, already running tasks are forced to spill, but since they are not reseting the `LongArray`, this is resulting in OOM.

core/src/main/java/org/apache/spark/unsafe/map/BytesToBytesMap.java
```diff
@@ -170,6 +170,8 @@
 
   private long peakMemoryUsedBytes = 0L;
 
+  private int initialCapacity;
```
sitalkedia: nit: final

davies:
@jiexiong The longArray will not grow indefinitely, it only grow when the number of keys reach 50% of it's size. Another assumption is that the memory used by longArray should be much smaller than the pages (longArray take 32 bytes per key, the pages take 56 bytes for a aggregate with 3 grouping key and 1 aggregate) . Is that true for your workload?

If the bookkeeping in the memory manager is right, it may do more spilling (because longArray is using more memory than expected), should not OOM. It's true that this patch could fix the OOM you saw in that query, but changing the memory factor (or other configs) should also fix that, I'm worrying there could be another bug in other places that cause the problem than this one. Could you dump logging how the memory is used when OOM happened?

sitalkedia:
@davies - Yes, we dumped the logging and confirmed that the OOM is because we are not freeing the `LongArray` while reseting the `BytesToBytesMap`. The job which used to fail because of OOM runs fine with this change. 

As explained above, the situation will lead to OOM when already running task is allocated more than fair share of its memory as a result of delay in scheduling by the scheduler. The `LongArray` itself can grown beyond the fair share of memory for the task (We have use cases where the `LongArray` is consuming significant portion of total memory because of too many keys) and later when the task spills, the `LongArray` is not freed as a result of which subsequent memory allocation request is denied by the memory manager resulting in OOM.

davies:
@sitalkedia I see your point, even the longArray only take a small factor of memory of the executor, but still could make other concurrent tasks fail to acquire enough initial memory, that make sense.

LGTM

sitalkedia:
The test case failure does not seem to be related to this change. May be flaky test?

hvanhovell:
retest this please

hvanhovell:
@jiexiong could you improve the PR description and add a description of what is actually causing this bug (a summary of the discussions in the PR would probably suffice), and how this PR fixes that?

hvanhovell:
ping @jiexiong

jiexiong:
@hvanhovell , I have already updated the description and explained how the PR fixed it. Could you please take another look?

dongjoon-hyun:
Retest this please.

core/src/main/java/org/apache/spark/unsafe/map/BytesToBytesMap.java
dongjoon-hyun: @jiexiong Could you fix this, too?

sitalkedia:
jenkins retest this please.

jiexiong:
Please retest!

hvanhovell:
retest this please

hvanhovell:
@jiexiong PR descriptions are used in git commit messages, and should be clear and concise. The fix LGTM, but the description should be improved for future reference. How about we change it into the following (I only redacted the comments in the PR):

`BytesToBytesMap` currently does not release the in-memory storage (the `longArray` variable) after it spills to disk. This is typically not a problem during aggregation because the longArray should be much smaller than the pages, and because we grow the `longArray` at a conservative rate.

However this can lead to an OOM when an already running task is allocated more than its fair share, this can happen because of a scheduling delay. In this case the `longArray` can grow beyond the fair share of memory for the task. This becomes problematic when the task spills and the long array is not freed, that causes subsequent memory allocation requests to be denied by the memory manager resulting in an OOM.

This PR fixes this issuing by freeing the `longArray` when the `BytesToBytesMap` spills.

## How was this patch tested?
Existing tests and tested on realworld workloads.

jiexiong:
@hvanhovell, thanks for your suggestion. I will change the PR description as suggested.

hvanhovell:
retest this please

hvanhovell:

PR #15724, opened 2016-11-02
[SPARK-18216][SQL] Make Column.expr public

rxin:
Column.expr is private[sql], but it's an actually really useful field to have for debugging. We should open it up, similar to how we use QueryExecution.

## How was this patch tested?
N/A - this is a simple visibility change.


marmbrus:
LGTM

ssimeonov:
👍

rxin:
Thanks - merging in master.

PR #15725, opened 2016-11-02
[SPARK-18167] Re-enable the non-flaky parts of SQLQuerySuite

ericl:

It seems the proximate cause of the test failures is that `cast(str as decimal)` in derby will raise an exception instead of returning NULL. This is a problem since Hive sometimes inserts `__HIVE_DEFAULT_PARTITION__` entries into the partition table as documented here: https://github.com/apache/hive/blob/trunk/metastore/src/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java#L1034

Basically, when these special default partitions are present, partition pruning pushdown using the SQL-direct mode will fail due this cast exception. As commented on in `MetaStoreDirectSql.java` above, this is normally fine since Hive falls back to JDO pruning, however when the pruning predicate contains an unsupported operator such as `>`, that will fail as well.

The only remaining question is why this behavior is nondeterministic. We know that when the test flakes, retries do not help, therefore the cause must be environmental. The current best hypothesis is that some config is different between different jenkins runs, which is why this PR prints out the Spark SQL and Hive confs for the test. The hope is that by comparing the config state for failure vs success we can isolate the root cause of the flakiness.

**Update:** we could not isolate the issue. It does not seem to be due to configuration differences. As such, I'm going to enable the non-flaky parts of the test since we are fairly confident these issues only occur with Derby (which is not used in production).

## How was this patch tested?

N/A


ericl:
This also enables the test. @marmbrus feel free to revert this patch as soon as we have a sample failure output.

ericl:
@rxin

ericl:
Oh cool it flaked on that run

jenkins retest this please. // let's try to get a successful run for comparison

JoshRosen:
Jenkins, retest this please

viirya:
`spark.sql.hive.metastorePartitionPruning` is seen in failed test, not in successful test.

ericl:
Oh wow, so it's deterministically broken, but some suite must be leaking a
conf that disables the flag. This explains why I couldn't find a preceding
suite to blame for the flake - it's merely the absence of a suite that
causes the failure.

viirya:
Should be HiveTableScanSuite. I think.

viirya:
`HiveTableScanSuite` changes the config `spark.sql.hive.metastorePartitionPruning` to false in the end.

In the failed test run, `HiveTableScanSuite` runs before `SQLQuerySuite`. In the successful run, `HiveTableScanSuite` runs after `SQLQuerySuite`.

ericl:
After some local experimental, I think that is actually a red herring. `spark.sql.hive.metastorePartitionPruning` defaults to true now, so omitting it has no effect. Also, I am not able to reproduce the failure locally by setting any of the spark confs. (Haven't tried the hive confs since there are so many, but I don't see anything obviously different there).

yhuai:
lgtm. merging to master and branch 2.1.

HyukjinKwon:
Matbe this one is also flaky..
https://amplab.cs.berkeley.edu/jenkins/job/SparkPullRequestBuilder/68198/consoleFull
https://amplab.cs.berkeley.edu/jenkins/job/SparkPullRequestBuilder/68185/consoleFull

jiangxb1987:
This is also flaky..
https://amplab.cs.berkeley.edu/jenkins/job/SparkPullRequestBuilder/68219/consoleFull

PR #15727, opened 2016-11-02
[SPARK-17895] Improve doc for rangeBetween and rowsBetween

david-weiluo-ren:

Copied description for row and range based frame boundary from https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/window/WindowExec.scala#L56

Added examples to show different behavior of rangeBetween and rowsBetween when involving duplicate values.


rxin:
@david-weiluo-ren this already has a conflict. Can you update it?

sql/core/src/main/scala/org/apache/spark/sql/expressions/Window.scala
```diff
@@ -91,7 +91,7 @@ object Window {
    * {{{
    *   import org.apache.spark.sql.expressions.Window
    *   val df = Seq((1, "a"), (1, "a"), (2, "a"), (1, "b"), (2, "b"), (3, "b")).toDF("id", "category")
-   *   df.withColumn("sum", sum('id) over Window.partitionBy('category).orderBy('id).rowsBetween(0,1)).show
+   *   df.withColumn("sum", sum('id) over Window.partitionBy('category).orderBy('id).rowsBetween(0,1)).show()
```
rxin: also you should probably use `.over(...)` rather than infix notation here

rxin:

PR #15729, opened 2016-11-02
[SPARK-18133] [branch-2.0] [Examples] [ML] [Python ML Pipeline Exampl…

jagadeesanas2:

[Fix] [branch-2.0] In Python 3, there is only one integer type (i.e., int), which mostly behaves like the long type in Python 2. Since Python 3 won't accept "L", so removed "L" in all examples.

## How was this patch tested?

Unit tests.

…e has syntax errors]


srowen:
You can close this PR now. I merged it.

PR #15731, opened 2016-11-02
[SPARK-18219] Move commit protocol API (internal) from sql/core to core module

rxin:
This patch moves the new commit protocol API from sql/core to core module, so we can use it in the future in the RDD API.

As part of this patch, I also moved the speficiation of the random uuid for the write path out of the commit protocol, and instead pass in a job id.

## How was this patch tested?
N/A


rxin:
cc @jiangxb1987

core/src/main/scala/org/apache/spark/internal/io/FileCommitProtocol.scala
```diff
@@ -0,0 +1,131 @@
...
+
+object FileCommitProtocol {
+  class TaskCommitMessage(val obj: Any) extends Serializable
+
+  object EmptyTaskCommitMessage extends TaskCommitMessage(null)
+
+  /**
+   * Instantiates a FileCommitProtocol using the given className.
+   */
+  def instantiate(className: String, jobId: String, outputPath: String, isAppend: Boolean)
```
rxin: the main change here is the introduction of a "jobId".

mridulm:
I am not sure I follow, why do we need to redefine OutputCommitter from mapreduce with slightly differing api ?

rxin:
It is insufficient for a lot of use cases, e.g. streaming.

mridulm:
Thx for clarifying @rxin, looks good to me.

rxin:
Merging in master/branch-2.1.

jiangxb1987:
I plan to push my work tonight.

jiangxb1987:
Still working on it, sorry for delay.
The listed works are:
1. Port `SparkHadoopWriter` to use commit protocol;
2. Write a new `SparkNewHadoopWriter` to write `RDD` through new Hadoop API;
3. Modify `PairRDDFunctions` and `PairRDDFunctionsSuite` to use the above writers properly.

Still working on 2 & 3, will try to solve them ASAP.

PR #15734, opened 2016-11-02
[SPARK-18222][CORE][SQL][ML] Use math instead of Math

zhengruifeng:
Use `scala.math` instead of `java.lang.Math`.
Using cmd `find . -name '*.scala' | xargs -i bash -c 'egrep "Math\." -n {} && echo {}'` to generate modification candidates.

## How was this patch tested?
existing tests

### DIFF
--- a/common/unsafe/src/test/scala/org/apache/spark/unsafe/types/UTF8StringPropertyCheckSuite.scala
+++ b/common/unsafe/src/test/scala/org/apache/spark/unsafe/types/UTF8StringPropertyCheckSuite.scala
@@ -77,7 +77,7 @@ class UTF8StringPropertyCheckSuite extends FunSuite with GeneratorDrivenProperty
 
   test("compare") {
     forAll { (s1: String, s2: String) =>
-      assert(Math.signum(toUTF8(s1).compareTo(toUTF8(s2))) === Math.signum(s1.compareTo(s2)))
+      assert(math.signum(toUTF8(s1).compareTo(toUTF8(s2))) === math.signum(s1.compareTo(s2)))
     }
   }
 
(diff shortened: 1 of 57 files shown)


srowen:
These are actually usages of java.lang.Math. I do kind of like standardizing in one go, but if so, I wonder if you can add a scalastyle rule to forbid direct use of java.lang.Math?

zhengruifeng:
@srowen You are right. What should I do next: close this PR for `scala.Math` is not used ? or view it as a kind of usage unification?

zhengruifeng:
@srowen OK, I know what you mean. I wil try to make the scalastyle rule.

zhengruifeng:
I add a scalastyle rule and it work in my local env with `./dev/lint-scala`

rxin:
Actually why would we do this at all? Often the scala implementation are just slower than the Java one (sometimes because they are wrappers).

srowen:
Yeah `scala.math` is all wrappers. It can't be faster though I think it won't matter after inlining. I suppose it's nicer stylewise to standardize, and `scala.math` is a bit more natural. If so it deserves a style rule. I don't feel that strongly about it. This PR started under a different assumption about the effect of this change.

rxin:
A lot of wrappers in Scala are slower even with JIT, e.g. at some point Spark used Scala's StringBuffer and some user found it was 30% slower than Java's. Anyway even if there is 0% performance difference, I don't think it is that big of a deal to have some code using scala.math and some using java math.

zhengruifeng:
this pr/jira started with a wrong point: I mistakenly view `Math` as `scala.Math`. I will close it.

PR #15736, opened 2016-11-02
[SPARK-18224] [CORE] Optimise PartitionedPairBuffer implementation

a-roberts:
This change is very similar to my pull request for improving PartitionedPairAppendOnlyMap: https://github.com/apache/spark/pull/15735

Summarising (more detail above), we avoid the slow iterator wrapping in favour of helping the inliner. We observed that this, when combined with the above change, leads to a 3% performance increase on the HiBench large PageRank benchmark with both IBM's SDK for Java and with OpenJDK 8

## How was this patch tested?
Existing unit tests and HiBench large profile with both IBM's SDK for Java and OpenJDK 8, the PageRank benchmark specifically

### DIFF
--- a/core/src/main/scala/org/apache/spark/util/collection/PartitionedAppendOnlyMap.scala
+++ b/core/src/main/scala/org/apache/spark/util/collection/PartitionedAppendOnlyMap.scala
@@ -30,7 +30,21 @@ private[spark] class PartitionedAppendOnlyMap[K, V]
 
   def partitionedDestructiveSortedIterator(keyComparator: Option[Comparator[K]])
     : Iterator[((Int, K), V)] = {
-    val comparator = keyComparator.map(partitionKeyComparator).getOrElse(partitionComparator)
+    val comparator : Comparator[(Int, K)] =
+      if (keyComparator.isEmpty) {
+        partitionComparator
+      } else {
+        new Comparator[(Int, K)] {
+          override def compare(a: (Int, K), b: (Int, K)): Int = {
+          val partitionDiff = a._1 - b._1
+          if (partitionDiff != 0) {
+            partitionDiff
+          } else {
+            keyComparator.get.compare(a._2, b._2)
+          }
+        }
+      }
+    }
     destructiveSortedIterator(comparator)
   }
 
--- a/core/src/main/scala/org/apache/spark/util/collection/PartitionedAppendOnlyMap.scala
+++ b/core/src/main/scala/org/apache/spark/util/collection/PartitionedAppendOnlyMap.scala
@@ -37,4 +37,3 @@ private[spark] class PartitionedAppendOnlyMap[K, V]
     update((partition, key), value)
   }
 }
-
--- a/core/src/main/scala/org/apache/spark/util/collection/PartitionedAppendOnlyMap.scala
+++ b/core/src/main/scala/org/apache/spark/util/collection/PartitionedAppendOnlyMap.scala
@@ -30,25 +30,11 @@ private[spark] class PartitionedAppendOnlyMap[K, V]
 
   def partitionedDestructiveSortedIterator(keyComparator: Option[Comparator[K]])
     : Iterator[((Int, K), V)] = {
-    val comparator : Comparator[(Int, K)] =
-      if (keyComparator.isEmpty) {
-        partitionComparator
-      } else {
-        new Comparator[(Int, K)] {
-          override def compare(a: (Int, K), b: (Int, K)): Int = {
-          val partitionDiff = a._1 - b._1
-          if (partitionDiff != 0) {
-            partitionDiff
-          } else {
-            keyComparator.get.compare(a._2, b._2)
-          }
-        }
-      }
-    }
-    destructiveSortedIterator(comparator)
+    destructiveSortedIterator(getComparator(keyComparator))
   }
 
   def insert(partition: Int, key: K, value: V): Unit = {
     update((partition, key), value)
   }
 }
+
--- a/core/src/main/scala/org/apache/spark/util/collection/PartitionedPairBuffer.scala
+++ b/core/src/main/scala/org/apache/spark/util/collection/PartitionedPairBuffer.scala
@@ -74,8 +74,8 @@ private[spark] class PartitionedPairBuffer[K, V](initialCapacity: Int = 64)
   /** Iterate through the data in a given order. For this class this is not really destructive. */
   override def partitionedDestructiveSortedIterator(keyComparator: Option[Comparator[K]])
     : Iterator[((Int, K), V)] = {
-    new Sorter(new KVArraySortDataFormat[(Int, K),
-      AnyRef]).sort(data, 0, curSize, getComparator(keyComparator))
+    val comparator = getComparator(keyComparator)
+    new Sorter(new KVArraySortDataFormat[(Int, K), AnyRef]).sort(data, 0, curSize, comparator)
     iterator
   }
 
--- a/core/src/main/scala/org/apache/spark/util/collection/PartitionedPairBuffer.scala
+++ b/core/src/main/scala/org/apache/spark/util/collection/PartitionedPairBuffer.scala
@@ -74,22 +74,8 @@ private[spark] class PartitionedPairBuffer[K, V](initialCapacity: Int = 64)
   /** Iterate through the data in a given order. For this class this is not really destructive. */
   override def partitionedDestructiveSortedIterator(keyComparator: Option[Comparator[K]])
     : Iterator[((Int, K), V)] = {
-    val comparator : Comparator[(Int, K)] =
-      if (keyComparator.isEmpty) {
-        partitionComparator
-    } else {
-      new Comparator[(Int, K)] {
-        override def compare(a: (Int, K), b: (Int, K)): Int = {
-          val partitionDiff = a._1 - b._1
-            if (partitionDiff != 0) {
-              partitionDiff
-            } else {
-              keyComparator.get.compare(a._2, b._2)
-            }
-        }
-      }
-    }
-    new Sorter(new KVArraySortDataFormat[(Int, K), AnyRef]).sort(data, 0, curSize, comparator)
+    new Sorter(new KVArraySortDataFormat[(Int, K),
+      AnyRef]).sort(data, 0, curSize, getComparator(keyComparator))
     iterator
   }
 
@@ -112,3 +98,4 @@ private[spark] class PartitionedPairBuffer[K, V](initialCapacity: Int = 64)
 private object PartitionedPairBuffer {
   val MAXIMUM_CAPACITY = Int.MaxValue / 2 // 2 ^ 30 - 1
 }
+
--- a/core/src/main/scala/org/apache/spark/util/collection/PartitionedPairBuffer.scala
+++ b/core/src/main/scala/org/apache/spark/util/collection/PartitionedPairBuffer.scala
@@ -98,4 +98,3 @@ private[spark] class PartitionedPairBuffer[K, V](initialCapacity: Int = 64)
 private object PartitionedPairBuffer {
   val MAXIMUM_CAPACITY = Int.MaxValue / 2 // 2 ^ 30 - 1
 }
-
--- a/core/src/main/scala/org/apache/spark/util/collection/PartitionedPairBuffer.scala
+++ b/core/src/main/scala/org/apache/spark/util/collection/PartitionedPairBuffer.scala
@@ -74,7 +74,20 @@ private[spark] class PartitionedPairBuffer[K, V](initialCapacity: Int = 64)
   /** Iterate through the data in a given order. For this class this is not really destructive. */
   override def partitionedDestructiveSortedIterator(keyComparator: Option[Comparator[K]])
     : Iterator[((Int, K), V)] = {
-    val comparator = keyComparator.map(partitionKeyComparator).getOrElse(partitionComparator)
+    val comparator : Comparator[(Int, K)] = 
+      if (keyComparator.isEmpty) {
+        partitionComparator
+    } else
+      new Comparator[(Int, K)] {
+        override def compare(a: (Int, K), b: (Int, K)): Int = {
+          val partitionDiff = a._1 - b._1
+            if (partitionDiff != 0) {
+              partitionDiff
+            } else {
+              keyComparator.get.compare(a._2, b._2)
+            }
+        }
+    }
     new Sorter(new KVArraySortDataFormat[(Int, K), AnyRef]).sort(data, 0, curSize, comparator)
     iterator
   }
--- a/core/src/main/scala/org/apache/spark/util/collection/PartitionedPairBuffer.scala
+++ b/core/src/main/scala/org/apache/spark/util/collection/PartitionedPairBuffer.scala
@@ -74,10 +74,10 @@ private[spark] class PartitionedPairBuffer[K, V](initialCapacity: Int = 64)
   /** Iterate through the data in a given order. For this class this is not really destructive. */
   override def partitionedDestructiveSortedIterator(keyComparator: Option[Comparator[K]])
     : Iterator[((Int, K), V)] = {
-    val comparator : Comparator[(Int, K)] = 
+    val comparator : Comparator[(Int, K)] =
       if (keyComparator.isEmpty) {
         partitionComparator
-    } else
+    } else {
       new Comparator[(Int, K)] {
         override def compare(a: (Int, K), b: (Int, K)): Int = {
           val partitionDiff = a._1 - b._1
@@ -87,6 +87,7 @@ private[spark] class PartitionedPairBuffer[K, V](initialCapacity: Int = 64)
               keyComparator.get.compare(a._2, b._2)
             }
         }
+      }
     }
     new Sorter(new KVArraySortDataFormat[(Int, K), AnyRef]).sort(data, 0, curSize, comparator)
     iterator
--- a/core/src/main/scala/org/apache/spark/util/collection/WritablePartitionedPairCollection.scala
+++ b/core/src/main/scala/org/apache/spark/util/collection/WritablePartitionedPairCollection.scala
@@ -74,42 +74,23 @@ private[spark] object WritablePartitionedPairCollection {
     }
   }
 
-  /* Takes an optional parameter (keyComparator), use if provided
+  /**
+   * Takes an optional parameter (keyComparator), use if provided
    * and returns a comparator for the partitions
    */
-  def getComparator[K](keyComparator: Option[Comparator[K]]) : Comparator[(Int, K)] = {
-    val comparator : Comparator[(Int, K)] =
-      if (keyComparator.isEmpty) {
-        partitionComparator
-      } else {
-        new Comparator[(Int, K)] {
-          // We know we have a non-empty comparator here
-          val ourKeyComp = keyComparator.get
-          override def compare(a: (Int, K), b: (Int, K)): Int = {
-            val partitionDiff = a._1 - b._1
-            if (partitionDiff != 0) {
-              partitionDiff
-            } else {
-              ourKeyComp.compare(a._2, b._2)
-            }
+  def getComparator[K](keyComparator: Option[Comparator[K]]): Comparator[(Int, K)] = {
+    if (!keyComparator.isDefined) return partitionComparator
+    else {
+      val theKeyComp = keyComparator.get
+      new Comparator[(Int, K)] {
+        // We know we have a non-empty comparator here
+        override def compare(a: (Int, K), b: (Int, K)): Int = {
+          val partitionDiff = a._1 - b._1
+          if (partitionDiff != 0) {
+            partitionDiff
+          } else {
+            theKeyComp.compare(a._2, b._2)
           }
-        }
-      }
-    comparator
-  }
-
-
-  /**
-   * A comparator for (Int, K) pairs that orders them both by their partition ID and a key ordering.
-   */
-  def partitionKeyComparator[K](keyComparator: Comparator[K]): Comparator[(Int, K)] = {
-    new Comparator[(Int, K)] {
-      override def compare(a: (Int, K), b: (Int, K)): Int = {
-        val partitionDiff = a._1 - b._1
-        if (partitionDiff != 0) {
-          partitionDiff
-        } else {
-          keyComparator.compare(a._2, b._2)
         }
       }
     }
--- a/core/src/main/scala/org/apache/spark/util/collection/WritablePartitionedPairCollection.scala
+++ b/core/src/main/scala/org/apache/spark/util/collection/WritablePartitionedPairCollection.scala
@@ -74,6 +74,31 @@ private[spark] object WritablePartitionedPairCollection {
     }
   }
 
+  /* Takes an optional parameter (keyComparator), use if provided
+   * and returns a comparator for the partitions
+   */
+  def getComparator[K](keyComparator: Option[Comparator[K]]) : Comparator[(Int, K)] = {
+    val comparator : Comparator[(Int, K)] =
+      if (keyComparator.isEmpty) {
+        partitionComparator
+      } else {
+        new Comparator[(Int, K)] {
+          // We know we have a non-empty comparator here
+          val ourKeyComp = keyComparator.get
+          override def compare(a: (Int, K), b: (Int, K)): Int = {
+            val partitionDiff = a._1 - b._1
+            if (partitionDiff != 0) {
+              partitionDiff
+            } else {
+              ourKeyComp.compare(a._2, b._2)
+            }
+          }
+        }
+      }
+    comparator
+  }
+
+
   /**
    * A comparator for (Int, K) pairs that orders them both by their partition ID and a key ordering.
    */
@@ -102,3 +127,4 @@ private[spark] trait WritablePartitionedIterator {
 
   def nextPartition(): Int
 }
+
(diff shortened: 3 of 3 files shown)


core/src/main/scala/org/apache/spark/util/collection/PartitionedPairBuffer.scala
```diff
@@ -74,7 +74,20 @@ private[spark] class PartitionedPairBuffer[K, V](initialCapacity: Int = 64)
...
   override def partitionedDestructiveSortedIterator(keyComparator: Option[Comparator[K]])
     : Iterator[((Int, K), V)] = {
-    val comparator = keyComparator.map(partitionKeyComparator).getOrElse(partitionComparator)
+    val comparator : Comparator[(Int, K)] = 
+      if (keyComparator.isEmpty) {
+        partitionComparator
+    } else
+      new Comparator[(Int, K)] {
+        override def compare(a: (Int, K), b: (Int, K)): Int = {
+          val partitionDiff = a._1 - b._1
```
srowen: There are some indentation problems here and the else clause is missing a brace. I think you can omit the type of `comparator`; no space before the colon in any event.

This subtraction can overflow in theory and give the wrong answer, but the existing code does it, so, pass on that.

While optimizing, do you want to call keyComparator.get outside the class definition?

There's a similar construct in PartitionedAppendOnlyMap that should be changed too. Can this be refactored maybe?

Can the method partitionKeyComparator go away? I think the whole WritablePartitionedPairCollection object goes away after this if you care to 'inline' it too in the one refactored instance.

a-roberts:
Will be adding the commit from #15735 here upon addressing the feedback

a-roberts:
Addressed the scalastyle comments and added the PartitionedAppendOnlyMap change here as per the above suggestions, will look at the review comments next

Two unrelated asides
1) wary I'm hogging the build machines, would be useful to not autotest everytime
2) dev/scalastyle should accept a parameter so we can quickly check just the one file, takes a long time typically

core/src/main/scala/org/apache/spark/util/collection/PartitionedAppendOnlyMap.scala
```diff
@@ -30,7 +30,21 @@ private[spark] class PartitionedAppendOnlyMap[K, V]
...
+        partitionComparator
+      } else {
+        new Comparator[(Int, K)] {
+          override def compare(a: (Int, K), b: (Int, K)): Int = {
+          val partitionDiff = a._1 - b._1
+          if (partitionDiff != 0) {
+            partitionDiff
+          } else {
+            keyComparator.get.compare(a._2, b._2)
+          }
```
mridulm: You can dereference the option to avoid get in inner loop

srowen: Yeah I think Github collapsed it but there's this and other suggestions at ... https://github.com/apache/spark/pull/15736#pullrequestreview-6801163

srowen:
To recap some of my feedback here, I think this will be a fine change but it can be refactored further.

I think we can refactor this logic that appears twice in one place, perhaps in `object WritablePartitionedPairCollection`? that's where its support code currently lives, anyway.

Since the two methods there are only used by these call sites that are changing, they can be 'inlined' into the one common implementation, which might open up more optimization.

It'd be nice to fix the subtraction issue while we're here unless someone is convinced the difference can never overflow.

`keyComparator.get` can be lifted out of the compare() method.

srowen:
@a-roberts this does look like a worthy change, what do you think of the further simplifications here?

a-roberts:
Sean, they are great suggestions, thanks -- I'll find the time (like for the other outstanding pull requests) to get your feedback integrated, tested and profiled, currently caught up in packaging our own Apache Spark releases for both 1.6.3 and 2.0.2. I also have a JIRA to create proposing regular performance runs using the latest Spark snapshot builds to track regressions (I have this all set up with scripts already)

srowen:
I know you're busy but this does look like a good change to finish off. That it's a win is self-evident, just a question of how much, and benchmarks you have already show it is an improvement. I can take it on (credit remains with you) or will just wait if you're getting back to it.

a-roberts:
I'm resuming the work for all of these related PRs again this week after the London Spark meetup on Wednesday, if you are keen to take it on I'm more than happy to help out and will share some information here that yourself and others should find useful.

**Useful tools**
As well as simple microbenchmarking we use the [Linux perf tools](https://perf.wiki.kernel.org/index.php/Main_Page), [tprof](http://perfinsp.sourceforge.net/tprof.html) with [Visual Performance Analyzer](http://www.scalabiliti.com/library/visual_performance_analyzer) and also IBM's [Healthcenter](https://www.ibm.com/developerworks/java/jdk/tools/healthcenter/) for Java for method profiling (this is bundled with the JDK and you provide -Xhealthcenter as a driver/executor option then open the files in said tool).

**Benchmarks**
We'd want to run our improvement ideas with and without the changes using[ HiBench 6](https://github.com/intel-hadoop/HiBench) (large profile) and [SparkSqlPerf](https://github.com/databricks/spark-sql-perf) against all 100 TPCDS queries.

a-roberts:
Back to working on the performance related JIRAs now, so based on the above helpful comments here's what I'll do

Remove the .get.compare from the loop as suggested above - we'll do a .get  upfront to get our comparator to use, eliminating the .get later

Move the duplicated code into the WritablePartitionedPairCollection object so the two methods optimised here will call the above new method (let's say it's called getComparator) before returning accordingly (both methods are the same apart from the final few lines).

PartitionedAppendOnlyMap returns 
```
destructiveSortedIterator(comparator)
```

and PartitionedPairBuffer returns:

```
new Sorter(new KVArraySortDataFormat[(Int, K), AnyRef]).sort(data, 0, curSize, comparator)
iterator
```

I'll then build/test/profile this again

core/src/main/scala/org/apache/spark/util/collection/WritablePartitionedPairCollection.scala
```diff
@@ -74,6 +74,31 @@ private[spark] object WritablePartitionedPairCollection {
     }
   }
 
+  /* Takes an optional parameter (keyComparator), use if provided
```
srowen: Javadoc/scaladoc starts with `/**` and usually you leave that alone on one line and start documentation on the next.

core/src/main/scala/org/apache/spark/util/collection/WritablePartitionedPairCollection.scala
```diff
@@ -74,6 +74,31 @@ private[spark] object WritablePartitionedPairCollection {
...
+   * and returns a comparator for the partitions
+   */
+  def getComparator[K](keyComparator: Option[Comparator[K]]) : Comparator[(Int, K)] = {
+    val comparator : Comparator[(Int, K)] =
```
srowen: `comparator` is now entirely redundant. The whole body is just the `if` statement

core/src/main/scala/org/apache/spark/util/collection/PartitionedPairBuffer.scala
```diff
@@ -74,8 +74,8 @@ private[spark] class PartitionedPairBuffer[K, V](initialCapacity: Int = 64)
   /** Iterate through the data in a given order. For this class this is not really destructive. */
   override def partitionedDestructiveSortedIterator(keyComparator: Option[Comparator[K]])
     : Iterator[((Int, K), V)] = {
-    val comparator = keyComparator.map(partitionKeyComparator).getOrElse(partitionComparator)
-    new Sorter(new KVArraySortDataFormat[(Int, K), AnyRef]).sort(data, 0, curSize, comparator)
+    new Sorter(new KVArraySortDataFormat[(Int, K),
```
srowen: I think breaking the line here is odd. If necesasry, pull out the result of `getComparator` to a statement above to shorten this line, like it was before.

core/src/main/scala/org/apache/spark/util/collection/WritablePartitionedPairCollection.scala
```diff
@@ -74,6 +74,31 @@ private[spark] object WritablePartitionedPairCollection {
...
+      if (keyComparator.isEmpty) {
+        partitionComparator
+      } else {
+        new Comparator[(Int, K)] {
+          // We know we have a non-empty comparator here
+          val ourKeyComp = keyComparator.get
```
srowen: I think this should be outside the body of the anonymous class. You don't need a reference to the `Option` here even in the anonymous class.

core/src/main/scala/org/apache/spark/util/collection/WritablePartitionedPairCollection.scala
```diff
@@ -74,6 +74,31 @@ private[spark] object WritablePartitionedPairCollection {
     }
   }
 
+  /* Takes an optional parameter (keyComparator), use if provided
+   * and returns a comparator for the partitions
+   */
+  def getComparator[K](keyComparator: Option[Comparator[K]]) : Comparator[(Int, K)] = {
```
srowen: Nit: no space before colon

core/src/main/scala/org/apache/spark/util/collection/WritablePartitionedPairCollection.scala
```diff
@@ -74,6 +74,31 @@ private[spark] object WritablePartitionedPairCollection {
...
+    val comparator : Comparator[(Int, K)] =
+      if (keyComparator.isEmpty) {
```
srowen: `isDefined` is probably a tiny bit more conventional (and then flip the logic here of course)

core/src/main/scala/org/apache/spark/util/collection/WritablePartitionedPairCollection.scala
```diff
@@ -74,6 +74,31 @@ private[spark] object WritablePartitionedPairCollection {
...
+            } else {
+              ourKeyComp.compare(a._2, b._2)
+            }
+          }
+        }
+      }
+    comparator
+  }
+
+
```
srowen: You can delete partitionKeyComparator below now, right?

core/src/main/scala/org/apache/spark/util/collection/WritablePartitionedPairCollection.scala
```diff
@@ -74,6 +74,31 @@ private[spark] object WritablePartitionedPairCollection {
     }
   }
 
+  /* Takes an optional parameter (keyComparator), use if provided
+   * and returns a comparator for the partitions
+   */
+  def getComparator[K](keyComparator: Option[Comparator[K]]) : Comparator[(Int, K)] = {
+    val comparator : Comparator[(Int, K)] =
+      if (keyComparator.isEmpty) {
+        partitionComparator
```
srowen: This can be inlined now

core/src/main/scala/org/apache/spark/util/collection/WritablePartitionedPairCollection.scala
```diff
@@ -74,6 +74,31 @@ private[spark] object WritablePartitionedPairCollection {
...
+      } else {
+        new Comparator[(Int, K)] {
+          // We know we have a non-empty comparator here
+          val ourKeyComp = keyComparator.get
+          override def compare(a: (Int, K), b: (Int, K)): Int = {
+            val partitionDiff = a._1 - b._1
```
srowen: I'm still not thrilled about the subtraction here but maybe leave it for now

srowen (changes requested):

a-roberts:
I've conducted a lot of performance tests and gathered .hcd files so I can investigate this next week, but it looks like either the first commit is the best for performance or my current configuration with this benchmark results in us being unable to infer if our changes really make a difference.

Sharing some raw data, the format is as follows.

Benchmark name, date, time, data size in bytes (the same each run), the elapsed time and the throughput (bytes per second).

**With the above suggestions for Partitioned*Buffer**
```
ScalaSparkPagerank 2016-11-25 18:49:23 259928115            49.577               5242917              
ScalaSparkPagerank 2016-11-25 18:56:55 259928115            49.946               5204182              
ScalaSparkPagerank 2016-11-25 19:00:04 259928115            46.510               5588650              
ScalaSparkPagerank 2016-11-25 19:02:23 259928115            49.018               5302707              
ScalaSparkPagerank 2016-11-25 19:05:25 259928115            49.270               5275585              
```

**Vanilla, no changes at all**
```
ScalaSparkPagerank 2016-11-25 19:08:45 259928115            48.068               5407508              
ScalaSparkPagerank 2016-11-25 19:11:20 259928115            47.712               5447856              
ScalaSparkPagerank 2016-11-25 19:13:50 259928115            44.517               5838850              
ScalaSparkPagerank 2016-11-25 19:16:07 259928115            49.942               5204599              
ScalaSparkPagerank 2016-11-25 19:19:08 259928115            48.521               5357023              
```

**Original commit**
```
ScalaSparkPagerank 2016-11-25 19:47:59 259928115            45.486               5714464              
ScalaSparkPagerank 2016-11-25 19:50:48 259928115            48.507               5358569              
ScalaSparkPagerank 2016-11-25 19:53:09 259928115            47.063               5522982              
ScalaSparkPagerank 2016-11-25 19:56:58 259928115            46.154               5631757              
ScalaSparkPagerank 2016-11-25 20:00:01 259928115            48.935               5311701        
```

In Healthcenter I do see that these methods are still great candidates for optimisation as they are all very commonly used.

Open to more suggestions, I have exclusive access to lots of hardware, can easily churn out more custom builds and have lots of profiling software we can use. I'll be committing code for the SizeEstimator soon as that's a good candidate for optimisation here as well.

core/src/main/scala/org/apache/spark/util/collection/WritablePartitionedPairCollection.scala
```diff
@@ -75,16 +75,22 @@ private[spark] object WritablePartitionedPairCollection {
...
-    new Comparator[(Int, K)] {
-      override def compare(a: (Int, K), b: (Int, K)): Int = {
-        val partitionDiff = a._1 - b._1
-        if (partitionDiff != 0) {
-          partitionDiff
-        } else {
-          keyComparator.compare(a._2, b._2)
+  def getComparator[K](keyComparator: Option[Comparator[K]]): Comparator[(Int, K)] = {
+    if (!keyComparator.isDefined) return partitionComparator
+    else {
```
srowen: Style is off here -- you need braces in both clauses, `return` is redundant, and there's no point in inverting the condition as opposed to just flipping the clauses.

core/src/main/scala/org/apache/spark/util/collection/WritablePartitionedPairCollection.scala
```diff
@@ -75,16 +75,22 @@ private[spark] object WritablePartitionedPairCollection {
...
-  def partitionKeyComparator[K](keyComparator: Comparator[K]): Comparator[(Int, K)] = {
-    new Comparator[(Int, K)] {
-      override def compare(a: (Int, K), b: (Int, K)): Int = {
-        val partitionDiff = a._1 - b._1
-        if (partitionDiff != 0) {
-          partitionDiff
-        } else {
-          keyComparator.compare(a._2, b._2)
+  def getComparator[K](keyComparator: Option[Comparator[K]]): Comparator[(Int, K)] = {
+    if (!keyComparator.isDefined) return partitionComparator
```
srowen: inline and remove partitionComparator, as I think it's not used

core/src/main/scala/org/apache/spark/util/collection/WritablePartitionedPairCollection.scala
```diff
@@ -75,16 +75,22 @@ private[spark] object WritablePartitionedPairCollection {
...
+    else {
+      val theKeyComp = keyComparator.get
+      new Comparator[(Int, K)] {
+        // We know we have a non-empty comparator here
+        override def compare(a: (Int, K), b: (Int, K)): Int = {
+          val partitionDiff = a._1 - b._1
+          if (partitionDiff != 0) {
```
srowen: Probably very very slightly better to say

```
if (a._1 == b._1) {
  theKeyComp.compare(a._2, b._2)
} else {
  a._1 - b._1
}
```

srowen (changes requested):
Hm, I don't see why this would be slower than the original version. It should be nearly identical anyway or better, as it further inlines a few things. It could be some weird interactions with the JIT and benchmark or whatever, or maybe some difference in how it was tested.

Try one more round of changes here and benchmark again. In any event it would be worthwhile just for the code streamlining.

a-roberts:
Before progressing I've discussed what I'm seeing with our JIT compiler team, with the refactoring to reduce code duplication, the following occurs which solves some of the mystery -- although it's bad news as, like you, I wanted to remove the duplicate method.

Summarising:

> By having both of these classes share the getComparator method one level up in the hierarchy, the JIT profiling won't function as expected. 
> 
> Let's assume we have A (calling method) -> getComparator -> B (returns the comparator) and then you have C (another calling method) -> getComparator -> D (returns the comparator).
> 
> A and B are the actual methods calling getComparator. B and D are the comparators that are passed in.
> 
> As a JIT compiler if we profile getComparator on its I will see it calling B **and** calling D since profiling in most JITs is context insensitive and we think that's happening here. 
> 
> When we inline from A into getComparator and C into getComparator, I don't know if I should inline B or D, and given that inlining is critical for performance we see the slight drop in performance. Inlining is critical for eliminating call overheads, improving code locality, and the scope for optimisation.

srowen:
If that's true then again doesn't my suggestion to inline `partitionComparator` fix it?

a-roberts:
@srowen how about this for profiling?

```
private[spark] object WritablePartitionedPairCollection {
  /**
   * Takes an optional parameter (keyComparator), use if provided
   * and returns a comparator for the partitions
   */
  def getComparator[K](keyComparator: Option[Comparator[K]]): Comparator[(Int, K)] = {
    if (keyComparator.isDefined) {
      val theKeyComp = keyComparator.get
      new Comparator[(Int, K)] {
        // We know we have a non-empty comparator here
        override def compare(a: (Int, K), b: (Int, K)): Int = {
          if (a._1 == b._1) {
            theKeyComp.compare(a._2, b._2)
          } else {
            a._1 - b._1
          }
        }
      }
    } else return new Comparator[(Int, K)] {
      override def compare(a: (Int, K), b: (Int, K)): Int = {
        a._1 - b._1
      }
    }
  }
}
```

srowen:
Looks right except you just want to write

```
if (...) {
  ...
} else {
  new Comparator...
}
```

a-roberts:
Good point, done, I can get profiling the below code then? Builds fine and no scalastyle problems

```
  def getComparator[K](keyComparator: Option[Comparator[K]]): Comparator[(Int, K)] = {
    if (keyComparator.isDefined) {
      val theKeyComp = keyComparator.get
      new Comparator[(Int, K)] {
        // We know we have a non-empty comparator here
        override def compare(a: (Int, K), b: (Int, K)): Int = {
          if (a._1 == b._1) {
            theKeyComp.compare(a._2, b._2)
          } else {
            a._1 - b._1
          }
        }
      }
    } else {
      new Comparator[(Int, K)] {
        override def compare(a: (Int, K), b: (Int, K)): Int = {
          a._1 - b._1
        }
      }
    }
  }
```

mridulm:
(Particularly) as the number of partitions increase, "if (a._1 != b._1)" might be better for bpt reasons.

a-roberts:
I see, so doing the != comparison first which is likely to be true more of the time so we're not consistently failing this check then entering the else

```
  def getComparator[K](keyComparator: Option[Comparator[K]]): Comparator[(Int, K)] = {
    if (keyComparator.isDefined) {
      val theKeyComp = keyComparator.get
      new Comparator[(Int, K)] {
        // We know we have a non-empty comparator here
        override def compare(a: (Int, K), b: (Int, K)): Int = {
          if (a._1 != b._1) {
           a._1 - b._1
          } else {
           theKeyComp.compare(a._2, b._2)
          }
        }
      }
    } else {
      new Comparator[(Int, K)] {
        override def compare(a: (Int, K), b: (Int, K)): Int = {
          a._1 - b._1
        }
      }
    }
  }
```

Again that builds fine

srowen:
Hm why does the order matter - maybe helps branch prediction? I doubt we even know how the bytecode orders this let alone how it is JITted and whether it will gather branching info on this one branch. Either way. I usually prefer == for code clarity all else equal. No need to benchmark both just pick one.

a-roberts:
Passed on your question to our JIT developers

> The sense* of the test can have an impact of the VM interpreter performance, but that is not usually much of a component of actual throughput since important methods will be JIT'd very quickly regardless of which specific JVM you use. The J9 VM is capable of profiling the branch and flipping the sense when JITing the code
> 
> The sense refers to the way a code branches, so either down the equals branch or not equals branch

Numbers for us

**Refactored further as above**
```
ScalaSparkPagerank 2016-11-30 14:27:49 259928115            49.841               5215146
ScalaSparkPagerank 2016-11-30 14:29:52 259928115            51.310               5065837
ScalaSparkPagerank 2016-11-30 14:31:59 259928115            52.086               4990364
ScalaSparkPagerank 2016-11-30 14:34:05 259928115            50.667               5130126
ScalaSparkPagerank 2016-11-30 14:36:04 259928115            47.096               5519112
ScalaSparkPagerank 2016-11-30 14:38:04 259928115            48.244               5387781
ScalaSparkPagerank 2016-11-30 14:40:10 259928115            48.734               5333609
ScalaSparkPagerank 2016-11-30 14:42:12 259928115            49.295               5272910
397.273 / 8 = 49.659 sec average
```
**initial commit**
```
ScalaSparkPagerank 2016-11-30 14:48:01 259928115            46.442               5596832
ScalaSparkPagerank 2016-11-30 14:50:06 259928115            50.016               5196899
ScalaSparkPagerank 2016-11-30 14:52:12 259928115            51.113               5085362
ScalaSparkPagerank 2016-11-30 14:54:12 259928115            46.424               5599002
ScalaSparkPagerank 2016-11-30 14:56:15 259928115            47.604               5460215
ScalaSparkPagerank 2016-11-30 14:58:14 259928115            46.802               5553782
ScalaSparkPagerank 2016-11-30 15:00:16 259928115            47.021               5527915
ScalaSparkPagerank 2016-11-30 15:02:16 259928115            47.072               5521926
382.494 / 8 = 47.811s average
```

The first commit performs better on average, I'd like to next add the improved compare code as above and "push this down" into the subclasses to see how this performs

srowen:
I'd certainly be curious to see a benchmark of the 'final' version with inlined comparator. I would honestly be surprised if that's not fastest of all.

a-roberts:
New data for us, inlined comparator scores here (code provided below to check I've not profiled something useless!):
```
ScalaSparkPagerank 2016-12-05 13:44:41 259928115            48.149               5398411              5398411
ScalaSparkPagerank 2016-12-05 13:46:43 259928115            46.897               5542531              5542531
ScalaSparkPagerank 2016-12-05 13:48:46 259928115            49.130               5290619              5290619
ScalaSparkPagerank 2016-12-05 13:50:49 259928115            49.793               5220173              5220173
ScalaSparkPagerank 2016-12-05 13:52:50 259928115            48.061               5408296              5408296
ScalaSparkPagerank 2016-12-05 13:54:52 259928115            46.468               5593701              5593701
ScalaSparkPagerank 2016-12-05 13:56:56 259928115            51.385               5058443              5058443
ScalaSparkPagerank 2016-12-05 13:58:59 259928115            47.857               5431349              5431349
ScalaSparkPagerank 2016-12-05 14:00:59 259928115            46.515               5588049              5588049
ScalaSparkPagerank 2016-12-05 14:03:03 259928115            47.791               5438850              5438850
Avg 48.2046s
```

Remember our "vanilla" average time is 47.752s and our first commit averaged 47.229s (so not much of a difference really).

I think we're splitting hairs and I've got another PR I am seeing good results on that I plan to focus on instead: the SizeEstimator. 

This is what I've benchmarked, PartitionedAppendOnlyMap first, so let me know if there any further suggestions, otherwise I propose leaving this one for later as actually against the Spark master codebase I'm not noticing anything exciting.

```
  def partitionedDestructiveSortedIterator(keyComparator: Option[Comparator[K]])
    : Iterator[((Int, K), V)] = {
    val comparator = {
      if (keyComparator.isDefined) {
        val theKeyComp = keyComparator.get
        new Comparator[(Int, K)] {
          // We know we have a non-empty comparator here
          override def compare(a: (Int, K), b: (Int, K)): Int = {
            if (a._1 != b._1) {
              a._1 - b._1
            } else {
             theKeyComp.compare(a._2, b._2)
            }
          }
        }
      } else {
        new Comparator[(Int, K)] {
          override def compare(a: (Int, K), b: (Int, K)): Int = {
            a._1 - b._1
          }
        }
      }
    }
    destructiveSortedIterator(comparator)
  }
```

In PartitionedPairBuffer

```

  /** Iterate through the data in a given order. For this class this is not really destructive. */
  override def partitionedDestructiveSortedIterator(keyComparator: Option[Comparator[K]])
    : Iterator[((Int, K), V)] = {
    val comparator = {
      if (keyComparator.isDefined) {
        val theKeyComp = keyComparator.get
        new Comparator[(Int, K)] {
          // We know we have a non-empty comparator here
          override def compare(a: (Int, K), b: (Int, K)): Int = {
            if (a._1 != b._1) {
              a._1 - b._1
            } else {
             theKeyComp.compare(a._2, b._2)
            }
          }
        }
      } else {
        new Comparator[(Int, K)] {
          override def compare(a: (Int, K), b: (Int, K)): Int = {
            a._1 - b._1
          }
        }
      }
    }
    new Sorter(new KVArraySortDataFormat[(Int, K), AnyRef]).sort(data, 0, curSize, comparator)
    iterator
  }
```

WritablePartitionedPairCollection remains unchanged.

srowen:
It does seem like nice cleanup in any event. I am not sure why the first commit was faster as this seems like a 'superset' of optimization. We can't use that one in any event. If you want to update the PR with what you posted above, I think it'd be OK to commit just for the code simplification.

srowen:
@a-roberts let's either finish the thought and merge this as mostly a code cleanup and maybe marginal win, or just close it.

srowen:
Ping @a-roberts to resolve this

srowen:
I'm going to manually close this

PR #15737, opened 2016-11-02
[SPARK-18212][SS][KAFKA] increase executor poll timeout

koeninger:

Increase poll timeout to try and address flaky test

## How was this patch tested?

Ran existing unit tests

### DIFF
--- a/external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/KafkaSource.scala
+++ b/external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/KafkaSource.scala
@@ -90,7 +90,7 @@ private[kafka010] case class KafkaSource(
 
   private val pollTimeoutMs = sourceOptions.getOrElse(
     "kafkaConsumer.pollTimeoutMs",
-    (sc.conf.getTimeAsSeconds("spark.network.timeout", "120s") * 1000).toString
+    sc.conf.getTimeAsSeconds("spark.network.timeout", "120s").toString
   ).toLong
 
   private val maxOffsetFetchAttempts =
--- a/external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/KafkaSource.scala
+++ b/external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/KafkaSource.scala
@@ -90,7 +90,7 @@ private[kafka010] case class KafkaSource(
 
   private val pollTimeoutMs = sourceOptions.getOrElse(
     "kafkaConsumer.pollTimeoutMs",
-    sc.conf.getTimeAsSeconds("spark.network.timeout", "120s").toString
+    sc.conf.getTimeAsMs("spark.network.timeout", "120s").toString
   ).toLong
 
   private val maxOffsetFetchAttempts =
--- a/external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/KafkaSource.scala
+++ b/external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/KafkaSource.scala
@@ -88,7 +88,10 @@ private[kafka010] case class KafkaSource(
 
   private val sc = sqlContext.sparkContext
 
-  private val pollTimeoutMs = sourceOptions.getOrElse("kafkaConsumer.pollTimeoutMs", "512").toLong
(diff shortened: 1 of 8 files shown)


marmbrus:
Is there are reason to not change the user facing default?  Its failing our tests and it was failing for me for real with the current default.  When we are poling on the executors, the offsets should always be available, so I don't think increasing the timeout will hurt anything.

koeninger:
In most cases poll should be returning prefetched data from the buffer, not
waiting to talk to kafka over the network.

I could see increasing it a little bit, but I don't think it should be
increased to 10 seconds by default.  That's a really long time for a 1
second batch streaming job to silently wait, and the only thing you'll be
able to observe is that your job starts running slow for some unknown
reason.  At least with a smaller default any network or Kafka issues will
be obvious, and you can either diagnose them or tune upwards.

marmbrus:
If we have prefetched the timeout should never kick in (I hope?), so we probably only care about the network here.  If my choices are "fail the job" or "wait a little bit longer", I don't see why you would ever chose "fail the job" for transient network issues.  If its not a transient issue, do you really think that knowing about it 9.5 seconds earlier outweighs that cost of false negatives?

I guess I don't understand what kind of Kafka issues you are optimizing for?

marmbrus:
I don't think these cases are comparable, silently swallowing offset out of range is dropping data on the floor.  In contrast, networks _are_ transiently flakey and we should be robust to that here like we are in all of the other parts of Spark.  Specifically, we should probably use `spark.network.timeout` if they haven't overridden this particular setting (which defaults to 120s).

koeninger:
Ok, I'll update the default in both places to use spark.network.timeout and leave the test config at 10 seconds

external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/KafkaSource.scala
```diff
@@ -88,7 +88,10 @@ private[kafka010] case class KafkaSource(
 
   private val sc = sqlContext.sparkContext
 
-  private val pollTimeoutMs = sourceOptions.getOrElse("kafkaConsumer.pollTimeoutMs", "512").toLong
+  private val pollTimeoutMs = sourceOptions.getOrElse(
+    "kafkaConsumer.pollTimeoutMs",
+    (sc.conf.getTimeAsSeconds("spark.network.timeout", "120s") * 1000).toString
```
marmbrus: Why not `getTimeAsMs`?

koeninger: My thinking was just to make it really obvious it was the same default as spark.network.timeout usually has, since I didn't find a standalone constant somewhere for that value.  If you think it's clearer as ms, I don't feel strongly either way.

marmbrus: Sorry, I think I was unclear and it is really minor.  I agree we should keep it as `120s` for consistency.  I was only suggesting we use the built conversion done by `getTimeAsMs` rather than embedding the magic `* 1000` here.

```scala
sc.conf.getTimeAsMs("spark.network.timeout", "120s")
```

marmbrus:
Two comments:
 - I think `sc.conf.getTimeAsSeconds("spark.network.timeout", "120s").toString` is still wrong.  The suggestion is to make it the following:

```scala
val pollTimeoutMs = sourceOptions.getOrElse(
  "kafkaConsumer.pollTimeoutMs",
  sc.conf.getTimeAsMs("spark.network.timeout", "120s").toString).toLong
```
or maybe

```scala
val pollTimeoutMs = sourceOptions
  .get("kafkaConsumer.pollTimeoutMs")
  .map(_.toLong)
  .getOrElse(sc.conf.getTimeAsMs("spark.network.timeout", "120s"))
```

 - I'm not sure about the test changes.  Unless there is a good reason, I'd leave the tests alone and just use the default configuration there as well.

koeninger:
Good catch, I just mistakenly changed to AsMS in one place but not both.

On the test changes, do you want tests waiting up to 2 minutes * however many kafka calls are being made?  If so I can rip everything else out, just let me know.

marmbrus:
Yeah, I'd rather just keep it the same.  Failures here should be uncommon.

marmbrus:
Thanks, merging to master and 2.1!

PR #15738, opened 2016-11-02
SPARK-18086: Add support for Hive session vars.

rdblue:

This adds support for Hive variables:

* Makes values set via `spark-sql --hivevar name=value` accessible
* Adds `getHiveVar` and `setHiveVar` to the `HiveClient` interface
* Adds a SessionVariables trait for sessions like Hive that support variables (including Hive vars)
* Adds SessionVariables support to variable substitution
* Adds SessionVariables support to the SET command

## How was this patch tested?

* Adds a test to all supported Hive versions for accessing Hive variables
* Adds HiveVariableSubstitutionSuite

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/command/SetCommand.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/command/SetCommand.scala
@@ -60,6 +60,13 @@ case class SetCommand(kv: Option[(String, Option[String])]) extends RunnableComm
       }
       (keyValueOutput, runFunc)
 
+    case Some((key @ SetCommand.VariableName(name), Some(value))) =>
+      val runFunc = (sparkSession: SparkSession) => {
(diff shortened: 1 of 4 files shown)


rxin:
How are the session vars different from our configs? Can't we just use that?

rdblue:
The purpose of this is to ensure the same behavior as Hive so users can move to SparkSQL without rewriting queries. That's why I'm trying to keep behavior as close as possible to the behavior of Hive (and Spark 1.6.1).

Session vars are kept in a different namespace, so collisions should not overwrite configs and session vars take precedence in Hive. Because we still use a Hive SessionState and pass some commands to Hive, I thought it was also important to store the hive variables in the session state like Hive would. That way any command that are passed to Hive that have variable references work as well.

rxin:
Thanks - merging in master/branch-2.1!

rdblue:
Thanks for reviewing @rxin!

PR #15741, opened 2016-11-02
[SPARK-18200][GRAPHX] Support zero as an initial capacity in OpenHashSet

dongjoon-hyun:

[SPARK-18200](https://issues.apache.org/jira/browse/SPARK-18200) reports Apache Spark 2.x raises `java.lang.IllegalArgumentException: requirement failed: Invalid initial capacity` while running `triangleCount`. The root cause is that `VertexSet`, a type alias of `OpenHashSet`, does not allow zero as a initial size. This PR loosens the restriction to allow zero.

## How was this patch tested?

Pass the Jenkins test with a new test case in `OpenHashSetSuite`.

### DIFF
--- a/core/src/main/scala/org/apache/spark/util/collection/OpenHashSet.scala
+++ b/core/src/main/scala/org/apache/spark/util/collection/OpenHashSet.scala
@@ -42,15 +42,13 @@ import org.apache.spark.annotation.Private
  */
 @Private
 class OpenHashSet[@specialized(Long, Int) T: ClassTag](
-    var initialCapacity: Int,
+    initialCapacity: Int,
     loadFactor: Double)
   extends Serializable {
 
-  if (initialCapacity == 0) initialCapacity = 1
-
   require(initialCapacity <= OpenHashSet.MAX_CAPACITY,
     s"Can't make capacity bigger than ${OpenHashSet.MAX_CAPACITY} elements")
-  require(initialCapacity >= 1, "Invalid initial capacity")
+  require(initialCapacity >= 0, "Invalid initial capacity")
   require(loadFactor < 1.0, "Load factor must be less than 1.0")
   require(loadFactor > 0.0, "Load factor must be greater than 0.0")
 
@@ -273,7 +271,7 @@ class OpenHashSet[@specialized(Long, Int) T: ClassTag](
   private def hashcode(h: Int): Int = Hashing.murmur3_32().hashInt(h).asInt()
 
   private def nextPowerOf2(n: Int): Int = {
-    val highBit = Integer.highestOneBit(n)
+    val highBit = Integer.highestOneBit(if (n == 0) 1 else n)
     if (highBit == n) n else highBit << 1
   }
 }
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/apache/spark/util/collection/OpenHashSet.scala
```diff
@@ -42,10 +42,12 @@ import org.apache.spark.annotation.Private
  */
 @Private
 class OpenHashSet[@specialized(Long, Int) T: ClassTag](
-    initialCapacity: Int,
+    var initialCapacity: Int,
```
rxin: This looks really hacky and makes it a public variable. Why don't we just change the require to >= 0, and then change nextPowerOf2 to pick a value that's greater than 0?

dongjoon-hyun: Yep. I'll change like that.

core/src/main/scala/org/apache/spark/util/collection/OpenHashSet.scala
```diff
@@ -271,7 +271,7 @@ class OpenHashSet[@specialized(Long, Int) T: ClassTag](
   private def hashcode(h: Int): Int = Hashing.murmur3_32().hashInt(h).asInt()
 
   private def nextPowerOf2(n: Int): Int = {
-    val highBit = Integer.highestOneBit(n)
+    val highBit = Integer.highestOneBit(if (n == 0) 1 else n)
```
rxin: can you just do ...

```
if (n == 0) {
  2
} else {
  // original code
}
```

dongjoon-hyun: Sure.
BTW, for the test failure, I'll look at that soon.

dongjoon-hyun:
The only failure seems to be irrelevant.
```
[info] BasicSchedulerIntegrationSuite:
[info] - job with fetch failure *** FAILED *** (1 second, 889 milliseconds)
[info]   java.util.concurrent.TimeoutException: Futures timed out after [1 second]
```

dongjoon-hyun:
Retest this please.

rxin:
LGTM pending Jenkins.

dongjoon-hyun:
Thank you for review and merging, @rxin .

core/src/main/scala/org/apache/spark/util/collection/OpenHashSet.scala
```diff
@@ -271,8 +271,12 @@ class OpenHashSet[@specialized(Long, Int) T: ClassTag](
   private def hashcode(h: Int): Int = Hashing.murmur3_32().hashInt(h).asInt()
 
   private def nextPowerOf2(n: Int): Int = {
-    val highBit = Integer.highestOneBit(n)
-    if (highBit == n) n else highBit << 1
+    if (n == 0) {
+      2
```
mridulm: This should have been 1 perhaps ?
nextPowerOf2(1) < nextPowerOf2(0) now ...

srowen: True or just make the condition <= 1. It won't really matter much but is a little inconsistent.

srowen: No I think the issue was that right now nextPowerOf2(1) is 1 and that seems inconsistent if 0 gives 2. Both should result in 2?

dongjoon-hyun: Then, should I fix this like the following?
```
-      2
+      1
```

PR #15742, opened 2016-11-02
[SPARK-16808][Core] History Server main page does not honor APPLICATION_WEB_PROXY_BASE

vijoshi:

Application links generated on the history server UI no longer (regression from 1.6) contain the configured spark.ui.proxyBase in the links. To address this, made the uiRoot available globally to all javascripts for Web UI. Updated the mustache template (historypage-template.html) to include the uiroot for rendering links to the applications. 

The existing test was not sufficient to verify the scenario where ajax call is used to populate the application listing template, so added a new selenium test case to cover this scenario.

## How was this patch tested?

Existing tests and a new unit test. 
No visual changes to the UI.

### DIFF
--- a/core/src/main/scala/org/apache/spark/deploy/history/HistoryServer.scala
+++ b/core/src/main/scala/org/apache/spark/deploy/history/HistoryServer.scala
@@ -143,6 +143,12 @@ class HistoryServer(
     appCache.stop()
   }
 
+  // For testing - override stop timeout used by jetty
+  private[history] def setStopTimeout(timeout: Long): Unit = {
+    assert(serverInfo.isDefined, "HistoryServer must be bound before setting stop timeout")
+    serverInfo.get.server.setStopTimeout(timeout)
+  }
+
   /** Attach a reconstructed UI to this server. Only valid after bind(). */
   override def attachSparkUI(
       appId: String,
--- a/core/src/main/scala/org/apache/spark/deploy/history/HistoryServer.scala
+++ b/core/src/main/scala/org/apache/spark/deploy/history/HistoryServer.scala
@@ -143,12 +143,6 @@ class HistoryServer(
     appCache.stop()
   }
 
-  // For testing - override stop timeout used by jetty
-  private[history] def setStopTimeout(timeout: Long): Unit = {
-    assert(serverInfo.isDefined, "HistoryServer must be bound before setting stop timeout")
-    serverInfo.get.server.setStopTimeout(timeout)
-  }
-
   /** Attach a reconstructed UI to this server. Only valid after bind(). */
   override def attachSparkUI(
       appId: String,
--- a/core/src/test/scala/org/apache/spark/deploy/history/HistoryServerSuite.scala
+++ b/core/src/test/scala/org/apache/spark/deploy/history/HistoryServerSuite.scala
@@ -310,7 +310,7 @@ class HistoryServerSuite extends SparkFunSuite with BeforeAndAfter with Matchers
         }
 
         val proxyidx = sb.indexOf(uiRoot)
-        sb.delete(proxyidx, proxyidx+uiRoot.length).toString
+        sb.delete(proxyidx, proxyidx + uiRoot.length).toString
       }
     }
 
--- a/core/src/test/scala/org/apache/spark/deploy/history/HistoryServerSuite.scala
+++ b/core/src/test/scala/org/apache/spark/deploy/history/HistoryServerSuite.scala
@@ -321,29 +321,32 @@ class HistoryServerSuite extends SparkFunSuite with BeforeAndAfter with Matchers
       getWebClient.getOptions.setThrowExceptionOnScriptError(false)
     }
 
-    val url = s"http://localhost:$port"
+    try {
+      val url = s"http://localhost:$port"
 
-    go to s"$url$uiRoot"
+      go to s"$url$uiRoot"
 
-    // expect the ajax call to finish in 5 seconds
-    implicitlyWait(org.scalatest.time.Span(5, org.scalatest.time.Seconds))
+      // expect the ajax call to finish in 5 seconds
+      implicitlyWait(org.scalatest.time.Span(5, org.scalatest.time.Seconds))
 
-    // once this findAll call returns, we know the ajax load of the table completed
-    findAll(ClassNameQuery("odd"))
+      // once this findAll call returns, we know the ajax load of the table completed
+      findAll(ClassNameQuery("odd"))
 
-    val links = findAll(TagNameQuery("a"))
-      .map(_.attribute("href"))
-      .filter(_.isDefined)
-      .map(_.get)
-      .filter(_.startsWith(url)).toList
+      val links = findAll(TagNameQuery("a"))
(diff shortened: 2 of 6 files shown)


vijoshi:
@ajbozarth @srowen - for your review if possible.

ajbozarth:
Nice catch, from a quick look this seems good. My issue is one of consistency, in previous changes I've made I've used a setter rather that declaring a var on the scala side. I think using a default val and a setter is safer than checking for undefined. An example can be seen at the top of `historyserver.js`

vijoshi:
@ajbozarth updated the code. Guess, there is agreement to keep `uiRoot` in `webui.js` to make it available to any more templates / javascript that need it. I removed the `undefined` check for `uiRoot` as that in a way makes the expectation that `uiRoot` is available explicit. In case it isn't, the testcases would fail which would be better than defaulting to a value that may not be correct.

ajbozarth:
LGTM, putting the var and setter in `webui.js` is a good idea

core/src/test/scala/org/apache/spark/deploy/history/HistoryServerSuite.scala
```diff
@@ -275,6 +277,79 @@ class HistoryServerSuite extends SparkFunSuite with BeforeAndAfter with Matchers
     all (siteRelativeLinks) should startWith (uiRoot)
   }
 
+  test("ajax rendered relative links are prefixed with uiRoot (spark.ui.proxyBase)") {
+
+    val proxyBaseBeforeTest = System.getProperty("spark.ui.proxyBase")
+    val uiRoot = "/testwebproxybase"
+    System.setProperty("spark.ui.proxyBase", uiRoot)
```
vanzin: Do you need to use system properties for this? Isn't it enough to set it in `SparkConf`?

If you need to use system properties, you should use `ResetSystemProperties` since if your test fails, it will leave the modified system properties behind.

core/src/test/scala/org/apache/spark/deploy/history/HistoryServerSuite.scala
```diff
@@ -275,6 +277,79 @@ class HistoryServerSuite extends SparkFunSuite with BeforeAndAfter with Matchers
...
+
+    val url = s"http://localhost:$port"
+
+    go to s"$url$uiRoot"
+
+    // expect the ajax call to finish in 5 seconds
+    implicitlyWait(org.scalatest.time.Span(5, org.scalatest.time.Seconds))
+
+    // once this findAll call returns, we know the ajax load of the table completed
+    findAll (ClassNameQuery("odd"))
```
vanzin: nit: no space before `(` in method calls , also happens below.

vanzin (commented):
Couple of comments on the test, otherwise LGTM.

core/src/test/scala/org/apache/spark/deploy/history/HistoryServerSuite.scala
vijoshi: fixed.

core/src/test/scala/org/apache/spark/deploy/history/HistoryServerSuite.scala
vijoshi: I do need to use system properties as that's how this particular property seems to work. The test suite `HistoryServerSuite` already uses `ResetSystemProperties`, so I have fixed the tests to not try to store/reset system properties on their own.

vijoshi:
@vanzin @ajbozarth besides your review comments, I noticed that the test case was getting blocked in the `after { }` where the history server is stopped. Tracked it to Jetty waiting for 30 secs for graceful killing of open connections. Not needed during testing, so added a change for that.

vijoshi:
retest this please

core/src/main/scala/org/apache/spark/deploy/history/HistoryServer.scala
```diff
@@ -143,6 +143,12 @@ class HistoryServer(
     appCache.stop()
   }
 
+  // For testing - override stop timeout used by jetty
+  private[history] def setStopTimeout(timeout: Long): Unit = {
```
ajbozarth: I think that adding a function just to speed up tests by 30sec is unnecessary additional code in this case

vijoshi:
retest this please

core/src/main/scala/org/apache/spark/deploy/history/HistoryServer.scala
vijoshi: this was 30 secs additional for just the one test i added. if more such tests are added the time would add up - so i felt the option to bypass the wait was useful to have. assuming more of the UI could transition to ajax etc I could move this utility function up into the base `WebUI` class so it's available more generally for future UI tests - not just history server?

vanzin: I think this is a little fishy, because other tests run Jetty and don't seem to suffer from this problem (or at least no one has complained).

Maybe, if you explicitly close the selenium driver in the test, this will go away without the need for this method?

vijoshi: This looks to be due to the ProxyServlet handler, the way it handles requests. I am able to get around the problem by explicitly stopping the servlet handler before the test ends. No need to alter Jetty stop timeout with this change. Updated the pull request.

core/src/test/scala/org/apache/spark/deploy/history/HistoryServerSuite.scala
```diff
@@ -258,23 +260,92 @@ class HistoryServerSuite extends SparkFunSuite with BeforeAndAfter with Matchers
...
+    // once this findAll call returns, we know the ajax load of the table completed
+    findAll(ClassNameQuery("odd"))
+
+    val links = findAll(TagNameQuery("a"))
+      .map(_.attribute("href"))
+      .filter(_.isDefined)
+      .map(_.get)
+      .filter(_.startsWith(url)).toList
+
+    contextHandler.stop()
```
vanzin: Do you need to do this in a `finally` block? And do you also need to call `server.stop()`?

ajbozarth:
LGTM

core/src/test/scala/org/apache/spark/deploy/history/HistoryServerSuite.scala
vijoshi: Okay, moved it to a `finally` block. This should make both good and bad testcase runs consistent in their cleanup.

vanzin:
LGTM pending tests.

vanzin:
retest this please

vanzin:
Merging to master / 2.1.

mariobriggs:
@vanzin since this a regression bux in 2.0, any particular reason it is merged only to 2.1 . I believe this should be in 2.0.2/3 as well

vanzin:
It doesn't merge cleanly into branch-2.0. You can file a separate PR if you want it fixed there.

mariobriggs:
ok. will look into that

vijoshi:
@vanzin @mariobriggs Opened https://github.com/apache/spark/pull/15855 for merging to branch-2.0

PR #11751, opened 2016-03-16
[SPARK-13920][BUILD] MIMA checks should apply to @Experimental and @DeveloperAPI APIs

dongjoon-hyun:
## What changes were proposed in this pull request?

We are able to change `@Experimental` and `@DeveloperAPI` API freely but also should monitor and manage those API carefully. This PR for [SPARK-13920](https://issues.apache.org/jira/browse/SPARK-13920) enables MiMa check and adds filters for them.
## How was this patch tested?

Pass the Jenkins tests (including MiMa).

PR #15743, opened 2016-11-03
[SPARK-18236] Reduce duplicate objects in Spark UI and HistoryServer

JoshRosen:

When profiling heap dumps from the HistoryServer and live Spark web UIs, I found a large amount of memory being wasted on duplicated objects and strings. This patch's changes remove most of this duplication, resulting in over 40% memory savings for some benchmarks.

- **Task metrics** (6441f0624dfcda9c7193a64bfb416a145b5aabdf): previously, every `TaskUIData` object would have its own instances of `InputMetricsUIData`, `OutputMetricsUIData`, `ShuffleReadMetrics`, and `ShuffleWriteMetrics`, but for many tasks these metrics are irrelevant because they're all zero. This patch changes how we construct these metrics in order to re-use a single immutable "empty" value for the cases where these metrics are empty.
- **TaskInfo.accumulables** (ade86db901127bf13c0e0bdc3f09c933a093bb76): Previously, every `TaskInfo` object had its own empty `ListBuffer` for holding updates from named accumulators. Tasks which didn't use named accumulators still paid for the cost of allocating and storing this empty buffer. To avoid this overhead, I changed the `val` with a mutable buffer into a `var` which holds an immutable Scala list, allowing tasks which do not have named accumulator updates to share the same singleton `Nil` object.
- **String.intern() in JSONProtocol** (7e05630e9a78c455db8c8c499f0590c864624e05): in the HistoryServer, executor hostnames and ids are deserialized from JSON, leading to massive duplication of these string objects. By calling `String.intern()` on the deserialized values we can remove all of this duplication. Since Spark now requires Java 7+ we don't have to worry about string interning exhausting the permgen (see http://java-performance.info/string-intern-in-java-6-7-8/).

## How was this patch tested?

I ran 

```
sc.parallelize(1 to 100000, 100000).count()
```

in `spark-shell` with event logging enabled, then loaded that event log in the HistoryServer, performed a full GC, and took a heap dump. According to YourKit, the changes in this patch reduced memory consumption by roughly 28 megabytes (or 770k Java objects):

image

Here's a table illustrating the drop in objects due to deduplication (the drop is <100k for some objects because some events were dropped from the listener bus; this is a separate, existing bug that I'll address separately after CPU-profiling):

image

### DIFF
--- a/core/src/main/scala/org/apache/spark/scheduler/DAGScheduler.scala
+++ b/core/src/main/scala/org/apache/spark/scheduler/DAGScheduler.scala
@@ -1089,7 +1089,7 @@ class DAGScheduler(
         // To avoid UI cruft, ignore cases where value wasn't updated
         if (acc.name.isDefined && !updates.isZero) {
           stage.latestInfo.accumulables(id) = acc.toInfo(None, Some(acc.value))
-          event.taskInfo.accumulables ::= acc.toInfo(Some(updates.value), Some(acc.value))
+          event.taskInfo._accumulables ::= acc.toInfo(Some(updates.value), Some(acc.value))
         }
       }
     } catch {
--- a/core/src/main/scala/org/apache/spark/scheduler/DAGScheduler.scala
+++ b/core/src/main/scala/org/apache/spark/scheduler/DAGScheduler.scala
@@ -1089,7 +1089,8 @@ class DAGScheduler(
         // To avoid UI cruft, ignore cases where value wasn't updated
         if (acc.name.isDefined && !updates.isZero) {
           stage.latestInfo.accumulables(id) = acc.toInfo(None, Some(acc.value))
-          event.taskInfo.accumulables ::= acc.toInfo(Some(updates.value), Some(acc.value))
+          event.taskInfo.setAccumulables(
+            acc.toInfo(Some(updates.value), Some(acc.value)) +: event.taskInfo.accumulables)
         }
       }
     } catch {
--- a/core/src/main/scala/org/apache/spark/scheduler/DAGScheduler.scala
+++ b/core/src/main/scala/org/apache/spark/scheduler/DAGScheduler.scala
@@ -1089,7 +1089,7 @@ class DAGScheduler(
         // To avoid UI cruft, ignore cases where value wasn't updated
         if (acc.name.isDefined && !updates.isZero) {
           stage.latestInfo.accumulables(id) = acc.toInfo(None, Some(acc.value))
-          event.taskInfo._accumulables ::= acc.toInfo(Some(updates.value), Some(acc.value))
+          event.taskInfo.accumulables ::= acc.toInfo(Some(updates.value), Some(acc.value))
         }
       }
     } catch {
--- a/core/src/main/scala/org/apache/spark/scheduler/DAGScheduler.scala
+++ b/core/src/main/scala/org/apache/spark/scheduler/DAGScheduler.scala
@@ -1089,7 +1089,7 @@ class DAGScheduler(
         // To avoid UI cruft, ignore cases where value wasn't updated
         if (acc.name.isDefined && !updates.isZero) {
           stage.latestInfo.accumulables(id) = acc.toInfo(None, Some(acc.value))
-          event.taskInfo.accumulables += acc.toInfo(Some(updates.value), Some(acc.value))
+          event.taskInfo.accumulables ::= acc.toInfo(Some(updates.value), Some(acc.value))
         }
       }
     } catch {
--- a/core/src/main/scala/org/apache/spark/scheduler/TaskInfo.scala
+++ b/core/src/main/scala/org/apache/spark/scheduler/TaskInfo.scala
@@ -17,6 +17,8 @@
 
 package org.apache.spark.scheduler
 
+import scala.collection.mutable.ListBuffer
+
 import org.apache.spark.TaskState
 import org.apache.spark.TaskState.TaskState
 import org.apache.spark.annotation.DeveloperApi
@@ -52,7 +54,12 @@ class TaskInfo(
    * accumulable to be updated multiple times in a single task or for two accumulables with the
    * same name but different IDs to exist in a task.
    */
-  var accumulables: List[AccumulableInfo] = Nil
+  lazy val accumulables: ListBuffer[AccumulableInfo] = ListBuffer(_accumulables: _*)
+
+  // Note: all internal uses should use _accumulables; the above `accumulables` is only exposed
+  // for backwards-compatibility purposes for non-Spark-users of this class.
+  // See SPARK-18236 for discussion.
+  private[spark] var _accumulables: List[AccumulableInfo] = Nil
 
   /**
    * The time when the task has completed successfully (including the time to remotely fetch
--- a/core/src/main/scala/org/apache/spark/scheduler/TaskInfo.scala
+++ b/core/src/main/scala/org/apache/spark/scheduler/TaskInfo.scala
@@ -17,8 +17,6 @@
 
 package org.apache.spark.scheduler
 
-import scala.collection.mutable.ListBuffer
-
 import org.apache.spark.TaskState
 import org.apache.spark.TaskState.TaskState
 import org.apache.spark.annotation.DeveloperApi
@@ -54,12 +52,7 @@ class TaskInfo(
    * accumulable to be updated multiple times in a single task or for two accumulables with the
    * same name but different IDs to exist in a task.
    */
-  lazy val accumulables: ListBuffer[AccumulableInfo] = ListBuffer(_accumulables: _*)
-
-  // Note: all internal uses should use _accumulables; the above `accumulables` is only exposed
-  // for backwards-compatibility purposes for non-Spark-users of this class.
-  // See SPARK-18236 for discussion.
-  private[spark] var _accumulables: List[AccumulableInfo] = Nil
+  var accumulables: List[AccumulableInfo] = Nil
 
   /**
    * The time when the task has completed successfully (including the time to remotely fetch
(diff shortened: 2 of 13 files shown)


srowen:
Aside: if this and other processes see a lot of memory used by duplicate strings then maybe we should recommend trying -XX+UseStringDeduplication for a related and automatic win.

mridulm:
+1 on UseStringDeduplication ! Though jdk8 specific ....

core/src/main/scala/org/apache/spark/util/JsonProtocol.scala
```diff
@@ -885,8 +885,8 @@ private[spark] object JsonProtocol {
     if (json == JNothing) {
       return null
     }
-    val executorId = (json \ "Executor ID").extract[String]
-    val host = (json \ "Host").extract[String]
+    val executorId = (json \ "Executor ID").extract[String].intern()
+    val host = (json \ "Host").extract[String].intern()
```
mridulm: In general intern'ing can be dangerous.
I dont expect issues with host, but executorId (for long running jobs) can essentially OOM the driver.

mridulm: scratch that, jdk7 improvements help.

core/src/main/scala/org/apache/spark/scheduler/DAGScheduler.scala
```diff
@@ -1089,7 +1089,7 @@ class DAGScheduler(
         // To avoid UI cruft, ignore cases where value wasn't updated
         if (acc.name.isDefined && !updates.isZero) {
           stage.latestInfo.accumulables(id) = acc.toInfo(None, Some(acc.value))
-          event.taskInfo.accumulables += acc.toInfo(Some(updates.value), Some(acc.value))
+          event.taskInfo.accumulables ::= acc.toInfo(Some(updates.value), Some(acc.value))
```
mridulm: This should be equivalent, no ?
ListBuffer.+= does essentially :: on last

mridulm: Ok, type change ...

core/src/main/scala/org/apache/spark/ui/jobs/UIData.scala
```diff
@@ -197,8 +196,32 @@ private[spark] object UIData {
...
+    def apply(metrics: InputMetrics): InputMetricsUIData = {
+      if (metrics.bytesRead == 0 && metrics.recordsRead == 0) {
+        EMPTY
+      } else {
+        new InputMetricsUIData(
+          bytesRead = metrics.bytesRead,
+          recordsRead = metrics.recordsRead)
+      }
+    }
+    val EMPTY = InputMetricsUIData(0, 0)
```
mridulm: private ?

mridulm:
Looks good to me, some minor queries and comments.

core/src/main/scala/org/apache/spark/scheduler/DAGScheduler.scala
JoshRosen: Something I was a bit unsure of but want to revisit now:

It looks like the loop 

```
 for (accumulableInfo <- info.accumulables) {
        stageData.accumulables(accumulableInfo.id) = accumulableInfo
      }
```

in `JobProgressListener.onTaskEnd()` may assume that accumulator updates are appended to the list so that the newest update for a particular accumulator appears last, but _prepending_ here will change that. I don't think that this behavior is covered by any existing unit tests, though. Instead of using a list here, I might want to use an immutable `Queue`. Let me go ahead and make this change now.

core/src/main/scala/org/apache/spark/ui/jobs/UIData.scala
```diff
@@ -197,8 +196,32 @@ private[spark] object UIData {
...
+  }
 
   case class OutputMetricsUIData(bytesWritten: Long, recordsWritten: Long)
+  object OutputMetricsUIData {
+    def apply(metrics: OutputMetrics): OutputMetricsUIData = {
+      if (metrics.bytesWritten == 0 && metrics.recordsWritten == 0) {
```
JoshRosen: For OutputMetrics, I'd actually assume the opposite: these metrics are referring to bytes written to an external system, not bytes written to shuffle, so the majority of tasks won't have non-zero values for this metric (all but the last stage in a multi-stage job, for example).

core/src/main/scala/org/apache/spark/ui/jobs/UIData.scala
JoshRosen: Done.

core/src/main/scala/org/apache/spark/scheduler/DAGScheduler.scala
JoshRosen: On the other hand, code on the `StagePage` uses `accumulables` like this:

```
    val externalAccumulableReadable = info.accumulables
      .filterNot(_.internal)
      .flatMap { a =>
        (a.name, a.update) match {
          case (Some(name), Some(update)) => Some(StringEscapeUtils.escapeHtml4(s"$name: $update"))
          case _ => None
        }
      }
```

This seems to be assuming that you'll only have one update per task.

JoshRosen:
@srowen, that's a great tip about `UseStringDeduplication`; I'll give this a try in my deployment.

JoshRosen:
Jenkins, retest this please

core/src/main/scala/org/apache/spark/scheduler/TaskInfo.scala
```diff
@@ -54,7 +52,7 @@ class TaskInfo(
    * accumulable to be updated multiple times in a single task or for two accumulables with the
    * same name but different IDs to exist in a task.
    */
-  val accumulables = ListBuffer[AccumulableInfo]()
```
JoshRosen: In order to maintain binary compatibility, I could rewrite this to be a `lazy val` that returns a `ListBuffer` formed from the "real" `accumulables` which can remain private. I might go ahead and do that just to avoid any chance of incompatibility-related problems, although I don't anticipate this being an issue in practice.

JoshRosen:
Jenkins, retest this please

ajbozarth:
Looking through the code this looks good to me, but I can only talk to the code for the first two bullets, I don't know enough about the intern function to comment on its use. This will give us some great speed ups to go with our other recent SHS updates.

core/src/main/scala/org/apache/spark/scheduler/TaskInfo.scala
```diff
@@ -54,7 +54,12 @@ class TaskInfo(
...
+  lazy val accumulables: ListBuffer[AccumulableInfo] = ListBuffer(_accumulables: _*)
```
vanzin: Isn't a `def` here the right thing? Otherwise, if `_accumulables` changes after this lazy val is evaluated, things will get out of sync.

Or given how the semantics here now are getting pretty weird, maybe changing it to an immutable list and just breaking compatibility would be better. Then you can just return `_accumulables`.

vanzin (approved):

core/src/main/scala/org/apache/spark/scheduler/TaskInfo.scala
JoshRosen: Meh, kind of a toss-up. My original patch broke compatibility and I did this in 4c867f10ff5d8d80c7e8d44b4702785c0a017b64 to restore binary compatibility, but it risks changing semantics. I think I'll just revert 4c867f10ff5d8d80c7e8d44b4702785c0a017b64 and add the minor compatibility break to a lesser-used `@DeveloperAPI`.

JoshRosen: Actually, though, if we're going to expose a public API for this at all then I think it should be read-only. Let me go ahead and refactor this to expose a read-only `Seq`.

JoshRosen:
Alright, just pushed an update which re-introduces the binary compatibility break but chooses to expose a nicer + narrower public interface.

project/MimaExcludes.scala
```diff
@@ -86,7 +86,10 @@ object MimaExcludes {
       // [SPARK-18034] Upgrade to MiMa 0.1.11 to fix flakiness.
       ProblemFilters.exclude[InheritedNewAbstractMethodProblem]("org.apache.spark.ml.param.shared.HasAggregationDepth.aggregationDepth"),
       ProblemFilters.exclude[InheritedNewAbstractMethodProblem]("org.apache.spark.ml.param.shared.HasAggregationDepth.getAggregationDepth"),
-      ProblemFilters.exclude[InheritedNewAbstractMethodProblem]("org.apache.spark.ml.param.shared.HasAggregationDepth.org$apache$spark$ml$param$shared$HasAggregationDepth$_setter_$aggregationDepth_=")
+      ProblemFilters.exclude[InheritedNewAbstractMethodProblem]("org.apache.spark.ml.param.shared.HasAggregationDepth.org$apache$spark$ml$param$shared$HasAggregationDepth$_setter_$aggregationDepth_="),
+
+      // [SPARK-18236] Reduce duplicate objects in Spark UI and HistoryServer
+      ProblemFilters.exclude[IncompatibleResultTypeProblem]("org.apache.spark.scheduler.TaskInfo.accumulables")
```
vanzin: Hmm, I guess mima still fails to ignore `@DeveloperApi` changes?

vanzin (approved):
LGTM.

core/src/main/scala/org/apache/spark/ui/jobs/UIData.scala
mridulm: sounds good

project/MimaExcludes.scala
JoshRosen: Yep, that's intentional; we removed the logic for excluding that annotation in #11751. The rationale is discussed in https://issues.apache.org/jira/browse/SPARK-13920:

> Our MIMA binary compatibility checks currently ignore APIs which are marked as `@Experimental`or `@DeveloperApi`, but I don't think this makes sense. Even if those annotations _reserve_ the right to break binary compatibility, we should still avoid compatibility breaks whenever possible and should be informed explicitly when compatibility breaks.

mridulm:
LGTM

ajbozarth:
LGTM

JoshRosen:
Thanks for reviewing! I'm going to merge this to master.

PR #15744, opened 2016-11-03
[SPARK-18237][HIVE] hive.exec.stagingdir have no effect

classjtt:
hive.exec.stagingdir have no effect in spark2.0.1，
Hive confs in hive-site.xml will be loaded in `hadoopConf`, so we should use `hadoopConf` in `InsertIntoHiveTable` instead of `SessionState.conf`


rxin:
LGTM pending Jenkins.

But note that we might remove the idea of staging directory for writes completely in the future, when we do https://issues.apache.org/jira/browse/SPARK-15691

rxin:
Merging in master/branch-2.1.

rxin:
BTW you might want to add the email you used with the commit (fuxing@wacai.com) to your github account so the commit shows up as yours.

PR #15748, opened 2016-11-03
[SPARK-18240][ML][PySpark] Add Summary of BiKMeans and GMM in pyspark

zhengruifeng:
1, create class `ClusteringSummary`
2, add summary of BiKMeans and GMM by extends `ClusteringSummary`
3, remove two unused imports

## How was this patch tested?
added tests in clustering.py and local tests in pyspark

### DIFF
--- a/python/pyspark/ml/clustering.py
+++ b/python/pyspark/ml/clustering.py
@@ -133,7 +133,7 @@ class GaussianMixtureModel(JavaModel, JavaMLWritable, JavaMLReadable):
 @inherit_doc
 class GaussianMixtureSummary(ClusteringSummary):
     """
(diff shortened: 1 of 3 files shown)


zhengruifeng:
@yanboliang @jkbradley Could you please make a review?

jkbradley:
I do want to add this at some point, though we need to focus on QA for now

zhengruifeng:
@jkbradley  There is somewhat misunderstanding. I close this PR because there is a similar one. Sethah created SPARK-18282 just one day after this one. And people reviewed that one. So I think it's better to close this PR, and Sethah take it over. I will also help to reveiw that PR and QA. ☺

PR #15751, opened 2016-11-03
[SPARK-18246][SQL] Throws an exception before execution for unsupported types in Json, CSV and text functionailities

HyukjinKwon:

This PR includes several fixes as below:

**Case 1 `read.json(rdd)`** - throws an exception before the execution for unsupported types.

```scala
val rdd = spark.sparkContext.parallelize(1 to 100).map(i => s"""{"a": "str$i"}""")
val schema = new StructType().add("a", CalendarIntervalType)
spark.read.schema(schema).option("mode", "FAILFAST").json(rdd).show()
```

**Case 2 `read.json(path)`** - throws an exception before the execution for unsupported types.

```scala
val path = "/tmp/aa"
val rdd = spark.sparkContext.parallelize(1 to 100).map(i => s"""{"a": "str$i"}""").saveAsTextFile(path)
val schema = new StructType().add("a", CalendarIntervalType)
spark.read.schema(schema).option("mode", "FAILFAST").json(path).show()
```

**Case 3 `read.csv(path)`** - throws an exception before the execution for unsupported types.

```scala
val path = "/tmp/bb"
val rdd = spark.sparkContext.parallelize(1 to 100).saveAsTextFile(path)
val schema = new StructType().add("a", CalendarIntervalType)
spark.read.schema(schema).option("mode", "FAILFAST").csv(path).show()
```

**Case 4 `read.text(path)`** -  throws an exception before the execution for unsupported types  rather than printing incorrect values.

```scala
val path = "/tmp/cc"
val rdd = spark.sparkContext.parallelize(1 to 100).saveAsTextFile(path)
val schema = new StructType().add("a", LongType)
spark.read.schema(schema).text(path).show()
```

currently this prints as below:

```
+-----------+
|          a|
+-----------+
|68719476738|
|68719476738|
|68719476738|
...
```

whereas actual content is

```
1
2
3
...
```

**Case 5 `from_json(...)`** - throws analysis exception for unsupported types (`to_json` is already throwing an analysis exception).

```scala
import org.apache.spark.sql.functions._
import spark.implicits._

val df = Seq("""{"a" 1}""").toDS()
val schema = new StructType().add("a", CalendarIntervalType)
df.select(from_json($"value", schema)).collect()
```

**Case 6 `write.json(path)` _(this is a potential issue)_** - adds the schema check in writing

Currently, it seems JSON conversion does not support only `CalendarIntervalType` but this case seems already covered as below:

```scala
sql("SELECT interval 1 seconds as a").write.json("tmp/123")
```

```
Cannot save interval data type into external storage.;
org.apache.spark.sql.AnalysisException: Cannot save interval data type into external storage.;
	at org.apache.spark.sql.execution.datasources.DataSource.write(DataSource.scala:462)
	at org.apache.spark.sql.DataFrameWriter.save(DataFrameWriter.scala:215)
	at org.apache.spark.sql.DataFrameWriter.save(DataFrameWriter.scala:198)
```

However, it'd be safe if we add this check like the other text-based datasources. We might add some more types in the future.

In more details for **Case 1**, **Case 2** and **Case 3**, there are parsing modes, `FAILFAST`, `DROPMALFORMED` and `PERMISSIVE`. The original behaviour is,

  - `FAILFAST` - fails if it meets the unsupported type when parsing

  - `DROPMALFORMED` -  drops record having non-null values in the unsupported types. Otherwise, it reads it as `null`.

  - `PERMISSIVE` -  allows to read the values as null for unsupported types.

In case of `FAILFAST`, we can fail right after only checking the schema.

## How was this patch tested?

Unit tests in `JsonSuite`, `CSVSuite`, `TextSuite` and `JsonFunctionsSuite`.

### DIFF
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/jsonExpressions.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/jsonExpressions.scala
@@ -496,6 +496,21 @@ case class JsonToStruct(schema: StructType, options: Map[String, String], child:
 
   override def dataType: DataType = schema
 
+  override def checkInputDataTypes(): TypeCheckResult = {
+    if (StringType.acceptsType(child.dataType)) {
+      try {
+        JacksonUtils.verifySchema(schema)
+        TypeCheckResult.TypeCheckSuccess
+      } catch {
+        case e: UnsupportedOperationException =>
+          TypeCheckResult.TypeCheckFailure(e.getMessage)
+      }
+    } else {
+      TypeCheckResult.TypeCheckFailure(
+        s"$prettyName requires that the expression is a string expression.")
+    }
+  }
+
   override def nullSafeEval(json: Any): Any = {
     try parser.parse(json.toString).head catch {
       case _: SparkSQLJsonProcessingException => null
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/csv/CSVFileFormat.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/csv/CSVFileFormat.scala
@@ -143,6 +143,15 @@ class CSVFileFormat extends TextBasedFileFormat with DataSourceRegister {
     val broadcastedHadoopConf =
       sparkSession.sparkContext.broadcast(new SerializableConfiguration(hadoopConf))
 
+    if (csvOptions.failFast) {
+      // We can fail before starting to parse in cast of "FAILFAST" mode. In case of "PERMISIVE"
+      // mode, allows to read values as null for unsupported types. In case of "DROPMALFORMED"
+      // mode, drops records only containing non-null values in unsupported types. We should use
+      // `requiredSchema` instead of whole schema `dataSchema` here to not to break the original
+      // behaviour.
+      verifySchema(requiredSchema)
+    }
+
     (file: PartitionedFile) => {
       val lineIterator = {
         val conf = broadcastedHadoopConf.value.value
@@ -222,18 +231,18 @@ class CSVFileFormat extends TextBasedFileFormat with DataSourceRegister {
   }
 
   private def verifySchema(schema: StructType): Unit = {
-    def verifyType(dataType: DataType): Unit = dataType match {
+    def verifyType(name: String, dataType: DataType): Unit = dataType match {
         case ByteType | ShortType | IntegerType | LongType | FloatType |
              DoubleType | BooleanType | _: DecimalType | TimestampType |
              DateType | StringType =>
 
-        case udt: UserDefinedType[_] => verifyType(udt.sqlType)
+        case udt: UserDefinedType[_] => verifyType(name, udt.sqlType)
 
         case _ =>
           throw new UnsupportedOperationException(
-            s"CSV data source does not support ${dataType.simpleString} data type.")
+            s"Unable to convert column $name of type ${dataType.simpleString} to CSV.")
     }
 
-    schema.foreach(field => verifyType(field.dataType))
+    schema.foreach(field => verifyType(field.name, field.dataType))
   }
 }
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/csv/CSVFileFormat.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/csv/CSVFileFormat.scala
@@ -144,7 +144,7 @@ class CSVFileFormat extends TextBasedFileFormat with DataSourceRegister {
       sparkSession.sparkContext.broadcast(new SerializableConfiguration(hadoopConf))
 
     if (csvOptions.failFast) {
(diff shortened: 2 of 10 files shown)


sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/csv/CSVFileFormat.scala
```diff
@@ -222,18 +231,18 @@ class CSVFileFormat extends TextBasedFileFormat with DataSourceRegister {
...
+    def verifyType(name: String, dataType: DataType): Unit = dataType match {
         case ByteType | ShortType | IntegerType | LongType | FloatType |
              DoubleType | BooleanType | _: DecimalType | TimestampType |
              DateType | StringType =>
 
-        case udt: UserDefinedType[_] => verifyType(udt.sqlType)
+        case udt: UserDefinedType[_] => verifyType(name, udt.sqlType)
 
         case _ =>
           throw new UnsupportedOperationException(
```
HyukjinKwon: I am unhappy that CSV currently throws `UnsupportedOperation` but text datasource throws `AnalysisException`. I just matched this to `UnsupportedOperation`; however, I am happy to match this to `AnalysisException` if anyone thinks so.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/jsonExpressions.scala
```diff
@@ -487,6 +487,21 @@ case class JsonToStruct(schema: StructType, options: Map[String, String], child:
   override def dataType: DataType = schema
   override def children: Seq[Expression] = child :: Nil
 
+  override def checkInputDataTypes(): TypeCheckResult = {
+    if (StringType.acceptsType(child.dataType)) {
+      try {
+        JacksonUtils.verifySchema(schema)
```
HyukjinKwon: In this case, we don't have to worry about parsing mode, `mode` because `from_json` produces `null` with the default parse mode, `FAILFAST`.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/csv/CSVFileFormat.scala
```diff
@@ -143,6 +143,15 @@ class CSVFileFormat extends TextBasedFileFormat with DataSourceRegister {
     val broadcastedHadoopConf =
       sparkSession.sparkContext.broadcast(new SerializableConfiguration(hadoopConf))
 
+    if (csvOptions.failFast) {
+      // We can fail before starting to parse in case of "FAILFAST" mode. In case of "PERMISIVE"
+      // mode, it allows to read values as null for unsupported types. In case of "DROPMALFORMED"
+      // mode, it drops records only containing non-null values in unsupported types. We should use
+      // `requiredSchema` instead of whole schema `dataSchema` here to not to break the original
+      // behaviour.
+      verifySchema(requiredSchema)
```
HyukjinKwon: Here, it only checks projected columns for not changing the existing behaviour (we are not really checking the other columns when parsing already).

HyukjinKwon:
cc @marmbrus, could you please take a look? This is the one schema verification we talked when adding `to_json` .

HyukjinKwon:
Let me try to explain in more details and add some tests for reviewing.

HyukjinKwon:
cc @rxin and @marmbrus, actual change is only adding `JacksonUtils.verifySchema(schema)`, `CSVRelation.verifySchema(schema)` and `TextFileFormat.verifySchema(schema)` and the others are only test codes. Could you please take a look?

srowen (approved):

HyukjinKwon:
retest this please

marmbrus:
Thanks for working on this!  My meta question however is why don't we just support this instead?  We can parse interval type from strings and I think it would be really easy to write a converter that produces a string for write.

HyukjinKwon:
@marmbrus Sure! makes sense. One thing I instantly worry is, though, IIUC, it seems we should unset the check[1] for `CalendarIntervalType` for writing too and I am worried of other side effects. I can take a look into this deeper and then try to create a JIRA.

Anyway, this PR only makes the datasources throw the exceptions quicker and introduces some more test cases not existing before. So, I hope this is okay as it is if you are all fine.

[1]https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala#L355-L357

marmbrus:
Sure, we could possibly merge this first, and you are right, we'd need to remove checks.  I'll try to find some time to look this over as its a larger patch and I'm focusing on 2.1 bugs right now.  If you have time you might beat me with the better solution :)

HyukjinKwon:
retest this please

HyukjinKwon:
Hi @marmbrus, would there be something you are worried of and I should take a careful look for?

HyukjinKwon:
Do we want this change? If there is an approval, I will rebase. It seems easily making a conflict.

HyukjinKwon:
I will close this for now and make a new one soon.

PR #15757, opened 2016-11-03
[SPARK-18257][SS] Improve error reporting for FileStressSuite

rxin:
This patch improves error reporting for FileStressSuite, when there is an error in Spark itself (not user code). This works by simply tightening the exception verification, and gets rid of the unnecessary thread for starting the stream.

Also renamed the class FileStreamStressSuite to make it more obvious it is a streaming suite.

## How was this patch tested?
This is a test only change and I manually verified error reporting by injecting some bug in the addBatch code for FileStreamSink.


marmbrus:
LGTM

rxin:
Merging in master/branch-2.1.

PR #15758, opened 2016-11-03
[MINOR] Remove calculation of bucket spec when not expected

jaceklaskowski:

Remove calculation of bucket spec when asserted not to have one earlier.

## How was this patch tested?

Local build.

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/DataFrameWriter.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/DataFrameWriter.scala
@@ -209,11 +209,12 @@ final class DataFrameWriter[T] private[sql](ds: Dataset[T]) {
       df.sparkSession,
       className = source,
       partitionColumns = partitioningColumns.getOrElse(Nil),
-      bucketSpec = getBucketSpec,
+      bucketSpec = None,
       options = extraOptions.toMap)
 
     dataSource.write(mode, df)
   }
+
   /**
    * Inserts the content of the [[DataFrame]] to the specified table. It requires that
    * the schema of the [[DataFrame]] is the same as the schema of the table.


rxin:
Why is this the right thing to do?

jaceklaskowski:
Thanks @rxin for looking into it. While reviewing that code I noticed the call and thought I'd push it for review here since...`getBucketSpec` is superfluous given `assertNotBucketed("save")` at the very beginning. Correct me if I'm wrong (that will further help me understand the code).

I don't want to push the change for this one single line and wouldn't mind if you could merge it with some other more important change(s). Do whatever is best for the code. Thanks!

rxin:
where's the assert?

```


  private def getBucketSpec: Option[BucketSpec] = {
    if (sortColumnNames.isDefined) {
      require(numBuckets.isDefined, "sortBy must be used together with bucketBy")
    }

    for {
      n <- numBuckets
    } yield {
      require(n > 0 && n < 100000, "Bucket number must be greater than 0 and less than 100000.")

      // partitionBy columns cannot be used in bucketBy
      if (normalizedParCols.nonEmpty &&
        normalizedBucketColNames.get.toSet.intersect(normalizedParCols.get.toSet).nonEmpty) {
          throw new AnalysisException(
            s"bucketBy columns '${bucketColumnNames.get.mkString(", ")}' should not be part of " +
            s"partitionBy columns '${partitioningColumns.get.mkString(", ")}'")
      }

      BucketSpec(n, normalizedBucketColNames.get, normalizedSortColNames.getOrElse(Nil))
    }
  }
```

rxin:
ah you meant the assert in save itself.

I don't think we should remove it, since it is largely noop and might be useful once we allow bucket in save...

rxin:
Mind closing this?

PR #15761, opened 2016-11-03
[SPARK-17337][SQL] Do not pushdown predicates through filters with  predicate subqueries

hvanhovell:
The `PushDownPredicate` rule can create a wrong result if we try to push a filter containing a predicate subquery through a project when the subquery and the project share attributes (have the same source).

The current PR fixes this by making sure that we do not push down when there is a predicate subquery that outputs the same attributes as the filters new child plan.

## How was this patch tested?
Added a test to `SubquerySuite`. @nsyca has done previous work this. I have taken test from his initial PR.


nsyca:
A patch nicely done on this local issue. Minimal code but effective. The bigger problem remains there though, tracked by SPARk-17154.

sameeragarwal (approved):
LGTM

hvanhovell:
Merging this to master/branch-2.0. Thanks for the reviews!

PR #15763, opened 2016-11-04
[SPARK-17348][SQL] Incorrect results from subquery transformation

nsyca:

Return an Analysis exception when there is a correlated non-equality predicate in a subquery and the correlated column from the outer reference is not from the immediate parent operator of the subquery. This PR prevents incorrect results from subquery transformation in such case.

Test cases, both positive and negative tests, are added.

## How was this patch tested?

sql/test, catalyst/test, hive/test, and scenarios that will produce incorrect results without this PR and product correct results when subquery transformation does happen.


sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
```diff
@@ -977,6 +977,13 @@ class Analyzer(
         localPredicateReferences -- p.outputSet
       }
 
+      // SPARK-17348
+      // Report a non-supported case where there exists a correlated predicate
+      // that is not an equality predicate and the outer reference of the correlated predicate
+      // is not at the immediate parent operator.
+      def failOnNonEqualityPredicate(p: LogicalPlan): Unit = {
```
rxin: why define this since it is used only once?

hvanhovell:
ok to test

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
nsyca: yeah. I tried to make the error message reporting in line with existing ones, which have their own "def". I will collapse it in the main code.

nsyca:
Hmm. don't know what's wrong with the test failures reported in hive/test here. I will address @rxin's comment and submit a new version. That should trigger a new build.

nsyca:
I'd like to a note that a piece of existing code that closes to address this problem but only done for the scalar subquery context can be found by searching this pattern in CheckAnalysis.scala (line 126):
_"The correlated scalar subquery can only contain equality predicates:"_

An existing test coverage for that scenario is at
_SubquerySuite/non-equal correlated scalar subquery_

hvanhovell:
@nsyca you can retrigger builds by issuing `retest this please` this works as soon as an `ok to test` has been issued by an admin.

It might be a good idea to extend the logic in CheckAnalysis to also cover this case, that might make this more concise.

hvanhovell:
retest this please

nsyca:
@hvanhovell thank you for the trick.

Exploring your idea further, I looked at CheckAnalysis. This is a trait that Analyzer extends from. The checks in def CheckAnalysis of trait CheckAnalysis walk the operators in a LogicalPlan tree and look for violations specified to each operator type. The checks in Analyzer/ResolveSubquery has the context of a subquery, specifically the connect of those correlated predicates in the subquery, and look at everything from the subquery's viewpoint to see if there exists any violation then report an exception. I guess finding a correlated subquery for each operator type and flagging a violation is more expensive than here we having a correlated subquery, making sure no violation.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
```diff
@@ -1044,6 +1044,34 @@ class Analyzer(
...
       }
+
+      // SPARK-17348
+      // Looking for a potential incorrect result case.
+      // When a correlated predicate is a non-equality predicate
+      // it must be placed at the immediate child operator.
+      // Otherwise, the pull up of the correlated predicate
+      // will generate a plan with a different semantics
+      // which could return incorrect result.
+      var continue : Boolean = true
```
hvanhovell: Is it fair to say that we should never pull out non-equals predicates from an aggregate? It might be much easier to check this.

nsyca: That would regress a scenario currently support. The following queries demonstrate the good case that is working (giving correct results) today.

Seq((1,1)).toDF("c1","c2").createOrReplaceTempView("t1")
Seq((1,1),(2,0)).toDF("c1","c2").createOrReplaceTempView("t2")
sql("select c1 from t1 where c1 in (select t2.c1 from t2 where t1.c2 >= t2.c2)").show
sql("select t1.c1 from t1, t1 as t3 where t1.c1 = t3.c1 and (t1.c1 in (select t2.c1 from t2 where t1.c2 >= t2.c2 or t3.c2 < t2.c2) or t1.c2 >= 0)").show

Those two positive test cases have been added into SubquerySuite as part of this PR.

We would want to extend the support blocked by this PR in the future once we implement the correlated subquery processing.

hvanhovell: those are not aggregates right? I am saying that we should ban cases in which we pull out a predicate from an aggregate for example `t1.c2 >= t2.c2` in the following example:
```sql
select t1.c1
from   t1
where  t1.c1 in (select max(t2.c1)
                 from   t2
                 where  t1.c2 >= t2.c2)
```

nsyca: That scenario above is banned with this PR (and added as a new, negative, test case). However, the PR has a greater scope, Window function is another operator, we cannot pull up any correlated predicates, such as the example below:

    // Case 2: window func
    Seq((2,1)).toDF("c1","c2").createOrReplaceTempView("t1")
    Seq((1,1),(2,2)).toDF("c1","c2").createOrReplaceTempView("t2")
    sql("select c1 from t1 where c1 in (select max(t2.c1) over () from t2 where t1.c2 >= t2.c2)").show

hvanhovell: I am wondering if we should support `Window` at all. We can pull equals through `Window`, but that sounds pretty bizarre. What does DB2 support?

nsyca: ;-) An explicit question, eh?

DB2 uses a different approach. It does not aggressively transform subquery into join. I will speak of it in Spark terms. DB2 does not perform any query transformation during the Analyzer phase where parts of the LogicalPlan are still unresolved. This is because it doesn't know what the semantics of the entire query is. DB2 does a smalerl scope, comparing to what Spark does today, of transforming subquery to join in (rule-based) Optimizer phase, only with equality correlated predicate and when subquery referencing columns from immediate parents. Therefore DB2 does not encounter the problem we discuss in this PR.

One reason DB2 does differently is DB2 supports subquery processing at run-time so any correlated subquery left from the compiler/optimizer still can be processed.

I will touch on a broader scope of subquery transformation in the next update

nsyca: One technique that I know of being used to transform correlation queries to queries with no correlation is outlined in this 1996 IEEE Data Engineering paper.

    Complex query decorrelation
    P. Seshadri; H. Pirahesh; T. Y. C. Leung
    Data Engineering, 1996. Proceedings of the Twelfth International Conference on
    Pages: 450 - 458

Distributed systems aggravate the performance impact of correlated queries from the movement of the entire data set of the subqueries to where the data of the outer tables reside. This processing is similar to the `BroadcastNestedLoopJoinExec` in Spark.

The idea behind the paper is to build a duplicate portion of the outer tables and de-correlate the original subquery by joining the duplicate portion within the subquery. The algorithm is claimed to be generic and can be applied to all forms of correlations, both shallow correlation where the correlated point is immediately below the operation over the outer table(s), and deep correlation, where the correlated point is at arbitrary level below the operation over the outer tables.

nsyca:
retest this please

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
rxin: Well there is also this http://www.btw-2015.de/res/proceedings/Hauptband/Wiss/Neumann-Unnesting_Arbitrary_Querie.pdf

nsyca: Going back to @hvanhovell 's comment on the support of `Window`, I think we should leave it here for the equality predicates, as the current behaviour gives correct results. We should only close off the incorrect results from this PR.

nsyca: @rxin, Shall we continue the discussion of the general algorithm to resolve the correlated subquery issue in a new thread? Would you be interested in putting this item into 2.2?

nsyca:
I think I have addressed all the pending comments on this PR. Would be there anything left to do?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
```diff
@@ -1044,6 +1044,34 @@ class Analyzer(
...
+      for (pm <- predicateMap if continue) {
+        assert(pm._2.nonEmpty, "Correlated predicate(s) does not exist.")
+        for (p <- pm._2 if continue)
+          p match {
+            case EqualTo(_, _) | EqualNullSafe(_, _) =>
+              None
+            case _ =>
+              assert(transformed.children.nonEmpty)
+              if (!(transformed.isInstanceOf[Project]) ||
```
hvanhovell: This only allows a `Project` in combination with non-equals predicates. Is this not to limiting? I thought the only problems are with Aggregate and Window?

hvanhovell: Do you think we can put this in the previous block? That might be easier.

srinathshankar:
As it stands the check seems overly aggressive. The following works:
```sql
select t1.c1
from t1
where t1.c1 in (
  select c1 + 1 as col 
  from t2 
  where t1.c2 < t2.c2)
```
```
scala> sql("select t1.c1 from t1 where t1.c1 in (select c1 + 1 as col from t2 where t1.c2 < t2.c2)").show()
+---+
| c1|
+---+
+---+
```
But the following, when I wrap the subselect with a project, does not
```sql
select t1.c1
from t1
where t1.c1 in (
  select col + 1
  from (
    select c1 as col
    from t2
    where t1.c2 < t2.c2))
```
```
scala> sql("select t1.c1 from t1 where t1.c1 in (select col + 1 from (select c1 as col from t2 where t1.c2 < t2.c2))").show()
org.apache.spark.sql.AnalysisException: Correlated column is not allowed in a non-equality predicate:
Project [(col#29 + 1) AS (col + 1)#32]
+- Project [(c1#16 + 1) AS col#29]
   +- Filter (outer(c2#6) < c2#17)
      +- SubqueryAlias t2
         +- Project [_1#13 AS c1#16, _2#14 AS c2#17]
            +- LocalRelation [_1#13, _2#14]
;
```

Is this intentional ?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
nsyca: @hvanhovell I totally agree with you on sharing different perspectives and I don't mind to share what I can. I would like to draw up a plan for subquery processing but I hesitate to commit to a full support in 2.2. When is the rough date for 2.2 code cut-off? Given that I am relatively new to Spark code base, I would seek the commitment from someone like you (or any committers) who has experience in Spark SQL to review my design proposal and code. You would be an ideal candidate since you implemented most, if not all, the subquery functionality in 2.0.

On the second part of your comment, I do not know the code that does the semantics resolution from a unresolved LogicalPlan to a fully resolved LogicalPlan. I make this comment based on my general knowledge of SQL compilation. I lean toward to agree with your last statement that a complete isolation between resolving a reference (current or outer) and the rewrite. I believe this is an important principle we want to keep in the design of a SQL compiler and optimizer. How to do that cleanly? The above SQL is equivalent to  
````
select ... 
from (
      select b.key, min(b.value) as min_value
      from src b
      group by b.key) as Q
where exists (select a.key
              from src a
              where a.value > 'val_9' and a.value = Q.min_value)
````

To generalize further, the expression(s) referencing the outer reference, B in this case, in the HAVING clause needs not to be the same as in the SELECT clause, i.e., the aggregate function in the EXISTS subquery can be any aggregate function on any column in B. Spark supports this today.

What we may need is to recognize that there is an expression referencing outer reference in the subquery and that expression, if does not exist, need to be added to the projection list of the outer reference and the identifier of that expression will be used to substitute the whole expression in the subquery.

I'd like to propose we spin off the discussion of this topic in a separate thread, both on the technical detail and the schedule.

**Comment by nsyca** [comment]
@srinathshankar It is intentional. It is impossible to do an analysis on what in-between operations we can allow and what we cannot. Correlated predicates can be placed in any arbitrary level of depth in a subquery. Spark may not support more than one level of correlations today but it may in the next version. So my argument is if we cannot proof that pulling up a correlated predicate thru any operations will still preserve its original semantics, then we should not do it. The paper @rxin mentioned in the previous note (http://www.btw-2015.de/res/proceedings/Hauptband/Wiss/Neumann-Unnesting_Arbitrary_Querie.pdf) makes this claim in the paragraph after Q2 in page 2.

This PR is not a full solution. It is intended to be a temporary stop-gap solution to close off the incorrect result cases Spark exposes today. It could be argued that your example above is a regression but the statement can be rewritten to make it work by collapsing the 2 levels of subselects. It is harder to implement a solution that will walk the whole plan tree of a subquery to determine which operations are fine to let the correlated predicate pulled through.

A permanent solution, as I proposed in one of my comments above, is to separate the transformation of any correlated predicates to Optimizer phase and leave Analyzer phase to just resolve the references and validate the input SQL is valid. This way, the two subselects in your example will probably be merged and then the correlated predicate pull up follows.

File: sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala (hunk 3 above)
Comment (nsyca) [inline sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala]: @hvanhovell The key point of this PR is on the second condition, not the check for the `Project`. Its intention is if the correlated predicate is not at the immediate child, we are about to pull the correlated predicate up through a number of operations. It is very hard to prove that which operations we can pull the predicate through without changing the meaning of the original SQL. Yes, there are cases that it may regress, like the example @srinathshankar posted.

I do not understand your second comment on putting in the previous block. Could you please explain more?

**Comment by nsyca** [comment]
@hvanhovell ping... Are there anything I need to do to close this incorrect result problem?

**Comment by hvanhovell** [comment]
@nsyca I feel that the current approach is to restrictive. I would prefer to just close the gap for Window and Aggregate.

**Comment by nsyca** [comment]
@hvanhovell Then we will need to walk from the top of the operator hosting the outer reference to the operator hosting the correlation to ensure there is no Aggregate or Window operator.

If this is the algorithm you have in mind, I will see how I could implement it in an effective way.

Examples are:
````
// Case 1: agg func
Seq((1,1)).toDF("c1","c2").createOrReplaceTempView("t1")
Seq((1,1),(2,0)).toDF("c1","c2").createOrReplaceTempView("t2")
// Case 1A: Simplest case Aggregate is immediately under Project and over the correlation point
sql("""
| select t1.c1
| from t1
| where t1.c1 in (select max(t2.c1) 
|                 from t2 where t1.c2 >= t2.c2)
""".stripMargin).show
// Case 1B: Project - Filter - Aggregate - [correlation point]
// Project is the operator immediately over T1, the source of the outer reference
// There is a Filter (hosting the HAVING logic) in between
sql("""
| select t1.c1
| from t1
| where t1.c1 in (select max(t2.c1)
|                 from t2
|                 where t1.c2 >= t2.c2
|                 having count(*) > 0 )
""".stripMargin).show

// And the list can go on and on such as this one that
// Case 1C:  Insert a layer of SELECT on top of Case 1B
sql("""
| select t1.c1 
| from t1
| where t1.c1 in (select maxc+1
|                 from (select max(t2.c1) as maxc
|                       from t2
|                       where t1.c2 >= t2.c2
|                       having count(*) > 0 ))
""".stripMargin).show
````
````
// Here is the LogicalPlan at the point of checking for Case 1C
Project [(maxc#161 + 1) AS (maxc + 1)#167, c2#103]  <-- This is the Project hosting the outer reference T1.C1
+- Project [maxc#161, c2#103]
   +- Filter (count(1)#166L > cast(0 as bigint))
      +- Aggregate [c2#103], [max(c1#102) AS maxc#161, count(1) AS count(1)#166L, c2#103]
         +- SubqueryAlias t2, `t2`   <-- The boundary of Subquery
            +- Project [_1#99 AS c1#102, _2#100 AS c2#103]
               +- LocalRelation [_1#99, _2#100]
````

**Comment by hvanhovell** [comment]
@nsyca it should be relatively straighforward to implement this here: https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala#L1045-L1111

**Comment by hvanhovell** [comment]
I think it is also better that we start whitelisting operators instead blacklisting them.

**Comment by nsyca** [comment]
Thanks for the tip on the code. I will work on it.

**Comment by nsyca** [comment]
@hvanhovell could you please review the latest PR?

File: sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala [hunk 4]
```diff
@@ -1041,12 +1070,24 @@ class Analyzer(
...
 
+      var foundNonEqualCorrelatedPred : Boolean = false
+
       // Simplify the predicates before pulling them out.
       val transformed = BooleanSimplification(sub) transformUp {
         case f @ Filter(cond, child) =>
           // Find all predicates with an outer reference.
           val (correlated, local) = splitConjunctivePredicates(cond).partition(containsOuter)
 
+          // Find any non-equality correlated predicates
```
hvanhovell: This is a bit more concise and will terminate as soon as we find a non-equal predicate.
```scala
foundNonEqualCorrelatedPred = foundNonEqualCorrelatedPred || correlated.exists {
  case _: EqualTo | _: EqualNullSafe => false
  case _ => true
}
```

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
```diff
@@ -1069,11 +1110,19 @@ class Analyzer(
         case a @ Aggregate(grouping, expressions, child) =>
           failOnOuterReference(a)
           val referencesToAdd = missingReferences(a)
+          if (foundNonEqualCorrelatedPred) {
```
hvanhovell: Move the `if (foundNonEqualCorrelatedPred)` into the `failOnNonEqualCorrelatedPredicate` function

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
```diff
@@ -1069,11 +1110,19 @@ class Analyzer(
...
           if (referencesToAdd.nonEmpty) {
             Aggregate(grouping ++ referencesToAdd, expressions ++ referencesToAdd, child)
           } else {
             a
           }
+        case w : Window =>
+          if (foundNonEqualCorrelatedPred) {
+            failOnNonEqualCorrelatedPredicate(w)
+          }
+          w
```
hvanhovell: Should we rewrite the aggregates in the window function? For example:
```sql
select *
from  tbl_x
where value in (select max() over() from tbl_y where tbl_y.id = tbl_x.id)
```
Should we rewrite `max() over()` into `max() over(partition by tbl_y.id)`?

hvanhovell (changes requested):
@nsyca this is much better! I left a few minor comments regarding style. I was also wondering if we still need the following checkAnalysis lines: https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/CheckAnalysis.scala#L122-L127

I do want to discuss the semantics of Window in subqueries, see my comment about that.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
nsyca: Thank you for this version of code.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
nsyca: We could do that if all the correlated predicates are equality predicates. It is as if the computation of the subquery is bound by each value from the outer reference and therefore can be bound in the PARTITION BY clause of a Window function.

I guess we want to implement this transformation outside of this PR, preferrably in Optimizer phase.

nsyca:
@hvanhovell 

I removed the redundant checking in `ScalarSubquery` as pointed out in your earlier comment (copied below).

"I was also wondering if we still need the following checkAnalysis lines: https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/CheckAnalysis.scala#L122-L127"

I will submit another PR for this work, including updating some existing test result.

nsyca:
@hvanhovell could you please review the PR again? Hopefully this is the last one.

hvanhovell:

@nsyca could you open a JIRA discuss a generic subquery operator.

PR #15764, opened 2016-11-04
[SPARK-18217] [SQL] Disallow creating permanent views based on temporary views or UDFs

gatorsmile:
Based on the discussion in [SPARK-18209](https://issues.apache.org/jira/browse/SPARK-18209). It doesn't really make sense to create permanent views based on temporary views or temporary UDFs. 

To disallow the supports and issue the exceptions, this PR needs to detect whether a temporary view/UDF is being used when defining a permanent view. Basically, this PR can be split to two sub-tasks:

**Task 1:** detecting a temporary view from the query plan of view definition.
When finding an unresolved temporary view, Analyzer replaces it by a `SubqueryAlias` with the corresponding logical plan, which is stored in an in-memory HashMap. After replacement, it is impossible to detect whether the `SubqueryAlias` is added/generated from a temporary view. Thus, to detect the usage of a temporary view in view definition, this PR traverses the unresolved logical plan and uses the name of an `UnresolvedRelation` to detect whether it is a (global) temporary view.

**Task 2:** detecting a temporary UDF from the query plan of view definition.
Detecting usage of a temporary UDF in view definition is not straightfoward. 

First, in the analyzed plan, we are having different forms to represent the functions. More importantly, some classes (e.g., `HiveGenericUDF`) are not accessible from `CreateViewCommand`, which is part of  `sql/core`. Thus, we used the unanalyzed plan `child` of `CreateViewCommand` to detect the usage of a temporary UDF. Because the plan has already been successfully analyzed, we can assume the functions have been defined/registered. 

Second, in Spark, the functions have four forms: Spark built-in functions, built-in hash functions, permanent UDFs and temporary UDFs. We do not have any direct way to determine whether a function is temporary or not. Thus, we introduced a function `isTemporaryFunction` in `SessionCatalog`. This function contains the detailed logics to determine whether a function is temporary or not.

Added test cases.

### DIFF
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/SessionCatalog.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/SessionCatalog.scala
@@ -553,16 +553,16 @@ class SessionCatalog(
       val relationAlias = alias.getOrElse(table)
       if (db == globalTempViewManager.database) {
         globalTempViewManager.get(table).map { viewDef =>
-          SubqueryAlias(relationAlias, viewDef, Some(name))(isGeneratedByTempTable = true)
+          SubqueryAlias(relationAlias, viewDef, Some(name))
         }.getOrElse(throw new NoSuchTableException(db, table))
       } else if (name.database.isDefined || !tempTables.contains(table)) {
         val metadata = externalCatalog.getTable(db, table)
         val view = Option(metadata.tableType).collect {
           case CatalogTableType.VIEW => name
         }
-        SubqueryAlias(relationAlias, SimpleCatalogRelation(db, metadata), view)()
+        SubqueryAlias(relationAlias, SimpleCatalogRelation(db, metadata), view)
       } else {
-        SubqueryAlias(relationAlias, tempTables(table), Option(name))(isGeneratedByTempTable = true)
+        SubqueryAlias(relationAlias, tempTables(table), Option(name))
       }
     }
   }
@@ -926,7 +926,7 @@ class SessionCatalog(
   /**
    * Returns whether it is a temporary function.
    */
-  def isTempFunction(name: FunctionIdentifier): Boolean = {
+  def isTemporaryFunction(name: FunctionIdentifier): Boolean = {
     // copied from HiveSessionCatalog
     val hiveFunctions = Seq(
       "hash",
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/SessionCatalog.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/SessionCatalog.scala
@@ -553,16 +553,16 @@ class SessionCatalog(
       val relationAlias = alias.getOrElse(table)
       if (db == globalTempViewManager.database) {
         globalTempViewManager.get(table).map { viewDef =>
-          SubqueryAlias(relationAlias, viewDef, Some(name))
+          SubqueryAlias(relationAlias, viewDef, Some(name))(isGeneratedByTempTable = true)
         }.getOrElse(throw new NoSuchTableException(db, table))
       } else if (name.database.isDefined || !tempTables.contains(table)) {
         val metadata = externalCatalog.getTable(db, table)
         val view = Option(metadata.tableType).collect {
           case CatalogTableType.VIEW => name
         }
-        SubqueryAlias(relationAlias, SimpleCatalogRelation(db, metadata), view)
+        SubqueryAlias(relationAlias, SimpleCatalogRelation(db, metadata), view)()
       } else {
-        SubqueryAlias(relationAlias, tempTables(table), Option(name))
+        SubqueryAlias(relationAlias, tempTables(table), Option(name))(isGeneratedByTempTable = true)
       }
     }
   }
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/SessionCatalog.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/SessionCatalog.scala
@@ -923,6 +923,24 @@ class SessionCatalog(
     }
   }
 
+  /**
+   * Returns whether it is a temporary function.
+   */
+  def isTempFunction(name: FunctionIdentifier): Boolean = {
+    // copied from HiveSessionCatalog
+    val hiveFunctions = Seq(
+      "hash",
+      "histogram_numeric",
+      "percentile")
+
+    // A temporary function is a function that has been registered in functionRegistry
+    // without a database name, and is neither a built-in function nor a Hive function
+    name.database.isEmpty &&
+      functionRegistry.functionExists(name.funcName) &&
+      !FunctionRegistry.builtin.functionExists(name.funcName) &&
+      !hiveFunctions.contains(name.funcName.toLowerCase)
+  }
+
   protected def failFunctionLookup(name: String): Nothing = {
     throw new NoSuchFunctionException(db = currentDb, func = name)
   }
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/SessionCatalog.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/SessionCatalog.scala
@@ -924,7 +924,7 @@ class SessionCatalog(
   }
 
   /**
-   * Returns whether it is a temporary function.
+   * Returns whether it is a temporary function. If not existed, returns false.
    */
   def isTemporaryFunction(name: FunctionIdentifier): Boolean = {
     // copied from HiveSessionCatalog
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/basicLogicalOperators.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/basicLogicalOperators.scala
@@ -709,10 +709,13 @@ case class LocalLimit(limitExpr: Expression, child: LogicalPlan) extends UnaryNo
 case class SubqueryAlias(
     alias: String,
     child: LogicalPlan,
-    view: Option[TableIdentifier])
+    view: Option[TableIdentifier])(
+    val isGeneratedByTempTable: java.lang.Boolean = false)
   extends UnaryNode {
 
   override def output: Seq[Attribute] = child.output.map(_.withQualifier(Some(alias)))
+
+  override protected def otherCopyArgs: Seq[AnyRef] = isGeneratedByTempTable :: Nil
 }
 
 /**
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/basicLogicalOperators.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/basicLogicalOperators.scala
@@ -709,13 +709,10 @@ case class LocalLimit(limitExpr: Expression, child: LogicalPlan) extends UnaryNo
 case class SubqueryAlias(
     alias: String,
     child: LogicalPlan,
-    view: Option[TableIdentifier])(
-    val isGeneratedByTempTable: java.lang.Boolean = false)
+    view: Option[TableIdentifier])
   extends UnaryNode {
 
   override def output: Seq[Attribute] = child.output.map(_.withQualifier(Some(alias)))
-
-  override protected def otherCopyArgs: Seq[AnyRef] = isGeneratedByTempTable :: Nil
 }
 
 /**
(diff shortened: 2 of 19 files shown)


sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/basicLogicalOperators.scala
```diff
@@ -709,10 +709,13 @@ case class LocalLimit(limitExpr: Expression, child: LogicalPlan) extends UnaryNo
 case class SubqueryAlias(
     alias: String,
     child: LogicalPlan,
-    view: Option[TableIdentifier])
+    view: Option[TableIdentifier])(
+    val isGeneratedByTempTable: java.lang.Boolean = false)
```
rxin: is there a way to do it without introducing this?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/SessionCatalog.scala
```diff
@@ -923,6 +923,24 @@ class SessionCatalog(
     }
   }
 
+  /**
+   * Returns whether it is a temporary function.
+   */
+  def isTempFunction(name: FunctionIdentifier): Boolean = {
+    // copied from HiveSessionCatalog
```
gatorsmile: Will do. Thanks!

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/basicLogicalOperators.scala
gatorsmile: Just updated the PR description. This might be the cleanest way. The reason is explained below.

> When finding an unresolved temporary view, Analyzer replace it by a `SubqueryAlias` with the corresponding logical plan, which is stored in an in-memory HashMap. After replacement, it is impossible to detect whether the SubqueryAlias is added/generated from a temporary view. Thus, to detect the usage of a temporary view in view definition, we added an extra flag `isGeneratedByTempTable` into SubqueryAlias. The flag is added into the curried arguments. Via this extra flag, we can easily detect the usage of temporary view from a logical plan traversal.

Also cc @cloud-fan and @liancheng Do you have any better solution?

sql/core/src/main/scala/org/apache/spark/sql/execution/command/views.scala
```diff
@@ -131,6 +131,26 @@ case class CreateViewCommand(
...
+            if sparkSession.sessionState.catalog.isTemporaryTable(s.tableIdentifier) =>
+          throw new AnalysisException(s"Not allowed to create a permanent view $name by " +
+            s"referencing a temp view ${s.tableIdentifier}. " +
+            originalText.map(sql => s"""SQL: "$sql".""").getOrElse(""))
+        case other if !other.resolved => other.expressions.flatMap(_.collect {
+          // Disallow creating permanent views based on temporary UDFs.
+          case e: UnresolvedFunction
+              if sparkSession.sessionState.catalog.isTemporaryFunction(e.name) =>
+            throw new AnalysisException(s"Not allowed to create a permanent view $name by " +
+              s"referencing a temp function `${e.name}`. " +
```
rxin: temp -> temporary

sql/core/src/main/scala/org/apache/spark/sql/execution/command/views.scala
```diff
@@ -131,6 +131,26 @@ case class CreateViewCommand(
...
 
+    // When creating a permanent view, not allowed to reference temporary objects.
+    if (!isTemporary) {
+      child.collect {
+        // Disallow creating permanent views based on temporary views.
+        case s: UnresolvedRelation
+            if sparkSession.sessionState.catalog.isTemporaryTable(s.tableIdentifier) =>
+          throw new AnalysisException(s"Not allowed to create a permanent view $name by " +
+            s"referencing a temp view ${s.tableIdentifier}. " +
+            originalText.map(sql => s"""SQL: "$sql".""").getOrElse(""))
```
rxin: do we really need to show the entire sql? it's basically what the user just typed in

sql/core/src/main/scala/org/apache/spark/sql/execution/command/views.scala
```diff
@@ -131,6 +131,26 @@ case class CreateViewCommand(
...
+        case other if !other.resolved => other.expressions.flatMap(_.collect {
+          // Disallow creating permanent views based on temporary UDFs.
+          case e: UnresolvedFunction
+              if sparkSession.sessionState.catalog.isTemporaryFunction(e.name) =>
+            throw new AnalysisException(s"Not allowed to create a permanent view $name by " +
+              s"referencing a temp function `${e.name}`. " +
+              originalText.map(sql => s"""SQL: "$sql".""").getOrElse(""))
```
rxin: same here do we really need to show the entire sql? it's basically what the user just typed in

sql/core/src/main/scala/org/apache/spark/sql/execution/command/views.scala
```diff
@@ -131,6 +131,26 @@ case class CreateViewCommand(
         s"specified by CREATE VIEW (num: `${userSpecifiedColumns.length}`).")
     }
 
+    // When creating a permanent view, not allowed to reference temporary objects.
+    if (!isTemporary) {
```
rxin: also i'd move the check into a function, so it is more obvious what's going on with the main workflow.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/SessionCatalog.scala
```diff
@@ -923,6 +923,24 @@ class SessionCatalog(
     }
   }
 
+  /**
+   * Returns whether it is a temporary function.
+   */
+  def isTemporaryFunction(name: FunctionIdentifier): Boolean = {
```
rxin: add a unit test for this function?

rxin: also what's the behavior if the function doesn't exist?

sql/core/src/main/scala/org/apache/spark/sql/execution/command/views.scala
```diff
@@ -131,6 +131,26 @@ case class CreateViewCommand(
...
+      child.collect {
+        // Disallow creating permanent views based on temporary views.
```
rxin: i'd also copy paste what you put in the pr description in here, on why you are traversing the unresolved plan.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/SessionCatalog.scala
gatorsmile: Will resolve this tomorrow.

sql/hive/src/test/scala/org/apache/spark/sql/hive/execution/SQLViewSuite.scala
```diff
@@ -462,4 +487,65 @@ class SQLViewSuite extends QueryTest with SQLTestUtils with TestHiveSingleton {
       }
     }
   }
+
+  test("create a permanent/temp view using a hive function") {
```
rxin: while it's good to have separate test cases, i find these three separate cases a bit too verbose. why not just put hive, built-in, and permanent into a single query?

gatorsmile: Sure, will do it.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/SessionCatalog.scala
gatorsmile: Like `isTemporaryTable`, we return false when the function/table does not exist

rxin: Yea please docuemnt it.

rxin:
Looks good but I didn't read super carefully.

cc @hvanhovell and @cloud-fan

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/catalog/SessionCatalogSuite.scala
```diff
@@ -919,6 +919,34 @@ class SessionCatalogSuite extends SparkFunSuite {
...
+    val info1 = new ExpressionInfo("tempFunc1", "temp1")
+    sessionCatalog.createTempFunction("temp1", info1, tempFunc1, ignoreIfExists = false)
+
+    // Returns true when the function is temporary
+    assert(sessionCatalog.isTemporaryFunction(FunctionIdentifier("temp1")))
+
+    // Returns false when the function is permanent
+    assert(externalCatalog.listFunctions("db2", "*").toSet == Set("func1"))
+    assert(!sessionCatalog.isTemporaryFunction(FunctionIdentifier("func1", Some("db2"))))
+    assert(!sessionCatalog.isTemporaryFunction(FunctionIdentifier("db2.func1")))
```
cloud-fan: it's not a permanent function right? it's a function called `db2.func1` which doesn't exist

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/SessionCatalog.scala
```diff
@@ -923,6 +923,24 @@ class SessionCatalog(
...
+    val hiveFunctions = Seq(
+      "hash",
+      "histogram_numeric",
+      "percentile")
+
+    // A temporary function is a function that has been registered in functionRegistry
+    // without a database name, and is neither a built-in function nor a Hive function
+    name.database.isEmpty &&
+      functionRegistry.functionExists(name.funcName) &&
+      !FunctionRegistry.builtin.functionExists(name.funcName) &&
```
cloud-fan: `toLowerCase`?

sql/core/src/main/scala/org/apache/spark/sql/execution/command/views.scala
```diff
@@ -173,6 +177,34 @@ case class CreateViewCommand(
   }
 
   /**
+   * Permanent views are not allowed to reference temp objects, including temp function and views
+   */
+  private def verifyTemporaryObjectsNotExists(sparkSession: SparkSession): Unit = {
+    if (!isTemporary) {
+      // This func traverses the unresolved plan `child`. Below are the reasons:
```
cloud-fan: is it possible the `child` is resolved? e.g. by `DataFrame` API

cloud-fan: nvm, we can only create permanent view by SQL API

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/SessionCatalog.scala
gatorsmile: Our built-in function registry is using `SimpleFunctionRegistry`, which is using [case insensitive string key hash map](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/FunctionRegistry.scala#L71-L72). 

Thus, no need to add `toLowerCase`. However, we need to add a test case to ensure we can capture the potential change in this part.

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/catalog/SessionCatalogSuite.scala
gatorsmile: The [related codes](https://github.com/apache/spark/blob/96cc1b5675273c276e04c4dc19ef9033a314292d/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/SessionCatalog.scala#L938) in `lookupFunction` looks confusing. 

Here, you are right, `functionRegistry` does not have such a function. Let me remove it. Thanks!

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/SessionCatalog.scala
gatorsmile: Added a test case: https://github.com/gatorsmile/spark/blob/fec0066331aaf085f8cd0318be23844560e72b4b/sql/hive/src/test/scala/org/apache/spark/sql/hive/execution/SQLViewSuite.scala#L496

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/SessionCatalog.scala
```diff
@@ -923,6 +923,24 @@ class SessionCatalog(
...
+      !hiveFunctions.contains(name.funcName.toLowerCase)
```
tejasapatil: I thought `HiveSessionCatalog` is used in context of Hive and `SessionCatalog` has nothing to do with Hive. So if we remove this from here and override `isTemporaryFunction` in `HiveSessionCatalog`, this would look clean. Feels like I am missing something obvious here.

rxin: It is true - but we are working towards getting rid of HiveSessionCatalog though (including getting rid of the 3 fallback functions), so in practice this will make no difference soon.

rxin:

PR #15765, opened 2016-11-04
[MINOR][Documentation] Fix some minor descriptions in functions consistently with expressions

HyukjinKwon:

This PR proposes to improve documentation and fix some descriptions equivalent to several minor fixes identified in https://github.com/apache/spark/pull/15677

Also, this suggests to change `Note:` and `NOTE:` to `.. note::` consistently with the others which marks up pretty.

## How was this patch tested?

Jenkins tests and manually. 

For PySpark, `Note:` and `NOTE:` to `.. note::` make the document as below:

**From**

**To**

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/functions.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/functions.scala
@@ -1119,6 +1119,7 @@ object functions {
   /**
    * Generate a random column with independent and identically distributed (i.i.d.) samples
    * from U[0.0, 1.0].
+   *
    * Note that this is indeterministic when data partitions are not fixed.
    *
    * @group normal_funcs
--- a/sql/core/src/main/scala/org/apache/spark/sql/functions.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/functions.scala
@@ -1117,8 +1117,8 @@ object functions {
   def not(e: Column): Column = !e
 
   /**
-   * Generate a random column with i.i.d. samples from U[0.0, 1.0].
-   *
+   * Generate a random column with independent and identically distributed (i.i.d.) samples
+   * from U[0.0, 1.0].
    * Note that this is indeterministic when data partitions are not fixed.
    *
    * @group normal_funcs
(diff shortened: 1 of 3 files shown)


HyukjinKwon:
It seems something wrongly gone in my local so I can't run cran-check in my local. Please let me run this via Jenkins.

HyukjinKwon:
Please let me cc @felixcheung and @srowen though if you don't mind.

HyukjinKwon:
Oh, actually, it seems I should fix the ones in `functions.scala` too.

sql/core/src/main/scala/org/apache/spark/sql/functions.scala
```diff
@@ -1145,15 +1147,16 @@ object functions {
...
+   * Generate a column with independent and identically distributed (i.i.d.) samples from
+   * the standard normal distribution.
    *
    * @group normal_funcs
    * @since 1.4.0
    */
   def randn(): Column = randn(Utils.random.nextLong)
 
   /**
-   * Partition ID of the Spark task.
```
HyukjinKwon: This one was missed in https://github.com/apache/spark/commit/02f203107b8eda1f1576e36c4f12b0e3bc5e910e and https://github.com/apache/spark/commit/a08463b1d32348a81d0f148dfaf22741d5c23b1a

srowen (approved):

HyukjinKwon:
It seems the newly added test seems too flaky. the specific test seems often failed. `randomized aggregation test - [typed, with partial + safe] - with grouping keys - with non-empty input` cc @liancheng

HyukjinKwon:
retest this please

felixcheung:
changes LGTM. thanks.
to nit pick, I find this "(Signed) shift the given value numBits right." confusing. Perhaps we should clarify this across all doc?

srowen:
How would you write it? we do need to specify that it's a signed shift.

felixcheung:
hmm, like "shift the signed value numBits to the right"?

HyukjinKwon:
Ah, IMHO, it seems both are fine but actually it seems there is an unsigned shift one which is described as..

```
Unsigned shift the given value numBits right.
```

So.. if any of you does not feel strongly which one is better, I hope leave as it is because it seems at least consistent.

```
(Signed) shift the given value numBits right.
```

felixcheung:
I think both sound odd, but it's a nit

srowen:
It's not a shift of a signed value, but a 'signed' shift, meaning one that preserves the sign bit. I recognize the terms "unsigned shift" and "signed shift", FWIW.

felixcheung:
Sure then.
Merged to master/branch-2.1

HyukjinKwon:
Thank you both!

PR #14498, opened 2016-08-04
[SPARK-16904] [SQL] Removal of Hive Built-in Hash Functions and TestHiveFunctionRegistry

gatorsmile:
### What changes were proposed in this pull request?

Currently, the Hive built-in `hash` function is not being used in Spark since Spark 2.0. The public interface does not allow users to unregister the Spark built-in functions. Thus, users will never use Hive's built-in `hash` function. 

The only exception here is `TestHiveFunctionRegistry`, which allows users to unregister the built-in functions. Thus, we can load Hive's hash function in the test cases. If we disable it, 10+ test cases will fail because the results are different from the Hive golden answer files.

This PR is to remove `hash` from the list of `hiveFunctions` in `HiveSessionCatalog`. It will also remove `TestHiveFunctionRegistry`. This removal makes us easier to remove `TestHiveSessionState` in the future.
### How was this patch tested?
N/A

PR #15766, opened 2016-11-04
[SPARK-18271][SQL]hash udf in HiveSessionCatalog.hiveFunctions is redundant

windpiger:

when lookupfunction in HiveSessionCatalog, first look up in spark's build-in functionRegistry, if it not existed, then look up the Hive'S build-in functions which are listed in Seq(hiveFunctions).

But the [hash] function is already in spark's build-in functionRegistry list, so it will never to go to look up in Hive's hiveFunctions list, so the [hash] function which is in Hive's hiveFunctions is redundant, we can remove it.

### DIFF
--- a/sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveSessionCatalog.scala
+++ b/sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveSessionCatalog.scala
@@ -227,7 +227,6 @@ private[sql] class HiveSessionCatalog(
   // in_file, index, matchpath, ngrams, noop, noopstreaming, noopwithmap,
   // noopwithmapstreaming, parse_url_tuple, reflect2, windowingtablefunction.
   private val hiveFunctions = Seq(
-    "hash",
     "histogram_numeric",
     "percentile"
   )


cloud-fan:
ok to test

cloud-fan:
LGTM

andrewor14:
retest this please

rxin:
Don't we actually unregister this when testing Hive?

cloud-fan:
@rxin  good catch! We do unregister the spark builtin hash in test: https://github.com/apache/spark/blob/master/sql/hive/compatibility/src/test/scala/org/apache/spark/sql/hive/execution/HiveCompatibilitySuite.scala#L60-L61

So we have a little more work to do here, we should register the hive hash function in that test suite.

gatorsmile:
I am working on the related issue in https://github.com/apache/spark/pull/14498

windpiger:
@rxin  @cloud-fan you are rigth,hash should be unregistered and replace with Hive's hash, or we could put the failed hash testcase into blacklist as @gatorsmile  's work #14498 . I will close the PR

windpiger:
@cloud-fan I will appreciate that you can help to close this PR~

cloud-fan:
Only the owner(yourself) can close this PR.

PR #15329, opened 2016-10-03
[SPARK-17763][SQL] JacksonParser silently parses null as 0 when the field is not nullable

HyukjinKwon:
## What changes were proposed in this pull request?

This PR proposes handling not nullable field properly in `JacksonParser`. For example, the codes below:

``` scala
val testJson = """{"nullInt":null}""" :: Nil
val testSchema = StructType(StructField("nullInt", IntegerType, false) :: Nil)
val data = spark.sparkContext.parallelize(testJson)
spark.read.schema(testSchema).json(data).show()
```

prints

**Before**

```
+-------+
|nullInt|
+-------+
|      0|
+-------+
```

**After**

```
org.apache.spark.sql.catalyst.json.NotAllowedNullException: Null not allowed: {"nullInt":null}
    at org.apache.spark.sql.catalyst.json.JacksonParser.failedRecord(JacksonParser.scala:131)
    at org.apache.spark.sql.catalyst.json.JacksonParser.parse(JacksonParser.scala:472)
...
```
## How was this patch tested?

Unit test in`JsonSuite`.

PR #14124, opened 2016-07-10
[SPARK-16472][SQL] Force user specified schema to the nullable one

HyukjinKwon:
## What changes were proposed in this pull request?

This PR proposes to force the user-specified schema to the nullable one, as Spark SQL can't validate it.

## How was this patch tested?

Unit tests adde in `FileStreamSourceSuite.scala` and `DataFrameReaderWriterSuite.scala`.

PR #15767, opened 2016-11-04
[SPARK-18269][SQL] CSV datasource should read null properly when schema is lager than parsed tokens

HyukjinKwon:

Currently, there are the three cases when reading CSV by datasource when it is `PERMISSIVE` parse mode.

- schema == parsed tokens (from each line)
  No problem to cast the value in the tokens to the field in the schema as they are equal.

- schema < parsed tokens (from each line)
  It slices the tokens into the number of fields in schema.  

- schema > parsed tokens (from each line)
  It appends `null` into parsed tokens so that safely values can be casted with the schema.

However, when `null` is appended in the third case, we should take `null` into account when casting the values.

In case of `StringType`, it is fine as `UTF8String.fromString(datum)` produces `null` when the input is `null`. Therefore, this case will happen only when schema is explicitly given and schema includes data types that are not `StringType`.

The codes below:

```scala
val path = "/tmp/a"
Seq("1").toDF().write.text(path.getAbsolutePath)
val schema = StructType(
  StructField("a", IntegerType, true) ::
  StructField("b", IntegerType, true) :: Nil)
spark.read.schema(schema).option("header", "false").csv(path).show()
```

prints

**Before**

```
java.lang.NumberFormatException: null
at java.lang.Integer.parseInt(Integer.java:542)
at java.lang.Integer.parseInt(Integer.java:615)
at scala.collection.immutable.StringLike$class.toInt(StringLike.scala:272)
at scala.collection.immutable.StringOps.toInt(StringOps.scala:29)
at org.apache.spark.sql.execution.datasources.csv.CSVTypeCast$.castTo(CSVInferSchema.scala:24)
```

**After**

```
+---+----+
|  a|   b|
+---+----+
|  1|null|
+---+----+
```

## How was this patch tested?

Unit test in `CSVSuite.scala` and `CSVTypeCastSuite.scala`

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/csv/CSVInferSchema.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/csv/CSVInferSchema.scala
@@ -221,13 +221,18 @@ private[csv] object CSVTypeCast {
    * Currently we do not support complex types (ArrayType, MapType, StructType).
    *
    * For string types, this is simply the datum. For other types.
-   * For other nullable types, this is null if the string datum is empty.
+   * For other nullable types, returns null if it is null or equals to the value specified
+   * in `nullValue` option.
    *
    * @param datum string value
-   * @param castType SparkSQL type
+   * @param name field name in schema.
+   * @param castType data type to cast `datum` into.
+   * @param nullable nullability for the field.
+   * @param options CSV options.
    */
   def castTo(
       datum: String,
+      name: String,
       castType: DataType,
       nullable: Boolean = true,
       options: CSVOptions = CSVOptions()): Any = {
@@ -235,7 +240,7 @@ private[csv] object CSVTypeCast {
     // datum can be null if the number of fields found is less than the length of the schema
     if (datum == options.nullValue || datum == null) {
       if (!nullable) {
-        throw new RuntimeException("null value found but the field is not nullable.")
+        throw new RuntimeException(s"null value found but field $name is not nullable.")
       }
       null
     } else {
@@ -285,7 +290,7 @@ private[csv] object CSVTypeCast {
               DateTimeUtils.millisToDays(DateTimeUtils.stringToTime(datum).getTime)
             }
         case _: StringType => UTF8String.fromString(datum)
-        case udt: UserDefinedType[_] => castTo(datum, udt.sqlType, nullable, options)
+        case udt: UserDefinedType[_] => castTo(datum, name, udt.sqlType, nullable, options)
         case _ => throw new RuntimeException(s"Unsupported type: ${castType.typeName}")
       }
     }
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/csv/CSVInferSchema.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/csv/CSVInferSchema.scala
@@ -231,8 +231,8 @@ private[csv] object CSVTypeCast {
       castType: DataType,
       nullable: Boolean = true,
       options: CSVOptions = CSVOptions()): Any = {
-
-    if (datum == null || nullable && datum == options.nullValue) {
+    val isNull = datum == options.nullValue || datum == null
+    if (nullable && isNull) {
       null
     } else {
       castType match {
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/csv/CSVInferSchema.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/csv/CSVInferSchema.scala
@@ -232,8 +232,11 @@ private[csv] object CSVTypeCast {
       nullable: Boolean = true,
       options: CSVOptions = CSVOptions()): Any = {
 
-    val isNull = datum == options.nullValue || datum == null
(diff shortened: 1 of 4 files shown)


rxin:
Can you get rid of those links from the pr description? They become stale immediately after merging this.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/csv/CSVInferSchema.scala
```diff
@@ -232,7 +232,8 @@ private[csv] object CSVTypeCast {
       nullable: Boolean = true,
       options: CSVOptions = CSVOptions()): Any = {
 
-    if (nullable && datum == options.nullValue) {
+    val isNull = datum == options.nullValue || datum == null
+    if (nullable && isNull) {
```
rxin: isn't this more clear if you do
```
// datum can be null if the number of fields found is less than the length of the schema
if (datum == options.nullValue || datum == null) {
  if (!nullable) {
    throw some exception saying field is not null but null value found
  }
  null
} else {
  ..
}
```

HyukjinKwon: Sure, I will.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/csv/CSVInferSchema.scala
```diff
@@ -232,7 +232,11 @@ private[csv] object CSVTypeCast {
       nullable: Boolean = true,
       options: CSVOptions = CSVOptions()): Any = {
 
-    if (nullable && datum == options.nullValue) {
+    // datum can be null if the number of fields found is less than the length of the schema
+    if (datum == options.nullValue || datum == null) {
+      if (!nullable) {
+        throw new RuntimeException("null value found but the field is not nullable.")
```
srowen: Nit: This could be `require(nullable, ...)` which would throw a better exception, `IllegalArgumentException`, too. (Even NPE would be reasonable.) But I don't feel strongly about it.

HyukjinKwon: Let me please leave as it is because reading/writing in CSV and JSON throw `RuntimeException` during execution in similar cases (if you are fine with this).

HyukjinKwon:
retest this please

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/csv/CSVInferSchema.scala
```diff
@@ -232,7 +232,9 @@ private[csv] object CSVTypeCast {
       nullable: Boolean = true,
       options: CSVOptions = CSVOptions()): Any = {
 
-    if (nullable && datum == options.nullValue) {
+    // datum can be null if the number of fields found is less than the length of the schema
+    if (datum == options.nullValue || datum == null) {
+      require(nullable, "null value found but the field is not nullable.")
```
rxin: sorry it doesn't make sense to throw illegalargumentexception at runtime here. What's the illegal argument? It is usually used when some parameters are out of bound or invalid. What's happening here is that we found exceptions during runtime that invalidates the assumption the user made (the field is nullable).

HyukjinKwon: Oh, I thought users set a non-nullable field in the schema inappropriate for data having null values and that is an illigal argument.. FWIW, `IllegalArgumentException` seems extending `RuntimeException` too..
Let me revert it back. I don't feel strongly about this.

HyukjinKwon: (Other related PRs with the comment, https://github.com/apache/spark/pull/15767#discussion_r86675671, are  https://github.com/apache/spark/pull/15329  and https://github.com/apache/spark/pull/14124)

HyukjinKwon:
retest this please

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/csv/CSVInferSchema.scala
rxin: Sorry I thought more about this - the current error message doesn't give the user a way to know which field is causing the problem. Can you add at least the field name to this?

HyukjinKwon: Sure, I can. Will try to make this neat up.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/csv/CSVInferSchema.scala
srowen: @rxin well, the user input is the illegal argument I guess, but I don't feel strongly about it. I was really arguing against `RuntimeException` which is never really the right thing to throw. `NullPointerException`? `IllegalStateException`? take your pick of anything else more specific.

rxin:

HyukjinKwon:
Thank you @rxin!

PR #15769, opened 2016-11-04
[SPARK-18191][CORE] Port RDD API to use commit protocol

jiangxb1987:

This PR port RDD API to use commit protocol, the changes made here:
1. Add new internal helper class that saves an RDD using a Hadoop OutputFormat named `SparkNewHadoopWriter`, it's similar with `SparkHadoopWriter` but uses commit protocol. This class supports the newer `mapreduce` API, instead of the old `mapred` API which is supported by `SparkHadoopWriter`;
2. Rewrite `PairRDDFunctions.saveAsNewAPIHadoopDataset` function, so it uses commit protocol now.

## How was this patch tested?
Exsiting test cases.

### DIFF
--- a/core/src/main/scala/org/apache/spark/SparkHadoopWriter.scala
+++ b/core/src/main/scala/org/apache/spark/SparkHadoopWriter.scala
@@ -162,9 +162,12 @@ class SparkHadoopWriter(jobConf: JobConf) extends Logging with Serializable {
 private[spark]
 object SparkHadoopWriter {
   def createJobID(time: Date, id: Int): JobID = {
-    val formatter = new SimpleDateFormat("yyyyMMddHHmmss", Locale.US)
-    val jobtrackerID = formatter.format(time)
+    val jobtrackerID = createJobTrackerID(time)
     new JobID(jobtrackerID, id)
+  }
+
+  def createJobTrackerID(time: Date): String = {
+    new SimpleDateFormat("yyyyMMddHHmmss", Locale.US).format(time)
   }
 
   def createPathFromString(path: String, conf: JobConf): Path = {
--- a/core/src/main/scala/org/apache/spark/SparkHadoopWriter.scala
+++ b/core/src/main/scala/org/apache/spark/SparkHadoopWriter.scala
@@ -23,11 +23,11 @@ import java.text.SimpleDateFormat
 import java.util.{Date, Locale}
 
 import org.apache.hadoop.fs.FileSystem
-import org.apache.hadoop.fs.Path
 import org.apache.hadoop.mapred._
 import org.apache.hadoop.mapreduce.TaskType
 
 import org.apache.spark.internal.Logging
+import org.apache.spark.internal.io.SparkNewHadoopWriterUtils
 import org.apache.spark.mapred.SparkHadoopMapRedUtil
 import org.apache.spark.rdd.HadoopRDD
 import org.apache.spark.util.SerializableJobConf
@@ -153,32 +153,8 @@ class SparkHadoopWriter(jobConf: JobConf) extends Logging with Serializable {
     splitID = splitid
     attemptID = attemptid
 
-    jID = new SerializableWritable[JobID](SparkHadoopWriter.createJobID(now, jobid))
+    jID = new SerializableWritable[JobID](SparkNewHadoopWriterUtils.createJobID(now, jobid))
     taID = new SerializableWritable[TaskAttemptID](
         new TaskAttemptID(new TaskID(jID.value, TaskType.MAP, splitID), attemptID))
-  }
-}
-
-private[spark]
-object SparkHadoopWriter {
-  def createJobID(time: Date, id: Int): JobID = {
-    val jobtrackerID = createJobTrackerID(time)
-    new JobID(jobtrackerID, id)
-  }
-
-  def createJobTrackerID(time: Date): String = {
-    new SimpleDateFormat("yyyyMMddHHmmss", Locale.US).format(time)
-  }
-
-  def createPathFromString(path: String, conf: JobConf): Path = {
-    if (path == null) {
-      throw new IllegalArgumentException("Output path is null")
-    }
-    val outputPath = new Path(path)
-    val fs = outputPath.getFileSystem(conf)
-    if (fs == null) {
-      throw new IllegalArgumentException("Incorrectly formatted output path")
-    }
-    outputPath.makeQualified(fs.getUri, fs.getWorkingDirectory)
   }
 }
--- a/core/src/main/scala/org/apache/spark/SparkHadoopWriter.scala
+++ b/core/src/main/scala/org/apache/spark/SparkHadoopWriter.scala
@@ -27,7 +27,7 @@ import org.apache.hadoop.mapred._
 import org.apache.hadoop.mapreduce.TaskType
 
 import org.apache.spark.internal.Logging
-import org.apache.spark.internal.io.SparkNewHadoopWriterUtils
+import org.apache.spark.internal.io.SparkHadoopWriterUtils
 import org.apache.spark.mapred.SparkHadoopMapRedUtil
 import org.apache.spark.rdd.HadoopRDD
 import org.apache.spark.util.SerializableJobConf
@@ -153,7 +153,7 @@ class SparkHadoopWriter(jobConf: JobConf) extends Logging with Serializable {
     splitID = splitid
     attemptID = attemptid
 
-    jID = new SerializableWritable[JobID](SparkNewHadoopWriterUtils.createJobID(now, jobid))
+    jID = new SerializableWritable[JobID](SparkHadoopWriterUtils.createJobID(now, jobid))
     taID = new SerializableWritable[TaskAttemptID](
         new TaskAttemptID(new TaskID(jID.value, TaskType.MAP, splitID), attemptID))
   }
--- a/core/src/main/scala/org/apache/spark/SparkNewHadoopWriter.scala
+++ b/core/src/main/scala/org/apache/spark/SparkNewHadoopWriter.scala
@@ -1,0 +1,132 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.spark
+
+import java.util.Date
+
+import org.apache.hadoop.conf.Configuration
+import org.apache.hadoop.mapreduce._
+import org.apache.hadoop.mapreduce.task.TaskAttemptContextImpl
+import org.apache.spark.internal.Logging
+import org.apache.spark.internal.io.HadoopMapReduceCommitProtocol
+import org.apache.spark.util.SerializableConfiguration
+
+/**
+ * Internal helper class that saves an RDD using a Hadoop OutputFormat
+ * (from the newer mapreduce API, not the old mapred API).
+ *
+ * Saves the RDD using a JobConf, which should contain an output key class, an output value class,
+ * a filename to write to, etc, exactly like in a Hadoop MapReduce job.
+ *
+ * Use a [[HadoopMapReduceCommitProtocol]] to handle output commit, which, unlike Hadoop's
+ * OutputCommitter, is serializable.
+ */
+private[spark]
+class SparkNewHadoopWriter(
+    jobConf: Configuration,
+    committer: HadoopMapReduceCommitProtocol) extends Logging with Serializable {
+
+  private val now = new Date()
+  private val conf = new SerializableConfiguration(jobConf)
+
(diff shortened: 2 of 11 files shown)


core/src/main/scala/org/apache/spark/SparkHadoopWriter.scala
```diff
@@ -162,11 +162,14 @@ class SparkHadoopWriter(jobConf: JobConf) extends Logging with Serializable {
...
-    val formatter = new SimpleDateFormat("yyyyMMddHHmmss", Locale.US)
-    val jobtrackerID = formatter.format(time)
+    val jobtrackerID = createJobTrackerID(time)
     new JobID(jobtrackerID, id)
   }
 
+  def createJobTrackerID(time: Date): String = {
+    new SimpleDateFormat("yyyyMMddHHmmss", Locale.US).format(time)
+  }
+
```
jiangxb1987: We need to generate jobTrackerID seprately in `SparkNewHadoopWriter `

core/src/test/scala/org/apache/spark/rdd/PairRDDFunctionsSuite.scala
```diff
@@ -509,21 +509,6 @@ class PairRDDFunctionsSuite extends SparkFunSuite with SharedSparkContext {
       (2, ArrayBuffer(1))))
   }
 
-  test("saveNewAPIHadoopFile should call setConf if format is configurable") {
```
jiangxb1987: This logic is no longer needed.

core/src/test/scala/org/apache/spark/rdd/PairRDDFunctionsSuite.scala
```diff
@@ -725,8 +710,7 @@ class PairRDDFunctionsSuite extends SparkFunSuite with SharedSparkContext {
 }
 
 /*
-  These classes are fakes for testing
-    "saveNewAPIHadoopFile should call setConf if format is configurable".
+  These classes are fakes for testing saveAsHadoopFile/saveNewAPIHadoopFile.
```
jiangxb1987: This comment have been out-of-date for a while.

jiangxb1987 (commented):
I didn't touch `SparkHadoopWriter` and `PairRDDFunctions.saveAsHadoopDataset`, because they use the older `mapred` API, which is not supported by the `FileCommitProtocol` framework now.

core/src/main/scala/org/apache/spark/SparkNewHadoopWriter.scala
```diff
@@ -0,0 +1,133 @@
...
+ * (from the newer mapreduce API, not the old mapred API).
+ *
+ * Saves the RDD using a JobConf, which should contain an output key class, an output value class,
+ * a filename to write to, etc, exactly like in a Hadoop MapReduce job.
+ *
+ * Use a [[HadoopMapReduceCommitProtocol]] to handle output commit, which, unlike Hadoop's
+ * OutputCommitter, is serializable.
+ */
+private[spark]
+class SparkNewHadoopWriter(
```
rxin: move this into internal/io

rxin:
Doesn't need to be in this PR, but can you also implement a HadoopMapRedCommitProtocol that supports the older mapred package's commiter?

jiangxb1987:
Moved the logic from `saveAsNewAPIHadoopDataset` into `SparkNewHadoopWriter`, also added `SparkNewHadoopWriterUtils` to store functions used by Hadoop/NewHadoop writer.
In a followup PR, I'll implement a similar protocol to support the older mapred package's committer.

core/src/main/scala/org/apache/spark/rdd/PairRDDFunctions.scala
```diff
@@ -1092,37 +1093,38 @@ class PairRDDFunctions[K, V](self: RDD[(K, V)])
       jobFormat.checkOutputSpecs(job)
     }
 
+    // Instantiate writer
+    val committer = FileCommitProtocol.instantiate(
+      className = classOf[HadoopMapReduceCommitProtocol].getName,
+      jobId = stageId.toString,
```
jiangxb1987: In fact I'm not sure what value should be assigned to `jobId` here, should it be `jobTrackerId` combined with `jobId`, or any value else? I failed to find some code to follow on this topic.

core/src/main/scala/org/apache/spark/SparkHadoopWriter.scala
```diff
@@ -23,11 +23,11 @@ import java.text.SimpleDateFormat
 import java.util.{Date, Locale}
 
 import org.apache.hadoop.fs.FileSystem
-import org.apache.hadoop.fs.Path
 import org.apache.hadoop.mapred._
 import org.apache.hadoop.mapreduce.TaskType
 
 import org.apache.spark.internal.Logging
```
rxin: i'd move this file into spark/internal/io as well to be closer to SparkNewHadoopWriter.scala

rxin: we can do it in a separate pr when you do the mapred committer.

core/src/main/scala/org/apache/spark/internal/io/SparkNewHadoopWriter.scala
```diff
@@ -0,0 +1,248 @@
...
+
+/**
+ * A helper object that saves an RDD using a Hadoop OutputFormat
+ * (from the newer mapreduce API, not the old mapred API).
+ */
+private[spark]
+object SparkNewHadoopWriter extends Logging {
+
+  /** A shared job description for all the write tasks. */
+  private class WriteJobDescription[K, V](
```
rxin: maybe just remove this, since you have only 3 items here. In SQL there were a lot of items.

core/src/main/scala/org/apache/spark/internal/io/SparkNewHadoopWriter.scala
```diff
@@ -0,0 +1,248 @@
...
+   *    rows within an RDD partition.
+   * 3. If no exception is thrown in a task, commits that task, otherwise aborts that task;  If any
+   *    exception is thrown during task commitment, also aborts that task.
+   * 4. If all tasks are committed, commit the job, otherwise aborts the job;  If any exception is
+   *    thrown during job commitment, also aborts the job.
+   */
+  def write[K, V: ClassTag](
+      sparkContext: SparkContext,
+      rdd: RDD[(K, V)],
+      committer: HadoopMapReduceCommitProtocol,
```
rxin: i'd move the creation of the commit protocol here here. The reason I put it outside in SQL was because streaming and batch needed to specify different protocols, but that problem doesn't exist in core.

core/src/main/scala/org/apache/spark/SparkHadoopWriter.scala
jiangxb1987: Yes I'll do it later.

jiangxb1987:
The failed test case is imported by #15725, which is not related to our changes in this PR.

core/src/main/scala/org/apache/spark/internal/io/SparkHadoopMapReduceWriter.scala
```diff
@@ -0,0 +1,242 @@
...
+        s"$committer may be an output committer that writes data directly to " +
+          "the final location. Because speculation is enabled, this output committer may " +
+          "cause data loss (see the case in SPARK-10063). If possible, please use an output " +
+          "committer that does not have this behavior (e.g. FileOutputCommitter)."
+      logWarning(warningMessage)
+    }
+
+    // Try to write all RDD partitions as a Hadoop OutputFormat.
+    try {
+      sparkContext.runJob(rdd, (context: TaskContext, iter: Iterator[(K, V)]) => {
```
rxin: we need to collect the result coming from the commit protocol here, and pass it into commitJob, don't we?

rxin:
The rest looks good.

cc @ericl for another look too

rxin:
Also @mridulm who has looked at one of the prs for commit protocol too.

@mridulm can you help review this too?

jiangxb1987:
retest this please.

jiangxb1987:
@rxin I've addressed your comment. Seems the test case is still failing... @ericl

rxin:
I've disabled the test.

core/src/main/scala/org/apache/spark/internal/io/SparkHadoopMapReduceWriter.scala
```diff
@@ -0,0 +1,243 @@
...
+        logError(s"Task ${taskContext.getTaskAttemptID} aborted.")
+      }, finallyBlock = writer.close(taskContext))
+    } catch {
+      case t: Throwable =>
+        throw new SparkException("Task failed while writing rows", t)
+    } finally {
+      outputMetricsAndBytesWrittenCallback.foreach { case (om, callback) =>
+        om.setBytesWritten(callback())
+        om.setRecordsWritten(recordsWritten)
+      }
```
mridulm: This looks like a behavior change - metric's are getting updated even when exceptions are thrown.
Do it after Utils.tryWithSafeFinallyAndFailureCallbacks completes, not in finally

core/src/main/scala/org/apache/spark/internal/io/SparkHadoopMapReduceWriter.scala
```diff
@@ -0,0 +1,243 @@
...
+
+    // Set up a job.
+    val jobTrackerId = SparkHadoopWriterUtils.createJobTrackerID(new Date())
+    val jobAttemptId = new TaskAttemptID(jobTrackerId, stageId, TaskType.MAP, 0, 0)
+    val jobContext = new TaskAttemptContextImpl(conf.value, jobAttemptId)
+    val format = jobContext.getOutputFormatClass
+
+    if (SparkHadoopWriterUtils.isOutputSpecValidationEnabled(sparkConf)) {
+      // FileOutputFormat ignores the filesystem parameter
+      val jobFormat = format.newInstance
```
mridulm: If it is Configurable, we should invoke setConf on it - the earlier code was doing this.
The tests for that seem to have been removed as well.
Any reason why ?

core/src/main/scala/org/apache/spark/internal/io/SparkHadoopMapReduceWriter.scala
```diff
@@ -0,0 +1,243 @@
...
+      recordsWritten: Long): Unit = {
+    if (recordsWritten % RECORDS_BETWEEN_BYTES_WRITTEN_METRIC_UPDATES == 0) {
+      outputMetricsAndBytesWrittenCallback.foreach { case (om, callback) =>
+        om.setBytesWritten(callback())
+        om.setRecordsWritten(recordsWritten)
+      }
+    }
+  }
+
+  val RECORDS_BETWEEN_BYTES_WRITTEN_METRIC_UPDATES = 256
```
mridulm: private and move it to top of the object ?

mridulm (commented):
Overall looks good, added a few comments though.

core/src/main/scala/org/apache/spark/internal/io/SparkHadoopMapReduceWriter.scala
jiangxb1987: The logic is to create a `OutputCommitter` from the given hadoop conf, which is now handled by `HadoopMapReduceCommitProtocol`, so this code and corresponding tests are no longer needed.

jiangxb1987:
@mridulm I've added most of your comments. Thank you!

rxin:
Thanks - merging in master.

core/src/main/scala/org/apache/spark/internal/io/SparkHadoopMapReduceWriter.scala
mridulm: I think I added the comment to the wrong location (though it is relevant here, it is probably less serious ?).
HadoopMapReduceCommitProtocol.setupCommitter() should be doing a setConf if OutputFormat is Configurable.

This needs to be fixed to ensure custom OutputFormat's work.
I see that the PR has already been committed - can you please file a bug and fix it ?
The test will also need to be re-added.

jiangxb1987: Oh...I see the problem now. Will add that in a follow up, thank you for clarifying.

core/src/main/scala/org/apache/spark/internal/io/SparkHadoopMapReduceWriter.scala
```diff
@@ -0,0 +1,249 @@
...
+   *    exception is thrown during task commitment, also aborts that task.
+   * 4. If all tasks are committed, commit the job, otherwise aborts the job;  If any exception is
+   *    thrown during job commitment, also aborts the job.
+   */
+  def write[K, V: ClassTag](
+      rdd: RDD[(K, V)],
+      hadoopConf: Configuration): Unit = {
+    // Extract context and configuration from RDD.
+    val sparkContext = rdd.context
+    val stageId = rdd.id
```
rezasafi: Is this accurate? Seems weird that satgeId is set to be equal to rdd.id. What is the commit protocol here?

jiangxb1987: It follows the previous behavior, what's your concern here?

rezasafi: We had a test failure using Spark 2.2 that seems to happen after this commit. The failure didn't occur before using Spark 2.0. We thought there could be a problem since the jobContext passed to commiter.setupJob() (line 84) has a different JobID comparing to the task context that is passed to to commiter.setupTask() (line 126)  in the former, it comes from rdd.id and in the later, its from stage.id. Just wanted to check what is the protocol here that requires such a difference? Thanks.

rezasafi: I investigated the problem a little more and I filed a jira & fix, see SPARK-22162. Thank you in advance.

PR #15770, opened 2016-11-04
[SPARK-15784][ML]:Add Power Iteration Clustering to spark.ml

wangmiao1981:

As we discssed in the JIRA, `PowerIterationClustering` is added as a `Transformer`. The `featureCol` is `vector` type. In the `transform` method, it calls `MLlibPowerIterationClustering().run(rdd)` method and transforms the return value `assignments` (the Kmeans output of the pseudo-eigenvector) as a Dataframe (`id`: `LongType`, `cluster`: `IntegerType`).

## How was this patch tested?
Add new unit tests similar to `MLlibPowerIterationClustering`.


wangmiao1981:
ping @yanboliang

yanboliang:
@wangmiao1981 Thanks for working on this. Since we start QA for 2.1, I will review this after the release.

mllib/src/main/scala/org/apache/spark/ml/clustering/PowerIterationClustering.scala
```diff
@@ -0,0 +1,182 @@
...
+  def getK: Int = $(k)
+
+  /**
+   * Param for the initialization algorithm. This can be either "random" to use a random vector
+   * as vertex properties, or "degree" to use normalized sum similarities. Default: random.
+   */
+  @Since("2.2.0")
+  final val initMode = new Param[String](this, "initMode", "The initialization algorithm. " +
+    "Supported options: 'random' and 'degree'.",
+    (value: String) => validateInitMode(value))
```
zhengruifeng: What about use validator `ParamValidators.inArray[String](...)` instead?

wangmiao1981: what is the difference? Thanks!

mllib/src/main/scala/org/apache/spark/ml/clustering/PowerIterationClustering.scala
```diff
@@ -0,0 +1,182 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.spark.ml.clustering
+
+import org.apache.spark.annotation.{Experimental, Since}
+import org.apache.spark.ml.Transformer
+import org.apache.spark.ml.linalg.{Vector}
```
thunterdb: no need for brackets

mllib/src/main/scala/org/apache/spark/ml/clustering/PowerIterationClustering.scala
```diff
@@ -0,0 +1,182 @@
...
+import org.apache.spark.ml.param._
+import org.apache.spark.ml.param.shared._
+import org.apache.spark.ml.util._
+import org.apache.spark.mllib.clustering.{PowerIterationClustering => MLlibPowerIterationClustering}
+import org.apache.spark.mllib.clustering.PowerIterationClustering.Assignment
+import org.apache.spark.rdd.RDD
+import org.apache.spark.sql.{DataFrame, Dataset, Row}
+import org.apache.spark.sql.functions.{col}
```
thunterdb: same here

mllib/src/main/scala/org/apache/spark/ml/clustering/PowerIterationClustering.scala
thunterdb: You do not need to use write a function as you do below after that, it will allow more user-friendly error messages in the future.

mllib/src/main/scala/org/apache/spark/ml/clustering/PowerIterationClustering.scala
```diff
@@ -0,0 +1,182 @@
...
+  /**
+   * Param for the initialization algorithm. This can be either "random" to use a random vector
+   * as vertex properties, or "degree" to use normalized sum similarities. Default: random.
+   */
+  @Since("2.2.0")
+  final val initMode = new Param[String](this, "initMode", "The initialization algorithm. " +
+    "Supported options: 'random' and 'degree'.",
+    (value: String) => validateInitMode(value))
+
+  private[spark] def validateInitMode(initMode: String): Boolean = {
```
thunterdb: No need with comment above

mllib/src/main/scala/org/apache/spark/ml/clustering/PowerIterationClustering.scala
```diff
@@ -0,0 +1,182 @@
...
+ *
+ * Note that we implement [[PowerIterationClustering]] as a transformer. The [[transform]] is an
+ * expensive operation, because it uses PIC algorithm to cluster the whole input dataset.
+ *
+ * @see [[http://en.wikipedia.org/wiki/Spectral_clustering Spectral clustering (Wikipedia)]]
+ */
+@Since("2.2.0")
+@Experimental
+class PowerIterationClustering private[clustering] (
+                                @Since("2.2.0") override val uid: String)
```
thunterdb: indentation

mllib/src/main/scala/org/apache/spark/ml/clustering/PowerIterationClustering.scala
```diff
@@ -0,0 +1,182 @@
...
+  val idCol = new Param[String](this, "idCol", "column name for ids.")
+
+  /** @group getParam */
+  def getIdCol: String = $(idCol)
+
+  /**
+   * Validates the input schema
+   * @param schema input schema
+   */
+  protected def validateSchema(schema: StructType): Unit = {
```
thunterdb: Instead of just validating the schema, we should validate and transform. You can follow the example in https://github.com/apache/spark/blob/master/mllib/src/main/scala/org/apache/spark/ml/clustering/KMeans.scala#L92

mllib/src/main/scala/org/apache/spark/ml/clustering/PowerIterationClustering.scala
```diff
@@ -0,0 +1,182 @@
...
+  /** @group expertGetParam */
+  @Since("2.2.0")
+  def getInitMode: String = $(initMode)
+
+  /**
+   * Param for the column name for ids returned by [[PowerIterationClustering.transform()]].
+   * Default: "id"
+   * @group param
+   */
+  val idCol = new Param[String](this, "idCol", "column name for ids.")
```
thunterdb: Instead of making an 'id' column, which does not convey much information, we should follow the example of `K-Means` and call it `prediction`. You already include the trait for that.

mllib/src/main/scala/org/apache/spark/ml/clustering/PowerIterationClustering.scala
```diff
@@ -0,0 +1,182 @@
...
+  /** @group setParam */
+  @Since("2.2.0")
+  def setMaxIter(value: Int): this.type = set(maxIter, value)
+
+  /** @group setParam */
+  @Since("2.2.0")
+  def setIdCol(value: String): this.type = set(idCol, value)
+
+  @Since("2.2.0")
+  override def transform(dataset: Dataset[_]): DataFrame = {
```
thunterdb: One issue with the current implementation is that extra columns are going to be discarded (this is also the case with most graphframes algorithms), which breaks the general contract of transformers. After making the transform, you should join back on the missing columns.
Also, a test for this use case would be useful.

mllib/src/test/scala/org/apache/spark/ml/clustering/PowerIterationClusteringSuite.scala
```diff
@@ -0,0 +1,153 @@
...
+  test("parameters validation") {
+    intercept[IllegalArgumentException] {
+      new PowerIterationClustering().setK(1)
+    }
+    intercept[IllegalArgumentException] {
+      new PowerIterationClustering().setInitMode("no_such_a_mode")
+    }
+  }
+
+  test("power iteration clustering") {
```
thunterdb: can you also add a test with a dataframe that has some extra data in it?

wangmiao1981:
@thunterdb Thanks for your review! I will address the comments soon.

thunterdb:
@wangmiao1981 thanks a lot! I would be very happy to see that PR in Spark 2.2 and I will gladly help you for that.

mllib/src/main/scala/org/apache/spark/ml/clustering/PowerIterationClustering.scala
wangmiao1981: In the MLLIB implementation, the clustering result is `case class Assignment(id: Long, cluster: Int)`. `idCol` is the node id and the `cluster` is the cluster id that this node belongs to. I think `id` is still useful to represent the node. Otherwise, we need to make sure that the output order of nodes is the same as input order, which means `id` is implicitly inferred from the Row Number.

mllib/src/main/scala/org/apache/spark/ml/clustering/PowerIterationClustering.scala
wangmiao1981: The missing column is the `featureCol`. You want the `featureCol` is in the result `DataFrame`. Right? Thanks!

wangmiao1981: @thunterdb The prediction/transformation result `(id, cluster)` has different size of the `features`. It is because `id` denotes the node id in the graph, which is equal to the number of nodes in the graph. `features` are the graph weights and `|features|` is roughly `|nodes|^2/2`. So I can't directly join the result dataframe with the input dataset. Please correct me if I misunderstood your comments. Thanks!

mllib/src/main/scala/org/apache/spark/ml/clustering/PowerIterationClustering.scala
wangmiao1981: PIC is different compared with K-Means. K-Means `transform` applies `predict` method to each row of the input (i.e., each data point). While PIC `run` is applying the K-Means to the pseudo-eigenvector from `powerIter` method. This is not one-to-one map from Input dataset to result. Please also see the comments below.

mllib/src/test/scala/org/apache/spark/ml/clustering/PowerIterationClusteringSuite.scala
wangmiao1981: Add it.

thunterdb:
You are right, I had forgotten that for this algorithm, the input is the edges, and the output is the label for each of the vertices.

This is a tricky algorithm to put as a transformer, since it does not follow the usual convention that data should only be appended to the dataframe. I suggest we follow the same example as ALS the mllib implementation of PIC:
 - let's make it an estimator that returns a model: the model contains the labels for each of the points in a dataframe (the current output of transform)
 - the model's transform method now takes points with an id, and joins it with the models to append a column of labels. This is the same as ALS.

If we do not follow this pattern, then the model selection algorithms are not going to work. What do you think?

wangmiao1981:
@thunterdb Thanks for your response. In the original JIRA, we have discussed why we want it to be a transformer. Let me find it and post it here.

wangmiao1981:
Joseph K. Bradley added a comment - 31/Oct/16 18:14

Miao Wang Sorry for the slow response here. I do want us to add PIC to spark.ml, but we should discuss the design before the PR. Could you please close the PR for now but save the branch to re-open after discussion?

Let's have a design discussion first.

I agree that the big issue is that there isn't a clear way to make predictions on new data points. In fact, I've never heard of people trying to do so. Has anyone else?

Assuming that prediction is not meaningful for PIC, then I don't think the algorithm fits within the Pipeline framework, though it's debatable. I see a few options:

    Put PIC in Pipelines as a Transformer, not an Estimator. We would just need to document that it is a very expensive Transformer.
    Put PIC in spark.ml as a static method. We may have to do this anyways to support all of spark.mllib's Statistics.
    Put PIC in GraphFrames (and push harder for GraphFrames to be merged back into Spark, which will include a much longer set of improvements).

My top choice is PIC as a Transformer. What do you think?

CC Yanbo Liang Seth Hendrickson Nick Pentreath opinions?
sethah Seth Hendrickson added a comment - 31/Oct/16 22:40

This seems like it fits the framework of a feature transformer. We could generate a real-valued feature column using PIC algorithm where the values are just the components of the pseudo-eigenvector. Alternatively we could pipeline a KMeans clustering on the end, but I think it makes more sense to let users do that themselves - but that's up for debate.

wangmiao1981:
Yanbo Liang added a comment - 02/Nov/16 09:30 - edited

I'm prefer to #1 and #3, but it looks like we can achieve both goals.
Graph can be represented by GraphX/GraphFrame or DataFrame/RDD. PIC model can be trained on both of them, but we use GraphX operators in the internal implementation which means input data should be converted to GraphX representation if it's RDD of tuples. So it's straight forward to make PIC as one of the algorithms in GraphX(or GraphFrame when it is merged back into Spark). However, users may load their graph as DataFrame/RDD and transform via ML Pipeline which should also be supported, so it's better we can wrap PIC of GraphX/GraphFrame as an Pipeline stage and then ML users can use it as well.
For some historical reasons(we don't want to add new features to GraphX), I propose to split this task into the following step:

    Put PIC in Pipeline as a Transformer, use the GraphX operators in the implementation (This is consistent with Joseph K. Bradley's proposal).
    Add PIC algorithms to GraphFrames when it is merged into Spark.
    Make the ML PIC as a wrapper to call the GraphFrames PIC implementation.

I think this scenario should be better for different users(ML users and GraphFrames users), but still open to hear your thoughts. Thanks.

wangmiao1981:
I am checking ALS out to understand your suggestions. Thanks!

thunterdb:
@wangmiao1981 yes I had seen the discussions there. I believe that eventually PIC should be moved into graphframes, but we can have a simple API in `spark.ml` for the time being.

wangmiao1981:
@thunterdb Per discussion with Yanbo, there is one concern of making it an Estimator. For every `transform`, there is an additional data shuffle. cc @yanboliang @jkbradley Thanks!

yanboliang:
@thunterdb I have two concerns about making PIC as an Estimator:
* As your suggestion, model's transform will introduce extra data shuffle compared with the original implementation.
* It can not fit into the pipeline well, since both the estimator ```fit``` and model ```transform``` method should work on the dataset with same schema, like:
```
val model = new ALS().fit(dataset)
model.transform(dataset)
```
But in your suggestion, the input of PIC estimator is the edges or RDD[(Long, Long, Double)], the input of PIC model is the vertices ID. 

I totally agree making PIC as a Transformer is tricky, but I did not find a better way. Thanks.

wangmiao1981:
@thunterdb @yanboliang Do we reach an agreement on whether to make it a transformer or an estimator now?

jkbradley:
Sorry for my absence from recent conversation!

I agree there is no clear answer for handling input and output schema.  Some options:
* Option 1: same as RDD/GraphX-based API
  * Input: Each Row is 1 edge.  DataFrame represents a graph.
  * Output: [node ID, cluster ID]
* Option 2: take adjacency list
  * Input: Each Row has [node ID, list of neighbor IDs, list of neighbor edge weights].  DataFrame represents a graph.
  * Output: Append column for "cluster ID"
* Option 3: take feature vectors --> This is another use of PIC from the original paper.  Given feature vectors, compute a similarity matrix using some distance, and then run PIC as we do now.  We won't do this now, but we could add it as an option in the future.
  * Input: Each Row has [node ID, feature vector].  DataFrame does *not* represent a graph.
  * Output: Append columns for "cluster ID" and "adjacency list"

It sounds like Option 1 is the main one being considered right now.  What do you think of Option 2 though to keep it as a row->row Transformer?

wangmiao1981:
@jkbradley Thanks for your reply! I quickly go through your suggestions. If I understand correctly, you prefer making it a `Transformer`, as we previously discussed, but changing the input data to fit into to the pipeline. Right? Let me think about details and evaluate each options before making the changes. Thanks!

jkbradley:
Yep, that's correct.  Everyone, please let me know if you disagree.

Also, if we do go with Option 2 above, then the input schema could be a few possible things:
* list of (neighbor ID, weight)
  * pro: ensures # neighbors = # weights
  * con: nesting can be annoying to work with
* (list of neighbor IDs, list of weights)
  * pro: Easier to support unweighted graphs by omitting list of weights.  Avoid nesting.
  * con: could have invalid input with mismatched numbers of neighbors, weights
* Vector (of length # neighbors)
  * pro: Familiar data type.  Ensures # neighbors = # weights.
  * con: If we later support passing raw features which are used to compute a similarity matrix within PIC, then this could confuse users.

I'm ambivalent about these schema options.  What do you think?

thunterdb:
Note that any of these formats would cause trouble for a graph with high centrality (lady gaga in the twitter graph). That being said, I do not have a strong opinion as to which option we pick, in order to move things along.

wangmiao1981:
Option 2 doesn't break our pipeline scheme which only appends the result column to the input dataframe. Besides the discussions above, the graph is undirected and the weight list will appear twice for two nodes, while Option 1 only stores the weights once. 

Let me try Option 2 today.

Thanks!

wangmiao1981:
Update: 
Sorry for the delay. I am working on some other items.

Now, I am using  (node-id, list of neighbor IDs, list of weights) by adding two additional columns. 

I will push soon.

wangmiao1981:
@jkbradley I changed the input data format by using (list of neighbor IDs, list of weights), which are two columns of the input dataset. For the result, I appended the predicted cluster ids to the input dataset by joining the assignment dataset by col `id`. 

This is an initial implementation. Please take a look. Thanks!

wangmiao1981:
It passed at local. I will fix the issue.

wangmiao1981:
@jkbradley can you take a look? Thanks!

wangmiao1981:
gentle ping @jkbradley @yanboliang @thunterdb

felixcheung:
Any update on this?

wangmiao1981:
@yanboliang Can you take a look first? Thanks!

mllib/src/main/scala/org/apache/spark/ml/clustering/PowerIterationClustering.scala
```diff
@@ -0,0 +1,213 @@
...
+ */
+private[clustering] trait PowerIterationClusteringParams extends Params with HasMaxIter
+  with HasFeaturesCol with HasPredictionCol with HasWeightCol {
+
+  /**
+   * The number of clusters to create (k). Must be > 1. Default: 2.
+   * @group param
+   */
+  @Since("2.2.0")
+  final val k = new IntParam(this, "k", "The number of clusters to create. " +
```
WeichenXu123: change since to 2.3

mllib/src/main/scala/org/apache/spark/ml/clustering/PowerIterationClustering.scala
```diff
@@ -0,0 +1,213 @@
...
+  @Since("2.2.0")
+  def setNeighborCol(value: String): this.type = set(neighborCol, value)
+
+  @Since("2.2.0")
+  override def transform(dataset: Dataset[_]): DataFrame = {
+    val sparkSession = dataset.sparkSession
+
+    val rdd: RDD[(Long, Long, Double)] =
+      dataset.select(col($(idCol)), col($(neighborCol)), col($(weightCol))).rdd.map {
+        case Row(id: Long, nbr: Vector, weight: Vector) => (id, nbr, weight)
```
WeichenXu123: Why need this redundant `map` ?

mllib/src/main/scala/org/apache/spark/ml/clustering/PowerIterationClustering.scala
```diff
@@ -0,0 +1,213 @@
...
+    }.flatMap{ case (id, nbr, weight) =>
+        require(nbr.size == weight.size,
+          "The length of neighbor list must be equal to the the length of the weight list.")
+        val ids = Array.fill(nbr.size)(id)
+        ids.zip(nbr.toArray).zip(weight.toArray)}.map {case ((i, j), k) => (i, j.toLong, k)}
```
WeichenXu123: Avoid use `Array.zip`, it will generate temporary array. Use `while` instead.

wangmiao1981:
@WeichenXu123 Thanks for reviewing! I will address the comments soon.

mllib/src/main/scala/org/apache/spark/ml/clustering/PowerIterationClustering.scala
wangmiao1981: I need to transfer the `Row` data to the `RDD[(Long, Long, Double)]`. Any other ways of doing this? Thanks! @WeichenXu123

wangmiao1981: I think I get your point now. I can directly `flatMap`. Thanks!

wangmiao1981:
weird. Local style test passed. Anyway, I changed the order as required by Jenkins.

wangmiao1981:
Jenkins, retest please

wangmiao1981:
retest please

wangmiao1981:
info] Main Scala API documentation successful.
[error] (spark/javaunidoc:doc) javadoc returned nonzero exit code
[error] Total time: 95 s, completed Aug 15, 2017 4:59:59 PM
[error] running /home/jenkins/workspace/SparkPullRequestBuilder/build/sbt -Phadoop-2.6 -Pmesos -Pkinesis-asl -Pyarn -Phive-thriftserver -Phive unidoc ; received return code 1

It seems irrelevant.

wangmiao1981:
Jenkins, retest this please.

wangmiao1981:
@WeichenXu123 I have made changes based on your comments. Thanks!

wangmiao1981:
ping @WeichenXu123

wangmiao1981:
ping @WeichenXu123

WeichenXu123:
@wangmiao1981 Sorry for delay, I will take a look later, thanks!

mllib/src/main/scala/org/apache/spark/ml/clustering/PowerIterationClustering.scala
```diff
@@ -0,0 +1,216 @@
...
+   */
+  @Since("2.3.0")
+  def setNeighborCol(value: String): this.type = set(neighborCol, value)
+
+  @Since("2.3.0")
+  override def transform(dataset: Dataset[_]): DataFrame = {
+    val sparkSession = dataset.sparkSession
+    val rdd: RDD[(Long, Long, Double)] =
+      dataset.select(col($(idCol)), col($(neighborCol)), col($(weightCol))).rdd.flatMap {
+        case Row(id: Long, nbr: Vector, weight: Vector) =>
```
WeichenXu123: The PIC require input graph matrix to be symmetric, and the weight should be non-negative. It is better to check them here. But checking symmetric seems cost too much, I have no good idea for now. cc @jkbradley Do you have some thoughts ?

mllib/src/main/scala/org/apache/spark/ml/clustering/PowerIterationClustering.scala
```diff
@@ -0,0 +1,216 @@
...
+        require(nbr.size == weight.size,
+          "The length of neighbor list must be equal to the the length of the weight list.")
+        val ids = Array.fill(nbr.size)(id)
+        for (i <- 0 until ids.size) yield (ids(i), nbr(i).toLong, weight(i))}
```
WeichenXu123: This line code do not performs well, why you expand the var `id` into a array filled with the same value ?
You can use:
```
nbr.toArray.toIterator.zip(weight.toArray.toIterator).map(x => (id, x._1.toLong, x._2.toLong))
```

WeichenXu123 (commented):
And I concern about the case of high centrality graph mentioned by @thunterdb . 
And it seems we need a helper function, converting Graph structure into adjacency list format dataframe. so that user can use this transformer conveniently ?

wangmiao1981:
I will address the review comments soon. Thanks! @WeichenXu123

mllib/src/main/scala/org/apache/spark/ml/clustering/PowerIterationClustering.scala
wangmiao1981: I think checking symmetric is too much for PIC in this data format. Maybe, we can omit the check and put a comment and INFO on console to let users take care of it. @WeichenXu123

mllib/src/main/scala/org/apache/spark/ml/clustering/PowerIterationClustering.scala
wangmiao1981: I remember last time you mentioned not using zip. I might get it wrong. By the way,` it seems we need a helper function`, do you want a helper function takes GraphFrame and returns the RDD used by PIC?

WeichenXu123: Ah, I don't like `Array.zip` because it generate a temp array But `Array.toIterator.zip` do not have this problem, it is iterator zip.

wangmiao1981:
@WeichenXu123 , for the graph helper, the Mllib has a version takes `Graph[Double, Double]` as a parameter for training. In ML, do we have to provide `DataSet` of `Graph`? Can you specify the requirement? I have addressed your other comments. Thanks!

mllib/src/main/scala/org/apache/spark/ml/clustering/PowerIterationClustering.scala
WeichenXu123: OK I agree.

WeichenXu123:
@wangmiao1981 oh, not a big deal, what I thought is that, user is possible to use `graphx` package to get the `Graph[Double, Double]`, and in `ml` package it cannot accept this format, require user to convert it into dataframe (because `mllib` API is deprecate maybe user do not want to use it).
But, this can be discussed in separated JIRA.

wangmiao1981:
@WeichenXu123 Thanks for your review and reply! I agree with you that the helper can be discussed later for potential enhancement.

wangmiao1981:
@weichenXu123 Any other comments? Thanks!

WeichenXu123:
LGTM. ping @yanboliang

wangmiao1981:
ping @yanboliang

wangmiao1981:
ping @yanboliang

mllib/src/main/scala/org/apache/spark/ml/clustering/PowerIterationClustering.scala
```diff
@@ -0,0 +1,216 @@
...
+import org.apache.spark.rdd.RDD
+import org.apache.spark.sql.{DataFrame, Dataset, Row}
+import org.apache.spark.sql.functions.col
+import org.apache.spark.sql.types.{IntegerType, LongType, StructField, StructType}
+
+/**
+ * Common params for PowerIterationClustering
+ */
+private[clustering] trait PowerIterationClusteringParams extends Params with HasMaxIter
+  with HasFeaturesCol with HasPredictionCol with HasWeightCol {
```
jkbradley: We should not use weightCol, which is for instance weights, not for this kind of adjacency.  Let's add a new Param here, perhaps called neighborWeightCol.

mllib/src/main/scala/org/apache/spark/ml/clustering/PowerIterationClustering.scala
jkbradley: +1
Also:
* This should check other input columns to make sure they are defined.
* This should add predictionCol, not check that it exists in the input.

mllib/src/main/scala/org/apache/spark/ml/clustering/PowerIterationClustering.scala
```diff
@@ -0,0 +1,216 @@
...
+   */
+  @Since("2.3.0")
+  val neighborCol = new Param[String](this, "neighbor", "column name for neighbors.")
+
+  /** @group getParam */
+  @Since("2.3.0")
+  def getNeighborCol: String = $(neighborCol)
+
+  /**
+   * Validates the input schema
```
jkbradley: nit: No need for doc like this which is explained by the method title

mllib/src/main/scala/org/apache/spark/ml/clustering/PowerIterationClustering.scala
jkbradley: Also, featuresCol is not used, so it should be removed.

mllib/src/main/scala/org/apache/spark/ml/clustering/PowerIterationClustering.scala
```diff
@@ -0,0 +1,216 @@
...
+    initMode -> "random",
+    idCol -> "id",
+    weightCol -> "weight",
+    neighborCol -> "neighbor")
+
+  @Since("2.3.0")
+  override def copy(extra: ParamMap): PowerIterationClustering = defaultCopy(extra)
+
+  @Since("2.3.0")
+  def this() = this(Identifiable.randomUID("PowerIterationClustering"))
```
jkbradley: nit: Put constructors first, before other methods (like copy), to match the style of the rest of MLlib.

mllib/src/main/scala/org/apache/spark/ml/clustering/PowerIterationClustering.scala
```diff
@@ -0,0 +1,216 @@
...
+      .setK($(k))
+      .setInitializationMode($(initMode))
+      .setMaxIterations($(maxIter))
+    val model = algorithm.run(rdd)
+
+    val rows: RDD[Row] = model.assignments.map {
+      case assignment: Assignment => Row(assignment.id, assignment.cluster)
+    }
+
+    val schema = transformSchema(new StructType(Array(StructField($(idCol), LongType),
```
jkbradley: You should not need to explicitly create a schema here.

mllib/src/test/scala/org/apache/spark/ml/clustering/PowerIterationClusteringSuite.scala
```diff
@@ -0,0 +1,171 @@
...
+import org.apache.spark.ml.linalg.{Vector, Vectors}
+import org.apache.spark.ml.util.DefaultReadWriteTest
+import org.apache.spark.mllib.util.MLlibTestSparkContext
+import org.apache.spark.sql.{DataFrame, Dataset, Row, SparkSession}
+
+class PowerIterationClusteringSuite extends SparkFunSuite
+  with MLlibTestSparkContext with DefaultReadWriteTest {
+
+  @transient var data: Dataset[_] = _
+  @transient var malData: Dataset[_] = _
```
jkbradley: Not used

mllib/src/test/scala/org/apache/spark/ml/clustering/PowerIterationClusteringSuite.scala
```diff
@@ -0,0 +1,171 @@
...
+      .setMaxIter(10)
+      .setInitMode("degree")
+      .transform(data)
+    val predictions2 = Array.fill(2)(mutable.Set.empty[Long])
+    result2.select("id", "prediction").collect().foreach {
+      case Row(id: Long, cluster: Integer) => predictions2(cluster) += id
+    }
+    assert(predictions2.toSet == Set((1 until n1).toSet, (n1 until n).toSet))
+
+    val expectedColumns = Array("id", "prediction")
```
jkbradley: No need to check this since it's already checks above by result2.select(...)

jkbradley (commented):
I just reviewed this.  Overall, it looks good.  In addition to some comments here, I had some comments about the text in docstrings.  For that text, I figured it'd be easiest to send a PR to your PR here: https://github.com/wangmiao1981/spark/pull/4

It'd be nice to add a test for invalid input:
* a negative weight
* mismatched lengths for neighbors & weights in one row
* (not symmetry since we don't check that)

Thanks @wangmiao1981 !

jkbradley:
@wangmiao1981  Do let me know if you're too busy now to resume this; I know it's been a long time.  Thanks!

jkbradley:
I don't mind; I'll take it.  But I'll mark @wangmiao1981 as the main contributor for the PR.  Would you mind closing this issue @wangmiao1981  and I'll reopen a new PR under the same JIRA?

wangmiao1981:
@jkbradley Sorry for missing your comments. Anyway, I will close it now. I will choose another one to work on. Thanks!

wangmiao1981:
@jkbradley I close this one now. Thanks!

jkbradley:
OK sorry to push @wangmiao1981 !  I just want to make sure this gets in before I no longer have bandwidth for it.  If you have the time, would you mind checking the updates I made in the new PR?   Thanks!

PR #15771, opened 2016-11-04
[SPARK-18260] Make from_json null safe

brkyvz:

`from_json` is currently not safe against `null` rows. This PR adds a fix and a regression test for it.

## How was this patch tested?

Regression test

### DIFF
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/jsonExpressions.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/jsonExpressions.scala
@@ -498,8 +498,7 @@ case class JsonToStruct(schema: StructType, options: Map[String, String], child:
   override def children: Seq[Expression] = child :: Nil
 
   override def eval(input: InternalRow): Any = {
-    try parser.parse(child.eval(input).toString).head catch {
-      case _: NullPointerException => null
+    try Option(child.eval(input)).map(json => parser.parse(json.toString).head).orNull catch {
       case _: SparkSQLJsonProcessingException => null
     }
   }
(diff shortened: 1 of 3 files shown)


sql/core/src/test/scala/org/apache/spark/sql/JsonFunctionsSuite.scala
```diff
@@ -114,6 +116,18 @@ class JsonFunctionsSuite extends QueryTest with SharedSQLContext {
       Row(Row(null)) :: Nil)
   }
 
+  test("from_json null column with parquet source") {
+    withTempDir { dir =>
+      Seq("""{"a": 1}""").toDF("value").write.parquet(new File(dir, "p=x").toString)
+      Seq("""{"a": 2}""").toDF("record").write.parquet(new File(dir, "p=y").toString)
+      val schema = new StructType().add("a", IntegerType)
+      val df = spark.read.parquet(dir.toString)
```
marmbrus: Why parquet?  Seems this is testing more than it needs to (and thus runs slower).

brkyvz:
@marmbrus Addressed

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/jsonExpressions.scala
```diff
@@ -498,7 +498,7 @@ case class JsonToStruct(schema: StructType, options: Map[String, String], child:
   override def children: Seq[Expression] = child :: Nil
 
   override def eval(input: InternalRow): Any = {
-    try parser.parse(child.eval(input).toString).head catch {
+    try Option(child.eval(input)).map(json => parser.parse(json.toString).head).orNull catch {
```
marmbrus: Sorry to keep iterating... I think you want to avoid an extra allocation in the execution path.  I would use an `if` statement.

marmbrus:
LGTM, pending tests

rxin:
Merging in master/branch-2.1.

PR #15775, opened 2016-11-04
[SPARK-18280][Core]Fix potential deadlock in `StandaloneSchedulerBackend.dead`

zsxwing:

"StandaloneSchedulerBackend.dead" is called in a RPC thread, so it should not call "SparkContext.stop" in the same thread. "SparkContext.stop" will block until all RPC threads exit, if it's called inside a RPC thread, it will be dead-lock.

This PR add a thread local flag inside RPC threads. `SparkContext.stop` uses it to decide if launching a new thread to stop the SparkContext.

## How was this patch tested?

Jenkins

### DIFF
--- a/core/src/main/scala/org/apache/spark/scheduler/cluster/StandaloneSchedulerBackend.scala
+++ b/core/src/main/scala/org/apache/spark/scheduler/cluster/StandaloneSchedulerBackend.scala
@@ -139,13 +139,7 @@ private[spark] class StandaloneSchedulerBackend(
         scheduler.error(reason)
       } finally {
         // Ensure the application terminates, as we can no longer run jobs.
-        new Thread("stop-spark-context") {
-          setDaemon(true)
-
-          override def run(): Unit = {
-            sc.stop()
-          }
-        }.start()
+        sc.stop()
       }
     }
   }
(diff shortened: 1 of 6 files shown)


rxin:
How can we prevent issues like this from happening again in the future? Any exception we can throw better or linter to do?

zsxwing:
How about throw an exception if the current thread is an RPC thread in `sc.stop`?

rxin:
Yea I think that's a good idea ...

core/src/main/scala/org/apache/spark/scheduler/cluster/StandaloneSchedulerBackend.scala
```diff
@@ -139,7 +139,13 @@ private[spark] class StandaloneSchedulerBackend(
         scheduler.error(reason)
       } finally {
         // Ensure the application terminates, as we can no longer run jobs.
-        sc.stop()
+        new Thread("stop-spark-context") {
```
rxin: also document why we need this.

zsxwing:
I added a thread local flag inside RPC threads. Instead of throwing an exception, `SparkContext.stop` will always launch a new thread to stop `SparkContext`.

I think this is better since we always launch a new thread for such issue.

zsxwing:
retest this please

zsxwing:
retest this please

zsxwing:
Thanks! Merging to master, 2.1 and 2.0.

PR #15777, opened 2016-11-05
[SPARK-18282][ML][PYSPARK] Add python clustering summaries for GMM and BKM

sethah:

Add model summary APIs for `GaussianMixtureModel` and `BisectingKMeansModel` in pyspark.

## How was this patch tested?

Unit tests.

### DIFF
--- a/mllib/src/main/scala/org/apache/spark/ml/clustering/BisectingKMeans.scala
+++ b/mllib/src/main/scala/org/apache/spark/ml/clustering/BisectingKMeans.scala
@@ -133,7 +133,7 @@ class BisectingKMeansModel private[ml] (
   private var trainingSummary: Option[BisectingKMeansSummary] = None
 
   private[clustering] def setSummary(summary: BisectingKMeansSummary): this.type = {
-    this.trainingSummary = Some(summary)
+    this.trainingSummary = Option(summary)
     this
   }
 
--- a/mllib/src/main/scala/org/apache/spark/ml/clustering/BisectingKMeans.scala
+++ b/mllib/src/main/scala/org/apache/spark/ml/clustering/BisectingKMeans.scala
@@ -95,8 +95,7 @@ class BisectingKMeansModel private[ml] (
   @Since("2.0.0")
   override def copy(extra: ParamMap): BisectingKMeansModel = {
     val copied = copyValues(new BisectingKMeansModel(uid, parentModel), extra)
-    if (trainingSummary.isDefined) copied.setSummary(trainingSummary.get)
-    copied.setParent(this.parent)
+    copied.setSummary(trainingSummary).setParent(this.parent)
   }
 
   @Since("2.0.0")
--- a/mllib/src/main/scala/org/apache/spark/ml/clustering/BisectingKMeans.scala
+++ b/mllib/src/main/scala/org/apache/spark/ml/clustering/BisectingKMeans.scala
@@ -132,8 +132,8 @@ class BisectingKMeansModel private[ml] (
 
   private var trainingSummary: Option[BisectingKMeansSummary] = None
 
-  private[clustering] def setSummary(summary: BisectingKMeansSummary): this.type = {
-    this.trainingSummary = Option(summary)
+  private[clustering] def setSummary(summary: Option[BisectingKMeansSummary]): this.type = {
+    this.trainingSummary = summary
     this
   }
 
@@ -265,7 +265,7 @@ class BisectingKMeans @Since("2.0.0") (
     val model = copyValues(new BisectingKMeansModel(uid, parentModel).setParent(this))
     val summary = new BisectingKMeansSummary(
       model.transform(dataset), $(predictionCol), $(featuresCol), $(k))
-    model.setSummary(summary)
+    model.setSummary(Some(summary))
     instr.logSuccess(model)
     model
   }
--- a/python/pyspark/ml/classification.py
+++ b/python/pyspark/ml/classification.py
@@ -309,13 +309,16 @@ class LogisticRegressionModel(JavaModel, JavaClassificationModel, JavaMLWritable
     @since("2.0.0")
     def summary(self):
         """
-        Gets summary (e.g. residuals, mse, r-squared ) of model on
-        training set. An exception is thrown if
-        `trainingSummary is None`.
+        Gets summary (e.g. accuracy/precision/recall, objective history, total iterations) of model
+        trained on the training set. An exception is thrown if `trainingSummary is None`.
         """
-        java_blrt_summary = self._call_java("summary")
-        # Note: Once multiclass is added, update this to return correct summary
-        return BinaryLogisticRegressionTrainingSummary(java_blrt_summary)
+        if self.hasSummary:
+            java_blrt_summary = self._call_java("summary")
+            # Note: Once multiclass is added, update this to return correct summary
+            return BinaryLogisticRegressionTrainingSummary(java_blrt_summary)
+        else:
+            raise RuntimeError("No training summary available for this %s" %
+                               self.__class__.__name__)
 
     @property
     @since("2.0.0")
--- a/python/pyspark/ml/clustering.py
+++ b/python/pyspark/ml/clustering.py
@@ -172,6 +172,13 @@ class GaussianMixture(JavaEstimator, HasFeaturesCol, HasPredictionCol, HasMaxIte
     >>> gm = GaussianMixture(k=3, tol=0.0001,
     ...                      maxIter=10, seed=10)
     >>> model = gm.fit(df)
+    >>> model.hasSummary
+    True
+    >>> summary = model.summary
+    >>> summary.k
+    3
+    >>> summary.clusterSizes
+    [2, 2, 2]
     >>> weights = model.weights
     >>> len(weights)
     3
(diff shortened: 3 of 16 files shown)


python/pyspark/ml/classification.py
```diff
@@ -309,13 +309,16 @@ def interceptVector(self):
...
         """
-        java_blrt_summary = self._call_java("summary")
-        # Note: Once multiclass is added, update this to return correct summary
-        return BinaryLogisticRegressionTrainingSummary(java_blrt_summary)
+        if self.hasSummary:
+            java_blrt_summary = self._call_java("summary")
+            # Note: Once multiclass is added, update this to return correct summary
+            return BinaryLogisticRegressionTrainingSummary(java_blrt_summary)
+        else:
+            raise RuntimeError("No training summary available for this %s" %
```
sethah: Before, this would throw a `Py4JJavaError`. I think it's slightly better to throw a `RuntimeError` here as is done in Scala.

python/pyspark/ml/tests.py
```diff
@@ -1097,6 +1097,42 @@ def test_logistic_regression_summary(self):
...
+        self.assertTrue(isinstance(s.predictions, DataFrame))
+        self.assertEqual(s.probabilityCol, "probability")
+        self.assertTrue(isinstance(s.probability, DataFrame))
+        self.assertEqual(s.featuresCol, "features")
+        self.assertEqual(s.predictionCol, "prediction")
+        self.assertTrue(isinstance(s.cluster, DataFrame))
+        self.assertEqual(len(s.clusterSizes), 2)
+        self.assertEqual(s.k, 2)
+
+        # TODO: test when there is no summary
```
sethah: We should test that `hasSummary` returns `False` when there is no summary available, and that `summary` throws an exception. The problem I'm having is that I'm not sure how to create this test case. The only way to get a model is by calling fit, which will produce a model with a summary. Calling `model._call_java("setSummary", None)` doesn't work either. Is there some way that I'm missing?

python/pyspark/ml/classification.py
holdenk: I think thats generally a good improvement, the `Py4J` errors are often confusing to end users.

python/pyspark/ml/tests.py
holdenk: It might make sense to update `setSummary` to treat null as empty (e.g. `Option` instead of `Some`)) for easy testing.

holdenk:
Thanks for working on this @sethah - more work towards increased parity is good :)

sethah:
ping @yanboliang

mllib/src/main/scala/org/apache/spark/ml/clustering/BisectingKMeans.scala
```diff
@@ -132,7 +132,7 @@ class BisectingKMeansModel private[ml] (
   private var trainingSummary: Option[BisectingKMeansSummary] = None
 
   private[clustering] def setSummary(summary: BisectingKMeansSummary): this.type = {
-    this.trainingSummary = Some(summary)
+    this.trainingSummary = Option(summary)
```
sethah: per @holdenk's suggestion, I changed the `setSummary` to use `Option.apply` which treats `null` as `None`. This allows us to exercise the test case for `hasSummary` when `summary == None`. This was never tested in Scala either so I added unit tests for both Scala and Python.

yanboliang:
@sethah I will take a look in a few days. Thanks.

mllib/src/main/scala/org/apache/spark/ml/clustering/BisectingKMeans.scala
yanboliang: I'd more prefer to make the argument as ```Option[BisectingKMeansSummary]``` like:
```
private[clustering] def setSummary(summary: Option[BisectingKMeansSummary]): this.type = {
    this.trainingSummary = summary
    this
}
```
And test ```summary == None``` by:
```
model.setSummary(None)
assert(!model.hasSummary)
```
Since I think ```setSummary(null)``` and test whether it existing is very tricky. The type of ```summary``` is ```Option[BisectingKMeansSummary]``` and with ```None``` as default value, so ```setSummary(None)``` should make more sense for the scenario that the model does not have summary.
I saw the reason for make this change is that you want to call ```setSummary``` at Python side, and Python None would be converted to ```null``` in Scala. But I think this is private function, we don't need to run test across Scala and Python, since private function should not be called by users.

python/pyspark/ml/clustering.py
```diff
@@ -17,16 +17,75 @@
...
 from pyspark import since, keyword_only
 from pyspark.ml.util import *
-from pyspark.ml.wrapper import JavaEstimator, JavaModel
+from pyspark.ml.wrapper import JavaEstimator, JavaModel, JavaWrapper
 from pyspark.ml.param.shared import *
 from pyspark.ml.common import inherit_doc
 
-__all__ = ['BisectingKMeans', 'BisectingKMeansModel',
+__all__ = ['BisectingKMeans', 'BisectingKMeansModel', 'BisectingKMeansSummary',
+           'ClusteringSummary',
```
yanboliang: +1 @zhengruifeng

python/pyspark/ml/classification.py
yanboliang: I like this change, we should always throw an exception easy to understand by users.

python/pyspark/ml/clustering.py
```diff
@@ -346,6 +453,27 @@ def computeCost(self, dataset):
...
+
+    @property
+    @since("2.1.0")
+    def summary(self):
+        """
+        Gets summary (e.g. cluster assignments, cluster sizes) of the model trained on the
+        training set. An exception is thrown if no summary exists.
+        """
+        if self.hasSummary:
+            return GaussianMixtureSummary(self._call_java("summary"))
```
yanboliang: Typo, should be ```BisectingKMeansSummary```?

python/pyspark/ml/tests.py
```diff
@@ -1097,6 +1097,44 @@ def test_logistic_regression_summary(self):
...
+        self.assertTrue(isinstance(s.predictions, DataFrame))
+        self.assertEqual(s.probabilityCol, "probability")
+        self.assertTrue(isinstance(s.probability, DataFrame))
+        self.assertEqual(s.featuresCol, "features")
+        self.assertEqual(s.predictionCol, "prediction")
+        self.assertTrue(isinstance(s.cluster, DataFrame))
+        self.assertEqual(len(s.clusterSizes), 2)
+        self.assertEqual(s.k, 2)
+
+        model._call_java("setSummary", None)
```
yanboliang: Why we test this? Actually ```setSummary``` is private and it should not generate a model w/o summary in ordinary case. I think we should only test the public API for Python. Further more, if we want to test model w/o summary, we need to write a dummy Scala model w/o summary, and check ```hasSummary``` directly at Python side. I think if the Scala function is not public, we may not confirm Scala/Python compatibility, so testing is also not very make sense.

sethah: This came out of a suggestion in the other PR to add `KMeansSummary`. Adding a `hasSummary` method implies that models can both have and not have summaries. How can we test that `hasSummary` works properly when we can only exercise one of its test cases? Maybe the real problem is that `hasSummary` doesn't make sense since it will always return `True`.

python/pyspark/ml/clustering.py
sethah: Wow, good catch!

python/pyspark/ml/tests.py
yanboliang: Yeah, after loading a saved model, the ```summary``` should be ```None``` and ```hasSummary``` return false. I think this is the correct test route, although with some extra cost. What about add ```hasSummary``` test near ```save/load``` doc test? Then it should not need extra effort.

yanboliang: Yeah, I think it makes sense to add summary related doc tests for algorithms to illustrate the output of summary. So one more line to check ```hasSummary``` does not seam to have much impact. @jkbradley What's your opinion? Thanks.

mllib/src/main/scala/org/apache/spark/ml/clustering/BisectingKMeans.scala
```diff
@@ -95,8 +95,7 @@ class BisectingKMeansModel private[ml] (
   @Since("2.0.0")
   override def copy(extra: ParamMap): BisectingKMeansModel = {
     val copied = copyValues(new BisectingKMeansModel(uid, parentModel), extra)
-    if (trainingSummary.isDefined) copied.setSummary(trainingSummary.get)
-    copied.setParent(this.parent)
+    copied.setSummary(trainingSummary).setParent(this.parent)
```
yanboliang: This looks better. Could you make the change for Scala LiR, LoR, GLM and KMeans as well? I think they should be consistent. Thanks.

yanboliang (commented):
@sethah Only last comment, otherwise, LGTM. I'd like to get this in before 2.1. Thanks.

mllib/src/main/scala/org/apache/spark/ml/clustering/BisectingKMeans.scala
sethah: Updated. I also added tests.

yanboliang:

jkbradley:
@yanboliang I noticed the JIRA is still pending.  Are there follow-up tasks?

yanboliang:
No follow-up, forgetting to close it. Thanks for reminding.

PR #15779, opened 2016-11-05
[SPARK-17748][ML] Minor cleanups to one-pass linear regression with elastic net

jkbradley:

* Made SingularMatrixException private ml
* WeightedLeastSquares: Changed to allow tol >= 0 instead of only tol > 0

## How was this patch tested?

existing tests

### DIFF
--- a/mllib/src/main/scala/org/apache/spark/ml/regression/LinearRegression.scala
+++ b/mllib/src/main/scala/org/apache/spark/ml/regression/LinearRegression.scala
@@ -160,14 +160,13 @@ class LinearRegression @Since("1.3.0") (@Since("1.3.0") override val uid: String
   /**
    * Set the solver algorithm used for optimization.
    * In case of linear regression, this can be "l-bfgs", "normal" and "auto".
-   * "l-bfgs" denotes Limited-memory BFGS which is a limited-memory quasi-Newton
-   * optimization method. "normal" denotes using Normal Equation as an analytical
-   * solution to the linear regression problem.
-   * The default value is "auto" which means that the solver algorithm is
-   * selected automatically.
-   *
-   * The Normal Equation solver is limited to a few thousand features; when needed,
-   * [[LinearRegression]] will automatically fall back to iterative optimization methods.
+   *  - "l-bfgs" denotes Limited-memory BFGS which is a limited-memory quasi-Newton
+   *    optimization method.
+   *  - "normal" denotes using Normal Equation as an analytical solution to the linear regression
+   *    problem.  This solver is limited to [[LinearRegression.MAX_FEATURES_FOR_NORMAL_SOLVER]].
+   *  - "auto" (default) means that the solver algorithm is selected automatically.
+   *    The Normal Equations solver will be used when possible, but this will automatically fall
+   *    back to iterative optimization methods when needed.
    *
    * @group setParam
    */
@@ -407,6 +406,13 @@ object LinearRegression extends DefaultParamsReadable[LinearRegression] {
 
   @Since("1.6.0")
   override def load(path: String): LinearRegression = super.load(path)
+
+  /**
+   * When using [[LinearRegression.solver]] == "normal", the solver must limit the number of
+   * features to at most this number.
+   */
+  @Since("2.1.0")
+  val MAX_FEATURES_FOR_NORMAL_SOLVER: Int = WeightedLeastSquares.MAX_NUM_FEATURES
 }
 
 /**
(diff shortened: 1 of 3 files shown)


jkbradley:
@sethah @yanboliang I just saw your PRs for SPARK-17748.  Awesome change.  I just saw a few nits along the way.

The only major item is making SingularMatrixException private ml.  This would be the first public Exception type we have in MLlib, I believe.  If it is to be public, I'd prefer it live in spark.ml.linalg, so we could move it there instead of making it private.

Also, I'm not a big fan of using Exceptions to handle logic.  I'd prefer to change the use of CholeskySolver to pass the failure code, rather than relying on exceptions.  If that sounds good to you, then I'll make a follow up JIRA for it.

Thanks!

sethah:
+1 on removing the use of exceptions. I thought it was a bit of an awkward solution to begin with. Thanks a lot for this pr, I will take a look soon.

mllib/src/main/scala/org/apache/spark/ml/regression/LinearRegression.scala
```diff
@@ -166,6 +166,9 @@ class LinearRegression @Since("1.3.0") (@Since("1.3.0") override val uid: String
    * The default value is "auto" which means that the solver algorithm is
    * selected automatically.
    *
+   * The Normal Equation solver is limited to a few thousand features; when needed,
```
sethah: Is this a bit misleading? If I set `solver = "normal"` and my dataset contains >4096 features, an exception is thrown instead of automatically falling back to iterative methods.

sethah (commented):
LGTM other than the two comments.

mllib/src/main/scala/org/apache/spark/ml/regression/LinearRegression.scala
yanboliang: +1 @sethah It will only fall back when ```solver = "auto"```, we should clarify it more clear.

yanboliang:
@jkbradley @sethah The idea of using failure code rather than exception sounds good to me. I thought users could get the exception and infer the cause of failure from it. After rethinking it, I think it's an expert exception that few users would like to intercept it, so using failure code should make more sense. Thanks for the follow-up work.

srowen:
I agree that exceptions can become, quite surprisingly, part of a public API. It has to be done with care. I am not sure I agree that failure codes are better; they're exceptions that don't use a language API and are even more ignorable. If it's simply status, that makes sense; if it's indicating failure, you want it to be hard-er to ignore.

jkbradley:
I should have clarified: I meant for us to use the failure code internally.  If the linear algebra ops become more public, then we may need to consider better ways to return result codes.  I'll make a JIRA for it.

mllib/src/main/scala/org/apache/spark/ml/regression/LinearRegression.scala
```diff
@@ -404,6 +406,13 @@ object LinearRegression extends DefaultParamsReadable[LinearRegression] {
 
   @Since("1.6.0")
   override def load(path: String): LinearRegression = super.load(path)
+
+  /**
+   * When using [[LinearRegression.solver]] == "normal", the solver must limit the number of
+   * features to at most this number.
+   */
+  @Since("2.1.0")
+  val MAX_FEATURES_FOR_NORMAL_SOLVER: Int = WeightedLeastSquares.MAX_NUM_FEATURES
```
yanboliang: ```MAX_NUM_FEATURES_FOR_NORMAL_SOLVER``` should be better?

yanboliang (commented):
Only the last nit, otherwise, LGTM. Thanks.

mllib/src/main/scala/org/apache/spark/ml/regression/LinearRegression.scala
```diff
@@ -404,6 +406,13 @@ object LinearRegression extends DefaultParamsReadable[LinearRegression] {
 
   @Since("1.6.0")
   override def load(path: String): LinearRegression = super.load(path)
+
+  /**
+   * When using [[LinearRegression.solver]] == "normal", the solver must limit the number of
+   * features to at most this number.
```
jkbradley: Will do

sethah:
LGTM

jkbradley:
Thanks!  I'll merge this with master and branch-2.1

jkbradley:
Linking commit for branch-2.1 since it hasn't linked automatically: [https://github.com/apache/spark/commit/21bbf94b41fbd193e370a3820131e449aaf0e3db]

PR #15780, opened 2016-11-05
[SPARK-18284][SQL] Make ExpressionEncoder.serializer.nullable precise

kiszk:

This PR makes `ExpressionEncoder.serializer.nullable` for flat encoder for a primitive type `false`. Since it is `true` for now, it is too conservative.
While `ExpressionEncoder.schema` has correct information (e.g. `<IntegerType, false>`), `serializer.head.nullable` of `ExpressionEncoder`, which got from `encoderFor[T]`, is always false. It is too conservative.

This is accomplished by checking whether a type is one of primitive types. If it is `true`, `nullable` should be `false`.

## How was this patch tested?

Added new tests for encoder and dataframe

### DIFF
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/JavaTypeInference.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/JavaTypeInference.scala
@@ -396,12 +396,15 @@ object JavaTypeInference {
 
         case _ if mapType.isAssignableFrom(typeToken) =>
           val (keyType, valueType) = mapKeyValueType(typeToken)
+          val (_, valueNullable) = inferDataType(valueType)
+
           ExternalMapToCatalyst(
             inputObject,
             ObjectType(keyType.getRawType),
             serializerFor(_, keyType),
             ObjectType(valueType.getRawType),
-            serializerFor(_, valueType)
+            serializerFor(_, valueType),
+            valueNullable
           )
 
         case other =>
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/JavaTypeInference.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/JavaTypeInference.scala
@@ -396,7 +396,6 @@ object JavaTypeInference {
 
         case _ if mapType.isAssignableFrom(typeToken) =>
           val (keyType, valueType) = mapKeyValueType(typeToken)
-          val (_, valueNullable) = inferDataType(valueType)
 
           ExternalMapToCatalyst(
             inputObject,
@@ -404,7 +403,7 @@ object JavaTypeInference {
             serializerFor(_, keyType),
             ObjectType(valueType.getRawType),
             serializerFor(_, valueType),
-            valueNullable
+            true
           )
 
         case other =>
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/JavaTypeInference.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/JavaTypeInference.scala
@@ -403,7 +403,7 @@ object JavaTypeInference {
             serializerFor(_, keyType),
             ObjectType(valueType.getRawType),
             serializerFor(_, valueType),
-            true
+            valueNullable = true
           )
 
         case other =>
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/ScalaReflection.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/ScalaReflection.scala
@@ -499,7 +499,7 @@ object ScalaReflection extends ScalaReflection {
           serializerFor(_, keyType, keyPath),
           dataTypeFor(valueType),
           serializerFor(_, valueType, valuePath),
-          !valueType.typeSymbol.asClass.isPrimitive)
+          valueNullable = !valueType.typeSymbol.asClass.isPrimitive)
 
       case t if t <:< localTypeOf[String] =>
         StaticInvoke(
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/ScalaReflection.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/ScalaReflection.scala
@@ -592,7 +592,8 @@ object ScalaReflection extends ScalaReflection {
           }
 
           // we know primitive type takes only non-null, or
-          // we can infer correct nullability for struct's fieldValue by a guard using If(IsNull())
+          // we can infer correct nullability for struct's fieldValue by a guard
+          // using If(IsNull()) at the last line of this section
           val returnNonNull = fieldType.typeSymbol.asClass.isPrimitive ||
             definedByConstructorParams(fieldType)
           val fieldValue = Invoke(inputObject, fieldName, dataTypeFor(fieldType),
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/ScalaReflection.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/ScalaReflection.scala
@@ -591,11 +591,9 @@ object ScalaReflection extends ScalaReflection {
               "cannot be used as field name\n" + walkedTypePath.mkString("\n"))
           }
 
-          // primitive take only non-null or struct takes non-null object guarded by isNull
-          val returnNonNull = fieldType.typeSymbol.asClass.isPrimitive ||
-            definedByConstructorParams(fieldType)
-          val fieldValue = Invoke(inputObject, fieldName, dataTypeFor(fieldType),
-            returnNullable = !returnNonNull)
+          val fieldValue = Invoke(
+            AssertNotNull(inputObject, walkedTypePath), fieldName, dataTypeFor(fieldType),
+            returnNullable = !fieldType.typeSymbol.asClass.isPrimitive)
           val clsName = getClassNameFromType(fieldType)
           val newPath = s"""- field (class: "$clsName", name: "$fieldName")""" +: walkedTypePath
           expressions.Literal(fieldName) :: serializerFor(fieldValue, fieldType, newPath) :: Nil
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/ScalaReflection.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/ScalaReflection.scala
@@ -591,7 +591,7 @@ object ScalaReflection extends ScalaReflection {
           }
 
           val fieldValue = Invoke(inputObject, fieldName, dataTypeFor(fieldType),
-            returnNullable = inputObject.nullable || !fieldType.typeSymbol.asClass.isPrimitive)
+            returnNullable = !fieldType.typeSymbol.asClass.isPrimitive)
           val clsName = getClassNameFromType(fieldType)
           val newPath = s"""- field (class: "$clsName", name: "$fieldName")""" +: walkedTypePath
           expressions.Literal(fieldName) :: serializerFor(fieldValue, fieldType, newPath) :: Nil
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/ScalaReflection.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/ScalaReflection.scala
@@ -590,7 +590,12 @@ object ScalaReflection extends ScalaReflection {
               "cannot be used as field name\n" + walkedTypePath.mkString("\n"))
           }
 
-          val fieldValue = Invoke(inputObject, fieldName, dataTypeFor(fieldType))
+          val fieldValue = if (!fieldType.typeSymbol.asClass.isPrimitive) {
+            Invoke(inputObject, fieldName, dataTypeFor(fieldType))
+          } else {
+            AssertNotNull(Invoke(inputObject, fieldName, dataTypeFor(fieldType)),
+              Seq("primitive type should not have null"))
+          }
           val clsName = getClassNameFromType(fieldType)
           val newPath = s"""- field (class: "$clsName", name: "$fieldName")""" +: walkedTypePath
           expressions.Literal(fieldName) :: serializerFor(fieldValue, fieldType, newPath) :: Nil
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/ScalaReflection.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/ScalaReflection.scala
@@ -590,12 +590,8 @@ object ScalaReflection extends ScalaReflection {
               "cannot be used as field name\n" + walkedTypePath.mkString("\n"))
           }
 
-          val fieldValue = if (!fieldType.typeSymbol.asClass.isPrimitive) {
-            Invoke(inputObject, fieldName, dataTypeFor(fieldType))
-          } else {
-            AssertNotNull(Invoke(inputObject, fieldName, dataTypeFor(fieldType)),
-              Seq("primitive type should not have null"))
-          }
+          val fieldValue = Invoke(inputObject, fieldName, dataTypeFor(fieldType),
+            returnNullable = inputObject.nullable || !fieldType.typeSymbol.asClass.isPrimitive)
           val clsName = getClassNameFromType(fieldType)
           val newPath = s"""- field (class: "$clsName", name: "$fieldName")""" +: walkedTypePath
           expressions.Literal(fieldName) :: serializerFor(fieldValue, fieldType, newPath) :: Nil
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/ScalaReflection.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/ScalaReflection.scala
@@ -591,8 +591,11 @@ object ScalaReflection extends ScalaReflection {
               "cannot be used as field name\n" + walkedTypePath.mkString("\n"))
           }
 
+          // primitive take only non-null or struct takes non-null object guarded by isNull
+          val returnNonNull = fieldType.typeSymbol.asClass.isPrimitive ||
+            definedByConstructorParams(fieldType)
           val fieldValue = Invoke(inputObject, fieldName, dataTypeFor(fieldType),
-            returnNullable = !fieldType.typeSymbol.asClass.isPrimitive)
+            returnNullable = !returnNonNull)
           val clsName = getClassNameFromType(fieldType)
           val newPath = s"""- field (class: "$clsName", name: "$fieldName")""" +: walkedTypePath
           expressions.Literal(fieldName) :: serializerFor(fieldValue, fieldType, newPath) :: Nil
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/ScalaReflection.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/ScalaReflection.scala
@@ -591,13 +591,9 @@ object ScalaReflection extends ScalaReflection {
               "cannot be used as field name\n" + walkedTypePath.mkString("\n"))
           }
 
-          // we know primitive type takes only non-null, or
-          // we can infer correct nullability for struct's fieldValue by a guard
-          // using If(IsNull()) at the last line of this section
-          val returnNonNull = fieldType.typeSymbol.asClass.isPrimitive ||
-            definedByConstructorParams(fieldType)
-          val fieldValue = Invoke(inputObject, fieldName, dataTypeFor(fieldType),
-            returnNullable = !returnNonNull)
+          val fieldValue = Invoke(
+            AssertNotNull(inputObject, walkedTypePath), fieldName, dataTypeFor(fieldType),
+            returnNullable = !fieldType.typeSymbol.asClass.isPrimitive)
           val clsName = getClassNameFromType(fieldType)
           val newPath = s"""- field (class: "$clsName", name: "$fieldName")""" +: walkedTypePath
           expressions.Literal(fieldName) :: serializerFor(fieldValue, fieldType, newPath) :: Nil
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/ScalaReflection.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/ScalaReflection.scala
@@ -591,9 +591,12 @@ object ScalaReflection extends ScalaReflection {
               "cannot be used as field name\n" + walkedTypePath.mkString("\n"))
           }
 
-          val fieldValue = Invoke(
-            AssertNotNull(inputObject, walkedTypePath), fieldName, dataTypeFor(fieldType),
-            returnNullable = !fieldType.typeSymbol.asClass.isPrimitive)
+          // we know primitive type takes only non-null, or
+          // we can infer correct nullability for struct's fieldValue by a guard using If(IsNull())
+          val returnNonNull = fieldType.typeSymbol.asClass.isPrimitive ||
+            definedByConstructorParams(fieldType)
+          val fieldValue = Invoke(inputObject, fieldName, dataTypeFor(fieldType),
+            returnNullable = !returnNonNull)
           val clsName = getClassNameFromType(fieldType)
           val newPath = s"""- field (class: "$clsName", name: "$fieldName")""" +: walkedTypePath
           expressions.Literal(fieldName) :: serializerFor(fieldValue, fieldType, newPath) :: Nil
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/ScalaReflection.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/ScalaReflection.scala
@@ -498,7 +498,8 @@ object ScalaReflection extends ScalaReflection {
           dataTypeFor(keyType),
           serializerFor(_, keyType, keyPath),
           dataTypeFor(valueType),
-          serializerFor(_, valueType, valuePath))
+          serializerFor(_, valueType, valuePath),
+          !valueType.typeSymbol.asClass.isPrimitive)
 
       case t if t <:< localTypeOf[String] =>
         StaticInvoke(
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/encoders/ExpressionEncoder.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/encoders/ExpressionEncoder.scala
@@ -61,10 +61,7 @@ object ExpressionEncoder {
     val serializer = ScalaReflection.serializerFor[T](nullSafeInput)
     val deserializer = ScalaReflection.deserializerFor[T]
 
-    val schema = ScalaReflection.schemaFor[T] match {
-      case ScalaReflection.Schema(s: StructType, _) => s
-      case ScalaReflection.Schema(dt, nullable) => new StructType().add("value", dt, nullable)
-    }
+    val schema = serializer.dataType
 
     new ExpressionEncoder[T](
       schema,
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/encoders/ExpressionEncoder.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/encoders/ExpressionEncoder.scala
@@ -50,7 +50,7 @@ object ExpressionEncoder {
     val cls = mirror.runtimeClass(tpe)
     val flat = !ScalaReflection.definedByConstructorParams(tpe)
 
-    val inputObject = BoundReference(0, ScalaReflection.dataTypeFor[T], nullable = true)
+    val inputObject = BoundReference(0, ScalaReflection.dataTypeFor[T], nullable = !cls.isPrimitive)
     val nullSafeInput = if (flat) {
       inputObject
     } else {
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/ReferenceToExpressions.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/ReferenceToExpressions.scala
@@ -74,8 +74,7 @@ case class ReferenceToExpressions(result: Expression, children: Seq[Expression])
         ctx.addMutableState("boolean", classChildVarIsNull, "")
 
         val classChildVar =
-          LambdaVariable(classChildVarName, classChildVarIsNull, child.dataType,
-            childGen.isNull != "false")
+          LambdaVariable(classChildVarName, classChildVarIsNull, child.dataType, child.nullable)
 
         val initCode = s"${classChildVar.value} = ${childGen.value};\n" +
           s"${classChildVar.isNull} = ${childGen.isNull};"
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/ReferenceToExpressions.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/ReferenceToExpressions.scala
@@ -74,7 +74,8 @@ case class ReferenceToExpressions(result: Expression, children: Seq[Expression])
         ctx.addMutableState("boolean", classChildVarIsNull, "")
 
         val classChildVar =
-          LambdaVariable(classChildVarName, classChildVarIsNull, child.dataType)
+          LambdaVariable(classChildVarName, classChildVarIsNull, child.dataType,
+            childGen.isNull != "false")
 
         val initCode = s"${classChildVar.value} = ${childGen.value};\n" +
           s"${classChildVar.isNull} = ${childGen.isNull};"
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
@@ -406,10 +406,11 @@ case class WrapOption(child: Expression, optType: DataType)
  * A place holder for the loop variable used in [[MapObjects]].  This should never be constructed
  * manually, but will instead be passed into the provided lambda function.
  */
-case class LambdaVariable(value: String, isNull: String, dataType: DataType) extends LeafExpression
+case class LambdaVariable(value: String, isNull: String, dataType: DataType,
+    valueNullable: Boolean = true) extends LeafExpression
   with Unevaluable with NonSQLExpression {
 
-  override def nullable: Boolean = true
+  override def nullable: Boolean = valueNullable
 
   override def genCode(ctx: CodegenContext): ExprCode = {
     ExprCode(code = "", value = value, isNull = isNull)
@@ -593,7 +594,8 @@ object ExternalMapToCatalyst {
       keyType: DataType,
       keyConverter: Expression => Expression,
       valueType: DataType,
-      valueConverter: Expression => Expression): ExternalMapToCatalyst = {
+      valueConverter: Expression => Expression,
+      valueNullable: Boolean): ExternalMapToCatalyst = {
     val id = curId.getAndIncrement()
     val keyName = "ExternalMapToCatalyst_key" + id
     val valueName = "ExternalMapToCatalyst_value" + id
@@ -606,7 +608,7 @@ object ExternalMapToCatalyst {
       valueName,
       valueIsNull,
       valueType,
-      valueConverter(LambdaVariable(valueName, valueIsNull, valueType)),
+      valueConverter(LambdaVariable(valueName, valueIsNull, valueType, valueNullable)),
       inputMap
     )
   }
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
@@ -413,7 +413,7 @@ case class LambdaVariable(value: String, isNull: String, dataType: DataType,
   override def nullable: Boolean = valueNullable
 
   override def genCode(ctx: CodegenContext): ExprCode = {
-    ExprCode(code = "", value = value, isNull = isNull)
+    ExprCode(code = "", value = value, isNull = if (nullable) isNull else "false")
   }
 }
 
@@ -604,7 +604,7 @@ object ExternalMapToCatalyst {
     ExternalMapToCatalyst(
       keyName,
       keyType,
-      keyConverter(LambdaVariable(keyName, "false", keyType)),
+      keyConverter(LambdaVariable(keyName, "false", keyType, false)),
       valueName,
       valueIsNull,
       valueType,
@@ -647,12 +647,7 @@ case class ExternalMapToCatalyst private(
   override def foldable: Boolean = false
 
   override def dataType: MapType = {
-    val isPrimitiveType = valueType match {
-      case BooleanType | ByteType | ShortType | IntegerType | LongType |
-            FloatType | DoubleType => true
-      case _ => false
-    }
-    MapType(keyConverter.dataType, valueConverter.dataType, !isPrimitiveType)
+    MapType(keyConverter.dataType, valueConverter.dataType, valueConverter.nullable)
   }
 
   override def eval(input: InternalRow): Any =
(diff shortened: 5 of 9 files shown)


hvanhovell:
cc @cloud-fan

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/encoders/ExpressionEncoder.scala
```diff
@@ -50,7 +50,7 @@ object ExpressionEncoder {
     val cls = mirror.runtimeClass(tpe)
     val flat = !ScalaReflection.definedByConstructorParams(tpe)
 
-    val inputObject = BoundReference(0, ScalaReflection.dataTypeFor[T], nullable = true)
+    val inputObject = BoundReference(0, ScalaReflection.dataTypeFor[T], nullable = !cls.isPrimitive)
```
cloud-fan: does it work for nested primitive types? e.g. `Seq(1, 2, 3)`, should have data type `ArrayType(IntegerType, containsNull = false)`

kiszk: Yes, it works. I have just added [a new test](https://github.com/kiszk/spark/blob/38991d00cbaa50ffc9d22c54f643ed03e51b4785/sql/core/src/test/scala/org/apache/spark/sql/DatasetSuite.scala#L944-L946) for this.

kiszk:
cc @ueshin

sql/core/src/test/scala/org/apache/spark/sql/DatasetSuite.scala
```diff
@@ -935,6 +935,20 @@ class DatasetSuite extends QueryTest with SharedSQLContext {
     assert(dataset.collect() sameElements Array(resultValue, resultValue))
   }
 
+  test("SPARK-18284: Serializer should have correct nullable value") {
+    val df1 = sparkContext.parallelize(Seq(1, 2, 3, 4), 1).toDF
```
cloud-fan: why not use `Seq(1, 2, 3, 4).toDF`? Does the number of partitions matters here?

cloud-fan:
LGTM

sql/core/src/test/scala/org/apache/spark/sql/DatasetSuite.scala
kiszk: No matters, and it looks simple. done

cloud-fan:
Some more thoughts: in `ExpressionEncoder` we get the schema of encoder by
```
val schema = ScalaReflection.schemaFor[T] match {
  case ScalaReflection.Schema(s: StructType, _) => s
  case ScalaReflection.Schema(dt, nullable) => new StructType().add("value", dt, nullable)
}
```
now we fix the nullability of the serializer expressions, can we just use `val schema = serializer.schema` and see if tests can pass?

kiszk:
Sure, let me try this

kiszk:
@cloud-fan , this optimization leads to failure of the program `Seq((0, 1.0, 3.0), (2, 2.0, 5.0)).toDF("id", "v").printSchema` .
While the previous code returns `false`, this optimization returns `true`.

kiszk:
ping @cloud-fan

cloud-fan:
> this optimization leads to failure of the program Seq((0, 1.0, 3.0), (2, 2.0, 5.0)).toDF("id", "v").printSchema

This is not an optimization, we should use `serializer.schema` as encoder schema. Can you look into it?

kiszk:
I see. I will look at this carefully.
While I looked at this, I did not understand why only a tuple causes a failure. Do you have any insight on how a tuple is handled?

ueshin:
Needs `AssertNotNull()` [here](https://github.com/kiszk/spark/blob/38991d00cbaa50ffc9d22c54f643ed03e51b4785/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/ScalaReflection.scala#L577) if the field type is primitive?
I don't know this is enough for `serializer.schema` as encoder schema, though.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
```diff
@@ -603,7 +603,14 @@ case class ExternalMapToCatalyst private(
...
   override def foldable: Boolean = false
 
-  override def dataType: MapType = MapType(keyConverter.dataType, valueConverter.dataType)
+  override def dataType: MapType = {
+    val isPrimitiveType = valueType match {
+      case BooleanType | ByteType | ShortType | IntegerType | LongType |
+            FloatType | DoubleType => true
+      case _ => false
+    }
+    MapType(keyConverter.dataType, valueConverter.dataType, !isPrimitiveType)
```
ueshin: I think this approach would break nullable constraint.
For example, even if `valueType` is primitive, `valueConverter` can return `null`.

kiszk: Are you talking about another case rather than [this test](https://github.com/apache/spark/pull/15780/files#diff-3836bd1fc5ae9c8b8ac4bdb2b9944159R972) ?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/ScalaReflection.scala
```diff
@@ -574,7 +574,12 @@ object ScalaReflection extends ScalaReflection {
               "cannot be used as field name\n" + walkedTypePath.mkString("\n"))
           }
 
-          val fieldValue = Invoke(inputObject, fieldName, dataTypeFor(fieldType))
+          val fieldValue = if (!fieldType.typeSymbol.asClass.isPrimitive) {
+            Invoke(inputObject, fieldName, dataTypeFor(fieldType))
+          } else {
+            AssertNotNull(Invoke(inputObject, fieldName, dataTypeFor(fieldType)),
```
cloud-fan: maybe we can just add a `returnPrimitiveType` parameter in `Invoke` and update its `nullable`. The `AssertNotNull` here will never get evaluated and is only used to correct the `nullable`.

ueshin: @cloud-fan How about `returnNullable` for general cases?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
kiszk: Sorry. I cannot imagine a program that produces this condition. Would it be possible to give me an example program?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/ScalaReflection.scala
viirya: +1 for `returnNullable`.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
cloud-fan: yea, `BooleanType` can be primitive boolean but can also be `java.lang.Boolean`, we need an extra parameter to decide if the map can contain null values.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/ScalaReflection.scala
cloud-fan: LGTM

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
ueshin: I think we should use `valueConverter.nullable` to determine `valueContainsNull`.

kiszk: `(Tuple1(Map(2 -> 3)) :: Nil).toDF("m")` returns `true` in `valueConverter.nullable`. Am I wrong?

cloud-fan: it should be false, do we return true now?

kiszk: Yes, it returns `true` now. I think that this is because `LambdaVariable.nullable` always return `true`.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/ScalaReflection.scala
```diff
@@ -574,7 +574,8 @@ object ScalaReflection extends ScalaReflection {
               "cannot be used as field name\n" + walkedTypePath.mkString("\n"))
           }
 
-          val fieldValue = Invoke(inputObject, fieldName, dataTypeFor(fieldType))
+          val fieldValue = Invoke(inputObject, fieldName, dataTypeFor(fieldType),
+            returnNullable = inputObject.nullable || !fieldType.typeSymbol.asClass.isPrimitive)
```
cloud-fan: we use just write `returnNullable = !fieldType.typeSymbol.asClass.isPrimitive` and check `inputObject.nullable` in `Invoke.nullable`

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
kiszk: Since `LambdaVariable` receives only `DataType`, it is not easy to set `nullable` precisely.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/ScalaReflection.scala
kiszk: This change seem to lead to a failure at the following assertion
```java
val schema = ExpressionEncoder[(Int, (String, Int))].schema
assert(schema(1).dataType.asInstanceOf[StructType](1).nullable === false)
```

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
kiszk: @cloud-fan @ueshin @viirya Is there any idea to set correct value into `LambdaVariable.nullable`?
Since I am busy on Sunday and Monday, I will be back to this on Tuesday.

cloud-fan: how about we define `LambdaVariable.nullable` as `!isNull == "false"`?

ueshin: @cloud-fan I'm afraid I don't think it's a good idea..

There are 3 cases that we use `LambdaVariable`:

- As a placeholder of element for `MapObjects`.
- As a placeholder of key for `ExternalMapToCatalyst`
- As a placeholder of value for `ExternalMapToCatalyst`

Only 1 case we can fix nullability by it is the second case and it would not be the cause of this problem.

cloud-fan: looking at the places we create `ExternalMapToCatalyst`, I think we do know what the map value type is and if it can be null. How about we add a `valueNullable: Boolean` parameter in `ExternalMapToCatalyst` like what we did for `Invoke`?

ueshin: I agree with it, but how about this patch (https://github.com/ueshin/apache-spark/commit/1462a767c9643c04b08cb29abaf94f8a5090a90a) for now?
All tests in catalyst/sql passed in my local environment after patch.

I'd like to separate the following issues to other PRs:

- Add `returnNullable: Boolean` to `StaticInvoke` and `Invoke` and fix nullable of them.
  - and then remove `AssertNotNull()` at [here](https://github.com/ueshin/apache-spark/commit/1462a767c9643c04b08cb29abaf94f8a5090a90a#diff-89643554d9757dd3e91abff1cc6096c7R581) like the current implementation.
- Add `valueNullable: Boolean` to `ExternalMapToCatalyst`.
  - and then remove `AssertNotNull()` at [here](https://github.com/ueshin/apache-spark/commit/1462a767c9643c04b08cb29abaf94f8a5090a90a#diff-89643554d9757dd3e91abff1cc6096c7R485) and add `valueNullable = valueType.typeSymbol.asClass.isPrimitive`.

What do you think?

cloud-fan: There are only 2 places that create `ExternalMapToCatalyst`, seems the change won't be too large? I think incorrect nullability is not a critical/blocker issue and we don't need to rush.

ueshin: Hm, I see. Let's go ahead.

kiszk: I agree with adding a flag.
I will try to add a `valueNullable: Boolean` parameter in `ExternalMapToCatalyst` in this PR since it would be not large change.

viirya: Actually we already do the similar thing at JavaTypeInference https://github.com/apache/spark/blob/f00df40cfefef0f3fc73f16ada1006e4dcfa5a39/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/JavaTypeInference.scala#L107.

Looks like we can infer value nullability at the places we create `ExternalMapToCatalyst` more precisely. It would not be a large change with the current type inference utils.

kiszk: I fixed failures for a nested struct based on the [suggested approach](https://github.com/apache/spark/pull/15780#discussion_r88587549) adding an argument to `Invoke`.
@cloud-fan @ueshin @viirya Is it OK?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/JavaTypeInference.scala
```diff
@@ -396,12 +396,15 @@ object JavaTypeInference {
 
         case _ if mapType.isAssignableFrom(typeToken) =>
           val (keyType, valueType) = mapKeyValueType(typeToken)
+          val (_, valueNullable) = inferDataType(valueType)
```
cloud-fan: I think the value of java map is always nullable?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/ScalaReflection.scala
```diff
@@ -590,7 +591,11 @@ object ScalaReflection extends ScalaReflection {
               "cannot be used as field name\n" + walkedTypePath.mkString("\n"))
           }
 
-          val fieldValue = Invoke(inputObject, fieldName, dataTypeFor(fieldType))
+          // primitive take only non-null or struct takes non-null object guarded by isNull
+          val returnNonNull = fieldType.typeSymbol.asClass.isPrimitive ||
+            definedByConstructorParams(fieldType)
```
cloud-fan: explain this?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
```diff
@@ -605,7 +608,7 @@ object ExternalMapToCatalyst {
       valueName,
       valueIsNull,
       valueType,
-      valueConverter(LambdaVariable(valueName, valueIsNull, valueType)),
+      valueConverter(LambdaVariable(valueName, valueIsNull, valueType, valueNullable)),
```
cloud-fan: shall we update the `LambdaVariable` for key too? It's always non-nullable

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
```diff
@@ -643,8 +646,14 @@ case class ExternalMapToCatalyst private(
 
   override def foldable: Boolean = false
 
-  override def dataType: MapType = MapType(
-    keyConverter.dataType, valueConverter.dataType, valueContainsNull = valueConverter.nullable)
+  override def dataType: MapType = {
+    val isPrimitiveType = valueType match {
```
cloud-fan: why we still have this? can't we use `valueConverter.nullable`?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/ScalaReflection.scala
cloud-fan: can you investigate why it fails?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/ScalaReflection.scala
```diff
@@ -590,7 +591,11 @@ object ScalaReflection extends ScalaReflection {
               "cannot be used as field name\n" + walkedTypePath.mkString("\n"))
           }
 
-          val fieldValue = Invoke(inputObject, fieldName, dataTypeFor(fieldType))
+          // primitive take only non-null or struct takes non-null object guarded by isNull
```
viirya: +1

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
```diff
@@ -405,10 +406,11 @@ case class WrapOption(child: Expression, optType: DataType)
...
-case class LambdaVariable(value: String, isNull: String, dataType: DataType) extends LeafExpression
+case class LambdaVariable(value: String, isNull: String, dataType: DataType,
+    valueNullable: Boolean = true) extends LeafExpression
   with Unevaluable with NonSQLExpression {
 
-  override def nullable: Boolean = true
+  override def nullable: Boolean = valueNullable
 
   override def genCode(ctx: CodegenContext): ExprCode = {
     ExprCode(code = "", value = value, isNull = isNull)
```
viirya: If `valueNullable` is false, can we replace `isNull` as constant "false" directly?

kiszk: good catch

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
kiszk: Sure

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
kiszk: Yeah, done

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
```diff
@@ -405,13 +406,14 @@ case class WrapOption(child: Expression, optType: DataType)
  * A place holder for the loop variable used in [[MapObjects]].  This should never be constructed
  * manually, but will instead be passed into the provided lambda function.
  */
-case class LambdaVariable(value: String, isNull: String, dataType: DataType) extends LeafExpression
+case class LambdaVariable(value: String, isNull: String, dataType: DataType,
+    valueNullable: Boolean = true) extends LeafExpression
```
ueshin: We can use `nullable: Boolean` here.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
```diff
@@ -177,9 +177,10 @@ case class Invoke(
     functionName: String,
     dataType: DataType,
     arguments: Seq[Expression] = Nil,
-    propagateNull: Boolean = true) extends InvokeLike {
+    propagateNull: Boolean = true,
+    returnNullable : Boolean = true) extends InvokeLike {
```
ueshin: Add a `@param` document for `returnNullable`.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/ScalaReflection.scala
viirya: The result is correct. But it looks a bit weird because actually the returned Tuple can be null.

You set it as non-null because we can infer correct nullability for its fields by that.

This part is little tricky to me. We may need more clear comment to explain this if we want to do this. Can you rephrase the current comment?

ueshin (changes requested):

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/ScalaReflection.scala
ueshin: I think we can only guarantee that the `inputObject` is not null in the false case, so we should use `AssertNotNull()` for `inputObject` and the `fieldValue` will be like:

```scala
val fieldValue = Invoke(
  AssertNotNull(inputObject, walkedTypePath), fieldName, dataTypeFor(fieldType),
  returnNullable = !fieldType.typeSymbol.asClass.isPrimitive)
```

kiszk: Yes, I agree that this part is little tricky. After waiting for other comments, I will rephrase the comment on Tuesday.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
```diff
@@ -177,9 +177,10 @@ case class Invoke(
...
 
-  override def nullable: Boolean = true
+  override def nullable: Boolean = targetObject.nullable || returnNullable
```
viirya: When `returnNullable` is false, meaning the invoked method doesn't return a null, and `targetObject` is also not null, why `Invoke.nullable` is true?

viirya: OK. I see. What you want is to have `Invoke` as null if any input argument is null, no matters what the invoked method returns.

ueshin: Yes, that's right.
When `needNullCheck == true`, i.e. `propagateNull && arguments.exists(_.nullable)`, if there is null argument, `Invoke` returns null.

viirya: That makes sense.

kiszk: i see. done

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
kiszk: I see. Now caller explicitly passes a value to `valueNullable`.

ueshin: I meant that we could use the parameter name `nullable` like:

```scala
case class LambdaVariable(value: String, isNull: String, dataType: DataType,
    nullable: Boolean = true) extends LeafExpression
```

and remove `override def nullable: Boolean = valueNullable`.

kiszk: I see. thanks.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
```diff
@@ -643,8 +645,9 @@ case class ExternalMapToCatalyst private(
 
   override def foldable: Boolean = false
 
-  override def dataType: MapType = MapType(
-    keyConverter.dataType, valueConverter.dataType, valueContainsNull = valueConverter.nullable)
+  override def dataType: MapType = {
+    MapType(keyConverter.dataType, valueConverter.dataType, valueConverter.nullable)
+  }
```
ueshin: Looks no difference here?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/ReferenceToExpressions.scala
```diff
@@ -74,7 +74,8 @@ case class ReferenceToExpressions(result: Expression, children: Seq[Expression])
         ctx.addMutableState("boolean", classChildVarIsNull, "")
 
         val classChildVar =
-          LambdaVariable(classChildVarName, classChildVarIsNull, child.dataType)
+          LambdaVariable(classChildVarName, classChildVarIsNull, child.dataType,
+            childGen.isNull != "false")
```
ueshin: Use `child.nullable` if you want to specify it here.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/JavaTypeInference.scala
kiszk: good catch. done

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
kiszk: good catch. done

sql/core/src/test/scala/org/apache/spark/sql/streaming/FileStreamSinkSuite.scala
```diff
@@ -86,7 +86,7 @@ class FileStreamSinkSuite extends StreamTest {
 
       val outputDf = spark.read.parquet(outputDir)
       val expectedSchema = new StructType()
-        .add(StructField("value", IntegerType))
+        .add(StructField("value", IntegerType, nullable = false))
         .add(StructField("id", IntegerType))
```
ueshin: BTW, do you know why `id` is not `nullable == false`?
Looks both `value` and `id` are `nullable == false`.

kiszk: Unfortunately, [this code](https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/PartitioningUtils.scala#L139) makes 'id' nullable. The column specified by `partitionBy()` will have `nullable=true`.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/ScalaReflection.scala
kiszk: Since this code looks simple, I used this without any comment. Thank you very much.

sql/core/src/test/scala/org/apache/spark/sql/streaming/FileStreamSinkSuite.scala
ueshin: I see, thanks.

kiszk:
ping @cloud-fan

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/ScalaReflection.scala
ueshin: Hmm, I found something wrong with the combination of `Option` and something the serializer of which is made by `t if definedByConstructorParams(t)` case like `Option[Tuple2[Int, String]]` (even at master).

If we add a following test to `ExpressionEncoderSuite` (at master):

```scala
encodeDecodeTest(Option.empty[Tuple2[Int, Sring]], "empty option of tuple2 of int and string")
```

it fails as:

```
Encoded/Decoded data does not match input data

in:  None
out: Some((0,null))
types: scala.Tuple2

Encoded Data: [null,null]
Schema: _1#381,_2#382
root
-- _1: integer (nullable = false)
-- _2: string (nullable = true)
```

(BTW, there are no `Option` related tests in `ExpressionEncoderSuite`..)

The failure of `roundtrip in to_json and from_json` would be related to this.

viirya: ExpressionEncoder can't encode `None` at the root level, as Spark SQL doesn't allow top-level row to be null.

viirya: My previous comment is not correct.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/ScalaReflection.scala
```diff
@@ -590,7 +591,12 @@ object ScalaReflection extends ScalaReflection {
               "cannot be used as field name\n" + walkedTypePath.mkString("\n"))
           }
 
-          val fieldValue = Invoke(inputObject, fieldName, dataTypeFor(fieldType))
+          // we know primitive type takes only non-null, or
+          // we can infer correct nullability for struct's fieldValue by a guard using If(IsNull())
```
cloud-fan: where do we add the `If(IsNull)`?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
```diff
@@ -405,13 +407,12 @@ case class WrapOption(child: Expression, optType: DataType)
  * A place holder for the loop variable used in [[MapObjects]].  This should never be constructed
  * manually, but will instead be passed into the provided lambda function.
  */
-case class LambdaVariable(value: String, isNull: String, dataType: DataType) extends LeafExpression
+case class LambdaVariable(value: String, isNull: String, dataType: DataType,
+    nullable: Boolean = true) extends LeafExpression
```
cloud-fan: nit: code style should be
```
case class xxx(
  param1: xxx
  param2: xxx) extends ...
```

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/encoders/ExpressionEncoderSuite.scala
```diff
@@ -338,6 +338,18 @@ class ExpressionEncoderSuite extends PlanTest with AnalysisTest {
     }
   }
 
+  test("nullable of encoder serializer") {
+    def checkNullable[T: Encoder](nullable: Boolean*): Unit = {
```
cloud-fan: this is over design, we only pass a single parameter, or are you going to add some more non-flat cases?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/ScalaReflection.scala
viirya: The concept of `None` can't well fit with Tuple, except for Tuple1.

For a Tuple 2, for example, we encoder it as a row of 2 columns. If it is None, should we encode it as `[null]` or `[null, null]`? Conceptually, `[null]` looks the correct answer. However, practically it becomes a row with only one column and there is a conflict in data format.

Currently, if given a None for a Tuple2 data, we will encode it as `[null, null]`, as you seen in the test @ueshin mentioned.

viirya: So I think we can't very well modeling optional data type like Option[Tuple2[Int, String]] with a row.

cloud-fan: using `Dataset[Option[xxx]]` is going to be banned, see https://github.com/apache/spark/pull/15979

ueshin: +1

`Option[Int]`, `Option[String]`, `Option[Tuple1[Int]]`, `Option[Tuple1[String]]` are ok?
If so I'd like you to add the following tests to `ExpressionEncoderSuite`:

```scala
encodeDecodeTest(Option(31), "option of int")
encodeDecodeTest(Option("abc"), "option of string")
encodeDecodeTest(Option.empty[Int], "empty option of int")
encodeDecodeTest(Option.empty[String], "empty option of string")
encodeDecodeTest(Option(Tuple1(31)), "option of tuple1 of int")
encodeDecodeTest(Option.empty[Tuple1[Int]], "empty option of tuple1 of int")
encodeDecodeTest(Option(Tuple1("abc")), "option of tuple1 of string")
encodeDecodeTest(Option.empty[Tuple1[String]], "empty option of tuple1 of string")
```

ueshin: I see that `Option[Tuple1[Int]]`, `Option[Tuple1[String]]` are invalid.

ueshin: Anyway, after #15979, I believe we can use the previous code I suggested:

```scala
val fieldValue = Invoke(
  AssertNotNull(inputObject, walkedTypePath), fieldName, dataTypeFor(fieldType),
  returnNullable = !fieldType.typeSymbol.asClass.isPrimitive)
```

What do you think?

viirya: Yeah, looks so. And this looks more reasonable than setting `returnNullable` when `definedByConstructorParams` is true.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/ScalaReflection.scala
viirya: Once #15979 is merged, we can use AssertNotNull approach in above discussion. This comment can be simplified then to remove the second case.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
```diff
@@ -171,15 +171,17 @@ case class StaticInvoke(
  * @param arguments An optional list of expressions, whos evaluation will be passed to the function.
  * @param propagateNull When true, and any of the arguments is null, null will be returned instead
  *                      of calling the function.
+ * @param returnNullable When false, return value must be non-null.
```
viirya: Can we refine this comment a bit? Like `When false, indicating the invoked method will return non-null value.`

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/ScalaReflection.scala
cloud-fan: +1

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
kiszk: thanks, done

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/encoders/ExpressionEncoderSuite.scala
kiszk: sure, done

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
kiszk: done

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/ScalaReflection.scala
kiszk: I see. Updated the comment

cloud-fan:
hi @kiszk can you address this comment? https://github.com/apache/spark/pull/15780#discussion_r90199899 thanks!

kiszk:
Of course. I overlooked this comment.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
```diff
@@ -171,15 +171,17 @@ case class StaticInvoke(
  * @param arguments An optional list of expressions, whos evaluation will be passed to the function.
  * @param propagateNull When true, and any of the arguments is null, null will be returned instead
  *                      of calling the function.
+ * @param returnNullable When false, indicating the invoked method will return non-null value.
```
cloud-fan: `will always return non-null value`

sql/core/src/test/scala/org/apache/spark/sql/DatasetSuite.scala
```diff
@@ -969,6 +972,53 @@ class DatasetSuite extends QueryTest with SharedSQLContext {
...
+    assert(df3.schema(0).nullable == true)
+    assert(df3.schema(0).dataType.asInstanceOf[ArrayType].containsNull == false)
+    val df4 = Seq(Seq("a", "b"), Seq("c", "d")).toDF
+    assert(df4.schema(0).nullable == true)
+    assert(df4.schema(0).dataType.asInstanceOf[ArrayType].containsNull == true)
+
+    val df5 = Seq((0, 1.0), (2, 2.0)).toDF("id", "v")
+    assert(df5.schema(0).nullable == false)
+    assert(df5.schema(1).nullable == false)
+    val df6 = Seq((0, 1.0, 3.0), (2, 2.0, 5.0)).toDF("id", "v1", "v2")
```
cloud-fan: to improve test coverage, we should test `(int, double, string)` here, or testing `df6` is same with testing `df5`

cloud-fan:
looks pretty good, left 2 minor comments, thanks for working on it!

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
```diff
@@ -171,15 +171,18 @@ case class StaticInvoke(
  * @param arguments An optional list of expressions, whos evaluation will be passed to the function.
  * @param propagateNull When true, and any of the arguments is null, null will be returned instead
  *                      of calling the function.
+ * @param returnNullable When false, indicating the invoked method will always return
+ *                      non-null value.
```
viirya: nit: indent looks wrong. an extra space here.

viirya:
LGTM except a very minor style comment.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/JavaTypeInference.scala
```diff
@@ -396,12 +396,14 @@ object JavaTypeInference {
...
           val (keyType, valueType) = mapKeyValueType(typeToken)
+
           ExternalMapToCatalyst(
             inputObject,
             ObjectType(keyType.getRawType),
             serializerFor(_, keyType),
             ObjectType(valueType.getRawType),
-            serializerFor(_, valueType)
+            serializerFor(_, valueType),
+            true
```
ueshin: nit: I prefer named argument here.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/ScalaReflection.scala
```diff
@@ -498,7 +498,8 @@ object ScalaReflection extends ScalaReflection {
           dataTypeFor(keyType),
           serializerFor(_, keyType, keyPath),
           dataTypeFor(valueType),
-          serializerFor(_, valueType, valuePath))
+          serializerFor(_, valueType, valuePath),
+          !valueType.typeSymbol.asClass.isPrimitive)
```
ueshin: nit: and named argument here, too.

sql/core/src/test/scala/org/apache/spark/sql/DatasetSuite.scala
kiszk: i see. done

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
kiszk: sure done.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/ScalaReflection.scala
kiszk: ditto

cloud-fan:
merging to master/2.1, thanks for working on it!

rxin:
@cloud-fan why did we merge this in branch-2.1? Does this fix any bugs?

The change looks invasive.

rxin:
I'm going to revert this.

cloud-fan:
oh sorry I thought incorrect nullability is also a bug...  yea this change is too large to backport, I'll be more careful in the future.

PR #15783, opened 2016-11-05
[Minor][DOC] Unify example marks

zhengruifeng:
1, `**Example**` => `**Examples**`, because more algos use `**Examples**`.
2,  delete `### Examples` in `Isotonic regression`, because it's not that special in http://spark.apache.org/docs/latest/ml-classification-regression.html
3, add missing marks for `LDA` and other algos.

## How was this patch tested?
No tests for it only modify doc


docs/ml-tuning.md
```diff
@@ -62,7 +62,7 @@ To help construct the parameter grid, users can use the [`ParamGridBuilder`](api
 
 After identifying the best `ParamMap`, `CrossValidator` finally re-fits the `Estimator` using the best `ParamMap` and the entire dataset.
 
-## Example: model selection via cross-validation
+## Examples: model selection via cross-validation
```
srowen: I'm OK with this standardization overall. Here you still have a `##` section though, was that intended?

zhengruifeng: I hesitated about where to modify this place. For there are not may links in http://spark.apache.org/docs/latest/ml-tuning.html like those in featureAlgos http://spark.apache.org/docs/latest/ml-features.html.
But now, I tend to use `**examples...**` for standardization.

srowen (approved):

srowen:
Merged to master

PR #15784, opened 2016-11-05
[SPARK-18287][SQL] Move hash expressions from misc.scala into hash.scala

rxin:
As the title suggests, this patch moves hash expressions from misc.scala into hash.scala, to make it easier to find the hash functions. I wanted to do this a while ago but decided to wait for the branch-2.1 cut so the chance of conflicts will be smaller.

## How was this patch tested?
Test cases were also moved out of MiscFunctionsSuite into HashExpressionsSuite.


rxin:
cc @hvanhovell 

This is a straight move of code. No logic change.

hvanhovell:

PR #15786, opened 2016-11-05
[SPARK-18261][Structured Streaming] Add statistics to MemorySink for joining 

lw-lin:

Right now, there is no way to join the output of a memory sink with any table:

> UnsupportedOperationException: LeafNode MemoryPlan must implement statistics

This patch adds statistics to MemorySink, making joining snapshots of memory streams with tables possible.

## How was this patch tested?

Added a test case.

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/memory.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/memory.scala
@@ -27,7 +27,7 @@ import org.apache.spark.internal.Logging
 import org.apache.spark.sql._
 import org.apache.spark.sql.catalyst.encoders.encoderFor
 import org.apache.spark.sql.catalyst.expressions.Attribute
-import org.apache.spark.sql.catalyst.plans.logical.LeafNode
+import org.apache.spark.sql.catalyst.plans.logical.{LeafNode, Statistics}
 import org.apache.spark.sql.streaming.OutputMode
 import org.apache.spark.sql.types.StructType
 import org.apache.spark.util.Utils
@@ -212,4 +212,8 @@ class MemorySink(val schema: StructType, outputMode: OutputMode) extends Sink wi
  */
 case class MemoryPlan(sink: MemorySink, output: Seq[Attribute]) extends LeafNode {
   def this(sink: MemorySink) = this(sink, sink.schema.toAttributes)
+
+  private val sizePerRow = sink.schema.toAttributes.map(_.dataType.defaultSize).sum
+
+  override def statistics: Statistics = Statistics(sizePerRow * sink.allData.size)
 }
--- a/sql/core/src/test/scala/org/apache/spark/sql/streaming/MemorySinkSuite.scala
+++ b/sql/core/src/test/scala/org/apache/spark/sql/streaming/MemorySinkSuite.scala
@@ -187,6 +187,31 @@ class MemorySinkSuite extends StreamTest with BeforeAndAfter {
     query.stop()
   }
 
+  test("MemoryPlan statistics for joining") {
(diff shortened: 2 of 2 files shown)


sql/core/src/test/scala/org/apache/spark/sql/streaming/MemorySinkSuite.scala
```diff
@@ -187,6 +187,31 @@ class MemorySinkSuite extends StreamTest with BeforeAndAfter {
     query.stop()
   }
 
+  test("MemoryPlan statistics for joining") {
```
brkyvz: it would also be nicer to just implement a test on the MemoryPlan itself, without needing to start a stream, purely on statistics calculation

lw-lin:
@brkyvz would you take another look

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/memory.scala
```diff
@@ -212,4 +212,8 @@ class MemorySink(val schema: StructType, outputMode: OutputMode) extends Sink wi
  */
 case class MemoryPlan(sink: MemorySink, output: Seq[Attribute]) extends LeafNode {
   def this(sink: MemorySink) = this(sink, sink.schema.toAttributes)
+
+  private val sizePerRow = sink.schema.toAttributes.map(_.dataType.defaultSize).sum
+
+  override def statistics: Statistics = Statistics(sizePerRow * sink.allData.size)
```
brkyvz: everywhere else, this is defined as a lazy val. Can you make it such that:
```scala
override lazy val statistics: Statistics  = {
  val sizePerRow = sink.schema.toAttributes.map(_.dataType.defaultSize).sum
  Statistics(sizeInBytes = sizePerRow * sink.allData.size)
}
```

sql/core/src/test/scala/org/apache/spark/sql/streaming/MemorySinkSuite.scala
```diff
@@ -187,6 +187,22 @@ class MemorySinkSuite extends StreamTest with BeforeAndAfter {
     query.stop()
   }
 
+  test("MemoryPlan statistics") {
+    implicit val schema = new StructType().add(new StructField("value", IntegerType))
```
brkyvz: thanks! This will be much faster than a test where we actually join DataFrames.

brkyvz:
Thanks @lw-lin ! Left one last comment. @davies Can you also please take a look? I think you implemented most of the statistics code

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/memory.scala
lw-lin: thanks @brkyvz. I'm afraid it should not be `lazy val` as the statistics should vary from batch to batch -- `sink.allData.size`should vary from batch to batch. `sizePerRow`, however, does not vary from batch to batch, thus it's `private val` as it is now.

lw-lin:
@zsxwing could you also take a look

sql/core/src/test/scala/org/apache/spark/sql/streaming/MemorySinkSuite.scala
lw-lin: happy to do this

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/memory.scala
brkyvz: oh yeah! you're right! good catch!

rxin:
LGTM - merging in master/branch-2.1.

PR #15788, opened 2016-11-06
[SPARK-18291][SparkR][ML] SparkR glm predict should output original label when family = binomial.

yanboliang:
SparkR ```spark.glm``` predict should output original label when family = "binomial".

## How was this patch tested?
Add unit test.
You can also run the following code to test:
```R
training <- suppressWarnings(createDataFrame(iris))
training <- training[training$Species %in% c("versicolor", "virginica"), ]
model <- spark.glm(training, Species ~ Sepal_Length + Sepal_Width,family = binomial(link = "logit"))
showDF(predict(model, training))
```
Before this change:
```
+------------+-----------+------------+-----------+----------+-----+-------------------+
|Sepal_Length|Sepal_Width|Petal_Length|Petal_Width|   Species|label|         prediction|
+------------+-----------+------------+-----------+----------+-----+-------------------+
|         7.0|        3.2|         4.7|        1.4|versicolor|  0.0| 0.8271421517601544|
|         6.4|        3.2|         4.5|        1.5|versicolor|  0.0| 0.6044595910413112|
|         6.9|        3.1|         4.9|        1.5|versicolor|  0.0| 0.7916340858281998|
|         5.5|        2.3|         4.0|        1.3|versicolor|  0.0|0.16080518180591158|
|         6.5|        2.8|         4.6|        1.5|versicolor|  0.0| 0.6112229217050189|
|         5.7|        2.8|         4.5|        1.3|versicolor|  0.0| 0.2555087295500885|
|         6.3|        3.3|         4.7|        1.6|versicolor|  0.0| 0.5681507664364834|
|         4.9|        2.4|         3.3|        1.0|versicolor|  0.0|0.05990570219972002|
|         6.6|        2.9|         4.6|        1.3|versicolor|  0.0| 0.6644434078306246|
|         5.2|        2.7|         3.9|        1.4|versicolor|  0.0|0.11293577405862379|
|         5.0|        2.0|         3.5|        1.0|versicolor|  0.0|0.06152372321585971|
|         5.9|        3.0|         4.2|        1.5|versicolor|  0.0|0.35250697207602555|
|         6.0|        2.2|         4.0|        1.0|versicolor|  0.0|0.32267018290814303|
|         6.1|        2.9|         4.7|        1.4|versicolor|  0.0|  0.433391153814592|
|         5.6|        2.9|         3.6|        1.3|versicolor|  0.0| 0.2280744262436993|
|         6.7|        3.1|         4.4|        1.4|versicolor|  0.0| 0.7219848389339459|
|         5.6|        3.0|         4.5|        1.5|versicolor|  0.0|0.23527698971404695|
|         5.8|        2.7|         4.1|        1.0|versicolor|  0.0|  0.285024533520016|
|         6.2|        2.2|         4.5|        1.5|versicolor|  0.0| 0.4107047877447493|
|         5.6|        2.5|         3.9|        1.1|versicolor|  0.0|0.20083561961645083|
+------------+-----------+------------+-----------+----------+-----+-------------------+
```
After this change: 
```
+------------+-----------+------------+-----------+----------+-----+----------+
|Sepal_Length|Sepal_Width|Petal_Length|Petal_Width|   Species|label|prediction|
+------------+-----------+------------+-----------+----------+-----+----------+
|         7.0|        3.2|         4.7|        1.4|versicolor|  0.0| virginica|
|         6.4|        3.2|         4.5|        1.5|versicolor|  0.0| virginica|
|         6.9|        3.1|         4.9|        1.5|versicolor|  0.0| virginica|
|         5.5|        2.3|         4.0|        1.3|versicolor|  0.0|versicolor|
|         6.5|        2.8|         4.6|        1.5|versicolor|  0.0| virginica|
|         5.7|        2.8|         4.5|        1.3|versicolor|  0.0|versicolor|
|         6.3|        3.3|         4.7|        1.6|versicolor|  0.0| virginica|
|         4.9|        2.4|         3.3|        1.0|versicolor|  0.0|versicolor|
|         6.6|        2.9|         4.6|        1.3|versicolor|  0.0| virginica|
|         5.2|        2.7|         3.9|        1.4|versicolor|  0.0|versicolor|
|         5.0|        2.0|         3.5|        1.0|versicolor|  0.0|versicolor|
|         5.9|        3.0|         4.2|        1.5|versicolor|  0.0|versicolor|
|         6.0|        2.2|         4.0|        1.0|versicolor|  0.0|versicolor|
|         6.1|        2.9|         4.7|        1.4|versicolor|  0.0|versicolor|
|         5.6|        2.9|         3.6|        1.3|versicolor|  0.0|versicolor|
|         6.7|        3.1|         4.4|        1.4|versicolor|  0.0| virginica|
|         5.6|        3.0|         4.5|        1.5|versicolor|  0.0|versicolor|
|         5.8|        2.7|         4.1|        1.0|versicolor|  0.0|versicolor|
|         6.2|        2.2|         4.5|        1.5|versicolor|  0.0|versicolor|
|         5.6|        2.5|         3.9|        1.1|versicolor|  0.0|versicolor|
+------------+-----------+------------+-----------+----------+-----+----------+
```

### DIFF
--- a/mllib/src/main/scala/org/apache/spark/ml/r/GeneralizedLinearRegressionWrapper.scala
+++ b/mllib/src/main/scala/org/apache/spark/ml/r/GeneralizedLinearRegressionWrapper.scala
@@ -26,8 +26,13 @@ import org.apache.spark.ml.{Pipeline, PipelineModel}
 import org.apache.spark.ml.attribute.{Attribute, AttributeGroup, NominalAttribute}
 import org.apache.spark.ml.feature.{IndexToString, RFormula}
 import org.apache.spark.ml.regression._
+import org.apache.spark.ml.Transformer
+import org.apache.spark.ml.param.ParamMap
+import org.apache.spark.ml.param.shared._
 import org.apache.spark.ml.util._
 import org.apache.spark.sql._
+import org.apache.spark.sql.functions._
+import org.apache.spark.sql.types._
 
 private[r] class GeneralizedLinearRegressionWrapper private (
     val pipeline: PipelineModel,
@@ -56,6 +61,7 @@ private[r] class GeneralizedLinearRegressionWrapper private (
   def transform(dataset: Dataset[_]): DataFrame = {
     if (rFamily == "binomial") {
       pipeline.transform(dataset)
+        .drop(PREDICTED_LABEL_PROB_COL)
         .drop(PREDICTED_LABEL_INDEX_COL)
         .drop(glm.getFeaturesCol)
     } else {
@@ -71,6 +77,7 @@ private[r] class GeneralizedLinearRegressionWrapper private (
 private[r] object GeneralizedLinearRegressionWrapper
   extends MLReadable[GeneralizedLinearRegressionWrapper] {
 
+  val PREDICTED_LABEL_PROB_COL = "pred_label_prob"
   val PREDICTED_LABEL_INDEX_COL = "pred_label_idx"
   val PREDICTED_LABEL_COL = "prediction"
 
@@ -89,9 +96,6 @@ private[r] object GeneralizedLinearRegressionWrapper
     val rFormulaModel = rFormula.fit(data)
     // get labels and feature names from output schema
     val schema = rFormulaModel.transform(data).schema
-    val labelAttr = Attribute.fromStructField(schema(rFormulaModel.getLabelCol))
-      .asInstanceOf[NominalAttribute]
-    val labels = labelAttr.values.get
     val featureAttrs = AttributeGroup.fromStructField(schema(rFormula.getFeaturesCol))
       .attributes.get
     val features = featureAttrs.map(_.name.get)
@@ -106,12 +110,22 @@ private[r] object GeneralizedLinearRegressionWrapper
       .setRegParam(regParam)
       .setFeaturesCol(rFormula.getFeaturesCol)
     val pipeline = if (family == "binomial") {
+      // Convert prediction from probability to label index.
+      val probToPred = new ProbabilityToPrediction()
+        .setInputCol(PREDICTED_LABEL_PROB_COL)
+        .setOutputCol(PREDICTED_LABEL_INDEX_COL)
+      // Convert prediction from label index to original label.
+      val labelAttr = Attribute.fromStructField(schema(rFormulaModel.getLabelCol))
+        .asInstanceOf[NominalAttribute]
+      val labels = labelAttr.values.get
       val idxToStr = new IndexToString()
         .setInputCol(PREDICTED_LABEL_INDEX_COL)
         .setOutputCol(PREDICTED_LABEL_COL)
         .setLabels(labels)
+
       new Pipeline()
-        .setStages(Array(rFormulaModel, glr.setPredictionCol(PREDICTED_LABEL_INDEX_COL), idxToStr))
+        .setStages(Array(rFormulaModel, glr.setPredictionCol(PREDICTED_LABEL_PROB_COL),
+          probToPred, idxToStr))
         .fit(data)
     } else {
       new Pipeline().setStages(Array(rFormulaModel, glr)).fit(data)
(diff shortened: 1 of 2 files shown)


felixcheung:
LGTM

mllib/src/main/scala/org/apache/spark/ml/r/GeneralizedLinearRegressionWrapper.scala
```diff
@@ -200,3 +237,27 @@ private[r] object GeneralizedLinearRegressionWrapper
     }
   }
 }
+
+/**
+ * This utility transformer converts the predicted value of GeneralizedLinearRegressionModel
```
yanboliang: This is an inherent feature of other classification algorithms who extends ```Classifier``` currently, only GLM use the separate conversion since ```GeneralizedLinearRegression extends Regressor```. So it should not be reused, and we can move it to RWrapperUtils.scala when it's necessary. Thanks.

yanboliang:
Merged into master and branch-2.1. Thanks for review.

jkbradley:
@felixcheung @yanboliang  + @shivaram @mengxr  Just saw this change.  IIRC we intentionally made spark.glm match R's glm behavior for the binomial family, though I see it was a little different: glm outputs log-space predictions (rawPrediction) by default, and spark.glm output probabilities by default before this PR.  This PR makes spark.glm match Scala/Java/Python MLlib APIs instead of R libraries.

We should pick a clear strategy here: match the rest of MLlib or match familiar R libraries.

mllib/src/main/scala/org/apache/spark/ml/r/GeneralizedLinearRegressionWrapper.scala
```diff
@@ -52,7 +59,15 @@ private[r] class GeneralizedLinearRegressionWrapper private (
   def residuals(residualsType: String): DataFrame = glm.summary.residuals(residualsType)
 
   def transform(dataset: Dataset[_]): DataFrame = {
-    pipeline.transform(dataset).drop(glm.getFeaturesCol)
+    if (rFamily == "binomial") {
+      pipeline.transform(dataset)
+        .drop(PREDICTED_LABEL_PROB_COL)
```
jkbradley: This is going to make R models incompatible with other languages when you persist them.  (They already are because of having hidden Pipelines, but that is fixable.)  This, however, will encode special behavior which is only respected when the model is loaded from R, not from other languages.

One option is to encode this in a SQLTransformer.

I'm also worried that these hard-coded columns names will lead to future bug reports about conflicting input column names.

It looks like this same issue appears in other PRs for R, such as [https://issues.apache.org/jira/browse/SPARK-18401].  Given the pervasiveness and that we're in QA right now, I'd recommend we not worry about it for 2.1 and delay fixing it until 2.2.

yanboliang:
@jkbradley Thanks for your comments. I fully understand and also carefully considered your concerns. I totally agree we should make ```spark.glm``` match R’s ```glm``` as much as possible. However, I found the output prediction of ```spark.glm``` for the binomial family is meaningless if we only provides the probability value.

Let’s look at how R ```glm``` works for binomial family(with string label):
```R
data <- iris[iris$Species %in% c("versicolor", "virginica"), ]
model <- glm(Species ~ ., data = data, family = binomial(link = "logit”))
predict(model, data, type = "response”)
f <- factor(data$Species)
prediction <- round(predict(model, data, type = "response"))
factor(prediction, labels = levels(f))
```
The label is string and R ```glm``` will encode it into factor automatically when training. The default prediction output is log-space prediction, but we can convert it into probability by specifying ```type = "response"```, then we can get the prediction value in the form ```0``` and ```1```.
However, native R ```factor``` method can get the map between the string values and the converted numeric values. For example, ```versicolor``` maps to ```0``` and ```virginica``` maps to ```1```. Users can convert the numeric prediction value into string value and vice versa.

In Spark, we use ```StringIndexer``` to encode label and ```IndexToString``` to convert back. This was wrapped in ```RFormula``` and SparkR users can not use it which means they can not get the map between string label and numeric label. I think getting the original label as prediction value is one of the most important use cases in SparkR users, so I proposed to make this change.
Another option is to implement ```spark.factor``` as a function to SparkR users, then they will be able to use it as ```factor``` in native R. But I think the internal of ```RFormula```(which is closely related with the implementation of ```spark.factor```) is private for SparkR users, and we are not ready to make it public(We are still making improvements to ```RFormula```, see [SPARK-15540](https://issues.apache.org/jira/browse/SPARK-15540)), so I give up this opinion.

mllib/src/main/scala/org/apache/spark/ml/r/GeneralizedLinearRegressionWrapper.scala
yanboliang: Yeah, I totally agree the hard-coded column names issues should be fixed, and already have some ideas in my mind to improve SparkR ML wrappers(which include this). This can be placed in the plan of next release version and I will write simple design documents for reviewing.
For the ```ProbabilityToPrediction``` issue, the idea of ```SQLTransformer``` sounds good and I will try to fix it in follow-up PR. Thanks.

jkbradley:
I agree that R makes it much easier to convert back to the string label.  My main worries are:
* changing the API, which may break some user code
* hiding probabilities from R users  (since I'd expect many would want to analyze the probabilities)

One alternative would be to output both the string prediction and the probability as 2 columns, which would make the SparkR API a little closer to the other language APIs.

yanboliang:
@jkbradley I agree that exposing probabilities is also useful to R users. Think more about this issue, I think we can add a new argument ```type``` to ```predict```, this argument controls the type of prediction required. This will match the behavior of R [```glm.predict```](https://stat.ethz.ch/R-manual/R-devel/library/stats/html/predict.glm.html) and other native R predict method.

PR #15790, opened 2016-11-06
[SPARK-18264][SPARKR] build vignettes with package, update vignettes for CRAN release build and add info on release

felixcheung:

Changes to DESCRIPTION to build vignettes.
Changes the metadata for vignettes to generate the recommended format (which is about <10% of size before). Unfortunately it does not look as nice
(before - left, after - right)

image

image

Also add information on how to run build/release to CRAN later.

## How was this patch tested?

manually, unit tests

@shivaram 

We need this for branch-2.1

### DIFF
--- a/R/CRAN_RELEASE.md
+++ b/R/CRAN_RELEASE.md
@@ -5,7 +5,7 @@ To release SparkR as a package to CRAN, we would use the `devtools` package. Ple
 
 ### Release
 
-First, check that the `Version:` field in the `pkg/DESCRIPTION` file is updated. Also, check for stale file not under source control.
+First, check that the `Version:` field in the `pkg/DESCRIPTION` file is updated. Also, check for stale files not under source control.
 
 Note that while `check-cran.sh` is running `R CMD check`, it is doing so with `--no-manual --no-vignettes`, which skips a few vignettes or PDF checks - therefore it will be preferred to run `R CMD check` on the source package built manually before uploading a release.
 
@@ -56,6 +56,23 @@ man/
 
 vignettes/
 sparkr-vignettes.Rmd
+```
+
+#### Test source package
+
+To install, run this:
+
+```sh
+R CMD INSTALL SparkR_2.1.0.tar.gz
+```
+
+With "2.1.0" replaced with the version of SparkR.
+
+This command installs SparkR to the default libPaths. Once that is done, you should be able to start R and run:
+
+```R
+library(SparkR)
+vignette("sparkr-vignettes", package="SparkR")
 ```
 
 #### Build binary package
--- a/R/CRAN_RELEASE.md
+++ b/R/CRAN_RELEASE.md
@@ -1,0 +1,70 @@
+# SparkR CRAN Release
+
+To release SparkR as a package to CRAN, we would use the `devtools` package. Please work with the
+`dev@spark.apache.org` community and R package maintainer on this.
+
+### Release
+
+First, check that the `Version:` field in the `pkg/DESCRIPTION` file is updated. Also, check for stale file not under source control.
+
+To upload a release, we would need to update the `cran-comments.md`. This should generally contain the results from running the `check-cran.sh` script along with comments on status of all `WARNING` (should not be any) or `NOTE`. As a part of `check-cran.sh` and the release process, the vignettes is build - make sure `SPARK_HOME` is set and Spark jars are accessible.
+
+Once everything is in place, run in R under the `SPARK_HOME/R` directory:
+
+```R
+paths <- .libPaths(); .libPaths(c("lib", paths)); Sys.setenv(SPARK_HOME=tools::file_path_as_absolute("..")); devtools::release(); .libPaths(paths)
+```
+
+For more information please refer to http://r-pkgs.had.co.nz/release.html#release-check
+
+### Testing: build package manually
+
+To build package manually such as to inspect the resulting `.tar.gz` file content, we would also use the `devtools` package.
+
+Source package is what get released to CRAN. CRAN would then build platform-specific binary packages from the source package.
+
+#### Build source package
+
+To build source package locally without releasing to CRAN, run in R under the `SPARK_HOME/R` directory:
+
+```R
+paths <- .libPaths(); .libPaths(c("lib", paths)); Sys.setenv(SPARK_HOME=tools::file_path_as_absolute("..")); devtools::build("pkg"); .libPaths(paths)
+```
+
+(http://r-pkgs.had.co.nz/vignettes.html#vignette-workflow-2)
+
+For example, this should be the content of the source package:
+
+```sh
+DESCRIPTION	R		inst		tests
+NAMESPACE	build		man		vignettes
+
+inst/doc/
+sparkr-vignettes.html
+sparkr-vignettes.Rmd
+sparkr-vignettes.Rman
+
+build/
+vignette.rds
+
+man/
+ *.Rd files...
+
+vignettes/
+sparkr-vignettes.Rmd
+```
+
+#### Build binary package
+
+To build binary package locally, run in R under the `SPARK_HOME/R` directory:
+
+```R
+paths <- .libPaths(); .libPaths(c("lib", paths)); Sys.setenv(SPARK_HOME=tools::file_path_as_absolute("..")); devtools::build("pkg", binary = TRUE); .libPaths(paths)
+```
+
+For example, this should be the content of the binary package:
+
+```sh
+DESCRIPTION	Meta		R		html		tests
+INDEX		NAMESPACE	help		profile		worker
+```
--- a/R/CRAN_RELEASE.md
+++ b/R/CRAN_RELEASE.md
@@ -7,6 +7,8 @@ To release SparkR as a package to CRAN, we would use the `devtools` package. Ple
 
 First, check that the `Version:` field in the `pkg/DESCRIPTION` file is updated. Also, check for stale file not under source control.
 
+Note that while `check-cran.sh` is running `R CMD check`, it is doing so with `--no-manual --no-vignettes`, which skips a few vignettes or PDF checks - therefore it will be preferred to run `R CMD check` on the source package built manually before uploading a release.
+
 To upload a release, we would need to update the `cran-comments.md`. This should generally contain the results from running the `check-cran.sh` script along with comments on status of all `WARNING` (should not be any) or `NOTE`. As a part of `check-cran.sh` and the release process, the vignettes is build - make sure `SPARK_HOME` is set and Spark jars are accessible.
 
 Once everything is in place, run in R under the `SPARK_HOME/R` directory:
@@ -32,6 +34,8 @@ paths <- .libPaths(); .libPaths(c("lib", paths)); Sys.setenv(SPARK_HOME=tools::f
 ```
 
 (http://r-pkgs.had.co.nz/vignettes.html#vignette-workflow-2)
+
+Similarly, the source package is also created by `check-cran.sh` with `R CMD build pkg`.
 
 For example, this should be the content of the source package:
 
--- a/R/check-cran.sh
+++ b/R/check-cran.sh
@@ -37,7 +37,7 @@ fi
 echo "USING R_HOME = $R_HOME"
 
 # Build the latest docs, but not vignettes, which is built with the package next
-NO_VIGNETTES=1 $FWDIR/create-docs.sh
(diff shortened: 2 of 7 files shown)


shivaram:
This is great @felixcheung - Taking a closer look now

felixcheung:
interesting. jenkins failed because it is building vignettes again right after it is built correctly, and the second time it didn't have the Spark jar (so trying to download it as per install.spark)

I'll try to track down where it is doing that. 

```
  Use Petal_Width instead of Petal.Width  as column name
/home/jenkins/workspace/SparkPullRequestBuilder/R
* checking for file '/home/jenkins/workspace/SparkPullRequestBuilder/R/pkg/DESCRIPTION' ... OK
* preparing 'SparkR':
* checking DESCRIPTION meta-information ... OK
* installing the package to build vignettes
* creating vignettes ... ERROR

Attaching package: 'SparkR'

The following objects are masked from 'package:stats':

The following objects are masked from 'package:base':

    as.data.frame, colnames, colnames<-, drop, intersect, rank,
    rbind, sample, subset, summary, transform, union

Spark not found in SPARK_HOME: 
Spark not found in the cache directory. Installation will start.
MirrorUrl not provided.
Looking for preferred site from apache website...
Preferred mirror site found: http://mirror.nexcess.net/apache/spark
Downloading spark-2.1.0 for Hadoop 2.7 from:
- http://mirror.nexcess.net/apache/spark/spark-2.1.0/spark-2.1.0-bin-hadoop2.7.tgz
trying URL 'http://mirror.nexcess.net/apache/spark/spark-2.1.0/spark-2.1.0-bin-hadoop2.7.tgz'
Fetch failed from http://mirror.nexcess.net/apache/spark
To use backup site...
Downloading spark-2.1.0 for Hadoop 2.7 from:
- http://www-us.apache.org/dist/spark/spark-2.1.0/spark-2.1.0-bin-hadoop2.7.tgz
trying URL 'http://www-us.apache.org/dist/spark/spark-2.1.0/spark-2.1.0-bin-hadoop2.7.tgz'
Fetch failed from http://www-us.apache.org/dist/spark
Quitting from lines 31-32 (sparkr-vignettes.Rmd) 
Error: processing vignette 'sparkr-vignettes.Rmd' failed with diagnostics:
Unable to download Spark spark-2.1.0 for Hadoop 2.7. Please check network connection, Hadoop version, or provide other mirror sites.
Execution halted
```

felixcheung:
Basically it is building vignettes multiple times - once in create-doc.sh, once with `R CMD build`.

I'll skip the one in create-doc.sh to get it to create vignettes together with the source package, which I think is the more nature or common behavior.

Also, we are currently passing `--no-manual` to `R CMD check` (in check-cran.sh while running from run-tests.sh), otherwise it will once again try to build vignettes (this time from the source package)

felixcheung:
I looked the `qpdf` warning is from a fairly strict [check](https://github.com/wch/r-source/blob/tags/R-3-3-2/src/library/tools/R/check.R#L325) - I don't see a way around it other than getting qpdf on the system (it's a tool, not a R package).

shivaram:
Can you open a Spark JIRA for installing `qpdf` and cc @shaneknapp on it ? We can install it on the Jenkins machines.

felixcheung:
opened SPARK-18347

shaneknapp:
test this please

shaneknapp:
ok, qpdf is installed per the jira.   no environment variables set up yet though.

R/run-tests.sh
```diff
@@ -48,7 +48,8 @@ if [[ $FAILED != 0 || $NUM_TEST_WARNING != 0 ]]; then
 else
     # We have 2 existing NOTEs for new maintainer, attach()
     # We have one more NOTE in Jenkins due to "No repository set"
-    if [[ $NUM_CRAN_WARNING != 0 || $NUM_CRAN_ERROR != 0 || $NUM_CRAN_NOTES -gt 3 ]]; then
+    # We have one warning on ‘qpdf’ is needed for checks on size reduction of PDFs
+    if [[ $NUM_CRAN_WARNING != 1 || $NUM_CRAN_ERROR != 0 || $NUM_CRAN_NOTES -gt 3 ]]; then
```
shivaram: I think we can remove this now. This was the cause for the latest test failure ?

felixcheung:
this is good to go in?

felixcheung:
ping @shivaram

shivaram:
Sorry I got caught up with some other stuff - Will take a look at this today.

R/CRAN_RELEASE.md
```diff
@@ -0,0 +1,74 @@
+# SparkR CRAN Release
+
+To release SparkR as a package to CRAN, we would use the `devtools` package. Please work with the
+`dev@spark.apache.org` community and R package maintainer on this.
+
+### Release
+
+First, check that the `Version:` field in the `pkg/DESCRIPTION` file is updated. Also, check for stale file not under source control.
```
shivaram: nit - `stale files`

R/pkg/vignettes/sparkr-vignettes.Rmd
```diff
@@ -1,12 +1,13 @@
...
 title: "SparkR - Practical Guide"
 output:
-  html_document:
-    theme: united
+  rmarkdown::html_vignette:
     toc: true
     toc_depth: 4
-    toc_float: true
-    highlight: textmate
+vignette: >
```
shivaram: So I think the `theme` was the one giving us the nice looking HTML . Is that not supported with `rmarkdown::html_vignette` ?

shivaram (commented):
Thanks @felixcheung - Change looks pretty good to me. I just had a couple of minor inline comments. I also want to try out the scripts on my machine - but I can do that after the merge as well.

R/check-cran.sh
```diff
@@ -36,11 +36,27 @@ if [ ! -z "$R_HOME" ]
...
 
-# Build the latest docs
-$FWDIR/create-docs.sh
+# Build the latest docs, but not vignettes, which is built with the package next
+NO_VIGNETTES=1 $FWDIR/create-docs.sh
 
-# Build a zip file containing the source package
-"$R_SCRIPT_PATH/"R CMD build $FWDIR/pkg
+# Build source package with vignettes
+SPARK_HOME="$(cd "`dirname "$0"`"/..; pwd)"
```
shivaram: So I ran into a problem while trying to run this locally. It seems to be getting `SPARK_HOME` wrong for some reason
````
./R/check-cran.sh: line 45: /Users/shivaram/spark-1/R/bin/load-spark-env.sh: No such file or directory
````

**Comment by shivaram** [comment]
@felixcheung I noticed one more thing - We are somehow not registering the vignette correctly with the R package. So for example if I launch `./bin/sparkR` and then run `vignette(package="dplyr")`
I see a list of vignettes that I can then launch with `vignette("introduction", package="dplyr")`. However this doesn't seem to work with our vignette - I'm not sure what we need to do to get this to work though.

File: R/pkg/vignettes/sparkr-vignettes.Rmd (hunk 3 above)
Comment (felixcheung) [inline R/pkg/vignettes/sparkr-vignettes.Rmd]: right, it seems to have a theme already and is complaining about theme already defined

File: R/check-cran.sh (hunk 4 above)
Comment (felixcheung) [inline R/check-cran.sh]: that's odd, it works in my environment.
do you happen to have check-cran.sh under a subdirectory of R?

this is the code:
```
SPARK_HOME="$(cd "`dirname "$0"`"/..; pwd)"
. "${SPARK_HOME}"/bin/load-spark-env.sh
```

so for /opt/spark/R/check-cran.sh, it should look for /opt/spark/bin/load-spark-env.sh
and SPARK_HOME should be /opt/spark

Comment (shivaram) [inline R/check-cran.sh]: So I think this needs to be at the top of the file - The current working directory is probably changed in the middle by `create-docs.sh` or the `pushd $FWDIR`  ?

**Comment by felixcheung** [comment]
Tested that more, I think the vignettes works only with installed package (or installed built package)

```
R CMD INSTALL SparkR_2.1.0.tar.gz
```

and then
```
library(SparkR)
vignette("sparkr-vignettes", package="SparkR")
```

File: R/run-tests.sh [hunk 5]
```diff
@@ -48,6 +48,7 @@ if [[ $FAILED != 0 || $NUM_TEST_WARNING != 0 ]]; then
 else
     # We have 2 existing NOTEs for new maintainer, attach()
     # We have one more NOTE in Jenkins due to "No repository set"
+    # We have one warning on ‘qpdf’ is needed for checks on size reduction of PDFs
```
shivaram: Nit: This comment can be removed now ?

R/pkg/vignettes/sparkr-vignettes.Rmd
shivaram: I see - I think thats fine then.

R/create-docs.sh
```diff
@@ -52,21 +52,28 @@ Rscript -e 'libDir <- "../../lib"; library(SparkR, lib.loc=libDir); library(knit
...
+  echo "Skipping R vignettes"
 else
-  SPARK_JARS_DIR="${SPARK_HOME}/assembly/target/scala-$SPARK_SCALA_VERSION/jars"
-fi
 
-# Only create vignettes if Spark JARs exist
-if [ -d "$SPARK_JARS_DIR" ]; then
-  # render creates SparkR vignettes
-  Rscript -e 'library(rmarkdown); paths <- .libPaths(); .libPaths(c("lib", paths)); Sys.setenv(SPARK_HOME=tools::file_path_as_absolute("..")); render("pkg/vignettes/sparkr-vignettes.Rmd"); .libPaths(paths)'
+  # Find Spark jars.
```
shivaram: This code is now duplicated in two files. Do you think we should just create a `build-vignette.sh` and call it in two places ?

shivaram (commented):
Thanks for the update and especially for figuring out the install step. I just had a few more minor comments.

R/check-cran.sh
felixcheung: what that command is doing is
1. get the dirname of the currently running file, check-cran.sh (`dirname "$0"`)
2. cd to one level up (`cd x/..`)
3. print current directory `pwd`
4. set that to SPARK_HOME

so it shouldn't matter what the current directory is but only with where the check-cran.sh is found?

I'm running this on Ubuntu, it's possible if you are running on Mac the behavior is different somehow?

R/create-docs.sh
felixcheung: right, I thought about that. I think there's value in knitting html and vignettes in create-doc.sh, it is a bit duplicated to have vignettes in 2 places but 

in create-doc.sh
```
...
render("pkg/vignettes/sparkr-vignettes.Rmd");
```

in check-cran.sh
```
  SPARK_HOME="${SPARK_HOME}" "$R_SCRIPT_PATH/"R CMD build $FWDIR/pkg
```

in the latter it is building the full package along with vignettes, so the actual command and behavior isn't exactly the same.

Perhaps we should just take vignettes build out of create-doc.sh?

felixcheung: I changed this up a bit, not sure if it helps but both works on my setup.

felixcheung: I changed this up a bit, not sure if it helps but both works on my setup

R/check-cran.sh
felixcheung: I changed this up a bit, not sure if it helps but both works on my setup

R/create-docs.sh
shivaram: Yeah that seems fine to me. the `create-doc.sh` is used to generate docs for the Spark website etc. so its probably ok to not do the vignette as a part of that.

felixcheung:
I think https://github.com/apache/spark/blob/master/dev/make-distribution.sh should change too but I'm not 100% how the R package is built there.

shivaram:
So one proposal I was thinking of is to just check in a built version of the vignette in to the source tree. That way the release packaging wouldn't need to change. The only thing to keep in mind is that whenever we update the vignette we will need to rebuild it. Thoughts ?

felixcheung:
Problem is the required and generated `vignette.rds` RDS file is a binary file?
I'm not sure about checking in binaries in git, that would show up in a source-only release?
Maybe create-distribution.sh should run the equivalent of `R CMD install SparkR.tar.gz` to generate the binaries necessary that would only go to a binary release?

How about we merge this PR first - I can test out the release mechanism more next week.

shivaram:
Sure - Sounds good. LGTM. Merging this to master and branch-2.1

PR #15787, opened 2016-11-06
[SPARK-18286][ML] Add Scala/Java examples for MinHash and RandomProjection

jiayue-zhang:
## What changes were proposed in this pull request?

Add Scala/Java/Python examples for MinHash and RandomProjection

## How was this patch tested?

`./bin/run-example ml.MinHashExample`
`./bin/run-example ml.JavaMinHashExample`
`./bin/run-example ml.RandomProjectionExample`
`./bin/run-example ml.JavaRandomProjectionExample`

PR #15795, opened 2016-11-07
[SPARK-18081][ML][DOCS] Add user guide for Locality Sensitive Hashing(LSH)

Yunni:
The user guide for LSH is added to ml-features.md, with several scala/java examples in spark-examples.

## How was this patch tested?
Doc has been generated through Jekyll, and checked through manual inspection.


examples/src/main/java/org/apache/spark/examples/ml/JavaRandomProjectionExample.java
```diff
@@ -0,0 +1,72 @@
...
+import org.apache.spark.sql.types.DataTypes;
+import org.apache.spark.sql.types.Metadata;
+import org.apache.spark.sql.types.StructField;
+import org.apache.spark.sql.types.StructType;
+// $example off$
+
+public class JavaRandomProjectionExample {
+    public static void main(String[] args) {
+        SparkSession spark = SparkSession
+                .builder()
```
srowen: Fix the indentation in many of the Java examples -- 4 space for continuation.

docs/ml-features.md
```diff
@@ -1396,3 +1396,134 @@ for more details on the API.
 {% include_example python/ml/chisq_selector_example.py %}
 </div>
 </div>
+
+# Locality Sensitive Hashing
+[Locality Sensitive Hashing(LSH)](https://en.wikipedia.org/wiki/Locality-sensitive_hashing) is a class of dimension reduction hash families, which can be used as both feature transformation and machine-learned ranking. Difference distance metric has its own LSH family class in `spark.ml`, which can transform feature columns to hash values as new columns. Besides feature transforming, `spark.ml` also implemented approximate nearest neighbor algorithm and approximate similarity join algorithm using LSH.
```
srowen: Despite the opening sentence of the wikipedia article, I wouldn't class LSH as a dimensionality reduction technique? It's a set of hashing techniques where the hash preserves some properties. Maybe it's just my taste. But the rest of the text talks about the output as hash values.

What does "machine-learned ranking" refer to here? as this isn't a ranking technique per se.

I think this is missing a broad summary statement that indicates why LSH is even of interest: it provides a hash function where hashed values are in some sense close when the input values are close according to some metric. And then the variations below plug in different definitions of "close" and "input".

docs/ml-features.md
```diff
@@ -1396,3 +1396,134 @@ for more details on the API.
...
+\]`
+Its LSH family projects features onto a random unit vector and divide the projected results to hash buckets:
+`\[
+h(\mathbf{x}) = \lfloor \frac{\mathbf{x} \cdot \mathbf{v}}{r} \rfloor
+\]`
+where `v` is a normalized random unit vector and `r` is user-defined bucket length.
+
+The input features in Euclidean space are represented in vectors. Both sparse and dense vectors are supported.
+
+The bucket length can be used to trade off the performance of random projection. A larger bucket lowers the false negative rate but usually increases running time and false positive rate.
```
srowen: Isn't the point here that near vectors end up in nearby buckets? I feel like this is skipping the intuition of why you would care about this technique or when you'd use it. Not like this needs a whole paragraph, just a few sentences of pointers?

docs/ml-features.md
```diff
@@ -1396,3 +1396,134 @@ for more details on the API.
...
+[MinHash](https://en.wikipedia.org/wiki/MinHash) is the LSH family in `spark.ml` for Jaccard distance where input features are sets of natural numbers. Jaccard distance of two sets is defined by the cardinality of their intersection and union:
+`\[
+d(\mathbf{A}, \mathbf{B}) = 1 - \frac{|\mathbf{A} \cap \mathbf{B}|}{|\mathbf{A} \cup \mathbf{B}|}
+\]`
+As its LSH family, MinHash applies a random [perfect hash function](https://en.wikipedia.org/wiki/Perfect_hash_function) `g` to each elements in the set and take the minimum of all hashed values:
+`\[
+h(\mathbf{A}) = \min_{a \in \mathbf{A}}(g(a))
+\]`
+
+Input sets for MinHash is represented in vectors which dimension equals the total number of elements in the space. Each dimension of the vectors represents the status of an elements: zero value means the elements is not in the set; non-zero value means the set contains the corresponding elements. For example, `Vectors.sparse(10, Array[(2, 1.0), (3, 1.0), (5, 1.0)])` means there are 10 elements in the space. This set contains elem 2, elem 3 and elem 5.
```
srowen: Same sort of comment: isn't the intuition that MinHash approximates the Jaccard similarity without actually computing it completely?

This may again reduce to my different understanding of the technique, but MinHash isn't a hashing technique per se. it relies on a given family of hash functions to approximate set similarity. I'm finding it a little hard to view it as an 'LSH family'

docs/ml-features.md
```diff
@@ -1396,3 +1396,134 @@ for more details on the API.
...
+<div data-lang="java" markdown="1">
+
+Refer to the [RandomProjection Java docs](api/java/org/apache/spark/ml/feature/RandomProjection.html)
+for more details on the API.
+
+{% include_example java/org/apache/spark/examples/ml/JavaRandomProjectionExample.java %}
+</div>
+</div>
+
+## MinHash for Jaccard Distance
```
srowen: Similarity, not distance, right? it's higher when they overlap more.

docs/ml-features.md
```diff
@@ -1396,3 +1396,134 @@ for more details on the API.
...
 </div>
 </div>
+
+# Locality Sensitive Hashing
+[Locality Sensitive Hashing(LSH)](https://en.wikipedia.org/wiki/Locality-sensitive_hashing) is a class of dimension reduction hash families, which can be used as both feature transformation and machine-learned ranking. Difference distance metric has its own LSH family class in `spark.ml`, which can transform feature columns to hash values as new columns. Besides feature transforming, `spark.ml` also implemented approximate nearest neighbor algorithm and approximate similarity join algorithm using LSH.
+
+In this section, we call a pair of input features a false positive if the two features are hashed into the same hash bucket but they are far away in distance, and we define false negative as the pair of features when their distance are close but they are not in the same hash bucket.
+
+## Random Projection for Euclidean Distance
+**Note:** Please note that this is different than the [Random Projection for cosine distance](https://en.wikipedia.org/wiki/Locality-sensitive_hashing#Random_projection).
```
srowen: Just an aside, but this random projection + cosine distance technique is the main thing I think of when I think of "LSH". Is that not implemented?

srowen (changes requested):

srowen:
Also this seems to overlap with https://github.com/apache/spark/pull/15787

examples/src/main/java/org/apache/spark/examples/ml/JavaRandomProjectionExample.java
```diff
@@ -0,0 +1,72 @@
...
+                .appName("JavaMinHashExample")
```
bravo-zhang: JavaRandomProjectionExample

examples/src/main/scala/org/apache/spark/examples/ml/RandomProjectionExample.scala
```diff
@@ -0,0 +1,56 @@
...
+import org.apache.spark.ml.linalg.Vectors
+// $example off$
+import org.apache.spark.sql.SparkSession
+
+object RandomProjectionExample {
+  def main(args: Array[String]): Unit = {
+    // Creates a SparkSession
+    val spark = SparkSession
+      .builder
+      .appName("MinHashExample")
```
bravo-zhang: RandomProjectionExample

bravo-zhang:
Examples are usually class/algorithm based, not interface/use case based. Maybe we can summarize the 5 classes into 2? Do you mind to modify the examples in #15787 after it is merged instead?

docs/ml-features.md
Yunni: Not sure how MinHash approximates the Jaccard similarity? It is true that `Pr(min(h(A)) = min(h(B)))` is equal to Jaccard similarity when `h` is picked from a universal hash family. But I think we are not computing `Pr(min(h(A)) = min(h(B)))` in MinHash, we are only use this property to construct an LSH function.

docs/ml-features.md
Yunni: Jaccard Distance is just `1 - JaccardSimilarity`(https://en.wikipedia.org/wiki/Jaccard_index)

There are 2 reasons we use Jaccard Distance instead of similarity:
(1) It's cleaner to map each distance metric to their LSH family by the definition of LSH.
(2) In `approxNearestNeighbor` and `approxSimilarityJoin`, the returned dataset has a `distCol` showing the distance values.

examples/src/main/java/org/apache/spark/examples/ml/JavaRandomProjectionExample.java
Yunni: Used 2 space as in other java examples.

docs/ml-features.md
Yunni: Rephrased. PTAL

examples/src/main/java/org/apache/spark/examples/ml/JavaRandomProjectionExample.java
Yunni: Done

examples/src/main/scala/org/apache/spark/examples/ml/RandomProjectionExample.scala
Yunni: Done

Yunni:
@bravo-zhang @srowen I am OK to use the example in #15787.
But I still think `approxNearestNeighbor` and `approxSimilarityJoin` are different algorithms and it would be easier for user to understand in separate examples.

docs/ml-features.md
```diff
@@ -1396,3 +1396,149 @@ for more details on the API.
...
+d(\mathbf{A}, \mathbf{B}) = 1 - \frac{|\mathbf{A} \cap \mathbf{B}|}{|\mathbf{A} \cup \mathbf{B}|}
+\]`
+As its LSH family, MinHash applies a random perfect hash function `g` to each elements in the set and take the minimum of all hashed values:
+`\[
+h(\mathbf{A}) = \min_{a \in \mathbf{A}}(g(a))
+\]`
+
+Input sets for MinHash is represented in vectors which dimension equals the total number of elements in the space. Each dimension of the vectors represents the status of an elements: zero value means the elements is not in the set; non-zero value means the set contains the corresponding elements. For example, `Vectors.sparse(10, Array[(2, 1.0), (3, 1.0), (5, 1.0)])` means there are 10 elements in the space. This set contains elem 2, elem 3 and elem 5.
+
+**Note:** Empty sets cannot be transformed by MinHash, which means any input vector must have at least 1 non-zero indices.
```
MLnick: "... non-zero entry" perhaps

docs/ml-features.md
```diff
@@ -1396,3 +1396,149 @@ for more details on the API.
...
+</div>
+
+When multiple hash functions are picked, it's very useful for users to apply [amplification](https://en.wikipedia.org/wiki/Locality-sensitive_hashing#Amplification) to trade off between false positive and false negative rate.
+* AND-amplifications: Two input vectors are defined to be in the same bucket only if ALL of the hash values match. This will decrease the false positive rate but increase the false negative rate.
+* OR-amplifications: Two input vectors are defined to be in the same bucket as long as ANY one of the hash value matches. This will increase the false positive rate but decrease the false negative rate.
+
+Approximate nearest neighbor and approximate similarity join use OR-amplification.
+
+## Approximate Similarity Join
+Approximate similarity join takes two datasets, and approximately returns row pairs which distance is smaller than a user-defined threshold. Approximate Similarity Join supports both joining two different datasets and self joining.
```
MLnick: Row pairs of what? Does it return all columns or just the vector columns? 

I think we need to be specific about "distance between two input vectors is smaller".

docs/ml-features.md
```diff
@@ -1396,3 +1396,149 @@ for more details on the API.
...
+
+Approximate similarity join allows users to cache the transformed columns when necessary: If the `outputCol` is missing, the method will transform the data; if the `outputCol` exists, it will use the `outputCol` directly.
```
MLnick: I don't think it's totally clear what this means. Let's be more specific about the steps involved:

1. transform the input dataset(s) to create the hash signature in `LSH.outputCol`.
3. if an untransformed dataset is used as input, it will be transformed automatically

Because (1) is expensive, the transformed dataset can be cached if it will be re-used many times.

docs/ml-features.md
```diff
@@ -1396,3 +1396,149 @@ for more details on the API.
...
+
+A distance column will be added in the output dataset of approximate similarity join to show the distance between each output row pairs.
```
MLnick: Again, row pairs of what? Specify what is being compared here to be clear.

docs/ml-features.md
```diff
@@ -1396,3 +1396,149 @@ for more details on the API.
...
+</div>
+
+<div data-lang="java" markdown="1">
+
+{% include_example java/org/apache/spark/examples/ml/JavaApproxSimilarityJoinExample.java %}
+</div>
+</div>
+
+## Approximate Nearest Neighbor Search
+Approximate nearest neighbor search takes a dataset and a key, and approximately returns a number of rows in the dataset that are closest to the key. The number of rows to return are defined by user.
```
MLnick: Can simplify to "... returns a specified number of rows ..." and drop the last sentence.

Are we supporting arbitrary keys? I don't think so, so perhaps just call it "vector"?

docs/ml-features.md
```diff
@@ -1396,3 +1396,149 @@ for more details on the API.
...
+
+Approximate nearest neighbor search allows users to cache the transformed columns when necessary: If the `outputCol` is missing, the method will transform the data; if the `outputCol` exists, it will use the `outputCol` directly.
```
MLnick: Same comment as above for similarity join applies here.

docs/ml-features.md
```diff
@@ -1396,3 +1396,149 @@ for more details on the API.
...
+[MinHash](https://en.wikipedia.org/wiki/MinHash) is the LSH family in `spark.ml` for Jaccard distance where input features are sets of natural numbers. Jaccard distance of two sets is defined by the cardinality of their intersection and union:
+`\[
+d(\mathbf{A}, \mathbf{B}) = 1 - \frac{|\mathbf{A} \cap \mathbf{B}|}{|\mathbf{A} \cup \mathbf{B}|}
+\]`
+As its LSH family, MinHash applies a random perfect hash function `g` to each elements in the set and take the minimum of all hashed values:
+`\[
+h(\mathbf{A}) = \min_{a \in \mathbf{A}}(g(a))
+\]`
+
+Input sets for MinHash is represented in vectors which dimension equals the total number of elements in the space. Each dimension of the vectors represents the status of an elements: zero value means the elements is not in the set; non-zero value means the set contains the corresponding elements. For example, `Vectors.sparse(10, Array[(2, 1.0), (3, 1.0), (5, 1.0)])` means there are 10 elements in the space. This set contains elem 2, elem 3 and elem 5.
```
MLnick: Do we require (check) in `MinHash` that the input vectors are binary? Or do we just treat any non-zero value as 1 effectively? Maybe mention it whichever it is.

MLnick (changes requested):

jkbradley:
Linking [https://issues.apache.org/jira/browse/SPARK-18392] since it will alter the public API for LSH

docs/ml-features.md
```diff
@@ -1396,3 +1396,149 @@ for more details on the API.
 {% include_example python/ml/chisq_selector_example.py %}
 </div>
 </div>
+
+# Locality Sensitive Hashing
+[Locality Sensitive Hashing(LSH)](https://en.wikipedia.org/wiki/Locality-sensitive_hashing) Locality Sensitive Hashing(LSH) is an important class of hashing techniques, which is commonly used in clustering and outlier detection with large datasets.
```
Yunni: Done.

docs/ml-features.md
```diff
@@ -1396,3 +1396,149 @@ for more details on the API.
...
+d(p,q) > r2 \Rightarrow Pr(h(p)=h(q)) \leq p1
+\]`
+This LSH family is called `(r1, r2, p1, p2)`-sensitive.
+
+In this section, we call a pair of input features a false positive if the two features are hashed into the same hash bucket but they are far away in distance, and we define false negative as the pair of features when their distance are close but they are not in the same hash bucket.
+
+## Random Projection for Euclidean Distance
+**Note:** Please note that this is different than the [Random Projection for cosine distance](https://en.wikipedia.org/wiki/Locality-sensitive_hashing#Random_projection).
+
+[Random Projection](https://en.wikipedia.org/wiki/Locality-sensitive_hashing#Stable_distributions) is the LSH family in `spark.ml` for Euclidean distance. The Euclidean distance is defined as follows:
```
Yunni: My understanding is 2-stable distribution is how we choose the random vector, not the name of metric space or the hash function name.

See: https://en.wikipedia.org/wiki/Locality-sensitive_hashing#Stable_distributions

docs/ml-features.md
```diff
@@ -1396,3 +1396,149 @@ for more details on the API.
...
+`\[
+d(\mathbf{x}, \mathbf{y}) = \sqrt{\sum_i (x_i - y_i)^2}
+\]`
+Its LSH family projects features onto a random unit vector and divide the projected results to hash buckets:
+`\[
+h(\mathbf{x}) = \lfloor \frac{\mathbf{x} \cdot \mathbf{v}}{r} \rfloor
+\]`
+where `v` is a normalized random unit vector and `r` is user-defined bucket length. The bucket length can be used to control the average size of hash buckets. A larger bucket length means higher probability for features to be in the same bucket.
+
+The input features in Euclidean space are represented in vectors. Both sparse and dense vectors are supported.
```
Yunni: Done.

docs/ml-features.md
Yunni: Done.

docs/ml-features.md
Yunni: Add some description in L1501

docs/ml-features.md
Yunni: How about now?

docs/ml-features.md
Yunni: Added some description.

docs/ml-features.md
Yunni: Done.

docs/ml-features.md
```diff
@@ -1396,3 +1396,149 @@ for more details on the API.
 {% include_example python/ml/chisq_selector_example.py %}
 </div>
 </div>
+
+# Locality Sensitive Hashing
+[Locality Sensitive Hashing(LSH)](https://en.wikipedia.org/wiki/Locality-sensitive_hashing) Locality Sensitive Hashing(LSH) is an important class of hashing techniques, which is commonly used in clustering and outlier detection with large datasets. 
+
+The general idea of LSH is to use a family of functions (we call them LSH families) to hash data points into buckets, so that the data points which are close to each other are in the same buckets with high probability, while data points that are far away from each other are very likely in different buckets. A formal definition of LSH family is as follows:
+
+In a metric space `(M, d)`, an LSH family is a family of functions `h` that satisfy the following properties:
```
Yunni: Done.

docs/ml-features.md
```diff
@@ -1396,3 +1396,149 @@ for more details on the API.
...
+`\[
+\forall p, q \in M,\\
+d(p,q) < r1 \Rightarrow Pr(h(p)=h(q)) \geq p1\\
+d(p,q) > r2 \Rightarrow Pr(h(p)=h(q)) \leq p1
```
Yunni: Done.

docs/ml-features.md
```diff
@@ -1396,3 +1396,149 @@ for more details on the API.
...
+{% include_example scala/org/apache/spark/examples/ml/LSHTransformationExample.scala %}
+</div>
+
+<div data-lang="java" markdown="1">
+
+{% include_example java/org/apache/spark/examples/ml/JavaLSHTransformationExample.java %}
+</div>
+</div>
+
+When multiple hash functions are picked, it's very useful for users to apply [amplification](https://en.wikipedia.org/wiki/Locality-sensitive_hashing#Amplification) to trade off between false positive and false negative rate.
```
Yunni: Sorry, I did not get it?

Yunni: This section is removed since it will be fully implemented in https://issues.apache.org/jira/browse/SPARK-18450

sethah:
Is this still targeted for 2.1?

Yunni:
@sethah I think so. I have made changes for the docs but I haven't made changes to the examples. Please take a look when you get a chance.

docs/ml-features.md
```diff
@@ -1478,3 +1478,139 @@ for more details on the API.
...
+h(\mathbf{x}) = \lfloor \frac{\mathbf{x} \cdot \mathbf{v}}{r} \rfloor
+\]`
+where `v` is a normalized random unit vector and `r` is user-defined bucket length. The bucket length can be used to control the average size of hash buckets. A larger bucket length means higher probability for features to be in the same bucket.
+
+Bucketed Random Projection accepts arbitrary vectors as input features, and supports both sparse and dense vectors.
+
+<div class="codetabs">
+<div data-lang="scala" markdown="1">
+
+Refer to the [RandomProjection Scala docs](api/scala/index.html#org.apache.spark.ml.feature.RandomProjection)
```
MLnick: This Scaladoc link should be for `BucketedRandomProjection` now

docs/ml-features.md
```diff
@@ -1478,3 +1478,139 @@ for more details on the API.
...
+for more details on the API.
+
+{% include_example scala/org/apache/spark/examples/ml/BucketedRandomProjectionLSHExample.scala %}
+</div>
+
+<div data-lang="java" markdown="1">
+
+Refer to the [RandomProjection Java docs](api/java/org/apache/spark/ml/feature/RandomProjection.html)
```
MLnick: Same here

docs/ml-features.md
```diff
@@ -1478,3 +1478,139 @@ for more details on the API.
...
+\]`
+
+The input sets for MinHash are represented as binary vectors, where the vector indices represent the elements themselves and the non-zero values in the vector represent the presence of that element in the set. While both dense and sparse vectors are supported, typically sparse vectors are recommended for efficiency. For example, `Vectors.sparse(10, Array[(2, 1.0), (3, 1.0), (5, 1.0)])` means there are 10 elements in the space. This set contains elem 2, elem 3 and elem 5. All non-zero values are treated as binary "1" values.
+
+**Note:** Empty sets cannot be transformed by MinHash, which means any input vector must have at least 1 non-zero entry.
+
+<div class="codetabs">
+<div data-lang="scala" markdown="1">
+
+Refer to the [MinHash Scala docs](api/scala/index.html#org.apache.spark.ml.feature.MinHash)
```
MLnick: Should be updated to `MinHashLSH`?

docs/ml-features.md
```diff
@@ -1478,3 +1478,139 @@ for more details on the API.
...
+for more details on the API.
+
+{% include_example java/org/apache/spark/examples/ml/JavaMinHashLSHExample.java %}
+</div>
+</div>
+
+## Feature Transformation
+Feature Transformation is the base functionality to add hash results as a new column. Users can specify input column name and output column name by setting `inputCol` and `outputCol`. LSH in `spark.ml` also supports multiple LSH hash tables. Users can specify the number of hash tables by setting `numHashTables`.
+
+The output type of feature type is `Array[Vector]` where the dimension of the array equals `numHashTables`, and the dimensions of the vectors are currently set to 1.
```
MLnick: The type of `outputCol` is ...

examples/src/main/scala/org/apache/spark/examples/ml/ApproxSimilarityJoinExample.scala
```diff
@@ -0,0 +1,67 @@
...
+
+    val mh = new MinHashLSH()
+      .setNumHashTables(5)
+      .setInputCol("keys")
+      .setOutputCol("values")
+
+    val model = mh.fit(dfA)
+    model.approxSimilarityJoin(dfA, dfB, 0.6).show()
+
+    // Cache the transformed columns
```
MLnick: This mentions caching but doesn't cache.

examples/src/main/scala/org/apache/spark/examples/ml/ApproxSimilarityJoinExample.scala
```diff
@@ -0,0 +1,67 @@
...
+    val transformedA = model.transform(dfA)
+    val transformedB = model.transform(dfB)
+    model.approxSimilarityJoin(transformedA, transformedB, 0.6).show()
+
+    // Self Join
+    model.approxSimilarityJoin(dfA, dfA, 0.6).filter("datasetA.id < datasetB.id").show()
```
MLnick: Just a note - will `approxSimilarityJoin` return duplicates? We should think about removing them automatically then?

examples/src/main/scala/org/apache/spark/examples/ml/MinHashLSHExample.scala
```diff
@@ -0,0 +1,54 @@
...
+// scalastyle:off println
+package org.apache.spark.examples.ml
+
+// $example on$
+import org.apache.spark.ml.feature.MinHashLSH
+import org.apache.spark.ml.linalg.Vectors
+// $example off$
+import org.apache.spark.sql.SparkSession
+
+object MinHashLSHExample {
```
MLnick: This and the min hash transformation example are almost the same. Perhaps we can remove the transformation example, and adjust the user guide section to refer (and link) to the code examples for `MinHashLSH` and `BucketedRandomProjectionLSH`

docs/ml-features.md
```diff
@@ -1478,3 +1478,139 @@ for more details on the API.
...
+<div data-lang="java" markdown="1">
+
+Refer to the [MinHash Java docs](api/java/org/apache/spark/ml/feature/MinHash.html)
+for more details on the API.
+
+{% include_example java/org/apache/spark/examples/ml/JavaMinHashLSHExample.java %}
+</div>
+</div>
+
+## Feature Transformation
```
MLnick: We don't really mention anything here about the amplification approach taken. Perhaps we can mention that we hash into multiple hash tables, which are used for OR-amplification in similarity join and ANN. We can mention what the impact is on accuracy / performance of `numHashTables`.

We can also mention that AND-amplification will be added in future.

MLnick: It would also be good to mention that the transformed dataset can be cached, since `transform` can be expensive. We can either mention it here, or perhaps mention it (twice) in the join and ANN sections below.

docs/ml-features.md
```diff
@@ -1478,3 +1478,139 @@ for more details on the API.
...
+
+<div data-lang="java" markdown="1">
+
+{% include_example java/org/apache/spark/examples/ml/JavaLSHTransformationExample.java %}
+</div>
+</div>
+
+## Approximate Similarity Join
+Approximate similarity join takes two datasets, and approximately returns pairs of rows in the origin datasets which distance is smaller than a user-defined threshold. Approximate Similarity Join supports both joining two different datasets and self joining.
+
```
MLnick: Will self join produce duplicates? If so we should note that.

docs/ml-features.md
```diff
@@ -1478,3 +1478,139 @@ for more details on the API.
...
+<div data-lang="java" markdown="1">
+
+{% include_example java/org/apache/spark/examples/ml/JavaApproxSimilarityJoinExample.java %}
+</div>
+</div>
+
+## Approximate Nearest Neighbor Search
+Approximate nearest neighbor search takes a dataset and a vector, and approximately returns a specified number of rows in the dataset that are closest to the vector.
+
+Approximate nearest neighbor accepts both transformed and untransformed datasets as input. If an untransformed dataset is used, it will be transformed automatically. In this case, the hash signature will be created as outputCol.
```
MLnick: backticks around "outputCol"

docs/ml-features.md
```diff
@@ -1478,3 +1478,139 @@ for more details on the API.
...
+<div data-lang="java" markdown="1">
+
+{% include_example java/org/apache/spark/examples/ml/JavaLSHTransformationExample.java %}
+</div>
+</div>
+
+## Approximate Similarity Join
+Approximate similarity join takes two datasets, and approximately returns pairs of rows in the origin datasets which distance is smaller than a user-defined threshold. Approximate Similarity Join supports both joining two different datasets and self joining.
+
+Approximate similarity join accepts both transformed and untransformed datasets as input. If an untransformed dataset is used, it will be transformed automatically. In this case, the hash signature will be created as outputCol.
```
MLnick: backticks around "outputCol"

docs/ml-features.md
```diff
@@ -1478,3 +1478,139 @@ for more details on the API.
...
+
+In the joined dataset, the origin datasets can be queried in `datasetA` and `datasetB`. A distance column will be added in the output dataset of approximate similarity join to show the distance between each output pairs of rows in the origin datasets .
```
MLnick: nit - space at end before `.`

MLnick (changes requested):
Made a pass - I think we can consolidate at least one of the examples.

Also think we need a little more detail in places.

jkbradley:
Could you please add tags "[ML][DOCS]" to the PR title?

jkbradley:
+1 for consolidating the examples.  The boilerplate of creating a dataset and setting algorithm parameters takes up most of the example.  I would create 1 example per algorithm which does transform, approxNearestNeighbor, and approxSimilarityJoin.  The only reason not to would be if those demos required different datasets or algorithm settings, but I suspect they could be done in a unified manner.

MLnick:
ok to test

docs/ml-features.md
Yunni: Done.

docs/ml-features.md
Yunni: Done.

docs/ml-features.md
Yunni: Both Done.

Yunni: This doc is in L1509 and L1516

docs/ml-features.md
Yunni: Done.

docs/ml-features.md
Yunni: Done.

docs/ml-features.md
Yunni: Done.

docs/ml-features.md
Yunni: Done.

examples/src/main/scala/org/apache/spark/examples/ml/ApproxSimilarityJoinExample.scala
Yunni: Fixed.

examples/src/main/scala/org/apache/spark/examples/ml/ApproxSimilarityJoinExample.scala
Yunni: Duplicates are only in self join, where both (RowA, RowB) and (RowB, RowA) will appear in the results.

Yunni:
@MLnick @jkbradley I have changed the examples to be 1 example per algorithm which does transform, approxNearestNeighbor, and approxSimilarityJoin. PTAL.

jkbradley:
I can take a look

jkbradley:
I found myself wanting to make a number of tiny comments, so I thought it'd be easier to send a PR.  Could you please take a look at this one?
Thanks!

jkbradley:
LGTM
merging with master and branch-2.1
Thanks all!

PR #15797, opened 2016-11-07
[SPARK-17990][SPARK-18302][SQL] correct several partition related behaviours of ExternalCatalog

cloud-fan:

This PR corrects several partition related behaviors of `ExternalCatalog`:

1. default partition location should not always lower case the partition column names in path string(fix `HiveExternalCatalog`)
2. rename partition should not always lower case the partition column names in updated partition path string(fix `HiveExternalCatalog`)
3. rename partition should update the partition location only for managed table(fix `InMemoryCatalog`)
4. create partition with existing directory should be fine(fix `InMemoryCatalog`)
5. create partition with non-existing directory should create that directory(fix `InMemoryCatalog`)
6. drop partition from external table should not delete the directory(fix `InMemoryCatalog`)

## How was this patch tested?

new tests in `ExternalCatalogSuite`

### DIFF
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/ExternalCatalogUtils.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/ExternalCatalogUtils.scala
@@ -1,0 +1,92 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.spark.sql.catalyst.catalog
+
+import org.apache.hadoop.fs.Path
+import org.apache.hadoop.util.Shell
+
+import org.apache.spark.sql.catalyst.catalog.CatalogTypes.TablePartitionSpec
+
+object ExternalCatalogUtils {
+  // This duplicates default value of Hive `ConfVars.DEFAULTPARTITIONNAME`, since catalyst doesn't
+  // depend on Hive.
+  val DEFAULT_PARTITION_NAME = "__HIVE_DEFAULT_PARTITION__"
+
+  //////////////////////////////////////////////////////////////////////////////////////////////////
+  // The following string escaping code is mainly copied from Hive (o.a.h.h.common.FileUtils).
+  //////////////////////////////////////////////////////////////////////////////////////////////////
+
+  val charToEscape = {
+    val bitSet = new java.util.BitSet(128)
+
+    /**
+     * ASCII 01-1F are HTTP control characters that need to be escaped.
+     * \u000A and \u000D are \n and \r, respectively.
+     */
+    val clist = Array(
+      '\u0001', '\u0002', '\u0003', '\u0004', '\u0005', '\u0006', '\u0007', '\u0008', '\u0009',
+      '\n', '\u000B', '\u000C', '\r', '\u000E', '\u000F', '\u0010', '\u0011', '\u0012', '\u0013',
+      '\u0014', '\u0015', '\u0016', '\u0017', '\u0018', '\u0019', '\u001A', '\u001B', '\u001C',
+      '\u001D', '\u001E', '\u001F', '"', '#', '%', '\'', '*', '/', ':', '=', '?', '\\', '\u007F',
+      '{', '[', ']', '^')
+
+    clist.foreach(bitSet.set(_))
+
+    if (Shell.WINDOWS) {
+      Array(' ', '<', '>', '|').foreach(bitSet.set(_))
+    }
+
+    bitSet
+  }
+
+  def needsEscaping(c: Char): Boolean = {
+    c >= 0 && c < charToEscape.size() && charToEscape.get(c)
+  }
+
+  def escapePathName(path: String): String = {
+    val builder = new StringBuilder()
+    path.foreach { c =>
+      if (needsEscaping(c)) {
+        builder.append('%')
+        builder.append(f"${c.asInstanceOf[Int]}%02X")
+      } else {
+        builder.append(c)
+      }
+    }
+
+    builder.toString()
+  }
+
+  def generatePartitionPath(
+      spec: TablePartitionSpec,
+      partitionColumnNames: Seq[String],
+      tablePath: Path): Path = {
+    val partitionPathStrings = partitionColumnNames.map { col =>
+      val partitionValue = spec(col)
+      val partitionString = if (partitionValue == null) {
+        DEFAULT_PARTITION_NAME
+      } else {
+        escapePathName(partitionValue)
+      }
+      escapePathName(col) + "=" + partitionString
+    }
+    partitionPathStrings.foldLeft(tablePath) { (totalPath, nextPartPath) =>
+      new Path(totalPath, nextPartPath)
+    }
+  }
+}
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/ExternalCatalogUtils.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/ExternalCatalogUtils.scala
@@ -72,6 +72,35 @@ object ExternalCatalogUtils {
     builder.toString()
   }
 
+
+  def unescapePathName(path: String): String = {
+    val sb = new StringBuilder
+    var i = 0
+
+    while (i < path.length) {
+      val c = path.charAt(i)
+      if (c == '%' && i + 2 < path.length) {
+        val code: Int = try {
+          Integer.parseInt(path.substring(i + 1, i + 3), 16)
+        } catch {
+          case _: Exception => -1
+        }
+        if (code >= 0) {
+          sb.append(code.asInstanceOf[Char])
+          i += 3
+        } else {
+          sb.append(c)
+          i += 1
+        }
+      } else {
+        sb.append(c)
+        i += 1
+      }
+    }
+
+    sb.toString()
+  }
+
   def generatePartitionPath(
       spec: TablePartitionSpec,
       partitionColumnNames: Seq[String],
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/InMemoryCatalog.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/InMemoryCatalog.scala
@@ -231,7 +231,7 @@ class InMemoryCatalog(
         assert(tableMeta.storage.locationUri.isDefined,
           "Managed table should always have table location, as we will assign a default location " +
             "to it if it doesn't have one.")
-        val dir = new Path(tableMeta.storage.locationUri.get)
+        val dir = new Path(tableMeta.location)
         try {
           val fs = dir.getFileSystem(hadoopConfig)
           fs.delete(dir, true)
@@ -259,7 +259,7 @@ class InMemoryCatalog(
       assert(oldDesc.table.storage.locationUri.isDefined,
         "Managed table should always have table location, as we will assign a default location " +
           "to it if it doesn't have one.")
-      val oldDir = new Path(oldDesc.table.storage.locationUri.get)
+      val oldDir = new Path(oldDesc.table.location)
       val newDir = new Path(catalog(db).db.locationUri, newName)
       try {
         val fs = oldDir.getFileSystem(hadoopConfig)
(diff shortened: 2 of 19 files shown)


sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveExternalCatalog.scala
```diff
@@ -226,106 +208,139 @@ private[spark] class HiveExternalCatalog(conf: SparkConf, hadoopConf: Configurat
...
-        val location = if (tableDefinition.tableType == EXTERNAL) {
-          // When we hit this branch, we are saving an external data source table with hive
-          // compatible format, which means the data source is file-based and must have a `path`.
-          require(tableDefinition.storage.locationUri.isDefined,
-            "External file-based data source table must have a `path` entry in storage properties.")
-          Some(new Path(tableDefinition.storage.locationUri.get).toUri.toString)
-        } else {
-          None
-        }
+  private def createDataSourceTable(
```
cloud-fan: This method just copies the code in the above if branch: https://github.com/apache/spark/pull/15797/files#diff-159191585e10542f013cb3a714f26075R199

cloud-fan:
cc @mallman this should fix https://issues.apache.org/jira/browse/SPARK-17990
cc @yhuai @gatorsmile

mallman:
@cloud-fan Thanks for taking this on. I will test out a build from your branch to verify that it fixes the problem I was seeing.

mallman:
I tried your patch and it works for me.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/InMemoryCatalog.scala
```diff
@@ -423,33 +424,35 @@ class InMemoryCatalog(
...
-        }.mkString("/")
-        val newPath = partitionColumnNames.flatMap { col =>
-          newSpec.get(col).map(col + "=" + _)
-        }.mkString("/")
+      val oldPartition = getPartition(db, table, oldSpec)
+      val newPartition = if (shouldUpdatePartitionLocation) {
+        val oldPartPath = new Path(oldPartition.storage.locationUri.get)
+        val newPartPath = new Path(
+          tableLocation,
+          partitionColumnNames.map(col => col + "=" + newSpec(col)).mkString("/"))
```
ericl: Should we pull this into PartitioningUtils?

sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveExternalCatalog.scala
```diff
@@ -810,13 +840,37 @@ private[spark] class HiveExternalCatalog(conf: SparkConf, hadoopConf: Configurat
...
       db, table, specs.map(lowerCasePartitionSpec), newSpecs.map(lowerCasePartitionSpec))
+
+    val tableMeta = getTable(db, table)
+    val partitionColumnNames = tableMeta.partitionColumnNames
+    // Hive metastore is not case preserving and keeps partition columns with lower cased names.
+    // When Hive rename partition for managed tables, it will update the partition location with
+    // a default value generate by the new spec and lower cased partition column names. This is
+    // unexpected and we need to call `alterPartitions` to fix the partition location.
+    val hasUpperCasePartitionColumn = partitionColumnNames.exists(col => col.toLowerCase != col)
+    if (tableMeta.tableType == MANAGED && hasUpperCasePartitionColumn) {
```
ericl: Would this apply to external tables as well?

sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveExternalCatalog.scala
```diff
@@ -810,13 +840,37 @@ private[spark] class HiveExternalCatalog(conf: SparkConf, hadoopConf: Configurat
...
+      val tableLocation = tableMeta.storage.locationUri.get
+      val newParts = newSpecs.map { spec =>
+        val partition = client.getPartition(db, table, lowerCasePartitionSpec(spec))
+        val partitionPath = new Path(
+          tableLocation,
+          partitionColumnNames.map(col => col + "=" + spec(col)).mkString("/"))
```
ericl: What if it had a custom location?

ericl (changes requested):

sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveExternalCatalog.scala
cloud-fan: no, Hive only update the partition location when renaming partitions for managed table.

sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveExternalCatalog.scala
cloud-fan: RENAME PARTITION can only change the partition spec, not partition path. The SQL syntax is :`ALTER TABLE table_name PARTITION partition_spec RENAME TO PARTITION partition_spec;`, users can't specify partition location

cloud-fan: if the partition to be renamed has custom location, we will move the directory too.

Actually there is no such a concept called `custom location` for partitions. Partitions always have a location, if users didn't specify, we will generate a default one. The generated one and specified one are treated equally.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/ExternalCatalogUtils.scala
```diff
@@ -0,0 +1,92 @@
...
+ */
+
+package org.apache.spark.sql.catalyst.catalog
+
+import org.apache.hadoop.fs.Path
+import org.apache.hadoop.util.Shell
+
+import org.apache.spark.sql.catalyst.catalog.CatalogTypes.TablePartitionSpec
+
+object ExternalCatalogUtils {
```
cloud-fan: @ericl `PartitioningUtils` is not accessible in catalyst module, so I created this.

cloud-fan:
retest this please

gatorsmile:
retest this please

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/PartitioningUtils.scala
```diff
@@ -451,51 +444,6 @@ object PartitioningUtils {
...
-        builder.append(f"${c.asInstanceOf[Int]}%02X")
-      } else {
-        builder.append(c)
-      }
-    }
-
-    builder.toString()
-  }
-
   def unescapePathName(path: String): String = {
```
gatorsmile: If `escapePathName` is moved, I think we also need to move `unescapePathName` to the same file?

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/catalog/ExternalCatalogSuite.scala
```diff
@@ -591,25 +673,25 @@ abstract class ExternalCatalogSuite extends SparkFunSuite with BeforeAndAfterEac
     val catalog = newBasicCatalog()
     val db = catalog.getDatabase("db1")
     val table = CatalogTable(
-      identifier = TableIdentifier("my_table", Some("db1")),
```
gatorsmile: Could we keep the original names unchanged? Now, the table/database names allow `alphanumeric characters and underscores`. Thus, I think they are named using underscore on purpose?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/interface.scala
```diff
@@ -171,6 +171,10 @@ case class CatalogTable(
     throw new AnalysisException(s"table $identifier did not specify database")
   }
 
+  def location: String = storage.locationUri.getOrElse {
```
gatorsmile: Like the other functions in this file, add the function descriptions too?

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/catalog/ExternalCatalogSuite.scala
cloud-fan: Actually this test was written by me, and I picked `myTable` at the first time, but changed to `my_table` because of this bug.

sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveExternalCatalog.scala
```diff
@@ -810,13 +825,44 @@ private[spark] class HiveExternalCatalog(conf: SparkConf, hadoopConf: Configurat
...
+    // When Hive rename partition for managed tables, it will create the partition location with
+    // a default path generate by the new spec with lower cased partition column names. This is
+    // unexpected and we need to rename them manually and alter the partition location.
+    val hasUpperCasePartitionColumn = partitionColumnNames.exists(col => col.toLowerCase != col)
+    if (tableMeta.tableType == MANAGED && hasUpperCasePartitionColumn) {
+      val tablePath = new Path(tableMeta.location)
+      val fs = tablePath.getFileSystem(hadoopConf)
+      val newParts = newSpecs.map { spec =>
+        val partition = client.getPartition(db, table, lowerCasePartitionSpec(spec))
+        val wrongPath = new Path(partition.storage.locationUri.get)
```
gatorsmile: How about creating a function `location` for partitions too, just like [what this PR did for table's location](https://github.com/cloud-fan/spark/blob/d939b37e6d032ef61f6d0d383f18049ffed7c78a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/interface.scala#L174-L177)?

sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveExternalCatalog.scala
```diff
@@ -810,13 +825,44 @@ private[spark] class HiveExternalCatalog(conf: SparkConf, hadoopConf: Configurat
...
+        val rightPath = ExternalCatalogUtils.generatePartitionPath(
+          spec, partitionColumnNames, tablePath)
+        try {
+          fs.rename(wrongPath, rightPath)
+        } catch {
+          case e: IOException =>
+            throw new SparkException(s"Unable to rename partition path $wrongPath", e)
```
gatorsmile: Maybe we also need to add `rightPath` here

ericl:
I just tested this PR with https://github.com/apache/spark/pull/15814 modified with mixed case partition names in unit tests, and it fixes the issues there too. I'll leave it to @gatorsmile to lgtm since he has a few more comments.

gatorsmile:
Will carefully review it tonight. Thanks!

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/InMemoryCatalog.scala
```diff
@@ -231,7 +231,7 @@ class InMemoryCatalog(
         assert(tableMeta.storage.locationUri.isDefined,
           "Managed table should always have table location, as we will assign a default location " +
             "to it if it doesn't have one.")
-        val dir = new Path(tableMeta.storage.locationUri.get)
+        val dir = new Path(tableMeta.location)
         try {
           val fs = dir.getFileSystem(hadoopConfig)
           fs.delete(dir, true)
```
gatorsmile: Since the deletion is recursive and thus it can delete the lcation/directory of a partitioned table. I checked the implementation of `delete`. It sounds like it does not guaranttee the atomicity. The deletion could be partial. We might need to also put a [TODO](https://github.com/cloud-fan/spark/blob/56f3073162f0b53e4a3de686bba31c8eb6048123/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/InMemoryCatalog.scala#L399-L400) here, like what we did for `dropPartition`

gatorsmile:
In [`renamePartition` of `HiveExternalCatalog`](https://github.com/cloud-fan/spark/blob/56f3073162f0b53e4a3de686bba31c8eb6048123/sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveExternalCatalog.scala#L829-L853), we update the file names to preserve the case after renaming partitions. How about another command `loadPartition`?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/InMemoryCatalog.scala
cloud-fan: it's minor and not related to this PR, I'd like to not touch it.

cloud-fan:
> In renamePartition of HiveExternalCatalog, we update the file names to preserve the case after renaming partitions. How about another command loadPartition? More generally, if we insert new data into a partitioned table, Hive will create a new directory for non-existent partitions. These newly created file path are not case-preserving, right?

This PR tries its best to fix some partition related behaviours, I think it's OK to leave some hard works for follow-ups.

gatorsmile:
LGTM

rxin:
Merging in master/branch-2.1.

sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveExternalCatalog.scala
```diff
@@ -810,13 +825,43 @@ private[spark] class HiveExternalCatalog(conf: SparkConf, hadoopConf: Configurat
...
+    val hasUpperCasePartitionColumn = partitionColumnNames.exists(col => col.toLowerCase != col)
+    if (tableMeta.tableType == MANAGED && hasUpperCasePartitionColumn) {
+      val tablePath = new Path(tableMeta.location)
+      val newParts = newSpecs.map { spec =>
+        val partition = client.getPartition(db, table, lowerCasePartitionSpec(spec))
+        val wrongPath = new Path(partition.location)
+        val rightPath = ExternalCatalogUtils.generatePartitionPath(
+          spec, partitionColumnNames, tablePath)
+        try {
+          tablePath.getFileSystem(hadoopConf).rename(wrongPath, rightPath)
```
gatorsmile: Found an issue here... When we call `rename`, not all the file systems have the same behaviors. For example, on mac OS, when we doing this `.../tbl/partcol1=5/partcol2=6` -> `.../tbl/partCol1=5/partCol2=6` . The result is `.../tbl/partcol1=5/partCol2=6`. However, the file system used in Jenkin does not have such an issue.

cloud-fan: good catch! let me think about how to fix it.

PR #15798, opened 2016-11-07
[SPARK-18262][BUILD][SQL] JSON.org license is now CatX

srowen:

Try excluding org.json:json from hive-exec dep as it's Cat X now. It may be the case that it's not used by the part of Hive Spark uses anyway.

## How was this patch tested?

Existing tests

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -421,9 +421,6 @@ Copyright (c) 2011, Terrence Parr.
 This product includes/uses ASM (http://asm.ow2.org/),
 Copyright (c) 2000-2007 INRIA, France Telecom.
 
-This product includes/uses org.json (http://www.json.org/java/index.html),
-Copyright (c) 2002 JSON.org
-
 This product includes/uses JLine (http://jline.sourceforge.net/),
 Copyright (c) 2002-2006, Marc Prud'hommeaux <mwp1@cornell.edu>.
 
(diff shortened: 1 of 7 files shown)


busbey:
apologies if this is just a test run. If review comments are welcome currently, you should also update the top level NOTICE file to remove the blurb for org.json:

```
This product includes/uses org.json (http://www.json.org/java/index.html),
Copyright (c) 2002 JSON.org
```

srowen:
Yeah just a test run right now (`[WIP]`) to see if it works at all; see a little more background on the JIRA. Indeed I'd clean out all of those refs before committing something like this.

srowen:
Hm, so indeed Jenkins tests find this is fine to exclude. I ran into local test failures that looked unrelated, even if they were persistent, so could be noise.

My next step will be to make all the other license changes to make this a possibly mergeable change, just in case we do want to merge.

rxin:
SGTM. Thanks for doing this.

srowen:
I removed the entry in `NOTICE`. Interestingly we're supposed to preserve the NOTICE file contents of projects that are redistributed and this comes from Hive. However I suspect it is not required to retain portions of NOTICE that will not apply to Spark's distribution.

busbey:
That interpretation lines up with my understanding; we only need to recreate the portions of NOTICE that are applicable.

rxin:
Is this still WIP?

srowen:
I think it's mergeable. I'm slightly suspicious it was just this easy, but tests have passed.

rxin:
OK let's see. Merging in master/branch-2.1.

PR #15801, opened 2016-11-07
[SPARK-18337] Complete mode memory sinks should be able to recover from checkpoints

brkyvz:

It would be nice if memory sinks can also recover from checkpoints. For correctness reasons, the only time we should support it is in `Complete` OutputMode. We can support this in CompleteMode, because the output of the StateStore is already persisted in the checkpoint directory.

## How was this patch tested?

Unit test

### DIFF
--- a/sql/core/src/test/scala/org/apache/spark/sql/streaming/test/DataStreamReaderWriterSuite.scala
+++ b/sql/core/src/test/scala/org/apache/spark/sql/streaming/test/DataStreamReaderWriterSuite.scala
@@ -477,35 +477,49 @@ class DataStreamReaderWriterSuite extends StreamTest with BeforeAndAfter {
     val checkpointDir = new File(checkpointLoc, "offsets")
     checkpointDir.mkdirs()
     assert(checkpointDir.exists())
-    def startStream: StreamingQuery = {
+    val tableName = "test"
+    def startQuery: StreamingQuery = {
       df.groupBy("a")
         .count()
         .writeStream
         .format("memory")
-        .queryName("test")
+        .queryName(tableName)
         .option("checkpointLocation", checkpointLoc)
         .outputMode("complete")
         .start()
     }
     // no exception here
-    val q = startStream
+    val q = startQuery
     ms.addData(0, 1)
     q.processAllAvailable()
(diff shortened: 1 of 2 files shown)


sql/core/src/test/scala/org/apache/spark/sql/streaming/test/DataStreamReaderWriterSuite.scala
```diff
@@ -467,4 +468,37 @@ class DataStreamReaderWriterSuite extends StreamTest with BeforeAndAfter {
...
+    val q = df
+      .groupBy("a")
+      .count()
+      .writeStream
+      .format("memory")
+      .queryName("test")
+      .option("checkpointLocation", checkpointLoc)
+      .outputMode("complete")
+      .start()
+
```
tdas: This test should also check that the memory sink data is recovered to be the same as before failure.

brkyvz:
@tdas Added test for the data as well

sql/core/src/test/scala/org/apache/spark/sql/streaming/test/DataStreamReaderWriterSuite.scala
```diff
@@ -467,4 +468,54 @@ class DataStreamReaderWriterSuite extends StreamTest with BeforeAndAfter {
...
+      spark.table("test"),
+      Seq(Row(0, 1), Row(1, 1))
+    )
+
+    val q2 = startStream
+    q2.processAllAvailable()
+    checkAnswer(
+      spark.table("test"),
+      Seq(Row(0, 1), Row(1, 1))
+    )
```
tdas: its not clear if this table was regenerated by the query, or just the same table that existed before when the query was stopped earlier. so it would be good it if before restarting you remove the table, or rewrite the table to something different (and verify that), so that we are sure that the table is actually being regenerated.

brkyvz:
@tdas  Addressed your comments. Test time increased to 2.5 seconds though, fyi.

tdas:
LGTM. Merging this to master and 2.1

tdas:
and 2.0 as well. seems like small enough fix.

PR #15802, opened 2016-11-07
[SPARK-18338][SQL][test-maven] Fix test case initialization order under Maven builds

liancheng:

Test case initialization order under Maven and SBT are different. Maven always creates instances of all test cases and then run them all together.

This fails `ObjectHashAggregateSuite` because the randomized test cases there register a temporary Hive function right before creating a test case, and can be cleared while initializing other successive test cases. In SBT, this is fine since the created test case is executed immediately after creating the temporary function.

To fix this issue, we should put initialization/destruction code into `beforeAll()` and `afterAll()`.

## How was this patch tested?

Existing tests.

### DIFF
--- a/sql/hive/src/test/scala/org/apache/spark/sql/hive/execution/ObjectHashAggregateSuite.scala
+++ b/sql/hive/src/test/scala/org/apache/spark/sql/hive/execution/ObjectHashAggregateSuite.scala
@@ -25,11 +25,10 @@ import org.scalatest.Matchers._
 import org.apache.spark.sql._
 import org.apache.spark.sql.catalyst.FunctionIdentifier
 import org.apache.spark.sql.catalyst.analysis.UnresolvedFunction
-import org.apache.spark.sql.catalyst.expressions.{ExpressionEvalHelper, ExpressionInfo, Literal}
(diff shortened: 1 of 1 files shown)


yhuai:
Seems it does not work with sbt?

zsxwing:
Could you add `[test-maven]` to the title to run the PR with maven build?

zsxwing:
> Seems it does not work with sbt?

I saw the same `java.lang.OutOfMemoryError: Java heap space` in my PR as well. I think it's just because these tests are flaky.

liancheng:
@cloud-fan already reported the OOM issue. I'm trying to reproduce it locally.

Added the `[test-maven]` tag to trigger Maven tests.

liancheng:
retest this please

liancheng:
retest this please.

liancheng:
The last build failure was caused by an irrelevant flaky test.

BTW, I've reproduced the OOM issue locally by running `ObjectHashAggregateSuite` 200 times within a single SBT REPL session. Analyzing the heap dump.

yhuai:
I am going to merge this to fix maven build.

PR #15804, opened 2016-11-08
[SPARK-18342] Make rename failures fatal in HDFSBackedStateStore

brkyvz:

If the rename operation in the state store fails (`fs.rename` returns `false`), the StateStore should throw an exception and have the task retry. Currently if renames fail, nothing happens during execution immediately. However, you will observe that snapshot operations will fail, and then any attempt at recovery (executor failure / checkpoint recovery) also fails.

## How was this patch tested?

Unit test

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/state/HDFSBackedStateStoreProvider.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/state/HDFSBackedStateStoreProvider.scala
@@ -254,7 +254,9 @@ private[state] class HDFSBackedStateStoreProvider(
   private def commitUpdates(newVersion: Long, map: MapType, tempDeltaFile: Path): Path = {
     synchronized {
       val finalDeltaFile = deltaFile(newVersion)
-      fs.rename(tempDeltaFile, finalDeltaFile)
+      if (!fs.rename(tempDeltaFile, finalDeltaFile)) {
+        throw new IOException(s"Failed to rename $tempDeltaFile to $finalDeltaFile")
+      }
       loadedMaps.put(newVersion, map)
       finalDeltaFile
     }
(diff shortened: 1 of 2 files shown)


brkyvz:
cc @tdas

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/state/HDFSBackedStateStoreProvider.scala
```diff
@@ -525,7 +527,7 @@ private[state] class HDFSBackedStateStoreProvider(
 
         val deltaFiles = allFiles.filter { file =>
           file.version > snapshotFile.version && file.version <= version
-        }
+        }.toList
```
brkyvz: the line returns a `Stream` which is a Scala Iterator like class, therefore the logging below doesn't work

sql/core/src/test/scala/org/apache/spark/sql/execution/streaming/state/StateStoreSuite.scala
```diff
@@ -598,3 +612,19 @@ private[state] object StateStoreSuite {
     }}.toSet
   }
 }
+
+/** Fake FileSystem to test whether the method `fs.exists` is called during
+ * `DataSource.resolveRelation`.
+ */
```
tdas: Wrong docs.

sql/core/src/test/scala/org/apache/spark/sql/execution/streaming/state/StateStoreSuite.scala
```diff
@@ -455,6 +456,18 @@ class StateStoreSuite extends SparkFunSuite with BeforeAndAfter with PrivateMeth
     }
   }
 
+  test("commit fails when rename fails") {
```
tdas: can you add the jira number here.

tdas:
looks good. to minor comments.

brkyvz:
@tdas Addressed

tdas:
LGTM.

tdas:
Merge this to master and 2.0

PR #15805, opened 2016-11-08
[SPARK-18346][SQL] TRUNCATE TABLE should fail if no partition is matched for the given non-partial partition spec

cloud-fan:

a follow up of https://github.com/apache/spark/pull/15688

## How was this patch tested?

updated test in `DDLSuite`

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/command/tables.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/command/tables.scala
@@ -30,13 +30,13 @@ import org.apache.hadoop.fs.Path
 
 import org.apache.spark.sql.{AnalysisException, Row, SparkSession}
 import org.apache.spark.sql.catalyst.TableIdentifier
+import org.apache.spark.sql.catalyst.analysis.NoSuchPartitionException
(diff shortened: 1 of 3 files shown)


cloud-fan:
cc @gatorsmile

rxin:
LGTM otherwise.

sql/core/src/main/scala/org/apache/spark/sql/execution/command/tables.scala
```diff
@@ -324,38 +324,45 @@ case class TruncateTableCommand(
...
         val normalizedSpec = partitionSpec.map { spec =>
           PartitioningUtils.normalizePartitionSpec(
             spec,
-            table.partitionColumnNames,
+            partCols,
             table.identifier.quotedString,
             spark.sessionState.conf.resolver)
         }
-        catalog.listPartitions(table.identifier, normalizedSpec).map(_.storage.locationUri)
+        val parts =
```
gatorsmile: Nit: How about `partLocations`?

gatorsmile:
LGTM except a minor comment and test pending Jenkins

rxin:
LGTM pending Jenkins.

cloud-fan:

PR #15806, opened 2016-11-08
[SPARK-18345][STRUCTURED STREAMING] Structured Streaming quick examples fails with default configuration

oza:

This PR fixes failure of a quick example of structured streaming with default configuration, because of failure of HDFS connection. 
* Fix to add prefix "file:///" when useTempCheckpointLocation is enabled.

## How was this patch tested?

I test this with manual test: running a quick example of structured streaming with default configuration. With this fix, it works well without configuring HDFS.

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/streaming/StreamingQueryManager.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/streaming/StreamingQueryManager.scala
@@ -209,7 +209,7 @@ class StreamingQueryManager private[sql] (sparkSession: SparkSession) {
         }
       }.getOrElse {
         if (useTempCheckpointLocation) {
-          Utils.createTempDir(namePrefix = s"temporary").getCanonicalPath
+          "file:///" + Utils.createTempDir(namePrefix = "temporary").getCanonicalPath
         } else {
           throw new AnalysisException(
             "checkpointLocation must be specified either " +


rxin:
cc @brkyvz

sql/core/src/main/scala/org/apache/spark/sql/streaming/StreamingQueryManager.scala
```diff
@@ -219,10 +219,11 @@ class StreamingQueryManager private[sql] (sparkSession: SparkSession) {
       }
 
       // If offsets have already been created, we trying to resume a query.
-      if (!recoverFromCheckpointLocation) {
+      if (recoverFromCheckpointLocation) {
```
brkyvz: this line is correct. Please don't change this line

sql/core/src/main/scala/org/apache/spark/sql/streaming/StreamingQueryManager.scala
```diff
@@ -219,10 +219,11 @@ class StreamingQueryManager private[sql] (sparkSession: SparkSession) {
...
         val checkpointPath = new Path(checkpointLocation, "offsets")
         val fs = checkpointPath.getFileSystem(df.sparkSession.sessionState.newHadoopConf())
         if (fs.exists(checkpointPath)) {
+          // Currently, offset recovery from checkpoint is not supported.
```
brkyvz: why not? it is supported

examples/src/main/scala/org/apache/spark/examples/sql/streaming/StructuredNetworkWordCount.scala
```diff
@@ -68,6 +70,8 @@ object StructuredNetworkWordCount {
     val query = wordCounts.writeStream
       .outputMode("complete")
       .format("console")
+      .option("checkpointLocation",
```
brkyvz: we don't support checkpoint recovery for console sinks though

brkyvz:
Hi @oza, thank you for this patch, however I'm not sure if this patch fixes anything. What was your problem, what didn't work for you?

oza:
@brkyvz Thanks for taking a look. The problem I faced is: without HDFS configuration, the [quick example](https://spark.apache.org/docs/latest/structured-streaming-programming-guide.html#quick-example) results in failure with connection error to try to connect to HDFS(full stack trace is attached on [jira](https://issues.apache.org/jira/browse/SPARK-18345)). How to reproduce is:

* Download spark 2.0.1 or build trunk code.
* Run the [quick example](http://spark.apache.org/docs/latest/quick-start.html): results in fail

The documentation is described as quick start, so I think we should avoid any configuration.

sql/core/src/main/scala/org/apache/spark/sql/streaming/StreamingQueryManager.scala
oza: Maybe I understood the code path wrongly. I thought that the recovery will be done if fs.exists(checkpointPath) is true. However, it will only throw the exception, so I thought that offset recovery from checkpoint is not supported. Am I missing something?

sql/core/src/main/scala/org/apache/spark/sql/streaming/StreamingQueryManager.scala
kiszk: If I correctly interpret [this comment](https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/streaming/StreamingQueryManager.scala#L181-L183), an exception will be thrown when `recoverFromCheckpointLocation` == `false` and `fs.exists(checkpointPath)` == `true` are satisfied.
Is [the comment](https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/streaming/StreamingQueryManager.scala#L181-L183) wrong if this line is correct?

oza:
Now I understanding the behaviour. Based on this, I would like to suggest to add an option to disable checkpoint for the testing or quick start example.  What do you think?

brkyvz:
I think the change you should make should be for:
https://github.com/apache/spark/pull/15806/files#diff-7dc261474784c6402f7020ffe7f61038R212
where you always append `file:///` otherwise, it will use the default FS, which may be hdfs, or whatever

cc @marmbrus is there a reason we can't just create the temp checkpoint directory to be local? I understand that if someone's running the test in a distributed environment, then all machines will create checkpoint directories, but since it's temporary, and is not meant for recovery, isn't that okay?

examples/src/main/scala/org/apache/spark/examples/sql/streaming/StructuredNetworkWordCount.scala
```diff
@@ -68,6 +70,8 @@ object StructuredNetworkWordCount {
...
+        "file:///" + System.getProperty("java.io.tmpdir") + "/spark_streaming_" + UUID.randomUUID())
```
srowen: It's a bit nicer to use `Files.createTempDirectory` and call toURI if you need to rather than assemble it directly. It will deal with making the dir unique and all that.

oza:
@brkyvz thanks for explanation, I will fix the line to work it on local mode without any breakage.

oza:
@brkyvz @srowen could you check the latest PR? I added prefix "file:///" before the result of Utils.createTempDir, since Utils.createTempDir only creates local directory via java.io.File.

sql/core/src/main/scala/org/apache/spark/sql/streaming/StreamingQueryManager.scala
```diff
@@ -209,7 +209,7 @@ class StreamingQueryManager private[sql] (sparkSession: SparkSession) {
         }
       }.getOrElse {
         if (useTempCheckpointLocation) {
-          Utils.createTempDir(namePrefix = s"temporary").getCanonicalPath
+          "file:///" + Utils.createTempDir(namePrefix = s"temporary").getCanonicalPath
```
kiszk: Is this a good chance to remove 's' before `"temporary"`?

oza:
Updating patch based on kiszk's comment:
https://github.com/apache/spark/pull/15806#discussion_r87393524

oza:
I'm sorry, but I had a hadoop configuration on my local: I noticed it on PR about SPARK-18345. I will check whether it happens after clearing configuration.

oza:
This was my mistake of configuration. After clearing the configuration about Hadoop, it works well. Closing this as not a problem. Thanks all for helping me, reviewing patch!

PR #15276, opened 2016-09-28
[SPARK-17703][SQL] Add unnamed version of addReferenceObj for minor objects.

ueshin:
## What changes were proposed in this pull request?

There are many minor objects in references, which are extracted to the generated class field, e.g. `errMsg` in `GetExternalRowField` or `ValidateExternalType`, but number of fields in class is limited so we should reduce the number.
This pr adds unnamed version of `addReferenceObj` for these minor objects not to store the object into field but refer it from the `references` field at the time of use.
## How was this patch tested?

Existing tests.

PR #15808, opened 2016-11-08
[SPARK-17703][SQL][BACKPORT-2.0] Add unnamed version of addReferenceObj for minor objects.

ueshin:

This is a backport of #15276.

There are many minor objects in references, which are extracted to the generated class field, e.g. `errMsg` in `GetExternalRowField` or `ValidateExternalType`, but number of fields in class is limited so we should reduce the number.
This pr adds unnamed version of `addReferenceObj` for these minor objects not to store the object into field but refer it from the `references` field at the time of use.
## How was this patch tested?

Existing tests.


rxin:
Thanks - merging in. Can you close this?

ueshin:
Sure. Thanks.

PR #15809, opened 2016-11-08
[SPARK-18268][ML][MLLib] ALS fail with better message if ratings is empty rdd

techaddict:
ALS.run fail with better message if ratings is empty rdd
ALS.train and ALS.trainImplicit are also affected

## How was this patch tested?
added new tests


srowen:
What about putting this check in `.ml.ALS.train()`? I think that's the method at the 'heart' of all the various entry points. That is, you'd catch any usage that didn't go through `.mllib.ALS` too.

A tiny unit test wouldn't hurt.

techaddict:
@srowen done 👍

thunterdb:
LGTM

aditya1702:
What about the python code? this was for scala only right?

srowen:
Python API calls the Scala API so I think this is sufficient.

aditya1702:
@srowen Ok. Thanks 👍

PR #15627, opened 2016-10-25
[SPARK-18099][YARN] Fail if same files added to distributed cache for --files and --archives

kishorvpatil:
## What changes were proposed in this pull request?

During spark-submit, if yarn dist cache is instructed to add same file under --files and --archives, This code change ensures the spark yarn distributed cache behaviour is retained i.e. to warn and fail if same files is mentioned in both --files and --archives.
## How was this patch tested?

Manually tested: 
1. if same jar is mentioned in --jars and --files it will continue to submit the job. 
- basically functionality [SPARK-14423] #12203 is unchanged 
  1. if same file is mentioned in --files and --archives it will fail to submit the job.

Please review https://cwiki.apache.org/confluence/display/SPARK/Contributing+to+Spark before opening a pull request.

… under archives and files

PR #15810, opened 2016-11-08
[SPARK-18357] Fix yarn files/archive broken issue andd unit tests

kishorvpatil:

The #15627 broke functionality with yarn --files --archives does not accept any files.
This patch ensures that --files and --archives accept unique files.

## How was this patch tested?

A. I added unit tests.
B. Also, manually tested --files with --archives to throw exception if duplicate files are specified and continue if unique files are specified.


tgravescs:
+1

PR #15811, opened 2016-11-08
[SPARK-18361] [PySpark] Expose RDD localCheckpoint in PySpark

gabrielhuang:

Expose RDD's localCheckpoint() and associated functions in PySpark.

## How was this patch tested?

I added a UnitTest in python/pyspark/tests.py which passes.

I certify that this is my original work, and I license it to the project under the project's open source license.

Gabriel HUANG
Developer at Cardabel (http://cardabel.com/)

### DIFF
--- a/python/pyspark/rdd.py
+++ b/python/pyspark/rdd.py
@@ -181,7 +181,6 @@ class RDD(object):
         self._jrdd = jrdd
         self.is_cached = False
         self.is_checkpointed = False
-        self.is_locally_checkpointed = False
         self.ctx = ctx
         self._jrdd_deserializer = jrdd_deserializer
         self._id = jrdd.id()
@@ -287,7 +286,6 @@ class RDD(object):
 
         The checkpoint directory set through L{SparkContext.setCheckpointDir()} is not used.
         """
-        self.is_locally_checkpointed = True
         self._jrdd.rdd().localCheckpoint()
 
     def isLocallyCheckpointed(self):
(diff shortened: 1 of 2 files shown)


andrewor14:
ok to test

python/pyspark/rdd.py
```diff
@@ -181,6 +181,7 @@ def __init__(self, jrdd, ctx, jrdd_deserializer=AutoBatchedSerializer(PickleSeri
         self._jrdd = jrdd
         self.is_cached = False
         self.is_checkpointed = False
+        self.is_locally_checkpointed = False
```
andrewor14: is this read anywhere?

andrewor14:
Looks good, just one question.

python/pyspark/rdd.py
gabrielhuang: No it's not. I just mirrored the behavior of self.is_checkpointed: the flag self.is_locally_checkpointed is set by the Python wrappers, but isLocallyChecked() wraps around the scala equivalent, which I imagine reads itself a scala flag.

gabrielhuang:
There's a sleep(1.) Instruction that causes the test to time out. Is that a problem?

viirya:
Looks like just an IO problem when archiving logs after successful tests.

viirya:
LGTM

viirya:
retest this please.

python/pyspark/rdd.py
andrewor14: so we probably don't need it since we're always just calling the scala version

gabrielhuang:
Ok, I removed the Python flag.

gabrielhuang:
retest this please.

andrewor14:
LGTM merging into master thanks.

rxin:
@andrewor14 did you also merge this in branch-2.1?

andrewor14:
I did

PR #15813, opened 2016-11-08
[SPARK-18362][SQL] Use TextFileFormat in implementation of CSVFileFormat

JoshRosen:

This patch significantly improves the IO / file listing performance of schema inference in Spark's built-in CSV data source.

Previously, this data source used the legacy `SparkContext.hadoopFile` and `SparkContext.hadoopRDD` methods to read files during its schema inference step, causing huge file-listing bottlenecks on the driver.

This patch refactors this logic to use Spark SQL's `text` data source to read files during this step. The text data source still performs some unnecessary file listing (since in theory we already have resolved the table prior to schema inference and therefore should be able to scan without performing _any_ extra listing), but that listing is much faster and takes place in parallel. In one production workload operating over tens of thousands of files, this change managed to reduce schema inference time from 7 minutes to 2 minutes.

A similar problem also affects the JSON file format and this patch originally fixed that as well, but I've decided to split that change into a separate patch so as not to conflict with changes in another JSON PR.

## How was this patch tested?

Existing unit tests, plus manual benchmarking on a production workload.


sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/csv/CSVFileFormat.scala
```diff
@@ -210,14 +196,21 @@ class CSVFileFormat extends TextBasedFileFormat with DataSourceRegister {
...
+      inputPaths: Seq[String]): Dataset[String] = {
     if (Charset.forName(options.charset) == StandardCharsets.UTF_8) {
-      sparkSession.sparkContext.textFile(location)
+      sparkSession.baseRelationToDataFrame(
+        DataSource.apply(
+          sparkSession,
+          paths = inputPaths,
+          className = classOf[TextFileFormat].getName
+        ).resolveRelation(checkFilesExist = false))
+        .select("value").as[String](Encoders.STRING)
```
JoshRosen: I copied this logic from the `text` method in `DataFrameReader`, so that's where the `value` came from.

JoshRosen:
Jenkins, retest this please

JoshRosen:
Jenkins, retest this please

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/csv/CSVFileFormat.scala
```diff
@@ -173,35 +178,17 @@ class CSVFileFormat extends TextBasedFileFormat with DataSourceRegister {
...
-
   /**
    * Returns the first line of the first non-empty file in path
    */
-  private def findFirstLine(options: CSVOptions, rdd: RDD[String]): String = {
+  private def findFirstLine(options: CSVOptions, lines: Dataset[String]): String = {
     if (options.isCommentSet) {
       val comment = options.comment.toString
-      rdd.filter { line =>
+      lines.filter { line =>
```
liancheng: Using untyped `filter` can be more performant here since we don't need to pay for the extra de/serialization costs:

```scala
lines.filter(length(trim($"value")) > 0 && $"value".startsWith(comment))
```

NathanHowell:
Any thoughts on modifying `JsonToStruct` to support arrays (and options), then parsing could be something like:

```
dataset.select(
  Column(Inline(
    JsonToValue(
      ArrayType(schema),
      options,
      Column("value").expr))))
```

Likely out of scope for this pull request, but if there is a push to migrate from `RDD[T]` to `Dataset[T]` it would clean things up a bit.

JoshRosen:
@NathanHowell, I've gone ahead and removed the JSON changes from this PR; now it only touches CSV and thus should not conflict with your work.

@liancheng, want to give this a final review? I've addressed your earlier comments.

rxin:
Merging in master.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/csv/CSVFileFormat.scala
```diff
@@ -173,51 +179,37 @@ class CSVFileFormat extends TextBasedFileFormat with DataSourceRegister {
...
+          sparkSession,
+          paths = inputPaths,
+          className = classOf[TextFileFormat].getName
+        ).resolveRelation(checkFilesExist = false))
+        .select("value").as[String](Encoders.STRING)
     } else {
       val charset = options.charset
-      sparkSession.sparkContext
-        .hadoopFile[LongWritable, Text, TextInputFormat](location)
+      val rdd = sparkSession.sparkContext
```
rxin: @JoshRosen do you know why the special handling for non-utf8 encoding is needed? I would think TextFileFormat itself already supports that since it is reading it in from Hadoop Text.

JoshRosen: I'm not sure; I think this was a carryover from `spark-csv`.

rxin: cc @falaki 
Can you chime in?

HyukjinKwon: @rxin, I made a PR to address it at https://github.com/apache/spark/pull/29063 FYI.

PR #15815, opened 2016-11-08
[DOCS][SPARK-18365] Improve Sample Method Documentation

bllchmbrs:

I found the documentation for the sample method to be confusing, this adds more clarification across all languages.

- [x] Scala
- [x] Python
- [x] R
- [x] RDD Scala
- [ ] RDD Python with SEED
- [X] RDD Java
- [x] RDD Java with SEED
- [x] RDD Python

## How was this patch tested?

NA

### DIFF
--- a/R/pkg/R/DataFrame.R
+++ b/R/pkg/R/DataFrame.R
@@ -937,7 +937,7 @@ setMethod("unique",
 #' Sample
 #'
 #' Return a sampled subset of this SparkDataFrame using a random seed. 
-#' Note that this is not guaranteed to provide exactly the fraction specified
+#' Note: that this is not guaranteed to provide exactly the fraction specified
 #' of the total count of of the given SparkDataFrame.
 #'
 #' @param x A SparkDataFrame
--- a/R/pkg/R/DataFrame.R
+++ b/R/pkg/R/DataFrame.R
@@ -937,7 +937,7 @@ setMethod("unique",
 #' Sample
 #'
 #' Return a sampled subset of this SparkDataFrame using a random seed. 
-#' Note: that this is not guaranteed to provide exactly the fraction specified
+#' Note: this is not guaranteed to provide exactly the fraction specified
 #' of the total count of of the given SparkDataFrame.
 #'
 #' @param x A SparkDataFrame
--- a/R/pkg/R/DataFrame.R
+++ b/R/pkg/R/DataFrame.R
@@ -936,7 +936,9 @@ setMethod("unique",
 
 #' Sample
 #'
-#' Return a sampled subset of this SparkDataFrame using a random seed.
+#' Return a sampled subset of this SparkDataFrame using a random seed. 
+#' Note that this is not guaranteed to provide exactly the fraction specified
+#' of the total count of of the given SparkDataFrame.
 #'
 #' @param x A SparkDataFrame
 #' @param withReplacement Sampling with replacement or not
--- a/core/src/main/scala/org/apache/spark/api/java/JavaRDD.scala
+++ b/core/src/main/scala/org/apache/spark/api/java/JavaRDD.scala
@@ -98,7 +98,7 @@ class JavaRDD[T](val rdd: RDD[T])(implicit val classTag: ClassTag[T])
   def repartition(numPartitions: Int): JavaRDD[T] = rdd.repartition(numPartitions)
 
   /**
-   * Return a sampled subset of this RDD.
+   * Return a sampled subset of this RDD with a random seed.
    * Note: this is NOT guaranteed to provide exactly the fraction of the count
    * of the given [[RDD]].
    *
@@ -111,7 +111,9 @@ class JavaRDD[T](val rdd: RDD[T])(implicit val classTag: ClassTag[T])
     sample(withReplacement, fraction, Utils.random.nextLong)
 
   /**
-   * Return a sampled subset of this RDD.
+   * Return a sampled subset of this RDD, with a user-supplied seed.
+   * Note: this is NOT guaranteed to provide exactly the fraction of the count
+   * of the given [[RDD]].
    *
    * @param withReplacement can elements be sampled multiple times (replaced when sampled out)
    * @param fraction expected size of the sample as a fraction of this RDD's size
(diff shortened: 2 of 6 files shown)


brkyvz:
@anabranch I don't see how the documentation was wrong. The second argument doesn't take the seed as a parameter, therefore the seed is random

brkyvz:
LGTM

sql/core/src/main/scala/org/apache/spark/sql/Dataset.scala
```diff
@@ -1612,7 +1612,9 @@ class Dataset[T] private[sql](
   }
 
   /**
-   * Returns a new Dataset by sampling a fraction of rows.
+   * Returns a new Dataset by sampling a fraction of rows, using a user-supplied seed.
+   * Note: this is NOT guaranteed to provide exactly the fraction specified of the 
+   * Dataset.
```
HyukjinKwon: Should we change this to `[[Dataset]]` instead to mark down pretty?

sql/core/src/main/scala/org/apache/spark/sql/Dataset.scala
```diff
@@ -1612,7 +1612,9 @@ class Dataset[T] private[sql](
   }
 
   /**
-   * Returns a new Dataset by sampling a fraction of rows.
+   * Returns a new Dataset by sampling a fraction of rows, using a user-supplied seed.
+   * Note: this is NOT guaranteed to provide exactly the fraction specified of the
```
HyukjinKwon: Maybe, we need a newline between description and `Note:` just to be consistent with the others in here and other places such as `functions.scala` if more commits should be pushed.

brkyvz: +1

srowen:
@anabranch OK, the purpose of this JIRA and PR have changed significantly. Could you update all the titles and descriptions? 

I think the right thing to do is make the documentation consistent across all sample methods, including RDD and DataFrame and in Python and R. They all describe this a bit differently. For example, the RDD method covers this, in a way, by talking about _expected_ sample size.

I'd merge this if it were making every related method's docs consistent.

anabranch:
Sounds good to me. I will update it shortly.

sql/core/src/main/scala/org/apache/spark/sql/Dataset.scala
anabranch: Done in next commit.

sql/core/src/main/scala/org/apache/spark/sql/Dataset.scala
anabranch: Added a new line.

anabranch:
@srowen Think this is probably ready.

- [ ] Updated All Languages
- [ ] Updated Ticket Description

R/pkg/R/DataFrame.R
```diff
@@ -936,7 +936,9 @@ setMethod("unique",
 
 #' Sample
 #'
-#' Return a sampled subset of this SparkDataFrame using a random seed.
+#' Return a sampled subset of this SparkDataFrame using a random seed. 
+#' Note that this is not guaranteed to provide exactly the fraction specified
```
HyukjinKwon: @anabranch it is already probably fine (I hope this does not sound like nitpicking) but maybe we would better just match this up as below (there are some examples for this in https://github.com/anabranch/spark/blob/b4f2611214e00e11040d78fa33b8772588897264/R/pkg/R/functions.R#L2299-L2300 and https://github.com/anabranch/spark/blob/b4f2611214e00e11040d78fa33b8772588897264/R/pkg/R/mllib.R#L606-L607):

```
#' Return a sampled subset of this SparkDataFrame using a random seed. 
#' 
#' Note: that this is not guaranteed to provide exactly the fraction specified
```

I understand `Note:` and `NOTE:` are mixed across the documentation (including annotations and indentation). Let me try to deal with this across all documentation in a separate PR.

HyukjinKwon:
For Python documenation, it seems fine.

R/pkg/R/DataFrame.R
anabranch: Yeah I see what you're saying but I think it might be a bit out of scope for this PR. What your goal is definitely makes sense as there should be a standard, but I don't think I'm degrading anything significantly. I will add the `:` character.

anabranch:
The test failure seems quite unrelated but we'll see if it happens again.

felixcheung:
LGTM thanks!

srowen (changes requested):
I think we should give `RDD`, `JavaRDD` and `rdd.py` a similar treatment for consistency. They also have a sample method.

python/pyspark/sql/dataframe.py
```diff
@@ -549,6 +549,11 @@ def distinct(self):
     def sample(self, withReplacement, fraction, seed=None):
         """Returns a sampled subset of this :class:`DataFrame`.
 
+        .. note::
```
srowen: Tiny question about this syntax (I don't know it) -- I see other instances of this in the code base have to use a line continuation `\` when it spans several lines? and then indent continuation lines flush with "..".

core/src/main/scala/org/apache/spark/api/java/JavaRDD.scala
```diff
@@ -99,6 +99,8 @@ class JavaRDD[T](val rdd: RDD[T])(implicit val classTag: ClassTag[T])
 
   /**
    * Return a sampled subset of this RDD.
+   * Note: this is NOT guaranteed to provide exactly the fraction of the count
```
srowen: There are a couple overloads of sample here; update them all and maybe apply the clarification about seed you added below?

anabranch: Yes. Will add this.

anabranch: There is also another method I forgot in the python RDD that I will fix now as well.

anabranch: Fixed, the Python one didn't need it once i re-read the docs.

anabranch:
failures also seem unrelated.

srowen:
Merged to master/2.1

PR #15816, opened 2016-11-08
SPARK-18368: Fix regexp_replace with task serialization.

rdblue:

This makes the result value both transient and lazy, so that if the RegExpReplace object is initialized then serialized, `result: StringBuffer` will be correctly initialized.

## How was this patch tested?

* Verified that this patch fixed the query that found the bug.
* Added a test case that fails without the fix.

### DIFF
--- a/sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/RegexpExpressionsSuite.scala
+++ b/sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/RegexpExpressionsSuite.scala
@@ -17,8 +17,7 @@
 
 package org.apache.spark.sql.catalyst.expressions
 
-import org.apache.spark.{SparkConf, SparkFunSuite}
-import org.apache.spark.serializer.JavaSerializer
+import org.apache.spark.SparkFunSuite
 import org.apache.spark.sql.catalyst.dsl.expressions._
 import org.apache.spark.sql.types.StringType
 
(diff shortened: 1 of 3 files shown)


hvanhovell:
Wow I am actually surprised that this wasn't caught before. Integration testing should have uncovered this. Could you add one?

hvanhovell:
Sorry I missed your update.

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/RegexpExpressionsSuite.scala
```diff
@@ -191,4 +192,17 @@ class RegexpExpressionsSuite extends SparkFunSuite with ExpressionEvalHelper {
     checkEvaluation(StringSplit(s1, s2), null, row3)
   }
 
+  test("RegExpReplace serialization") {
+    val serializer = new JavaSerializer(new SparkConf()).newInstance
```
hvanhovell: Maybe we should add such a similar test to `ExpressionEvalHelper`.

rdblue:
@hvanhovell, I updated the test so that all expressions passed through `checkEvaluation` are serialized and deserialized before the checks are run. I'll open an issue if it catches anything else. I'd rather not hold up this PR if it finds lots of other bugs.

rxin:
Merging in master/branch-2.1/branch-2.0.

rxin:
I'm surprised too that we haven't caught this earlier ...

rdblue:
Thanks @rxin and @hvanhovell! I appreciate the quick reviews.

yhuai:
I am wondering if it breaks some tests?

```
org.apache.spark.sql.catalyst.expressions.MathExpressionsSuite.e
org.apache.spark.sql.catalyst.expressions.MathExpressionsSuite.pi
org.apache.spark.sql.catalyst.expressions.MathExpressionsSuite.binary log
org.apache.spark.sql.catalyst.expressions.StringExpressionsSuite.format_number / FormatNumber
```

http://spark-tests.appspot.com/test-details?suite_name=org.apache.spark.sql.catalyst.expressions.MathExpressionsSuite&test_name=e

https://amplab.cs.berkeley.edu/jenkins/job/spark-master-test-sbt-hadoop-2.6/1952/

yhuai:
oh, seems the last commit did not pass build. 

Sorry. I am going to revert this patch.

yhuai:
Reverted from master/branch-2.1/branch-2.0.

yhuai:
@rdblue Can you send a new pr with the fix? Thanks!

zsxwing:
If `result` is called before serialized, it has the same issue. Right?

PR #12770, opened 2016-04-28
[SPARK-14994][SQL] Remove execution hive from HiveSessionState

rxin:
## What changes were proposed in this pull request?

This patch removes executionHive from HiveSessionState and HiveSharedState.
## How was this patch tested?

Updated test cases.

PR #15819, opened 2016-11-09
[SPARK-18372][SQL][Branch-1.6].Staging directory fail to be removed 

merlintang:

This fix is related to be bug: https://issues.apache.org/jira/browse/SPARK-18372 . 
The insertIntoHiveTable would generate a .staging directory, but this directory  fail to be removed in the end. 

This is backport from spark 2.0.x code, and is related to PR #12770

## How was this patch tested?
manual tests

Author: Mingjie Tang <mtang@hortonworks.com>

### DIFF
--- a/sql/hive/src/main/scala/org/apache/spark/sql/hive/execution/InsertIntoHiveTable.scala
+++ b/sql/hive/src/main/scala/org/apache/spark/sql/hive/execution/InsertIntoHiveTable.scala
@@ -27,6 +27,8 @@ import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.fs.{FileSystem, Path}
 import org.apache.hadoop.hive.common.FileUtils
 
+import scala.util.control.NonFatal
+
 import scala.collection.JavaConverters._
 import org.apache.hadoop.hive.conf.HiveConf
 import org.apache.hadoop.hive.conf.HiveConf.ConfVars
@@ -61,6 +63,7 @@ case class InsertIntoHiveTable(
   @transient private lazy val hiveContext = new Context(sc.hiveconf)
   @transient private lazy val catalog = sc.catalog
 
+  @transient var createdTempDir: Option[Path] = None
   val stagingDir = new HiveConf().getVar(HiveConf.ConfVars.STAGINGDIR)
 
   private def executionId: String = {
@@ -88,6 +91,7 @@ case class InsertIntoHiveTable(
       if (!FileUtils.mkdir(fs, dir, true, hadoopConf)) {
         throw new IllegalStateException("Cannot create staging directory  '" + dir.toString + "'")
       }
+      createdTempDir = Some(dir)
       fs.deleteOnExit(dir)
     }
     catch {
@@ -321,6 +325,15 @@ case class InsertIntoHiveTable(
         qualifiedTableName,
         overwrite,
         holdDDLTime)
+    }
+
+    // Attempt to delete the staging directory and the inclusive files. If failed, the files are
+    // expected to be dropped at the normal termination of VM since deleteOnExit is used.
+    try {
+      createdTempDir.foreach { path => path.getFileSystem(jobConf).delete(path, true) }
+    } catch {
+      case NonFatal(e) =>
+        logWarning(s"Unable to delete staging directory: $stagingDir.\n" + e)
     }
 
     // Invalidate the cache.
--- a/sql/hive/src/main/scala/org/apache/spark/sql/hive/execution/InsertIntoHiveTable.scala
+++ b/sql/hive/src/main/scala/org/apache/spark/sql/hive/execution/InsertIntoHiveTable.scala
@@ -70,7 +70,7 @@ case class InsertIntoHiveTable(
     val rand: Random = new Random
     val format: SimpleDateFormat = new SimpleDateFormat("yyyy-MM-dd_HH-mm-ss_SSS")
     val executionId: String = "hive_" + format.format(new Date) + "_" + Math.abs(rand.nextLong)
-     executionId
+    executionId
   }
 
   private def getStagingDir(inputPath: Path, hadoopConf: Configuration): Path = {
--- a/sql/hive/src/main/scala/org/apache/spark/sql/hive/execution/InsertIntoHiveTable.scala
+++ b/sql/hive/src/main/scala/org/apache/spark/sql/hive/execution/InsertIntoHiveTable.scala
@@ -70,7 +70,7 @@ case class InsertIntoHiveTable(
     val rand: Random = new Random
     val format: SimpleDateFormat = new SimpleDateFormat("yyyy-MM-dd_HH-mm-ss_SSS")
     val executionId: String = "hive_" + format.format(new Date) + "_" + Math.abs(rand.nextLong)
-    return executionId
+     executionId
   }
 
   private def getStagingDir(inputPath: Path, hadoopConf: Configuration): Path = {
--- a/sql/hive/src/main/scala/org/apache/spark/sql/hive/execution/InsertIntoHiveTable.scala
+++ b/sql/hive/src/main/scala/org/apache/spark/sql/hive/execution/InsertIntoHiveTable.scala
@@ -17,24 +17,31 @@
 
 package org.apache.spark.sql.hive.execution
 
+import java.io.IOException
+import java.net.URI
+import java.text.SimpleDateFormat
 import java.util
+import java.util.{Date, Random}
+
+import org.apache.hadoop.conf.Configuration
+import org.apache.hadoop.fs.{FileSystem, Path}
+import org.apache.hadoop.hive.common.FileUtils
 
 import scala.collection.JavaConverters._
-
 import org.apache.hadoop.hive.conf.HiveConf
 import org.apache.hadoop.hive.conf.HiveConf.ConfVars
+import org.apache.hadoop.hive.ql.exec.TaskRunner
 import org.apache.hadoop.hive.ql.plan.TableDesc
 import org.apache.hadoop.hive.ql.{Context, ErrorMsg}
 import org.apache.hadoop.hive.serde2.Serializer
 import org.apache.hadoop.hive.serde2.objectinspector.ObjectInspectorUtils.ObjectInspectorCopyOption
 import org.apache.hadoop.hive.serde2.objectinspector._
 import org.apache.hadoop.mapred.{FileOutputCommitter, FileOutputFormat, JobConf}
-
 import org.apache.spark.rdd.RDD
 import org.apache.spark.sql.Row
 import org.apache.spark.sql.catalyst.InternalRow
 import org.apache.spark.sql.catalyst.expressions.Attribute
-import org.apache.spark.sql.execution.{UnaryNode, SparkPlan}
+import org.apache.spark.sql.execution.{SparkPlan, UnaryNode}
 import org.apache.spark.sql.hive.HiveShim.{ShimFileSinkDesc => FileSinkDesc}
 import org.apache.spark.sql.hive._
 import org.apache.spark.sql.types.DataType
@@ -53,6 +60,61 @@ case class InsertIntoHiveTable(
   @transient lazy val outputClass = newSerializer(table.tableDesc).getSerializedClass
   @transient private lazy val hiveContext = new Context(sc.hiveconf)
   @transient private lazy val catalog = sc.catalog
+
+  val stagingDir = new HiveConf().getVar(HiveConf.ConfVars.STAGINGDIR)
+
+  private def executionId: String = {
+    val rand: Random = new Random
+    val format: SimpleDateFormat = new SimpleDateFormat("yyyy-MM-dd_HH-mm-ss_SSS")
+    val executionId: String = "hive_" + format.format(new Date) + "_" + Math.abs(rand.nextLong)
+    return executionId
+  }
+
+  private def getStagingDir(inputPath: Path, hadoopConf: Configuration): Path = {
+    val inputPathUri: URI = inputPath.toUri
+    val inputPathName: String = inputPathUri.getPath
+    val fs: FileSystem = inputPath.getFileSystem(hadoopConf)
+    val stagingPathName: String =
+      if (inputPathName.indexOf(stagingDir) == -1) {
+        new Path(inputPathName, stagingDir).toString
+      } else {
+        inputPathName.substring(0, inputPathName.indexOf(stagingDir) + stagingDir.length)
+      }
+    val dir: Path =
+      fs.makeQualified(
+        new Path(stagingPathName + "_" + executionId + "-" + TaskRunner.getTaskRunnerID))
+    logDebug("Created staging dir = " + dir + " for path = " + inputPath)
+    try {
+      if (!FileUtils.mkdir(fs, dir, true, hadoopConf)) {
+        throw new IllegalStateException("Cannot create staging directory  '" + dir.toString + "'")
+      }
+      fs.deleteOnExit(dir)
+    }
+    catch {
+      case e: IOException =>
+        throw new RuntimeException(
+          "Cannot create staging directory '" + dir.toString + "': " + e.getMessage, e)
+
+    }
+    return dir
+  }
+
+  private def getExternalScratchDir(extURI: URI, hadoopConf: Configuration): Path = {
+    getStagingDir(new Path(extURI.getScheme, extURI.getAuthority, extURI.getPath), hadoopConf)
+  }
+
+  def getExternalTmpPath(path: Path, hadoopConf: Configuration): Path = {
+    val extURI: URI = path.toUri
+    if (extURI.getScheme == "viewfs") {
+      getExtTmpPathRelTo(path.getParent, hadoopConf)
+    } else {
+      new Path(getExternalScratchDir(extURI, hadoopConf), "-ext-10000")
+    }
+  }
+
+  def getExtTmpPathRelTo(path: Path, hadoopConf: Configuration): Path = {
+    new Path(getStagingDir(path, hadoopConf), "-ext-10000") // Hive uses 10000
+  }
 
   private def newSerializer(tableDesc: TableDesc): Serializer = {
     val serializer = tableDesc.getDeserializerClass.newInstance().asInstanceOf[Serializer]
@@ -129,7 +191,9 @@ case class InsertIntoHiveTable(
     // instances within the closure, since Serializer is not serializable while TableDesc is.
     val tableDesc = table.tableDesc
     val tableLocation = table.hiveQlTable.getDataLocation
-    val tmpLocation = hiveContext.getExternalTmpPath(tableLocation)
+    val jobConf = new JobConf(sc.hiveconf)
+    val tmpLocation = getExternalTmpPath(tableLocation, jobConf)
+
     val fileSinkConf = new FileSinkDesc(tmpLocation.toString, tableDesc, false)
     val isCompressed = sc.hiveconf.getBoolean(
       ConfVars.COMPRESSRESULT.varname, ConfVars.COMPRESSRESULT.defaultBoolVal)
@@ -175,7 +239,6 @@ case class InsertIntoHiveTable(
       }
     }
 
-    val jobConf = new JobConf(sc.hiveconf)
     val jobConfSer = new SerializableJobConf(jobConf)
 
     // When speculation is on and output committer class name contains "Direct", we should warn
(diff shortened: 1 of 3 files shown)


rxin:
Can you add some documentation? The current code is very difficult to follow.

cloud-fan:
do you have a unit test to reproduce this bug?

merlintang:
Actually, I do not have the unit test, but the code list below (same as we
posted in the JIRA) can reproduce this bug.

The related code would be this way:
val sqlContext = new org.apache.spark.sql.hive.HiveContext(sc)
sqlContext.sql("CREATE TABLE IF NOT EXISTS T1 (key INT, value STRING)")
sqlContext.sql("LOAD DATA LOCAL INPATH
'../examples/src/main/resources/kv1.txt' INTO TABLE T1")
sqlContext.sql("CREATE TABLE IF NOT EXISTS T2 (key INT, value STRING)")
val sparktestdf = sqlContext.table("T1")
val dfw = sparktestdf.write
dfw.insertInto("T2")
val sparktestcopypydfdf = sqlContext.sql("""SELECT * from T2 """)
sparktestcopypydfdf.show

Our customer and ourself also have manually reproduced this bug for spark
1.6.x and 1.5.x.

For the unit test, because we do not know how to find the hive directory
for the related table in the test case, we can not check the computed
directory in the end.

The solution is that we reuse three functions in the 2.0.2 to create the
staging directory, then this bug is fixed.

sql/hive/src/main/scala/org/apache/spark/sql/hive/execution/InsertIntoHiveTable.scala
```diff
@@ -54,6 +61,61 @@ case class InsertIntoHiveTable(
   @transient private lazy val hiveContext = new Context(sc.hiveconf)
   @transient private lazy val catalog = sc.catalog
 
+  val stagingDir = new HiveConf().getVar(HiveConf.ConfVars.STAGINGDIR)
+
+  private def executionId: String = {
+    val rand: Random = new Random
+    val format: SimpleDateFormat = new SimpleDateFormat("yyyy-MM-dd_HH-mm-ss_SSS")
+    val executionId: String = "hive_" + format.format(new Date) + "_" + Math.abs(rand.nextLong)
+    return executionId
```
fidato13: Can the extra noise (return statement) in scala code be removed please.

sql/hive/src/main/scala/org/apache/spark/sql/hive/execution/InsertIntoHiveTable.scala
```diff
@@ -54,6 +61,61 @@ case class InsertIntoHiveTable(
...
+      }
+      fs.deleteOnExit(dir)
+    }
+    catch {
+      case e: IOException =>
+        throw new RuntimeException(
+          "Cannot create staging directory '" + dir.toString + "': " + e.getMessage, e)
+
+    }
+    return dir
```
fidato13: Can the extra noise (return statement) in scala code be removed please.

sql/hive/src/main/scala/org/apache/spark/sql/hive/execution/InsertIntoHiveTable.scala
merlintang: hi @fidato13 this is ok, since the part of this code is reused from spark 2.0.2.

merlintang:
@cloud-fan @rxin can you review this code? since several customers are complaining about the hive generated empty staging files in the HDFS.

sql/hive/src/main/scala/org/apache/spark/sql/hive/execution/InsertIntoHiveTable.scala
fidato13: @merlintang  Can we take this opportunity to rectify at other places as well, Adding a return statement at the end of a simple method where no complex control flows are introduced would rather make it seem like java style coding. Check on the below link for Scala style guide for reference:
https://github.com/databricks/scala-style-guide#return-statements

sql/hive/src/main/scala/org/apache/spark/sql/hive/execution/InsertIntoHiveTable.scala
```diff
@@ -54,6 +61,61 @@ case class InsertIntoHiveTable(
   @transient private lazy val hiveContext = new Context(sc.hiveconf)
   @transient private lazy val catalog = sc.catalog
 
+  val stagingDir = new HiveConf().getVar(HiveConf.ConfVars.STAGINGDIR)
+
+  private def executionId: String = {
+    val rand: Random = new Random
+    val format: SimpleDateFormat = new SimpleDateFormat("yyyy-MM-dd_HH-mm-ss_SSS")
+    val executionId: String = "hive_" + format.format(new Date) + "_" + Math.abs(rand.nextLong)
```
srowen: Why all this -- just us a UUID? you also have a redundant return and types here.

sql/hive/src/main/scala/org/apache/spark/sql/hive/execution/InsertIntoHiveTable.scala
```diff
@@ -54,6 +61,61 @@ case class InsertIntoHiveTable(
...
+    logDebug("Created staging dir = " + dir + " for path = " + inputPath)
+    try {
+      if (!FileUtils.mkdir(fs, dir, true, hadoopConf)) {
+        throw new IllegalStateException("Cannot create staging directory  '" + dir.toString + "'")
+      }
+      fs.deleteOnExit(dir)
+    }
+    catch {
+      case e: IOException =>
+        throw new RuntimeException(
```
srowen: Don't use RuntimeException; why even handle this?

srowen (commented):
I don't quite see how this removes the staging dir. Just the deleteOnExit? does it need this complexity then?

sql/hive/src/main/scala/org/apache/spark/sql/hive/execution/InsertIntoHiveTable.scala
merlintang: You can find the reason that we use this code is because (1) the old version need to use the hive package to create the staging directory, in the hive code, this staging directory is storied in a hash map, and then these staging directories would be removed when the session is closed. however, our spark code do not trigger the hive session close, then, these directories will not be removed. (2) you can find the pushed code just simulate the hive way to create the staging directory inside the spark rather than based on the hive. Then, the staging directory will be removed. (3) I will fix the return type issue, thanks for your comments @srowen

sql/hive/src/main/scala/org/apache/spark/sql/hive/execution/InsertIntoHiveTable.scala
merlintang: thanks, I will fix it.

fidato13: Cheers.

sql/hive/src/main/scala/org/apache/spark/sql/hive/execution/InsertIntoHiveTable.scala
fidato13: Thanks.

merlintang:
yes, exactly. This path is only for spark 1.x. what i proposed here is that
we need to use the code of spark 2.0.x o fix the bug of spark 1.x. you can
see this message from the my previous replies. I do not want to change the
code, since it will make the 1.x and 2.x in great different.

gatorsmile:
@merlintang Could you please add [Branch-1.6] in your PR title?

merlintang:
it is updated.

cloud-fan:
OK so the problem becomes, do we want to backport this to 1.6? cc @rxin

rxin:
If it is a bug fix and low risk, sure.

merlintang:
this bug is related to 1.5.x as well as 1.6.x.  please backport to 1.5.x as
well.

rxin:
We have stopped making new releases for 1.5 so it makes no sense to backport.

merlintang:
Ok.

cloud-fan:
ok @merlintang can you find out which PR adds these codes to 2.0? Then other people can know what we are backporting in this PR

merlintang:
@cloud-fan  this is related to this PR in the 2.0.x
https://github.com/apache/spark/pull/12770

lichenglin:
I'm using spark 2.0.2 I got a really big hive-stage folder.
May I delete the folder Manually ?
does it make any bad affect on  warehouse?

gatorsmile:
Could you post the layout of that staging folder?

lichenglin:
here is some result for `du -h --max-depth=1 .`
3.3G    ./.hive-staging_hive_2016-12-06_18-17-48_899_1400956608265117052-5
13G     ./.hive-staging_hive_2016-12-06_15-43-35_928_6647980494630196053-5
8.6G    ./.hive-staging_hive_2016-12-06_17-05-51_951_8422682528744006964-5
9.7G    ./.hive-staging_hive_2016-12-06_17-14-44_748_6947381677226271245-5
9.2G    ./day=2016-12-01
8.5G    ./day=2016-11-19

I run a sql like `insert overwrite db.table partition(day='2016-12-06') select * from tmpview`  everyday
each sql create a "hive-staging folder".

Can I delete the folders manually??

merlintang:
do you exit the spark shell ?  I have tested on this, and this staging file
would be removed after we exit the spark shell under spark 2.0.x.

meanwhile, the staging file are used for hive to write data, and if one
hive insert data fail in the middle, the staging file could be used.

lichenglin:
In fact,I'm using zeppelin to run sql.
When I restart spark interpreter , the folders are deleted.
Thank you a lot

gatorsmile:
@lichenglin Another PR https://github.com/apache/spark/pull/16134 is trying to delete the staging directory and the temporary data files (which is pretty big in your case) after each insert.

merlintang:
@gatorsmile what is going on this patch? this is a backport code, thus, can you merge this patch into 1.6.x ?  more than one users are running into this issue in the spark 1.6.x.

gatorsmile:
The current fix does not resolve the issue when users hitting abnormal termination of JVM. In addition, if the JVM does not stop, these temporary files could consume a lot of spaces. Thus, I think https://github.com/apache/spark/pull/16134 needs to be added too.

This is just my opinion. Also need to get the feedbacks from the other Committers.

cloud-fan:
yea, I think we should backport a complete staging dir cleanup functionality to 1.6, let's wait for #16134

merlintang:
Great, once the  #16134 <https://github.com/apache/spark/pull/16134> is
done, we can backport them together.

merlintang:
@cloud-fan @gatorsmile this patch is related to #16134, It seems #16134 would be merged soon. Meanwhile,  should we backport #16104 into 1.6.x? please advise. or else, I just backport #16134 and #12770 to the spark 1.6.x?

merlintang:
@gatorsmile one more customer is running into this issue in the spark 1.6.x. I backport the code #16134 to here and test it manually. Please verify.

merlintang:
@cloud-fan @gatorsmile  I have backport the code from #16134, can you verify and backport this to spark 1.6.x?

gatorsmile:
Let me backport it to Spark 2.0.x at first. We need to follow the backport ordering

merlintang:
@gatorsmile Great! thanks so much, because I was pinged multiple times for this bug. :)

cloud-fan:
https://github.com/apache/spark/pull/16399 has been merged, feel free if you wanna backport to 1.6

sql/hive/src/test/scala/org/apache/spark/sql/hive/InsertIntoHiveTableSuite.scala
```diff
@@ -76,6 +76,8 @@ class InsertIntoHiveTableSuite extends QueryTest with TestHiveSingleton with Bef
       sql("SELECT * FROM createAndInsertTest"),
       testData.collect().toSeq
     )
+
+
```
gatorsmile: Nit: please remove these two lines.

gatorsmile:
Is that possible to backport the test cases in https://github.com/apache/spark/pull/16399?

merlintang:
yes, let me backport the test cases for checking the staging file.

merlintang:
@gatorsmile I have backport the test case in #16339 with small modification. because the "INSERT OVERWRITE TABLE tab SELECT '$i'" will bring the issue from hive side e.g., https://issues.apache.org/jira/browse/HIVE-12200. Thus, I just create a temp table and insert data from the select temp table. please double check and verify.

gatorsmile:
retest this please

sql/hive/src/main/scala/org/apache/spark/sql/hive/execution/InsertIntoHiveTable.scala
```diff
@@ -54,6 +63,63 @@ case class InsertIntoHiveTable(
...
   @transient private lazy val catalog = sc.catalog
 
+  @transient var createdTempDir: Option[Path] = None
+  val stagingDir = new HiveConf().getVar(HiveConf.ConfVars.STAGINGDIR)
+
+  private def executionId: String = {
+    val rand: Random = new Random
+    val format: SimpleDateFormat = new SimpleDateFormat("yyyy-MM-dd_HH-mm-ss_SSS")
+    val executionId: String = "hive_" + format.format(new Date) + "_" + Math.abs(rand.nextLong)
+    return executionId
```
gatorsmile: Please remove the return?

sql/hive/src/test/scala/org/apache/spark/sql/hive/client/VersionsSuite.scala
```diff
@@ -216,5 +218,37 @@ class VersionsSuite extends SparkFunSuite with Logging {
...
+    test(s"$version: Delete the temporary staging directory and files after each insert") {
+      withTempDir { tmpDir =>
+        withTable("tab", "tbl") {
+          sqlContext.sql(
+            s"""
+               |CREATE  TABLE tab(c1 string)
+               |location '${tmpDir.toURI.toString}'
+             """.stripMargin)
+
+          sqlContext.sql("CREATE TABLE tbl AS SELECT 1 AS a")
```
gatorsmile: you can create a temporary view, instead of creating another table.

sql/hive/src/main/scala/org/apache/spark/sql/hive/execution/InsertIntoHiveTable.scala
merlintang: done.

sql/hive/src/test/scala/org/apache/spark/sql/hive/client/VersionsSuite.scala
merlintang: does the temporary view supported in the 1.6.x?  I just used the hivecontext to create the view, but it does not work. because this is small test case, the created table here would be ok. please advise. thanks so much, Tao.

gatorsmile: In 1.6, the function is `registerTempTable`. The name was changed in 2.0 to temp view.

merlintang: thanks Tao, I have created a dataframe, then create registerTempTable as following.
 
 val df = sqlContext.createDataFrame((1 to 2).map(i => (i, "a"))).toDF("key", "value")
 df.select("value").repartition(1).registerTempTable("tbl")

it can work, but it looks like fuzzy. what do you think?

gatorsmile: How about the following line?
```Scala
Seq((1, "a")).toDF("key", "value").registerTempTable("tbl")
```
BTW, I am Xiao Li. : )

gatorsmile: You just want one column. Then, you can do it by
```Scala
Seq(Tuple1("a")).toDF("value").registerTempTable("tbl")
```

merlintang: Sorry Xiao, since one of my best friend is Tao. :). Sorry. It is updated.  Thanks again.

merlintang:
@gatorsmile can you retest the patch, then we can merge. Sorry to ping you multiple times since several users are asking this.

gatorsmile:
retest this please

sql/hive/src/test/scala/org/apache/spark/sql/hive/client/VersionsSuite.scala
```diff
@@ -216,5 +219,37 @@ class VersionsSuite extends SparkFunSuite with Logging {
...
+        assert(sqlContext.table("tbl").collect().toSeq == Seq(Row(1)))
+      }
+    }
+
+    test(s"$version: Delete the temporary staging directory and files after each insert") {
+      withTempDir { tmpDir =>
+        withTable("tab", "tbl") {
+          sqlContext.sql(
+            s"""
+               |CREATE  TABLE tab(c1 string)
```
gatorsmile: Nit: two spaces -> one space

sql/hive/src/main/scala/org/apache/spark/sql/hive/execution/InsertIntoHiveTable.scala
```diff
@@ -54,6 +63,63 @@ case class InsertIntoHiveTable(
...
   @transient private lazy val catalog = sc.catalog
 
+  @transient var createdTempDir: Option[Path] = None
+  val stagingDir = new HiveConf().getVar(HiveConf.ConfVars.STAGINGDIR)
+
+  private def executionId: String = {
+    val rand: Random = new Random
+    val format: SimpleDateFormat = new SimpleDateFormat("yyyy-MM-dd_HH-mm-ss_SSS")
+    val executionId: String = "hive_" + format.format(new Date) + "_" + Math.abs(rand.nextLong)
+     executionId
```
gatorsmile: Nit: an indent issue. Please remove one more space.

sql/hive/src/test/scala/org/apache/spark/sql/hive/client/VersionsSuite.scala
```diff
@@ -216,5 +219,37 @@ class VersionsSuite extends SparkFunSuite with Logging {
...
+               |location '${tmpDir.toURI.toString}'
+             """.stripMargin)
+
+          import sqlContext.implicits._
```
gatorsmile: Nit: move this import to line 231.

gatorsmile:
retest this please

gatorsmile:
Weird... Not sure why the build failed. The build works in my local environment. cc @srowen @JoshRosen

sql/hive/src/test/scala/org/apache/spark/sql/hive/client/VersionsSuite.scala
merlintang: thanks, it is updated.

sql/hive/src/main/scala/org/apache/spark/sql/hive/execution/InsertIntoHiveTable.scala
merlintang: Done! thanks xiao.

sql/hive/src/test/scala/org/apache/spark/sql/hive/client/VersionsSuite.scala
merlintang: Done

gatorsmile:
retest this please

sql/hive/src/test/scala/org/apache/spark/sql/hive/client/VersionsSuite.scala
```diff
@@ -216,5 +219,37 @@ class VersionsSuite extends SparkFunSuite with Logging {
...
+
+    test(s"$version: CREATE TABLE AS SELECT") {
+      withTable("tbl") {
+        sqlContext.sql("CREATE TABLE tbl AS SELECT 1 AS a")
+        assert(sqlContext.table("tbl").collect().toSeq == Seq(Row(1)))
+      }
+    }
+
+    test(s"$version: Delete the temporary staging directory and files after each insert") {
+      import sqlContext.implicits._
```
gatorsmile: Let us roll back to the way you did in the last run, instead of using the temp table. I am not sure whether this trigger the build issue.

merlintang: thanks, xiao, I have reverted that and test locally.

gatorsmile:
retest this please

gatorsmile:
LGTM pending test

gatorsmile:
Thanks! Merging to 1.6

gatorsmile:
@merlintang Can you close this PR? Thanks!

merlintang:
Many thanks, Xiao. I learnt lots.

PR #15820, opened 2016-11-09
[SPARK-18373][SPARK-18529][SS][Kafka]Make failOnDataLoss=false work with Spark jobs

zsxwing:

This PR adds `CachedKafkaConsumer.getAndIgnoreLostData` to handle corner cases of `failOnDataLoss=false`.

It also resolves [SPARK-18529](https://issues.apache.org/jira/browse/SPARK-18529) after refactoring codes: Timeout will throw a TimeoutException.

## How was this patch tested?

Because I cannot find any way to manually control the Kafka server to clean up logs, it's impossible to write unit tests for each corner case. Therefore, I just created `test("stress test for failOnDataLoss=false")` which should cover most of corner cases.

I also modified some existing tests to test for both `failOnDataLoss=false` and `failOnDataLoss=true` to make sure it doesn't break existing logic.

### DIFF
--- a/external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/CachedKafkaConsumer.scala
+++ b/external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/CachedKafkaConsumer.scala
@@ -57,65 +57,53 @@ private[kafka010] case class CachedKafkaConsumer private(
    * Get the record for the given offset, waiting up to timeout ms if IO is necessary.
    * Sequential forward access will use buffers, but random access will be horribly inefficient.
    *
-   * If `failOnDataLoss` is `false`, it will try to get the earliest record in
-   * `[offset, untilOffset)` when some illegal state happens. Otherwise, an `IllegalStateException`
-   * will be thrown.
+   * When `failOnDataLoss` is `true`, this will either return record at offset if available, or
+   * throw exception.
    *
-   * It returns `null` only when `failOnDataLoss` is `false` and it cannot fetch any record between
-   * [offset, untilOffset).
+   * When `failOnDataLoss` is `false`, this will either return record at offset if available, or
+   * return the next earliest available record less than untilOffset, or null. It will not throw
+   * any exception.
    */
   def get(
       offset: Long,
       untilOffset: Long,
       pollTimeoutMs: Long,
       failOnDataLoss: Boolean): ConsumerRecord[Array[Byte], Array[Byte]] = {
-    require(offset < untilOffset, s"offset: $offset, untilOffset: $untilOffset")
+    require(offset < untilOffset,
+      s"offset must always be less than untilOffset [offset: $offset, untilOffset: $untilOffset]")
     logDebug(s"Get $groupId $topicPartition nextOffset $nextOffsetInFetchedData requested $offset")
-    try {
-      fetchDataIfNeeded(offset, pollTimeoutMs)
-      getRecordFromFetchedData(offset, untilOffset, failOnDataLoss)
-    } catch {
-      case e: OffsetOutOfRangeException =>
-        val message =
-          if (failOnDataLoss) {
-            s"""Cannot fetch offset $offset. (GroupId: $groupId, TopicPartition: $topicPartition).
-               | $INSTRUCTION_FOR_FAIL_ON_DATA_LOSS_TRUE""".stripMargin
-          } else {
-            s"""Cannot fetch offset $offset. Some data may be lost. Recovering from the earliest
-               | offset (GroupId: $groupId, TopicPartition: $topicPartition).""".stripMargin
-          }
-        reportDataLoss(failOnDataLoss, message, e)
-        advanceToEarliestOffsetAndFetch(offset, untilOffset, pollTimeoutMs)
+    var toFetchOffset = offset
+    while (toFetchOffset != UNKNOWN_OFFSET) {
+      try {
+        val record = fetchData(toFetchOffset, untilOffset, pollTimeoutMs, failOnDataLoss)
+        if (record == null) {
+          reset()
+        }
+        return record
+      } catch {
+        case e: OffsetOutOfRangeException =>
+          val message =
+            if (failOnDataLoss) {
+              s"Cannot fetch offset $toFetchOffset"
+            } else {
+              s"Cannot fetch offset $toFetchOffset. Some data may be lost. " +
+                "Recovering from the earliest offset"
+            }
+          reportDataLoss(failOnDataLoss, message, e)
+          toFetchOffset = getNextEarliestOffset(toFetchOffset, untilOffset)
+      }
     }
+    reset()
+    null
   }
 
   /**
-   * Check the pre-fetched data with `offset` and try to fetch from Kafka if they don't match.
+   * Return the next earliest available offset in [offset, untilOffset). If all offsets in
+   * [offset, untilOffset) are invalid (e.g., the topic is deleted and recreated), it will return
+   * `UNKNOWN_OFFSET`.
    */
-  private def fetchDataIfNeeded(offset: Long, pollTimeoutMs: Long): Unit = {
-    if (offset != nextOffsetInFetchedData) {
-      logInfo(s"Initial fetch for $topicPartition $offset")
-      seek(offset)
-      poll(pollTimeoutMs)
-    } else if (!fetchedData.hasNext()) {
-      // The last pre-fetched data has been drained.
-      // Seek to the offset because we may call seekToBeginning or seekToEnd before this.
-      seek(offset)
-      poll(pollTimeoutMs)
-    }
-  }
-
-  /**
-   * Try to advance to the beginning offset and fetch again. `earliestOffset` should be in
-   * `[offset, untilOffset]`. If not, it will try to fetch `offset` again if it's in
-   * `[earliestOffset, latestOffset)`. Otherwise, it will return null and reset the pre-fetched
-   * data.
-   */
-  private def advanceToEarliestOffsetAndFetch(
-      offset: Long,
-      untilOffset: Long,
-      pollTimeoutMs: Long): ConsumerRecord[Array[Byte], Array[Byte]] = {
-    val (earliestOffset, latestOffset) = getCurrentOffsetRange()
+  private def getNextEarliestOffset(offset: Long, untilOffset: Long): Long = {
+    val (earliestOffset, latestOffset) = getAvailableOffsetRange()
     if (offset >= latestOffset || earliestOffset >= untilOffset) {
       // [offset, untilOffset) and [earliestOffset, latestOffset) have no overlap,
       // either
@@ -133,12 +121,10 @@ private[kafka010] case class CachedKafkaConsumer private(
         s"""
           |The current available offset range is [$earliestOffset, $latestOffset).
           | Offset ${offset} is out of range, and records in [$offset, $untilOffset) will be
-          | skipped (GroupId: $groupId, TopicPartition: $topicPartition).
-          | $INSTRUCTION_FOR_FAIL_ON_DATA_LOSS_FALSE
+          | skipped ${additionalMessage(failOnDataLoss = false)}
         """.stripMargin
       logWarning(warningMessage)
-      reset()
-      null
+      UNKNOWN_OFFSET
     } else if (offset >= earliestOffset) {
       // -----------------------------------------------------------------------------
       //         ^            ^                  ^                                 ^
@@ -147,13 +133,10 @@ private[kafka010] case class CachedKafkaConsumer private(
       //
       // This will happen when a topic is deleted and recreated, and new data are pushed very fast,
       // then we will see `offset` disappears first then appears again. Although the parameters
-      // are same, the state in Kafka cluster is changed, so it's not an endless loop.
-      //
-      // In addition, the stack here won't be deep unless the user keeps deleting and creating the
-      // topic very fast.
-      //
-      // Therefore, this recursive call is safe.
-      get(offset, untilOffset, pollTimeoutMs, failOnDataLoss = false)
+      // are same, the state in Kafka cluster is changed, so the outer loop won't be endless.
+      logWarning(s"Found a disappeared offset $offset. " +
+        s"Some data may be lost ${additionalMessage(failOnDataLoss = false)}")
+      offset
     } else {
       // ------------------------------------------------------------------------------
       //      ^           ^                       ^                                 ^
@@ -163,23 +146,29 @@ private[kafka010] case class CachedKafkaConsumer private(
         s"""
            |The current available offset range is [$earliestOffset, $latestOffset).
            | Offset ${offset} is out of range, and records in [$offset, $earliestOffset) will be
-           | skipped (GroupId: $groupId, TopicPartition: $topicPartition).
-           | $INSTRUCTION_FOR_FAIL_ON_DATA_LOSS_FALSE
+           | skipped ${additionalMessage(failOnDataLoss = false)}
         """.stripMargin
       logWarning(warningMessage)
-      get(earliestOffset, untilOffset, pollTimeoutMs, failOnDataLoss = false)
+      earliestOffset
     }
   }
 
   /**
-   * Get the earliest record in [offset, untilOffset) from the fetched data. If there is no such
-   * record, return null. Note that this must be called after some data has already been fetched
-   * using poll.
+   * Get the earliest record in [offset, untilOffset). If there is not such record, return null and
+   * clear the fetched data.
    */
-  private def getRecordFromFetchedData(
+  private def fetchData(
       offset: Long,
       untilOffset: Long,
+      pollTimeoutMs: Long,
       failOnDataLoss: Boolean): ConsumerRecord[Array[Byte], Array[Byte]] = {
+    if (offset != nextOffsetInFetchedData || !fetchedData.hasNext()) {
+      // This is the first fetch, or the last pre-fetched data has been drained.
+      // Seek to the offset because we may call seekToBeginning or seekToEnd before this.
+      seek(offset)
+      poll(pollTimeoutMs)
+    }
+
     if (!fetchedData.hasNext()) {
       // We cannot fetch anything after `poll`. Two possible cases:
       // - `earliestOffset` is `offset` but there is nothing for `earliestOffset` right now.
@@ -188,61 +177,20 @@ private[kafka010] case class CachedKafkaConsumer private(
       // partition.
       val message =
         if (failOnDataLoss) {
-          s"""
-             |Cannot fetch record for offset ${offset}
-             | (GroupId: $groupId, TopicPartition: $topicPartition).
-             | $INSTRUCTION_FOR_FAIL_ON_DATA_LOSS_TRUE""".stripMargin
+          s"Cannot fetch record for offset $offset"
         } else {
-          s"""
-             |Cannot fetch record for offset ${offset}. Records in [$offset, $untilOffset) will be
-             | skipped (GroupId: $groupId, TopicPartition: $topicPartition).
-             | $INSTRUCTION_FOR_FAIL_ON_DATA_LOSS_FALSE""".stripMargin
+          s"Cannot fetch record for offset $offset. " +
+            s"Records in [$offset, $untilOffset) will be skipped"
         }
       reportDataLoss(failOnDataLoss, message)
-      reset()
       null
     } else {
       val record = fetchedData.next()
-      if (record.offset >= untilOffset) {
-        // This may happen when records are aged out.
-        val message =
-          if (failOnDataLoss) {
-            s"""
-               |The offset range is [$offset, $untilOffset), but the fetched record offset
-               | ${record.offset} is out of range
-               | (GroupId: $groupId, TopicPartition: $topicPartition).
-               | $INSTRUCTION_FOR_FAIL_ON_DATA_LOSS_TRUE""".stripMargin
-          } else {
-            s"""
-               |The fetched record offset in fetched data ${record.offset} is out of range. Records
-               | in [$offset, $untilOffset) will be skipped
-               | (GroupId: $groupId, TopicPartition: $topicPartition).
-               | $INSTRUCTION_FOR_FAIL_ON_DATA_LOSS_FALSE""".stripMargin
-          }
-        reportDataLoss(failOnDataLoss, message)
-        reset()
-        null
-      } else {
-        if (record.offset != offset) {
-          // This may happen when records are aged out.
-          val message =
-            if (failOnDataLoss) {
-              s"""
-                 |The fetched record offset ${record.offset()} is not the requested offset $offset
-                 | (GroupId: $groupId, TopicPartition: $topicPartition).
-                 | $INSTRUCTION_FOR_FAIL_ON_DATA_LOSS_TRUE""".stripMargin
-            } else {
-              s"""
-                 |The fetched record offset ${record.offset()} is not the requested offset $offset.
-                 | Records in [$offset, ${record.offset()}) will be skipped
-                 | (GroupId: $groupId, TopicPartition: $topicPartition).
-                 | $INSTRUCTION_FOR_FAIL_ON_DATA_LOSS_FALSE""".stripMargin
-            }
-          reportDataLoss(failOnDataLoss, message)
-        }
-        nextOffsetInFetchedData = record.offset + 1
-        record
-      }
+      nextOffsetInFetchedData = record.offset + 1
+      // `seek` is always called before "poll". So "record.offset" must be same as "offset".
+      assert(record.offset == offset,
+        s"The fetched data has a different offset: expected $offset but was ${record.offset}")
+      record
     }
   }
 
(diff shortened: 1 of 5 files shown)


external/kafka-0-10-sql/src/test/scala/org/apache/spark/sql/kafka010/KafkaSourceSuite.scala
```diff
@@ -626,3 +776,106 @@ class KafkaSourceStressSuite extends KafkaSourceTest {
...
+      override def process(value: Int): Unit = {
+        // Slow down the processing speed so that messages may be aged out.
+        Thread.sleep(Random.nextInt(100))
+      }
+
+      override def close(errorOrNull: Throwable): Unit = {
+      }
+    }).start()
+
+    val testTime = 1.minutes
```
zsxwing: I changed this to 20 minutes and test it locally. It passed.

zsxwing:
Use https://github.com/apache/spark/pull/15820/files?w=1 to review PR to ignore space changes.

zsxwing:
cc @tdas @marmbrus @koeninger

external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/CachedKafkaConsumer.scala
```diff
@@ -83,6 +86,113 @@ private[kafka010] case class CachedKafkaConsumer private(
...
+        logWarning(s"Buffer miss for $groupId $topicPartition [$offset, $untilOffset)")
+        reset()
+        return null
+      } else {
+        if (record.offset != offset) {
+          // Case 1
+          logWarning(s"Buffer miss for $groupId $topicPartition [$offset, ${record.offset})")
+        }
+        nextOffsetInFetchedData = record.offset + 1
+        return record
```
koeninger: Are these two early returns actually necessary?

external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/CachedKafkaConsumer.scala
```diff
@@ -83,6 +86,113 @@ private[kafka010] case class CachedKafkaConsumer private(
...
+      }
+    }
+  }
+
+  private def reset(): Unit = {
+    nextOffsetInFetchedData = -2
```
koeninger: This use of -2 as a magic number here and earlier in the file is a little misleading, since the new consumer won't actually let you seek to -2 as a means of indicating earliest

external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/CachedKafkaConsumer.scala
```diff
@@ -83,6 +86,113 @@ private[kafka010] case class CachedKafkaConsumer private(
...
+        val latestOffset = getLatestOffset()
+        if (latestOffset <= offset) {
+          // Case 3 or 6
+          logWarning(s"Offset ${offset} is later than the latest offset $latestOffset. " +
+            s"Skipped [$offset, $untilOffset)")
+          reset()
+          return null
+        } else {
+          // Case 4 or 5
+          getAndIgnoreLostData(offset, untilOffset, pollTimeoutMs)
```
koeninger: - Why is this not an early return?
- The arguments to the recursive function have not changed at this point, right?  Why does it terminate?

external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/CachedKafkaConsumer.scala
```diff
@@ -83,6 +86,113 @@ private[kafka010] case class CachedKafkaConsumer private(
...
+    // scalastyle:on
+    if (offset >= untilOffset) {
+      // Case 2 or 8
+      // We seek to beginningOffset but beginningOffset >= untilOffset
+      reset()
+      return null
+    }
+
+    logDebug(s"Get $groupId $topicPartition nextOffset $nextOffsetInFetchedData requested $offset")
+    var outOfOffset = false
```
koeninger: Can't this var be eliminated with a singly try around the following if/else?

external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/CachedKafkaConsumer.scala
```diff
@@ -83,6 +86,113 @@ private[kafka010] case class CachedKafkaConsumer private(
...
+      try {
+        poll(pollTimeoutMs)
+      } catch {
+        case e: OffsetOutOfRangeException =>
+          logWarning(s"Cannot fetch offset $offset, try to recover from the beginning offset", e)
+          outOfOffset = true
+      }
+    } else if (!fetchedData.hasNext()) {
+      // The last pre-fetched data has been drained.
+      seek(offset)
```
koeninger: I don't think it's necessary to seek every time the fetched data is empty, in normal operation the poll should return the next offset, right?

external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/CachedKafkaConsumer.scala
koeninger: The use of -2 as a magic number here and elsewhere in the file is misleading, since the new consumer won't actually let you seek to -2 as an indicator of earliest.

external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/CachedKafkaConsumer.scala
koeninger: Can this var be eliminated by just using a single try around the if / else?  It's the same catch condition in either case

external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/CachedKafkaConsumer.scala
koeninger: - Why isn't this an early return?
- Unless I'm misreading, this is a recursive call without changing the arguments.  Why is it guaranteed to terminate?

koeninger (commented):
In general, getAndIgnoreLostData is hard to read due to the length and early returns, and I'm pretty sure I've missed something.  
I know the actual underlying cases are complicated to get right, but is it possible to refactor it?

koeninger:
Wow, looks like the new github comment interface did all kinds of weird things, apologies about that.

zsxwing:
@koeninger Thanks for reviewing. Refactored the codes to avoid using early returns and addressed your comments.

external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/CachedKafkaConsumer.scala
```diff
@@ -83,6 +86,129 @@ private[kafka010] case class CachedKafkaConsumer private(
...
+      val latestOffset = getLatestOffset()
+      if (latestOffset <= offset) {
+        // beginningOffset <= latestOffset - 1 < offset <= untilOffset - 1
+        logWarning(s"Offset ${offset} is later than the latest offset $latestOffset. " +
+          s"Skipped [$offset, $untilOffset)")
+        reset()
+        null
+      } else {
+        // beginningOffset <= offset <= min(latestOffset - 1, untilOffset - 1)
+        getAndIgnoreLostData(offset, untilOffset, pollTimeoutMs)
```
koeninger: I'm clearer on why this terminates, but I think it's worth a comment, since it's a mutually recursive call without changing arguments.

external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/CachedKafkaConsumer.scala
```diff
@@ -83,6 +86,129 @@ private[kafka010] case class CachedKafkaConsumer private(
...
+        seek(offset)
+        poll(pollTimeoutMs)
+      } else if (!fetchedData.hasNext()) {
+        // The last pre-fetched data has been drained.
+        poll(pollTimeoutMs)
+      }
+      getRecordFromFetchedData(offset, untilOffset)
+    } catch {
+      case e: OffsetOutOfRangeException =>
+        logWarning(s"Cannot fetch offset $offset, try to recover from the beginning offset", e)
```
koeninger: I think it's worth the warning explicitly stating that data has been lost

external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/CachedKafkaConsumer.scala
```diff
@@ -83,6 +86,129 @@ private[kafka010] case class CachedKafkaConsumer private(
...
+   * Get the earliest record in [offset, untilOffset) from the fetched data. If there is no such
+   * record, returns null. Must be called after `poll`.
+   */
+  private def getRecordFromFetchedData(
+      offset: Long,
+      untilOffset: Long): ConsumerRecord[Array[Byte], Array[Byte]] = {
+    if (!fetchedData.hasNext()) {
+      // We cannot fetch anything after `poll`. Two possible cases:
+      // - `beginningOffset` is `offset` but there is nothing for `beginningOffset` right now.
+      // - Cannot fetch any date before timeout.
```
koeninger: As a user, I'm not sure that setting failOnDataLoss=false would make me know that a timeout would cause me to miss data in my spark job (that might otherwise still be in kafka)

koeninger (commented):
Thanks, I think that version is easier to read, and hopefully in normal operation won't be recursing much so the @tailrec loss wont be an efficiency hit.

I'm mostly concerned about clarifying the timeout situation at this point.

external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/CachedKafkaConsumer.scala
marmbrus: +1

external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/CachedKafkaConsumer.scala
huitseeker: +1
Especially since with the loss of the use of `@tailrec`, this must now prove it will terminate within a limited stack size, and should prove it will under most stack size configurations.

external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/CachedKafkaConsumer.scala
zsxwing: If throwing an exception here, `test("access offset 0 in Spark job but the topic has been deleted")` will fail. It seems a reasonable case.

external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/CachedKafkaConsumer.scala
```diff
@@ -83,6 +86,139 @@ private[kafka010] case class CachedKafkaConsumer private(
...
+  /**
+   * Get the record at the `offset`. If it doesn't exist, try to get the earliest record in
+   * `[offset, untilOffset)`.
+   */
+  def getAndIgnoreLostData(
+      offset: Long,
+      untilOffset: Long,
+      pollTimeoutMs: Long): ConsumerRecord[Array[Byte], Array[Byte]] = {
+    // scalastyle:off
+    // When `failOnDataLoss` is `false`, we need to handle the following cases (note: untilOffset and latestOffset are exclusive):
```
tdas: I dont get this, there is no reference to `latestOffset` in this method.

zsxwing: `latestOffset` means the latest offset that we get from the Kafka.

tdas: Also is beginningOffset the earlier offset available in Kafka? Then lets rename this as earliestOffset, and "earliest" and "latest" and more popular terms in the context of Kafka

zsxwing: Yes. Will update the terms.

tdas (changes requested):
Two main high level points
1. Should not have separate code paths for failOnDataLoss = true/false.
2. Merge the if conditions in getRecordFromFetchedData to reduce it from 4 cases to 3.
3. Needs better inline documentation as the code is not easy to understand. Sometimes better to document using ascii art ;)

external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/CachedKafkaConsumer.scala
koeninger: I don't think topic deletion is anywhere near as common as timeouts, and topic deletion is something user initiated.  As a user, is skipping the rest of the offsets in a timeout really what you would want to happen (it isn't for me)?  If so, does this need a separate configuration?

external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/CachedKafkaConsumer.scala
```diff
@@ -47,40 +51,229 @@ private[kafka010] case class CachedKafkaConsumer private(
...
+        }
+      reportDataLoss(failOnDataLoss, message)
+      reset()
+      null
+    } else {
+      val record = fetchedData.next()
+      if (record.offset >= untilOffset) {
+        // This may happen when records are aged out.
+        val message =
+          if (failOnDataLoss) {
```
tdas: There is ALREADY a if statement in reportDataLoss on failOnDataLoss. having another one, looks too complicated. i think best to not have an if statement here. and same for other places. this too much complication for documentation.

tdas (changes requested):
As we spoke offline, this can further simplified. I am posting my detailed comments anyways, all of them are not applicable if you are going to refactor the code.

external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/CachedKafkaConsumer.scala
```diff
@@ -47,40 +51,191 @@ private[kafka010] case class CachedKafkaConsumer private(
...
-    assert(fetchedData.hasNext(),
-      s"Failed to get records for $groupId $topicPartition $offset " +
-        s"after polling for $pollTimeoutMs")
-    var record = fetchedData.next()
+  /**
+   * Return the next earliest available offset in [offset, untilOffset). If all offsets in
+   * [offset, untilOffset) are invalid (e.g., the topic is deleted and recreated), it will return
+   * `UNKNOWN_OFFSET`.
+   */
+  private def getNextEarliestOffset(offset: Long, untilOffset: Long): Long = {
```
tdas: nit: `getEarliestAvailableOffsetBetween(...)`

external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/CachedKafkaConsumer.scala
```diff
@@ -47,40 +51,191 @@ private[kafka010] case class CachedKafkaConsumer private(
 
   /** Iterator to the already fetch data */
   private var fetchedData = ju.Collections.emptyIterator[ConsumerRecord[Array[Byte], Array[Byte]]]
-  private var nextOffsetInFetchedData = -2L
+  private var nextOffsetInFetchedData = UNKNOWN_OFFSET
 
   /**
    * Get the record for the given offset, waiting up to timeout ms if IO is necessary.
```
tdas: nit: update docs to clarify earlier that this may not return offset 
Get the record for the given offset if available. Otherwise it will either throw error (if failOnDataLoss = true), or return the next available offset within [offset, untilOffset].

Use @param to explain pollTimeoutMs and others in more detail

external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/CachedKafkaConsumer.scala
```diff
@@ -47,40 +51,191 @@ private[kafka010] case class CachedKafkaConsumer private(
...
 
-    if (record.offset != offset) {
-      logInfo(s"Buffer miss for $groupId $topicPartition $offset")
+  /**
+   * Get the earliest record in [offset, untilOffset). If there is not such record, return null and
+   * clear the fetched data.
+   */
+  private def fetchData(
+      offset: Long,
+      untilOffset: Long,
```
tdas: untilOffset and failOnDataLoss not really needed.

external/kafka-0-10-sql/src/test/scala/org/apache/spark/sql/kafka010/KafkaSourceSuite.scala
```diff
@@ -626,3 +776,106 @@ class KafkaSourceStressSuite extends KafkaSourceTest {
...
+    val startTime = System.currentTimeMillis()
+    val topics = mutable.ArrayBuffer[String]()
+    while (System.currentTimeMillis() - testTime.toMillis < startTime) {
+      Random.nextInt(6) match {
+        case 0 =>
+          val topic = newTopic()
+          topics += topic
+          testUtils.createTopic(topic, partitions = 1)
+        case 1 =>
```
tdas: Update test to recreate same topics.

external/kafka-0-10-sql/src/test/scala/org/apache/spark/sql/kafka010/KafkaSourceSuite.scala
```diff
@@ -223,52 +228,184 @@ class KafkaSourceSuite extends KafkaSourceTest {
...
+      // the Kafka consumer will just block until timeout and return an empty result.
+      // So set the timeout to 1 second to make this test fast.
+      .option("kafkaConsumer.pollTimeoutMs", "1000")
+      .option("startingOffsets", "earliest")
+      .option("failOnDataLoss", "false")
+    val kafka = reader.load()
+      .selectExpr("CAST(key AS STRING)", "CAST(value AS STRING)")
+      .as[(String, String)]
+    KafkaSourceSuite.globalTestUtils = testUtils
+    val query = kafka.map(kv => kv._2.toInt).writeStream.foreach(new ForeachWriter[Int] {
```
tdas: add explanation.

external/kafka-0-10-sql/src/test/scala/org/apache/spark/sql/kafka010/KafkaSourceSuite.scala
```diff
@@ -223,52 +228,184 @@ class KafkaSourceSuite extends KafkaSourceTest {
...
+    }).start()
+    query.processAllAvailable()
+    assert(KafkaSourceSuite.collectedData.isEmpty)
+    query.stop()
+    // `failOnDataLoss` is `false`, we should not fail the query
+    assert(query.exception.isEmpty)
   }
 
-  test("subscribing topic by name from specific offsets") {
+  test("access non-zero offset in Spark job but the topic has been deleted") {
```
tdas: remove this test.

external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/CachedKafkaConsumer.scala
```diff
@@ -47,40 +51,190 @@ private[kafka010] case class CachedKafkaConsumer private(
...
+      failOnDataLoss: Boolean): ConsumerRecord[Array[Byte], Array[Byte]] = {
+    require(offset < untilOffset,
+      s"offset must always be less than untilOffset [offset: $offset, untilOffset: $untilOffset]")
     logDebug(s"Get $groupId $topicPartition nextOffset $nextOffsetInFetchedData requested $offset")
-    if (offset != nextOffsetInFetchedData) {
-      logInfo(s"Initial fetch for $topicPartition $offset")
-      seek(offset)
-      poll(pollTimeoutMs)
+    var toFetchOffset = offset
+    while (toFetchOffset != UNKNOWN_OFFSET) {
```
tdas: add docs for this logic.

koeninger:
Because the comment made by me and +1'ed by marmbrus is hidden at this point, I just want to re-iterate that this patch should not skip the rest of the partition in the case that a timeout happens.

external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/CachedKafkaConsumer.scala
```diff
@@ -34,53 +38,211 @@ import org.apache.spark.internal.Logging
...
-      record = fetchedData.next()
+    }
+
+    if (!fetchedData.hasNext()) {
+      // We cannot fetch anything after `poll`. Two possible cases:
+      // - `offset` is out of range so that Kafka returns nothing. Just throw
+      // `OffsetOutOfRangeException` to let the caller handle it.
+      // - Cannot fetch any data before timeout. TimeoutException will be thrown.
+      val (earliestOffset, latestOffset) = getAvailableOffsetRange()
+      if (offset < earliestOffset || offset >= latestOffset) {
```
zsxwing: @koeninger Just updated the timeout logic. It will check the current available offset range and use it to distinguish these two cases.

external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/CachedKafkaConsumer.scala
```diff
@@ -34,53 +38,211 @@ import org.apache.spark.internal.Logging
...
+      failOnDataLoss: Boolean): ConsumerRecord[Array[Byte], Array[Byte]] = {
+    require(offset < untilOffset,
+      s"offset must always be less than untilOffset [offset: $offset, untilOffset: $untilOffset]")
     logDebug(s"Get $groupId $topicPartition nextOffset $nextOffsetInFetchedData requested $offset")
-    if (offset != nextOffsetInFetchedData) {
-      logInfo(s"Initial fetch for $topicPartition $offset")
-      seek(offset)
-      poll(pollTimeoutMs)
+    // The following loop is basically for `failOnDataLoss = false`. When `failOnDataLoss` is
+    // `false`, firstly, we will try to fetch the record at `offset`. If no such record, then we
```
tdas: nit: 
firstly --> first,
no such record exists

overall +1, thanks for this explanation.

external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/CachedKafkaConsumer.scala
```diff
@@ -34,53 +38,211 @@ import org.apache.spark.internal.Logging
...
+   * @param untilOffset the max offset to fetch. Exclusive.
+   * @param pollTimeoutMs timeout in milliseconds to poll data from Kafka.
+   * @param failOnDataLoss When `failOnDataLoss` is `true`, this method will either return record at
+   *                       offset if available, or throw exception.when `failOnDataLoss` is `false`,
+   *                       this method will either return record at offset if available, or return
+   *                       the next earliest available record less than untilOffset, or null. It
+   *                       will not throw any exception.
    */
-  def get(offset: Long, pollTimeoutMs: Long): ConsumerRecord[Array[Byte], Array[Byte]] = {
+  def get(
```
tdas: Can you also document that it can return null.

external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/CachedKafkaConsumer.scala
```diff
@@ -34,53 +38,211 @@ import org.apache.spark.internal.Logging
...
+    // If `failOnDataLoss` is `true`, the loop body will be executed only once.
+    var toFetchOffset = offset
+    while (toFetchOffset != UNKNOWN_OFFSET) {
+      try {
+        return fetchData(toFetchOffset, pollTimeoutMs)
+      } catch {
+        case e: OffsetOutOfRangeException =>
+          // When there is some error thrown, it's better to use a new consumer to drop all cached
+          // states in the old consumer. We don't need to worry about the performance because this
+          // is not a normal path.
```
tdas: normal --> common

external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/KafkaSource.scala
```diff
@@ -458,6 +457,21 @@ private[kafka010] case class KafkaSource(
 /** Companion object for the [[KafkaSource]]. */
 private[kafka010] object KafkaSource {
 
+  val INSTRUCTION_FOR_FAIL_ON_DATA_LOSS_FALSE =
+    """
+      |There may have been some data loss because some data may have been aged out in Kafka or
+      | the topic has been deleted and is therefore unavailable for processing. If you want your
```
tdas: nit: better grammar

Some data may have been lost because they are not available in Kafka any more; either the data was aged out by Kafka or the topic may have been deleted before all the data in the topic was processed. 

Similarly change below.

external/kafka-0-10-sql/src/test/scala/org/apache/spark/sql/kafka010/KafkaSourceSuite.scala
```diff
@@ -413,13 +451,59 @@ class KafkaSourceSuite extends KafkaSourceTest {
     )
   }
 
+  test("Delete a topic when a Spark job is running") {
```
tdas: nit: D -> d, for consistency.

tdas (commented):
LGTM, just a few documentation improvements. This is much cleaner and easier to understand now.

zsxwing:
@tdas I did some changes to make the stress test stable and ran it for 20 minutes without errors. I also confirmed the warning logs did appear in the unit test logs.

srowen:
@zsxwing should the title include SPARK-18529 per your comment on the JIRA?

zsxwing:
> @zsxwing should the title include SPARK-18529 per your comment on the JIRA?

Thanks for pointing out. Updated.

tdas:
LGTM. Merging this in master and 2.1

PR #15823, opened 2016-11-09
[SPARK-18191][CORE][FOLLOWUP] Call `setConf` if `OutputFormat` is `Configurable`.

jiangxb1987:

We should call `setConf` if `OutputFormat` is `Configurable`, this should be done before we create `OutputCommitter` and `RecordWriter`.
This is follow up of #15769, see discussion [here](https://github.com/apache/spark/pull/15769/files#r87064229)

## How was this patch tested?

Add test of this case in `PairRDDFunctionsSuite`.


jiangxb1987:
cc @rxin @mridulm

rxin:
LGTM otherwise.

mridulm:
LGTM. Not at my laptop, would be great if you can merge @rxin, thanks.

rxin:
Merging in master.

PR #15825, opened 2016-11-09
[SPARK-18377][SQL] warehouse path should be a static conf

cloud-fan:

it's weird that every session can set its own warehouse path at runtime, we should forbid it and make it a static conf.

## How was this patch tested?

existing tests.

### DIFF
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/SessionCatalog.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/SessionCatalog.scala
@@ -83,7 +83,7 @@ class SessionCatalog(
   // check whether the temporary table or function exists, then, if not, operate on
   // the corresponding item in the current database.
   @GuardedBy("this")
-  protected var currentDb = DEFAULT_DATABASE
+  protected var currentDb = formatDatabaseName(DEFAULT_DATABASE)
 
   /**
    * Format table name, taking into account case sensitivity.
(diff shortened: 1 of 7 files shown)


cloud-fan:
CC @yhuai @rxin @srown @gatorsmile

cloud-fan:
retest this please

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/SessionCatalog.scala
```diff
@@ -83,14 +83,7 @@ class SessionCatalog(
...
   // the corresponding item in the current database.
   @GuardedBy("this")
-  protected var currentDb = {
-    val defaultName = DEFAULT_DATABASE
-    val defaultDbDefinition =
-      CatalogDatabase(defaultName, "default database", conf.warehousePath, Map())
-    // Initialize default database if it doesn't already exist
-    createDatabase(defaultDbDefinition, ignoreIfExists = true)
-    formatDatabaseName(defaultName)
-  }
```
gatorsmile: To the other reviewers: this is moved to `SharedState.scala`. The default database is created once and shared by all the sessions.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/SessionCatalog.scala
```diff
@@ -83,14 +83,7 @@ class SessionCatalog(
...
+  protected var currentDb = DEFAULT_DATABASE
```
gatorsmile: It might be safer if we call `formatDatabaseName` like what we did before.

```Scala
protected var currentDb = formatDatabaseName (DEFAULT_DATABASE)
```

gatorsmile:
Regarding the code changes, LGTM except a couple of minor comments. 

BTW, this introduces reasonable external behavior changes. Maybe we need to document it in the release of Spark 2.1

cloud-fan:
retest this please

rxin:
Merging in master/branch-2.1.

cloud-fan:
I added a `releasenotes` label to the ticket, so that we won't forget

PR #15826, opened 2016-11-09
[SPARK-14077][ML][FOLLOW-UP] Minor refactor and cleanup for NaiveBayes

yanboliang:
* Refactor out ```trainWithLabelCheck``` and make ```mllib.NaiveBayes``` call into it.
* Avoid capturing the outer object for ```modelType```.
* Move ```requireNonnegativeValues``` and ```requireZeroOneBernoulliValues``` to companion object.

## How was this patch tested?
Existing tests.

### DIFF
--- a/mllib/src/main/scala/org/apache/spark/ml/classification/NaiveBayes.scala
+++ b/mllib/src/main/scala/org/apache/spark/ml/classification/NaiveBayes.scala
@@ -206,15 +206,15 @@ class NaiveBayes @Since("1.5.0") (
 @Since("1.6.0")
 object NaiveBayes extends DefaultParamsReadable[NaiveBayes] {
   /** String name for multinomial model type. */
-  private[spark] val Multinomial: String = "multinomial"
+  private[classification] val Multinomial: String = "multinomial"
 
   /** String name for Bernoulli model type. */
-  private[spark] val Bernoulli: String = "bernoulli"
+  private[classification] val Bernoulli: String = "bernoulli"
 
   /* Set of modelTypes that NaiveBayes supports */
-  private[spark] val supportedModelTypes = Set(Multinomial, Bernoulli)
+  private[classification] val supportedModelTypes = Set(Multinomial, Bernoulli)
 
-  def requireNonnegativeValues(v: Vector): Unit = {
+  private[NaiveBayes] val requireNonnegativeValues: Vector => Unit = (v: Vector) => {
     val values = v match {
       case sv: SparseVector => sv.values
       case dv: DenseVector => dv.values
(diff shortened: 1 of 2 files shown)


mllib/src/main/scala/org/apache/spark/ml/classification/NaiveBayes.scala
```diff
@@ -226,13 +206,33 @@ class NaiveBayes @Since("1.5.0") (
...
 
   /** String name for Bernoulli model type. */
-  private[spark] val Bernoulli: String = "bernoulli"
+  private[classification] val Bernoulli: String = "bernoulli"
 
   /* Set of modelTypes that NaiveBayes supports */
-  private[spark] val supportedModelTypes = Set(Multinomial, Bernoulli)
+  private[classification] val supportedModelTypes = Set(Multinomial, Bernoulli)
+
+  private[NaiveBayes] val requireNonnegativeValues: Vector => Unit = (v: Vector) => {
```
yanboliang: @thunterdb I define these two functions still by ```val``` which should be more appropriate, since there is difference between ```val``` and ```def``` to define a function. http://stackoverflow.com/questions/18887264/what-is-the-difference-between-def-and-val-to-define-a-function

yanboliang:
cc @thunterdb @zhengruifeng

mllib/src/main/scala/org/apache/spark/ml/classification/NaiveBayes.scala
thunterdb: @yanboliang indeed it makes a difference when defining methods inside another method. In the current case, you brought the function as a value of the companion object. You can safely turn it into a method

mllib/src/main/scala/org/apache/spark/ml/classification/NaiveBayes.scala
```diff
@@ -226,13 +206,33 @@ class NaiveBayes @Since("1.5.0") (
...
+    val values = v match {
+      case sv: SparseVector => sv.values
+      case dv: DenseVector => dv.values
+    }
+
+    require(values.forall(_ >= 0.0),
+      s"Naive Bayes requires nonnegative feature values but found $v.")
+  }
+
+  private[NaiveBayes] val requireZeroOneBernoulliValues: Vector => Unit = (v: Vector) => {
```
thunterdb: same thing here

mllib/src/main/scala/org/apache/spark/ml/classification/NaiveBayes.scala
```diff
@@ -110,21 +110,20 @@ class NaiveBayes @Since("1.5.0") (
...
-  private[spark] def setIsML(isML: Boolean): this.type = {
-    this.isML = isML
-    this
-  }
-
-  override protected def train(dataset: Dataset[_]): NaiveBayesModel = {
-    if (isML) {
+  private[spark] def trainWithLabelCheck(
+      dataset: Dataset[_],
+      positiveLabel: Boolean = true): NaiveBayesModel = {
```
thunterdb: Since it is an internal method, I suggest to not use a default argument and specify it explicitly above.

thunterdb (changes requested):
@yanboliang thanks! I have a few small comments.

yanboliang:
@thunterdb Any more comments? Thanks!

thunterdb:
@yanboliang that looks great, thank you. LGTM.

yanboliang:
Merged into master and branch-2.1. Thanks for reviewing.

PR #15827, opened 2016-11-09
[SPARK-18187][STREAMING] CompactibleFileStreamLog should not use "compactInterval" direcly with user setting.

uncleGen:

CompactibleFileStreamLog relys on "compactInterval" to detect a compaction batch. If the "compactInterval" is reset by user, CompactibleFileStreamLog will return wrong answer, resulting data loss. This PR procides a way to check the validity of 'compactInterval', and calculate an appropriate value.

## How was this patch tested?

When restart a stream, we change the 'spark.sql.streaming.fileSource.log.compactInterval' different with the former one.


uncleGen:
cc @zsxwing

zsxwing:
Thanks for doing this. I just merged #15852. Could you close this PR, please?

uncleGen:
@zsxwing OK，close it.

PR #15828, opened 2016-11-09
[SPARK-18187][SS] CompactibleFileStreamLog should not rely on "compactInterval" to detect a compaction batch

lw-lin:
### Problem

Right now `CompactibleFileStreamLog` uses `compactInterval` to check if a batch is a compaction batch. However, since this conf is controlled by the user, they may just change it, and `CompactibleFileStreamLog` will just silently return the wrong answer.

### Changes

With this patch the `compactInterval` configuration should only be used when deciding if we should perform a new compaction. The identification of a compaction vs a delta is done by looking for the compact suffix.

In more details, upon each start/restart, we first examine what metadata files were there, what files among them are compacted files. We introduce `knownCompactionBatches` to hold the existing compaction batches before this run, and `zeroBatch` to hold the first batch this run should write to.

Thus we can support the following situations:

```
  (1) a fresh run
  (2) the previous run with `compactInterval` = 2
      0
  (3) the previous run with `compactInterval` = 2
      0   1.compact
  (4) previous runs with `compactInterval` = 2 and `compactInterval` = 5
      0   1.compact    2   3.compact   4.compact
  (5) previous runs with `compactInterval` = 2 and `compactInterval` = 5
      0   1.compact    2   3.compact   4.compact   5   6   7   8
```
 with:
```
  (1) `knownCompactionBatches` = (), `zeroBatch` = 0
  (2) `knownCompactionBatches` = (), `zeroBatch` = 1
  (3) `knownCompactionBatches` = (1), `zeroBatch` = 2
  (4) `knownCompactionBatches` = (1, 3, 4), `zeroBatch` = 5
  (5) `knownCompactionBatches` = (1, 3, 4), `zeroBatch` = 9
```

Note this appoach does not require the `compactInteral` for a new run must be consistent with previous runs.

## How was this patch tested?

added test cases

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/CompactibleFileStreamLog.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/CompactibleFileStreamLog.scala
@@ -19,6 +19,7 @@ package org.apache.spark.sql.execution.streaming
 
 import java.io.{InputStream, IOException, OutputStream}
 import java.nio.charset.StandardCharsets.UTF_8
+import java.util
 
 import scala.io.{Source => IOSource}
 import scala.reflect.ClassTag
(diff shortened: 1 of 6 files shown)


uncleGen:
related to #15827

lw-lin:
rebased.
- the first & second commits were a simple refactor of the `	CompactibleFileStreamLogSuite.scala`
- **the third commit was the actual change of this patch**

lw-lin:
Jenkins retest this please

PR #15829, opened 2016-11-09
[SPARK-18379][SQL] Make the parallelism of parallelPartitionDiscovery configurable.

uncleGen:

The largest parallelism in PartitioningAwareFileIndex #listLeafFilesInParallel() is 10000 in hard code. We may need to make this number configurable. And in PR, I reduce it to 100.

## How was this patch tested?

Existing ut.

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala
@@ -396,6 +396,13 @@ object SQLConf {
       .intConf
       .createWithDefault(32)
 
+  val PARALLEL_PARTITION_DISCOVERY_PARALLELISM =
+    SQLConfigBuilder("spark.sql.sources.parallelPartitionDiscovery.parallelism")
+      .doc("The number of parallelism to list a collection of path recursively, Set the " +
+        "number to prevent file listing from generating too many tasks.")
+      .intConf
+      .createWithDefault(100)
+
   // Whether to automatically resolve ambiguity in join conditions for self-joins.
   // See SPARK-6231.
   val DATAFRAME_SELF_JOIN_AUTO_RESOLVE_AMBIGUITY =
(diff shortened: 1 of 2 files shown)


sql/core/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala
```diff
@@ -396,6 +396,13 @@ object SQLConf {
       .intConf
       .createWithDefault(32)
 
+  val PARALLEL_PARTITION_DISCOVERY_PARALLELISM =
+    SQLConfigBuilder("spark.sql.sources.parallelPartitionDiscovery.parallelism")
+      .doc("The number of parallelism to list a collection of path recursively, Set the " +
+        "number to prevent file listing from generating too many tasks.")
+      .intConf
+      .createWithDefault(100)
```
HyukjinKwon: Do you mind if I ask the reason to reduce this from 10000 to 100?

uncleGen: IMHO, 10000 is too large for the file discovery in consideration of scene where cluster is not very large, and it makes	 no contributions to improve performance in most scenes.

rxin: cc @liancheng @yhuai 

This should be an internal config.

yhuai: In general, I think we should have more smaller tasks than having a smaller number of tasks that need to handle lots of paths. 

If you have a small cluster and we have a large number of paths, with this change, we only have 100 tasks. Every task will take longer time to finish and other jobs in the cluster may need to wait for a longer time to get scheduled.

yhuai: btw, I think 100 is too small.

uncleGen: Yeah, this should be an internal config, and i will reserve the default 10000 setting.

uncleGen:
cc @yhuai @srowen

srowen (approved):
Looks OK with respect to the comments here

srowen:
@yhuai per request on the JIRA, are you OK with me backporting to 2.1? seems OK as it doesn't modify the default behavior.

srowen:
Merged to 2.1 as well

yhuai:
Sure. Thanks!

PR #15830, opened 2016-11-09
[SPARK-18375][SPARK-18383][BUILD][CORE]Upgrade netty to 4.0.42.Final

witgo:

One of the important changes for 4.0.42.Final is "Support any FileRegion implementation when using epoll transport netty/netty#5825".
In 4.0.42.Final, `MessageWithHeader` can work properly when `spark.[shuffle|rpc].io.mode` is set to epoll

## How was this patch tested?

Existing tests

### DIFF
--- a/core/src/main/scala/org/apache/spark/util/Utils.scala
+++ b/core/src/main/scala/org/apache/spark/util/Utils.scala
@@ -39,6 +39,7 @@ import scala.reflect.ClassTag
 import scala.util.Try
 import scala.util.control.{ControlThrowable, NonFatal}
 
+import _root_.io.netty.channel.unix.Errors.NativeIoException
 import com.google.common.cache.{CacheBuilder, CacheLoader, LoadingCache}
 import com.google.common.io.{ByteStreams, Files => GFiles}
 import com.google.common.net.InetAddresses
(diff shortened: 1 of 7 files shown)


core/src/main/scala/org/apache/spark/util/Utils.scala
```diff
@@ -2222,6 +2223,9 @@ private[spark] object Utils extends Logging {
         isBindCollision(e.getCause)
       case e: MultiException =>
         e.getThrowables.asScala.exists(isBindCollision)
+      case e: NativeIoException =>
+        e.getMessage().startsWith("bind() failed: ") ||
```
srowen: Seems OK. Do you need to check for a null message though, like above? In fact, the condition above for `BindException` could be `e.getMessage != null || isBindCollision(e.getCause)`

witgo: From the netty code, the return of `NativeIoException.getMessage` is not null.
The only way to  create a NativeIoException instance is to call [Errors.newIOException](https://github.com/netty/netty/blob/4.0/transport-native-epoll/src/main/java/io/netty/channel/unix/Errors.java#L116)

witgo: OK, I will modify the code.

srowen (approved):

srowen:
Merged to master/2.1

witgo:
@srowen thank you

PR #15833, opened 2016-11-09
[SPARK-18353][CORE] spark.rpc.askTimeout defalut value is not 120s

srowen:

Avoid hard-coding spark.rpc.askTimeout to non-default in Client; fix doc about spark.rpc.askTimeout default

## How was this patch tested?

Existing tests

### DIFF
--- a/core/src/main/scala/org/apache/spark/deploy/Client.scala
+++ b/core/src/main/scala/org/apache/spark/deploy/Client.scala
@@ -221,7 +221,6 @@ object Client {
     val conf = new SparkConf()
     val driverArgs = new ClientArguments(args)
 
-    conf.set("spark.rpc.askTimeout", "10")
     Logger.getRootLogger.setLevel(driverArgs.logLevel)
 
     val rpcEnv =
--- a/core/src/main/scala/org/apache/spark/deploy/Client.scala
+++ b/core/src/main/scala/org/apache/spark/deploy/Client.scala
@@ -221,6 +221,9 @@ object Client {
     val conf = new SparkConf()
     val driverArgs = new ClientArguments(args)
 
+    if (!conf.contains("spark.rpc.askTimeout")) {
+      conf.set("spark.rpc.askTimeout", "10s")
+    }
     Logger.getRootLogger.setLevel(driverArgs.logLevel)
 
     val rpcEnv =
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1175,7 +1175,7 @@ Apart from these, the following properties are also available, and may be useful
 </tr>
 <tr>
   <td><code>spark.rpc.askTimeout</code></td>
-  <td>120s</td>
+  <td><code>spark.network.timeout</code></td>
   <td>
     Duration for an RPC ask operation to wait before timing out.
   </td>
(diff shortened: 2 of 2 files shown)


core/src/main/scala/org/apache/spark/deploy/Client.scala
```diff
@@ -221,7 +221,6 @@ object Client {
     val conf = new SparkConf()
     val driverArgs = new ClientArguments(args)
 
-    conf.set("spark.rpc.askTimeout", "10")
```
rxin: cc @zsxwing was there a reason this timeout is hard coded here?

rxin: I guess part of the reason is that master/client have low memory usage and as a result unlikely to have long timeouts due to gc.

@srowen I'm not sure if this is a good change.

zsxwing: I don't know. @pwendell added it in https://github.com/apache/spark/commit/3d939e5fe8930368b46eb49659a9cca8f41c8768 and changed the value to 10 in another commit.

srowen: I see, does that mean that the default should be 10s? the real issue is just that the default ends up being "10" instead of advertised 120s.

rxin: This doesn't actually impact the default for Spark apps, but only the standalone cluster doesn't it?

srowen: I think that's right. I suspect that either it's OK to let them use the default timeout of 120s in the end, or, probably bears noting the different practical default of 10s in the docs.

rxin: Yea maybe we should just note it with a comment explaining why it is 10s.

srowen: OK, see my current version, that at least lets spark.rpc.askTimeout be set to something besides 10s in standalone. See https://issues.apache.org/jira/browse/SPARK-18353 discussion too.

core/src/main/scala/org/apache/spark/deploy/Client.scala
```diff
@@ -221,7 +221,9 @@ object Client {
...
+    if (!conf.contains("spark.rpc.askTimeout")) {
```
rxin: can this ever be set? I don't remember whether the standalone master/worker reads config files or not.

andrewor14: also we should probably add this to `config/package.scala` so we don't duplicate it

srowen: OK, to move forward on this: @rxin I think this ends up affecting application launch because this is the bit that runs applications rather than starts workers, if that's what you mean. It will process `--conf` from the user command line but then override it here. The report in the JIRA seems to confirm that. At the least, this change would not affect the current behavior.

@andrewor14 that makes sense, though I note that the other `spark.rpc` configs are not in `config/package.scala`. The `ConfigBuilder` doesn't quite seem able to express a property whose default is another property either. Should I still try to refactor that?

srowen:
Going to merge this because it doesn't change any default behavior, just appears to let this configuration be set by the user in some contexts where not possible before.

srowen:
Merged to master/2.1

PR #15834, opened 2016-11-09
[SPARK-18368] [SQL] Fix regexp replace when serialized

rdblue:

This makes the result value both transient and lazy, so that if the RegExpReplace object is initialized then serialized, `result: StringBuffer` will be correctly initialized.

## How was this patch tested?

* Verified that this patch fixed the query that found the bug.
* Added a test case that fails without the fix.


rdblue:
@yhuai, this replaces #15816. The ref, 	3536f6a, has already passed tests on that PR so it should be safe to merge this.

yhuai:
Awesome! btw looks like your original changes in `ExpressionEvalHelper.scala` (https://github.com/apache/spark/pull/15816/files#diff-41747ec3f56901eb7bfb95d2a217e94d) uncovered issues with other expressions. Want to send a new pr to fix them as well?

yhuai:
image

Since tests have already passed, I am merging this PR to master/branch-2.1/branch-2.0.

rdblue:
[SPARK-18387](https://issues.apache.org/jira/browse/SPARK-18387) tracks the other bugs.

yhuai:
Great. Thanks!

PR #14649, opened 2016-08-15
[SPARK-17059][SQL] Allow FileFormat to specify partition pruning strategy

a10y:
## What changes were proposed in this pull request?

Different FileFormat implementations may be able to make intelligent decisions about files that will need to be processed as part of a FileSourceScanExec based on the `pushedFilters` available. This PR implements a way to do that pluggably, with an implementation for Parquet that allows applying the summary metadata to prevent task creation.

This has a few performance benefits:
1. Reading of files is generally slow, especially for S3. In the current Parquet implementation the summary metadata is not used and so the footers are read directly. This can be very slow for large Parquet datasets, as even as of Hadoop 2.7 S3 reads will read the entire file by default (random reads are configurable only starting on 2.7 onwards, and is disabled by default)
2. Partitions that are found to contain no files can then be pruned out or coalesced depending on the FileFormat's implementation, allowing for fewer tasks being created.
## How was this patch tested?

Existing tests and Spark Shell, plus unit test for the Parquet implementation.

PR #15835, opened 2016-11-09
[SPARK-17059][SQL] Allow FileFormat to specify partition pruning strategy via splits

pwoody:

This is a follow up to changes in #14649 to meet the new codebase changes. It is slightly different in that it will not filter files explicitly, but it allows a FileFormat to prune splits (if applicable). This is implemented in ParquetFileFormat and every other format maintains the same behavior.

## How was this patch tested?

Passing current tests and added two new tests to validate the pruning and ensure that excessive filtering does not occur on malformed metadata.

Thanks!
Closes #14649

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/FileFormat.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/FileFormat.scala
@@ -21,6 +21,7 @@ import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.fs._
 import org.apache.hadoop.io.compress.{CompressionCodecFactory, SplittableCompressionCodec}
 import org.apache.hadoop.mapreduce.Job
+import org.apache.hadoop.mapreduce.lib.input.FileSplit
 
 import org.apache.spark.sql._
 import org.apache.spark.sql.catalyst.InternalRow
@@ -72,6 +73,18 @@ trait FileFormat {
       options: Map[String, String],
       path: Path): Boolean = {
     false
+  }
+
+  /**
+   * Allow a splittable FileFormat to produce a function to split individual files.
+   */
+  def buildSplitter(
+      sparkSession: SparkSession,
+      fileIndex: FileIndex,
+      filters: Seq[Filter],
+      schema: StructType,
+      hadoopConf: Configuration): (FileStatus => Seq[FileSplit]) = {
+    stat => Seq(new FileSplit(stat.getPath, 0, stat.getLen, Array.empty))
   }
 
   /**
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetFileFormat.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetFileFormat.scala
@@ -17,12 +17,15 @@
 
 package org.apache.spark.sql.execution.datasources.parquet
 
+import java.io.FileNotFoundException
 import java.net.URI
+import java.util.concurrent.{Callable, TimeUnit}
 
 import scala.collection.JavaConverters._
 import scala.collection.mutable
 import scala.util.{Failure, Try}
 
+import com.google.common.cache.{Cache, CacheBuilder}
 import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.fs.{FileStatus, Path}
 import org.apache.hadoop.mapreduce._
@@ -30,8 +33,10 @@ import org.apache.hadoop.mapreduce.lib.input.FileSplit
 import org.apache.hadoop.mapreduce.task.TaskAttemptContextImpl
 import org.apache.parquet.filter2.compat.FilterCompat
 import org.apache.parquet.filter2.predicate.FilterApi
+import org.apache.parquet.format.converter.ParquetMetadataConverter
 import org.apache.parquet.hadoop._
 import org.apache.parquet.hadoop.codec.CodecConfig
+import org.apache.parquet.hadoop.metadata.ParquetMetadata
 import org.apache.parquet.hadoop.util.ContextUtil
 import org.apache.parquet.schema.MessageType
 
@@ -278,6 +283,68 @@ class ParquetFileFormat
     true
   }
 
+  override def buildSplitter(
+    sparkSession: SparkSession,
+    fileIndex: FileIndex,
+    filters: Seq[Filter],
+    schema: StructType,
+    hadoopConf: Configuration): (FileStatus => Seq[FileSplit]) = {
+    val pruningEnabled = sparkSession.sessionState.conf.parquetPartitionPruningEnabled
+    val defaultSplitter = super.buildSplitter(sparkSession, fileIndex, filters, schema, hadoopConf)
+    if (!pruningEnabled || filters.isEmpty) {
+      // Return immediately to save FileSystem overhead
+      defaultSplitter
+    } else {
+      val splitters = fileIndex.rootPaths.map { root =>
+        val splits = ParquetFileFormat.fileSplits.get(root,
+          new Callable[ParquetFileSplitter] {
+            override def call(): ParquetFileSplitter =
+              createParquetFileSplits(root, hadoopConf, schema)
+          })
+        root -> splits.buildSplitter(filters)
+      }.toMap
+      val compositeSplitter: (FileStatus => Seq[FileSplit]) = { stat =>
+        val filePath = stat.getPath
+        val rootOption: Option[Path] = fileIndex.rootPaths
+          .find(root => filePath.toString.startsWith(root.toString))
+        val splitterForPath = rootOption.flatMap { root =>
+          splitters.get(root)
+        }.getOrElse(defaultSplitter)
+        splitterForPath(stat)
+      }
+      compositeSplitter
+    }
+  }
+
+  private def createParquetFileSplits(
+    root: Path,
+    hadoopConf: Configuration,
+    schema: StructType): ParquetFileSplitter = {
+    getMetadataForPath(root, hadoopConf)
+      .map(meta => new ParquetMetadataFileSplitter(root, meta.getBlocks.asScala, schema))
+      .getOrElse(ParquetDefaultFileSplitter)
+  }
+
+  private def getMetadataForPath(
+      rootPath: Path,
+      conf: Configuration): Option[ParquetMetadata] = {
+    val fs = rootPath.getFileSystem(conf)
+    try {
+      val stat = fs.getFileStatus(rootPath)
+      // Mimic Parquet behavior. If given a directory, find the underlying _metadata file
+      // If given a single file, check the parent directory for a _metadata file
+      val directory = if (stat.isDirectory) stat.getPath else stat.getPath.getParent
+      val metadataFile = new Path(directory, ParquetFileWriter.PARQUET_METADATA_FILE)
+      val metadata =
+        ParquetFileReader.readFooter(conf, metadataFile, ParquetMetadataConverter.NO_FILTER)
+      Option(metadata)
+    } catch {
+      case notFound: FileNotFoundException =>
+        log.debug(s"No _metadata file found in root $rootPath")
+        None
+    }
+  }
+
   override def buildReaderWithPartitionValues(
       sparkSession: SparkSession,
       dataSchema: StructType,
@@ -424,6 +491,13 @@ class ParquetFileFormat
 }
 
 object ParquetFileFormat extends Logging {
+
+  @transient private val fileSplits: Cache[Path, ParquetFileSplitter] =
+    CacheBuilder.newBuilder()
+      .expireAfterAccess(4, TimeUnit.HOURS)
+      .concurrencyLevel(1)
+      .build()
+
   private[parquet] def readSchema(
       footers: Seq[Footer], sparkSession: SparkSession): Option[StructType] = {
 
--- a/sql/core/src/test/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetQuerySuite.scala
+++ b/sql/core/src/test/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetQuerySuite.scala
@@ -18,8 +18,10 @@
 package org.apache.spark.sql.execution.datasources.parquet
 
 import java.io.File
+import java.net.URI
 
-import org.apache.hadoop.fs.{FileSystem, Path}
+import com.google.common.collect.{HashMultiset, Multiset}
+import org.apache.hadoop.fs.{FileSystem, FSDataInputStream, Path, RawLocalFileSystem}
 import org.apache.parquet.hadoop.ParquetOutputFormat
 
 import org.apache.spark.sql._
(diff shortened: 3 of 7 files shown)


pwoody:
@andreweduffy @robert3005 @HyukjinKwon

pwoody:
There is a performance issue here in that we re-fetch the metadata file for each file. My understanding is that FileFormat is meant to be stateless, but if we can add in a cache then that would be very useful.

HyukjinKwon:
Hi @pwoody, So, if I understood this correctly, the original PR only filters out the files to touch ahead but this one proposes also to filter splits via offsets from Parquet's metadata in driver-side, right?

IIRC, each task will already read the footer before it actually starts to read in executors and then will drop the splits at Parquet-side. This happens fine in both the Spark's vectorized parquet reader and normal parquet reader too. It might be worth reducing the files to touch for the reason I and other guys described in the original PR but I am not sure of pruning splits ahead.

Another potential problem I see here is It seems it does not consider bucketed table read whereas the original PR does.

In addition, I guess we really need a benchmark for the proposal to improve the performance. It is fine if I am wrong and this PR has a benchmark showing the performance improvement with a reasonable explanation.

Lastly, I guess it is a followup including the changes proposed in 14649. Maybe, we can wait until that is merged before submitting a followup. I guess it is being reviewed and I think @andreweduffy is still echoing fine.

HyukjinKwon:
I think we should cc @liancheng as well who is insightful in this area.

pwoody:
Hey @HyukjinKwon - appreciate the feedback!

Re: file touching - If I add the cache to the `_metadata` file, then this PR will end up touching at most one file per rootPath driver side (generally just one total).

Re: files v.s. splits - The main difference when pruning splits instead of files is when you have larger files you will end up spawning tasks that immediately will be filtered out executor-side after grabbing the footer. For simplicity, if we have maxSplitBytes == Parquet row group size then a single hit in a file will end up spawning a task for every row group even if the file only has one matching block. This overhead can end up being expensive in a setup w/ dynamicAllocation and multi-tenancy. I generally wish to reduce the total number of tasks.

Re: bucketed - yep sorry, will fix!

Re: benchmarks - yeah totally happy to poke around and make some benchmarks. ParquetReadBenchmark is the appropriate place I suppose?

Re: old PR - the code there is out of date and I don't believe Andrew is actively working on it based on his last comment. This is the follow up to original ER in what I believe is a more comprehensive and reliable way.

Thanks!

HyukjinKwon:
Ah, I missed the last comment from the old PR. Okay, we can make this shaped nicer. BTW, Spark collects small partitions for each task so I guess this would not introduce a lot of tasks always but yes, I guess it is still a valid point to reduce the number of tasks.

Right, I am fine with this. I thought the original PR was taken over without the courtesy of giving a notification or talking about this ahead. For the benchmark, I have a PR with a benchmark in a PR, 14660, which I also referred from other PRs.

I have just few minor notes which are, maybe `Closes #14649` can be added at the end of the PR description so that the merge script from committers could close the original one if this one gets merged. Another one is, there are some style guide lines I usually refer which are [wiki](https://cwiki.apache.org/confluence/display/SPARK/Spark+Code+Style+Guide) and [databricks/scala-style-guide](https://github.com/databricks/scala-style-guide). I will leave some comments on the changed files.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/FileFormat.scala
```diff
@@ -75,6 +76,17 @@ trait FileFormat {
   }
 
   /**
+   * For a file, return valid splits that may pass the given data filter.
+   */
+  def getSplits(fileIndex: FileIndex,
+                fileStatus: FileStatus,
+                filters: Seq[Filter],
+                schema: StructType,
+                hadoopConf: Configuration): Seq[FileSplit] = {
```
HyukjinKwon: We might be able to fix this and the same instances as below 
```scala
def getSplits(
    fileIndex: FileIndex,
    fileStatus: FileStatus,
...
```

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetFileFormat.scala
```diff
@@ -272,6 +275,83 @@ class ParquetFileFormat
...
+                             metadata: ParquetMetadata,
+                             metadataRoot: Path,
+                             filters: Seq[Filter],
+                             schema: StructType,
+                             hadoopConf: Configuration): Seq[FileSplit] = {
+    val metadataBlocks = metadata.getBlocks
+    val parquetSchema = metadata.getFileMetaData.getSchema
+    val filter = FilterCompat.get(filters
+          .flatMap(ParquetFilters.createFilter(schema, _))
+          .reduce(FilterApi.and))
```
HyukjinKwon: I guess this indentation should be double-spaced.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetFileFormat.scala
```diff
@@ -272,6 +275,83 @@ class ParquetFileFormat
...
+    val filteredMetadata =
+      RowGroupFilter.filterRowGroups(filter, metadataBlocks, parquetSchema).asScala
+    filteredMetadata.flatMap(bmd => {
```
HyukjinKwon: According to https://github.com/databricks/scala-style-guide#anonymous-methods, 

```scala
filteredMetadata.flatMap { bmd =>
  ...
```

seems encouraged.

HyukjinKwon (commented):
I just left some comments for obvious style nits to sweep out.

pwoody:
Ah awesome thanks - my IDE settings must not be properly catching the style - will fix and make sure this doesn't happen in the future.

Also just took a look at the bucketing. I've implemented the same logic, which will prune the file if no splits exist, but it is still all or nothing for a given file because it launches tasks based on bucket. Commit coming soon.

Thanks again.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetFileFormat.scala
```diff
@@ -272,6 +275,83 @@ class ParquetFileFormat
...
+      // Mimic Parquet behavior. If given a directory, find the underlying _metadata file
+      // If given a single file, check the parent directory for a _metadata file
+      val directory = if (stat.isDirectory) stat.getPath else stat.getPath.getParent
+      val metadataFile = new Path(directory, ParquetFileWriter.PARQUET_METADATA_FILE)
+      val metadata =
+        ParquetFileReader.readFooter(conf, metadataFile, ParquetMetadataConverter.NO_FILTER)
+
+      // Ensure that the metadata has an entry for the file.
+      // If it does not, do not filter at this stage.
+      val matchingBlockExists = metadata.getBlocks.asScala.exists(bmd => {
```
andreweduffy: I'm a little unclear as to why this branch is necessary. Under what conditions do you have a parquet dataset where one of its files isn't in the summary metadata? Is this the case where you write out the dataset with metadata, then add a new partition, and write without updating the metadata?

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetFileFormat.scala
```diff
@@ -272,6 +275,83 @@ class ParquetFileFormat
...
+      val fsPath = fileStatus.getPath
+      if (bmdPath == fsPath) {
+        Some(new FileSplit(bmdPath, bmd.getStartingPos, bmd.getTotalByteSize, Array.empty))
+      } else {
+        None
+      }
+    })
+  }
+
+  private def getMetadataForPath(filePath: Path,
```
andreweduffy: I think this should be cacheable, as the lifetime of a `FileFormat` is tied to a particular `HadoopFsRelation`, which again is tied to a `Dataset`. Something may have changed about the API but from a quick look through the relevant classes I don't see anything to refute this. I don't see any written guarantees in the FileFormat API that concrete impls need to be stateless, but neither did I find any reference to caching/state in other implementations. Still I think it should be safe to cache on the ParquetFileFormat for this use case.

sql/core/src/test/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetQuerySuite.scala
```diff
@@ -703,6 +703,34 @@ class ParquetQuerySuite extends QueryTest with ParquetTest with SharedSQLContext
...
+        val zeroPartitions = spark.read.parquet(path.getCanonicalPath).where("x = 0")
+        assert(zeroPartitions.rdd.partitions.length == 0)
+
+        val onePartition = spark.read.parquet(path.getCanonicalPath).where("x = 1")
+        assert(onePartition.rdd.partitions.length == 1)
+      }
+    }
+  }
+
+  test("Do not filter out parquet file when missing in _metadata file") {
```
andreweduffy: I'm not sure I understand what this test accomplishes. If you write out twice with summary metadata, why would anything be missing in `_metadata` file?

a10y (commented):
Agree this is better than the original file-based pruning strategy in my old PR. I have a few comments which probably are largely due to a lack of understanding on my part, overall things look good.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetFileFormat.scala
pwoody: By default, writes won't merge parquet metadata unfortunately. So if you use SaveMode.Append on an existing parquet directory it will re-write it.

sql/core/src/test/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetQuerySuite.scala
pwoody: See other comment, but yeah - it is when you overwrite the metadata file.

pwoody:
Cool - I've added the caching, fixed style issues, and added pruning to the bucketed reads.

pwoody:
I've pushed up the ability to configure this feature being enabled as well. Here is a benchmark when writing out 200 files with this code:

```
withSQLConf(ParquetOutputFormat.ENABLE_JOB_SUMMARY -> "true") {
      withTempPath { path =>
        spark.range(0, numPartitions, 1, numPartitions)
          .write.parquet(path.getCanonicalPath)
        val benchmark = new Benchmark("Parquet partition pruning benchmark", numPartitions)

        benchmark.addCase("Parquet partition pruning enabled") { iter =>
          spark.read.parquet(path.getCanonicalPath).filter("id = 0").collect()
        }

        benchmark.addCase("Parquet partition pruning disabled") { iter =>
          withSQLConf(SQLConf.PARQUET_PARTITION_PRUNING_ENABLED.key -> "false") {
            spark.read.parquet(path.getCanonicalPath).filter("id = 0").collect()
          }
        }

        benchmark.run()
      }
    }
```


```
Running benchmark: Parquet partition pruning benchmark
  Running case: Parquet partition pruning enabled
  Stopped after 12 iterations, 2049 ms
  Running case: Parquet partition pruning disabled
  Stopped after 5 iterations, 2177 ms

Java HotSpot(TM) 64-Bit Server VM 1.8.0_20-b26 on Mac OS X 10.10.5
Intel(R) Core(TM) i7-3635QM CPU @ 2.40GHz

Parquet partition pruning benchmark:     Best/Avg Time(ms)    Rate(M/s)   Per Row(ns)   Relative
Parquet partition pruning enabled              145 /  171          0.0      723119.3       1.0X
Parquet partition pruning disabled             414 /  436          0.0     2070279.4       0.3X


Process finished with exit code 0

```

sql/core/src/test/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetQuerySuite.scala
```diff
@@ -703,6 +705,81 @@ class ParquetQuerySuite extends QueryTest with ParquetTest with SharedSQLContext
...
+    withSQLConf(ParquetOutputFormat.ENABLE_JOB_SUMMARY -> "true") {
+      withTempPath { path =>
+        spark.sparkContext.parallelize(Seq(1, 2, 3), 3)
+          .toDF("x").write.parquet(path.getCanonicalPath)
+        spark.sparkContext.parallelize(Seq(4))
+          .toDF("x").write.mode(SaveMode.Append).parquet(path.getCanonicalPath)
```
ash211: do you need to turn off `ParquetOutputFormat.ENABLE_JOB_SUMMARY` to prevent the `_metadata` file from being updated on the second write, or is it normally not updated in `SaveMode.Append` ?

I would've expected the second write here to update the `_metadata` file in which case the test name wouldn't match the behavior.  But if the second write _doesn't_ update `_metadata` then the file has 3 out of 4 file partitions in it, matching the test name.

pwoody: Yeah - i've edited the test to clarify the intended behavior. i've also change how we cache the metadata to reflect it as well.

pwoody:
@rxin @HyukjinKwon ready for more review on my end.

HyukjinKwon:
I guess we should ping @liancheng as he was reviewing the previous one.

pwoody:
I've updated the structure of the PR to change caching to be global across instances of FileFormat, have expiry, and reuse known filters. Here is a new benchmark to highlight the filter re-use (I flipped the results to make it easier to read)

```
    withSQLConf(ParquetOutputFormat.ENABLE_JOB_SUMMARY -> "true") {
      withTempPath { path =>
        spark.range(0, 200, 1, 200)
          .write.parquet(path.getCanonicalPath)
        val benchmark = new Benchmark("Parquet partition pruning benchmark", 200)
        benchmark.addCase("Parquet partition pruning disabled") { iter =>
          withSQLConf(SQLConf.PARQUET_PARTITION_PRUNING_ENABLED.key -> "false") {
            var df = spark.read.parquet(path.getCanonicalPath).filter("id = 0")
            for (i <- 1 to 10) {
              df = df.filter(s"id < $i")
              df.collect()
            }
          }
        }
        benchmark.addCase("Parquet partition pruning enabled") { iter =>
          var df = spark.read.parquet(path.getCanonicalPath).filter("id = 0")
          for (i <- 1 to 10) {
            df = df.filter(s"id < $i")
            df.collect()
          }
        }
        benchmark.run()
      }
    }
```


```
Running benchmark: Parquet partition pruning benchmark
  Running case: Parquet partition pruning disabled
  Stopped after 2 iterations, 8744 ms
  Running case: Parquet partition pruning enabled
  Stopped after 5 iterations, 2187 ms

Java HotSpot(TM) 64-Bit Server VM 1.8.0_20-b26 on Mac OS X 10.10.5
Intel(R) Core(TM) i7-3635QM CPU @ 2.40GHz

Parquet partition pruning benchmark:     Best/Avg Time(ms)    Rate(M/s)   Per Row(ns)   Relative
Parquet partition pruning disabled            4332 / 4372          0.0    21659450.2       1.0X
Parquet partition pruning enabled              399 /  438          0.0     1995877.1      10.9X
```

rxin:
This creates huge problems when the table is big doesn't it? We just did a big change to get rid of the per table file status cache, because its existence made Spark unstable with dealing with tables (tables with a lot of files).

pwoody:
Hey - yeah definitely a real concern as it needs driver heap to scale with the size of the metadata of the table you are going to read in.

We could be creative to add heuristics around reading the metadata at all if it is too large or add the ability to spill the existing metadata to disk. Do you have any preferences/thoughts?

PR #15836, opened 2016-11-09
[SPARK-18391] Openstack deployment mode for any Spark version since 1.0

al-indigo:
This change adds a link to universal way to deploy Apache Spark since 1.0 up to 2.0.1 with any supported version of Hadoop in Openstack environment.

We utilize vanilla images (nothing pre-installed is needed; note it works only for cloud images of OS [but it's reasonable])

We support YARN-mode, Apache Ignite, NFS-mount, Jupyter notebooks, etc as optional on-demand options.

All the work is licensed Apache 2.0 explicitly.

The scenario was developed at Institute for System Programming of the Russian Academy of Sciences.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -107,6 +107,7 @@ options for deployment:
 * [Submitting Applications](submitting-applications.html): packaging and deploying applications
 * Deployment modes:
   * [Amazon EC2](https://github.com/amplab/spark-ec2): scripts that let you launch a cluster on EC2 in about 5 minutes
+  * [Openstack](https://github.com/ispras/spark-openstack): scripts that let you launch a cluster with any version since 1.0 in about 10 minutes (and optional additional tools)
   * [Standalone Deploy Mode](spark-standalone.html): launch a standalone cluster quickly without a third-party cluster manager
   * [Mesos](running-on-mesos.html): deploy a private cluster using
       [Apache Mesos](http://mesos.apache.org)


HyukjinKwon:
Could you please fix the title to the one that summerises what this PR proposes?

rxin:
Hm we typically don't link to 3rd party things that way. Sorry please close this.

rxin:
@al-indigo mind closing the pull request?

srowen:
@al-indigo you can add this to spark-packages.org or I can put it on https://cwiki.apache.org/confluence/display/SPARK/Third+Party+Projects

al-indigo:
@rxin I just thought that it makes sense to have links for Amazon EC2 and Openstack at the same place since the deployments are for the same purpose and provides the same CLI options for different cloud platforms

I thought that amplab scripts are 3rd party too, and we had a conversation in incubator-spark repository in October-November 2013 about it; as far as I remember Matei Zaharia recommended us to make a separate repository and have a link to it from Spark docs.

If you insist, I will close this PR and use @srowen ; I'm just worried about Spark users can discover it in docs much easier than in wiki or spark-packages

srowen:
@al-indigo this document describes integrations that are part of Spark (or were in the case of EC2, which was only halfway in Spark and kind of spun back out, and could arguably be removed from this list) rather than 3rd-party project.

al-indigo:
@srowen Ok, got your point, thanks

al-indigo:
@srowen excuse me, I forgot to say that if you add it to Wiki, it would be great (you may close pr again as you see the comment)

PR #15838, opened 2016-11-10
[SPARK-18396][HISTORYSERVER]"Duration" column makes search result confused, maybe we should make it unsearchable

WangTaoTheTonic:

When we search data in History Server, it will check if any columns contains the search string. Duration is represented as long value in table, so if we search simple string like "003", "111", the duration containing "003", ‘111“ will be showed, which make not much sense to users.
We cannot simply transfer the long value to meaning format like "1 h", "3.2 min" because they are also used for sorting. Better way to handle it is ban "Duration" columns from searching.

## How was this patch tested

manually tests. 

Before("local-1478225166651" pass the filter because its duration in long value, which is "257244245" contains search string "244"): 
before

After:
after

### DIFF
--- a/core/src/main/resources/org/apache/spark/ui/static/historypage.js
+++ b/core/src/main/resources/org/apache/spark/ui/static/historypage.js
@@ -139,6 +139,9 @@ $(document).ready(function() {
                         {name: 'eighth'},
                         {name: 'ninth'},
                     ],
+                    "columnDefs": [
+                        {"searchable": false, "targets": [5]}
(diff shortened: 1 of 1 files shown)


srowen (approved):
I tend to agree because I can't see a use case for searching durations, really.

ajbozarth (approved):
Nice catch, LGTM

WangTaoTheTonic:
Is it good to go? @srowen @vanzin

PR #15839, opened 2016-11-10
[SPARK-18397][SQL]cannot create table by using the hive default fileformat

classjtt:
spark2.x cannot create table by using the hive default fileformat.
in hive-site.xml,I add properties like this: 
<property> 
<name>hive.default.fileformat</name> 
<value>RCFile</value> 
</property> 
but I create table ,this properties have no effect, the hive table use textfile fileformat.


gatorsmile:
Thanks for reporting this! Many parameters in HiveConf are not respected by Spark SQL. If users need to change the default file format, we suggest users to set it through our Spark SQL configuration. 

Could you close this PR at first? We can revisit this issue when we try to resolve the whole problem.

gatorsmile:
We are closing it due to inactivity. please do reopen if you want to push it forward. Thanks!

PR #15841, opened 2016-11-10
[SPARK-18399][SQL][DOCUMENTATION] Examples in SparkSQL/DataFrame/DataSets guide fails with default configuration

oza:

With default configuration, examples in Spark SQL, DataFrames and Datasets Guide result in a failure since they try to access HDFS while the paths of people.json, people.txt are assumed to be in local file system.

## How was this patch tested?

I test this with manual test: running examples of SparkSQL/DataFrame/DataSets with default configuration. With this fix, it works well without configuring HDFS.

### DIFF
--- a/examples/src/main/scala/org/apache/spark/examples/sql/SparkSQLExample.scala
+++ b/examples/src/main/scala/org/apache/spark/examples/sql/SparkSQLExample.scala
@@ -60,7 +60,8 @@ object SparkSQLExample {
 
   private def runBasicDataFrameExample(spark: SparkSession): Unit = {
     // $example on:create_df$
-    val df = spark.read.json("examples/src/main/resources/people.json")
+    val df = spark.read.json(
+      "file:///" + System.getProperty("user.dir") + "/examples/src/main/resources/people.json")
 
     // Displays the content of the DataFrame to stdout
     df.show()
@@ -179,7 +180,8 @@ object SparkSQLExample {
     primitiveDS.map(_ + 1).collect() // Returns: Array(2, 3, 4)
 
     // DataFrames can be converted to a Dataset by providing a class. Mapping will be done by name
-    val path = "examples/src/main/resources/people.json"
+    val path =
+      "file:///" + System.getProperty("user.dir") + "/examples/src/main/resources/people.json"
     val peopleDS = spark.read.json(path).as[Person]
     peopleDS.show()
     // +----+-------+
@@ -199,7 +201,8 @@ object SparkSQLExample {
 
     // Create an RDD of Person objects from a text file, convert it to a Dataframe
     val peopleDF = spark.sparkContext
-      .textFile("examples/src/main/resources/people.txt")
+      .textFile(
+        "file:///" + System.getProperty("user.dir") + "/examples/src/main/resources/people.txt")
       .map(_.split(","))
       .map(attributes => Person(attributes(0), attributes(1).trim.toInt))
       .toDF()
@@ -240,7 +243,8 @@ object SparkSQLExample {
     import spark.implicits._
     // $example on:programmatic_schema$
     // Create an RDD
-    val peopleRDD = spark.sparkContext.textFile("examples/src/main/resources/people.txt")
+    val peopleRDD = spark.sparkContext.textFile(
+      "file:///" + System.getProperty("user.dir") + "/examples/src/main/resources/people.txt")
 
     // The schema is encoded in a string
     val schemaString = "name age"


examples/src/main/scala/org/apache/spark/examples/sql/SparkSQLExample.scala
```diff
@@ -60,7 +60,8 @@ object SparkSQLExample {
 
   private def runBasicDataFrameExample(spark: SparkSession): Unit = {
     // $example on:create_df$
-    val df = spark.read.json("examples/src/main/resources/people.json")
+    val df = spark.read.json(
+      "file:///" + System.getProperty("user.dir") + "examples/src/main/resources/people.json")
```
srowen: I don't think this is correct because there's also not a reason to assume this is the location. Just write `file:examples/...`? does that not work?

oza: @srowen thanks for taking a look! Unfortunately, your suggestion doesn't work well in this case. I tried it on my local spark shell:

``` 
scala> val df = spark.read.json("file:examples/src/main/resources/people.json")
16/11/10 20:12:25 WARN datasources.DataSource: Error while looking for metadata directory.
java.lang.IllegalArgumentException: java.net.URISyntaxException: Relative path in absolute URI: file:examples/src/main/resources/people.json
  at org.apache.hadoop.fs.Path.initialize(Path.java:206)
  at org.apache.hadoop.fs.Path.<init>(Path.java:172)
  at org.apache.spark.sql.execution.datasources.DataSource$$anonfun$8.apply(DataSource.scala:283)
  at org.apache.spark.sql.execution.datasources.DataSource$$anonfun$8.apply(DataSource.scala:282)
  at scala.collection.TraversableLike$$anonfun$flatMap$1.apply(TraversableLike.scala:241)
  at scala.collection.TraversableLike$$anonfun$flatMap$1.apply(TraversableLike.scala:241)
  at scala.collection.TraversableLike$class.flatMap(TraversableLike.scala:241)
  at scala.collection.immutable.List.flatMap(List.scala:344)
  at org.apache.spark.sql.execution.datasources.DataSource.resolveRelation(DataSource.scala:282)
  at org.apache.spark.sql.DataFrameReader.load(DataFrameReader.scala:152)
  at org.apache.spark.sql.DataFrameReader.json(DataFrameReader.scala:297)
  at org.apache.spark.sql.DataFrameReader.json(DataFrameReader.scala:250)
```

As you mentioned, the fix is not correct if we are not in spark homw directory. How about fixing the documentation to do the comments in spark directory?

oza:
I found that my previous PR has a bug(forgot to add "/" before example). Resending PR with the fix.

oza:
Before the patch:
```
scala> val df = spark.read.json("examples/src/main/resources/people.json")
16/11/10 21:02:59 WARN datasources.DataSource: Error while looking for metadata directory.
java.net.ConnectException: Call From ozamac-2.local/10.129.56.104 to localhost:9000 failed on connection exception: java.net.ConnectException: Connection refused; For more details see:  http://wiki.apache.org/hadoop/ConnectionRefused
```

After the patch:
```
scala> val df = spark.read.json("file:///" + System.getProperty("user.dir") + "/examples/src/main/resources/people.json")
df: org.apache.spark.sql.DataFrame = [age: bigint, name: string]
```

examples/src/main/scala/org/apache/spark/examples/sql/SparkSQLExample.scala
HyukjinKwon: Isn't this supposed to be ran via https://github.com/apache/spark#example-programs ? I just ran some of them after manually building and it seems fine.

examples/src/main/scala/org/apache/spark/examples/sql/SparkSQLExample.scala
```diff
@@ -60,7 +60,8 @@ object SparkSQLExample {
 
   private def runBasicDataFrameExample(spark: SparkSession): Unit = {
     // $example on:create_df$
-    val df = spark.read.json("examples/src/main/resources/people.json")
+    val df = spark.read.json(
+      "file:///" + System.getProperty("user.dir") + "/examples/src/main/resources/people.json")
```
hvanhovell: This makes a very strong assumption about the location of the files right?

examples/src/main/scala/org/apache/spark/examples/sql/SparkSQLExample.scala
oza: @HyukjinKwon Thanks for your comment! Without HDFS configuration(including HADOOP_HOME), it result in failure.

```
./bin/run-example sql.SparkSQLExample
...
16/11/10 22:48:07 WARN datasources.DataSource: Error while looking for metadata directory.
Exception in thread "main" java.net.ConnectException: Call From ozamac-2.local/10.0.1.20 to localhost:9000 failed on connection exception: java.net.ConnectException: Connection refused; For more details see:  http://wiki.apache.org/hadoop/ConnectionRefused
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
```

After applying the patch, it works correctly.

examples/src/main/scala/org/apache/spark/examples/sql/SparkSQLExample.scala
oza: Yes. This also assumes that user's current directory is $SPARK_HOME. However, without this change, ``` ./bin/run-example sql.SparkSQLExample``` goes failure with default configuration.

Another option is to give a path of people.json/people.text via CLI. I would like to hear your thought.

hvanhovell: No it does not, the `user.dir` is simply the location java is started from. That can be anywhere.

oza: Oh, my bad. Thanks for pointing out. Then, how about passing the path of people.json/people.text via CLI?

oza: > This makes a very strong assumption about the location of the files right?

I thought so, but it results in failure because client try to access HDFS. The stack trace is here:
https://issues.apache.org/jira/browse/SPARK-18399?focusedCommentId=15653722&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-15653722

examples/src/main/scala/org/apache/spark/examples/sql/SparkSQLExample.scala
hvanhovell: I just downloaded Spark 2.0.1 and executed `./bin/run-example sql.SparkSQLExample` in the spark dir. I am on a Mac (on which platform are you?). It only becomes an issue when you try to run the examples from a different location; then it will fail.

examples/src/main/scala/org/apache/spark/examples/sql/SparkSQLExample.scala
hvanhovell: You wouldn't happen to some hadoop configuration on your system?

srowen:
We want the examples to be consistent, so wouldn't this affect all examples?
And yes they're supposed to be run with the given script.
Putting the path in the source code might help a bit for people reading it as source code, which is how these examples are really consumed, rather than referring to what the command line args are supposed to be. If it works with `spark-example` and so do other examples, I think we should leave them.

HyukjinKwon:
As far as I know, the default file system should be `file:///`.

I created a virtual machine with the OS `Ubuntu Server Trusty 14.04 amd64`, from the link https://oss-binaries.phusionpassenger.com/vagrant/boxes/latest/ubuntu-14.04-amd64-vbox.box

And then, I ran the commands as below:

```
sudo apt-get install openjdk-7-jdk -y
wget http://d3kbcqa49mib13.cloudfront.net/spark-2.0.1-bin-hadoop2.7.tgz
tar xzvf spark-2.0.1-bin-hadoop2.7.tgz spark-2.0.1-bin-hadoop2.7
cd spark-2.0.1-bin-hadoop2.7/
```

and I ran

```
./bin/run-example sql.SparkSQLExample
```

this seems working fine. Could you please try this?

examples/src/main/scala/org/apache/spark/examples/sql/SparkSQLExample.scala
HyukjinKwon: Please check `fs.default.name`. If the path does not specify the file system in the schema, then, I believe the property will be referred. If it is not found, it will use `file:///`[1].

[1]https://github.com/apache/hadoop/blob/7e521c5a49fbcf88285c102051ea2522edc847b9/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/CommonConfigurationKeysPublic.java#L58

oza:
Thanks a lot for your comments! I see, in fact, I had the problem on Mac OS. Okay, I will try it on your VM, and on my local linux. Please wait a moment.

oza:
@srowen it makes sense to me. In that case, we can update our documentation to avoid the problem.

examples/src/main/scala/org/apache/spark/examples/sql/SparkSQLExample.scala
oza: @hvanhovell @HyukjinKwon oh, it's completely my bad! You're right, I had a hadoop configuration. Closing this as not a problem.

oza:
I'm sorry, it's completely

hvanhovell:
No problem. Thanks for the effort!

HyukjinKwon:
+1 for no problem :).

PR #15843, opened 2016-11-10
[SPARK-18274][ML][PYSPARK] Memory leak in PySpark JavaWrapper

techaddict:
In`JavaWrapper `'s destructor make Java Gateway dereference object in destructor, using `SparkContext._active_spark_context._gateway.detach` 
Fixing the copying parameter bug, by moving the `copy` method from `JavaModel` to `JavaParams`

## How was this patch tested?
```scala
import random, string
from pyspark.ml.feature import StringIndexer

l = [(''.join(random.choice(string.ascii_uppercase) for _ in range(10)), ) for _ in range(int(7e5))]  # 700000 random strings of 10 characters
df = spark.createDataFrame(l, ['string'])

for i in range(50):
    indexer = StringIndexer(inputCol='string', outputCol='index')
    indexer.fit(df)
```
* Before: would keep StringIndexer strong reference, causing GC issues and is halted midway
After: garbage collection works as the object is dereferenced, and computation completes
* Mem footprint tested using profiler
* Added a parameter copy related test which was failing before.


techaddict:
cc: @jkbradley @davies @holdenk

jkbradley:
Thanks a lot for finding & reporting this!  The fix should probably go in JavaWrapper, not JavaModel, right?

I tested this manually (in JavaWrapper), and it seems to fix the problematic case with StringIndexer.

techaddict:
@jkbradley yes I did it for `JavaWrapper` first, but try running tests with it gives https://amplab.cs.berkeley.edu/jenkins/job/SparkPullRequestBuilder/68478/consoleFull

jkbradley:
You're right!  It's another bug: copy should be implemented in JavaParams, not JavaModel.  I'm sending this PR to fix that: https://github.com/techaddict/spark/pull/1

Can you please check it out and merge it into your PR if it looks OK to you?  All pyspark.ml tests ran successfully with it.

techaddict:
@jkbradley looks good, merged 👍

python/pyspark/ml/wrapper.py
```diff
@@ -33,6 +33,10 @@ def __init__(self, java_obj=None):
         super(JavaWrapper, self).__init__()
         self._java_obj = java_obj
 
+    def __del__(self):
+        if SparkContext._active_spark_context:
```
techaddict: checking if there is active spark context, got this error after `quit()` in `pyspark`
```
Exception ignored in: <bound method JavaWrapper.__del__ of
StringIndexer_4a75b9e8c92f56703aff>
File "/Users/xx/Project/Spark/python/pyspark/ml/wrapper.py", line
37, in __del__
SparkContext._active_spark_context._gateway.detach(self._java_obj)
AttributeError: 'NoneType' object has no attribute '_gateway'
Exception ignored in: <bound method JavaWrapper.__del__ of
StringIndexer_4a75b9e8c92f56703aff>
File "/Users/xx/Project/Spark/python/pyspark/ml/wrapper.py", line
37, in __del__
AttributeError: 'NoneType' object has no attribute '_gateway'
```

python/pyspark/ml/wrapper.py
```diff
@@ -180,6 +184,24 @@ def __get_class(clazz):
                                       % stage_name)
         return py_stage
 
+    def copy(self, extra=None):
+        """
+        Creates a copy of this instance with the same uid and some
+        extra params. This implementation first calls Params.copy and
+        then make a copy of the companion Java model with extra params.
```
viirya: nit: Java model -> Java pipeline component.

python/pyspark/ml/wrapper.py
```diff
@@ -180,6 +184,25 @@ def __get_class(clazz):
                                       % stage_name)
         return py_stage
 
+    def copy(self, extra=None):
+        """
+        Creates a copy of this instance with the same uid and some
+        extra params. This implementation first calls Params.copy and
+        then make a copy of the companion Java pipeline component with
+        extra params. So both the Python wrapper and the Java model get
```
viirya: nit: There is another "Java model" here too.

viirya:
LGTM with minor doc comment.

holdenk:
So this change looks good to me, but it seems like it fixes more than just the bug described in the JIRA & PR description with @jkbradley's change integrated (namely the issue with param copy which we have). For people who are looking for what's changed between versions it might make sense to explain the copy related fix the PR description as well since that is what is used in the commit log.

techaddict:
@holdenk updated the description.

holdenk:
LGTM thanks for fixing this @techaddict :D :)

holdenk:
ping @davies if you have time for final review/merge?

jkbradley:
Good point @holdenk --- @techaddict could you also please update the PR title to say "JavaWrapper" instead of "StringIndexer"?

jkbradley:
I'm now wondering if ```__del___``` should be in JavaParams instead of JavaWrapper since JavaWrapper does not override ```copy```.  Do yall agree it will be safer if it's moved there?

viirya:
@jkbradley Sounds making sense more.

holdenk:
I'm not so sure about that, we still would want to cleanup the underlying Java reference object on delete if it isn't needed anymore. I think the question is do we want to support shallow copy of javawrapper objects?

viirya:
Oh. I thought JavaWrapper is only used on JavaParams. But there are also others like LogisticRegressionSummary which directly inherits JavaWrapper.

jkbradley:
I don't see a need to do deep copies of model summaries, but I agree I don't like how JavaWrapper is ambiguous about whether it does shallow or deep copies.

I'd say the confusion comes from us having a mix of immutable Java types (like model summaries) and mutable Java types (like Params subclasses).  What do you think of these 2 options?
* Distinguish mutability within Python wrappers: JavaWrapper is usable for immutable types.  JavaParams (or other subtypes, if needed) is usable for mutable types.  I.e., ```__del__``` and ```copy``` go in JavaParams.
* Distinguish mutability within Java only: Use the same wrapper types for both in Python, and Java copy methods can do deep or shallow copies.  I.e., in JavaWrapper, implement copy() which copies the Java instance, and implement ```__del__``` to release that instance's handle.

I don't think either option does much for enforcing these semantics.  Barring GC issues, I'd pick option 1 since it's simpler.  But if option 2 is better for GC issues, then I'd vote for it.

Thoughts?

viirya:
I'd prefer option2 for safety since the model summaries should be an issue for GC. And looks like Java model summaries don't have copy method.

holdenk:
From the Py4J documentation it seems like we could be leaking memory with the first option, although perhaps not a lot of memory, but if it was being used in an iterative Python algorithm for training many models it could start to have some impact. I'd be in favor of option 2, but that could be done as a follow up issue if the required copy methods aren't generally available.

viirya:
@jkbradley @holdenk @techaddict Do we want to implement copy() in JavaWrapper in this PR too? Or separate it to follow ones with the required copy methods on JVM side?

jkbradley:
Sorry for the slow response on this.  Given the time pressure for 2.1, let's go with option 1 for now with a follow-up task to implement option 2.  It would be great to include this fix in 2.1.

@techaddict will you have time to update your PR quickly?  Thank you!

holdenk:
I agree, for a follow up (so we don't lose track of it) - I've created SPARK-18630 but option 1 for now is a strict improvement over the current situation. Thanks for all of your work on this @techaddict

techaddict:
@jkbradley @holdenk will update the PR with changes today.

techaddict:
@jkbradley @holdenk @viirya PR updated

viirya:
LGTM

jkbradley:
LGTM too
Thanks a lot!
Merging with master, branch-2.1, branch-2.0

Has anyone heard of complaints of this in current use cases of earlier branches?  If not, I won't backport it further than 2.0.

PR #15846, opened 2016-11-10
[CORE][Minor]:remove unused import in SparkContext.scala

wangmiao1981:

Found one unused import in SparkContext.scala while auditing ML code.

Remove it.
## How was this patch tested?

Remove the unused import and compile successfully


rxin:
hm thanks for submitting but we also want to discourage change this small without any user facing benefit.

wangmiao1981:
OK. I will close it. Thanks!

PR #15851, opened 2016-11-11
[SPARK-18412][SPARKR][ML] Fix exception for some SparkR ML algorithms training on libsvm data

yanboliang:
* Fix the following exceptions which throws when ```spark.randomForest```(classification), ```spark.gbt```(classification), ```spark.naiveBayes``` and ```spark.glm```(binomial family) were fitted on libsvm data.
```
java.lang.IllegalArgumentException: requirement failed: If label column already exists, forceIndexLabel can not be set with true.
```
See [SPARK-18412](https://issues.apache.org/jira/browse/SPARK-18412) for more detail about how to reproduce this bug.
* Refactor out ```getFeaturesAndLabels``` to RWrapperUtils, since lots of ML algorithm wrappers use this function.
* Drop some unwanted columns when making prediction.

## How was this patch tested?
Add unit test.

### DIFF
--- a/R/pkg/inst/tests/testthat/test_mllib.R
+++ b/R/pkg/inst/tests/testthat/test_mllib.R
@@ -971,6 +971,12 @@ test_that("spark.randomForest Classification", {
   predictions <- collect(predict(model, data))$prediction
   expect_equal(length(grep("1.0", predictions)), 50)
   expect_equal(length(grep("2.0", predictions)), 50)
+
+  # spark.randomForest classification can work on libsvm data
+  data <- read.df(absoluteSparkPath("data/mllib/sample_multiclass_classification_data.txt"),
+                source = "libsvm")
+  model <- spark.randomForest(data, label ~ features, "classification")
+  expect_equal(summary(model)$numFeatures, 4)
 })
 
 test_that("spark.gbt", {
@@ -1039,6 +1045,12 @@ test_that("spark.gbt", {
   expect_equal(iris2$NumericSpecies, as.double(collect(predict(m, df))$prediction))
   expect_equal(s$numFeatures, 5)
   expect_equal(s$numTrees, 20)
+
+  # spark.gbt classification can work on libsvm data
+  data <- read.df(absoluteSparkPath("data/mllib/sample_binary_classification_data.txt"),
+                source = "libsvm")
+  model <- spark.gbt(data, label ~ features, "classification")
+  expect_equal(summary(model)$numFeatures, 692)
 })
 
 sparkR.session.stop()
--- a/R/pkg/inst/tests/testthat/test_mllib.R
+++ b/R/pkg/inst/tests/testthat/test_mllib.R
@@ -881,7 +881,8 @@ test_that("spark.kstest", {
   expect_match(capture.output(stats)[1], "Kolmogorov-Smirnov test summary:")
 })
 
-test_that("spark.randomForest Regression", {
+test_that("spark.randomForest", {
+  # regression
   data <- suppressWarnings(createDataFrame(longley))
   model <- spark.randomForest(data, Employed ~ ., "regression", maxDepth = 5, maxBins = 16,
                               numTrees = 1)
@@ -923,9 +924,8 @@ test_that("spark.randomForest Regression", {
   expect_equal(stats$treeWeights, stats2$treeWeights)
 
   unlink(modelPath)
-})
 
-test_that("spark.randomForest Classification", {
+  # classification
   data <- suppressWarnings(createDataFrame(iris))
   model <- spark.randomForest(data, Species ~ Petal_Length + Petal_Width, "classification",
                               maxDepth = 5, maxBins = 16)
@@ -979,7 +979,8 @@ test_that("spark.randomForest Classification", {
   expect_equal(summary(model)$numFeatures, 4)
 })
 
-test_that("spark.gbt regression", {
+test_that("spark.gbt", {
+  # regression
   data <- suppressWarnings(createDataFrame(longley))
   model <- spark.gbt(data, Employed ~ ., "regression", maxDepth = 5, maxBins = 16, seed = 123)
   predictions <- collect(predict(model, data))
@@ -1008,9 +1009,8 @@ test_that("spark.gbt regression", {
   expect_equal(stats$treeWeights, stats2$treeWeights)
 
   unlink(modelPath)
-})
 
-test_that("spark.gbt classification", {
+  # classification
   # label must be binary - GBTClassifier currently only supports binary classification.
   iris2 <- iris[iris$Species != "virginica", ]
   data <- suppressWarnings(createDataFrame(iris2))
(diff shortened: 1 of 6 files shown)


R/pkg/inst/tests/testthat/test_mllib.R
```diff
@@ -971,10 +971,15 @@ test_that("spark.randomForest Classification", {
   predictions <- collect(predict(model, data))$prediction
   expect_equal(length(grep("1.0", predictions)), 50)
   expect_equal(length(grep("2.0", predictions)), 50)
+
+  # spark.randomForest classification can work on libsvm data
+  data <- read.df(absoluteSparkPath("data/mllib/sample_multiclass_classification_data.txt"),
+                source = "libsvm")
+  model <- spark.randomForest(data, label ~ features, "classification")
+  expect_equal(summary(model)$numFeatures, 4)
```
yanboliang: This only tests whether the fitting throws exception, so I did not check all result.

yanboliang: BTW, I only add tests of libsvm data for ```spark.gbt``` and ```spark.randomForest```. This PR fix the exception for ```spark.naiveBayes``` and ```spark.glm``` also, I did not add tests for those algorithms because we fix the issue at ```RWrapperUtils.checkDataColumns``` which was called by all algorithms. It means that if one algorithm works well, others should also work well.

mllib/src/main/scala/org/apache/spark/ml/r/GBTClassificationWrapper.scala
```diff
@@ -51,6 +51,7 @@ private[r] class GBTClassifierWrapper private (
     pipeline.transform(dataset)
       .drop(PREDICTED_LABEL_INDEX_COL)
       .drop(gbtcModel.getFeaturesCol)
+      .drop(gbtcModel.getLabelCol)
```
yanboliang: For all classification algorithms, we do ```forceIndexLabel``` to generate new label column from the original label column, so we need to drop the generated label to make the output more clear to be understand when making prediction.

yanboliang:
cc @felixcheung

mllib/src/main/scala/org/apache/spark/ml/r/RWrapperUtils.scala
```diff
@@ -41,5 +41,29 @@ object RWrapperUtils extends Logging {
         s"using new name $newFeaturesName instead")
       rFormula.setFeaturesCol(newFeaturesName)
     }
+
+    if (rFormula.getForceIndexLabel && data.schema.fieldNames.contains(rFormula.getLabelCol)) {
+      val newLabelName = s"${Identifiable.randomUID(rFormula.getLabelCol)}"
+      logWarning(s"data containing ${rFormula.getLabelCol} column and we force to index label, " +
```
felixcheung: one concern is this might happen a lot - esp. with libsvm data or similar. Also this is benign correct? Should we just make this INFO so this won't pop up as much?

R/pkg/inst/tests/testthat/test_mllib.R
felixcheung: one minor nit is though the tests for glm or naiveBayes are earlier on so we might want to test this there instead of a later test like randomForest or gbt.

R/pkg/inst/tests/testthat/test_mllib.R
```diff
@@ -971,10 +971,15 @@ test_that("spark.randomForest Classification", {
...
 })
 
-test_that("spark.gbt", {
-  # regression
+test_that("spark.gbt regression", {
```
felixcheung: my thought was actually the other way around - I was going to merge the randomForest ones into one tests.
on a related note, I think mllib.R and test_mllib.R are getting too long we might to break them up into more manageable groups. randomForest and gbt classification and regression (R APIs) are similarly enough that I thought it would make more sense to put them into a group.

But, I don't feel strongly about it either way.

R/pkg/inst/tests/testthat/test_mllib.R
```diff
@@ -1039,6 +1045,12 @@ test_that("spark.gbt", {
   expect_equal(iris2$NumericSpecies, as.double(collect(predict(m, df))$prediction))
   expect_equal(s$numFeatures, 5)
   expect_equal(s$numTrees, 20)
+
+  # spark.gbt classification can work on libsvm data
+  data <- read.df(absoluteSparkPath("data/mllib/sample_binary_classification_data.txt"),
+                source = "libsvm")
+  model <- spark.gbt(data, label ~ features, "classification")
+  expect_equal(summary(model)$numFeatures, 692)
```
felixcheung: ditto. good here, perhaps we test in glm or naiveBayes

mllib/src/main/scala/org/apache/spark/ml/r/RWrapperUtils.scala
```diff
@@ -41,5 +41,29 @@ object RWrapperUtils extends Logging {
...
+        s"using new name $newLabelName instead")
+      rFormula.setLabelCol(newLabelName)
+    }
+  }
+
+  /**
+   * Get the feature names and original labels from the schema
+   * of DataFrame transformed by RFormulaModel.
```
felixcheung: should this have @param doc ?

felixcheung:
Thanks for testing and fixing this. A couple of minor comments.

R/pkg/inst/tests/testthat/test_mllib.R
yanboliang: The reason behind this is there is no suitable libsvm data under ```data/mllib/``` for ```naiveBayes``` and ```glm```. Currently SparkR only support ```Bernoulli``` naive bayes which require all features data should be 0 or 1. And the binary classification data under ```data/mllib``` is collinear data, which is not very suitable to illustrate ```glm``` with binomial family. I'm working on another PR to fix this collinear issue, so I added test for ```randomForest``` and ```gbt```.

R/pkg/inst/tests/testthat/test_mllib.R
yanboliang: I don't have strong preference as well, I'll merge random forest tests into one as you suggested.

mllib/src/main/scala/org/apache/spark/ml/r/RWrapperUtils.scala
yanboliang: Good point. I strongly support to make this INFO since users should not concern about this issue. I'll update it soon.

mllib/src/main/scala/org/apache/spark/ml/r/RWrapperUtils.scala
yanboliang: It could have but not must, since it should be private. BTW, I found we should make object ```RWrapperUtils``` private[r], and will fix it soon.

yanboliang:
@felixcheung I addressed most of your comments, please let me know whether it's ok. Thanks.

felixcheung:
LGTM

yanboliang:
Merged into master and branch-2.1. Thanks for review.

PR #15852, opened 2016-11-11
[Spark-18187] [SQL] CompactibleFileStreamLog should not use "compactInterval" direcly with user setting.

ghost:
CompactibleFileStreamLog relys on "compactInterval" to detect a compaction batch. If the "compactInterval" is reset by user, CompactibleFileStreamLog will return wrong answer, resulting data loss. This PR procides a way to check the validity of 'compactInterval', and calculate an appropriate value.

## How was this patch tested?
When restart a stream, we change the 'spark.sql.streaming.fileSource.log.compactInterval' different with the former one.

The primary solution to this issue was given by @uncleGen 
Added extensions include an additional metadata field in OffsetSeq and CompactibleFileStreamLog APIs. @zsxwing


srowen:
(Update the title please; see others for format)

lw-lin:
@uncleGen @tcondie thanks for working on this.

My major concern is this approach might disallow changing the `compactInterval` once there were at least two compact files. Should we disallow it? Or as a alternative, what do you think of the approach taken in https://github.com/apache/spark/pull/15828?

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/CompactibleFileStreamLog.scala
```diff
@@ -63,7 +63,60 @@ abstract class CompactibleFileStreamLog[T <: AnyRef : ClassTag](
...
+    } else if (compactibleBatchIds.length == 1) {
+      // Case 3
+      val latestCompactBatchId = compactibleBatchIds(0).toInt
+      if (latestCompactBatchId + 1 <= defaultCompactInterval) {
+        interval = latestCompactBatchId + 1
+      } else if (defaultCompactInterval < (latestCompactBatchId + 1) / 2) {
+        // Find the first divisor >= default compact interval
+        def properDivisors(n: Int, min: Int) =
+          (min to n/2).filter(i => n % i == 0) :+ n
+
```
uncleGen: to => until ?

uncleGen:
LGTM overall. If this is accepted, then i will close #15827

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/CompactibleFileStreamLog.scala
```diff
@@ -63,7 +63,60 @@ abstract class CompactibleFileStreamLog[T <: AnyRef : ClassTag](
...
+  protected final lazy val compactInterval: Int = {
+    // SPARK-18187: "compactInterval" can be set by user via defaultCompactInterval.
+    // If there are existing log entries, then we should ensure a compatible compactInterval
+    // is used, irrespective of the defaultCompactInterval. There are three cases:
+    //
+    // 1. If there is no '.compact' file, we can use the default setting directly.
+    // 2. If there are two or more '.compact' files, we use the interval of patch id suffix with
+    // '.compact' as compactInterval. It is unclear whether this case will ever happen in the
+    // current code, since only the latest '.compact' file is retained i.e., other are garbage
+    // collected.
```
zsxwing: Right. Please update the comment accordingly.

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/CompactibleFileStreamLog.scala
```diff
@@ -63,7 +63,60 @@ abstract class CompactibleFileStreamLog[T <: AnyRef : ClassTag](
...
+      .reverse
+
+    // Case 1
+    var interval = defaultCompactInterval
+    if (compactibleBatchIds.length >= 2) {
+      // Case 2
+      val latestCompactBatchId = compactibleBatchIds(0)
+      val previousCompactBatchId = compactibleBatchIds(1)
+      interval = (latestCompactBatchId - previousCompactBatchId).toInt
+      logInfo(s"Compact interval case 2 = $interval")
```
zsxwing: nit: Please use a better message, like
```
Set the compact interval to XXX [the previous two batch Ids: XXX, XXX]
```

sql/core/src/test/scala/org/apache/spark/sql/streaming/FileStreamSourceSuite.scala
```diff
@@ -896,32 +898,35 @@ class FileStreamSourceSuite extends FileStreamSourceTest {
...
 
-  test("compacat metadata log") {
+  test("compact interval metadata log") {
     val _sources = PrivateMethod[Seq[Source]]('sources)
     val _metadataLog = PrivateMethod[FileStreamSourceLog]('metadataLog)
 
-    def verify(execution: StreamExecution)
-      (batchId: Long, expectedBatches: Int): Boolean = {
+    def verify(execution: StreamExecution, batchId: Long,
+               expectedBatches: Int, expectedCompactInterval: Int): Boolean = {
```
zsxwing: nit: the correct style should be
```Scala
def verify(
    execution: StreamExecution,
    batchId: Long,
    expectedBatches: Int,
    expectedCompactInterval: Int): Boolean = {
```

sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamTest.scala
```diff
@@ -161,7 +161,8 @@ trait StreamTest extends QueryTest with SharedSQLContext with Timeouts {
   /** Starts the stream, resuming if data has already been processed. It must not be running. */
   case class StartStream(
       trigger: Trigger = ProcessingTime(0),
-      triggerClock: Clock = new SystemClock)
+      triggerClock: Clock = new SystemClock,
+      pairs: mutable.Map[String, String] = mutable.Map.empty)
```
zsxwing: nit: use `Map[String, String]` instead since it won't be changed.

zsxwing: nit: it's better to rename `pairs` to `additionalConfs` for readability.

sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamTest.scala
```diff
@@ -330,14 +331,17 @@ trait StreamTest extends QueryTest with SharedSQLContext with Timeouts {
...
+          case StartStream(trigger, triggerClock, pairs) =>
             verify(currentStream == null, "stream already running")
             verify(triggerClock.isInstanceOf[SystemClock]
               || triggerClock.isInstanceOf[StreamManualClock],
               "Use either SystemClock or StreamManualClock to start the stream")
             if (triggerClock.isInstanceOf[StreamManualClock]) {
               manualClockExpectedTime = triggerClock.asInstanceOf[StreamManualClock].getTimeMillis()
             }
+
+            pairs.foreach(pair => spark.conf.set(pair._1, pair._2))
```
zsxwing: You need to also change the confs back at the end of this method to avoid affecting other tests sharing the same SparkSession.

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/CompactibleFileStreamLog.scala
```diff
@@ -63,7 +63,60 @@ abstract class CompactibleFileStreamLog[T <: AnyRef : ClassTag](
...
+    } else if (compactibleBatchIds.length == 1) {
+      // Case 3
+      val latestCompactBatchId = compactibleBatchIds(0).toInt
```
zsxwing: Could you pull this branch into a method in `object CompactibleFileStreamLog` so that it's easy to write tests for this complicated logic? And please add tests as well.

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/CompactibleFileStreamLog.scala
```diff
@@ -63,7 +63,60 @@ abstract class CompactibleFileStreamLog[T <: AnyRef : ClassTag](
...
+        def properDivisors(n: Int, min: Int) =
+          (min to n/2).filter(i => n % i == 0) :+ n
+
+        interval = properDivisors(latestCompactBatchId + 1, defaultCompactInterval)(0)
+      }
+      else {
+        // default compact interval > than any divisor other than latest compact id
+        interval = latestCompactBatchId + 1
+      }
+      logInfo(s"Compact interval case 3 = $interval")
```
zsxwing: nit: it's better to include all infos in the log, like
```
Set the compact interval to XXX [latestCompactBatchId: XXX, defaultCompactInterval: XXX]
```

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/CompactibleFileStreamLog.scala
zsxwing: I would use the following codes to avoid generating the real number sequence:

```Scala
def properDivisors(n: Int, min: Int) = (min to n/2).view.filter(n % _ == 0) :+ n

interval = properDivisors(latestCompactBatchId + 1, defaultCompactInterval).head
```

zsxwing (changes requested):
The approach looks good to me. I just left some style suggestions.

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/CompactibleFileStreamLog.scala
zsxwing: I think it's better to provide this one. It only outputs once and is pretty helpful when some bug happens here.

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/CompactibleFileStreamLog.scala
```diff
@@ -63,7 +63,60 @@ abstract class CompactibleFileStreamLog[T <: AnyRef : ClassTag](
 
   protected def isDeletingExpiredLog: Boolean
 
-  protected def compactInterval: Int
+  protected def defaultCompactInterval: Int
+
+  protected final lazy val compactInterval: Int = {
```
tcondie: FileStreamSourceLog uses compactInterval in multiple places. Please advise?

zsxwing: > FileStreamSourceLog uses compactInterval in multiple places. Please advise?

Sorry. My bad. Didn't notice that.

sql/core/src/test/scala/org/apache/spark/sql/streaming/FileStreamSourceSuite.scala
```diff
@@ -932,26 +940,28 @@ class FileStreamSourceSuite extends FileStreamSourceTest {
       ) {
         val fileStream = createFileStream("text", src.getCanonicalPath)
         val filtered = fileStream.filter($"value" contains "keep")
+        val updateConf = new mutable.HashMap[String, String]()
```
zsxwing: nit: Use `Map(SQLConf.FILE_SOURCE_LOG_COMPACT_INTERVAL.key -> "5")`

zsxwing (changes requested):
Looks good overall. Just some nits.

zsxwing:
LGTM. Thanks! Merging to master and 2.1.

PR #15853, opened 2016-11-11
[MINOR][SS] a minor refactor of getOffset in KafkaSource

CodingCat:

This PR factors the verbose code in the original implementation

## How was this patch tested?

regular unit test cases

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -901,7 +901,7 @@ Apart from these, the following properties are also available, and may be useful
   <td><code>spark.memory.offHeap.size</code></td>
   <td>0</td>
   <td>
-    The absolute amount of memory in bytes which can be used for off-heap allocation.
+    The absolute amount of memory (in terms by bytes) which can be used for off-heap allocation.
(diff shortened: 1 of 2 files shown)


CodingCat:
@zsxwing mind taking a look?

srowen:
It's OK by  me but is it really helping?

CodingCat:
Eh...depends on how you define help...remove some unnecessary code is the only benefit...I am ok if you close it...just a 1 min work when reading the source code

srowen:
It's not bad but there are a thousand things in the code that should be cleaned up. I think it could be OK if you sweep the code for similar patterns and fix all of some type. Just on principle and to be fair to other tiny PRs we've requested people close yes I think we maybe shouldn't change things ad-hoc like this if they have no real effect.

CodingCat:
Considering the current size of code base and the diversity of the required domain knowledge to ensure that the changes in all sub projects do not hurt, it is impossible to apply changes to the whole code base in once instead of fix issues progressively. Anyway, I will close it

PR #15857, opened 2016-11-11
[SPARK-18300][SQL] Do not apply foldable propagation with expand as a child.

hvanhovell:
The `FoldablePropagation` optimizer rule, pulls foldable values out from under an `Expand`. This breaks the `Expand` in two ways:

- It rewrites the output attributes of the `Expand`. We explicitly define output attributes for `Expand`, these are (unfortunately) considered as part of the expressions of the `Expand` and can be rewritten.
- Expand can actually change the column (it will typically re-use the attributes or the underlying plan). This means that we cannot safely propagate the expressions from under an `Expand`.

This PR fixes this and (hopefully) other issues by explicitly whitelisting allowed operators.

## How was this patch tested?
Added tests to `FoldablePropagationSuite` and to `SQLQueryTestSuite`.


gatorsmile:
cc @dongjoon-hyun @cloud-fan

sql/core/src/test/resources/sql-tests/inputs/group-by.sql
```diff
@@ -32,3 +32,6 @@ SELECT a + 1 + 1, COUNT(b) FROM testData GROUP BY a + 1;
 -- Aggregate with nulls.
 SELECT SKEWNESS(a), KURTOSIS(a), MIN(a), MAX(a), AVG(a), VARIANCE(a), STDDEV(a), SUM(a), COUNT(a)
 FROM testData;
+
+-- Aggregate with foldable input and multiple distinct groups.
+SELECT COUNT(DISTINCT b), COUNT(DISTINCT b, c) FROM (SELECT 1 AS a, 2 AS b, 3 AS c) GROUP BY a;
```
dongjoon-hyun: Ah. I see.

dongjoon-hyun:
Thank you for pining me, @gatorsmile . Thank you for fixing this, @hvanhovell .

dongjoon-hyun:
It looks good to me.

gatorsmile:
Could we add a comment above [`Union`](https://github.com/hvanhovell/spark/blob/d98f8f930330b368fbeade07d435197b0cdd2228/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/expressions.scala#L440) to explain why `Except` and `Intersect` are not handled? 

You know, my first reaction is why only handling `Union`? How about `Except` and `Intersect`? Then, I realize they have been replaced before this rule. That means, the order of these rules matters. : )

gatorsmile:
The test case [test("Propagate in subqueries of Union queries")](https://github.com/hvanhovell/spark/blob/d98f8f930330b368fbeade07d435197b0cdd2228/sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/optimizer/FoldablePropagationSuite.scala#L116-L130) is not resolvable. Thus, we do not have a test case to verify whether the propagation is stopped by [`Union`](https://github.com/hvanhovell/spark/blob/d98f8f930330b368fbeade07d435197b0cdd2228/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/expressions.scala#L440-L442).

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/expressions.scala
```diff
@@ -449,7 +452,7 @@ object FoldablePropagation extends Rule[LogicalPlan] {
           stop = true
           j
 
-        // These 3 operators take attributes as constructor parameters, and these attributes
+        // These 4 operators take attributes as constructor parameters, and these attributes
```
cloud-fan: shall we turn to whitelist? As I remember there have been more than 3 bugs reported on this optimizer rule.

dongjoon-hyun: If you think so, I have no objection for that.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/expressions.scala
```diff
@@ -428,43 +428,47 @@ object FoldablePropagation extends Rule[LogicalPlan] {
...
-          c
-        // For outer join, although its output attributes are derived from its children, they are
-        // actually different attributes: the output of outer join is not always picked from its
-        // children, but can also be null.
+        // Allow all leafnodes
+        case l: LeafNode =>
+          l
+
+        // Whitelist of all nodes we are allowed to apply this rule to.
+        case p @ (_: Project | _: Filter | _: SubqueryAlias | _: Aggregate | _: Window |
```
hvanhovell: @gatorsmile @dongjoon-hyun @cloud-fan I have created the following white list. Could you guys take a look?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/expressions.scala
```diff
@@ -428,43 +428,47 @@ object FoldablePropagation extends Rule[LogicalPlan] {
...
+                  _: Sample | _: GlobalLimit | _: LocalLimit | _: Generate | _: Distinct |
```
gatorsmile: `Distinct` will be replaced by our `Optimizer` at the beginning, right?

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/optimizer/FoldablePropagationSuite.scala
```diff
@@ -118,14 +118,30 @@ class FoldablePropagationSuite extends PlanTest {
       Seq(
         testRelation.select(Literal(1).as('x), 'a).select('x + 'a),
         testRelation.select(Literal(2).as('x), 'a).select('x + 'a)))
-      .select('x)
```
gatorsmile:
```
  test("Propagate in subqueries of Union queries") {
    val query = Union(
      Seq(
        testRelation.select(Literal(1).as('x), 'a).select('x, 'x + 'a),
        testRelation.select(Literal(2).as('x), 'a).select('x, 'x + 'a)))
      .select('x)
    val optimized = Optimize.execute(query.analyze)
    val correctAnswer = Union(
      Seq(
        testRelation.select(Literal(1).as('x), 'a)
          .select(Literal(1).as('x), (Literal(1).as('x) + 'a).as("(x + a)")),
        testRelation.select(Literal(2).as('x), 'a)
          .select(Literal(2).as('x), (Literal(2).as('x) + 'a).as("(x + a)"))))
      .select('x).analyze

    comparePlans(optimized, correctAnswer)
  }
```

How about this? The `x` (under `Union`) is transformed as the alias of a literal but the top `x` (above `Union`) is not propagated. This can validate the white list does not include `Union`.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/expressions.scala
hvanhovell: That is true. I only added the nodes of which I know they can be produced by analysis (like Distinct). I didn't want to rely on optimizer ordering. I can remove it if you want me to.

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/optimizer/FoldablePropagationSuite.scala
hvanhovell: Yeah let me change that

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/expressions.scala
gatorsmile: Let us keep it. I think your concern is right.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/expressions.scala
```diff
@@ -428,43 +428,47 @@ object FoldablePropagation extends Rule[LogicalPlan] {
...
-        case u: Union =>
-          stop = true
-          u
-        case c: Command =>
-          stop = true
-          c
-        // For outer join, although its output attributes are derived from its children, they are
-        // actually different attributes: the output of outer join is not always picked from its
-        // children, but can also be null.
+        // Allow all leafnodes
```
cloud-fan: `Allow all leafnodes` -> `Skip all leafnodes`?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/expressions.scala
```diff
@@ -428,43 +428,47 @@ object FoldablePropagation extends Rule[LogicalPlan] {
...
         // TODO(cloud-fan): It seems more reasonable to use new attributes as the output attributes
         // of outer join.
-        case j @ Join(_, _, LeftOuter | RightOuter | FullOuter, _) =>
+        case j @ Join(_, _, Inner, _) =>
+          j.transformExpressions(replaceFoldable)
+
+        // We can fold the projections an expand holds. However expand changes the output columns
+        // and often reuses the underlying attributes; so we cannot assume that a column is still
+        // foldable after the expand has been applied.
+        case expand: Expand if !stop =>
```
cloud-fan: should we add a TODO that `Expand` should always output new attributes?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/expressions.scala
hvanhovell: LeafNodes should not stop the folding process. That is what I am trying to dat.

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/optimizer/FoldablePropagationSuite.scala
```diff
@@ -118,14 +118,30 @@ class FoldablePropagationSuite extends PlanTest {
...
     val optimized = Optimize.execute(query.analyze)
     val correctAnswer = Union(
       Seq(
         testRelation.select(Literal(1).as('x), 'a).select((Literal(1).as('x) + 'a).as("(x + a)")),
         testRelation.select(Literal(2).as('x), 'a).select((Literal(2).as('x) + 'a).as("(x + a)"))))
-      .select('x).analyze
```
cloud-fan: how can this test pass before...

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/expressions.scala
cloud-fan: ah i see

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/optimizer/FoldablePropagationSuite.scala
gatorsmile: We are not checking whether it is resolvable or not. See the PR https://github.com/apache/spark/pull/15417

sql/core/src/test/resources/sql-tests/results/group-by.sql.out
```diff
@@ -131,3 +131,11 @@ FROM testData
 struct<skewness(CAST(a AS DOUBLE)):double,kurtosis(CAST(a AS DOUBLE)):double,min(a):int,max(a):int,avg(a):double,var_samp(CAST(a AS DOUBLE)):double,stddev_samp(CAST(a AS DOUBLE)):double,sum(a):bigint,count(a):bigint>
 -- !query 13 output
 -0.2723801058145729	-1.5069204152249134	1	3	2.142857142857143	0.8095238095238094	0.8997354108424372	15	7
+
+
+-- !query 14
+SELECT COUNT(DISTINCT b), COUNT(DISTINCT b, c) FROM (SELECT 1 AS a, 2 AS b, 3 AS c) GROUP BY a
```
hvanhovell: This PR fixes this bug (not really a regression, it always has been there). So I think I should at least add it.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/expressions.scala
hvanhovell: Improved the comment.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/expressions.scala
hvanhovell: I have added the TODO. I think that we really should revisit attribute reuse in general (it causes a lot of subtle bugs).

cloud-fan:
LGTM

gatorsmile:
LGTM

gatorsmile:

It sounds like we also need to backport it to 2.0

yhuai:
Seems this breaks the scala 2.10 build?

yhuai:
```
[error] [warn] /home/jenkins/workspace/spark-master-compile-sbt-scala-2.10/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/expressions.scala:439: Cannot check match for unreachability.
[error] (The analysis required more space than allowed. Please try with scalac -Dscalac.patmat.analysisBudget=512 or -Dscalac.patmat.analysisBudget=off.)
[error] [warn]       CleanupAliases(plan.transformUp {
[error] [warn] 
[error] [warn] /home/jenkins/workspace/spark-master-compile-sbt-scala-2.10/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/expressions.scala:439: Cannot check match for unreachability.
[error] (The analysis required more space than allowed. Please try with scalac -Dscalac.patmat.analysisBudget=512 or -Dscalac.patmat.analysisBudget=off.)
[error] [warn]       CleanupAliases(plan.transformUp {
[error] [warn] 
java.lang.RuntimeException: 2 fatal warnings
	at scala.sys.package$.error(package.scala:27)
	at SparkBuild$$anonfun$sharedSettings$18.apply(SparkBuild.scala:333)
	at SparkBuild$$anonfun$sharedSettings$18.apply(SparkBuild.scala:302)
	at scala.Function1$$anonfun$compose$1.apply(Function1.scala:47)
	at sbt.$tilde$greater$$anonfun$$u2219$1.apply(TypeFunctions.scala:40)
	at sbt.std.Transform$$anon$4.work(System.scala:63)
	at sbt.Execute$$anonfun$submit$1$$anonfun$apply$1.apply(Execute.scala:228)
	at sbt.Execute$$anonfun$submit$1$$anonfun$apply$1.apply(Execute.scala:228)
	at sbt.ErrorHandling$.wideConvert(ErrorHandling.scala:17)
	at sbt.Execute.work(Execute.scala:237)
	at sbt.Execute$$anonfun$submit$1.apply(Execute.scala:228)
	at sbt.Execute$$anonfun$submit$1.apply(Execute.scala:228)
	at sbt.ConcurrentRestrictions$$anon$4$$anonfun$1.apply(ConcurrentRestrictions.scala:159)
	at sbt.CompletionService$$anon$2.call(CompletionService.scala:28)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
[error] (catalyst/compile:compile) 2 fatal warnings
[error] Total time: 190 s, completed Nov 15, 2016 7:11:45 AM
Build step 'Execute shell' marked build as failure
Finished: FAILURE
```

hvanhovell:
I am going to revert this.

hvanhovell:
I am working on a fix for 2.10.

gatorsmile:
Thank you!

PR #15859, opened 2016-11-12
[SPARK-18416][Structured Streaming] Fixed temp file leak in state store

tdas:

StateStore.get() causes temporary files to be created immediately, even if the store is not used to make updates for new version. The temp file is not closed as store.commit() is not called in those cases, thus keeping the output stream to temp file open forever.

This PR fixes it by opening the temp file only when there are updates being made.

## How was this patch tested?

New unit test


zsxwing:
LGTM pending tests.

tdas:
@zsxwing can you take another look.

zsxwing:
LGTM. Merging to master, 2.1 and 2.0.

PR #15860, opened 2016-11-12
[SPARK-18418] Fix flags for make_binary_release for hadoop profile

holdenk:

Fix the flags used to specify the hadoop version

## How was this patch tested?

Manually tested as part of https://github.com/apache/spark/pull/15659 by having the build succeed.

cc @joshrosen

### DIFF
--- a/dev/create-release/release-build.sh
+++ b/dev/create-release/release-build.sh
@@ -187,10 +187,10 @@ if [[ "$1" == "package" ]]; then
   # We increment the Zinc port each time to avoid OOM's and other craziness if multiple builds
   # share the same Zinc server.
   FLAGS="-Psparkr -Phive -Phive-thriftserver -Pyarn -Pmesos"
-  make_binary_release "hadoop2.3" "-Phadoop2.3 $FLAGS" "3033" &
-  make_binary_release "hadoop2.4" "-Phadoop2.4 $FLAGS" "3034" &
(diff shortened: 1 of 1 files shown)


srowen (approved):

holdenk:
re-ping @joshrosen?

JoshRosen:
Looks good to me. I did a bit of digging and it looks like the bug that this fixes actually ended up causing silent failures in our nightly snapshot package builder:

- http://people.apache.org/~pwendell/spark-nightly/spark-branch-2.0-bin/spark-2.0.3-SNAPSHOT-2016_11_08_00_23-a39f8c1-bin/, which is built off of branch-2.0, has all of the expected packages
- http://people.apache.org/~pwendell/spark-nightly/spark-master-bin/spark-2.1.0-SNAPSHOT-2016_11_09_01_03-e256392-bin/ is missing the packages whose Maven profiles were invalid.

In the builder, the invalid profiles showed up as warnings in the logs:

```
+ VERSION='[WARNING] The requested profile "hadoop2.6" could not be activated because it does not exist.'
```

While in the longer term it might be nice to strengthen error-checks in that script to make sure that this doesn't silently break, I don't think it's necessary to do now or a super high priority in general (since any bug here will be caught prior to release), so this looks good to me and I'll merge it now.

PR #15863, opened 2016-11-12
[SPARK-18419][SQL] `JDBCRelation.insert` should not remove Spark options

dongjoon-hyun:

Currently, `JDBCRelation.insert` removes Spark options too early by mistakenly using `asConnectionProperties`. Spark options like `numPartitions` should be passed into `DataFrameWriter.jdbc` correctly. This bug have been **hidden** because `JDBCOptions.asConnectionProperties` fails to filter out the mixed-case options. This PR aims to fix both.

**JDBCRelation.insert**
```scala
override def insert(data: DataFrame, overwrite: Boolean): Unit = {
  val url = jdbcOptions.url
  val table = jdbcOptions.table
- val properties = jdbcOptions.asConnectionProperties
+ val properties = jdbcOptions.asProperties
  data.write
    .mode(if (overwrite) SaveMode.Overwrite else SaveMode.Append)
    .jdbc(url, table, properties)
```

**JDBCOptions.asConnectionProperties**
```scala
scala> import org.apache.spark.sql.execution.datasources.jdbc.JDBCOptions
scala> import org.apache.spark.sql.catalyst.util.CaseInsensitiveMap
scala> new JDBCOptions(Map("url" -> "jdbc:mysql://localhost:3306/temp", "dbtable" -> "t1", "numPartitions" -> "10")).asConnectionProperties
res0: java.util.Properties = {numpartitions=10}
scala> new JDBCOptions(new CaseInsensitiveMap(Map("url" -> "jdbc:mysql://localhost:3306/temp", "dbtable" -> "t1", "numPartitions" -> "10"))).asConnectionProperties
res1: java.util.Properties = {numpartitions=10}
```

## How was this patch tested?

Pass the Jenkins with a new testcase.

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JDBCOptions.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JDBCOptions.scala
@@ -41,6 +41,12 @@ class JDBCOptions(
       JDBCOptions.JDBC_TABLE_NAME -> table)))
   }
 
+  val asProperties: Properties = {
+    val properties = new Properties()
+    parameters.foreach { case (k, v) => properties.setProperty(k, v) }
+    properties
+  }
+
   val asConnectionProperties: Properties = {
     val properties = new Properties()
     // We should avoid to pass the options into properties. See SPARK-17776.
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JDBCOptions.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JDBCOptions.scala
@@ -44,7 +44,7 @@ class JDBCOptions(
   val asConnectionProperties: Properties = {
     val properties = new Properties()
     // We should avoid to pass the options into properties. See SPARK-17776.
-    parameters.filterKeys(!jdbcOptionNames.contains(_))
+    parameters.filterKeys(key => !jdbcOptionNames.contains(key.toLowerCase))
       .foreach { case (k, v) => properties.setProperty(k, v) }
     properties
   }
@@ -129,7 +129,7 @@ object JDBCOptions {
   private val jdbcOptionNames = ArrayBuffer.empty[String]
 
   private def newOption(name: String): String = {
-    jdbcOptionNames += name
+    jdbcOptionNames += name.toLowerCase
     name
   }
 
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JDBCOptions.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JDBCOptions.scala
@@ -20,8 +20,6 @@ package org.apache.spark.sql.execution.datasources.jdbc
 import java.sql.{Connection, DriverManager}
 import java.util.Properties
 
-import scala.collection.mutable.ArrayBuffer
-
 import org.apache.spark.sql.catalyst.util.CaseInsensitiveMap
 
 /**
@@ -41,16 +39,23 @@ class JDBCOptions(
       JDBCOptions.JDBC_TABLE_NAME -> table)))
   }
 
+  /**
+   * Returns a property with all options.
+   */
   val asProperties: Properties = {
     val properties = new Properties()
     parameters.foreach { case (k, v) => properties.setProperty(k, v) }
     properties
   }
 
+  /**
+   * Returns a property with all options except Spark internal data source options like `url`,
+   * `dbtable`, and `numPartition`. This should be used when invoking JDBC API like `Driver.connect`
+   * because each DBMS vendor has its own property list for JDBC driver. See SPARK-17776.
+   */
   val asConnectionProperties: Properties = {
     val properties = new Properties()
-    // We should avoid to pass the options into properties. See SPARK-17776.
-    parameters.filterKeys(key => !jdbcOptionNames.contains(key.toLowerCase))
+    parameters.filterKeys(key => !jdbcOptionNames(key.toLowerCase))
       .foreach { case (k, v) => properties.setProperty(k, v) }
     properties
   }
@@ -132,7 +137,7 @@ class JDBCOptions(
 }
 
 object JDBCOptions {
-  private val jdbcOptionNames = ArrayBuffer.empty[String]
+  private val jdbcOptionNames = collection.mutable.Set[String]()
 
   private def newOption(name: String): String = {
     jdbcOptionNames += name.toLowerCase
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JDBCRDD.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JDBCRDD.scala
@@ -54,7 +54,6 @@ object JDBCRDD extends Logging {
   def resolveTable(options: JDBCOptions): StructType = {
     val url = options.url
     val table = options.table
-    val properties = options.asConnectionProperties
     val dialect = JdbcDialects.get(url)
     val conn: Connection = JdbcUtils.createConnectionFactory(options)()
     try {
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JDBCRelation.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JDBCRelation.scala
@@ -131,7 +131,7 @@ private[sql] case class JDBCRelation(
   override def insert(data: DataFrame, overwrite: Boolean): Unit = {
     val url = jdbcOptions.url
     val table = jdbcOptions.table
-    val properties = jdbcOptions.asConnectionProperties
+    val properties = jdbcOptions.asProperties
     data.write
       .mode(if (overwrite) SaveMode.Overwrite else SaveMode.Append)
       .jdbc(url, table, properties)
--- a/sql/core/src/test/scala/org/apache/spark/sql/jdbc/JDBCSuite.scala
+++ b/sql/core/src/test/scala/org/apache/spark/sql/jdbc/JDBCSuite.scala
@@ -26,6 +26,7 @@ import org.scalatest.{BeforeAndAfter, PrivateMethodTester}
 
 import org.apache.spark.SparkFunSuite
 import org.apache.spark.sql.{DataFrame, Row}
+import org.apache.spark.sql.catalyst.util.CaseInsensitiveMap
 import org.apache.spark.sql.execution.DataSourceScanExec
 import org.apache.spark.sql.execution.command.ExplainCommand
 import org.apache.spark.sql.execution.datasources.LogicalRelation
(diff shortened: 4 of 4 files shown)


sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JDBCRDD.scala
```diff
@@ -54,7 +54,6 @@ object JDBCRDD extends Logging {
   def resolveTable(options: JDBCOptions): StructType = {
     val url = options.url
     val table = options.table
-    val properties = options.asConnectionProperties
```
dongjoon-hyun: This is unused.

HyukjinKwon:
Hi @dongjoon-hyun, I just happened to look at this PR and I am just curious. Would this be possible to override `filterKeys` in `CastInsensitiveMap` as something like below?

```scala
class CaseInsensitiveMap(map: Map[String, String]) extends Map[String, String]
  with MapLike[String, String, CaseInsensitiveMap] with Serializable {
  ...
  override def filterKeys(p: String => Boolean): Map[String, String] = {
    super.filterKeys({v: String => p(v.toLowerCase)})
  }
}
```

dongjoon-hyun:
Thank you for review, @HyukjinKwon . Let me see that.

HyukjinKwon:
Oh, it seems not. I just removed my suggestion. I will take a look at this again if it is possible in similar way. Thanks!

dongjoon-hyun:
Yes. It seems not to solve this problem. The problem is the following list is case-sensitive.
```
private val jdbcOptionNames = ArrayBuffer.empty[String]
```

gatorsmile:
Our user-specified JDBC options/parameters are case sensitive, right?

dongjoon-hyun:
Thank you for review, @gatorsmile .
In the Spark code, `JDBCOption` is always receiving `CaseInsensitiveMap`.
For example, here.
- https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala#L237

gatorsmile:
Not always. Try this code path.
 
```Scala
    df.write.format("jdbc")
      .option("URL", url1)
      .options(properties.asScala)
      .save()
```

I think we should make them consistent. Can you fix it in this PR?

dongjoon-hyun:
Oh, sure. I'll investigate it.

dongjoon-hyun:
I updated `DataSource` consistently to use `CaseInsensitiveMap` and am running tests.
After testing, this PR will include the followings. 
- Fix `DataSource` to use `CaseInsensitiveMap`.
- Fix `JDBCOptions.asConnectionProperties` to be case-insensitive.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
```diff
@@ -314,8 +311,7 @@ case class DataSource(
             catalogTable.get,
             catalogTable.get.stats.map(_.sizeInBytes.toLong).getOrElse(0L))
         } else {
-          new InMemoryFileIndex(
-            sparkSession, globbedPaths, options, partitionSchema)
+          new InMemoryFileIndex(sparkSession, globbedPaths, options, partitionSchema)
```
dongjoon-hyun: `InMemoryFileIndex` should use the original one because it extends `PartitioningAwareFileIndex` which uses the following line.
```scala
protected val hadoopConf = sparkSession.sessionState.newHadoopConfWithOptions(parameters)
```

Eventually, it created another **case-sensitive** Map from the `CaseInsensitiveMap`. In that case, it will fail to retrieve data from that map because the key value is changed as lowercases already in `CaseInsensitiveMap`.
```scala
def newHadoopConfWithOptions(options: Map[String, String]): Configuration = {
    val hadoopConf = newHadoopConf()
    options.foreach { case (k, v) =>
      if ((v ne null) && k != "path" && k != "paths") {
        hadoopConf.set(k, v)
      }
    }
    hadoopConf
  }
```

sql/core/src/test/scala/org/apache/spark/sql/jdbc/JDBCWriteSuite.scala
```diff
@@ -303,4 +303,13 @@ class JDBCWriteSuite extends SharedSQLContext with BeforeAndAfter {
     assert(e.contains("If 'partitionColumn' is specified then 'lowerBound', 'upperBound'," +
       " and 'numPartitions' are required."))
   }
+
+  test("SPARK-18419 JDBCOption keys should be case-insensitive") {
+    val df = spark.createDataFrame(sparkContext.parallelize(arr2x2), schema2)
+    df.write.format("jdbc")
+      .option("URL", url1)
```
viirya: nit: Use "Url" or "uRL" to show the case-insensitive more explicitly.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
```diff
@@ -157,11 +156,11 @@ case class DataSource(
   def createSource(metadataPath: String): Source = {
     providingClass.newInstance() match {
       case s: StreamSourceProvider =>
-        s.createSource(
-          sparkSession.sqlContext, metadataPath, userSpecifiedSchema, className, options)
+        s.createSource(sparkSession.sqlContext, metadataPath, userSpecifiedSchema, className,
+          caseInsensitiveOptions)
```
gatorsmile: The code style issue.
```
s.createSource(
  sparkSession.sqlContext,
  metadataPath, 
  userSpecifiedSchema, 
  className, 
  caseInsensitiveOptions)
```

sql/core/src/test/scala/org/apache/spark/sql/jdbc/JDBCWriteSuite.scala
dongjoon-hyun: Thank you for review, @viirya ! I updated it, too.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
dongjoon-hyun: Sure!

gatorsmile:
@rxin @cloud-fan Could you take a look at this PR, especially the external behavior impact? Thanks!

sql/core/src/test/scala/org/apache/spark/sql/jdbc/JDBCWriteSuite.scala
```diff
@@ -303,4 +303,13 @@ class JDBCWriteSuite extends SharedSQLContext with BeforeAndAfter {
     assert(e.contains("If 'partitionColumn' is specified then 'lowerBound', 'upperBound'," +
       " and 'numPartitions' are required."))
   }
+
+  test("SPARK-18419 JDBCOption keys should be case-insensitive") {
```
cloud-fan: is it consistent with other options like `JsonOtion`?

dongjoon-hyun:
Thank you for review, @cloud-fan .

Actually, this PR includes two different ones. The first commit is about a bug and the others (from seconds) is potential improvement.

First, I'll separate them to make each PRs more clearer.

dongjoon-hyun:
@cloud-fan and @gatorsmile .
For the `DataSource` options issue, I'm working on [SPARK-18433](https://issues.apache.org/jira/browse/SPARK-18433) for the followings.
- CSVOptions
- JDBCOptions
- JSONOptions
- OrcOptions

dongjoon-hyun:
I close this PR since this is already fixed.

dongjoon-hyun:
Oh, it's still there.
```scala
scala> new org.apache.spark.sql.execution.datasources.jdbc.JDBCOptions(Map("url" -> "jdbc:mysql://localhost:3306/temp", "dbtable" -> "t1", "numPartitions" -> "10")).asConnectionProperties
res0: java.util.Properties = {numpartitions=10}
```

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JDBCOptions.scala
```diff
@@ -129,7 +129,7 @@ object JDBCOptions {
   private val jdbcOptionNames = ArrayBuffer.empty[String]
```
dongjoon-hyun: `jdbcOptionNames` is the root cause.

dongjoon-hyun:
@cloud-fan and @gatorsmile .
How do you think about this? This is `asConnectionProperties` function specific issue.

dongjoon-hyun:
Also, cc @srowen . 
Sorry for the confusion.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JDBCRelation.scala
```diff
@@ -130,7 +130,7 @@ private[sql] case class JDBCRelation(
   override def insert(data: DataFrame, overwrite: Boolean): Unit = {
     val url = jdbcOptions.url
     val table = jdbcOptions.table
-    val properties = jdbcOptions.asConnectionProperties
```
dongjoon-hyun: Also, this should not use `asConnectionProperties`. `writer` needs options like `numPartitions`.

dongjoon-hyun:
I updated PR description and focus.

dongjoon-hyun:
The only failure seems to be irrelevant.
```
Had test failures in pyspark.streaming.tests with pypy; see logs.
```

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JDBCOptions.scala
cloud-fan: shall we make it a `Set`? seems we only use it for `contains` check

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JDBCOptions.scala
```diff
@@ -41,10 +41,16 @@ class JDBCOptions(
       JDBCOptions.JDBC_TABLE_NAME -> table)))
   }
 
+  val asProperties: Properties = {
+    val properties = new Properties()
+    parameters.foreach { case (k, v) => properties.setProperty(k, v) }
+    properties
+  }
+
   val asConnectionProperties: Properties = {
```
cloud-fan: let's add some document to explain when we should use `asConnectionProperties` and when `asProperties`

cloud-fan:
LGTM except 2 comments, including https://github.com/apache/spark/pull/15863/files#r90584811

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JDBCOptions.scala
dongjoon-hyun: Right. I'll update like that.

cloud-fan:

dongjoon-hyun:
Thank you for merging, @cloud-fan . Thank you for review, @gatorsmile , @viirya , @HyukjinKwon !

PR #15865, opened 2016-11-12
[SPARK-18420][BUILD] Fix the errors caused by lint check in Java

ConeyLiu:

Small fix, fix the errors caused by lint check in Java

- Clear unused objects and `UnusedImports`.
- Add comments around the method `finalize` of `NioBufferedFileInputStream`to turn off checkstyle.
- Cut the line which is longer than 100 characters into two lines.

## How was this patch tested?
Travis CI.
```
$ build/mvn -T 4 -q -DskipTests -Pyarn -Phadoop-2.3 -Pkinesis-asl -Phive -Phive-thriftserver install
$ dev/lint-java
```
Before:
```
Checkstyle checks failed at following occurrences:
[ERROR] src/main/java/org/apache/spark/network/util/TransportConf.java:[21,8] (imports) UnusedImports: Unused import - org.apache.commons.crypto.cipher.CryptoCipherFactory.
[ERROR] src/test/java/org/apache/spark/network/sasl/SparkSaslSuite.java:[516,5] (modifier) RedundantModifier: Redundant 'public' modifier.
[ERROR] src/main/java/org/apache/spark/io/NioBufferedFileInputStream.java:[133] (coding) NoFinalizer: Avoid using finalizer method.
[ERROR] src/main/java/org/apache/spark/sql/catalyst/expressions/UnsafeMapData.java:[71] (sizes) LineLength: Line is longer than 100 characters (found 113).
[ERROR] src/main/java/org/apache/spark/sql/catalyst/expressions/UnsafeArrayData.java:[112] (sizes) LineLength: Line is longer than 100 characters (found 110).
[ERROR] src/test/java/org/apache/spark/sql/catalyst/expressions/HiveHasherSuite.java:[31,17] (modifier) ModifierOrder: 'static' modifier out of order with the JLS suggestions.
[ERROR]src/main/java/org/apache/spark/examples/ml/JavaLogisticRegressionWithElasticNetExample.java:[64] (sizes) LineLength: Line is longer than 100 characters (found 103).
[ERROR] src/main/java/org/apache/spark/examples/ml/JavaInteractionExample.java:[22,8] (imports) UnusedImports: Unused import - org.apache.spark.ml.linalg.Vectors.
[ERROR] src/main/java/org/apache/spark/examples/ml/JavaInteractionExample.java:[51] (regexp) RegexpSingleline: No trailing whitespace allowed.
```

After:
```
$ build/mvn -T 4 -q -DskipTests -Pyarn -Phadoop-2.3 -Pkinesis-asl -Phive -Phive-thriftserver install
$ dev/lint-java
Using `mvn` from path: /home/travis/build/ConeyLiu/spark/build/apache-maven-3.3.9/bin/mvn
Checkstyle checks passed.
```

### DIFF
--- a/core/src/main/java/org/apache/spark/io/NioBufferedFileInputStream.java
+++ b/core/src/main/java/org/apache/spark/io/NioBufferedFileInputStream.java
@@ -130,8 +130,10 @@ public final class NioBufferedFileInputStream extends InputStream {
     StorageUtils.dispose(byteBuffer);
   }
 
+  //checkstyle.off: NoFinalizer
   @Override
   protected void finalize() throws IOException {
     close();
   }
+  //checkstyle.on: NoFinalizer
 }
--- a/dev/checkstyle.xml
+++ b/dev/checkstyle.xml
@@ -56,13 +56,13 @@
     If you wish to turn off checking for a section of code, you can put a comment in the source
     before and after the section, with the following syntax:
 
-      // checkstyle:off no.XXX (such as checkstyle:off NoFinalizer)
+      // checkstyle:off no.XXX (such as checkstyle.off: NoFinalizer)
       ...  // stuff that breaks the styles
       // checkstyle:on
     -->
     <module name="SuppressionCommentFilter">
-        <property name="offCommentFormat" value="checkstyle:off\: ([\w\|]+)"/>
-        <property name="onCommentFormat" value="checkstyle:on\: ([\w\|]+)"/>
+        <property name="offCommentFormat" value="checkstyle.off\: ([\w\|]+)"/>
+        <property name="onCommentFormat" value="checkstyle.on\: ([\w\|]+)"/>
         <property name="checkFormat" value="$1"/>
     </module>
 
--- a/dev/checkstyle.xml
+++ b/dev/checkstyle.xml
@@ -52,6 +52,20 @@
       <property name="file" value="dev/checkstyle-suppressions.xml"/>
     </module>
 
+    <!--
+    If you wish to turn off checking for a section of code, you can put a comment in the source
+    before and after the section, with the following syntax:
+
+      // checkstyle:off no.XXX (such as checkstyle:off NoFinalizer)
+      ...  // stuff that breaks the styles
+      // checkstyle:on
+    -->
+    <module name="SuppressionCommentFilter">
+        <property name="offCommentFormat" value="checkstyle:off\: ([\w\|]+)"/>
+        <property name="onCommentFormat" value="checkstyle:on\: ([\w\|]+)"/>
+        <property name="checkFormat" value="$1"/>
+    </module>
+
     <!-- Checks for whitespace                               -->
     <!-- See http://checkstyle.sf.net/config_whitespace.html -->
     <module name="FileTabCharacter">
@@ -168,5 +182,6 @@
         <module name="UnusedImports"/>
         <module name="RedundantImport"/>
         <module name="RedundantModifier"/>
+        <module name="FileContentsHolder"/>
     </module>
 </module>
--- a/examples/src/main/java/org/apache/spark/examples/ml/JavaInteractionExample.java
+++ b/examples/src/main/java/org/apache/spark/examples/ml/JavaInteractionExample.java
@@ -19,7 +19,6 @@ package org.apache.spark.examples.ml;
 
 import org.apache.spark.ml.feature.Interaction;
 import org.apache.spark.ml.feature.VectorAssembler;
-import org.apache.spark.ml.linalg.Vectors;
 import org.apache.spark.sql.*;
 import org.apache.spark.sql.types.DataTypes;
 import org.apache.spark.sql.types.Metadata;
@@ -48,7 +47,7 @@ public class JavaInteractionExample {
       RowFactory.create(5, 9, 2, 7, 10, 7, 3),
       RowFactory.create(6, 1, 1, 4, 2, 8, 4)
     );
-    
+
     StructType schema = new StructType(new StructField[]{
       new StructField("id1", DataTypes.IntegerType, false, Metadata.empty()),
       new StructField("id2", DataTypes.IntegerType, false, Metadata.empty()),
--- a/sql/catalyst/src/main/java/org/apache/spark/sql/catalyst/expressions/UnsafeArrayData.java
+++ b/sql/catalyst/src/main/java/org/apache/spark/sql/catalyst/expressions/UnsafeArrayData.java
@@ -109,7 +109,8 @@ public final class UnsafeArrayData extends ArrayData {
     // Read the number of elements from the first 8 bytes.
     final long numElements = Platform.getLong(baseObject, baseOffset);
     assert numElements >= 0 : "numElements (" + numElements + ") should >= 0";
-    assert numElements <= Integer.MAX_VALUE : "numElements (" + numElements + ") should <= Integer.MAX_VALUE";
+    assert numElements <= Integer.MAX_VALUE :
+      "numElements (" + numElements + ") should <= Integer.MAX_VALUE";
 
     this.numElements = (int)numElements;
     this.baseObject = baseObject;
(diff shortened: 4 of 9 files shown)


HyukjinKwon:
Hi @ConeyLiu, It seems 

```
[ERROR] src/main/java/org/apache/spark/io/NioBufferedFileInputStream.java:[133] (coding) NoFinalizer: Avoid using finalizer method.
```

is missed. If it is a required change, I guess we should disable the lint check for the lines.

ConeyLiu:
@HyukjinKwon I'm sorry for that, I've fixed it, please take a look. Thanks a lot.

core/src/main/java/org/apache/spark/io/NioBufferedFileInputStream.java
```diff
@@ -129,9 +129,4 @@ public synchronized void close() throws IOException {
     fileChannel.close();
     StorageUtils.dispose(byteBuffer);
   }
-
-  @Override
-  protected void finalize() throws IOException {
```
srowen: Why was this removed? it needs to be here for sure

sql/catalyst/src/test/java/org/apache/spark/sql/catalyst/expressions/HiveHasherSuite.java
```diff
@@ -28,7 +28,7 @@
 import java.util.Set;
 
 public class HiveHasherSuite {
-  private final static HiveHasher hasher = new HiveHasher();
+  private final HiveHasher hasher = new HiveHasher();
```
srowen: You made this non-static; that's not a style issue right? it may be OK but some of these changes don't seem to match the description

srowen:
They're caused by lint right? not checkstyle? and these are only Java, because that's all that's not automatically tested. Please update the title/description with more meaningful details, including the JIRA.

srowen (changes requested):

ConeyLiu:
@srowen  Thanks for the review, I've updated it.

HyukjinKwon:
@ConeyLiu Let us please make the title `[SPARK-18420][BUILD] Fix the errors caused by lint check in Java` for this PR and Jira you opened.

I think for the one below:

```
[ERROR] src/main/java/org/apache/spark/io/NioBufferedFileInputStream.java:[133] (coding) NoFinalizer: Avoid using finalizer method.
```

we should disable the lint for the lines because it seems a required one. I guess you should disable the lines or work around this.

Could we then remove `hasher` in `HiveHasherSuite.java` if it is not used? FWIW, I just checked that this private instance is not referenced in `HiveHasherSuite.java` for sure.

ConeyLiu:
@HyukjinKwon Thanks for the review and suggestion, I've updated it. Clear the unused object `hasher` and add suppression rules for the method `finalize` of `NioBufferedFileInputStream`. Please take a look, thanks a lot.

dev/checkstyle-suppressions.xml
```diff
@@ -30,6 +30,8 @@
 <suppressions>
     <suppress checks=".*"
               files="core/src/main/java/org/apache/spark/util/collection/TimSort.java"/>
+    <suppress checks="NoFinalizer"
+              files="core/src/main/java/org/apache/spark/io/NioBufferedFileInputStream.java"/>
```
HyukjinKwon: Ah, I thought we could disable it only for some lines. I am fine with this if it is only the way.

ConeyLiu: @HyukjinKwon  Also we could try `// scalastyle:off ...  // scalastyle:on`. I'll test this way.

srowen (approved):

sql/catalyst/src/main/java/org/apache/spark/sql/catalyst/expressions/UnsafeArrayData.java
```diff
@@ -109,7 +109,8 @@ public void pointTo(Object baseObject, long baseOffset, int sizeInBytes) {
     // Read the number of elements from the first 8 bytes.
     final long numElements = Platform.getLong(baseObject, baseOffset);
     assert numElements >= 0 : "numElements (" + numElements + ") should >= 0";
-    assert numElements <= Integer.MAX_VALUE : "numElements (" + numElements + ") should <= Integer.MAX_VALUE";
+    assert numElements <= Integer.MAX_VALUE :
+            "numElements (" + numElements + ") should <= Integer.MAX_VALUE";
```
HyukjinKwon: Hi @ConeyLiu, it is a super minor but it seems we should make the indentation for this (and the same instances here) double-spaced. I at least could find some examples for this such as in `UnsafeMemoryAllocator`

```java
    assert (memory.obj == null) :
      "baseObject not null; are you trying to use the off-heap allocator to free on-heap memory?";
```

or in `ShuffleSortDataFormat`

```java
    assert (length <= buffer.size()) :
      "the buffer is smaller than required: " + buffer.size() + " < " + length;
```

examples/src/main/java/org/apache/spark/examples/ml/JavaInteractionExample.java
```diff
@@ -48,8 +47,7 @@ public static void main(String[] args) {
       RowFactory.create(5, 9, 2, 7, 10, 7, 3),
       RowFactory.create(6, 1, 1, 4, 2, 8, 4)
     );
-
```
HyukjinKwon: If you are supposed to push more commits, let us revert this extra newline removal. I guess some guys don't like such changes. At least I was told twice before.

HyukjinKwon:
For sure, I ran 

```bash
$ ./dev/lint-java
Using `mvn` from path: .../mvn
Checkstyle checks passed.
```

It seems fine. It looks good to me for few minor comments above.

ConeyLiu:
@HyukjinKwon  I'm sorry for so late reply. I update the code and style followed by your suggestion, that's so useful. And also I change the method of disabling the check around `finalize`.  Please take a loot or give some suggestion.  Thanks a lot.

dev/checkstyle.xml
```diff
@@ -52,6 +52,20 @@
...
+      // checkstyle:off no.XXX (such as checkstyle.off: NoFinalizer)
+      ...  // stuff that breaks the styles
+      // checkstyle:on
+    -->
+    <module name="SuppressionCommentFilter">
+        <property name="offCommentFormat" value="checkstyle.off\: ([\w\|]+)"/>
+        <property name="onCommentFormat" value="checkstyle.on\: ([\w\|]+)"/>
+        <property name="checkFormat" value="$1"/>
+    </module>
+
```
HyukjinKwon: Ah, I am sorry @ConeyLiu. I should have taken a look ahead first. I actually rather meant agreeing the previous change if we should introduce such new property... Although I (as just one of contributors) think this might be worth, maybe this is orthogonal with what the PR tries to fix.

Maybe, I think we could keep this change and then propose a new separate PR later if other people also agree with this.

ConeyLiu: Ok, maybe we need more advices.

ConeyLiu:
@srowen hi owen, Can you give me some advice, and if so, what a great thing.

dev/checkstyle.xml
```diff
@@ -168,5 +182,6 @@
         <module name="UnusedImports"/>
         <module name="RedundantImport"/>
         <module name="RedundantModifier"/>
+        <module name="FileContentsHolder"/>
```
srowen: What is this needed for, the `checkstyle:off` mechanism?

ConeyLiu: @srowen `SuppressionCommentFilter`and `FileContentsHolder` must work together,http://stackoverflow.com/questions/5761188/checkstyle-suppressioncommentfilter-not-ignoring-specified-rule

srowen (approved):

HyukjinKwon:
I just checked for sure.

```
$ ./dev/lint-java
Using `mvn` from path: .../mvn
Checkstyle checks passed.
```

ConeyLiu:
@HyukjinKwon @srowen Thank you very much.

srowen:
Merged to master/2.1

PR #15868, opened 2016-11-13
[SPARK-18413][SQL] Add `maxConnections` JDBCOption

dongjoon-hyun:

This PR adds a new JDBCOption `maxConnections` which means the maximum number of simultaneous JDBC connections allowed. This option applies only to writing with coalesce operation if needed. It defaults to the number of partitions of RDD. Previously, SQL users cannot cannot control this while Scala/Java/Python users can use `coalesce` (or `repartition`) API. 

**Reported Scenario**

For the following cases, the number of connections becomes 200 and database cannot handle all of them.

```sql
CREATE OR REPLACE TEMPORARY VIEW resultview
USING org.apache.spark.sql.jdbc
OPTIONS (
  url "jdbc:oracle:thin:@10.129.10.111:1521:BKDB",
  dbtable "result",
  user "HIVE",
  password "HIVE"
);
-- set spark.sql.shuffle.partitions=200
INSERT OVERWRITE TABLE resultview SELECT g, count(1) AS COUNT FROM tnet.DT_LIVE_INFO GROUP BY g
```

## How was this patch tested?

Manual. Do the followings and see Spark UI.

**Step 1 (MySQL)**
```
CREATE TABLE t1 (a INT);
CREATE TABLE data (a INT);
INSERT INTO data VALUES (1);
INSERT INTO data VALUES (2);
INSERT INTO data VALUES (3);
```

**Step 2 (Spark)**
```scala
SPARK_HOME=$PWD bin/spark-shell --driver-memory 4G --driver-class-path mysql-connector-java-5.1.40-bin.jar
scala> sql("SET spark.sql.shuffle.partitions=3")
scala> sql("CREATE OR REPLACE TEMPORARY VIEW data USING org.apache.spark.sql.jdbc OPTIONS (url 'jdbc:mysql://localhost:3306/t', dbtable 'data', user 'root', password '')")
scala> sql("CREATE OR REPLACE TEMPORARY VIEW t1 USING org.apache.spark.sql.jdbc OPTIONS (url 'jdbc:mysql://localhost:3306/t', dbtable 't1', user 'root', password '', maxConnections '1')")
scala> sql("INSERT OVERWRITE TABLE t1 SELECT a FROM data GROUP BY a")
scala> sql("CREATE OR REPLACE TEMPORARY VIEW t1 USING org.apache.spark.sql.jdbc OPTIONS (url 'jdbc:mysql://localhost:3306/t', dbtable 't1', user 'root', password '', maxConnections '2')")
scala> sql("INSERT OVERWRITE TABLE t1 SELECT a FROM data GROUP BY a")
scala> sql("CREATE OR REPLACE TEMPORARY VIEW t1 USING org.apache.spark.sql.jdbc OPTIONS (url 'jdbc:mysql://localhost:3306/t', dbtable 't1', user 'root', password '', maxConnections '3')")
scala> sql("INSERT OVERWRITE TABLE t1 SELECT a FROM data GROUP BY a")
scala> sql("CREATE OR REPLACE TEMPORARY VIEW t1 USING org.apache.spark.sql.jdbc OPTIONS (url 'jdbc:mysql://localhost:3306/t', dbtable 't1', user 'root', password '', maxConnections '4')")
scala> sql("INSERT OVERWRITE TABLE t1 SELECT a FROM data GROUP BY a")
```

maxconnections

### DIFF
--- a/docs/sql-programming-guide.md
+++ b/docs/sql-programming-guide.md
@@ -1087,6 +1087,13 @@ the following case-sensitive options:
   </tr>
 
   <tr>
+     <td><code>maxConnections</code></td>
+     <td>
+       The number of JDBC connections, which specifies the maximum number of simultaneous JDBC connections that are allowed. This option applies only to writing. It defaults to the number of partitions of RDD.
+     </td>
+  </tr>
+
+  <tr>
      <td><code>isolationLevel</code></td>
      <td>
        The transaction isolation level, which applies to current connection. It can be one of <code>NONE<code>, <code>READ_COMMITTED<code>, <code>READ_UNCOMMITTED<code>, <code>REPEATABLE_READ<code>, or <code>SERIALIZABLE<code>, corresponding to standard transaction isolation levels defined by JDBC's Connection object, with default of <code>READ_UNCOMMITTED<code>. This option applies only to writing. Please refer the documentation in <code>java.sql.Connection</code>.
--- a/docs/sql-programming-guide.md
+++ b/docs/sql-programming-guide.md
@@ -1089,7 +1089,7 @@ the following case-sensitive options:
   <tr>
      <td><code>maxConnections</code></td>
      <td>
-       If set, limits the maximum number of JDBC connections that can be used. Only applies when writing. It works by limiting the operation's parallelism, which depends on the input's partition count. If its partition count exceeds this limit, the operation will coalesce the input to fewer partitions before writing.
+       The maximum number of concurrent JDBC connections that can be used, if set. Only applies when writing. It works by limiting the operation's parallelism, which depends on the input's partition count. If its partition count exceeds this limit, the operation will coalesce the input to fewer partitions before writing.
      </td>
   </tr>
 
--- a/docs/sql-programming-guide.md
+++ b/docs/sql-programming-guide.md
@@ -1089,7 +1089,7 @@ the following case-sensitive options:
   <tr>
      <td><code>maxConnections</code></td>
      <td>
-       The number of JDBC connections, which specifies the maximum number of simultaneous JDBC connections that are allowed. This option applies only to writing. It defaults to the number of partitions of RDD.
+       The number of JDBC connections, which specifies the maximum number of simultaneous JDBC connections that are allowed. This option applies only to writing with <code>coalesce<code/> operation if needed. It defaults to the number of partitions of RDD.
      </td>
   </tr>
 
--- a/docs/sql-programming-guide.md
+++ b/docs/sql-programming-guide.md
@@ -1089,7 +1089,7 @@ the following case-sensitive options:
   <tr>
      <td><code>maxConnections</code></td>
      <td>
-       The number of JDBC connections, which specifies the maximum number of simultaneous JDBC connections that are allowed. This option applies only to writing with <code>coalesce<code/> operation if needed. It defaults to the number of partitions of RDD.
+       If set, limits the maximum number of JDBC connections that can be used. Only applies when writing. It works by limiting the operation's parallelism, which depends on the input's partition count. If its partition count exceeds this limit, the operation will coalesce the input to fewer partitions before writing.
      </td>
   </tr>
 
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JDBCOptions.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JDBCOptions.scala
@@ -123,9 +123,9 @@ class JDBCOptions(
       case "SERIALIZABLE" => Connection.TRANSACTION_SERIALIZABLE
     }
   // the maximum number of connections
-  val maxConnections = parameters.getOrElse(JDBC_MAX_CONNECTIONS, null)
-  require(maxConnections == null || maxConnections.toInt > 0,
-    s"Invalid value `$maxConnections` for parameter `$JDBC_MAX_CONNECTIONS`. " +
+  val maxConnections = parameters.get(JDBC_MAX_CONNECTIONS)
+  require(maxConnections.isEmpty || maxConnections.get.toInt > 0,
+    s"Invalid value `${maxConnections.get}` for parameter `$JDBC_MAX_CONNECTIONS`. " +
       "The minimum value is 1.")
 }
 
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JDBCOptions.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JDBCOptions.scala
@@ -122,6 +122,11 @@ class JDBCOptions(
       case "REPEATABLE_READ" => Connection.TRANSACTION_REPEATABLE_READ
       case "SERIALIZABLE" => Connection.TRANSACTION_SERIALIZABLE
     }
+  // the maximum number of connections
+  val maxConnections = parameters.getOrElse(JDBC_MAX_CONNECTIONS, null)
+  require(maxConnections == null || maxConnections.toInt > 0,
+    s"Invalid value `$maxConnections` for parameter `$JDBC_MAX_CONNECTIONS`. " +
+      s"The minimum value is 1.")
 }
 
 object JDBCOptions {
@@ -144,4 +149,5 @@ object JDBCOptions {
   val JDBC_CREATE_TABLE_OPTIONS = newOption("createTableOptions")
   val JDBC_BATCH_INSERT_SIZE = newOption("batchsize")
   val JDBC_TXN_ISOLATION_LEVEL = newOption("isolationLevel")
+  val JDBC_MAX_CONNECTIONS = newOption("maxConnections")
 }
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JDBCOptions.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JDBCOptions.scala
@@ -123,8 +123,8 @@ class JDBCOptions(
       case "SERIALIZABLE" => Connection.TRANSACTION_SERIALIZABLE
     }
   // the maximum number of connections
-  val maxConnections = parameters.get(JDBC_MAX_CONNECTIONS)
-  require(maxConnections.isEmpty || maxConnections.get.toInt > 0,
+  val maxConnections = parameters.get(JDBC_MAX_CONNECTIONS).map(_.toInt)
+  require(maxConnections.isEmpty || maxConnections.get > 0,
     s"Invalid value `${maxConnections.get}` for parameter `$JDBC_MAX_CONNECTIONS`. " +
       "The minimum value is 1.")
 }
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JDBCOptions.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JDBCOptions.scala
@@ -126,7 +126,7 @@ class JDBCOptions(
   val maxConnections = parameters.getOrElse(JDBC_MAX_CONNECTIONS, null)
   require(maxConnections == null || maxConnections.toInt > 0,
     s"Invalid value `$maxConnections` for parameter `$JDBC_MAX_CONNECTIONS`. " +
-      s"The minimum value is 1.")
+      "The minimum value is 1.")
 }
 
 object JDBCOptions {
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
@@ -667,7 +667,14 @@ object JdbcUtils extends Logging {
     val getConnection: () => Connection = createConnectionFactory(options)
     val batchSize = options.batchSize
     val isolationLevel = options.isolationLevel
-    df.foreachPartition(iterator => savePartition(
+    val maxConnections = options.maxConnections
+    val repartitionedDF =
+      if (maxConnections != null && maxConnections.toInt < df.rdd.getNumPartitions) {
+        df.coalesce(maxConnections.toInt)
+      } else {
+        df
+      }
+    repartitionedDF.foreachPartition(iterator => savePartition(
       getConnection, table, iterator, rddSchema, nullTypes, batchSize, dialect, isolationLevel)
     )
   }
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
@@ -669,8 +669,8 @@ object JdbcUtils extends Logging {
     val isolationLevel = options.isolationLevel
     val maxConnections = options.maxConnections
     val repartitionedDF =
-      if (maxConnections != null && maxConnections.toInt < df.rdd.getNumPartitions) {
-        df.coalesce(maxConnections.toInt)
+      if (maxConnections.isDefined && maxConnections.get.toInt < df.rdd.getNumPartitions) {
+        df.coalesce(maxConnections.get.toInt)
       } else {
         df
       }
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
@@ -669,8 +669,8 @@ object JdbcUtils extends Logging {
     val isolationLevel = options.isolationLevel
     val maxConnections = options.maxConnections
     val repartitionedDF =
-      if (maxConnections.isDefined && maxConnections.get.toInt < df.rdd.getNumPartitions) {
-        df.coalesce(maxConnections.get.toInt)
+      if (maxConnections.isDefined && maxConnections.get < df.rdd.getNumPartitions) {
+        df.coalesce(maxConnections.get)
       } else {
         df
       }
--- a/sql/core/src/test/scala/org/apache/spark/sql/jdbc/JDBCWriteSuite.scala
+++ b/sql/core/src/test/scala/org/apache/spark/sql/jdbc/JDBCWriteSuite.scala
@@ -312,4 +312,16 @@ class JDBCWriteSuite extends SharedSQLContext with BeforeAndAfter {
       .options(properties.asScala)
       .save()
   }
+
+  test("SPARK-18413: Add `maxConnections` JDBCOption") {
+    val df = spark.createDataFrame(sparkContext.parallelize(arr2x2), schema2)
+    val e = intercept[IllegalArgumentException] {
+      df.write.format("jdbc")
+        .option("dbtable", "TEST.SAVETEST")
+        .option("url", url1)
+        .option(s"${JDBCOptions.JDBC_MAX_CONNECTIONS}", "0")
+        .save()
+    }.getMessage
+    assert(e.contains("Invalid value `0` for parameter `maxConnections`. The minimum value is 1"))
+  }
 }


lichenglin:
I'm sorry,my network is too bad to download dependencies from maven rep for building spark.
I have seen your PR and here is some suggestions：

1. I notice the PR has used CaseInsensitiveMap,becasuse "numPartitions" is still used by DataframeReader,we'd better to check whether jdbcreading work well.

2."Testing the Total Connections" may be difficulty,because when the tasks finish the connection will be closed.I have better way to check the  numPartitions if works, we can watch the last stage's task on spark ui,there should be "numPartitions" task in the last step.

dongjoon-hyun:
Thank you for review, @lichenglin .

For 1, Jenkins test that, too.
For 2, sure! It sounds reasonable and better.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
```diff
@@ -667,9 +667,15 @@ object JdbcUtils extends Logging {
     val getConnection: () => Connection = createConnectionFactory(options)
     val batchSize = options.batchSize
     val isolationLevel = options.isolationLevel
-    df.foreachPartition(iterator => savePartition(
-      getConnection, table, iterator, rddSchema, nullTypes, batchSize, dialect, isolationLevel)
-    )
+    if (options.numPartitions != null && options.numPartitions.toInt != df.rdd.getNumPartitions) {
+      df.repartition(options.numPartitions.toInt).foreachPartition(iterator => savePartition(
```
srowen: This repeats the `foreachPartition` part twice, when that could be outside the if-else.

I don't know enough to say whether we should add this property. It seems a little funny to set this globally to apply to all dataframes written to a database. I understand the use case is pure SQL, where perhaps that's the only option.

dongjoon-hyun: Thank you for review, @srowen .

First, the property `numPartitions` already exists in [JDBCOptions.scala: Optional parameters only for reading](https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JDBCOptions.scala#L73-L83).
This PR makes that option meaningful during write operation.

Second, for dataframe usecases, we can call `repartition` before saving to manage this. Actually, I asked @lichenglin that way. But, the main purpose of issue requested by @lichenglin is about providing pure SQL way to control the number of partitions for writing. In SQL, this looks reasonable to me.

dongjoon-hyun:
Hi, @srowen .
I addressed the comment and added an image in the PR description.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
```diff
@@ -667,7 +667,14 @@ object JdbcUtils extends Logging {
     val getConnection: () => Connection = createConnectionFactory(options)
     val batchSize = options.batchSize
     val isolationLevel = options.isolationLevel
-    df.foreachPartition(iterator => savePartition(
+    val numPartitions = options.numPartitions
+    val repartitionedDF = if (numPartitions != null &&
+        numPartitions.toInt != df.rdd.getNumPartitions) {
```
srowen: Tiny style point -- breaking the if statement that way looks a little funny to my eyes. I might do ...

```
val repartitionedDF =
  if (...) {
    ...
```

srowen (commented):
I still feel like I'd like someone else to comment on whether the property makes sense, but the change itself looks reasonable.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
dongjoon-hyun: Yep!

dongjoon-hyun:
Then, @gatorsmile .
Could you review this PR, too?

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
```diff
@@ -667,7 +667,14 @@ object JdbcUtils extends Logging {
     val getConnection: () => Connection = createConnectionFactory(options)
     val batchSize = options.batchSize
     val isolationLevel = options.isolationLevel
-    df.foreachPartition(iterator => savePartition(
+    val numPartitions = options.numPartitions
+    val repartitionedDF =
+      if (numPartitions != null && numPartitions.toInt != df.rdd.getNumPartitions) {
```
gatorsmile: Normally, based on my understanding, users only cares the maximal number of connections. Thus, no need to repartition it when `numPartitions.toInt >= df.rdd.getNumPartitions`, right?

gatorsmile:
`numPartitions` might be not a good name for this purpose. How about `maxConnections`?

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
```diff
@@ -667,7 +667,14 @@ object JdbcUtils extends Logging {
...
+        df.repartition(numPartitions.toInt)
```
gatorsmile: Is that ok to use `coalesce` here?

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
gatorsmile: Increasing the number of partitions can improve the insert performance in some scenarios, I think. However, `repartition` is not cheap.

dongjoon-hyun:
Thank you, @gatorsmile ! I'll update this soon.

dongjoon-hyun:
@gatorsmile , I addressed all comments.

gatorsmile:
retest this please

dongjoon-hyun:
Thank you for retriggering, @gatorsmile .

dongjoon-hyun:
Hi, @srowen and @gatorsmile .
How do you think about the updated PR?

docs/sql-programming-guide.md
```diff
@@ -1087,6 +1087,13 @@ the following case-sensitive options:
   </tr>
 
   <tr>
+     <td><code>maxConnection</code></td>
+     <td>
+       The maximum number of JDBC connections, which determines how many connections to occur. This can reduce the change of bottleneck on JDBC servers. This option applies only to writing. It defaults to the number of partitions of RDD.
```
gatorsmile: -> 
```
The number of JDBC connections, which specifies the maximum number of simultaneous JDBC connections that are allowed.  
```

dongjoon-hyun: Oops. I tried to mean `the chance of bottlenecks`.

dongjoon-hyun: I'll fix that. You could give me some better description for the whole. :)

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
```diff
@@ -358,7 +358,8 @@ case class DataSource(
 
     providingClass.newInstance() match {
       case dataSource: CreatableRelationProvider =>
-        dataSource.createRelation(sparkSession.sqlContext, mode, options, data)
+        dataSource.createRelation(
+          sparkSession.sqlContext, mode, new CaseInsensitiveMap(options), data)
```
gatorsmile: Can we keep this unchanged?

gatorsmile:
LGTM except one minor comment. Also CC @cloud-fan

gatorsmile:
Could we have a test case for this new parameter?

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
```diff
@@ -667,7 +667,14 @@ object JdbcUtils extends Logging {
     val getConnection: () => Connection = createConnectionFactory(options)
     val batchSize = options.batchSize
     val isolationLevel = options.isolationLevel
-    df.foreachPartition(iterator => savePartition(
+    val maxConnections = options.maxConnections
+    val repartitionedDF =
+      if (maxConnections != null && maxConnections.toInt < df.rdd.getNumPartitions) {
```
gatorsmile: If the inputs are less than 1, we should detect it at the beginning, right?

docs/sql-programming-guide.md
```diff
@@ -1087,6 +1087,13 @@ the following case-sensitive options:
   </tr>
 
   <tr>
+     <td><code>maxConnections</code></td>
+     <td>
+       The number of JDBC connections, which specifies the maximum number of simultaneous JDBC connections that are allowed. This option applies only to writing. It defaults to the number of partitions of RDD.
```
gatorsmile: What is the minimal value?

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
dongjoon-hyun: Sure. It's included in #15884 . After that, I can remove this.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
dongjoon-hyun: Yep. I'll add `require` like `batchsize`.

docs/sql-programming-guide.md
dongjoon-hyun: At the end of the description, I can add it. But, this will look different with other options in terms of level of depth. Is it okay?

dongjoon-hyun:
Just FYI, the only failure is irrelevant.
```
[info] StateStoreSuite:
[info] - maintenance *** FAILED *** (10 seconds, 120 milliseconds)
[info]   The code passed to eventually never returned normally. Attempted 648 times over 10.005010741 seconds. Last failure message: StateStore.isMaintenanceRunning was false Maintenance task is not running. (StateStoreSuite.scala:404)
```

dongjoon-hyun:
Hi, @gatorsmile .
Indeed, I really want to add a test case for this.
Could you give me some advice how to test this kind of features?

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
dongjoon-hyun: I removed this.

dongjoon-hyun:
I rebased and squashed to resolve conflicts. The PR is updated the followings.
- Revert `DataSource.scala` change.
- Rename `JDBC_MAX_CONNECTION` to `JDBC_MAX_CONNECTIONS`.
- Add `maxConnections` minimum value requirement.
```scala
scala> sql("CREATE OR REPLACE TEMPORARY VIEW t1 USING org.apache.spark.sql.jdbc OPTIONS (url 'jdbc:mysql://localhost:3306/t', dbtable 't1', user 'root', password '', maxConnections '-1')")
java.lang.IllegalArgumentException: requirement failed: Invalid value `-1` for parameter `maxConnections`. The minimum value is 1.
```

For the document and a test case, I need some advice.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
dongjoon-hyun: It's done.

dongjoon-hyun:
Could you review this PR again, @gatorsmile?

gatorsmile:
Yeah, add one negative test case and one positive test case for the newly added parm. For example, below is a negative test case.
```
    val e = intercept[java.lang.IllegalArgumentException] {
      df.write.format("jdbc")
        .option("url", url1)
        .option(s"${JDBCOptions.JDBC_MAX_CONNECTIONS}", "0")
        .save()
    }.getMessage
    assert(e.contains("Invalid value `0` for parameter `maxConnections`. The minimum value is 1"))
```

Very hard to test the number of connections Spark really makes, unless we change the source codes. Maybe we do not need it. You can manually checks the num of partitions of RDD.

docs/sql-programming-guide.md
gatorsmile: Ok. Let us keep it unchanged.

dongjoon-hyun:
Thank you, @gatorsmile . I added the negative test case.
But, for the positive test case, `DataFrameWriter` keeps it `private` and `save` function returns `Unit`.
```scala
private val df = ds.toDF()
```
Also, `JdbcUtils.saveTable` also returns `Unit` because the last statement is `foreachPartition`.
So, in the current function signature, we cannot check the partition of the repartitioned rdd.

gatorsmile:
LGTM. Maybe @srowen and @cloud-fan can double check the code. 

BTW, in the code path of regular SQL, we are able to explain the INSERT/CTAS. However, when users use the DataFrameWriter, this is not available. Maybe, in the future, we can add another option parameter (e.g., `explain`) for outputting the query plan? This is a low priority issue.

srowen:
OK I have an overall question about this approach. The point is to limit concurrent JDBC connections? this does that, but, requires repartitioning data to accomplish that, which is expensive. One could also just use a connection pool to cap this automatically and globally. That's, however, fairly driver-specific. OK, maybe that's a different topic.

Here though, why not just let the user repartition the DataFrame if desired? Is it worth a property?

gatorsmile:
Yeah, although we change the operator from `repartition` to `coalesce`, but yeah still costly. 

When users use JDBC source to connect RDBMS, the max number of concurrent connections is always a concern.  This is a common issue in the design of most RDBMS applications. We definitely can ask users to do it manually. I do not have a strong opinion on this. 

However, if we decide to not add such a property, we still need to document it. The current behavior is unclear to the external users. cc @rxin @cloud-fan @yhuai

gatorsmile:
FYI, SQOOP2 also [adds a feature](http://blog.cloudera.com/blog/2012/01/apache-sqoop-highlights-of-sqoop-2/) for limiting the max number of connections:

> By limiting the total number of physical Connections open at one time and with an option to disable Connections, resources can be managed.

dongjoon-hyun:
Sorry for late response, @gatorsmile and @srowen .

---

> However, when users use the DataFrameWriter, this is not available. Maybe, in the future, we can add another option parameter (e.g., explain) for outputting the query plan? This is a low priority issue.

Do you mean to print out the query plan of `df` of `DataFrameWriter` right?

* It's possible, but it should be done by some API like `save` instead of `option`.
* For the `save` operation, it makes another plan like `InsertIntoHadoopFsRelationCommand` or `foreachPartition` in JDBCUtils. So, it looks not easy to print out the whole logic.

---

> why not just let the user repartition the DataFrame if desired?

Yes. I asked the same question in the JIRA issue. But, the original demand of this issue is a **pure** SQL-based approach. We cannot call `repartition` in SQL language environment.

cloud-fan:
> We cannot call repartition in SQL language environment

This is a good point, we need to provide a workaround for sql users, and this PR did.

srowen:
I see, because a user can specify this option through SQL only? Ok, then that logic makes sense to me.

It's still kinda a coarse way to achieve a limit on concurrent connections. Some drivers let you configuring pooling in the driver string. That would cap connections globally, which is even better. Is that not an option? if it's readily available that's a much more direct solution.

At least, we ought to document that this is a potentially expensive operation; it doesn't actually just limit connections.

lichenglin:
In fact,I have idea .
We can change the param to "connections pool size on each executor",every partition task on the same executor use the same connection pool to save data.
we can create a map like ("jdbcviewid","jdbc connection pool").
So there is no need to call repartition.
I don't know if it works?

srowen:
Conceptually, it does seem nicer to resolve this with a connection pool. There is however some complexity in implementing our own connection pool.

I think you've pointed out a key issue: pools can only be per executor anyway, so, not sure that helps. Scratch that.

Now, this would only ever coalesce to smaller number of partitions right? that means it doesn't have to shuffle, right? especially if we make sure to set `shuffle = false` explicitly. In that case maybe this really is the best solution, especially if it's documented how it works.

dongjoon-hyun:
Thank you for review again, @cloud-fan , @liancheng , and @srowen .

I had the same opinion with @srowen . Also, let's think about the background of this issue.

Currently, in case of `Dataset`, the users uses `repartition` before saving, they don't use connection pool to workaround this issue. This PR is exactly the same with the previous Spark way. IMO, this option is not a worse behavior to do that.

dongjoon-hyun:
For document, may I add more description about the behavior to help the user understanding?

**BEFORE**
> The number of JDBC connections, which specifies the maximum number of simultaneous JDBC connections that are allowed. This option applies only to writing. It defaults to the number of partitions of RDD.

**AFTER**
> The number of JDBC connections, which specifies the maximum number of simultaneous JDBC connections that are allowed. This option applies only **to writing with additional `coalesce` operation**. It defaults to the number of partitions of RDD.

gatorsmile:
> This option applies only to writing with additional coalesce operation.

It is not clear. We do not always call `coalesce`. : )

dongjoon-hyun:
Ah, right. Now always!

gatorsmile:
Yeah, please improve the document comments and PR descriptions. 

It sounds like this is almost ready to merge after these changes.

dongjoon-hyun:
Documentation and PR descriptions are updated.

gatorsmile:
LGTM

docs/sql-programming-guide.md
```diff
@@ -1087,6 +1087,13 @@ the following case-sensitive options:
   </tr>
 
   <tr>
+     <td><code>maxConnections</code></td>
+     <td>
+       The number of JDBC connections, which specifies the maximum number of simultaneous JDBC connections that are allowed. This option applies only to writing with <code>coalesce<code/> operation if needed. It defaults to the number of partitions of RDD.
```
srowen: Pardon me for coming in with an additional comment here, but I would suggest yet another version of this text for more clarity, at least in my opinion:

If set, limits the maximum number of JDBC connections that can be used. Only applies when writing. It works by limiting the operation's parallelism, which depends on the input's partition count. If its partition count exceeds this limit, the operation will `coalesce` the input to fewer partitions before writing.

dongjoon-hyun: Thank you for revising that, @srowen ! This looks different from the other parameter description. This is the whole text, right?
> If set, limits the maximum number of JDBC connections that can be used. Only applies when writing. It works by limiting the operation's parallelism, which depends on the input's partition count. If its partition count exceeds this limit, the operation will coalesce the input to fewer partitions before writing.

dongjoon-hyun: It's done.

dongjoon-hyun:
Hi, @srowen .
Could you review this PR again?

srowen (approved):
I'm OK with it for 2.2. Let's see if anyone else has a comment before tomorrow

dongjoon-hyun:
Thank you, @srowen !

gatorsmile:
retest this please

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JDBCOptions.scala
```diff
@@ -122,6 +122,11 @@ class JDBCOptions(
       case "REPEATABLE_READ" => Connection.TRANSACTION_REPEATABLE_READ
       case "SERIALIZABLE" => Connection.TRANSACTION_SERIALIZABLE
     }
+  // the maximum number of connections
+  val maxConnections = parameters.getOrElse(JDBC_MAX_CONNECTIONS, null)
```
dongjoon-hyun: Sure! No problem. I'll update this PR again.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JDBCOptions.scala
```diff
@@ -122,6 +122,11 @@ class JDBCOptions(
       case "REPEATABLE_READ" => Connection.TRANSACTION_REPEATABLE_READ
       case "SERIALIZABLE" => Connection.TRANSACTION_SERIALIZABLE
     }
+  // the maximum number of connections
+  val maxConnections = parameters.get(JDBC_MAX_CONNECTIONS)
```
cloud-fan: `parameters.get(JDBC_MAX_CONNECTIONS).map(_.toInt)`, then we don't need to call `toInt` everytime we need to use this option.

dongjoon-hyun: Oops. Thank you!

dongjoon-hyun:
It seems not a test failure.
```
    main()
KeyError: -9
```

dongjoon-hyun:
Retest this please.

dongjoon-hyun:
Hi, @cloud-fan .
If there is something to do more, please let me know again.
Thank you so much.

cloud-fan:
LGTM, waiting @srowen for final sign off

srowen:
Merged to master

dongjoon-hyun:
Thank you for review and merging, @srowen , @gatorsmile , @cloud-fan , @lichenglin !

rxin:
Hm - wouldn't a more general solution be to introduce a parallelism hint or command (basically the coalesce equivalent) in SQL?

dongjoon-hyun:
Hi, @rxin .
Do you mean changing SQL syntax like `BROADCAST HINT` PR we tried before?

rxin:
Yes - otherwise it can get pretty confusing.

I'd argue the user has already explicitly defined the degree of parallelism in the JDBC source using other options. Having this maxConnections option doing "coalesce" is pretty weird.

dongjoon-hyun:
Could you give the example of existing method you mention?
> the user has already explicitly defined the degree of parallelism in the JDBC source using other options.

rxin:
Don't they already specify "numPartitions" ?

dongjoon-hyun:
It's read-only option. At the first commit, I reuse that option for writing. But, as you see, it changed according the review advices.

dongjoon-hyun:
Let me find the correct code. (It's weird I attached a wrong code example.)

rxin:
Got it - so this is to enforce the write side: in the future please make it clear this is for the write side. 

Coming back to the discussion: shouldn't we just use numPartitions, given it is already there? Why pick another one to confuse users?

dongjoon-hyun:
The code seems to be gone during resolving conflicts. This is the my comment at the first review.

> Thank you for review, @srowen .
> 
> First, the property numPartitions already exists in JDBCOptions.scala: Optional parameters only for reading.
> This PR makes that option meaningful during write operation.

rxin:
Basically right now if I were to define a table, I'd need to define both the read and the write side, using numPartitions and maxPartitions. Even if we want to have two separate configs for read vs write, it is absolutely not clear which one is which.

If I were to just hand you two config options: numPartitions and maxPartitions. Would you be able to tell me why one is for read and the other is for write?

dongjoon-hyun:
The rational is to give a name which has a clear meaning. Also, I agreed with the following advice.

> numPartitions might be not a good name for this purpose. How about maxConnections?

If you want to merge them together into `numPartition`, I'll do as a follow-up PR.

gatorsmile:
My fault. In my initial thoughts, `numPartitions` does not indicate the max allowed number of JDBC connections, and the write path does not always create the exact number of partitions. We only change the num of partitions if the number is larger than the threshold. 

I do not have a strong opinon on either way.

dongjoon-hyun:
Then, @rxin and @gatorsmile .
I'll make a PR to merge `maxConnection` into `numPartitions`.

PR #15869, opened 2016-11-13
[YARN][DOC] Remove non-Yarn specific configurations from running-on-yarn.md

weiqingy:

Remove `spark.driver.memory`, `spark.executor.memory`,  `spark.driver.cores`, and `spark.executor.cores` from `running-on-yarn.md` as they are not Yarn-specific, and they are also defined in`configuration.md`.

## How was this patch tested?
Build passed & Manually check.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -156,13 +156,6 @@ of the most common options to set are:
   </td>
 </tr>
 <tr>
- <td><code>spark.executor.instances</code></td>
-  <td><code>2</code></td>
-  <td>
-    The number of executors for static allocation. With <code>spark.dynamicAllocation.enabled</code>, the initial set of executors will be at least this large.
-  </td>
-</tr>
-<tr>
   <td><code>spark.extraListeners</code></td>
   <td>(none)</td>
   <td>
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -156,6 +156,13 @@ of the most common options to set are:
   </td>
 </tr>
 <tr>
+ <td><code>spark.executor.instances</code></td>
+  <td><code>2</code></td>
+  <td>
+    The number of executors for static allocation. With <code>spark.dynamicAllocation.enabled</code>, the initial set of executors will be at least this large.
+  </td>
+</tr>
+<tr>
   <td><code>spark.extraListeners</code></td>
   <td>(none)</td>
   <td>
--- a/docs/running-on-yarn.md
+++ b/docs/running-on-yarn.md
@@ -118,19 +118,6 @@ To use a custom metrics.properties for the application master and executors, upd
   </td>
 </tr>
 <tr>
-  <td><code>spark.driver.memory</code></td>
-  <td>1g</td>
-  <td>
-    Amount of memory to use for the driver process, i.e. where SparkContext is initialized.
-    (e.g. <code>1g</code>, <code>2g</code>).
-
-    <br /><em>Note:</em> In client mode, this config must not be set through the <code>SparkConf</code>
-    directly in your application, because the driver JVM has already started at that point.
-    Instead, please set this through the <code>--driver-memory</code> command line option
-    or in your default properties file.
-  </td>
-</tr>
-<tr>
   <td><code>spark.driver.cores</code></td>
   <td><code>1</code></td>
   <td>
@@ -238,20 +225,6 @@ To use a custom metrics.properties for the application master and executors, upd
   <td>1 in YARN mode, all the available cores on the worker in standalone mode.</td>
   <td>
     The number of cores to use on each executor. For YARN and standalone mode only.
-  </td>
-</tr>
-<tr>
- <td><code>spark.executor.instances</code></td>
-  <td><code>2</code></td>
-  <td>
-    The number of executors for static allocation. With <code>spark.dynamicAllocation.enabled</code>, the initial set of executors will be at least this large.
-  </td>
-</tr>
-<tr>
-  <td><code>spark.executor.memory</code></td>
-  <td>1g</td>
-  <td>
-    Amount of memory to use per executor process (e.g. <code>2g</code>, <code>8g</code>).
   </td>
 </tr>
 <tr>
@@ -493,6 +466,20 @@ To use a custom metrics.properties for the application master and executors, upd
   Java Regex to filter the log files which match the defined exclude pattern
   and those log files will not be aggregated in a rolling fashion. If the log file
   name matches both the include and the exclude pattern, this file will be excluded eventually.
+  </td>
+</tr>
+<tr>
+  <td><code>spark.yarn.report.interval</code></td>
+  <td>1s</td>
+  <td>
+  Interval between reports of the current app status in Yarn cluster mode.
+  </td>
+</tr>
+<tr>
+  <td><code>spark.yarn.services</code></td>
+  <td>Nil</td>
+  <td>
+  A comma-separated list of class names of services to add to the scheduler.
   </td>
 </tr>
 </table>
--- a/docs/running-on-yarn.md
+++ b/docs/running-on-yarn.md
@@ -118,15 +118,6 @@ To use a custom metrics.properties for the application master and executors, upd
   </td>
 </tr>
 <tr>
-  <td><code>spark.driver.cores</code></td>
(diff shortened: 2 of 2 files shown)


docs/configuration.md
```diff
@@ -156,6 +156,13 @@ of the most common options to set are:
   </td>
 </tr>
 <tr>
+ <td><code>spark.executor.instances</code></td>
```
srowen: I think this is actually YARN-specific right now? At least it is according to `SparkSubmit.scala`

docs/running-on-yarn.md
```diff
@@ -118,19 +118,6 @@ To use a custom metrics.properties for the application master and executors, upd
   </td>
 </tr>
 <tr>
-  <td><code>spark.driver.memory</code></td>
```
srowen: Yes, I see that these are duplicated in the YARN docs and main config docs. So are things like `spark.driver.cores`, but, the info on what it means in YARN is slightly more specific to YARN in the YARN docs. The `spark.driver.memory` doc isn't different, but, for consistency, maybe does not hurt to leave a note about these key, generic props in the YARN docs.

docs/running-on-yarn.md
```diff
@@ -495,6 +468,20 @@ To use a custom metrics.properties for the application master and executors, upd
   name matches both the include and the exclude pattern, this file will be excluded eventually.
   </td>
 </tr>
+<tr>
+  <td><code>spark.yarn.report.interval</code></td>
```
srowen: I am not sure whether these are meant to be exposed to users -- not secret exactly, but not sure if they're to be advertised. Maybe it's fine to document. I'm neutral.

docs/running-on-yarn.md
```diff
@@ -495,6 +468,20 @@ To use a custom metrics.properties for the application master and executors, upd
...
+  <td>1s</td>
+  <td>
+  Interval between reports of the current app status in Yarn cluster mode.
```
srowen: Yarn -> YARN

docs/running-on-yarn.md
```diff
@@ -495,6 +468,20 @@ To use a custom metrics.properties for the application master and executors, upd
...
+  </td>
+</tr>
+<tr>
+  <td><code>spark.yarn.services</code></td>
+  <td>Nil</td>
```
srowen: As far the user is concerned, I think the best way to express the default is "(none)". `Nil` is an implementation detail

srowen (changes requested):
CC @vanzin

docs/configuration.md
weiqingy: Thanks for reviewing this PR, @srowen 

Actually I am confused by this. From the information below, it should not be used for yarn only:

- https://github.com/apache/spark/blob/master/core/src/test/scala/org/apache/spark/SparkContextSuite.scala#L403
- In `configuration.md`:
```
<tr>
  <td><code>spark.dynamicAllocation.initialExecutors</code></td>
  <td><code>spark.dynamicAllocation.minExecutors</code></td>
  <td>
    Initial number of executors to run if dynamic allocation is enabled.
    <br /><br />
    If `--num-executors` (or `spark.executor.instances`) is set and larger than this value, it will
    be used as the initial number of executors.
  </td>
</tr>
```   

- Currently`spark.executor.instances` is defined in [config/package.scala](https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/internal/config/package.scala#L86). If it is for YARN-specific, it should be defined in `yarn/config.scala` instead of `config/package.scala`.

docs/running-on-yarn.md
weiqingy: Yes, in Yarn mode, the properties `spark.driver.cores`, `spark.driver.memory`, and `spark.executor.memory` are often used. But if searching them in Spark code (i.e. the information below), we will find they are not Yarn-specific.
- https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/deploy/SparkSubmit.scala#L448
- https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/deploy/SparkSubmit.scala#L477
- https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/deploy/SparkSubmit.scala#L469
- https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/deploy/SparkSubmit.scala#L469
- https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/deploy/rest/StandaloneRestServer.scala#L135
- https://github.com/apache/spark/blob/master/mesos/src/main/scala/org/apache/spark/deploy/rest/mesos/MesosRestServer.scala#L92
- ...

docs/running-on-yarn.md
weiqingy: Done.

docs/running-on-yarn.md
weiqingy: Done.

weiqingy:
Hi, @srowen I have replied your comments and updated the PR.  Could you please review it again? Thanks.

rxin:
Can you just put some of the more trivial pull requests into one, e.g. the two documentation update ones?

docs/running-on-yarn.md
tgravescs: If it wasn't documented, it was on purpose.  If users find it a useful config we can expose it.

docs/running-on-yarn.md
```diff
@@ -495,6 +468,20 @@ To use a custom metrics.properties for the application master and executors, upd
...
 </tr>
+<tr>
+  <td><code>spark.yarn.report.interval</code></td>
+  <td>1s</td>
+  <td>
+  Interval between reports of the current app status in YARN cluster mode.
+  </td>
+</tr>
+<tr>
+  <td><code>spark.yarn.services</code></td>
```
tgravescs: I don't think we should expose this.  I think this was added with a PR to start adding integration with ATS which I don't know we officially support yet, this config is just for testing at this point.  @steveloughran to verify.

weiqingy:
@rxin Thanks for commenting. The reason I created two PRs is that I noticed these issues not at the same time. This one is still during discussion, and I am not sure if it will be merged, however, the other one I am sure it will be merged. I did not think too much about this. I just though if I find an issue, no matter how small it is, just submit a PR directly. But, I will pay attention next time. So do you mean I should close PR #15886 and merge it into this PR? I can do that. Thanks.

docs/running-on-yarn.md
weiqingy: Thanks for pointing out this.

docs/running-on-yarn.md
weiqingy: Thanks for this information.

weiqingy:
Hi, @tgravescs What do you think about `spark.driver.memory`, `spark.executor.memory`, and `spark.executor.instances`? 
- Move `spark.executor.instances` from `running-on-yarn.md` to `configuration.md`.
- Remove spark.driver.memory and `spark.executor.memory` from `running-on-yarn.md`(they are also defined in `configuration.md`).
Actually I noticed this issue when I was working on PR #15563. So if they are documented in current way on purpose, I can just close this PR. I would like your suggestion. Thanks.

tgravescs:
- spark.driver.memory and spark.executor.memory is good to remove from yarn side as its duplicate since they were added for others. 
spark.driver.cores and spark.executor.cores could also be removed from yarn config docs as they are in the general ones now too.

- spark.executor.instances is still only used in YARN.  --num-executors is also YARN only.  I'm not sure the exact behavior of spark.dynamicAllocation.initialExecutors on mesos or standalone mode to know if the description for that should be updated for those.

weiqingy:
@tgravescs Thanks for the reply. I have updated the PR.

srowen (approved):
OK, but can you now update the title and description? "Update Yarn doc" is too generic, and the description doesn't match the change now.

steveloughran:
The plugin point is more generic than ATS integration; it lets you stick anything in to come up in the driver. Weakness: it's actually yarn specific; I could imagine uses in standalone too. The update interval flag? I don't remember this or what it does. Sorry

weiqingy:
Hi, @srowen I have updated the title and description. Thanks.

srowen:
Merged to master/2.1

PR #15871, opened 2016-11-13
[SPARK-17116][Pyspark] Allow parameters to be {string,value} dict at runtime

aditya1702:

This pull allows a dictionary containing string as keys to be passed in the fit() method. Presently it only allows an instance of the Estimator to be passed as key.

## How was this patch tested?

This patch was manually tested on a local PC.

### DIFF
--- a/python/pyspark/ml/base.py
+++ b/python/pyspark/ml/base.py
@@ -59,6 +59,12 @@ class Estimator(Params):
             return [self.fit(dataset, paramMap) for paramMap in params]
         elif isinstance(params, dict):
             if params:
+                if isinstance(params.keys()[0],str):
+                    param_new = dict()
+                    for param, value in params.items():
+                        p = getattr(self, param)
+                        param_new.update({p: value})
+                    params=param_new
                 return self.copy(params)._fit(dataset)
             else:
                 return self._fit(dataset)


aditya1702:
@srowen Could you take a look and check if this improvement is proper?

python/pyspark/ml/base.py
```diff
@@ -59,6 +59,12 @@ def fit(self, dataset, params=None):
             return [self.fit(dataset, paramMap) for paramMap in params]
         elif isinstance(params, dict):
             if params:
+                if type(params.keys()[0]) is str:
```
HyukjinKwon: Shouldn't we just use `isinstance` instead?

HyukjinKwon:
I think we need a test here maybe and update the argument description if this change is legitimate.

HyukjinKwon:
cc @holdenk

aditya1702:
@HyukjinKwon this is my first time with contributing in Spark and so I didnt know if we add tests to each pull. If it is proper I will add the tests :)

python/pyspark/ml/base.py
HyukjinKwon: Ah, I meant, `isinstance(params.keys()[0],str)`. Direct type comparison is a strict type comparison which does not take care of subclasses. I believe it is a common pattern to check the types. For example,

```python
>>> class A(str):
...     pass
...
>>> isinstance(A(), str)
True
>>> type(A()) is str
False
```

This might be major. You can imagine giving an argument as `OrderedDict` which extends `dict`.

Another reason is consistency across codebase. I believe `isinstance` is more widely used in the codes.

aditya1702: @HyukjinKwon Ah I get you. I will make the changes

aditya1702:
@HyukjinKwon I have updated the code. Could you please take a look

python/pyspark/ml/base.py
```diff
@@ -59,6 +59,12 @@ def fit(self, dataset, params=None):
             return [self.fit(dataset, paramMap) for paramMap in params]
         elif isinstance(params, dict):
             if params:
+                if isinstance(params.keys()[0],str):
+                    param_new = dict()
+                    for param, value in params.items():
+                        p = getattr(self, param)
+                        param_new.update({p: value})
+                    params=param_new
```
HyukjinKwon: Could we just make it just like the one as below?

```python
params = dict([(p, v) for p, v in params.items()])
```

BTW, I am not supposed to decide what should be added into Spark. We can wait for committer's review :)

aditya1702: @HyukjinKwon Ok sure.

holdenk:
Thanks for working on this! So at first glance I'm not super sure if we want to add this (is this being done to be more closely related to the skicitlearn apis or something?)

A good next step would probably be adding some tests for this as well :) Once we've got some tests we can ping some of the Python ML people and see what they think.

aditya1702:
@holdenk Sure. I will add the tests. I saw that as per the issue opened the setParams was taking in string dict while fit() method was not. Hence I thought it would be an improvement to do that to the fit() method as well.

holdenk:
@aditya1702 - thanks that sounds like a good reason for us to be consistent between the two. Let me know how adding the tests goes :)

HyukjinKwon:
Hi @aditya1702, do you mind if I ask whether you are still working on this? Maybe it should be closed for now if you are currently not able to work on this further. It seems inactive for few months.

PR #15875, opened 2016-11-14
[SPARK-18428][DOC] Update docs for GraphX

zhengruifeng:
1, Add link of `VertexRDD` and `EdgeRDD`
2, Notify in `Vertex and Edge RDDs` that not all methods are listed
3, `VertexID` -> `VertexId`

## How was this patch tested?
No tests, only docs is modified

### DIFF
--- a/docs/graphx-programming-guide.md
+++ b/docs/graphx-programming-guide.md
@@ -11,6 +11,7 @@ description: GraphX graph processing library guide for Spark SPARK_VERSION_SHORT
 <!-- All the documentation links  -->
 
 [EdgeRDD]: api/scala/index.html#org.apache.spark.graphx.EdgeRDD
+[VertexRDD]: api/scala/index.html#org.apache.spark.graphx.VertexRDD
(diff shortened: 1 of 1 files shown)


rxin:
So I think what we should really say is that this is just an incomplete list, and then link to the API docs. It is not great to duplicate API docs in two places.

zhengruifeng:
@rxin 
This has already been notified in `Summary List of Operators` : `Note that some function signatures have been simplified (e.g., default arguments and type constraints removed) and some more advanced functionality has been removed so please consult the API docs for the official list of operations.`

But it's not said in `VertexRDD`. I will revert this change and and a notification in `VertexRDD`.

zhengruifeng:
@rxin I update this PR to notify in `Vertex and Edge RDDs` that not all methods are listed and add link of `VertexRDD` and `EdgeRDD`.
BTW, use `VertexId` instead of `VertexID`

rxin:
Thanks - merging in master.

PR #15876, opened 2016-11-14
[SPARK-11496][GraphX][FOLLOWUP] Add param checking for runParallelPersonalizedPageRank

zhengruifeng:
add the param checking to keep in line with other algos

## How was this patch tested?
existing tests


srowen (approved):
Seems OK. @dbtsai ?

I know it's sometimes hard to see and comment on an issue before it's merged, but ideally we don't regularly have "FOLLOWUP" PRs like this. I've seen a number lately. Anything you can do to add your helpful comments while they're still in review would be great.

dbtsai:
LGTM. Thanks. Merged into master and 2.1 branch.

@srowen Are you referring the MLOR PR which has many followup PRs? If so, the changes in the main MLOR PR is very big, and many of the issues can be broke into smaller tasks. To avoid dealing with many conflicts, we decided to work on those smaller tasks which had been identified in the reviewing process in the early development of 2.1 so we can move faster (some of them may require extra infrastructure outside the algorithm so we don't want to do it together). Do you think we could have a guideline or checklist so the reviewers can consistently check how to do error handling (there are different type of exception being thrown for similar error) or coding style in addition to correctness of the algorithms? Any suggestion? Thanks.

PR #15877, opened 2016-11-14
[SPARK-18429] [SQL] implement a new Aggregate for CountMinSketch

wzhfy:

This PR implements a new Aggregate to generate count min sketch, which is a wrapper of CountMinSketch.

## How was this patch tested?

add test cases

### DIFF
--- a/common/sketch/src/main/java/org/apache/spark/util/sketch/CountMinSketchImpl.java
+++ b/common/sketch/src/main/java/org/apache/spark/util/sketch/CountMinSketchImpl.java
@@ -25,6 +25,7 @@ import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
 import java.io.OutputStream;
 import java.io.Serializable;
+import java.math.BigDecimal;
 import java.util.Arrays;
 import java.util.Random;
 
@@ -152,6 +153,16 @@ class CountMinSketchImpl extends CountMinSketch implements Serializable {
   public void add(Object item, long count) {
     if (item instanceof String) {
       addString((String) item, count);
+    } else if (item instanceof BigDecimal) {
+      addString(((BigDecimal) item).toString(), count);
+    } else if (item instanceof byte[]) {
+      addBinary((byte[]) item, count);
+    } else if (item instanceof Float) {
+      addLong(Float.floatToIntBits((Float) item), count);
+    } else if (item instanceof Double) {
+      addLong(Double.doubleToLongBits((Double) item), count);
+    } else if (item instanceof Boolean) {
+      addLong(((Boolean) item) ? 1L : 0L, count);
     } else {
       addLong(Utils.integralToLong(item), count);
     }
@@ -216,10 +227,6 @@ class CountMinSketchImpl extends CountMinSketch implements Serializable {
     return ((int) hash) % width;
   }
 
-  private static int[] getHashBuckets(String key, int hashCount, int max) {
-    return getHashBuckets(Utils.getBytesFromUTF8String(key), hashCount, max);
-  }
-
   private static int[] getHashBuckets(byte[] b, int hashCount, int max) {
     int[] result = new int[hashCount];
     int hash1 = Murmur3_x86_32.hashUnsafeBytes(b, Platform.BYTE_ARRAY_OFFSET, b.length, 0);
@@ -233,7 +240,18 @@ class CountMinSketchImpl extends CountMinSketch implements Serializable {
   @Override
   public long estimateCount(Object item) {
     if (item instanceof String) {
-      return estimateCountForStringItem((String) item);
+      return estimateCountForBinaryItem(Utils.getBytesFromUTF8String((String) item));
+    } else if (item instanceof BigDecimal) {
+      return estimateCountForBinaryItem(
+        Utils.getBytesFromUTF8String(((BigDecimal) item).toString()));
+    } else if (item instanceof byte[]) {
+      return estimateCountForBinaryItem((byte[]) item);
+    } else if (item instanceof Float) {
+      return estimateCountForLongItem(Float.floatToIntBits((Float) item));
+    } else if (item instanceof Double) {
+      return estimateCountForLongItem(Double.doubleToLongBits((Double) item));
+    } else if (item instanceof Boolean) {
+      return estimateCountForLongItem(((Boolean) item) ? 1L : 0L);
     } else {
       return estimateCountForLongItem(Utils.integralToLong(item));
     }
@@ -247,7 +265,7 @@ class CountMinSketchImpl extends CountMinSketch implements Serializable {
     return res;
   }
 
-  private long estimateCountForStringItem(String item) {
+  private long estimateCountForBinaryItem(byte[] item) {
     long res = Long.MAX_VALUE;
     int[] buckets = getHashBuckets(item, depth, width);
     for (int i = 0; i < depth; ++i) {
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/FunctionRegistry.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/FunctionRegistry.scala
@@ -261,6 +261,7 @@ object FunctionRegistry {
     expression[VarianceSamp]("var_samp"),
     expression[CollectList]("collect_list"),
     expression[CollectSet]("collect_set"),
+    expression[CountMinSketchAgg]("count_min_sketch"),
 
     // string functions
     expression[Ascii]("ascii"),
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/FunctionRegistry.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/FunctionRegistry.scala
@@ -262,7 +262,6 @@ object FunctionRegistry {
     expression[CollectList]("collect_list"),
     expression[CollectSet]("collect_set"),
     expression[CountMinSketchAgg]("count_min_sketch"),
-    expression[CountMinSketchAgg]("cmsketch"),
 
     // string functions
     expression[Ascii]("ascii"),
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/FunctionRegistry.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/FunctionRegistry.scala
@@ -262,6 +262,7 @@ object FunctionRegistry {
     expression[CollectList]("collect_list"),
     expression[CollectSet]("collect_set"),
     expression[CountMinSketchAgg]("count_min_sketch"),
+    expression[CountMinSketchAgg]("cmsketch"),
 
     // string functions
     expression[Ascii]("ascii"),
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/CountMinSketchAgg.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/CountMinSketchAgg.scala
@@ -23,7 +23,6 @@ import org.apache.spark.sql.catalyst.InternalRow
 import org.apache.spark.sql.catalyst.analysis.TypeCheckResult
 import org.apache.spark.sql.catalyst.analysis.TypeCheckResult.{TypeCheckFailure, TypeCheckSuccess}
 import org.apache.spark.sql.catalyst.expressions.{Expression, ExpressionDescription}
-import org.apache.spark.sql.catalyst.util.GenericArrayData
 import org.apache.spark.sql.types._
 import org.apache.spark.unsafe.types.UTF8String
 import org.apache.spark.util.sketch.CountMinSketch
@@ -32,6 +31,8 @@ import org.apache.spark.util.sketch.CountMinSketch
  * This function returns a count-min sketch of a column with the given esp, confidence and seed.
  * A count-min sketch is a probabilistic data structure used for summarizing streams of data in
  * sub-linear space, which is useful for equality predicates and join size estimation.
+ * The result returned by the function is an array of bytes, which should be deserialized to a
+ * `CountMinSketch` before usage.
  *
  * @param child child expression that can produce column value with `child.eval(inputRow)`
  * @param epsExpression relative error, must be positive
@@ -61,6 +62,11 @@ case class CountMinSketchAgg(
     this(child, epsExpression, confidenceExpression, seedExpression, 0, 0)
   }
 
+  // Mark as lazy so that they are not evaluated during tree transformation.
+  private lazy val eps: Double = epsExpression.eval().asInstanceOf[Double]
+  private lazy val confidence: Double = confidenceExpression.eval().asInstanceOf[Double]
+  private lazy val seed: Int = seedExpression.eval().asInstanceOf[Int]
+
   override def checkInputDataTypes(): TypeCheckResult = {
     val defaultCheck = super.checkInputDataTypes()
     if (defaultCheck.isFailure) {
@@ -72,16 +78,16 @@ case class CountMinSketchAgg(
     } else if (epsExpression.eval() == null || confidenceExpression.eval() == null ||
       seedExpression.eval() == null) {
       TypeCheckFailure("The eps, confidence or seed provided should not be null")
+    } else if (eps <= 0D) {
+      TypeCheckFailure(s"Relative error must be positive (current value = $eps)")
+    } else if (confidence <= 0D || confidence >= 1D) {
+      TypeCheckFailure(s"Confidence must be within range (0.0, 1.0) (current value = $confidence)")
     } else {
-      // parameter validity will be checked in CountMinSketchImpl
       TypeCheckSuccess
     }
   }
 
   override def createAggregationBuffer(): CountMinSketch = {
-    val eps: Double = epsExpression.eval().asInstanceOf[Double]
-    val confidence: Double = confidenceExpression.eval().asInstanceOf[Double]
-    val seed: Int = seedExpression.eval().asInstanceOf[Int]
     CountMinSketch.create(eps, confidence, seed)
   }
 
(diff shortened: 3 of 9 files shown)


wzhfy:
cc @rxin

rxin:
cc @liancheng

wzhfy:
retest this please

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/CountMinSketchAgg.scala
```diff
@@ -0,0 +1,131 @@
...
+      // UTF8String is a spark sql type, while CountMinSketch accepts String type
+      buffer.add(if (value.isInstanceOf[UTF8String]) value.toString else value)
+    }
+  }
+
+  override def merge(buffer: CountMinSketch, input: CountMinSketch): Unit = {
+    buffer.mergeInPlace(input)
+  }
+
+  override def eval(buffer: CountMinSketch): Any = new GenericArrayData(serialize(buffer))
```
hvanhovell: Is this an array of bytes?

wzhfy: yes

hvanhovell: It is better just to return the byte array and to change the datatype into a BinaryType

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/CountMinSketchAgg.scala
```diff
@@ -0,0 +1,131 @@
...
+  def this(
+      child: Expression,
+      epsExpression: Expression,
+      confidenceExpression: Expression,
+      seedExpression: Expression) = {
+    this(child, epsExpression, confidenceExpression, seedExpression, 0, 0)
+  }
+
+  override def checkInputDataTypes(): TypeCheckResult = {
+    val defaultCheck = super.checkInputDataTypes()
```
hvanhovell: I don't think we need to check this (the super class does not implement this).

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/CountMinSketchAgg.scala
```diff
@@ -0,0 +1,131 @@
...
+      seedExpression.eval() == null) {
+      TypeCheckFailure("The eps, confidence or seed provided should not be null")
+    } else {
+      // parameter validity will be checked in CountMinSketchImpl
+      TypeCheckSuccess
+    }
+  }
+
+  override def createAggregationBuffer(): CountMinSketch = {
+    val eps: Double = epsExpression.eval().asInstanceOf[Double]
```
hvanhovell: Should we cache this in lazy vals? I am not sure about the performance implications.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/CountMinSketchAgg.scala
```diff
@@ -0,0 +1,131 @@
...
+    val seed: Int = seedExpression.eval().asInstanceOf[Int]
+    CountMinSketch.create(eps, confidence, seed)
+  }
+
+  override def update(buffer: CountMinSketch, input: InternalRow): Unit = {
+    val value = child.eval(input)
+    // ignore empty rows
+    if (value != null) {
+      // UTF8String is a spark sql type, while CountMinSketch accepts String type
+      buffer.add(if (value.isInstanceOf[UTF8String]) value.toString else value)
```
hvanhovell: How bad would it be to add support for UTF8 string to CMS? Or to pass the UTF8 byte array to CMS?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/CountMinSketchAgg.scala
```diff
@@ -0,0 +1,131 @@
...
+import org.apache.spark.sql.catalyst.expressions.{Expression, ExpressionDescription}
+import org.apache.spark.sql.catalyst.util.GenericArrayData
+import org.apache.spark.sql.types._
+import org.apache.spark.unsafe.types.UTF8String
+import org.apache.spark.util.sketch.CountMinSketch
+
+/**
+ * This function returns a count-min sketch of a column with the given esp, confidence and seed.
+ * A count-min sketch is a probabilistic data structure used for summarizing streams of data in
+ * sub-linear space, which is useful for equality predicates and join size estimation.
```
hvanhovell: Maybe something on the return type? A developer should know how to work with these bytes.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/CountMinSketchAgg.scala
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  override def withNewMutableAggBufferOffset(newMutableAggBufferOffset: Int): CountMinSketchAgg =
+    copy(mutableAggBufferOffset = newMutableAggBufferOffset)
+
+  override def withNewInputAggBufferOffset(newInputAggBufferOffset: Int): CountMinSketchAgg =
+    copy(inputAggBufferOffset = newInputAggBufferOffset)
+
+  override def inputTypes: Seq[AbstractDataType] = {
+    // currently `CountMinSketch` supports integral and string types
```
hvanhovell: Should we expand this?

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/aggregate/CountMinSketchAggSuite.scala
```diff
@@ -0,0 +1,284 @@
...
+      group2.foreach { index =>
+        val input = InternalRow(allItems(sampledItemIndices(index)))
+        agg.update(group2Buffer, input)
+      }
+
+      val mergeBuffer = agg.createAggregationBuffer()
+      agg.merge(mergeBuffer, group1Buffer)
+      agg.merge(mergeBuffer, group2Buffer)
+      checkResult(agg.eval(mergeBuffer), allItems, exactFreq)
+    }
```
hvanhovell: This might also be a good place to test merging in a different order, and the merging of an empty partition.

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/aggregate/CountMinSketchAggSuite.scala
```diff
@@ -0,0 +1,284 @@
...
+      checkResult(agg.eval(mutableAggBuffer), allItems, exactFreq)
+    }
+  }
+
+  private def checkResult[T: ClassTag](
+      result: Any,
+      data: Array[T],
+      exactFreq: Map[T, Long]): Unit = {
+    result match {
+      case arrayData: ArrayData =>
```
hvanhovell: Add `case _ => fail("unexpected return type")` to have a nicer error when something goed wrong there

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/CountMinSketchAgg.scala
```diff
@@ -0,0 +1,131 @@
...
+
+  override def checkInputDataTypes(): TypeCheckResult = {
+    val defaultCheck = super.checkInputDataTypes()
+    if (defaultCheck.isFailure) {
+      defaultCheck
+    } else if (!epsExpression.foldable || !confidenceExpression.foldable ||
+      !seedExpression.foldable) {
+      TypeCheckFailure(
+        "The eps, confidence or seed provided must be a literal or constant foldable")
+    } else if (epsExpression.eval() == null || confidenceExpression.eval() == null ||
```
hvanhovell: Should we also check for negative eps and confidence values?

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/aggregate/CountMinSketchAggSuite.scala
```diff
@@ -0,0 +1,284 @@
...
+
+    // parameters are out of the valid range
+    Seq(0.0, -1000.0).foreach { invalidEps =>
+      val invalidAgg = new CountMinSketchAgg(
+        childExpression,
+        epsExpression = Literal(invalidEps),
+        confidenceExpression = Literal(confidence),
+        seedExpression = Literal(seed))
+      val err = intercept[IllegalArgumentException] {
+        invalidAgg.createAggregationBuffer()
```
hvanhovell: See my comment in the CMS agg. This is too late to throw such an error. I'd rather have driver side errors then executor side errors.

hvanhovell (changes requested):
This looks pretty good! I have left a few minor comments. Also consider to register this aggregate in the FunctionRegistry and to add it to functions.scala.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/CountMinSketchAgg.scala
wzhfy: Yes, that's better, thanks!

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/CountMinSketchAgg.scala
wzhfy: ExpectsInputTypes.checkInputDataTypes() checks validity of input types, right?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/CountMinSketchAgg.scala
wzhfy: Ok, i'll change them to lazy vals

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/CountMinSketchAgg.scala
hvanhovell: That is fair.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/CountMinSketchAgg.scala
wzhfy: ok, I wrote this in usage, but I'll add it here too, thanks.

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/aggregate/CountMinSketchAggSuite.scala
wzhfy: Yes, we should have driver side errors, thanks.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/CountMinSketchAgg.scala
wzhfy: @rxin suggested that for unsupported types, we hash it before count min sketch, i.e. CountMinSketchAgg(hash(col)).

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/CountMinSketchAgg.scala
wzhfy: Yes, we should pass the byte array to CMS.

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/aggregate/CountMinSketchAggSuite.scala
wzhfy: Ok, I'll also test these.

rxin:
yes please register a count_min_sketch and alias cmsketch in FunctionRegistry.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/CountMinSketchAgg.scala
```diff
@@ -0,0 +1,146 @@
...
+    val value = child.eval(input)
+    // ignore empty rows
+    if (value != null) {
+      child.dataType match {
+        // For string type, we can get bytes of our `UTF8String` directly, and call the `addBinary`
+        // instead of `addString` to avoid unnecessary conversion.
+        case StringType => buffer.addBinary(value.asInstanceOf[UTF8String].getBytes)
+        case ByteType => buffer.addLong(value.asInstanceOf[Byte])
+        case ShortType => buffer.addLong(value.asInstanceOf[Short])
+        case IntegerType => buffer.addLong(value.asInstanceOf[Int])
```
hvanhovell: Add DateType?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/CountMinSketchAgg.scala
```diff
@@ -0,0 +1,146 @@
...
+
+  override def createAggregationBuffer(): CountMinSketch = {
+    CountMinSketch.create(eps, confidence, seed)
+  }
+
+  override def update(buffer: CountMinSketch, input: InternalRow): Unit = {
+    val value = child.eval(input)
+    // ignore empty rows
+    if (value != null) {
+      child.dataType match {
```
rxin: although i don't know if it matters much here given we are going to run it through many hash functions.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/FunctionRegistry.scala
```diff
@@ -261,6 +261,8 @@ object FunctionRegistry {
     expression[VarianceSamp]("var_samp"),
     expression[CollectList]("collect_list"),
     expression[CollectSet]("collect_set"),
+    expression[CountMinSketchAgg]("count_min_sketch"),
+    expression[CountMinSketchAgg]("cmsketch"),
```
rxin: actually i take my word back. let's add only count_min_sketch. I don't think it's worth having an alias given this is sketch is difficult to consume (returning some binary)

sql/core/src/main/scala/org/apache/spark/sql/functions.scala
```diff
@@ -182,6 +182,36 @@ object functions {
...
+   * Aggregate function: returns a count-min sketch of a column.
+   *
+   * @param eps relative error, must be positive
+   * @param confidence confidence, must be positive and less than 1.0
+   * @param seed random seed
+   *
+   * @group agg_funcs
+   * @since 2.2.0
+   */
+  def count_min_sketch(e: Column, eps: Double, confidence: Double, seed: Int): Column = {
```
rxin: let's not add these for now.

wzhfy:
Hi @hvanhovell @rxin, I've updated this pr, does it look good to you now?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/CountMinSketchAgg.scala
```diff
@@ -0,0 +1,150 @@
...
+    copy(mutableAggBufferOffset = newMutableAggBufferOffset)
+
+  override def withNewInputAggBufferOffset(newInputAggBufferOffset: Int): CountMinSketchAgg =
+    copy(inputAggBufferOffset = newInputAggBufferOffset)
+
+  override def inputTypes: Seq[AbstractDataType] = {
+    // Currently `CountMinSketch` supports integral (date/timestamp is represented as int/long
+    // internally) and string types.
+    Seq(TypeCollection(IntegralType, StringType, DateType, TimestampType),
+      DoubleType, DoubleType, IntegerType)
```
hvanhovell: Also add `FloatType` (use Float.floatToIntBits), `DoubleType` (use Double.doubleToLongBits), `BooleanType` and `BinaryType`? We could also add support for Decimal, but that would be a bit harder to get right.

wzhfy: @rxin @hvanhovell If we really want to support all these types, is it better to move this conversion and pattern match logics into `CountMinSketch`? That is, make cms support these types itself. Then, when users do queries e.g. on float type, they don't need to do conversions like `cms.estimateCount(Float.floatToIntBits(value))`.

rxin: Yes if we want to add support for those I think it'd make sense to do it in count-min sketch itself too.

wzhfy:
cc @rxin @hvanhovell

common/sketch/src/main/java/org/apache/spark/util/sketch/CountMinSketchImpl.java
```diff
@@ -152,6 +153,16 @@ public void add(Object item) {
   public void add(Object item, long count) {
     if (item instanceof String) {
       addString((String) item, count);
+    } else if (item instanceof BigDecimal) {
+      addString(((BigDecimal) item).toString(), count);
```
rxin: Is this true? 

"1.0" and "1.00" is the same value but not the same string representation.

rxin:
Thanks - I'm going to merge this in master. I will submit a follow-up PR to simplify this a little bit, and remove the handling of float/double/decimal types and require explicit user action on how to turn that into long.

rxin:
Hey guys - after looking at the pr more, I'm afraid we have gone overboard with testing here. Most of the test cases written are just repeating each other and doing exactly the same thing. For testing something like this I'd probably just have some simple end-to-end test and then be done with it, because most of the complicated logics are isolated in the actual CountMinSketch implementation itself and already has good test coverage.

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/aggregate/CountMinSketchAggSuite.scala
```diff
@@ -0,0 +1,320 @@
...
+
+    // Check non-empty serialize and de-serialize
+    val random = new Random(31)
+    (0 until 10000).map(_ => random.nextInt(100)).foreach { value =>
+      buffer.add(value)
+    }
+    assert(buffer.equals(agg.deserialize(agg.serialize(buffer))))
+  }
+
+  def testHighLevelInterface[T: ClassTag](
```
rxin: @wzhfy can you comment on why we need to test both the high level interface and the low level interface?

PR #12136, opened 2016-04-03
[SPARK-14353] Dataset Time Window `window` API for Python, and SQL

brkyvz:
## What changes were proposed in this pull request?

The `window` function was added to Dataset with [this PR](https://github.com/apache/spark/pull/12008).
This PR adds the Python, and SQL, API for this function.

With this PR, SQL, Java, and Scala will share the same APIs as in users can use:
- `window(timeColumn, windowDuration)` 
- `window(timeColumn, windowDuration, slideDuration)` 
- `window(timeColumn, windowDuration, slideDuration, startTime)`

In Python, users can access all APIs above, but in addition they can do
- In Python:
  `window(timeColumn, windowDuration, startTime=...)`

that is, they can provide the startTime without providing the `slideDuration`. In this case, we will generate tumbling windows.
## How was this patch tested?

Unit tests + manual tests

PR #15878, opened 2016-11-14
[SPARK-18430] [SQL] Fixed Exception Messages when Hitting an Invocation Exception of Function Lookup

gatorsmile:
When the exception is an invocation exception during function lookup, we return a useless/confusing error message:

For example, 
```Scala
df.selectExpr("concat_ws()")
```
Below is the error message we got:
```
null; line 1 pos 0
org.apache.spark.sql.AnalysisException: null; line 1 pos 0
```

To get the meaningful error message, we need to get the cause. The fix is exactly the same as what we did in https://github.com/apache/spark/pull/12136. After the fix, the message we got is the exception issued in the constuctor of function implementation:
```
requirement failed: concat_ws requires at least one argument.; line 1 pos 0
org.apache.spark.sql.AnalysisException: requirement failed: concat_ws requires at least one argument.; line 1 pos 0
```

Added test cases.


sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/FunctionRegistry.scala
```diff
@@ -446,7 +446,10 @@ object FunctionRegistry {
         // If there is an apply method that accepts Seq[Expression], use that one.
         Try(varargCtor.get.newInstance(expressions).asInstanceOf[Expression]) match {
           case Success(e) => e
-          case Failure(e) => throw new AnalysisException(e.getMessage)
+          case Failure(e) =>
+            // the exception is an invocation exception. To get a meaningful message, we need the
+            // cause.
+            throw new AnalysisException(e.getCause.getMessage)
```
gatorsmile: Above is [the same fix](https://github.com/gatorsmile/spark/blob/a43acced66fa15228b6c32176bb9f28dacb6f024/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/FunctionRegistry.scala#L465-L468) in https://github.com/apache/spark/pull/12136.

rxin:
Does this apply to all functions? If yes, I'd put a test in SQLQueryTestSuite

gatorsmile:
Sure, let me do it.

rxin:

rxin:
Can you submit a backport for branch-2.0?

gatorsmile:
Sure, will do it. Thanks!

PR #15880, opened 2016-11-14
[SPARK-17913][SQL] compare atomic and string type column may return confusing result

cloud-fan:

Spark SQL follows MySQL to do the implicit type conversion for binary comparison: http://dev.mysql.com/doc/refman/5.7/en/type-conversion.html

However, this may return confusing result, e.g. `1 = 'true'` will return true, `19157170390056973L = '19157170390056971'` will return true.

I think it's more reasonable to follow postgres in this case, i.e. cast string to the type of the other side, but return null if the string is not castable to keep hive compatibility.

## How was this patch tested?

newly added tests.


cloud-fan:
cc @yhuai @gatorsmile

hvanhovell:
This might be a bad idea: should we follow the old casting strategy if we cannot cast from string to atomic datatype?

cloud-fan:
We can only know if this string is castable at runtime. BTW, other databases(like MySQL) have special implicit type conversion rules for constants, should we follow them?

hvanhovell:
Yeah, you are totally right about that. I like this approach, the only bothering me is that this breaks backwards compatibility.

dongjoon-hyun:
Hi, @cloud-fan . What about just showing warning for that?
For Hive 1.2.1 and 2.1.0, it's `true` like the old Spark.
```sql
hive> SELECT 19157170390056973L = "19157170390056971";
WARNING: Comparing a bigint and a string may result in a loss of precision.
OK
true
```

gatorsmile:
Below contains a section of `Implicit Data Conversion` in Oracle: https://docs.oracle.com/cd/B19306_01/server.102/b14200/sql_elements002.htm

It clearly documents the potential changes in implicit conversion and encourage users to do explicit casting.
> Algorithms for implicit conversion are subject to change across software releases and among Oracle products. Behavior of explicit conversions is more predictable.

cloud-fan:
Ok we need to make a decision here, to follow hive and give a warning message, or to follow postgres and cast string to the type of the other side.

Personally I prefer the postgres way, I think it's always better than blindly cast both side to double.

cc @rxin @marmbrust

hvanhovell:
+1 on the postgres approach

gatorsmile:
+1 on the postgres approach

dongjoon-hyun:
+1 on the postgres approach

rxin:
+1 on the postgres approach

sql/core/src/test/scala/org/apache/spark/sql/execution/datasources/json/JsonSuite.scala
```diff
@@ -448,7 +448,7 @@ class JsonSuite extends QueryTest with SharedSQLContext with TestJsonData {
 
     // Number and String conflict: resolve the type as number in this query.
     checkAnswer(
-      sql("select num_str + 1.2 from jsonTable where num_str > 14"),
+      sql("select num_str + 1.2 from jsonTable where num_str > 14d"),
```
cloud-fan: now we have a compatibility issue here: when comparing string of large numbers(larger than long) with int, we cast that string to int, which will return null(because it's not castable) and the filter condition becomes false. Is it a valid use case?

cloud-fan:
to be safe, how about we retargeting this ticket to 2.2?

dasbipulkumar:
@cloud-fan 
Will this handle UDFs?

cloud-fan:
`BinaryComparison` is special and we have special rules for it. For UDF we only apply the normal implicit type conversions. Have you met any problem with UDF?

dasbipulkumar:
Please refer this bug
https://issues.apache.org/jira/browse/SPARK-18489

cloud-fan:
I agree implicit cast is kind of evil, and for the example in SPARK-18489, it's better to follow postgres and fail the query because we can't cast "s" to integer. But this will break hive compatibility heavily, I think we shouldn't change it.

dasbipulkumar:
yes. As a Dev we can handle these things, but as Spark Sql is moving towards more abstracted APIs, it can be painful for general users. I hope that handing these issues can be planned in upcoming major releases.

gatorsmile:
In the future, more code changes might break Hive compatibility accidentally or intentionally. 

Maybe, we can introduce a configuration flag. For example, in DB2, we have a flag like
```
db2set DB2_COMPATIBILITY_VECTOR=ORA
```
Below is the link to the article:
https://www.ibm.com/developerworks/data/library/techarticle/dm-0907oracleappsondb2/

There is a section about implicit casting.

gatorsmile:
The [official Hive document](https://cwiki.apache.org/confluence/display/Hive/LanguageManual+Types#LanguageManualTypes-AllowedImplicitConversions) explains the implicit type casting rules. However, the rule is not exactly identical to ours, as shown above.

SourceType\CastToType | ByteType | ShortType | IntegerType | LongType | DoubleType | FloatType | Dec(10, 2) | BinaryType | BooleanType | StringType | DateType | TimestampType | ArrayType | MapType | StructType | NullType | CalendarIntervalType | DecimalType | NumericType | IntegralType
------------ | ------------ | ------------ | ------------ | ------------ | ------------ | ------------ | ------------ | ------------ | ------------ | ------------ | ------------ | ------------ | ------------ | ------------ | ------------ | ------------ | ------------ | ------------ | ------------ |  -----------
**ByteType** | ByteType | ShortType | IntegerType | LongType | DoubleType | FloatType | Dec(10, 2) | X    | X    | StringType | X    | X    | X    | X    | X    | X    | X    | DecimalType(3, 0) | ByteType | ByteType
**ShortType** | ByteType | ShortType | IntegerType | LongType | DoubleType | FloatType | Dec(10, 2) | X    | X    | StringType | X    | X    | X    | X    | X    | X    | X    | DecimalType(5, 0) | ShortType | ShortType
**IntegerType** | ByteType | ShortType | IntegerType | LongType | DoubleType | FloatType | Dec(10, 2) | X    | X    | StringType | X    | X    | X    | X    | X    | X    | X    | DecimalType(10, 0) | IntegerType | IntegerType
**LongType** | ByteType | ShortType | IntegerType | LongType | DoubleType | FloatType | Dec(10, 2) | X    | X    | StringType | X    | X    | X    | X    | X    | X    | X    | DecimalType(20, 0) | LongType | LongType
**DoubleType** | ByteType | ShortType | IntegerType | LongType | DoubleType | FloatType | Dec(10, 2) | X    | X    | StringType | X    | X    | X    | X    | X    | X    | X    | DecimalType(30, 15) | DoubleType | IntegerType
**FloatType** | ByteType | ShortType | IntegerType | LongType | DoubleType | FloatType | Dec(10, 2) | X    | X    | StringType | X    | X    | X    | X    | X    | X    | X    | DecimalType(14, 7) | FloatType | IntegerType
**Dec(10, 2)** | ByteType | ShortType | IntegerType | LongType | DoubleType | FloatType | Dec(10, 2) | X    | X    | StringType | X    | X    | X    | X    | X    | X    | X    | DecimalType(10, 2) | Dec(10, 2) | IntegerType
**BinaryType** | X    | X    | X    | X    | X    | X    | X    | BinaryType | X    | StringType | X    | X    | X    | X    | X    | X    | X    | X    | X    | X   
**BooleanType** | X    | X    | X    | X    | X    | X    | X    | X    | BooleanType | StringType | X    | X    | X    | X    | X    | X    | X    | X    | X    | X   
**StringType** | ByteType | ShortType | IntegerType | LongType | DoubleType | FloatType | Dec(10, 2) | BinaryType | X    | StringType | DateType | TimestampType | X    | X    | X    | X    | X    | DecimalType(38, 18) | DoubleType | X   
**DateType** | X    | X    | X    | X    | X    | X    | X    | X    | X    | StringType | DateType | TimestampType | X    | X    | X    | X    | X    | X    | X    | X   
**TimestampType** | X    | X    | X    | X    | X    | X    | X    | X    | X    | StringType | DateType | TimestampType | X    | X    | X    | X    | X    | X    | X    | X   
**ArrayType** | X    | X    | X    | X    | X    | X    | X    | X    | X    | X    | X    | X    | ArrayType* | X    | X    | X    | X    | X    | X    | X   
**MapType** | X    | X    | X    | X    | X    | X    | X    | X    | X    | X    | X    | X    | X    | MapType* | X    | X    | X    | X    | X    | X   
**StructType** | X    | X    | X    | X    | X    | X    | X    | X    | X    | X    | X    | X    | X    | X    | StructType* | X    | X    | X    | X    | X   
**NullType** | ByteType | ShortType | IntegerType | LongType | DoubleType | FloatType | Dec(10, 2) | BinaryType | BooleanType | StringType | DateType | TimestampType | ArrayType | MapType | StructType | NullType | CalendarIntervalType | DecimalType(38, 18) | DoubleType | IntegerType
**CalendarIntervalType** | X    | X    | X    | X    | X    | X    | X    | X    | X    | X    | X    | X    | X    | X    | X    | X    | CalendarIntervalType | X    | X    | X   
Note: ArrayType\*, MapType\*, StructType\* are castable only when the internal child types also match; otherwise, not castable

dasbipulkumar:
@gatorsmile 
 configuration flag would be good approach.

cloud-fan:
retest this please

cloud-fan:
retest this please

gatorsmile:
Just for your reference, below is the conversion charts of MS SQL Server. It includes both implicit and explicit conversion rules. 

screenshot 2017-01-02 23 18 56

Source: https://msdn.microsoft.com/en-us/library/ms191530.aspx

cloud-fan:
retest this please

cloud-fan:
retest this please

sql/core/src/test/scala/org/apache/spark/sql/DataFrameSuite.scala
```diff
@@ -1728,4 +1728,9 @@ class DataFrameSuite extends QueryTest with SharedSQLContext {
     val df = spark.createDataFrame(spark.sparkContext.makeRDD(rows), schema)
     assert(df.filter($"array1" === $"array2").count() == 1)
   }
+
+  test("SPARK-17913: compare long and string type column may return confusing result") {
+    val df = Seq(123L -> "123", 19157170390056973L -> "19157170390056971").toDF("i", "j")
```
gatorsmile: Like what the PR description mentioned, could we add an invalid test case? It sounds like we are returning `null` when it is not castable?

cloud-fan: "returning null when not castable" is an existing behavior, this PR just changes the target type of cast when comparing string and atomic types.

gatorsmile:
LGTM

The PR title is not right. BTW, we might need a release note for this PR. This will change the behaviors.

cloud-fan:
retest this please

cloud-fan:
I have updated the PR title and description, and added a release_notes label in the ticket.

gatorsmile:
Thanks! Merging to master.

PR #15881, opened 2016-11-14
[SPARK-18434][ML] Add missing ParamValidations for ML algos

zhengruifeng:
Add missing ParamValidations for ML algos
## How was this patch tested?
existing tests

### DIFF
--- a/mllib/src/main/scala/org/apache/spark/ml/regression/LinearRegression.scala
+++ b/mllib/src/main/scala/org/apache/spark/ml/regression/LinearRegression.scala
@@ -171,7 +171,10 @@ class LinearRegression @Since("1.3.0") (@Since("1.3.0") override val uid: String
    * @group setParam
    */
   @Since("1.6.0")
-  def setSolver(value: String): this.type = set(solver, value)
+  def setSolver(value: String): this.type = {
+    require(Array("auto", "l-bfgs", "normal").contains(value), s"Solver $value was not supported.")
+    set(solver, value)
+  }
   setDefault(solver -> "auto")
 
   /**
--- a/mllib/src/main/scala/org/apache/spark/ml/regression/LinearRegression.scala
+++ b/mllib/src/main/scala/org/apache/spark/ml/regression/LinearRegression.scala
@@ -172,7 +172,8 @@ class LinearRegression @Since("1.3.0") (@Since("1.3.0") override val uid: String
    */
   @Since("1.6.0")
   def setSolver(value: String): this.type = {
-    require(Array("auto", "l-bfgs", "normal").contains(value), s"Solver $value was not supported.")
+    require(Set("auto", "l-bfgs", "normal").contains(value),
+      s"Solver $value was not supported. Supported options: auto, l-bfgs, normal")
     set(solver, value)
   }
   setDefault(solver -> "auto")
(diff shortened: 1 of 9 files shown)


mllib/src/main/scala/org/apache/spark/ml/regression/LinearRegression.scala
```diff
@@ -171,7 +171,10 @@ class LinearRegression @Since("1.3.0") (@Since("1.3.0") override val uid: String
    * @group setParam
    */
   @Since("1.6.0")
-  def setSolver(value: String): this.type = set(solver, value)
+  def setSolver(value: String): this.type = {
+    require(Array("auto", "l-bfgs", "normal").contains(value), s"Solver $value was not supported.")
```
srowen: Maybe just a switch/case is simpler? I don't feel strongly though

srowen (approved):
These all look good to me. @jkbradley @sethah et al?

mllib/src/main/scala/org/apache/spark/ml/regression/LinearRegression.scala
sethah: This is a reasonable argument, but everywhere else we use a param validator and list the valid options in the param doc, AFAIK. If we want to change this, it seems like a discussion that should be had somewhere else, and we should stick to the current convention here. I don't see a good argument for changing the convention in one place, but not all places.

hhbyyh: Thanks for the reply. I assume `solver` is different from other params as it's inherited from a trait where we cannot define valid options. I don't mean to change any convention here. For this case, I simply mean we can add the available options in the error message in the `require`.

And surely we can discuss how to improve ParamValidator elsewhere.

sethah: Yes, good point about the param definition being inherited, I had overlooked that. I think it's ok to use a require here then. I would prefer `Set("auto", ...).contains` but it's a very minor issue.

zhengruifeng:
Thanks for reviewing. I update the checking in LiR's `solver` :  Array -> Set, output supported options in err message.

mllib/src/main/scala/org/apache/spark/ml/tree/treeParams.scala
```diff
@@ -73,11 +73,13 @@ private[ml] trait DecisionTreeParams extends PredictorParams
 
   /**
    * Minimum information gain for a split to be considered at a tree node.
+   * Should be >= 0.0.
    * (default = 0.0)
    * @group param
    */
   final val minInfoGain: DoubleParam = new DoubleParam(this, "minInfoGain",
-    "Minimum information gain for a split to be considered at a tree node.")
+    "Minimum information gain for a split to be considered at a tree node.",
```
hhbyyh: Sorry to miss this. Perhaps add >=0 at the end to be consistent with others.

hhbyyh (approved):
LGTM

mllib/src/main/scala/org/apache/spark/ml/tree/treeParams.scala
zhengruifeng: 👌， I will add it

holdenk:
Thanks for taking this on :) I did a quick skim and these all look pretty reasonable (in some cases we are directly calling an old mllib algorithm which would do the validation at training time and the values seem to line up).

yanboliang:
LGTM, thanks to all. Merged into master and branch-2.1.

PR #15882, opened 2016-11-14
[SPARK-18400][STREAMING] NPE when resharding Kinesis Stream

srowen:

Avoid NPE in KinesisRecordProcessor when shutdown happens without successful init

## How was this patch tested?

Existing tests

### DIFF
--- a/external/kinesis-asl/src/main/scala/org/apache/spark/streaming/kinesis/KinesisRecordProcessor.scala
+++ b/external/kinesis-asl/src/main/scala/org/apache/spark/streaming/kinesis/KinesisRecordProcessor.scala
@@ -27,7 +27,6 @@ import com.amazonaws.services.kinesis.clientlibrary.types.ShutdownReason
 import com.amazonaws.services.kinesis.model.Record
 
 import org.apache.spark.internal.Logging
-import org.apache.spark.streaming.Duration
 
 /**
  * Kinesis-specific implementation of the Kinesis Client Library (KCL) IRecordProcessor.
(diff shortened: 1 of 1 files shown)


external/kinesis-asl/src/main/scala/org/apache/spark/streaming/kinesis/KinesisRecordProcessor.scala
```diff
@@ -102,27 +101,29 @@ private[kinesis] class KinesisRecordProcessor[T](receiver: KinesisReceiver[T], w
...
     logInfo(s"Shutdown:  Shutting down workerId $workerId with reason $reason")
-    reason match {
-      /*
-       * TERMINATE Use Case.  Checkpoint.
-       * Checkpoint to indicate that all records from the shard have been drained and processed.
-       * It's now OK to read from the new shards that resulted from a resharding event.
-       */
-      case ShutdownReason.TERMINATE =>
-        receiver.removeCheckpointer(shardId, checkpointer)
+    if (shardId != null) { // null if not initialized before shutdown
```
srowen: This is the only substantive change; the rest is style/format fix

boneill42: cool.  only difference i had in my fix was a LOG message indicating that we hit this situation.  (not sure the log message is actionable, but it is informative)

srowen:
Merged to master/2.1/2.0

PR #15883, opened 2016-11-14
[SPARK-18438][SPARKR][ML] spark.mlp should support RFormula.

yanboliang:
```spark.mlp``` should support ```RFormula``` like other ML algorithm wrappers.
BTW, I did some cleanup and improvement for ```spark.mlp```.

## How was this patch tested?
Unit tests.

### DIFF
--- a/R/pkg/R/mllib.R
+++ b/R/pkg/R/mllib.R
@@ -231,7 +231,7 @@ setMethod("spark.glm", signature(data = "SparkDataFrame", formula = "formula"),
               stop("'family' not recognized")
             }
 
-            formula <- paste0(deparse(formula), collapse = "")
+            formula <- paste(deparse(formula), collapse = "")
             if (is.null(weightCol)) {
               weightCol <- ""
             }
@@ -525,7 +525,7 @@ setMethod("write.ml", signature(object = "LDAModel", path = "character"),
 #' @note spark.isoreg since 2.1.0
 setMethod("spark.isoreg", signature(data = "SparkDataFrame", formula = "formula"),
           function(data, formula, isotonic = TRUE, featureIndex = 0, weightCol = NULL) {
-            formula <- paste0(deparse(formula), collapse = "")
+            formula <- paste(deparse(formula), collapse = "")
 
             if (is.null(weightCol)) {
               weightCol <- ""
@@ -609,7 +609,7 @@ setMethod("summary", signature(object = "IsotonicRegressionModel"),
 #' @seealso \link{predict}, \link{read.ml}, \link{write.ml}
 setMethod("spark.kmeans", signature(data = "SparkDataFrame", formula = "formula"),
           function(data, formula, k = 2, maxIter = 20, initMode = c("k-means||", "random")) {
-            formula <- paste0(deparse(formula), collapse = "")
+            formula <- paste(deparse(formula), collapse = "")
             initMode <- match.arg(initMode)
             jobj <- callJStatic("org.apache.spark.ml.r.KMeansWrapper", "fit", data@sdf, formula,
                                 as.integer(k), as.integer(maxIter), initMode)
@@ -775,7 +775,7 @@ setMethod("spark.logit", signature(data = "SparkDataFrame", formula = "formula")
                    tol = 1E-6, fitIntercept = TRUE, family = "auto", standardization = TRUE,
                    thresholds = 0.5, weightCol = NULL, aggregationDepth = 2,
                    probabilityCol = "probability") {
-            formula <- paste0(deparse(formula), collapse = "")
+            formula <- paste(deparse(formula), collapse = "")
 
             if (is.null(weightCol)) {
               weightCol <- ""
@@ -872,7 +872,7 @@ setMethod("summary", signature(object = "LogisticRegressionModel"),
 #' @param ... additional arguments passed to the method.
 #' @return \code{spark.mlp} returns a fitted Multilayer Perceptron Classification Model.
 #' @rdname spark.mlp
-#' @aliases spark.mlp,SparkDataFrame-method,formula-method
+#' @aliases spark.mlp,SparkDataFrame,formula-method
 #' @name spark.mlp
 #' @seealso \link{read.ml}
 #' @export
@@ -901,7 +901,7 @@ setMethod("summary", signature(object = "LogisticRegressionModel"),
 setMethod("spark.mlp", signature(data = "SparkDataFrame", formula = "formula"),
           function(data, formula, layers, blockSize = 128, solver = "l-bfgs", maxIter = 100,
                    tol = 1E-6, stepSize = 0.03, seed = NULL, initialWeights = NULL) {
-            formula <- paste0(deparse(formula), collapse = "")
+            formula <- paste(deparse(formula), collapse = "")
             if (is.null(layers)) {
               stop ("layers must be a integer vector with length > 1.")
             }
@@ -999,7 +999,7 @@ setMethod("summary", signature(object = "MultilayerPerceptronClassificationModel
 #' @note spark.naiveBayes since 2.0.0
 setMethod("spark.naiveBayes", signature(data = "SparkDataFrame", formula = "formula"),
           function(data, formula, smoothing = 1.0) {
-            formula <- paste0(deparse(formula), collapse = "")
+            formula <- paste(deparse(formula), collapse = "")
             jobj <- callJStatic("org.apache.spark.ml.r.NaiveBayesWrapper", "fit",
             formula, data@sdf, smoothing)
             new("NaiveBayesModel", jobj = jobj)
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: R/pkg/R/mllib.R [hunk 1]
```diff
@@ -936,20 +939,23 @@ setMethod("predict", signature(object = "MultilayerPerceptronClassificationModel
...
 setMethod("summary", signature(object = "MultilayerPerceptronClassificationModel"),
           function(object) {
             jobj <- object@jobj
-            labelCount <- callJMethod(jobj, "labelCount")
             layers <- unlist(callJMethod(jobj, "layers"))
+            numOfInputs <- head(layers, n = 1)
+            numOfOutputs <- tail(layers, n = 1)
             weights <- callJMethod(jobj, "weights")
-            list(labelCount = labelCount, layers = layers, weights = weights)
+            list(numOfInputs = numOfInputs, numOfOutputs = numOfOutputs,
```
yanboliang: I changed the summary output to ```numOfInputs``` and ```numOfOutputs```, which is consistent with ```mlp``` in R ```RSNNS``` package. The original ```labelCount``` is actually the ```numOfOutputs```, but the later one should be more descriptive as well.

R/pkg/inst/tests/testthat/test_mllib.R
```diff
@@ -385,7 +386,7 @@ test_that("spark.mlp", {
   # Test predict method
   mlpTestDF <- df
   mlpPredictions <- collect(select(predict(model, mlpTestDF), "prediction"))
-  expect_equal(head(mlpPredictions$prediction, 6), c(0, 1, 1, 1, 1, 1))
+  expect_equal(head(mlpPredictions$prediction, 6), c("1.0", "0.0", "0.0", "0.0", "0.0", "0.0"))
```
yanboliang: Currently, SparkR ML classification algorithm wrappers supposes all labels are character, so the prediction is character as well. Users can cast it to numeric if it's applicable.

mllib/src/main/scala/org/apache/spark/ml/r/MultilayerPerceptronClassifierWrapper.scala
```diff
@@ -73,25 +90,25 @@ private[r] object MultilayerPerceptronClassifierWrapper
...
-      .setStages(Array(mlp))
+      .setStages(Array(rFormulaModel, mlp, idxToStr))
       .fit(data)
 
-    val multilayerPerceptronClassificationModel: MultilayerPerceptronClassificationModel =
-    pipeline.stages.head.asInstanceOf[MultilayerPerceptronClassificationModel]
-
-    val weights = multilayerPerceptronClassificationModel.weights.toArray
-    val layersFromPipeline = multilayerPerceptronClassificationModel.layers
-    val labelCount = data.select("label").distinct().count()
```
yanboliang: ```weights``` and ```layers``` can be get from the MLP model, so it's not necessary to store them. It's also not necessary to calculate ```labelCount``` by ```distinct``` which is very expensive. Actually we can get ```labelCount``` from the member variable ```layers``` of MLP model.

yanboliang:
cc @felixcheung

R/pkg/R/mllib.R
```diff
@@ -896,9 +898,10 @@ setMethod("summary", signature(object = "LogisticRegressionModel"),
 #' summary(savedModel)
 #' }
 #' @note spark.mlp since 2.1.0
-setMethod("spark.mlp", signature(data = "SparkDataFrame"),
-          function(data, layers, blockSize = 128, solver = "l-bfgs", maxIter = 100,
+setMethod("spark.mlp", signature(data = "SparkDataFrame", formula = "formula"),
+          function(data, formula, layers, blockSize = 128, solver = "l-bfgs", maxIter = 100,
```
felixcheung: if without `formula` works before, is/should `formula` be optional then? with this change it will require it

R/pkg/R/mllib.R
```diff
@@ -896,9 +898,10 @@ setMethod("summary", signature(object = "LogisticRegressionModel"),
 #' summary(savedModel)
 #' }
 #' @note spark.mlp since 2.1.0
```
felixcheung: we are targeting 2.1.0 for this change yes? otherwise it is a breaking signature change

R/pkg/inst/tests/testthat/test_mllib.R
felixcheung: isn't this an improvement we should make?

yanboliang:
Jenkins, test this please.

R/pkg/R/mllib.R
yanboliang: No, we must have ```formula```. The original PR is very tricky and even can be regarded as bug. We suppose the feature and label column as "features/label", but it should allow users to specify their own feature and label column by R formula. You can refer that almost all ML wrappers support R formula.

R/pkg/R/mllib.R
yanboliang: Yeah, I think so.

R/pkg/R/mllib.R
```diff
@@ -936,20 +939,23 @@ setMethod("predict", signature(object = "MultilayerPerceptronClassificationModel
 # Returns the summary of a Multilayer Perceptron Classification Model produced by \code{spark.mlp}
 
 #' @param object a Multilayer Perceptron Classification Model fitted by \code{spark.mlp}
-#' @return \code{summary} returns a list containing \code{labelCount}, \code{layers}, and
-#'         \code{weights}. For \code{weights}, it is a numeric vector with length equal to
-#'         the expected given the architecture (i.e., for 8-10-2 network, 100 connection weights).
+#' @return \code{summary} returns a list containing \code{numOfInputs}, \code{numOfOutputs},
+#'         \code{layers}, and \code{weights}. For \code{weights}, it is a numeric vector with
+#'         length equal to the expected given the architecture (i.e., for 8-10-2 network,
+#'         112 connection weights).
```
yanboliang: Here should be 112 rather than 110, it should be calculated by 8 * 10 + 10 * 2 + 10 + 2. You can refer the test case to verify it.

yanboliang:
@felixcheung I'm not quite understand the test failure, could you give me some suggestion? Thanks.
```
* checking for missing documentation entries ... WARNING
Undocumented S4 methods:
  generic 'spark.mlp' and siglist 'SparkDataFrame,formula'
All user-level objects in a package (including S4 classes and methods)
should have documentation entries.
See the chapter 'Writing R documentation files' in the 'Writing R
Extensions' manual.
* checking for code/documentation mismatches ... OK
* checking Rd \usage sections ... WARNING
Objects in \usage without \alias in documentation object 'spark.mlp':
  '\S4method{spark.mlp}{SparkDataFrame,formula}'

Functions with \usage entries need to have the appropriate \alias
entries, and all their arguments documented.
The \usage entries must correspond to syntactically valid R code.
See the chapter 'Writing R documentation files' in the 'Writing R
Extensions' manual.
* checking Rd contents ... OK
* checking for unstated dependencies in examples ... OK
* checking installed files from 'inst/doc' ... OK
* checking files in 'vignettes' ... OK
* checking examples ... OK
* checking for unstated dependencies in tests ... OK
* checking tests ... SKIPPED
* checking for unstated dependencies in vignettes ... OK
* checking package vignettes in 'inst/doc' ... OK
* checking running R code from vignettes ... SKIPPED
* checking re-building of vignette outputs ... SKIPPED

WARNING: There were 2 warnings.
NOTE: There were 3 notes.
See
  '/home/jenkins/workspace/SparkPullRequestBuilder@2/R/SparkR.Rcheck/00check.log'
for details.

Had CRAN check errors; see logs.
[error] running /home/jenkins/workspace/SparkPullRequestBuilder@2/R/run-tests.sh ; received return code 255
```

R/pkg/R/mllib.R
felixcheung: right, I can see that. basically it makes assumptions on the names of feature & label columns

R/pkg/R/mllib.R
```diff
@@ -870,7 +872,7 @@ setMethod("summary", signature(object = "LogisticRegressionModel"),
 #' @param ... additional arguments passed to the method.
 #' @return \code{spark.mlp} returns a fitted Multilayer Perceptron Classification Model.
 #' @rdname spark.mlp
-#' @aliases spark.mlp,SparkDataFrame-method
+#' @aliases spark.mlp,SparkDataFrame-method,formula-method
```
felixcheung: this should be `#' @aliases spark.mlp,SparkDataFrame,formula-method` with one `-method` at the end

felixcheung:
try my comment suggestion - it should fix the error

re: `paste0` it looks like there are more `paste` than `paste0`? perhaps we shouldn't change it then? what do you think?

R/pkg/R/mllib.R
yanboliang: Oops.

yanboliang:
@felixcheung Both ```paste``` and ```paste0``` should work well in this scenario, since ```RFormula``` can handle space between each term. I think it would be better to unify them, so I use ```paste``` across ```mllib.R```. Any more comments? Thanks.

felixcheung:
Yes thanks
LGTM

yanboliang:
Merged into master and branch-2.1. Thanks for reviewing.

PR #15884, opened 2016-11-15
[SPARK-18433][SQL] Improve DataSource option keys to be more case-insensitive

dongjoon-hyun:

This PR aims to improve DataSource option keys to be more case-insensitive

DataSource partially use CaseInsensitiveMap in code-path. For example, the following fails to find url.

```scala
df.write.format("jdbc")
    .option("UrL", url1)
    .options(properties.asScala)
    .save()
```

This PR makes DataSource options to use CaseInsensitiveMap internally and also makes DataSource to use CaseInsensitiveMap generally except `InMemoryFileIndex` and `InsertIntoHadoopFsRelationCommand`. We can not pass them CaseInsensitiveMap because they creates new case-sensitive HadoopConfs by calling newHadoopConfWithOptions(options) inside.

## How was this patch tested?

Pass the Jenkins test with newly added test cases.

### DIFF
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/json/JSONOptions.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/json/JSONOptions.scala
@@ -23,7 +23,7 @@ import com.fasterxml.jackson.core.{JsonFactory, JsonParser}
 import org.apache.commons.lang3.time.FastDateFormat
 
 import org.apache.spark.internal.Logging
-import org.apache.spark.sql.catalyst.util.{CompressionCodecs, ParseModes}
+import org.apache.spark.sql.catalyst.util.{CaseInsensitiveMap, CompressionCodecs, ParseModes}
 
 /**
  * Options for parsing JSON data into Spark SQL rows.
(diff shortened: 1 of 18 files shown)


sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/json/JSONOptions.scala
```diff
@@ -34,35 +34,42 @@ private[sql] class JSONOptions(
     @transient private val parameters: Map[String, String])
```
cloud-fan: can we just force users to pass a `CaseInsensitiveMap`?

dongjoon-hyun: Oh, sure! You mean `require(parameters.isInstanceOf[CaseInsensitiveMap])`, right?

dongjoon-hyun: I will update this PR in that way.

cloud-fan: `JSONOptions` is spark private, we can just declare the type is `CaseInsensitiveMap`, i.e. `parameters: CaseInsensitiveMap`

dongjoon-hyun:
Thank you, @cloud-fan .
It becomes much better. Could you review again please?

dongjoon-hyun:
Rebased to resolve conflicts.

sql/core/src/test/scala/org/apache/spark/sql/execution/datasources/json/JsonSuite.scala
```diff
@@ -1366,7 +1366,7 @@ class JsonSuite extends QueryTest with SharedSQLContext with TestJsonData {
 
   test("SPARK-6245 JsonRDD.inferSchema on empty RDD") {
     // This is really a test that it doesn't throw an exception
-    val emptySchema = InferSchema.infer(empty, "", new JSONOptions(Map()))
+    val emptySchema = InferSchema.infer(empty, "", new JSONOptions(Map.empty[String, String]))
```
cloud-fan: unnecessary change?

cloud-fan:
LGTM

sql/core/src/test/scala/org/apache/spark/sql/execution/datasources/json/JsonSuite.scala
dongjoon-hyun: In this case, we expect `Map[String,String]`, but `Map()` or `Map.empty` is `Map[Nothing,Nothing]`.
So, there was some compilation issue.

cloud-fan:

dongjoon-hyun:
Thank you so much, @cloud-fan ! Also, thank you for this issue, @gatorsmile .

PR #15888, opened 2016-11-15
[SPARK-18444][SPARKR] SparkR running in yarn-cluster mode should not download Spark package.

yanboliang:
When running SparkR job in yarn-cluster mode, it will download Spark package from apache website which is not necessary.
```
./bin/spark-submit --master yarn-cluster ./examples/src/main/r/dataframe.R
```
The following is output:
```
Attaching package: ‘SparkR’

The following objects are masked from ‘package:stats’:

The following objects are masked from ‘package:base’:

    as.data.frame, colnames, colnames<-, drop, endsWith, intersect,
    rank, rbind, sample, startsWith, subset, summary, transform, union

Spark not found in SPARK_HOME:
Spark not found in the cache directory. Installation will start.
MirrorUrl not provided.
Looking for preferred site from apache website...
......
```
There's no ```SPARK_HOME``` in yarn-cluster mode since the R process is in a remote host of the yarn cluster rather than in the client host. The JVM comes up first and the R process then connects to it. So in such cases we should never have to download Spark as Spark is already running.

## How was this patch tested?
Offline test.


R/pkg/R/sparkR.R
```diff
@@ -558,16 +558,18 @@ sparkCheckInstall <- function(sparkHome, master) {
       message(msg)
       NULL
     } else {
-      if (!nzchar(master) || isMasterLocal(master)) {
```
yanboliang: ```master``` will be always empty if the job was submitted by ```spark-submit```.

R/pkg/R/sparkR.R
```diff
@@ -558,16 +558,18 @@ sparkCheckInstall <- function(sparkHome, master) {
...
                       sparkHome)
         message(msg)
         packageLocalDir <- install.spark()
         packageLocalDir
-      } else {
+      } else if (master == "yarn-client") {
         msg <- paste0("Spark not found in SPARK_HOME: ",
                       sparkHome, "\n", installInstruction("remote"))
         stop(msg)
+      } else {
```
yanboliang: ```yarn-cluster``` mode, will do nothing.

yanboliang:
cc @felixcheung @shivaram @zjffdu @sun-rui I'm not sure this fix affects other scenarios, but it can fix my bug.

shivaram:
cc @junyangq

R/pkg/R/sparkR.R
```diff
@@ -558,16 +558,18 @@ sparkCheckInstall <- function(sparkHome, master) {
...
     } else {
-      if (!nzchar(master) || isMasterLocal(master)) {
+      if (isMasterLocal(master)) {
         msg <- paste0("Spark not found in SPARK_HOME: ",
                       sparkHome)
         message(msg)
         packageLocalDir <- install.spark()
         packageLocalDir
-      } else {
+      } else if (master == "yarn-client") {
```
felixcheung: should this be changed to hard coded to "yarn-client"?
if I recall, master can be "yarn" and deploy mode can "client" or "cluster" whereas it is possible to have "yarn-client"/"yarn-cluster" - both ways should work.

R/pkg/R/sparkR.R
felixcheung: also what happens to "mesos"? this is what we are saying we should support in SparkR 2.1.0

R/pkg/R/sparkR.R
```diff
@@ -550,24 +566,28 @@ processSparkPackages <- function(packages) {
...
         message(msg)
         packageLocalDir <- install.spark()
         packageLocalDir
-      } else {
+      } else if (nzchar(deployMode) && deployMode == "client") {
         msg <- paste0("Spark not found in SPARK_HOME: ",
                       sparkHome, "\n", installInstruction("remote"))
         stop(msg)
+      } else {
+        NULL
```
yanboliang: @felixcheung 
If we submit SparkR jobs by ```spark-submit```, then ```master``` and ```deployMode``` should be empty. ```sparkHome``` should be set correctly in ```client``` and ```local``` mode, but empty in ```cluster``` mode. ```sparkCheckInstall``` return NULL in all these scenarios.
If we start SparkR by ```SparkR.session``` and pass in corresponding arguments, we should download Spark package when ```master = local[*]```. If ```deployMode``` was set with ```client``` and w/o ```SPARK_HOME``` set correctly, we should stop due to misconfigure. Otherwise(deploy with ```yarn``` or ```mesos``` cluster mode) return NULL here.
From Spark 2.0, ```master = yarn-cluster``` was deprecated by ```master = yarn, deployMode = cluster```. Since ```SparkR.session``` was introduced from 2.0, I don't think we should support the old convention here.

felixcheung:
We do try to maintain compatibility with sparkR.init()

And we should avoid adding parameter in the middle of the parameter list - it breaks users depending on parameter order.

But like my other comment, setting deployMode has never been supported as a parameter and I think we should avoid adding that at this point. Users can always do so via sparkConfig parameter.

yanboliang:
@felixcheung That's OK, and will make the fix more simple. I think ```sparkCheckInstall``` can only handle some conditions by arguments passed in. However, if we can not judge the deploy mode from the given arguments, it should return NULL and let the following code to handle.

felixcheung:
That might be a reasonable approach.
In fact, thinking more about this I think it really should instantiate a sparkConf to read any conf file etc as well (which could set a value for master etc), however, in client mode and if SPARK_HOME is not set then we won't know how to kick off the JVM process..

yanboliang:
@felixcheung What about transferring ```SparkConf``` from JVM to R via environment when JVM creates R process? I guess it can work but need some time to verify and test. Since it's critical bug that users can not submit SparkR jobs via ```spark-submit```, I'd like to fix it firstly and do some refactor after 2.1. What's your opinion? Thanks.

felixcheung:
I think what you have here is reasonable but what about checking for spark.submit.deployMode from the sparkConfig list here too?

I'm all for fixing this ASAP - thinking longer term, I think perhaps we shouldn't auto install spark unless running as an interactive client (i.e. Not client in YARN cluster or client mode, or other cluster manager modes) as in those other cases we should just let sys admin to management it instead of messing with Spark there.

There are ways to check for interactive session in R or we could add an explicit parameter for auto install, or both.

felixcheung:
looks good. We don't have this but how about adding tests for `sparkCheckInstall`? we could at least check for the error cases it has right now?

R/pkg/R/sparkR.R
```diff
@@ -373,8 +373,13 @@ sparkR.session <- function(
     overrideEnvs(sparkConfigMap, paramMap)
   }
 
+  deployMode <- ""
+  if (exists("spark.submit.deployMode", envir = sparkConfigMap)) {
+    deployMode <- sparkConfigMap[["spark.submit.deployMode"]]
+  }
```
felixcheung: you could instead just do `deployMode <- sparkConfigMap[["spark.submit.deployMode"]]`
it will be set to NULL if the key is not set.

yanboliang: Yeah, but I'd like to keep consistent convention with ```sparkHome``` and ```master``` which use "" rather than NULL as default value. Then we don't need to deal with NULL in ```sparkCheckInstall```.

R/pkg/R/sparkR.R
```diff
@@ -550,24 +555,27 @@ processSparkPackages <- function(packages) {
...
-      if (!nzchar(master) || isMasterLocal(master)) {
-        msg <- paste0("Spark not found in SPARK_HOME: ",
-                      sparkHome)
+      if (isMasterLocal(master)) {
+        msg <- paste0("Spark not found in SPARK_HOME: ", sparkHome)
         message(msg)
         packageLocalDir <- install.spark()
         packageLocalDir
-      } else {
+      } else if (isClientMode(master) || deployMode == "client") {
```
yanboliang: ```endsWith``` only available for R version > 3.3, so I implement utility function to check whether it is client mode.

yanboliang (commented):
@felixcheung I added some basic tests for `sparkCheckInstall`. Any more comments, feel free to let me know. Thanks.

R/pkg/R/sparkR.R
felixcheung: fair enough

R/pkg/inst/tests/testthat/test_sparkR.R
```diff
@@ -0,0 +1,36 @@
...
+  deployMode <- ""
+  expect_true(is.null(sparkCheckInstall(sparkHome, master, deployMode)))
+  unlink(sparkHome, recursive = TRUE)
+
+  # "yarn-cluster, mesos-cluster" mode, SPARK_HOME was not set,
+  # and the SparkR job was submitted by "spark-submit"
+  sparkHome <- ""
+  master <- ""
+  deployMode <- ""
+  expect_true(is.null(sparkCheckInstall(sparkHome, master, deployMode)))
```
felixcheung: you could add tests for deployMode == "client" || master == "*-client" as well with expect_error()

yanboliang: Added.

felixcheung:
LGTM

yanboliang:
Merged into master, branch-2.1 and branch-2.0. Thanks for reviewing.

PR #15892, opened 2016-11-15
[SPARK-18300][SQL] Do not apply foldable propagation with expand as a child [BRANCH-2.0]

hvanhovell:
The `FoldablePropagation` optimizer rule, pulls foldable values out from under an `Expand`. This breaks the `Expand` in two ways:

- It rewrites the output attributes of the `Expand`. We explicitly define output attributes for `Expand`, these are (unfortunately) considered as part of the expressions of the `Expand` and can be rewritten.
- Expand can actually change the column (it will typically re-use the attributes or the underlying plan). This means that we cannot safely propagate the expressions from under an `Expand`.

This PR fixes this and (hopefully) other issues by explicitly whitelisting allowed operators.

This is a backport of https://github.com/apache/spark/pull/15857

## How was this patch tested?
Added tests to `FoldablePropagationSuite` and to `SQLQueryTestSuite`.


hvanhovell:
cc @gatorsmile

gatorsmile:
LGTM

rxin:
Merging in.

Can you close the ticket?

PR #15894, opened 2016-11-16
[SPARK-18188]  Add checksum for shuffle blocks

davies:

TBD

## How was this patch tested?

Existing tests.

### DIFF
--- a/common/network-common/src/main/java/org/apache/spark/network/buffer/FileSegmentManagedBuffer.java
+++ b/common/network-common/src/main/java/org/apache/spark/network/buffer/FileSegmentManagedBuffer.java
@@ -17,13 +17,12 @@
 
 package org.apache.spark.network.buffer;
 
-import java.io.File;
-import java.io.FileInputStream;
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.RandomAccessFile;
+import java.io.*;
 import java.nio.ByteBuffer;
 import java.nio.channels.FileChannel;
+import java.util.zip.Adler32;
+import java.util.zip.CheckedInputStream;
+import java.util.zip.Checksum;
 
 import com.google.common.base.Objects;
 import com.google.common.io.ByteStreams;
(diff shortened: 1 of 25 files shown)


common/network-common/src/main/java/org/apache/spark/network/buffer/FileSegmentManagedBuffer.java
```diff
@@ -92,12 +91,27 @@ public ByteBuffer nioByteBuffer() throws IOException {
   }
 
   @Override
-  public InputStream createInputStream() throws IOException {
+  public InputStream createInputStream(boolean checksum) throws IOException {
```
srowen: So, is this only for testing? because it's otherwise very expensive to compute, requiring two passes over the file.

If so, should it not be in some more package-private method only and not exposed?

davies: This is a good question.

Actually we already have checksum along with compression, we could move the decompression a little bit earlier to detect the corruption in block fetcher, will try that soon.

common/network-common/src/main/java/org/apache/spark/network/buffer/FileSegmentManagedBuffer.java
```diff
@@ -17,13 +17,12 @@
...
 
-import java.io.File;
-import java.io.FileInputStream;
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.RandomAccessFile;
+import java.io.*;
 import java.nio.ByteBuffer;
 import java.nio.channels.FileChannel;
+import java.util.zip.Adler32;
```
rxin: would be better to abstract out the checksum functions into some class so it's easier to change in the future.

core/src/main/scala/org/apache/spark/storage/ShuffleBlockFetcherIterator.scala
```diff
@@ -171,14 +179,30 @@ final class ShuffleBlockFetcherIterator(
         override def onBlockFetchSuccess(blockId: String, buf: ManagedBuffer): Unit = {
           // Only add the buffer to results queue if the iterator is not zombie,
           // i.e. cleanup() has not been called yet.
+          val in = try {
```
rxin: note: move this out of callback to not block network pool

davies:
Due to complexity and overhead here, close it in favor of https://github.com/apache/spark/pull/15923/.

PR #15896, opened 2016-11-16
[SPARK-18465] Add 'IF EXISTS' clause to 'UNCACHE' to not throw exceptions when table doesn't exist

brkyvz:

While this behavior is debatable, consider the following use case:
```sql
UNCACHE TABLE foo;
CACHE TABLE foo AS 
SELECT * FROM bar
```
The command above fails the first time you run it. But I want to run the command above over and over again, and I don't want to change my code just for the first run of it.
The issue is that subsequent `CACHE TABLE` commands do not overwrite the existing table.

Now we can do:
```sql
UNCACHE TABLE IF EXISTS foo;
CACHE TABLE foo AS 
SELECT * FROM bar
```

## How was this patch tested?

Unit tests

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/command/cache.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/command/cache.scala
@@ -19,6 +19,7 @@ package org.apache.spark.sql.execution.command
 
 import org.apache.spark.sql.{Dataset, Row, SparkSession}
 import org.apache.spark.sql.catalyst.TableIdentifier
+import org.apache.spark.sql.catalyst.analysis.NoSuchTableException
 import org.apache.spark.sql.catalyst.plans.QueryPlan
 import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan
 
@@ -49,10 +50,18 @@ case class CacheTableCommand(
 }
 
 
-case class UncacheTableCommand(tableIdent: TableIdentifier) extends RunnableCommand {
+case class UncacheTableCommand(
+    tableIdent: TableIdentifier,
+    ifExists: Boolean) extends RunnableCommand {
 
   override def run(sparkSession: SparkSession): Seq[Row] = {
-    sparkSession.catalog.uncacheTable(tableIdent.quotedString)
+    val tableId = tableIdent.quotedString
+    try {
+      sparkSession.catalog.uncacheTable(tableId)
+    } catch {
+      case _: NoSuchTableException if ifExists => // don't throw
+        logInfo(s"Asked to uncache table $tableId which doesn't exist.")
+    }
     Seq.empty[Row]
   }
 }
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/command/cache.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/command/cache.scala
@@ -60,7 +60,6 @@ case class UncacheTableCommand(
       sparkSession.catalog.uncacheTable(tableId)
     } catch {
       case _: NoSuchTableException if ifExists => // don't throw
-        logInfo(s"Asked to uncache table $tableId which doesn't exist.")
     }
     Seq.empty[Row]
   }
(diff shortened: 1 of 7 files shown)


brkyvz:
cc @gatorsmile I changed a test you added. Do you have any strong feelings on this?

gatorsmile:
After fixed the styles, LGTM pending testing.

brkyvz:
@gatorsmile Talking offline with several people, I may put this PR on hold for now since it is  a behavior change. I guess it would be better to go with Options 1 or 2 that I defined in the PR description.

gatorsmile:
@brkyvz Why not supporting both? : )

dongjoon-hyun:
Hi, @brkyvz and @gatorsmile .
Does this proceed to Option 1 or 2 for now? Or is this holding on for next month?

brkyvz:
On hold on my side. Will try to get back to it

dongjoon-hyun:
I see. Thank you for informing that, @brkyvz

andrewor14:
I personally think `UNCACHE TABLE IF EXISTS` is best. It preserves the old behavior but lets the user make sure a table is not cached if they really want.

brkyvz:
Hey @andrewor14 ! I went with your way. @hvanhovell can you take a quick look please? I would really like this to be available in Spark 2.1 (even though it is a new API)

sql/core/src/main/scala/org/apache/spark/sql/execution/command/cache.scala
```diff
@@ -49,10 +50,18 @@ case class CacheTableCommand(
...
+    ifExists: Boolean) extends RunnableCommand {
 
   override def run(sparkSession: SparkSession): Seq[Row] = {
-    sparkSession.catalog.uncacheTable(tableIdent.quotedString)
+    val tableId = tableIdent.quotedString
+    try {
+      sparkSession.catalog.uncacheTable(tableId)
+    } catch {
+      case _: NoSuchTableException if ifExists => // don't throw
+        logInfo(s"Asked to uncache table $tableId which doesn't exist.")
```
hvanhovell: Minor: Lets not log it, logs typically get swallowed by the environment anyway. See for example: https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala#L206

hvanhovell (changes requested):
One small comment, otherwise LGTM

brkyvz:
@hvanhovell Done! Thanks for the quick review!

hvanhovell:
LGTM pending jenkins

gatorsmile:
When the target is a view, it becomes a little bit more complex. If this PR does not handle it, we can do it in a separate PR.
```SQL
UNCACHE TABLE IF EXISTS viewName;
```

hvanhovell:
@gatorsmile could you elaborate a little? What am I missing?

hvanhovell:

gatorsmile:
```Scala
      spark.range(1, 10).toDF("id1").write.format("json").saveAsTable("jt1")
      spark.range(1, 10).toDF("id2").write.format("json").saveAsTable("jt2")
      sql("CREATE VIEW testView AS SELECT * FROM jt1 JOIN jt2 ON id1 == id2")
      // Cache is empty at the beginning
      assert(spark.sharedState.cacheManager.isEmpty)
      sql("CACHE TABLE testView")
      assert(spark.catalog.isCached("testView"))
      // Cache is not empty
      assert(!spark.sharedState.cacheManager.isEmpty)
      // drop a table referenced by a cached view
      sql("DROP TABLE jt1")

-- So far everything is fine

      // Failed to unache the view
      val e = intercept[AnalysisException] {
        sql("UNCACHE TABLE testView")
      }.getMessage
      assert(e.contains("Table or view not found: `default`.`jt1`"))

      // We are unable to drop it from the cache
      assert(!spark.sharedState.cacheManager.isEmpty)
```

@hvanhovell Above is the example.

hvanhovell:
Thanks! Isn't this an existing (separate) problem, that we should fix?

gatorsmile:
Yeah! That is just a relavant problem. Found it when I reviewed another PR: https://github.com/apache/spark/pull/15682

hvanhovell:
I agree that it is relevant, and that we should fix it for 2.1.

PR #15898, opened 2016-11-16
[SPARK-18457][SQL] ORC and other columnar formats using HiveShim read all columns when doing a simple count	

aray:

When reading zero columns (e.g., count(*)) from ORC or any other format that uses HiveShim, actually set the read column list to empty for Hive to use.

## How was this patch tested?

Query correctness is handled by existing unit tests. I'm happy to add more if anyone can point out some case that is not covered.

Reduction in data read can be verified in the UI when built with a recent version of Hadoop say:
```
build/mvn -Pyarn -Phadoop-2.7 -Dhadoop.version=2.7.0 -Phive -DskipTests clean package
```
However the default Hadoop 2.2 that is used for unit tests does not report actual bytes read and instead just full file sizes (see FileScanRDD.scala line 80). Therefore I don't think there is a good way to add a unit test for this.

I tested with the following setup using above build options
```
case class OrcData(intField: Long, stringField: String)
spark.range(1,1000000).map(i => OrcData(i, s"part-$i")).toDF().write.format("orc").save("orc_test")

sql(
      s"""CREATE EXTERNAL TABLE orc_test(
         |  intField LONG,
         |  stringField STRING
         |)
         |STORED AS ORC
         |LOCATION '${System.getProperty("user.dir") + "/orc_test"}'
       """.stripMargin)
```

## Results

query | Spark 2.0.2 | this PR
---|---|---
`sql("select count(*) from orc_test").collect`|4.4 MB|199.4 KB
`sql("select intField from orc_test").collect`|743.4 KB|743.4 KB
`sql("select * from orc_test").collect`|4.4 MB|4.4 MB


rxin:
cc @piaozhexiu I think you changed this last?

rxin:
Can we add a unit test for this function?

HyukjinKwon:
I was just watching the JIRA just for my curiosity. Actually, aren't we able to add test for `SparkOrcNewRecordReader` like the one as below?

```scala
test("Empty schema does not read data from ORC file") {
  val data = (1 to 2).zip(1 to 2)
  withOrcFile(data) { path =>
    val requrestedSchema = StructType(Nil)
    val conf = new Configuration()
    val physicalSchema = OrcFileOperator.readSchema(Seq(path), Some(conf)).get
    OrcRelation.setRequiredColumns(conf, physicalSchema, requrestedSchema)
    val maybeOrcReader = OrcFileOperator.getFileReader(path, Some(conf))
    assert(maybeOrcReader.isDefined)
    val orcRecordReader = new SparkOrcNewRecordReader(
      maybeOrcReader.get, conf, 0, maybeOrcReader.get.getContentLength)

    val recordsIterator = new RecordReaderIterator[OrcStruct](orcRecordReader)
    assert(recordsIterator.next().toString == "{null, null}")
      
    recordsIterator.close()
  }
}
```

aray:
The code that is being changed originated 2 years ago with the addition of Hive 0.13 support by @zhzhan, see 
https://github.com/apache/spark/commit/7c89a8f0c81ecf91dba34c1f44393f45845d438c#diff-77621c6c599cfdd24363967a500be1c5R109
Before that there was not even a null check.

I will add the test you are suggesting @HyukjinKwon

sql/hive/src/test/scala/org/apache/spark/sql/hive/orc/OrcQuerySuite.scala
```diff
@@ -577,4 +579,23 @@ class OrcQuerySuite extends QueryTest with BeforeAndAfterAll with OrcTest {
       assert(spark.table(tableName).schema == schema.copy(fields = expectedFields))
     }
   }
+
+  test("Empty schema does not read data from ORC file") {
+    val data = (1 to 2).zip(1 to 2)
+    withOrcFile(data) { path =>
+      val requrestedSchema = StructType(Nil)
```
tejasapatil: typo in variable name

sql/hive/src/test/scala/org/apache/spark/sql/hive/orc/OrcQuerySuite.scala
```diff
@@ -577,4 +579,23 @@ class OrcQuerySuite extends QueryTest with BeforeAndAfterAll with OrcTest {
       assert(spark.table(tableName).schema == schema.copy(fields = expectedFields))
     }
   }
+
+  test("Empty schema does not read data from ORC file") {
+    val data = (1 to 2).zip(1 to 2)
```
tejasapatil: feels weird to read. Would rather prefer simple `Seq((1,1), (2,2))`

sql/hive/src/test/scala/org/apache/spark/sql/hive/orc/OrcQuerySuite.scala
```diff
@@ -577,4 +579,23 @@ class OrcQuerySuite extends QueryTest with BeforeAndAfterAll with OrcTest {
...
+      OrcRelation.setRequiredColumns(conf, physicalSchema, requrestedSchema)
+      val maybeOrcReader = OrcFileOperator.getFileReader(path, Some(conf))
+      assert(maybeOrcReader.isDefined)
+      val orcRecordReader = new SparkOrcNewRecordReader(
+        maybeOrcReader.get, conf, 0, maybeOrcReader.get.getContentLength)
+
+      val recordsIterator = new RecordReaderIterator[OrcStruct](orcRecordReader)
+      assert(recordsIterator.next().toString == "{null, null}")
+
+      recordsIterator.close()
```
HyukjinKwon: Maybe, we should put some sugar on this something like `try {..} finally {recordsIterator.close()}`. I just wanted to show that it might be able to be tested :). if this test looks good to other guys as well.

tejasapatil:
Feels like this will only help for `select count(*) from table` without any projections and filters. 

Which version of ORC have you tested against ? In past I have seen PRs which do not work over older version of ORC and needed some fixes so it would be good to check.

aray:
@tejasapatil yes that is the use case where this applies. It's only tested against whatever version is included in the hadoop2.7+hive build configuration listed above. Is there anything in particular you were concerned about compatibility wise?

rxin:
@tejasapatil does this lgtu?

sql/hive/src/test/scala/org/apache/spark/sql/hive/orc/OrcQuerySuite.scala
```diff
@@ -577,4 +579,28 @@ class OrcQuerySuite extends QueryTest with BeforeAndAfterAll with OrcTest {
...
+      val maybeOrcReader = OrcFileOperator.getFileReader(path, Some(conf))
+      assert(maybeOrcReader.isDefined)
+      val orcRecordReader = new SparkOrcNewRecordReader(
+        maybeOrcReader.get, conf, 0, maybeOrcReader.get.getContentLength)
+
+      val recordsIterator = new RecordReaderIterator[OrcStruct](orcRecordReader)
+      try {
+        assert(recordsIterator.next().toString == "{null, null}")
+      } catch {
+        case e: Exception => fail(e)
```
rxin: why bother catching? the test case will fail anyway wouldn't it?

rxin:

PR #15901, opened 2016-11-16
[SPARK-18467][SQL] Extracts method for preparing arguments from StaticInvoke, Invoke and NewInstance and modify to short circuit if arguments have null when `needNullCheck == true`.

ueshin:

This pr extracts method for preparing arguments from `StaticInvoke`, `Invoke` and `NewInstance` and modify to short circuit if arguments have `null` when `propageteNull == true`.

The steps are as follows:

1. Introduce `InvokeLike` to extract common logic from `StaticInvoke`, `Invoke` and `NewInstance` to prepare arguments.
`StaticInvoke` and `Invoke` had a risk to exceed 64kb JVM limit to prepare arguments but after this patch they can handle them because they share the preparing code of NewInstance, which handles the limit well.

2. Remove unneeded null checking and fix nullability of `NewInstance`.
Avoid some of nullabilty checking which are not needed because the expression is not nullable.

3. Modify to short circuit if arguments have `null` when `needNullCheck == true`.
If `needNullCheck == true`, preparing arguments can be skipped if we found one of them is `null`, so modified to short circuit in the case.

## How was this patch tested?

Existing tests.

### DIFF
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
@@ -59,16 +59,20 @@ trait InvokeLike extends Expression with NonSQLExpression {
   def prepareArguments(ctx: CodegenContext): (String, String, String) = {
 
     val resultIsNull = if (needNullCheck) {
-      ctx.freshName("resultIsNull")
+      val resultIsNull = ctx.freshName("resultIsNull")
+      ctx.addMutableState("boolean", resultIsNull, "")
+      resultIsNull
     } else {
       "false"
     }
     val argValues = arguments.map { e =>
-      ctx.freshName("argValue")
+      val argValue = ctx.freshName("argValue")
+      ctx.addMutableState(ctx.javaType(e.dataType), argValue, "")
+      argValue
     }
 
     val argCodes = if (needNullCheck) {
-      val reset = s"boolean $resultIsNull = false;"
+      val reset = s"$resultIsNull = false;"
       val argCodes = arguments.zipWithIndex.map { case (e, i) =>
         val expr = e.genCode(ctx)
         val updateResultIsNull = if (e.nullable) {
@@ -77,7 +81,6 @@ trait InvokeLike extends Expression with NonSQLExpression {
           ""
         }
         s"""
-          ${ctx.javaType(e.dataType)} ${argValues(i)} = ${ctx.defaultValue(e.dataType)};
           if (!$resultIsNull) {
             ${expr.code}
             $updateResultIsNull
@@ -91,11 +94,11 @@ trait InvokeLike extends Expression with NonSQLExpression {
         val expr = e.genCode(ctx)
         s"""
           ${expr.code}
-          ${ctx.javaType(e.dataType)} ${argValues(i)} = ${expr.value};
+          ${argValues(i)} = ${expr.value};
         """
       }
     }
-    val argCode = argCodes.mkString("\n")
+    val argCode = ctx.splitExpressions(ctx.INPUT_ROW, argCodes)
 
     (argCode, argValues.mkString(", "), resultIsNull)
   }
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
@@ -150,7 +150,7 @@ case class StaticInvoke(
     val code = s"""
       $argCode
       boolean ${ev.isNull} = $resultIsNull;
-      final $javaType ${ev.value} = ${ev.isNull} ? ${ctx.defaultValue(dataType)} : $callFunc;
+      final $javaType ${ev.value} = $resultIsNull ? ${ctx.defaultValue(dataType)} : $callFunc;
       $postNullCheck
      """
     ev.copy(code = code)
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
@@ -59,10 +59,10 @@ trait InvokeLike extends Expression {
    */
   def prepareArguments(ctx: CodegenContext, ev: ExprCode): (String, String, String) = {
 
-    val argsHaveNull = if (propagatingNull) {
-      val argsHaveNull = ctx.freshName("argsHaveNull")
-      ctx.addMutableState("boolean", argsHaveNull, "")
-      argsHaveNull
+    val containsNullInArguments = if (propagatingNull) {
+      val containsNullInArguments = ctx.freshName("containsNullInArguments")
+      ctx.addMutableState("boolean", containsNullInArguments, "")
+      containsNullInArguments
     } else {
       ""
     }
@@ -73,16 +73,16 @@ trait InvokeLike extends Expression {
     }
 
     val argCodes = if (propagatingNull) {
-      val reset = s"$argsHaveNull = false;"
+      val reset = s"$containsNullInArguments = false;"
       val argCodes = arguments.zipWithIndex.map { case (e, i) =>
         val expr = e.genCode(ctx)
         s"""
-          if (!$argsHaveNull) {
+          if (!$containsNullInArguments) {
             ${expr.code}
         """ +
           (if (e.nullable) {
             s"""
-              $argsHaveNull = ${expr.isNull};
+              $containsNullInArguments = ${expr.isNull};
               ${argValues(i)} = ${expr.value};
             """
           } else {
@@ -105,7 +105,7 @@ trait InvokeLike extends Expression {
     val argCode = ctx.splitExpressions(ctx.INPUT_ROW, argCodes)
 
     val setIsNull = if (propagatingNull) {
-      s"${ev.isNull} = ${ev.isNull} || $argsHaveNull;"
+      s"${ev.isNull} = ${ev.isNull} || $containsNullInArguments;"
     } else {
       ""
     }
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
@@ -65,7 +65,7 @@ trait InvokeLike extends Expression with NonSQLExpression {
     } else {
       "false"
     }
-    val argValues = arguments.zipWithIndex.map { case (e, i) =>
+    val argValues = arguments.map { e =>
       val argValue = ctx.freshName("argValue")
       ctx.addMutableState(ctx.javaType(e.dataType), argValue, "")
       argValue
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
@@ -41,7 +41,7 @@ trait InvokeLike extends Expression with NonSQLExpression {
 
   def propagateNull: Boolean
 
-  protected lazy val propagatingNull: Boolean = propagateNull && arguments.exists(_.nullable)
+  protected lazy val needNullCheck: Boolean = propagateNull && arguments.exists(_.nullable)
 
   /**
    * Prepares codes for arguments.
@@ -50,7 +50,7 @@ trait InvokeLike extends Expression with NonSQLExpression {
    * - use ctx.splitExpressions() to not exceed 64kb JVM limit while preparing arguments.
    * - avoid some of nullabilty checking which are not needed because the expression is not
    *   nullable.
-   * - when progagatingNull == true, short circuit if we found one of arguments is null because
+   * - when needNullCheck == true, short circuit if we found one of arguments is null because
    *   preparing rest of arguments can be skipped in the case.
    *
    * @param ctx a [[CodegenContext]]
@@ -59,7 +59,7 @@ trait InvokeLike extends Expression with NonSQLExpression {
    */
   def prepareArguments(ctx: CodegenContext, ev: ExprCode): (String, String, String) = {
 
-    val containsNullInArguments = if (propagatingNull) {
+    val containsNullInArguments = if (needNullCheck) {
       val containsNullInArguments = ctx.freshName("containsNullInArguments")
       ctx.addMutableState("boolean", containsNullInArguments, "")
       containsNullInArguments
@@ -72,7 +72,7 @@ trait InvokeLike extends Expression with NonSQLExpression {
       argValue
     }
 
-    val argCodes = if (propagatingNull) {
+    val argCodes = if (needNullCheck) {
       val reset = s"$containsNullInArguments = false;"
       val argCodes = arguments.zipWithIndex.map { case (e, i) =>
         val expr = e.genCode(ctx)
@@ -101,7 +101,7 @@ trait InvokeLike extends Expression with NonSQLExpression {
     }
     val argCode = ctx.splitExpressions(ctx.INPUT_ROW, argCodes)
 
-    val setIsNull = if (propagatingNull) {
+    val setIsNull = if (needNullCheck) {
       s"${ev.isNull} = ${ev.isNull} || $containsNullInArguments;"
     } else {
       ""
@@ -297,7 +297,7 @@ case class NewInstance(
     outerPointer: Option[() => AnyRef]) extends InvokeLike {
   private val className = cls.getName
 
-  override def nullable: Boolean = propagatingNull
+  override def nullable: Boolean = needNullCheck
 
   override def children: Seq[Expression] = arguments
 
@@ -322,7 +322,7 @@ case class NewInstance(
     val outer = outerPointer.map(func => Literal.fromObject(func()).genCode(ctx))
 
     var isNull = ev.isNull
-    val prepareIsNull = if (propagatingNull) {
+    val prepareIsNull = if (needNullCheck) {
       s"boolean $isNull = false;"
     } else {
       isNull = "false"
@@ -341,7 +341,7 @@ case class NewInstance(
       $prepareIsNull
       $setIsNull
     """ +
-      (if (propagatingNull) {
+      (if (needNullCheck) {
         s"final $javaType ${ev.value} = $isNull ? ${ctx.defaultValue(javaType)} : $constructorCall;"
       } else {
         s"final $javaType ${ev.value} = $constructorCall;"
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
@@ -205,8 +205,9 @@ case class Invoke(
 
     val evaluateArguments = if (arguments.nonEmpty) {
       s"""
-        if (!${obj.isNull}) {
+        if (!${ev.isNull}) {
           $argCode
+          ${ev.isNull} = $resultIsNull;
         }
       """
     } else {
@@ -253,8 +254,8 @@ case class Invoke(
 
     val code = s"""
       ${obj.code}
+      boolean ${ev.isNull} = ${obj.isNull};
       $evaluateArguments
-      boolean ${ev.isNull} = ${obj.isNull} || $resultIsNull;
       $javaType ${ev.value} = ${ctx.defaultValue(dataType)};
       if (!${ev.isNull}) {
         $evaluate
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
@@ -35,7 +35,7 @@ import org.apache.spark.sql.types._
 /**
  * Common base class for [[StaticInvoke]], [[Invoke]], and [[NewInstance]].
  */
-trait InvokeLike extends Expression {
+trait InvokeLike {
 
   def arguments: Seq[Expression]
 
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
@@ -35,7 +35,7 @@ import org.apache.spark.sql.types._
 /**
  * Common base class for [[StaticInvoke]], [[Invoke]], and [[NewInstance]].
  */
-trait InvokeLike {
+trait InvokeLike extends Expression with NonSQLExpression {
 
   def arguments: Seq[Expression]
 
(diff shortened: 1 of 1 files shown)


rxin:
So the refactoring for simplification adds more code than it deletes? Did you introduce new features as well?

ueshin:
@rxin I introduced features as follows:

1. `StaticInvoke` and `Invoke` had a risk to exceed 64kb JVM limit to prepare arguments but after this patch they can handle them because they share the preparing code of `NewInstance`, which handles the limit well.
2. Avoid some of nullabilty checking which are not needed because the expression is not nullable.
3. If `progagateNull == true`, preparing arguments can be skipped if we found one of them is `null`, so modified to short circuit in the case.

rxin:
cc @cloud-fan

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
```diff
@@ -33,6 +33,73 @@ import org.apache.spark.sql.catalyst.util.{ArrayBasedMapData, GenericArrayData}
...
+trait InvokeLike extends Expression {
+
+  def arguments: Seq[Expression]
+
+  def propagateNull: Boolean
+
+  def prepareArguments(ctx: CodegenContext, ev: ExprCode): (String, String, String) = {
+
+    val argsHaveNull =
+      if (propagateNull && arguments.exists(_.nullable)) {
```
viirya: `propagateNull && arguments.exists(_.nullable)` is used many times, define a variable for it?

viirya:
This looks good overall. The more codes added are mostly used to deal with short circuit cases of nullability.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
```diff
@@ -109,7 +171,7 @@ case class Invoke(
     functionName: String,
     dataType: DataType,
     arguments: Seq[Expression] = Nil,
-    propagateNull: Boolean = true) extends Expression with NonSQLExpression {
+    propagateNull: Boolean = true) extends Expression with InvokeLike with NonSQLExpression {
```
viirya: The comment of `Invoke` doesn't have explanation for the param `propagateNull`. Can you also add it too?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
```diff
@@ -33,6 +33,73 @@ import org.apache.spark.sql.catalyst.util.{ArrayBasedMapData, GenericArrayData}
...
 /**
+ * Common base class for [[StaticInvoke]], [[Invoke]], and [[NewInstance]].
+ */
+trait InvokeLike extends Expression {
+
+  def arguments: Seq[Expression]
+
+  def propagateNull: Boolean
+
+  def prepareArguments(ctx: CodegenContext, ev: ExprCode): (String, String, String) = {
```
cloud-fan: +1

viirya:
The title is not well explained this pr and can be updated to reflect it.

The explanation of the features you added can be moved to the pr description.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
```diff
@@ -33,6 +33,73 @@ import org.apache.spark.sql.catalyst.util.{ArrayBasedMapData, GenericArrayData}
...
+
+    val argsHaveNull =
```
cloud-fan: this variable name looks like a boolean flag, can you think of a better name?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
cloud-fan: +1

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
```diff
@@ -33,6 +33,73 @@ import org.apache.spark.sql.catalyst.util.{ArrayBasedMapData, GenericArrayData}
 import org.apache.spark.sql.types._
 
 /**
+ * Common base class for [[StaticInvoke]], [[Invoke]], and [[NewInstance]].
+ */
+trait InvokeLike extends Expression {
```
viirya: Does it need to extend `Expression`?

ueshin:
@viirya @cloud-fan I addressed your comments and updated pr title and description.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
```diff
@@ -33,6 +33,88 @@ import org.apache.spark.sql.catalyst.util.{ArrayBasedMapData, GenericArrayData}
...
+
+  def propagateNull: Boolean
+
+  protected lazy val propagatingNull: Boolean = propagateNull && arguments.exists(_.nullable)
+
+  /**
+   * Prepares codes for arguments.
+   *
+   * - generate codes for argument.
+   * - use ctx.splitExpressions() not to exceed 64kb JVM limit while preparing arguments.
```
cloud-fan: nit: `not to` -> `to not`

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
```diff
@@ -33,6 +33,88 @@ import org.apache.spark.sql.catalyst.util.{ArrayBasedMapData, GenericArrayData}
...
+      val argValue = ctx.freshName("argValue")
+      ctx.addMutableState(ctx.javaType(e.dataType), argValue, "")
+      argValue
+    }
+
+    val argCodes = if (propagatingNull) {
+      val reset = s"$containsNullInArguments = false;"
+      val argCodes = arguments.zipWithIndex.map { case (e, i) =>
+        val expr = e.genCode(ctx)
+        s"""
```
cloud-fan: how about:
```
val updateContainsNull = if (e.nullable) {
  "$containsNull = ${expr.isNull};"
} else {
  ""
}
s"""
  if (!$containsNull) {
      ${expr.code}
      $updateContainsNull
      ${argValues(i)} = ${expr.value};
  }
"""
```

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
```diff
@@ -50,7 +132,7 @@ case class StaticInvoke(
     dataType: DataType,
     functionName: String,
     arguments: Seq[Expression] = Nil,
-    propagateNull: Boolean = true) extends Expression with NonSQLExpression {
+    propagateNull: Boolean = true) extends Expression with InvokeLike with NonSQLExpression {
```
cloud-fan: why not make `InvokeLike` extends `Expression` and `NonSQLExpression`, then we can just `extends `InvokeLike` here?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
```diff
@@ -245,51 +319,36 @@ case class NewInstance(
...
 
     var isNull = ev.isNull
-    val setIsNull = if (propagateNull && arguments.nonEmpty) {
-      s"""
-       boolean $isNull = false;
-       for (int idx = 0; idx < ${arguments.length}; idx++) {
-         if ($argIsNulls[idx]) { $isNull = true; break; }
-       }
-     """
+    val prepareIsNull = if (propagateNull && arguments.exists(_.nullable)) {
```
cloud-fan: use `propogatingNull`?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
ueshin: Thanks, fixed.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
ueshin: Thanks, I'll use it.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
ueshin: Makes sense, I'll modify them.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
ueshin: Oops, I missed it, fixed.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
```diff
@@ -33,6 +33,88 @@ import org.apache.spark.sql.catalyst.util.{ArrayBasedMapData, GenericArrayData}
...
+  protected lazy val propagatingNull: Boolean = propagateNull && arguments.exists(_.nullable)
+
+  /**
+   * Prepares codes for arguments.
+   *
+   * - generate codes for argument.
+   * - use ctx.splitExpressions() not to exceed 64kb JVM limit while preparing arguments.
+   * - avoid some of nullabilty checking which are not needed because the expression is not
+   *   nullable.
+   * - when progagateNull == true, short circuit if we found one of arguments is null because
```
ueshin: Thanks, I used `needNullCheck` for now.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
```diff
@@ -33,6 +33,85 @@ import org.apache.spark.sql.catalyst.util.{ArrayBasedMapData, GenericArrayData}
...
+   * - when needNullCheck == true, short circuit if we found one of arguments is null because
+   *   preparing rest of arguments can be skipped in the case.
+   *
+   * @param ctx a [[CodegenContext]]
+   * @param ev an [[ExprCode]] with unique terms.
+   * @return (code to prepare arguments, argument string, code to set isNull)
+   */
+  def prepareArguments(ctx: CodegenContext, ev: ExprCode): (String, String, String) = {
+
+    val containsNullInArguments = if (needNullCheck) {
```
cloud-fan: how about
```
val resultIsNull = if (needNullCheck) {
  ...
} else {
  "false"
}
```
Then we return this `resultIsNull`, and the caller side can just generate code like
```
${ev.isNull} = $resultIsNull
${ev.value} = ...
$postNullCheck
```

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
```diff
@@ -179,7 +249,8 @@ case class Invoke(
     }
     val code = s"""
       ${obj.code}
-      ${argGen.map(_.code).mkString("\n")}
+      $argCode
```
cloud-fan: we don't need to evaluate the arguments if `obj.isNull == true`

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
```diff
@@ -33,6 +33,85 @@ import org.apache.spark.sql.catalyst.util.{ArrayBasedMapData, GenericArrayData}
...
+      val reset = s"$containsNullInArguments = false;"
+      val argCodes = arguments.zipWithIndex.map { case (e, i) =>
+        val expr = e.genCode(ctx)
+        val updateContainsNull = if (e.nullable) {
+          s"$containsNullInArguments = ${expr.isNull};"
+        } else {
+          ""
+        }
+        s"""
+          if (!$containsNullInArguments) {
```
kiszk: if `containsNullInArguments == "true"`, we do not need generate this code block. Am I correct?

viirya: `containsNullInArguments` is known at runtime.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
```diff
@@ -245,51 +316,36 @@ case class NewInstance(
...
     val code = s"""
       $argCode
       ${outer.map(_.code).getOrElse("")}
+      $prepareIsNull
       $setIsNull
-      final $javaType ${ev.value} = $isNull ? ${ctx.defaultValue(javaType)} : $constructorCall;
-     """
+    """ +
+      (if (needNullCheck) {
+        s"final $javaType ${ev.value} = $isNull ? ${ctx.defaultValue(javaType)} : $constructorCall;"
```
viirya: If `needNullCheck` is true, we don't need to evaluate arguments if `$isNull == true`.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
ueshin: @kiszk I'm sorry but I didn't understand what you mean.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
ueshin: @cloud-fan Thanks, I updated.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
ueshin: @cloud-fan Thanks, I updated.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
ueshin: @viirya I think we do need because if `needNullCheck` is `true`, the `$isNull` stores the result of evaluating arguments.

ueshin: Or am I misunderstanding what you mean..?

viirya: yeah, I see. This is correct.

viirya:
LGTM cc @cloud-fan for another check.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
```diff
@@ -33,6 +33,79 @@ import org.apache.spark.sql.catalyst.util.{ArrayBasedMapData, GenericArrayData}
...
+   * - avoid some of nullabilty checking which are not needed because the expression is not
+   *   nullable.
+   * - when needNullCheck == true, short circuit if we found one of arguments is null because
+   *   preparing rest of arguments can be skipped in the case.
+   *
+   * @param ctx a [[CodegenContext]]
+   * @param ev an [[ExprCode]] with unique terms.
+   * @return (code to prepare arguments, argument string, result of argument null check)
+   */
+  def prepareArguments(ctx: CodegenContext, ev: ExprCode): (String, String, String) = {
```
cloud-fan: `ev` is not used.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
```diff
@@ -164,23 +233,20 @@ case class Invoke(
...
     }
+
     val code = s"""
       ${obj.code}
-      ${argGen.map(_.code).mkString("\n")}
-      $setIsNull
+      if (!${obj.isNull}) {
+        $argCode
+      }
+      boolean ${ev.isNull} = ${obj.isNull} || $resultIsNull;
```
cloud-fan: this assumes `resultIsNull` is a member variable not local variable, can we avoid doing this? We can make the code clearer and more robust like:
```
javaType ${ev.value} = defaultValue;
boolean ${ev.isNull} = true;
if (!${obj.isNull}) {
  $argCode
  ${ev.isNull} = $resultIsNull
  ...
  $postNullCheck
}

```

ueshin: I'm afraid not because if evaluating arguments is split to some methods, `resultIsNull` can't be referred from the split methods.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
```diff
@@ -245,51 +311,35 @@ case class NewInstance(
...
     val code = s"""
       $argCode
       ${outer.map(_.code).getOrElse("")}
-      $setIsNull
-      final $javaType ${ev.value} = $isNull ? ${ctx.defaultValue(javaType)} : $constructorCall;
-     """
+      $prepareIsNull
+    """ +
+      (if (needNullCheck) {
+        s"final $javaType ${ev.value} = $isNull ? ${ctx.defaultValue(javaType)} : $constructorCall;"
```
cloud-fan: is it necessary? I'm pretty sure javac can optimize `... = false ? xxx : xxx` for us.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
cloud-fan: I know `resultIsNull` must be a member variable now, but in the future we may make it a local variable if the argument code is not so long to exceed the 64kb limit. Here why not we assume `resultIsNull` can be local variable and make us more robust?

cloud-fan:
LGTM except some minor comments, thanks for working on it.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
kiszk: Do we use janino instead of javac?

cloud-fan: sorry I mean "java compiler"...  I think janino is smart enough about it.

viirya: Yeah. I go to check janino source code. It has the optimization: https://github.com/janino-compiler/janino/blob/22a7787e62037049e337cb1ce4064c29a4856022/janino/src/main/java/org/codehaus/janino/UnitCompiler.java#L4247.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
ueshin: Thanks, removed.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
ueshin: I see, updated. Thanks.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
ueshin: I modified the code as you mentioned.

But I'm sorry, I didn't understand what you want to do.

- You are assuming that the argument code is not so long, so use local variable without splitting evaluations? (could also make the argument variable as local variable)
  - I think the argument code would easily become large enough to exceed the 64kb limit.
- You want to modify only `resultIsNull` to local variable and pass it and return the value to the split methods?
  - We need to extend the method `ctx.splitExpressions()` to be able to pass and return value and handle it as #15480 is trying.
- You want to switch local variable or member variable by whether split or not?
  - We need to know whether split or not but we can't now. Modify `ctx.splitExpressions()` to return split or not?

Could you let me know if you have other thoughts?

cloud-fan: > You are assuming that the argument code is not so long, so use local variable without splitting evaluations? (could also make the argument variable as local variable)

This is what I mean. We can generate code first and see if it exceeds the 64kb limitation, so logically we do have a chance to use local variables.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
```diff
@@ -245,51 +311,31 @@ case class NewInstance(
...
     var isNull = ev.isNull
-    val setIsNull = if (propagateNull && arguments.nonEmpty) {
-      s"""
-       boolean $isNull = false;
-       for (int idx = 0; idx < ${arguments.length}; idx++) {
-         if ($argIsNulls[idx]) { $isNull = true; break; }
-       }
-     """
+    val prepareIsNull = if (needNullCheck) {
+      s"boolean $isNull = $resultIsNull;"
```
cloud-fan: one more optimization: we can just write `ev.isNull = resultIsNull`, then if `resultIsNull` is `false`, the final evaluation will become `... = false ? aaa : bbb` and get optimized by janino.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
```diff
@@ -164,28 +232,26 @@ case class Invoke(
...
+
     val code = s"""
       ${obj.code}
-      ${argGen.map(_.code).mkString("\n")}
-      $setIsNull
+      boolean ${ev.isNull} = true;
       $javaType ${ev.value} = ${ctx.defaultValue(dataType)};
-      if (!${ev.isNull}) {
-        $evaluate
+      if (!${obj.isNull}) {
```
ueshin: Yes, I agree with you. Thank you for your suggestion.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
ueshin: Ah, I see what you mean, modified. Thanks.

ueshin:
f8acda6 is a test to see if it exceeds the 64kb JVM limit or not, using only local variables.
We'll revert if some test failed.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala
```diff
@@ -82,8 +148,8 @@ case class StaticInvoke(
     }
 
     val code = s"""
-      ${argGen.map(_.code).mkString("\n")}
-      $setIsNull
+      $argCode
+      boolean ${ev.isNull} = $resultIsNull;
       final $javaType ${ev.value} = ${ev.isNull} ? ${ctx.defaultValue(dataType)} : $callFunc;
```
viirya: The `${ev.isNull}` at line 153 can be `$resultIsNull`. So the optimization for `false ? ${ctx.defaultValue(dataType)} : $callFunc;` can work here too?

viirya:
LGTM cc @cloud-fan for another look.

cloud-fan:

PR #15902, opened 2016-11-16
[SPARK-18430][SQL][BACKPORT-2.0] Fixed Exception Messages when Hitting an Invocation Exception of Function Lookup

gatorsmile:
This PR is to backport https://github.com/apache/spark/pull/15878

When the exception is an invocation exception during function lookup, we return a useless/confusing error message:

For example, 
```Scala
df.selectExpr("concat_ws()")
```
Below is the error message we got:
```
null; line 1 pos 0
org.apache.spark.sql.AnalysisException: null; line 1 pos 0
```

To get the meaningful error message, we need to get the cause. The fix is exactly the same as what we did in https://github.com/apache/spark/pull/12136. After the fix, the message we got is the exception issued in the constuctor of function implementation:
```
requirement failed: concat_ws requires at least one argument.; line 1 pos 0
org.apache.spark.sql.AnalysisException: requirement failed: concat_ws requires at least one argument.; line 1 pos 0
```

Added test cases.


hvanhovell:
LGTM - merging to branch-2.0. Can you close?

gatorsmile:
Thanks!

PR #15904, opened 2016-11-16
[SPARK-18470][STREAMING][WIP] Provide Spark Streaming Monitor Rest Api

uncleGen:

Provide Spark Streaming Monitor Rest Api

 - /api/v1/applications/[app-id]/streaming/batch
---
image

- /api/v1/applications/[app-id]/streaming/receivers
---
image

## How was this patch tested?

new unit test

### DIFF
--- a/core/src/main/scala/org/apache/spark/ui/SparkUI.scala
+++ b/core/src/main/scala/org/apache/spark/ui/SparkUI.scala
@@ -60,6 +60,8 @@ private[spark] class SparkUI private (
 
   var appId: String = _
 
+  var streamingJobProgressListener: Option[SparkListener] = None
+
   /** Initialize all components of the server. */
   def initialize() {
     val jobsTab = new JobsTab(this)
@@ -120,6 +122,12 @@ private[spark] class SparkUI private (
       ))
     ))
   }
+
+  def setStreamingListener(streamingJobProgressListener: SparkListener): Unit = {
+    this.streamingJobProgressListener = Some(streamingJobProgressListener)
+  }
+
+  def getStreamingListener: Option[SparkListener] = streamingJobProgressListener
 
   def getApplicationInfo(appId: String): Option[ApplicationInfo] = {
     getApplicationInfoList.find(_.id == appId)
--- a/streaming/src/main/scala/org/apache/spark/streaming/StreamingContext.scala
+++ b/streaming/src/main/scala/org/apache/spark/streaming/StreamingContext.scala
@@ -184,7 +184,11 @@ class StreamingContext private[streaming] (
 
   private[streaming] val waiter = new ContextWaiter
 
-  private[streaming] val progressListener = new StreamingJobProgressListener(this)
+  private[streaming] val progressListener = {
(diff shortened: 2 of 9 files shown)


BalaVuraMSM:
(Y)

BalaVuraMSM:
:+1:

ajbozarth:
I haven't had a chance to read your code yet, but someone opened a similar pr soon after you. Given you probably concurrently developed them you may wan to take a look.

uncleGen:
@ajbozarth Thank you for reminding me, i will take a look at it later.

uncleGen:
@saturday-shi  let us improve this.

core/src/main/scala/org/apache/spark/ui/SparkUI.scala
```diff
@@ -121,6 +123,12 @@ private[spark] class SparkUI private (
     ))
   }
 
+  def setStreamingListener(streamingJobProgressListener: SparkListener): Unit = {
+    this.streamingJobProgressListener = Some(streamingJobProgressListener)
```
vanzin: `Option(streamingJobProgressListener)`. Just in case.

streaming/src/main/scala/org/apache/spark/streaming/StreamingContext.scala
```diff
@@ -184,7 +184,11 @@ class StreamingContext private[streaming] (
 
   private[streaming] val waiter = new ContextWaiter
 
-  private[streaming] val progressListener = new StreamingJobProgressListener(this)
+  private[streaming] val progressListener = {
+    val streamingListener = new StreamingJobProgressListener(this)
+    sc.ui.get.setStreamingListener(streamingListener)
```
vanzin: This is probably what's causing the unit test failures. The UI might not exist, which is why it's an `Option[SparkUI]` and not just a plain `SparkUI`.

vanzin (commented):
Hey @uncleGen,

Just so I understand, are the resource classes being loaded automatically by jersey? Because I don't see any code to explicitly add the new endpoints to the API. Maybe the following line in the API setup code does it?

```
    holder.setInitParameter(ServerProperties.PROVIDER_PACKAGES, "org.apache.spark.status.api.v1")
```

I can't really comment much on the model, but it would be nice to compare it to the other PR and see what the differences are and whether it might make sense to implement things that the other PR does and this one does not.

Also, the streaming types can probably live in the streaming module too (instead of in `api.scala`).

uncleGen:
@vanzin Sorry for the delay, I will update as soon as possible on the basis of your comments.

saturday-shi:
@uncleGen 
It seems you don't have much time to go on with this. So I opened a new PR (#16253) inherit all functions from the old one, with the only change of merge them into the current api v1.

uncleGen:
@saturday-shi Well, we should not spend much time to do the repetitive work. So, if you make sure to complete this work, I will re-close this PR. OK?

saturday-shi:
@uncleGen 
Exactly I will try my best to complete this.

PR #15905, opened 2016-11-16

AnthonyTruchet:

block in-place.

## How was this patch tested?

Tests run by hand locally. 
(Setting up local infra to run the official Spark tests is in progress)

### DIFF
--- a/mllib/src/main/scala/org/apache/spark/mllib/optimization/LBFGS.scala
+++ b/mllib/src/main/scala/org/apache/spark/mllib/optimization/LBFGS.scala
@@ -241,16 +241,28 @@ object LBFGS extends Logging {
       val bcW = data.context.broadcast(w)
       val localGradient = gradient
 
-      val (gradientSum, lossSum) = data.treeAggregate((Vectors.zeros(n), 0.0))(
-          seqOp = (c, v) => (c, v) match { case ((grad, loss), (label, features)) =>
-            val l = localGradient.compute(
-              features, label, bcW.value, grad)
-            (grad, loss + l)
-          },
-          combOp = (c1, c2) => (c1, c2) match { case ((grad1, loss1), (grad2, loss2)) =>
-            axpy(1.0, grad2, grad1)
-            (grad1, loss1 + loss2)
-          })
+      /** Given (current accumulated gradient, current loss) and (label, features)
+       * tuples, updates the current gradient and current loss
+       */
+      val seqOp = (c: (Vector, Double), v: (Double, Vector)) => {
+            (c, v) match { case ((grad, loss), (label, features)) =>
+                val l = localGradient.compute(features, label, bcW.value, grad)
+                (grad, loss + l)
+            }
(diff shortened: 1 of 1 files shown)


mllib/src/main/scala/org/apache/spark/mllib/optimization/LBFGS.scala
```diff
@@ -241,16 +241,28 @@ object LBFGS extends Logging {
...
-          },
-          combOp = (c1, c2) => (c1, c2) match { case ((grad1, loss1), (grad2, loss2)) =>
-            axpy(1.0, grad2, grad1)
-            (grad1, loss1 + loss2)
-          })
+      /** Given (current accumulated gradient, current loss) and (label, features)
+       * tuples, updates the current gradient and current loss
+       */
+      val seqOp = (c: (Vector, Double), v: (Double, Vector)) => {
+            (c, v) match { case ((grad, loss), (label, features)) =>
```
srowen: Nit: unindent 2 spaces and you can remove the outer braces? same in the next function

mllib/src/main/scala/org/apache/spark/mllib/optimization/LBFGS.scala
```diff
@@ -241,16 +241,28 @@ object LBFGS extends Logging {
...
+
+      // Adds two (gradient, loss) tuples
+      val combOp = (c1: (Vector, Double), c2: (Vector, Double)) => {
+            (c1, c2) match { case ((grad1, loss1), (grad2, loss2)) =>
+                axpy(1.0, grad2, grad1)
+                (grad1, loss1 + loss2)
+           }
+      }
+
+      val (gradientSum, lossSum) = data.mapPartitions(it => {
```
srowen: Nit: `.mapPartitions { it =>`

srowen (commented):
Oh I get it, it's just that you find that the zero values accidentally gets into those function closures. Yes sounds good if it can be rewritten to avoid it. I think there are more instances of this pattern though, in NaiveBayes, ALS, etc. It would be cool to break out these operations even just for code clarity, but especially if it avoids some silent overhead.

mllib/src/main/scala/org/apache/spark/mllib/optimization/LBFGS.scala
AnthonyTruchet: ok will do it.

AnthonyTruchet:
By he way do you think that this should be addressed in core or just in each ML specific use ?

srowen:
I personally think it's good to be consistent. I think it's more readable to break out these function definitions, and, it seems like there's evidence it might avoid some unintended objects in a closure. Have a look for other instances of "seqOp = ..." etc and see which ones look like the same pattern that could be refactored.

AnthonyTruchet:
I missed part of my company guidelines. Closing this PR and creating a new one shortly from my company account. Sorry for the noise.

PR #15906, opened 2016-11-16
[YARN][DOC] Increasing NodeManager's heap size with External Shuffle Service

Devian-ua:

Suggest users to increase `NodeManager's` heap size if `External Shuffle Service` is enabled as
`NM` can spend a lot of time doing GC resulting in  shuffle operations being a bottleneck due to `Shuffle Read blocked time` bumped up. 
Also because of GC  `NodeManager` can use an enormous amount of CPU and cluster performance will suffer.
I have seen NodeManager using 5-13G RAM and up to 2700% CPU with `spark_shuffle` service on.

#### Added step 5:
shuffle_service


rxin:

PR #15907, opened 2016-11-16
[SPARK-18458][CORE] Fix signed integer overflow problem at an expression in RadixSort.java

kiszk:

This PR avoids that a result of an expression is negative due to signed integer overflow (e.g. 0x10?????? * 8 < 0). This PR casts each operand to `long` before executing a calculation. Since the result is interpreted as long, the result of the expression is positive.

## How was this patch tested?

Manually executed query82 of TPC-DS with 100TB

### DIFF
--- a/core/src/main/java/org/apache/spark/shuffle/sort/ShuffleInMemorySorter.java
+++ b/core/src/main/java/org/apache/spark/shuffle/sort/ShuffleInMemorySorter.java
@@ -176,7 +176,7 @@ final class ShuffleInMemorySorter {
     int offset = 0;
     if (useRadixSort) {
       offset = RadixSort.sort(
-        array, (long)pos,
+        array, pos,
         PackedRecordPointer.PARTITION_ID_START_BYTE_INDEX,
         PackedRecordPointer.PARTITION_ID_END_BYTE_INDEX, false, false);
     } else {
--- a/core/src/main/java/org/apache/spark/shuffle/sort/ShuffleInMemorySorter.java
+++ b/core/src/main/java/org/apache/spark/shuffle/sort/ShuffleInMemorySorter.java
@@ -177,8 +177,8 @@ final class ShuffleInMemorySorter {
     if (useRadixSort) {
       offset = RadixSort.sort(
         array, (long)pos,
-        (long)PackedRecordPointer.PARTITION_ID_START_BYTE_INDEX,
-        (long)PackedRecordPointer.PARTITION_ID_END_BYTE_INDEX, false, false);
+        PackedRecordPointer.PARTITION_ID_START_BYTE_INDEX,
+        PackedRecordPointer.PARTITION_ID_END_BYTE_INDEX, false, false);
     } else {
       MemoryBlock unused = new MemoryBlock(
         array.getBaseObject(),
--- a/core/src/main/java/org/apache/spark/shuffle/sort/ShuffleInMemorySorter.java
+++ b/core/src/main/java/org/apache/spark/shuffle/sort/ShuffleInMemorySorter.java
@@ -176,9 +176,9 @@ final class ShuffleInMemorySorter {
     int offset = 0;
     if (useRadixSort) {
       offset = RadixSort.sort(
-        array, pos,
-        PackedRecordPointer.PARTITION_ID_START_BYTE_INDEX,
-        PackedRecordPointer.PARTITION_ID_END_BYTE_INDEX, false, false);
+        array, (long)pos,
+        (long)PackedRecordPointer.PARTITION_ID_START_BYTE_INDEX,
+        (long)PackedRecordPointer.PARTITION_ID_END_BYTE_INDEX, false, false);
     } else {
       MemoryBlock unused = new MemoryBlock(
         array.getBaseObject(),
--- a/core/src/main/java/org/apache/spark/util/collection/unsafe/sort/RadixSort.java
+++ b/core/src/main/java/org/apache/spark/util/collection/unsafe/sort/RadixSort.java
@@ -242,7 +242,7 @@ public class RadixSort {
       boolean desc, boolean signed) {
     assert counts.length == 256;
     long[] offsets = transformCountsToOffsets(
-      counts, numRecords, array.getBaseOffset() + outIndex * 8, 16, desc, signed);
+      counts, numRecords, array.getBaseOffset() + outIndex * 8L, 16, desc, signed);
     Object baseObject = array.getBaseObject();
     long baseOffset = array.getBaseOffset() + inIndex * 8L;
     long maxOffset = baseOffset + numRecords * 16L;
--- a/core/src/main/java/org/apache/spark/util/collection/unsafe/sort/RadixSort.java
+++ b/core/src/main/java/org/apache/spark/util/collection/unsafe/sort/RadixSort.java
@@ -40,28 +40,28 @@ public class RadixSort {
    *         of always copying the data back to position zero for efficiency.
    */
   public static int sort(
-      LongArray array, int numRecords, int startByteIndex, int endByteIndex,
+      LongArray array, long numRecords, long startByteIndex, long endByteIndex,
       boolean desc, boolean signed) {
     assert startByteIndex >= 0 : "startByteIndex (" + startByteIndex + ") should >= 0";
     assert endByteIndex <= 7 : "endByteIndex (" + endByteIndex + ") should <= 7";
     assert endByteIndex > startByteIndex;
     assert numRecords * 2 <= array.size();
-    int inIndex = 0;
-    int outIndex = numRecords;
+    long inIndex = 0;
+    long outIndex = numRecords;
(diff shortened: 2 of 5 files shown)


kiszk:
@hvanhovell I found similar portions in `RadixSort.java` For example, at [line 85](https://github.com/apache/spark/blob/master/core/src/main/java/org/apache/spark/util/collection/unsafe/sort/RadixSort.java#L85).
Should we change these portions preventively while we do not have test suites?

hvanhovell:
cc @ericl

hvanhovell:
@kiszk yeah lets also address those

kiszk:
@hvanhovell as far as I know, I addressed

rxin:
Can you please have a more descriptive title? Presumably the pull request is not to add coredump.

kiszk:
@rxin Sure, I did.

ericl:
LGTM. I wonder if it's also worth changing all the ints to longs to be extra safe here with conversions, even if we know the values are bounded.

rxin:
Yea it might be a good idea to just use longs.

kiszk:
Ideally, it is a good idea. Do you want to do that change using this PR?

Another question is which scope do we apply such a change?
(1) Only in methods that I applied changes to.
(2) Only in 'RadixSort.java'
(3) Beyond public methods ('sort()' and 'sortKeyPrefixArray()')
Obviously, (3) requires a lot of changes to many files.

What do you think?

rxin:
I'm thinking 2. What do you think?

kiszk:
It would be good. 
In that case, is it better to insert explicit cast (from int to long) if a caller gives an int variable? This is why we want to explicitly express a specification of public APIs (`sort()` and `sortKeyPrefixArray()`).

core/src/main/java/org/apache/spark/util/collection/unsafe/sort/RadixSort.java
```diff
@@ -40,28 +40,28 @@
...
     assert numRecords * 2 <= array.size();
-    int inIndex = 0;
-    int outIndex = numRecords;
+    long inIndex = 0;
+    long outIndex = numRecords;
     if (numRecords > 0) {
       long[][] counts = getCounts(array, numRecords, startByteIndex, endByteIndex);
-      for (int i = startByteIndex; i <= endByteIndex; i++) {
-        if (counts[i] != null) {
+      for (long i = startByteIndex; i <= endByteIndex; i++) {
```
ericl: This should probably not be a long, since it ranges from 0..7

core/src/main/java/org/apache/spark/util/collection/unsafe/sort/RadixSort.java
```diff
@@ -40,28 +40,28 @@
    *         of always copying the data back to position zero for efficiency.
    */
   public static int sort(
-      LongArray array, int numRecords, int startByteIndex, int endByteIndex,
+      LongArray array, long numRecords, long startByteIndex, long endByteIndex,
```
ericl: I think we should keep the byte indices as ints, since they are in 0..7 and we assert these below.

ericl (commented):
How about keeping the very small numbers (0..7, 0..255) as ints, and everything else as a long.

core/src/main/java/org/apache/spark/util/collection/unsafe/sort/RadixSort.java
kiszk: Sure, I imagined that all of variables should be changed to long anyway.

ericl:
Looks good

sql/core/src/test/scala/org/apache/spark/sql/execution/benchmark/SortBenchmark.scala
```diff
@@ -80,7 +80,7 @@ class SortBenchmark extends BenchmarkBase {
       }
       val buf = new LongArray(MemoryBlock.fromLongArray(array))
       timer.startTiming()
-      RadixSort.sort(buf, size, 0, 7, false, false)
+      RadixSort.sort(buf, size.toLong, 0, 7, false, false)
```
rxin: why do we need these toLongs? Doesn't scala do the cast implicitly?

core/src/test/scala/org/apache/spark/util/collection/unsafe/sort/RadixSortSuite.scala
```diff
@@ -73,22 +73,22 @@ class RadixSortSuite extends SparkFunSuite with Logging {
...
+    val extended = ref ++ Array.fill[Long]((size * 2).toInt)(0)
     (new LongArray(MemoryBlock.fromLongArray(ref)),
      new LongArray(MemoryBlock.fromLongArray(extended)))
   }
 
-  private def collectToArray(array: LongArray, offset: Int, length: Int): Array[Long] = {
+  private def collectToArray(array: LongArray, offset: Int, length: Long): Array[Long] = {
     var i = 0
-    val out = new Array[Long](length)
+    val out = new Array[Long](length.toInt)
```
rxin: I'd add a checkedCast function similar to Guava's: https://google.github.io/guava/releases/19.0/api/docs/com/google/common/primitives/Ints.html#checkedCast(long)

and use those for all the toInt downcasts. Otherwise it'd be annoying to get mysterious negative integer exceptions.

core/src/main/java/org/apache/spark/shuffle/sort/ShuffleInMemorySorter.java
```diff
@@ -176,7 +176,7 @@ public ShuffleSorterIterator getSortedIterator() {
     int offset = 0;
     if (useRadixSort) {
       offset = RadixSort.sort(
-        array, pos,
+        array, (long)pos,
```
rxin: again doesn't java do the upcast implicitly?

core/src/test/scala/org/apache/spark/util/collection/unsafe/sort/RadixSortSuite.scala
kiszk: Great! done in these files.

rxin: Did you forget to push?

sql/core/src/test/scala/org/apache/spark/sql/execution/benchmark/SortBenchmark.scala
kiszk: As I wrote before, I wanted to emphasis the difference of types between caller and callee.
It is OK to reply on implicit cast, done.

core/src/main/java/org/apache/spark/shuffle/sort/ShuffleInMemorySorter.java
kiszk: It is OK to reply on implicit cast, done.

core/src/test/scala/org/apache/spark/util/collection/unsafe/sort/RadixSortSuite.scala
```diff
@@ -30,7 +32,7 @@ import org.apache.spark.util.collection.Sorter
 import org.apache.spark.util.random.XORShiftRandom
 
 class RadixSortSuite extends SparkFunSuite with Logging {
-  private val N = 10000  // scale this down for more readable results
+  private val N = 10000.toLong  // scale this down for more readable results
```
srowen: Write `10000L`

core/src/main/java/org/apache/spark/util/collection/unsafe/sort/RadixSort.java
```diff
@@ -40,28 +42,28 @@
...
             desc, signed && i == endByteIndex);
-          int tmp = inIndex;
+          long tmp = inIndex;
           inIndex = outIndex;
           outIndex = tmp;
         }
       }
     }
-    return inIndex;
+    return Ints.checkedCast(inIndex);
```
srowen: Rather than pull in a library method, how about `require` which lets you provide an actual error message?

srowen (changes requested):

core/src/test/scala/org/apache/spark/util/collection/unsafe/sort/RadixSortSuite.scala
kiszk: Sure, I will do that

core/src/main/java/org/apache/spark/util/collection/unsafe/sort/RadixSort.java
kiszk: cc @rxin

rxin: It might make sense to add a very simple if check in the beginning of this function. I will do it when I merge this.

rxin:
Merging in master/branch-2.1

PR #15908, opened 2016-11-16
[SPARK-18459][SPARK-18460][StructuredStreaming] Rename triggerId to batchId and add triggerDetails to json in StreamingQueryStatus (for branch-2.0)

tdas:
This is a fix for branch-2.0 for the earlier PR #15895 

SPARK-18459: triggerId seems like a number that should be increasing with each trigger, whether or not there is data in it. However, actually, triggerId increases only where there is a batch of data in a trigger. So its better to rename it to batchId.

SPARK-18460: triggerDetails was missing from json representation. Fixed it.
## How was this patch tested?

Updated tests

### DIFF
--- a/sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamingQueryStatusSuite.scala
+++ b/sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamingQueryStatusSuite.scala
@@ -48,12 +48,12 @@ class StreamingQueryStatusSuite extends SparkFunSuite {
         |    Processing rate 23.5 rows/sec
         |    Latency: 345.0 ms
         |    Trigger details:
+        |        batchId: 5
         |        isDataPresentInTrigger: true
         |        isTriggerActive: true
         |        latency.getBatch.total: 20
         |        latency.getOffset.total: 10
         |        numRows.input.total: 100
-        |        triggerId: 5
         |    Source statuses [1 source]:
         |        Source 1 - MySource1
         |            Available offset: #0
@@ -72,7 +72,11 @@ class StreamingQueryStatusSuite extends SparkFunSuite {
   test("json") {
     assert(StreamingQueryStatus.testStatus.json ===
       """
-        |{"sourceStatuses":[{"description":"MySource1","offsetDesc":"#0","inputRate":15.5,
+        |{"name":"query","id":1,"timestamp":123,"inputRate":15.5,"processingRate":23.5,
+        |"latency":345.0,"triggerDetails":{"latency.getBatch.total":"20",
+        |"numRows.input.total":"100","isTriggerActive":"true","batchId":"5",
+        |"latency.getOffset.total":"10","isDataPresentInTrigger":"true"},
+        |"sourceStatuses":[{"description":"MySource1","offsetDesc":"#0","inputRate":15.5,
         |"processingRate":23.5,"triggerDetails":{"numRows.input.source":"100",
         |"latency.getOffset.source":"10","latency.getBatch.source":"20"}}],
         |"sinkStatus":{"description":"MySink","offsetDesc":"[#1, -]"}}
@@ -84,6 +88,20 @@ class StreamingQueryStatusSuite extends SparkFunSuite {
       StreamingQueryStatus.testStatus.prettyJson ===
         """
           |{
+          |  "name" : "query",
+          |  "id" : 1,
+          |  "timestamp" : 123,
+          |  "inputRate" : 15.5,
+          |  "processingRate" : 23.5,
+          |  "latency" : 345.0,
+          |  "triggerDetails" : {
+          |    "latency.getBatch.total" : "20",
+          |    "numRows.input.total" : "100",
+          |    "isTriggerActive" : "true",
+          |    "batchId" : "5",
+          |    "latency.getOffset.total" : "10",
+          |    "isDataPresentInTrigger" : "true"
+          |  },
           |  "sourceStatuses" : [ {
           |    "description" : "MySource1",
           |    "offsetDesc" : "#0",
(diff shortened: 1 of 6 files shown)


zsxwing:
LGTM pending tests.

sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamingQueryStatusSuite.scala
```diff
@@ -84,6 +88,20 @@ class StreamingQueryStatusSuite extends SparkFunSuite {
       StreamingQueryStatus.testStatus.prettyJson ===
         """
           |{
+          |  "name" : "query",
+          |  "id" : 1,
```
marmbrus: I wonder if we should make this a `String`, both here and in the type-safe API.  The reasoning is, its already auto generated and I think its likely we will want to turn this into a UUID (maybe we should even just do that).  Thoughts?

sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamingQueryStatusSuite.scala
```diff
@@ -84,6 +88,20 @@ class StreamingQueryStatusSuite extends SparkFunSuite {
...
+          |    "latency.getBatch.total" : "20",
+          |    "numRows.input.total" : "100",
+          |    "isTriggerActive" : "true",
+          |    "batchId" : "5",
+          |    "latency.getOffset.total" : "10",
+          |    "isDataPresentInTrigger" : "true"
+          |  },
           |  "sourceStatuses" : [ {
           |    "description" : "MySource1",
           |    "offsetDesc" : "#0",
```
marmbrus: Are these not JSON now?

sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamingQueryStatusSuite.scala
```diff
@@ -84,6 +88,20 @@ class StreamingQueryStatusSuite extends SparkFunSuite {
...
+          |  "inputRate" : 15.5,
+          |  "processingRate" : 23.5,
+          |  "latency" : 345.0,
+          |  "triggerDetails" : {
+          |    "latency.getBatch.total" : "20",
+          |    "numRows.input.total" : "100",
+          |    "isTriggerActive" : "true",
+          |    "batchId" : "5",
+          |    "latency.getOffset.total" : "10",
+          |    "isDataPresentInTrigger" : "true"
```
marmbrus: Is this redundant with `numRows.input.total`?

sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamingQueryStatusSuite.scala
```diff
@@ -72,7 +72,11 @@ class StreamingQueryStatusSuite extends SparkFunSuite {
...
     assert(StreamingQueryStatus.testStatus.json ===
       """
-        |{"sourceStatuses":[{"description":"MySource1","offsetDesc":"#0","inputRate":15.5,
+        |{"name":"query","id":1,"timestamp":123,"inputRate":15.5,"processingRate":23.5,
+        |"latency":345.0,"triggerDetails":{"latency.getBatch.total":"20",
+        |"numRows.input.total":"100","isTriggerActive":"true","batchId":"5",
+        |"latency.getOffset.total":"10","isDataPresentInTrigger":"true"},
+        |"sourceStatuses":[{"description":"MySource1","offsetDesc":"#0","inputRate":15.5,
         |"processingRate":23.5,"triggerDetails":{"numRows.input.source":"100",
         |"latency.getOffset.source":"10","latency.getBatch.source":"20"}}],
```
marmbrus: I would consider removing `source`.  It is already nested inside of the list of sources.

sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamingQueryStatusSuite.scala
```diff
@@ -84,6 +88,20 @@ class StreamingQueryStatusSuite extends SparkFunSuite {
...
           |{
+          |  "name" : "query",
+          |  "id" : 1,
+          |  "timestamp" : 123,
+          |  "inputRate" : 15.5,
+          |  "processingRate" : 23.5,
+          |  "latency" : 345.0,
+          |  "triggerDetails" : {
+          |    "latency.getBatch.total" : "20",
+          |    "numRows.input.total" : "100",
```
marmbrus: Similar to removing `source` I'd consider removing `total` since this all of these metrics relate to the entire query.

sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamingQueryStatusSuite.scala
```diff
@@ -84,6 +88,20 @@ class StreamingQueryStatusSuite extends SparkFunSuite {
...
+          |    "isTriggerActive" : "true",
+          |    "batchId" : "5",
+          |    "latency.getOffset.total" : "10",
```
marmbrus: Where are the timestamps?  I only see latency here.

marmbrus (commented):
Thanks for fixing this!  A few other comments after looking at these metrics more closely.

marmbrus:
Oh sorry... I'm realizing I'm commenting on a PR for branch-2.0.  We don't have to address these comments here, but we should make sure we are happy with all the naming before 2.1 is released.

tdas:
@marmbrus I will make another PR to address those. Hopefully there wont be any more conflicts between master/2.1 and 2.0

zsxwing:
Sounds good. Merging to branch 2.0.

zsxwing:
@tdas Merged. Could you close this one, please?

PR #15910, opened 2016-11-17
[SPARK-18476][SPARKR][ML]:SparkR Logistic Regression should should support output original label.

wangmiao1981:

Similar to SPARK-18401, as a classification algorithm, logistic regression should support output original label instead of supporting index label.

In this PR, original label output is supported and test cases are modified and added. Document is also modified.

## How was this patch tested?

Unit tests.

### DIFF
--- a/R/pkg/R/mllib.R
+++ b/R/pkg/R/mllib.R
@@ -712,7 +712,6 @@ setMethod("predict", signature(object = "KMeansModel"),
 #'                        of L1 and L2. Default is 0.0 which is an L2 penalty.
 #' @param maxIter maximum iteration number.
 #' @param tol convergence tolerance of iterations.
-#' @param fitIntercept whether to fit an intercept term.
 #' @param family the name of family which is a description of the label distribution to be used in the model.
 #'               Supported options:
 #'                 \itemize{
@@ -783,7 +782,7 @@ setMethod("predict", signature(object = "KMeansModel"),
 #' @note spark.logit since 2.1.0
 setMethod("spark.logit", signature(data = "SparkDataFrame", formula = "formula"),
           function(data, formula, regParam = 0.0, elasticNetParam = 0.0, maxIter = 100,
-                   tol = 1E-6, fitIntercept = TRUE, family = "auto", standardization = TRUE,
+                   tol = 1E-6, family = "auto", standardization = TRUE,
                    thresholds = 0.5, weightCol = NULL, aggregationDepth = 2,
                    probabilityCol = "probability") {
             formula <- paste(deparse(formula), collapse = "")
@@ -795,10 +794,10 @@ setMethod("spark.logit", signature(data = "SparkDataFrame", formula = "formula")
             jobj <- callJStatic("org.apache.spark.ml.r.LogisticRegressionWrapper", "fit",
                                 data@sdf, formula, as.numeric(regParam),
                                 as.numeric(elasticNetParam), as.integer(maxIter),
-                                as.numeric(tol), as.logical(fitIntercept),
-                                as.character(family), as.logical(standardization),
-                                as.array(thresholds), as.character(weightCol),
-                                as.integer(aggregationDepth), as.character(probabilityCol))
+                                as.numeric(tol), as.character(family),
+                                as.logical(standardization), as.array(thresholds),
+                                as.character(weightCol), as.integer(aggregationDepth),
+                                as.character(probabilityCol))
             new("LogisticRegressionModel", jobj = jobj)
           })
 
--- a/R/pkg/R/mllib.R
+++ b/R/pkg/R/mllib.R
@@ -748,10 +748,10 @@ setMethod("predict", signature(object = "KMeansModel"),
 #' sparkR.session()
 #' # binary logistic regression
 #' label <- c(0.0, 0.0, 0.0, 1.0, 1.0)
-#' feature <- c(1.1419053, 0.9194079, -0.9498666, -1.1069903, 0.2809776)
-#' binary_data <- as.data.frame(cbind(label, feature))
+#' features <- c(1.1419053, 0.9194079, -0.9498666, -1.1069903, 0.2809776)
+#' binary_data <- as.data.frame(cbind(label, features))
 #' binary_df <- createDataFrame(binary_data)
-#' blr_model <- spark.logit(binary_df, label ~ feature, thresholds = 1.0)
+#' blr_model <- spark.logit(binary_df, label ~ features, thresholds = 1.0)
 #' blr_predict <- collect(select(predict(blr_model, binary_df), "prediction"))
 #'
 #' # summary of binary logistic regression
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/apache/spark/SparkContext.scala
```diff
@@ -19,7 +19,7 @@ package org.apache.spark
 
 import java.io._
 import java.lang.reflect.Constructor
-import java.net.{MalformedURLException, URI}
```
wangmiao1981: Removed unused import in this PR, because this one line change is not encouraged as a separate PR.

shivaram:
cc @yanboliang

wangmiao1981:
retest this please.

wangmiao1981:
the failure occurs in kafka-streaming. 

retest this please.

wangmiao1981:
retest this please

yanboliang:
Jenkins, test this please

R/pkg/inst/tests/testthat/test_mllib.R
```diff
@@ -637,31 +637,31 @@ test_that("spark.isotonicRegression", {
...
 
   # test summary of binary logistic regression
   blr_summary <- summary(blr_model)
   blr_fmeasure <- collect(select(blr_summary$fMeasureByThreshold, "threshold", "F-Measure"))
-  expect_equal(blr_fmeasure$threshold, c(0.8221347, 0.7884005, 0.6674709, 0.3785437, 0.3434487),
-               tolerance = 1e-4)
-  expect_equal(blr_fmeasure$"F-Measure", c(0.5000000, 0.8000000, 0.6666667, 0.8571429, 0.7500000),
-               tolerance = 1e-4)
+  expect_equal(blr_fmeasure$threshold, c(0.6565513, 0.6214563, 0.3325291, 0.2115995, 0.1778653),
+  tolerance = 1e-4)
```
felixcheung: nit: would be great to align the tolerance parameter with indentation

R/pkg/inst/tests/testthat/test_mllib.R
```diff
@@ -674,6 +674,16 @@ test_that("spark.logit", {
...
 
+  # test prediction label as text
+  training <- suppressWarnings(createDataFrame(iris))
+  binomial_training <- training[training$Species %in% c("versicolor", "virginica"), ]
+  binomial_model <- spark.logit(binomial_training, Species ~ Sepal_Length + Sepal_Width)
+  prediction <- predict(binomial_model, binomial_training)
+  expect_equal(typeof(take(select(prediction, "prediction"), 1)$prediction), "character")
+  expected <- c("virginica", "virginica", "virginica", "versicolor", "virginica",
+                "versicolor", "virginica", "versicolor", "virginica", "versicolor")
+  expect_equal(as.list(take(select(prediction, "prediction"), 10))[[1]], expected)
```
felixcheung: how reliable is this test? the order of rows is not guaranteed unless it is enforced by a sort or something, right?

mllib/src/main/scala/org/apache/spark/ml/r/LogisticRegressionWrapper.scala
```diff
@@ -84,14 +93,12 @@ private[r] object LogisticRegressionWrapper
...
-    // get feature names from output schema
-    val schema = rFormulaModel.transform(data).schema
-    val featureAttrs = AttributeGroup.fromStructField(schema(rFormulaModel.getFeaturesCol))
-      .attributes.get
-    val features = featureAttrs.map(_.name.get)
+    // get labels and feature names from output schema
+    val (features, labels) = getFeaturesAndLabels(rFormulaModel, data)
 
     // assemble and fit the pipeline
     val logisticRegression = new LogisticRegression()
```
yanboliang: off-topic, but I think it's a bug. We should not allow users pass ```fitIntercept``` to control whether to fit intercept, this should be handled by formula. For example, if users specify formula ```y ~ a + b + c - 1```, then the model should be fitted w/o intercept. Could you please fix this bug as well? Thanks.

R/pkg/R/mllib.R
```diff
@@ -736,7 +736,7 @@ setMethod("predict", signature(object = "KMeansModel"),
 #' \dontrun{
 #' sparkR.session()
 #' # binary logistic regression
-#' label <- c(1.0, 1.0, 1.0, 0.0, 0.0)
+#' label <- c(0.0, 0.0, 0.0, 1.0, 1.0)
 #' feature <- c(1.1419053, 0.9194079, -0.9498666, -1.1069903, 0.2809776)
```
yanboliang: Usually we name it as ```features```.

R/pkg/inst/tests/testthat/test_mllib.R
yanboliang: Theoretically, the order is not guaranteed. However, we did similar work from the first test case of mllib.R, but never had a problem until now. I'd like to enforce the tests here and other places, but may be in a separate work should be better since it involves lots of other tests?

felixcheung: sounds good, separate JIRA then. If tests haven't been failing perhaps it is not huge problem

wangmiao1981:
I am on travel now. I will address the comments asap. Thanks!

R/pkg/inst/tests/testthat/test_mllib.R
wangmiao1981: I will try to create follow-up jira for this.

mllib/src/main/scala/org/apache/spark/ml/r/LogisticRegressionWrapper.scala
wangmiao1981: OK. Fix it in this PR?

wangmiao1981:
@yanboliang @felixcheung I am back from vacation and made changes according to your comments.

Thanks!

yanboliang (commented):
LGTM

mllib/src/main/scala/org/apache/spark/ml/r/LogisticRegressionWrapper.scala
```diff
@@ -105,16 +113,23 @@ private[r] object LogisticRegressionWrapper
       .setWeightCol(weightCol)
       .setAggregationDepth(aggregationDepth)
       .setFeaturesCol(rFormula.getFeaturesCol)
+      .setLabelCol(rFormula.getLabelCol)
       .setProbabilityCol(probability)
+      .setPredictionCol(PREDICTED_LABEL_INDEX_COL)
```
felixcheung: shouldn't fitIntercept be set on LogisticRegression?

felixcheung:
LGTM. we need to get this in branch-2.1 because of the signature change

yanboliang:
Merged into master and branch-2.1. Thanks.

yanboliang:
I found the ```summary``` of ```spark.logit``` return incorrect result when reviewing this PR. Actually it should return ```coefficients``` rather than binary logistic regression summary that R users may not be interested. Meanwhile, the binary logistic regression summary will ignore ```weightCol``` and treats all instance weights as ```1.0```, we are discussing to resolve this issue at Scala side. I will send a PR to correct ```summary``` return value for ```spark.logit``` and hope it can catch 2.1. Thanks.

wangmiao1981:
The summary returns the same as scala side summary, including roc, areaUnderROC, pr, fMeasureByThreshold etc. I think we can add `coefficients` as additional item.

PR #15911, opened 2016-11-17
[SPARK-18477][SS]Enable interrupts for HDFS in HDFSMetadataLog

zsxwing:

HDFS `write` may just hang until timeout if some network error happens. It's better to enable interrupts to allow stopping the query fast on HDFS.

This PR just changes the logic to only disable interrupts for local file system, as HADOOP-10622 only happens for local file system.

## How was this patch tested?

Jenkins


tdas:
LGTM. Merging to master, 2.1, and 2.0

PR #15913, opened 2016-11-17
[SPARK-18481][ML] ML 2.1 QA: Remove deprecated methods for ML

yanboliang:
Remove deprecated methods for ML.

## How was this patch tested?
Existing tests.


yanboliang:
Jenkins, test this please

yanboliang:
Jenkins, test this please

yanboliang:
Jenkins, test this please.

yanboliang:
cc @jkbradley @sethah

mllib/src/main/scala/org/apache/spark/ml/tree/treeModels.scala
```diff
@@ -95,11 +95,6 @@ private[ml] trait TreeEnsembleModel[M <: DecisionTreeModel] {
   /** Trees in this ensemble. Warning: These have null parent Estimators. */
   def trees: Array[M]
 
-  /**
-   * Number of trees in ensemble
-   */
-  val getNumTrees: Int = trees.length
```
holdenk: Is there a reason this was moved? It seems unrelated to the cleanup described in the title.

python/pyspark/ml/util.py
```diff
@@ -78,7 +78,14 @@ def overwrite(self):
         raise NotImplementedError("MLWriter is not yet implemented for type: %s" % type(self))
 
     def context(self, sqlContext):
-        """Sets the SQL context to use for saving."""
+        """
+        Sets the SQL context to use for saving.
+        .. note:: Deprecated in 2.1, use session instead.
```
sethah: If we remove it from Scala in 2.2, then we have to remove it here in 2.2 right? I guess we should note it.

mllib/src/main/scala/org/apache/spark/ml/feature/ChiSqSelector.scala
```diff
@@ -216,13 +216,6 @@ final class ChiSqSelectorModel private[ml] (
   @Since("1.6.0")
   def setOutputCol(value: String): this.type = set(outputCol, value)
 
-  /**
-   * @group setParam
-   */
-  @Since("1.6.0")
-  @deprecated("labelCol is not used by ChiSqSelectorModel.", "2.0.0")
-  def setLabelCol(value: String): this.type = set(labelCol, value)
```
sethah: We need to remove it from Python as well.

sethah (commented):
Looks pretty good overall, a few small comments.

mllib/src/main/scala/org/apache/spark/ml/tree/treeModels.scala
yanboliang: @sethah You explained this problem very clearly and totally right. Thanks.

mllib/src/main/scala/org/apache/spark/ml/tree/treeParams.scala
```diff
@@ -443,12 +425,11 @@ private[ml] trait GBTParams extends TreeEnsembleParams with HasMaxIter with HasS
    * (default = 0.1)
    * @group setParam
    */
-  def setStepSize(value: Double): this.type = set(stepSize, value)
-
-  override def validateParams(): Unit = {
+  def setStepSize(value: Double): this.type = {
     require(ParamValidators.inRange(0, 1, lowerInclusive = false, upperInclusive = true)(
```
yanboliang: The original ```validateParams()``` was only used to check for interactions between parameters. Parameter value checks which do not depend on other parameters are handled by ````Param.validate()``` at the definition, i.e.
```
final val k = new IntParam(this, "k", "The number of clusters to create. " +
    "Must be > 1.", ParamValidators.gt(1))
```
However, ```stepSize``` was a trait and inherited by lots of sub-classes which have different constraints for this parameter, so I add the check at setter.

mllib/src/main/scala/org/apache/spark/ml/tree/treeParams.scala
```diff
@@ -317,8 +317,28 @@ private[ml] trait TreeEnsembleParams extends DecisionTreeParams {
...
+private[ml] trait RandomForestParams extends TreeEnsembleParams {
+
+  /**
+   * Number of trees to train (>= 1).
+   * If 1, then no bootstrapping is used.  If > 1, then bootstrapping is done.
+   * TODO: Change to always do bootstrapping (simpler).  SPARK-7130
+   * (default = 20)
+   * @group param
+   */
+  final val numTrees: IntParam = new IntParam(this, "numTrees", "Number of trees to train (>= 1)",
```
yanboliang: Good suggestion, added.

mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala
```diff
@@ -176,8 +176,12 @@ private[classification] trait LogisticRegressionParams extends ProbabilisticClas
     }
   }
 
-  override def validateParams(): Unit = {
+  override protected def validateAndTransformSchema(
```
yanboliang: I think it's a bug that ```validateParams ``` was never used. It should validate params interaction before fitting(if necessary), this is why we deprecate ```validateParams ``` and move what it does to ```transformSchema```. We do not have corresponding test cases before, so no test was broken when we deprecated ```validateParams```. I added test cases in this PR.

mllib/src/main/scala/org/apache/spark/ml/regression/GBTRegressor.scala
```diff
@@ -183,6 +183,11 @@ class GBTRegressionModel private[ml](
   @Since("1.4.0")
   override def trees: Array[DecisionTreeRegressionModel] = _trees
 
+  /**
+   * Number of trees in ensemble
+   */
+  val getNumTrees: Int = trees.length
```
yanboliang: Added.

mllib/src/main/scala/org/apache/spark/ml/feature/ChiSqSelector.scala
yanboliang: All python ML models don't support setter/getter params currently, so no need to remove it.

python/pyspark/ml/util.py
yanboliang: Added.

mllib/src/main/scala/org/apache/spark/ml/tree/treeParams.scala
holdenk: That makes more sense :) +1

yanboliang:
Jenkins, test this please.

mllib/src/main/scala/org/apache/spark/ml/param/params.scala
```diff
@@ -547,21 +547,6 @@ trait Params extends Identifiable with Serializable {
   }
 
   /**
-   * Validates parameter values stored internally.
-   * Raise an exception if any parameter value is invalid.
-   *
-   * This only needs to check for interactions between parameters.
```
jkbradley: Could you please add the relevant text from here to transformSchema?

mllib/src/main/scala/org/apache/spark/ml/tree/treeParams.scala
jkbradley: This validation should really be in the Param itself.  It's not since it's a sharedParam, but perhaps we should just copy the shared Param code here.  The problem with putting validation in the setter is that Params can get set in other ways too (such as ```gbt.set(stepSize, value)```

mllib/src/main/scala/org/apache/spark/ml/tree/treeParams.scala
```diff
@@ -443,12 +429,11 @@ private[ml] trait GBTParams extends TreeEnsembleParams with HasMaxIter with HasS
...
    */
-  def setStepSize(value: Double): this.type = set(stepSize, value)
-
-  override def validateParams(): Unit = {
+  def setStepSize(value: Double): this.type = {
     require(ParamValidators.inRange(0, 1, lowerInclusive = false, upperInclusive = true)(
-      getStepSize), "GBT parameter stepSize should be in interval (0, 1], " +
-      s"but it given invalid value $getStepSize.")
+      value), "GBT parameter stepSize should be in interval (0, 1], " +
+      s"but it given invalid value $value.")
```
jkbradley: "it given invalid value" -> "it was given an invalid value"

mllib/src/main/scala/org/apache/spark/ml/tree/treeParams.scala
```diff
@@ -317,8 +317,32 @@ private[ml] trait TreeEnsembleParams extends DecisionTreeParams {
...
+   * are a bit different.
+   * @group param
+   */
+  final val numTrees: IntParam = new IntParam(this, "numTrees", "Number of trees to train (>= 1)",
+    ParamValidators.gtEq(1))
+
+  setDefault(numTrees -> 20)
+
+  /** @group setParam */
+  def setNumTrees(value: Int): this.type = set(numTrees, value)
```
jkbradley: Are these setter methods in traits Java-compatible?

jkbradley (commented):
Thanks for the PR!  Adding a couple of comments

jkbradley:
One more thing: This breaks several public APIs, although in mostly minor ways and with good reasons.  Could you please update the JIRA to list the public API changes?

mllib/src/main/scala/org/apache/spark/ml/param/params.scala
yanboliang: Done.

mllib/src/main/scala/org/apache/spark/ml/tree/treeParams.scala
yanboliang: Yeah, we already have ```setNumTrees``` which calls ```super.setNumTrees``` in ```RandomForestClassifier``` and ```RandomForestRegressor```.

mllib/src/main/scala/org/apache/spark/ml/tree/treeParams.scala
```diff
@@ -432,24 +418,26 @@ private[ml] trait GBTParams extends TreeEnsembleParams with HasMaxIter with HasS
...
    */
+  final val stepSize: DoubleParam = new DoubleParam(this, "stepSize", "Step size " +
+    "(a.k.a. learning rate) in interval (0, 1] for shrinking the contribution of each estimator.",
+    ParamValidators.inRange(0, 1, lowerInclusive = false, upperInclusive = true))
+
+  /** @group getParam */
+  final def getStepSize: Double = $(stepSize)
+
+  /** @group setParam */
   def setStepSize(value: Double): this.type = set(stepSize, value)
```
jkbradley: I asked about this being Java-friendly because Param setter methods in traits used to have the wrong type in Java.  I wonder if this is no longer true.  Still, it might make sense to remove the setter method from the trait since it does not make sense to have it in the Model classes.  We could put the setter method in each subclass and then deprecate the method in the Model classes.

This issue also shows up in MimaExcludes.  If setFeatureSubsetStrategy were put in the concrete classes from the outset, then you would not need to include it in MimaExcludes now.

I'm Ok with doing the change now or in a follow-up PR.

I believe I was the one who incorrectly put the setters in the traits...

yanboliang:
Jenkins, test this please.

mllib/src/main/scala/org/apache/spark/ml/tree/treeParams.scala
yanboliang: Yeah, I understand what you mean. If we would like to correct the setter methods in traits, we  involves changes to lots of traits which include ```DecisionTreeParams```, ```TreeClassifierParams```, ```TreeRegressorParams```, ```RandomForestParams```, ```GBTParams```, etc. So i will merge this firstly after it pass Jenkins and address this issue in a separate follow-up PR. Thanks.

yanboliang:
Merged into master and branch-2.1. Thanks for all reviewing.

mllib/src/main/scala/org/apache/spark/ml/tree/treeParams.scala
yanboliang: @jkbradley I have sent #16017 to fix this issue, please feel free to comment that. Thanks.

PR #15915, opened 2016-11-17
[SPARK-18485][CORE] Underlying integer overflow when create ChunkedByteBufferOutputStream in MemoryStore

uncleGen:

There is an underlying integer overflow when create ChunkedByteBufferOutputStream in MemoryStore. This PR provide a check before cast.  

## How was this patch tested?

add new unit test


core/src/main/scala/org/apache/spark/storage/memory/MemoryStore.scala
```diff
@@ -331,7 +332,8 @@ private[spark] class MemoryStore(
     var unrollMemoryUsedByThisBlock = 0L
     // Underlying buffer for unrolling the block
     val redirectableStream = new RedirectableOutputStream
-    val bbos = new ChunkedByteBufferOutputStream(initialMemoryThreshold.toInt, allocator)
+    val bbos =
```
srowen: Rather than import a whole other class for this, you probably want to explicitly check and throw an exception.

But, you're basically saying that this threshold can't be >= 4GB. Should that not just be checked earlier?

uncleGen: make sense, thanks!

uncleGen: make sense, thanks

core/src/main/scala/org/apache/spark/storage/memory/MemoryStore.scala
```diff
@@ -331,7 +331,12 @@ private[spark] class MemoryStore(
     var unrollMemoryUsedByThisBlock = 0L
     // Underlying buffer for unrolling the block
     val redirectableStream = new RedirectableOutputStream
-    val bbos = new ChunkedByteBufferOutputStream(initialMemoryThreshold.toInt, allocator)
+    val chunkSize = if (initialMemoryThreshold >= Integer.MAX_VALUE) {
```
srowen: CC @joshrosen who I think wrote this part of the code for a check

uncleGen: Thanks for your feedback. Let us  listen to @joshrosen 's advice

JoshRosen:
To clarify, does this overflow only occur when `spark.storage.unrollMemoryThreshold` has been manually set to a huge value which doesn't fit into an int?

Setting the chunk size to `Integer.MAX_VALUE` probably isn't a good idea because it'll leave us OOM-prone. Instead, we should probably set some reasonable upper bound on the chunk size based on either the unsafe page size or something else.

srowen:
@JoshRosen sounds like you don't disagree with this change because anything more than  `Integer.MAX_VALUE` is going to cause an outright failure, but mean it could impose a significantly smaller limit. Should we do that here? you probably know better, is the right limit to impose the page size? or just punt on that and fix the corner case error only here?

uncleGen:
@JoshRosen I find some other usage of "chunkSize" in other code path, and we set it with **1024 * 1024 * 4**. Is there any reason？ Now, I use it as default value for "chunkSize"

uncleGen:
@JoshRosen The main cause for this issue is the mix-up of "chunkSize" and other "size"

uncleGen:
@SparkQA  retest again

uncleGen:
retest this please

JoshRosen:
Haven't forgotten about this; it's on my list to review post-Thanksgiving.

uncleGen:
Is there any update? @JoshRosen

srowen:
This introduces a whole new config property. I think the suggestion was to base it off of some existing property, like page size? @JoshRosen would be best positioned to comment on that. We can more narrowly fix the problem of exceeding Int.MaxValue as you did before if that is a non-trivial problem in real use.

uncleGen:
@srowen +1 to your comment.  

> I think the suggestion was to base it off of some existing property, like page size? 

Yes, introducing a new config property is a practice of lazy. As you said, we can reuse existing property. @JoshRosen can give some suggestion?

JoshRosen:
I'd use `SparkEnv.memoryManager.pageSizeBytes` (which is calculated at https://github.com/apache/spark/blob/bec0a9217b896a596c0a0919888da34589fc720d/core/src/main/scala/org/apache/spark/memory/MemoryManager.scala#L215).

srowen:
@uncleGen can you follow up on this so we can finish this out?

uncleGen:
@srowen Sorry for the delay, I will update it as soon as possible on the basis of comments.

core/src/main/scala/org/apache/spark/memory/MemoryManager.scala
```diff
@@ -223,8 +222,10 @@ private[spark] abstract class MemoryManager(
       case MemoryMode.OFF_HEAP => offHeapExecutionMemoryPool.poolSize
     }
     val size = ByteArrayMethods.nextPowerOf2(maxTungstenMemory / cores / safetyFactor)
-    val default = math.min(maxPageSize, math.max(minPageSize, size))
-    conf.getSizeAsBytes("spark.buffer.pageSize", default)
+    val maxPageSize = math.min(64L * minPageSize, math.max(minPageSize, size))
+    val userSetting = conf.getSizeAsBytes("spark.buffer.pageSize")
+    // In case of too large page size.
+    math.min(userSetting, maxPageSize)
   }
```
uncleGen: @JoshRosen The `SparkEnv.memoryManager.pageSizeBytes` returns `Long`, if we reuse it as chunk size, there is still underlying integer overflow, isn't it? Here, I restricted the upper limit of page size in case of too large.

core/src/main/scala/org/apache/spark/broadcast/TorrentBroadcast.scala
```diff
@@ -78,6 +80,7 @@ private[spark] class TorrentBroadcast[T: ClassTag](obj: T, id: Long)
     }
     // Note: use getSizeAsKb (not bytes) to maintain compatibility if no units are provided
     blockSize = conf.getSizeAsKb("spark.broadcast.blockSize", "4m").toInt * 1024
+    chunkSize = SparkEnv.get.memoryManager.pageSizeBytes.toInt
     checksumEnabled = conf.getBoolean("spark.broadcast.checksum", true)
```
uncleGen: @JoshRosen We use `SparkEnv.get.memoryManager.pageSizeBytes` as chunk size. As `SparkEnv.get.memoryManager.pageSizeBytes` returns `Long`, there is still underlying integer overflow issue, isn't it? Besides, users will never know the  low level details and the effect to chunk size when modify `pageSizeBytes`.

core/src/main/scala/org/apache/spark/broadcast/TorrentBroadcast.scala
```diff
@@ -78,6 +80,7 @@ private[spark] class TorrentBroadcast[T: ClassTag](obj: T, id: Long)
     }
     // Note: use getSizeAsKb (not bytes) to maintain compatibility if no units are provided
     blockSize = conf.getSizeAsKb("spark.broadcast.blockSize", "4m").toInt * 1024
```
viirya: `spark.broadcast.blockSize` has special meaning. I don't think we should replace it with `pageSizeBytes`.

core/src/main/scala/org/apache/spark/serializer/SerializerManager.scala
```diff
@@ -181,7 +181,7 @@ private[spark] class SerializerManager(
       blockId: BlockId,
       values: Iterator[_],
       classTag: ClassTag[_]): ChunkedByteBuffer = {
-    val bbos = new ChunkedByteBufferOutputStream(1024 * 1024 * 4, ByteBuffer.allocate)
+    val bbos = new ChunkedByteBufferOutputStream(chunkSize, ByteBuffer.allocate)
```
viirya: Previously it is a fixed size. When you use a configurable value converted from long to int, doesn't it increase the chance to overflow?

uncleGen:
Generally, there are three ways to fix this issue:

1. Check if `chunkSize` exceed `Int.MaxValue` to narrow this issue.
2. Provide a new config to set `chunkSize`.
3. Reuse existing config as `chunkSize`, like `pageSize`.

Basically, I don't like option 3, because users will never know the low level details and the effect to chunk size when modify `spark.buffer.pageSize`. Just like @viirya said:

> `spark.broadcast.blockSize` has special meaning. I don't think we should replace it with pageSizeBytes
 
Besides, `SparkEnv.get.memoryManager.pageSizeBytes` returns Long, there is still underlying integer overflow issue.

Introducing a new config is also a good idea. So, I'd like to do some check for exceeding `Int.MaxValue` to narrow this issue. 

Any suggestions will be appreciated.

@JoshRosen @srowen @viirya

viirya:
I'd like option 1 that to add some checks and throw error message for users.

core/src/main/scala/org/apache/spark/util/io/ChunkedByteBufferOutputStream.scala
```diff
@@ -30,9 +31,14 @@ import org.apache.spark.storage.StorageUtils
...
-    chunkSize: Int,
+    var chunkSize: Int,
     allocator: Int => ByteBuffer)
-  extends OutputStream {
+  extends OutputStream with Logging{
+
+  if (chunkSize < 0) {
+    logWarning(s"chunkSize should not be an negative value, replaced as 4MB default.")
+    chunkSize = 4 * 1024 * 1024
+  }
```
uncleGen: As `chunkSize` is passed from many code path, and there is underlying integer overflow when convert from `Long` to `Int`. As we do not have a better solution, introducing a protection check may be a tradeoff way in the first step.

@JoshRosen @srowen

core/src/test/scala/org/apache/spark/util/io/ChunkedByteBufferOutputStreamSuite.scala
```diff
@@ -119,4 +119,21 @@ class ChunkedByteBufferOutputStreamSuite extends SparkFunSuite {
...
+    assert(arrays.length === 3)
+    assert(arrays(0).length === 4 * 1024 * 1024)
+    assert(arrays(1).length === 4 * 1024 * 1024)
+    assert(arrays(2).length === 10 )
+
+    assert(arrays(0).toSeq === ref.slice(0, 4 * 1024 * 1024))
+    assert(arrays(1).toSeq === ref.slice(4 * 1024 * 1024, 8 * 1024 * 1024))
+    assert(arrays(2).toSeq === ref.slice(8 * 1024 * 1024, 8 * 1024 * 1024 + 10))
+  }
 }
```
uncleGen: Discovery starting.
Discovery completed in 42 seconds, 124 milliseconds.
Run starting. Expected test count is: 9
ChunkedByteBufferOutputStreamSuite:
- empty output
- write a single byte
- write a single near boundary
- write a single at boundary
- single chunk output
- single chunk output at boundary size
- multiple chunk output
- multiple chunk output at boundary size
- negative chunk size
Run completed in 42 seconds, 700 milliseconds.
Total number of tests run: 9
Suites: completed 2, aborted 0
Tests: succeeded 9, failed 0, canceled 0, ignored 0, pending 0
All tests passed.

core/src/main/scala/org/apache/spark/util/io/ChunkedByteBufferOutputStream.scala
uncleGen: For now though, there are not too may call sites. Checking Long value is not the best but feasible way to address this problem.

core/src/main/scala/org/apache/spark/storage/memory/MemoryStore.scala
```diff
@@ -331,7 +331,12 @@ private[spark] class MemoryStore(
     var unrollMemoryUsedByThisBlock = 0L
     // Underlying buffer for unrolling the block
     val redirectableStream = new RedirectableOutputStream
-    val bbos = new ChunkedByteBufferOutputStream(initialMemoryThreshold.toInt, allocator)
+    val chunkSize = if (initialMemoryThreshold > Int.MaxValue) {
+      4 * 1024 * 1024
```
viirya: Because `initialMemoryThreshold` is set to `unrollMemoryThreshold` which is configurable, if we want to use a default value in this case, we should log a warning message at least.

Another option is just throwing an exception. I am ok for either one.

uncleGen: I  prefer the first one, and the second one is too radical.

viirya:
Can you also add a test into MemoryStoreSuite?

uncleGen:
```
Discovery starting.
Discovery completed in 37 seconds, 984 milliseconds.
Run starting. Expected test count is: 10
MemoryStoreSuite:
- reserve/release unroll memory
- safely unroll blocks
- safely unroll blocks through putIteratorAsValues
- safely unroll blocks through putIteratorAsBytes
- set unrollMemoryThreshold a huge value larger than Int.MaxValue
- PartiallySerializedBlock.valuesIterator
- PartiallySerializedBlock.finishWritingToStream
- multiple unrolls by the same thread
- lazily create a big ByteBuffer to avoid OOM if it cannot be put into MemoryStore
- put a small ByteBuffer to MemoryStore
Run completed in 39 seconds, 91 milliseconds.
Total number of tests run: 10
Suites: completed 2, aborted 0
Tests: succeeded 10, failed 0, canceled 0, ignored 0, pending 0
All tests passed.
```

core/src/test/scala/org/apache/spark/storage/MemoryStoreSuite.scala
```diff
@@ -303,6 +303,34 @@ class MemoryStoreSuite
...
+      res
+    }
+
+    // Unroll with plenty of space. This should succeed and cache both blocks.
+    val result1 = putIteratorAsBytes("b1", smallIterator, ClassTag.Any)
+    val result2 = putIteratorAsBytes("b2", smallIterator, ClassTag.Any)
+    assert(memoryStore.contains("b1"))
+    assert(memoryStore.contains("b2"))
+    assert(result1.isRight) // unroll was successful
+    assert(result2.isRight)
```
viirya: Let's also check if the `currentUnrollMemoryForThisTask` is the default size 4MB.

viirya: nvm. It will be released once unrolling is successful.

viirya:
Current change looks good to me. cc @JoshRosen @srowen to check it again.

srowen:
It seems to do the job. It seems a little funny that if you specify a huge chunk size like 1GB, that's accepted, but if you specify ~2GB, it becomes 4MB. Is the logic that this probably happens when something else is left to a default value, so this is like a default value? 

it seems more natural to clamp to Int.MaxValue if we believe that the caller really means to specify a huge chunk size. But it sounds like the argument is that this can never be sensible. Is there a lower hard limit to enforce?

That is: is it too much to just cap this at 4MB?

uncleGen:
@srowen The truth is that `chunk size` is passed from different code path with different value. There is not a reasonable value when `chunk size` as a invalid value. Experiencedly, `chunk size` does not need a large value, although it can be. Besides, we can give it a special config, and restrict it a reasonable value, like not larger than 16MB or something else. But you said no to add new config.

srowen:
You could also cap it at 16MB here. Or 128MB or whatever. That is unrelated to introducing a new config. 

It seems funny that setting to allow Int.MaxValue but map anything else to something orders of magnitude smaller. Why not just cap at Int.MaxValue like in the original proposal? that's also less behavior change.

uncleGen:
OK, I will give an update as soon as possible.

core/src/main/scala/org/apache/spark/storage/memory/MemoryStore.scala
```diff
@@ -331,7 +331,15 @@ private[spark] class MemoryStore(
     var unrollMemoryUsedByThisBlock = 0L
     // Underlying buffer for unrolling the block
     val redirectableStream = new RedirectableOutputStream
-    val bbos = new ChunkedByteBufferOutputStream(initialMemoryThreshold.toInt, allocator)
+    val chunkSize = if (initialMemoryThreshold > Int.MaxValue) {
+      logWarning(s"Initial memory threshold of ${Utils.bytesToString(initialMemoryThreshold)} " +
+        s"is too large to be set as chunk size. It is safe to be cap at Int.MaxValue for chunk " +
```
srowen: How about `Chunk size has been capped to ${Utils.bytesToStringInt.MaxValue)}`instead, in the second sentence? that's more direct.

core/src/test/scala/org/apache/spark/storage/MemoryStoreSuite.scala
```diff
@@ -60,7 +60,7 @@ class MemoryStoreSuite
     SizeEstimator invokePrivate initialize()
   }
 
-  def makeMemoryStore(maxMem: Long): (MemoryStore, BlockInfoManager) = {
+  def makeMemoryStore(maxMem: Long, conf: SparkConf = conf): (MemoryStore, BlockInfoManager) = {
```
srowen: Is this change needed?

uncleGen: 👍

core/src/main/scala/org/apache/spark/storage/memory/MemoryStore.scala
uncleGen: 👍

core/src/test/scala/org/apache/spark/storage/MemoryStoreSuite.scala
viirya: I think it is for passing a SparkConf with modified `spark.storage.unrollMemoryThreshold`.

uncleGen: I changed this to pass a temp conf which set `spark.storage.unrollMemoryThreshold` a different value.

```
+    val tmpConf = conf.clone.set("spark.storage.unrollMemoryThreshold", s"${1L + Int.MaxValue}")
+    val (memoryStore, blockInfoManager) = makeMemoryStore(12000L + Int.MaxValue, tmpConf)
```

uncleGen:
@srowen Looks like it is not very convenient to test this case, with `java.lang.OutOfMemoryError: Requested array size exceeds VM limit` in jenkins. Do you think it is necessary to test Int.MaxValue chunk size or we can add a unit test in the follow-up PR if necessary?

core/src/main/scala/org/apache/spark/storage/memory/MemoryStore.scala
```diff
@@ -331,7 +331,15 @@ private[spark] class MemoryStore(
...
+        s"size instead.")
+      Int.MaxValue
+    } else {
+      initialMemoryThreshold.toInt
+    }
+    val bbos = new ChunkedByteBufferOutputStream(chunkSize, allocator)
```
viirya: Another option is to catch OOM exception here and then log a meaningful error message to hint users to lower `spark.storage.unrollMemoryThreshold` setting.

@srowen What do you think?

srowen: No, it's almost never a good idea, certainly not just to 'work around' bad config.
I think this is getting overly-complex. The net-net of a lot of discussion is that all we wanted to do was fix the actual bug, that a value > Int.MaxValue is not merely way too large, but read as the wrong number.

core/src/test/scala/org/apache/spark/storage/MemoryStoreSuite.scala
srowen: Oh, that's confusing because of the shadowing. The arg and member have the same name. Please name this arg differently if you do this.

core/src/main/scala/org/apache/spark/storage/memory/MemoryStore.scala
viirya: Yeah. As it causes OOM during jenkins test, we may not be able to test it like currently added test case if we cap it as `Int.MaxValue`.

viirya:
Because the actual change is far from complicated, if we can't have proper test case for it, can we skip the added test? @srowen @JoshRosen

srowen:
There's still one comment outstanding, but other than that, yes I think it's fine to make this change without another additional test, because it's hard to test and because the change is such a clear, narrow fix.

core/src/test/scala/org/apache/spark/storage/MemoryStoreSuite.scala
uncleGen: get it

uncleGen:
OK, we reach a consensus, I will give a update as soon as possible.

uncleGen:
There are some unrelated failures, retest it please.

uncleGen:
@srowen All tests passed,  take a look please.

srowen (approved):
OK by me. Will leave it open for a bit.

viirya:
LGTM

srowen:
Merged to master

PR #15874, opened 2016-11-14
[Spark-18408][ML] API Improvements for LSH

Yunni:
## What changes were proposed in this pull request?

(1) Change output schema to `Array of Vector` instead of `Vectors`
(2) Use `numHashTables` as the dimension of Array
(3) Rename `RandomProjection` to `BucketedRandomProjectionLSH`, `MinHash` to `MinHashLSH`
(4) Make `randUnitVectors/randCoefficients` private
(5) Make Multi-Probe NN Search and `hashDistance` private for future discussion

Saved for future PRs:
(1) AND-amplification and `numHashFunctions` as the dimension of Vector are saved for a future PR.
(2) `hashDistance` and MultiProbe NN Search needs more discussion. The current implementation is just a backward compatible one.

## How was this patch tested?
Related unit tests are modified to make sure the performance of LSH are ensured, and the outputs of the APIs meets expectation.

PR #15916, opened 2016-11-17
[SPARK-18487][SQL] Add completion listener to HashAggregate to avoid memory leak

viirya:

The methods such as Dataset.show and take use `Limit` (i.e., `CollectLimitExec`) which leverages SparkPlan.executeTake to efficiently collect required number of elements back to the driver.

However, under wholestage codege, we usually release resources after all elements are consumed (e.g., HashAggregate). In this case, we will not release the resources and cause memory leak with Dataset.show, for example.

An exception will be thrown in this case. This exception is thrown at `Executor` after it calls `taskMemoryManager.cleanUpAllAllocatedMemory` and finds there are memory not released after the task completion.

The exception looks like:

    [info] - SPARK-18487: Consume all elements for show/take to avoid memory leak *** FAILED *** (1 second, 73 milliseconds)
    [info]   org.apache.spark.SparkException: Job aborted due to stage failure: Task 0 in stage 179.0 failed 1 times, most recent failure: Lost task 0.0 in stage 179.0 (TID 501, localhost, executor driver): org.apache.spark.SparkException: Managed memory leak detected; size = 33816576 bytes, TID = 501
    [info]  at org.apache.spark.executor.Executor$TaskRunner.run(Executor.scala:295)
    [info]  at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
    [info]  at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
    [info]  at java.lang.Thread.run(Thread.java:745)
    [info] 
    [info] Driver stacktrace:
    [info]   at org.apache.spark.scheduler.DAGScheduler.org$apache$spark$scheduler$DAGScheduler$$failJobAndIndependentS tages(DAGScheduler.scala:1436)
    [info]   at org.apache.spark.scheduler.DAGScheduler$$anonfun$abortStage$1.apply(DAGScheduler.scala:1424)
    [info]   at org.apache.spark.scheduler.DAGScheduler$$anonfun$abortStage$1.apply(DAGScheduler.scala:1423)
    [info]   at scala.collection.mutable.ResizableArray$class.foreach(ResizableArray.scala:59)
    [info]   at scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:48)
    [info]   at org.apache.spark.scheduler.DAGScheduler.abortStage(DAGScheduler.scala:1423)
    [info]   at org.apache.spark.scheduler.DAGScheduler$$anonfun$handleTaskSetFailed$1.apply(DAGScheduler.scala:802)
    [info]   at org.apache.spark.scheduler.DAGScheduler$$anonfun$handleTaskSetFailed$1.apply(DAGScheduler.scala:802)
    [info]   at scala.Option.foreach(Option.scala:257)
    ...

This pr adds task completion listener to HashAggregate to avoid the memory leak.

## How was this patch tested?

Added test in DatasetSuite.

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/SparkPlan.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/SparkPlan.scala
@@ -234,12 +234,6 @@ abstract class SparkPlan extends QueryPlan[SparkPlan] with Logging with Serializ
         row.writeToStream(out, buffer)
         count += 1
       }
-      // If iterator has more elements, we should consume them all. Otherwise under wholestage
-      // codegen, as we release resources after consuming all elements (e.g., HashAggregate), it
-      // will cause problems such as memory leak.
-      while (iter.hasNext) {
-        iter.next()
-      }
       out.writeInt(-1)
       out.flush()
       out.close()
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/SparkPlan.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/SparkPlan.scala
@@ -234,6 +234,12 @@ abstract class SparkPlan extends QueryPlan[SparkPlan] with Logging with Serializ
         row.writeToStream(out, buffer)
         count += 1
       }
+      // If iterator has more elements, we should consume them all. Otherwise under wholestage
+      // codegen, as we release resources after consuming all elements (e.g., HashAggregate), it
+      // will cause problems such as memory leak.
+      while (iter.hasNext) {
+        iter.next()
+      }
       out.writeInt(-1)
       out.flush()
       out.close()
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/aggregate/HashAggregateExec.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/aggregate/HashAggregateExec.scala
@@ -295,6 +295,11 @@ case class HashAggregateExec(
   private var hashMapTerm: String = _
   private var sorterTerm: String = _
 
+  // Becasue Dataset.show/take methods will end of iteraton before reaching the end of all rows,
+  // we may not release resources then and cause memory leak. So we need to hold the reference
+  // of the hash map if it is created and release the resources after task completion.
+  private var hashMapToRelease: UnsafeFixedWidthAggregationMap = _
+
   /**
    * This is called by generated Java class, should be public.
    */
@@ -302,17 +307,23 @@ case class HashAggregateExec(
     // create initialized aggregate buffer
     val initExpr = declFunctions.flatMap(f => f.initialValues)
     val initialBuffer = UnsafeProjection.create(initExpr)(EmptyRow)
+    val context = TaskContext.get()
 
     // create hashMap
-    new UnsafeFixedWidthAggregationMap(
+    hashMapToRelease = new UnsafeFixedWidthAggregationMap(
       initialBuffer,
       bufferSchema,
       groupingKeySchema,
-      TaskContext.get().taskMemoryManager(),
+      context.taskMemoryManager(),
       1024 * 16, // initial capacity
-      TaskContext.get().taskMemoryManager().pageSizeBytes,
+      context.taskMemoryManager().pageSizeBytes,
       false // disable tracking of performance metrics
     )
+
+    // Release the resources of the hash map when the end of task.
+    context.addTaskCompletionListener(_ => hashMapToRelease.free())
+
+    hashMapToRelease
   }
 
   def getTaskMemoryManager(): TaskMemoryManager = {
--- a/sql/core/src/test/scala/org/apache/spark/sql/DatasetSuite.scala
+++ b/sql/core/src/test/scala/org/apache/spark/sql/DatasetSuite.scala
@@ -28,7 +28,7 @@ import org.apache.spark.sql.execution.streaming.MemoryStream
 import org.apache.spark.sql.functions._
 import org.apache.spark.sql.internal.SQLConf
 import org.apache.spark.sql.test.SharedSQLContext
-import org.apache.spark.sql.types.{IntegerType, StringType, StructField, StructType}
+import org.apache.spark.sql.types.{DataTypes, IntegerType, StringType, StructField, StructType}
 
 class DatasetSuite extends QueryTest with SharedSQLContext {
   import testImplicits._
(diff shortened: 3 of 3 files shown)


hvanhovell:
@viirya don't we use task completion listeners for this?

viirya:
@hvanhovell Thanks for the hint. I am trying it. However this listener would be in operators case by case, e.g., HashAggregate here. If there are any other operators needed to release resources after all rows iterated, we may need to add similar listeners too.

Originally I am little worrying the possible performance regression due to extra consuming of row. With task completion listener, this is not an issue.

rxin:
Did you actually see an issue with memory leak?

viirya:
@rxin Yeah, the added test case will report memory leak failure.

rxin:
What do you mean by "report"? A warning message was logged? If a warning message was logged, it is generated by the callback itself which just releases the memory.

viirya:
@rxin An exception will be thrown, not just a warning message. This exception is thrown at `Executor` after it calls `taskMemoryManager.cleanUpAllAllocatedMemory` and finds there are memory not released after the task completion.

The exception looks like:

    [info] - SPARK-18487: Consume all elements for show/take to avoid memory leak *** FAILED *** (1 second, 73 milliseconds)
    [info]   org.apache.spark.SparkException: Job aborted due to stage failure: Task 0 in stage 179.0 failed 1 times, most recent failure: Lost task 0.0 in stage 179.0 (TID 501, localhost, executor driver): org.apache.spark.SparkException: Managed memory leak detected; size = 33816576 bytes, TID = 501
    [info]  at org.apache.spark.executor.Executor$TaskRunner.run(Executor.scala:295)
    [info]  at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
    [info]  at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
    [info]  at java.lang.Thread.run(Thread.java:745)
    [info] 
    [info] Driver stacktrace:
    [info]   at org.apache.spark.scheduler.DAGScheduler.org$apache$spark$scheduler$DAGScheduler$$failJobAndIndependentS tages(DAGScheduler.scala:1436)
    [info]   at org.apache.spark.scheduler.DAGScheduler$$anonfun$abortStage$1.apply(DAGScheduler.scala:1424)
    [info]   at org.apache.spark.scheduler.DAGScheduler$$anonfun$abortStage$1.apply(DAGScheduler.scala:1423)
    [info]   at scala.collection.mutable.ResizableArray$class.foreach(ResizableArray.scala:59)
    [info]   at scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:48)
    [info]   at org.apache.spark.scheduler.DAGScheduler.abortStage(DAGScheduler.scala:1423)
    [info]   at org.apache.spark.scheduler.DAGScheduler$$anonfun$handleTaskSetFailed$1.apply(DAGScheduler.scala:802)
    [info]   at org.apache.spark.scheduler.DAGScheduler$$anonfun$handleTaskSetFailed$1.apply(DAGScheduler.scala:802)
    [info]   at scala.Option.foreach(Option.scala:257)
    ...

viirya:
Of course it is not actually a real memory leak because the memory is released at the end by calling `taskMemoryManager.cleanUpAllAllocatedMemory` in `Executor`. But with `spark.unsafe.exceptionOnMemoryLeak` as true by default, we will see the exception.

Or we just need to turn `spark.unsafe.exceptionOnMemoryLeak` to false by default?

rxin:
I'd set it maybe to false. You are just adding a completion listener, which is the same as taskMemoryManager.cleanUpAllAllocatedMemory anyway ...

viirya:
Yeah, I see. As I said in previous comment, the memory is released at the end anyway.

I would guess the default setting as true is to find potential memory leak during development. So turn it to false is a good idea?

This patch is coming from #15874 which hits the exception by @sethah.

Although `taskMemoryManager.cleanUpAllAllocatedMemory` can release memory for us, I think it is just a safety network. Operators should release memory themselves.

If you still think this is not necessary, I can close this.

viirya:
@hvanhovell @rxin What do you think? Please let me know if we need this or not. Thanks.

viirya:
Hello @rxin @hvanhovell Can you let me know if this is needed or not? So I can close it. Thank you.

rxin:
No it's not needed if you are using exactly the same way as the catch-all to release memory. It is basically just the catch-all itself.

viirya:
@rxin Thanks. Appreciate your feedback. I could close this now.

viirya:
@rxin BTW, I see you merged #15989 to downgrade error message level in TaskMemoryManager.

I'd like to modify the error message in Executor too, because the current one is little confusing to developers if they don't know this part exactly and they would think there is memory leak happened.

What do you think? If it is ok for you, I'd submit a pr for it.

Thanks.

rxin:
Can you show an example of a leak that would happen in Executor but not in the callback? Thanks.

viirya:
The test case I added in this pr:

    val rng = new scala.util.Random(42)
    val data = sparkContext.parallelize(Seq.tabulate(100) { i =>
      Row(Array.fill(10)(rng.nextInt(10)))
    })
    val schema = StructType(Seq(
      StructField("arr", DataTypes.createArrayType(DataTypes.IntegerType))
    ))
    val df = spark.createDataFrame(data, schema)
    val exploded = df.select(struct(col("*")).as("star"), explode(col("arr")).as("a"))
    val joined = exploded.join(exploded, "a").drop("a").distinct()
    joined.show()

would thrown an exception like this:

    [info] - SPARK-18487: Consume all elements for show/take to avoid memory leak *** FAILED *** (1 second, 73 milliseconds)
    [info]   org.apache.spark.SparkException: Job aborted due to stage failure: Task 0 in stage 179.0 failed 1 times, most recent failure: Lost task 0.0 in stage 179.0 (TID 501, localhost, executor driver): org.apache.spark.SparkException: Managed memory leak detected; size = 33816576 bytes, TID = 501
    [info]  at org.apache.spark.executor.Executor$TaskRunner.run(Executor.scala:295)
    [info]  at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
    [info]  at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
    [info]  at java.lang.Thread.run(Thread.java:745)

I submitted this pr because @sethah encountered this exception during his test. I think it might be other developers hit this in the future. If they don't know this part, from the error message they would think a memory leak happened. In order to avoid this and provide more useful info, I'd like to modify this error message too.

viirya:
Forget to say, of course, this example will thrown the exception only running in "test".

Other developers would possibly encounter this when they write test codes in the future. If we could provide more info in this error message, we could save their time to investigate this.

What do you think?

viirya:
ping @rxin Appreciate your feedback if you can let me know what you think of my suggestion. Thanks.

PR #15917, opened 2016-11-17
SPARK-18252: Using RoaringBitmap for bloom filters

ponkin:

+ Bloom filter is now using RoaringBitmap as bit vector - to save memory usage
+ Bloom filter is using Murmur 128 bit hash

## How was this patch tested?

All previous tests are valid
Added more accurate test for Byte type


common/sketch/src/main/java/org/apache/spark/util/sketch/Murmur3_128.java
```diff
@@ -0,0 +1,135 @@
...
+ */
+
+package org.apache.spark.util.sketch;
+
+/**
+ * 128-bit Murmur3 hasher.
+ * Best perfomance is on x86_64 platform
+ * Based on implementation https://github.com/aappleby/smhasher/blob/master/src/MurmurHash3.cpp
+ */
+final class Murmur3_128 {
```
srowen: Could you explain why we need to add a custom murmurhash implementation? it seems like conceptually we're just swapping bitmaps for bitmapss. What's the license of this code?

common/sketch/src/test/java/org/apache/spark/util/sketch/Murmur3_128Suite.java
```diff
@@ -0,0 +1,30 @@
+package org.apache.spark.util.sketch;
```
srowen: Missing a license

common/sketch/src/main/java/org/apache/spark/util/sketch/RoaringBitmapArray.java
```diff
@@ -0,0 +1,136 @@
...
+package org.apache.spark.util.sketch;
+
+import org.roaringbitmap.RoaringBitmap;
+
+import java.io.DataInputStream;
+import java.io.DataOutputStream;
+import java.io.IOException;
+import java.util.Arrays;
+
+class RoaringBitmapArray {
```
srowen: Probably a dumb question but why do we need an array of bitmaps? I thought RoaringBitmap was a BitArray counterpart?

ponkin: Unfortunately, RoaringBitmap supports only int indexes [javadoc](http://www.javadoc.io/doc/org.roaringbitmap/RoaringBitmap/0.6.27). Since we need to support bloom filters with size more than Int.MaxValue(BitArray supported 64*IntMaxValue) I decided to partition bloom filter with RoaringBitmap[]. Now Bloom filter can be Long.MaxValue large in size(number of bits).

srowen: Hm, I see: https://github.com/RoaringBitmap/RoaringBitmap/issues/109
I wonder if it's worth using RoaringBitmap then, but, of course making the improvements we want from RB would take work too, and, removing code helps.

common/sketch/src/main/java/org/apache/spark/util/sketch/Murmur3_128.java
ponkin: This decision come from fact that RoaringBitmap supports only int indexes(and can have only Int.MaxValue bits), since we use RoaringBitmap[] to represent vector way larger than Int.MaxValue, we need good hash function to generate long hash values. This implementation is based on [smhasher](https://github.com/aappleby/smhasher/blob/master/src/MurmurHash3.cpp)  and [solr implementation](https://github.com/yonik/java_util/blob/master/src/util/hash/MurmurHash3.java) which are both freely available I suppose.

srowen: Rather than make a new implementation, use Guava's Hashing? Scala and Spark itself already have murmur hash impls but they seem 32-bit only

common/sketch/src/main/java/org/apache/spark/util/sketch/RoaringBitmapArray.java
ponkin: well, the main point of this improvement is to save memory - now even filters with few elements acquire space needed for full bloom filter. With this improvement bloom filter size will increase as we insert values in it.

common/sketch/src/main/java/org/apache/spark/util/sketch/BloomFilter.java
```diff
@@ -233,4 +233,5 @@ public static BloomFilter create(long expectedNumItems, long numBits) {
 
     return new BloomFilterImpl(optimalNumOfHashFunctions(expectedNumItems, numBits), numBits);
   }
+
```
srowen: Go ahead and back out this spurious change; there's another newline at the end of the next file

common/sketch/src/main/java/org/apache/spark/util/sketch/Murmur3_128.java
srowen: OK, maybe that makes sense. If bothering with a custom impl, why not 64 bits instead of 128? it would avoid needing a long[2] to take the result?

srowen (changes requested):
CC @cloud-fan for an initial look before we get further

srowen:
Jenkins add to whitelist

common/sketch/src/main/java/org/apache/spark/util/sketch/Murmur3_128.java
ponkin: In fact the is Murmur hash version which produces 64 - it is called Murmur 2 [Wiki](https://en.wikipedia.org/wiki/MurmurHash), but it is previous version of Murmur, current is Murmur 3 which is faster and give better distribution(Murmur2 has some flaws). Since in bloom filter we always need 1 or more hash functions, creating 128 bit hash(two longs) is pretty good I suppose.

common/sketch/src/main/java/org/apache/spark/util/sketch/RoaringBitmapArray.java
```diff
@@ -0,0 +1,136 @@
...
+
+    private final RoaringBitmap[] data;
+    private long bitCount; // number of 1`s in bitset
```
ponkin: it is only need to calculate expectedFpp (false positive rate) which is not very frequent operation, I agree. will remove

rxin:
Please see my comment on the jira ticket. I'm not sure this is a good idea.

ponkin:
@srowen It seems like I have fixed all your comments.

rxin:
Also cc @lemire 

Can you take a look at the discussion on the JIRA ticket and chime in? https://issues.apache.org/jira/browse/SPARK-18252

lemire:
One of the core Roaring authors (@gssiyankai) worked specifically on Roaring and Bloom filters.  Any thoughts Gregory?

srowen:
Yes all good points, this is more complex than I had imagined, especially with the good point about needing to support older serialized forms. This may not be worth it, yes.

It would be nice to have just one implementation one way or the other, rather than 2. At the least, it may be fine to improve the existing implementation and punt on the rest.

rxin:
@ponkin mind closing the pull request given your findings on https://issues.apache.org/jira/browse/SPARK-18252 ?

PR #15919, opened 2016-11-17
[SPARK-18488][CORE]: Update the value of "spark.ui.port" after application started in cluster mode.

uncleGen:

In cluster mode, the "spark.ui.port" is initialized with 0, but not the real bound port.

after PR:

image

### DIFF
--- a/core/src/main/scala/org/apache/spark/SparkContext.scala
+++ b/core/src/main/scala/org/apache/spark/SparkContext.scala
@@ -2208,7 +2208,7 @@ class SparkContext(config: SparkConf) extends Logging {
   }
 
   /** Post the environment update event once the task scheduler is ready */
-  private def postEnvironmentUpdate() {
+  private[spark] def postEnvironmentUpdate() {
     if (taskScheduler != null) {
       val schedulingMode = getSchedulingMode.toString
       val addedJarPaths = addedJars.keys.toSeq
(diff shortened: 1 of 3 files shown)


srowen:
No, I disagree with this. It misrepresents the configuration supplied to the application, and that's what this tab is supposed to do. It is not trying to report the subsequent bound value.

uncleGen:
@srowen  This PR just works when application run in cluster mode.(https://github.com/apache/spark/blob/master/yarn/src/main/scala/org/apache/spark/deploy/yarn/ApplicationMaster.scala#L192) It is meaningless shown in ui for users. 
As you say, it may misrepresents the configuration, what about convert "spark.ui.port" to "spark.ui.port.bound" or just add another "spark.ui.port.bound"?

srowen:
No, because these are not config properties. This is not a way to show the bound port. This tab shows what the user's input config was.

uncleGen:
But, there are many other config which are set dynamically just after application submited, like "spark.app.id", "spark.driver.port" and so on. 
image

Indeed, this is not a user config property, but we should show some meaningful information, not just '0'.

srowen:
Isn't app ID set if not already set by the user? that's more of an internal config.
spark.driver.port is a good point, though it's set in order to pass the info along in the framework. I don't think that's the case for spark.ui.port?

uncleGen:
OK, I close this PR

PR #15921, opened 2016-11-17
[SPARK-18493] Add missing python APIs: withWatermark and checkpoint to dataframe

brkyvz:

This PR adds two of the newly added methods of `Dataset`s to Python:
`withWatermark` and `checkpoint`

## How was this patch tested?

Doc tests

### DIFF
--- a/python/pyspark/sql/dataframe.py
+++ b/python/pyspark/sql/dataframe.py
@@ -1692,7 +1692,6 @@ def _test():
     globs['sdf'] = sc.parallelize([Row(name='Tom', timestamp=1479441846),
                                    Row(name='Bob', timestamp=1479442946)]).toDF() \
         .select('name', from_unixtime('timestamp').alias('timestamp'))
-                
 
     (failure_count, test_count) = doctest.testmod(
         pyspark.sql.dataframe, globs=globs,
--- a/python/pyspark/sql/dataframe.py
+++ b/python/pyspark/sql/dataframe.py
@@ -322,6 +322,48 @@ class DataFrame(object):
     def __repr__(self):
         return "DataFrame[%s]" % (", ".join("%s: %s" % c for c in self.dtypes))
 
+    @since(2.1)
+    def checkpoint(self, eager=True):
+        """Returns a checkpointed version of this Dataset. Checkpointing can be used to truncate the
+        logical plan of this DataFrame, which is especially useful in iterative algorithms where the
+        plan may grow exponentially. It will be saved to a file inside the checkpoint
+        directory set with L{SparkContext.setCheckpointDir()}.
+
+        :param eager: Whether to checkpoint this DataFrame immediately
+
+        .. note:: Experimental
+        """
+        jdf = self._jdf.checkpoint(eager)
+        return DataFrame(jdf, self.sql_ctx)
+
+    @since(2.1)
+    def withWatermark(self, eventTime, delayThreshold):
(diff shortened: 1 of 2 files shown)


brkyvz:
cc @davies for PySpark changes
cc @liancheng for `checkpoint` API and javadoc update
cc @marmbrus for `withWatermark` API. Question here: should we throw an analysis exception if the Dataset that `withWatermark` is called on is non-streaming?

marmbrus:
No, I don't think we need to throw any exceptions.  Watermarks are defined at batch boundaries, so it would just have no affect for a batch job.

We should make sure that the batch planner knows that it can elide the operator (it probably does not today).

rxin:
@marmbrus Why don't we want to throw exceptions? Wouldn't it help users catch errors early.

sql/core/src/main/scala/org/apache/spark/sql/Dataset.scala
```diff
@@ -485,7 +485,10 @@ class Dataset[T] private[sql](
   def isStreaming: Boolean = logicalPlan.isStreaming
 
   /**
-   * Returns a checkpointed version of this Dataset.
+   * Eagerly checkpoint a Dataset and return the new Dataset. Checkpointing can be used to truncate
+   * the logical plan of this Dataset, which is especially useful in iterative algorithms where the
+   * plan may grow exponentially. It will be saved to a file inside the checkpoint
```
zsxwing: nit: `a file` -> `files`.

Each partition will be saved to one file.

sql/core/src/main/scala/org/apache/spark/sql/Dataset.scala
```diff
@@ -495,7 +498,10 @@ class Dataset[T] private[sql](
   def checkpoint(): Dataset[T] = checkpoint(eager = true)
 
   /**
-   * Returns a checkpointed version of this Dataset.
+   * Returns a checkpointed version of this Dataset. Checkpointing can be used to truncate the
+   * logical plan of this Dataset, which is especially useful in iterative algorithms where the
+   * plan may grow exponentially. It will be saved to a file inside the checkpoint
```
zsxwing: nit: `a file` -> `files`.

Each partition will be saved to one file.

python/pyspark/sql/dataframe.py
```diff
@@ -322,6 +322,53 @@ def show(self, n=20, truncate=True):
     def __repr__(self):
         return "DataFrame[%s]" % (", ".join("%s: %s" % c for c in self.dtypes))
 
+    @since(2.1)
+    def checkpoint(self, eager=True):
+        """Returns a checkpointed version of this Dataset. Checkpointing can be used to truncate the
+        logical plan of this DataFrame, which is especially useful in iterative algorithms where the
+        plan may grow exponentially. It will be saved to a file inside the checkpoint
```
zsxwing: nit: `a file` -> `files`.

Each partition will be saved to one file.

zsxwing (changes requested):
Looks good overall. Just several nits.

brkyvz:
thanks @gatorsmile and @tdas. I addressed your comments. The semantics look a lot cleaner now. That doesn't still mean it's clean though :P

brkyvz:
Hallelujah! @zsxwing shall we merge this?

zsxwing:
LGMT. Merging to master and 2.1.

PR #15923, opened 2016-11-18
[SPARK-4105] retry the fetch or stage if shuffle block is corrupt

davies:

There is an outstanding issue that existed for a long time: Sometimes the shuffle blocks are corrupt and can't be decompressed. We recently hit this in three different workloads, sometimes we can reproduce it by every try, sometimes can't. I also found that when the corruption happened, the beginning and end of the blocks are correct, the corruption happen in the middle. There was one case that the string of block id is corrupt by one character. It seems that it's very likely the corruption is introduced by some weird machine/hardware, also the checksum (16 bits) in TCP is not strong enough to identify all the corruption.

Unfortunately, Spark does not have checksum for shuffle blocks or broadcast, the job will fail if any corruption happen in the shuffle block from disk, or broadcast blocks during network. This PR try to detect the corruption after fetching shuffle blocks by decompressing them, because most of the compression already have checksum in them. It will retry the block, or failed with FetchFailure, so the previous stage could be retried on different (still random) machines.

Checksum for broadcast will be added by another PR.

## How was this patch tested?

Added unit tests

### DIFF
--- a/core/src/main/scala/org/apache/spark/storage/ShuffleBlockFetcherIterator.scala
+++ b/core/src/main/scala/org/apache/spark/storage/ShuffleBlockFetcherIterator.scala
@@ -17,19 +17,22 @@
 
 package org.apache.spark.storage
 
-import java.io.InputStream
+import java.io.{InputStream, IOException}
+import java.nio.ByteBuffer
 import java.util.concurrent.LinkedBlockingQueue
 import javax.annotation.concurrent.GuardedBy
 
+import scala.collection.mutable
 import scala.collection.mutable.{ArrayBuffer, HashSet, Queue}
 import scala.util.control.NonFatal
 
 import org.apache.spark.{SparkException, TaskContext}
 import org.apache.spark.internal.Logging
-import org.apache.spark.network.buffer.ManagedBuffer
+import org.apache.spark.network.buffer.{FileSegmentManagedBuffer, ManagedBuffer}
 import org.apache.spark.network.shuffle.{BlockFetchingListener, ShuffleClient}
 import org.apache.spark.shuffle.FetchFailedException
 import org.apache.spark.util.Utils
+import org.apache.spark.util.io.{ChunkedByteBufferInputStream, ChunkedByteBufferOutputStream}
 
 /**
  * An iterator that fetches multiple blocks. For local blocks, it fetches from the local block
@@ -56,6 +59,7 @@ final class ShuffleBlockFetcherIterator(
     shuffleClient: ShuffleClient,
     blockManager: BlockManager,
     blocksByAddress: Seq[(BlockManagerId, Seq[(BlockId, Long)])],
+    streamWrapper: (BlockId, InputStream) => InputStream,
     maxBytesInFlight: Long,
     maxReqsInFlight: Int)
   extends Iterator[(BlockId, InputStream)] with Logging {
@@ -107,6 +111,9 @@ final class ShuffleBlockFetcherIterator(
 
   /** Current number of requests in flight */
   private[this] var reqsInFlight = 0
+
+  /** The blocks that can't be decompressed successfully */
+  private[this] val corruptedBlocks = mutable.HashSet[String]()
 
   private[this] val shuffleMetrics = context.taskMetrics().createTempShuffleReadMetrics()
 
@@ -305,27 +312,74 @@ final class ShuffleBlockFetcherIterator(
    */
   override def next(): (BlockId, InputStream) = {
     numBlocksProcessed += 1
-    val startFetchWait = System.currentTimeMillis()
-    currentResult = results.take()
-    val result = currentResult
-    val stopFetchWait = System.currentTimeMillis()
-    shuffleMetrics.incFetchWaitTime(stopFetchWait - startFetchWait)
 
-    result match {
-      case SuccessFetchResult(_, address, size, buf, isNetworkReqDone) =>
-        if (address != blockManager.blockManagerId) {
-          shuffleMetrics.incRemoteBytesRead(buf.size)
-          shuffleMetrics.incRemoteBlocksFetched(1)
-        }
-        bytesInFlight -= size
-        if (isNetworkReqDone) {
-          reqsInFlight -= 1
-          logDebug("Number of requests in flight " + reqsInFlight)
-        }
-      case _ =>
+    var result: FetchResult = null
+    var input: InputStream = null
+    while (result == null) {
+      val startFetchWait = System.currentTimeMillis()
+      result = results.take()
+      val stopFetchWait = System.currentTimeMillis()
+      shuffleMetrics.incFetchWaitTime(stopFetchWait - startFetchWait)
+
+      result match {
+        case SuccessFetchResult(blockId, address, size, buf, isNetworkReqDone) =>
+          if (address != blockManager.blockManagerId) {
+            shuffleMetrics.incRemoteBytesRead(buf.size)
+            shuffleMetrics.incRemoteBlocksFetched(1)
+          }
+          bytesInFlight -= size
+          if (isNetworkReqDone) {
+            reqsInFlight -= 1
+            logDebug("Number of requests in flight " + reqsInFlight)
+          }
+
+          val in = try {
+            buf.createInputStream()
+          } catch {
+            // The exception could only be throwed by local shuffle block
+            case e: IOException if buf.isInstanceOf[FileSegmentManagedBuffer] =>
+              logError("Failed to create input stream from local block", e)
+              buf.release()
+              result = FailureFetchResult(blockId, address, e)
+              null
+          }
+          if (in != null) {
+            input = streamWrapper(blockId, in)
+            // Only copy the stream if it's wrapped by compression or encryption, also the size of
+            // block is small (the decompressed block is smaller than maxBytesInFlight)
+            if (!input.eq(in) && size < maxBytesInFlight / 3) {
+              val out = new ChunkedByteBufferOutputStream(64 * 1024, ByteBuffer.allocate)
+              try {
+                // Decompress the whole block at once to detect any corruption, which could increase
+                // the memory usage tne potential increase the chance of OOM.
+                // TODO: manage the memory used here, and spill it into disk in case of OOM.
+                Utils.copyStream(input, out)
+                input = out.toChunkedByteBuffer.toInputStream(true)
+              } catch {
+                case e: IOException =>
+                  buf.release()
+                  if (buf.isInstanceOf[FileSegmentManagedBuffer]
+                    || corruptedBlocks.contains(blockId.toString)) {
+                    result = FailureFetchResult(blockId, address, e)
+                  } else {
+                    logWarning(s"got an corrupted block $blockId from $address, fetch again")
+                    fetchRequests += FetchRequest(address, Array((blockId, size)))
+                    result = null
+                  }
+              } finally {
+                // TODO: release the buf here (earlier)
+                in.close()
+              }
+            }
+          }
+
+        case _ =>
+      }
+
+      // Send fetch requests up to maxBytesInFlight
+      fetchUpToMaxBytes()
     }
-    // Send fetch requests up to maxBytesInFlight
-    fetchUpToMaxBytes()
+    currentResult = result
 
     result match {
       case FailureFetchResult(blockId, address, e) =>
@@ -333,7 +387,7 @@ final class ShuffleBlockFetcherIterator(
 
       case SuccessFetchResult(blockId, address, _, buf, _) =>
         try {
-          (result.blockId, new BufferReleasingInputStream(buf.createInputStream(), this))
+          (result.blockId, new BufferReleasingInputStream(input, this))
         } catch {
           case NonFatal(t) =>
             throwFetchFailedException(blockId, address, t)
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/apache/spark/storage/ShuffleBlockFetcherIterator.scala
```diff
@@ -305,35 +312,82 @@ final class ShuffleBlockFetcherIterator(
...
-          shuffleMetrics.incRemoteBytesRead(buf.size)
-          shuffleMetrics.incRemoteBlocksFetched(1)
-        }
-        bytesInFlight -= size
-        if (isNetworkReqDone) {
-          reqsInFlight -= 1
-          logDebug("Number of requests in flight " + reqsInFlight)
-        }
-      case _ =>
+    var result: FetchResult = null
```
rxin: add documentation explaining what's going on here.

davies:
@joshrosen @zsxwing Could you help to review this one ?

core/src/main/scala/org/apache/spark/storage/ShuffleBlockFetcherIterator.scala
```diff
@@ -305,40 +312,82 @@ final class ShuffleBlockFetcherIterator(
...
+            case e: IOException =>
+              assert(buf.isInstanceOf[FileSegmentManagedBuffer])
+              logError("Failed to create input stream from local block", e)
+              buf.release()
+              throwFetchFailedException(blockId, address, e)
+          }
+
+          input = streamWrapper(blockId, in)
+          // Only copy the stream if it's wrapped by compression or encryption, also the size of
+          // block is small (the decompressed block is smaller than maxBytesInFlight)
```
zsxwing: Does this issue only happen for small blocks? Otherwise, only check small blocks seems not help. Why not add shuffle block checksum instead? Then we can just check the compressed block.

davies: The purpose of this PR is to reduce the possibility that failed job caused by network/disk corruption, without introduce other regression (OOM). Typically, the shuffle blocks are small, so we can have parallel fetching even with this maxBytesInFlight limit. For those few blocks (for example, data skew), we does not check that for now (at least, it's not worse than before).

davies: I tried to add checksum for shuffle blocks in https://github.com/apache/spark/pull/15894, that will have much more complexity and overhead, so in favor of this lighter one.

core/src/main/scala/org/apache/spark/storage/ShuffleBlockFetcherIterator.scala
```diff
@@ -56,8 +59,10 @@ final class ShuffleBlockFetcherIterator(
     shuffleClient: ShuffleClient,
```
JoshRosen: Could you update the Scaladoc to document the two new parameters here? I understand what `streamWrapper` means from context but it might be useful for new readers of this code.

core/src/main/scala/org/apache/spark/storage/ShuffleBlockFetcherIterator.scala
```diff
@@ -305,40 +312,82 @@ final class ShuffleBlockFetcherIterator(
...
+          // block is small (the decompressed block is smaller than maxBytesInFlight)
+          if (detectCorrupt && !input.eq(in) && size < maxBytesInFlight / 3) {
+            val out = new ChunkedByteBufferOutputStream(64 * 1024, ByteBuffer.allocate)
+            try {
+              // Decompress the whole block at once to detect any corruption, which could increase
+              // the memory usage tne potential increase the chance of OOM.
+              // TODO: manage the memory used here, and spill it into disk in case of OOM.
+              Utils.copyStream(input, out)
+              out.close()
+              input = out.toChunkedByteBuffer.toInputStream(true)
```
JoshRosen: Could you put `dispose = true` here to make the boolean parameter clearer?

core/src/main/scala/org/apache/spark/storage/ShuffleBlockFetcherIterator.scala
```diff
@@ -305,40 +312,82 @@ final class ShuffleBlockFetcherIterator(
...
+            } catch {
+              case e: IOException =>
+                buf.release()
+                if (buf.isInstanceOf[FileSegmentManagedBuffer]
+                  || corruptedBlocks.contains(blockId)) {
+                  throwFetchFailedException(blockId, address, e)
+                } else {
+                  logWarning(s"got an corrupted block $blockId from $address, fetch again")
```
JoshRosen: Can we log the `IOException` here? It looks like the exception isn't logged or rethrown from this branch and I think we'll need that information to help debug problems here.

core/src/main/scala/org/apache/spark/storage/ShuffleBlockFetcherIterator.scala
```diff
@@ -305,40 +312,82 @@ final class ShuffleBlockFetcherIterator(
...
+          input = streamWrapper(blockId, in)
+          // Only copy the stream if it's wrapped by compression or encryption, also the size of
+          // block is small (the decompressed block is smaller than maxBytesInFlight)
+          if (detectCorrupt && !input.eq(in) && size < maxBytesInFlight / 3) {
+            val out = new ChunkedByteBufferOutputStream(64 * 1024, ByteBuffer.allocate)
+            try {
+              // Decompress the whole block at once to detect any corruption, which could increase
+              // the memory usage tne potential increase the chance of OOM.
+              // TODO: manage the memory used here, and spill it into disk in case of OOM.
+              Utils.copyStream(input, out)
```
JoshRosen: Do we need to close the input stream here? There might be resources in the decompressor which need to be freed.

core/src/main/scala/org/apache/spark/storage/ShuffleBlockFetcherIterator.scala
JoshRosen: Once we start explicitly managing the memory and support spilling, will it be safe to do this for large blocks, too?

JoshRosen (changes requested):
I left a couple of comments regarding cleanup of decompression buffers and logging of exceptions.

core/src/main/scala/org/apache/spark/storage/ShuffleBlockFetcherIterator.scala
```diff
@@ -108,6 +113,9 @@ final class ShuffleBlockFetcherIterator(
   /** Current number of requests in flight */
   private[this] var reqsInFlight = 0
 
+  /** The blocks that can't be decompressed successfully */
```
lins05: What about add more explanation, for example:

```java
 /** The blocks that can't be decompressed successfully. 
  ** It is used to guarantee that we retry at most once for those corrupted blocks. 
  **/
```

core/src/main/scala/org/apache/spark/storage/ShuffleBlockFetcherIterator.scala
lins05: I think the IOException would be set as the cause of the `FetchFailedException`.

davies:
Manually test this patch with a job that usually failed because of corrupt stream, as the logging said:
```
16/11/20 08:32:07 WARN ShuffleBlockFetcherIterator: got an corrupted block shuffle_5_613_275 from BlockManagerId(6, 10.1.109.163, 34744), fetch again
16/11/20 08:32:07 WARN ShuffleBlockFetcherIterator: got an corrupted block shuffle_5_688_275 from BlockManagerId(6, 10.1.109.163, 34744), fetch again
16/11/20 08:32:07 WARN ShuffleBlockFetcherIterator: got an corrupted block shuffle_5_2434_275 from BlockManagerId(6, 10.1.109.163, 34744), fetch again
16/11/20 08:32:26 WARN ShuffleBlockFetcherIterator: got an corrupted block shuffle_5_878_275 from BlockManagerId(21, 10.1.107.237, 35876), fetch again
16/11/20 08:32:26 WARN ShuffleBlockFetcherIterator: got an corrupted block shuffle_5_1042_275 from BlockManagerId(21, 10.1.107.237, 35876), fetch again
16/11/20 08:32:26 WARN ShuffleBlockFetcherIterator: got an corrupted block shuffle_5_2301_275 from BlockManagerId(21, 10.1.107.237, 35876), fetch again
16/11/20 08:32:26 WARN ShuffleBlockFetcherIterator: got an corrupted block shuffle_5_2546_275 from BlockManagerId(21, 10.1.107.237, 35876), fetch again
16/11/20 08:32:27 WARN ShuffleBlockFetcherIterator: got an corrupted block shuffle_5_3160_275 from BlockManagerId(21, 10.1.107.237, 35876), fetch again
16/11/20 08:32:27 WARN ShuffleBlockFetcherIterator: got an corrupted block shuffle_5_3601_275 from BlockManagerId(21, 10.1.107.237, 35876), fetch again
...
16/11/20 08:32:41 INFO Executor: Finished task 275.0 in stage 26.0 (TID 22187). 5219 bytes result sent to driver
```
The shuffle fetcher got some corrupt blocks for partition 275, it retried once, then the task finally succeeded.

But the retry can not protect all the tasks, some failed as FetchFailed, then the stage is retried:
```
26	   2016/11/20 08:31:24	1.0min	403/1000 (2 failed)			205.6 GB	29.5 GB org.apache.spark.shuffle.FetchFailedException: Stream is corrupted

26 (retry 1)  2016/11/20 08:34:00	34s	200/629 (2 failed)			102.0 GB	14.6 GB	org.apache.spark.shuffle.FetchFailedException: Stream is corrupted

26 (retry 2)  2016/11/20 08:35:25	1.8min	461/461			235.1 GB	33.7 GB
```

The stage 26 succeeded after retried twice.

Another thing is that all the corruption happened only in 2 nodes out of 26. Also a few broadcast block is corrupt on them. They seems that the corruption happens on the receive (fetcher) side of network.

I will update the patch to address comments.

core/src/main/scala/org/apache/spark/storage/ShuffleBlockFetcherIterator.scala
davies: @JoshRosen  I think so.

core/src/main/scala/org/apache/spark/storage/ShuffleBlockFetcherIterator.scala
davies: @lins05 It's already set for FetchFailedException

core/src/main/scala/org/apache/spark/storage/ShuffleBlockFetcherIterator.scala
davies: done

rxin:
cc @zsxwing @JoshRosen does this look good?

davies:
ping @JoshRosen

JoshRosen:
This looks good overall, but one nit: it looks like we don't have any test coverage for the case where `detectCorrupt` is false. We should probably add a test to make sure that the feature flag works properly.

davies:
@JoshRosen Added a test for `detectCorrupt` is false.

zsxwing:
LGTM

zsxwing:
Merging to master.

core/src/main/scala/org/apache/spark/storage/ShuffleBlockFetcherIterator.scala
```diff
@@ -305,40 +316,84 @@ final class ShuffleBlockFetcherIterator(
...
-    // Send fetch requests up to maxBytesInFlight
-    fetchUpToMaxBytes()
 
-    result match {
-      case FailureFetchResult(blockId, address, e) =>
-        throwFetchFailedException(blockId, address, e)
+    var result: FetchResult = null
+    var input: InputStream = null
+    // Take the next fetched result and try to decompress it to detect data corruption,
+    // then fetch it one more time if it's corrupt, throw FailureFetchResult if the second fetch
```
fathersson: Could you elaborate a bit here. In my mind TCP provides pretty robust data transfer, which means that if there is an error, then it's been written to disk corrupted and fetch it one more time won't help.

rxin: We have observed in production a few failures related to this on virtualized environments. It is entirely possible there is a bug in the underlying networking stack, or a bug in Spark's networking stack. But either this way eliminates those issues.

davies: @fathersson The checksum in TCP is only 16 bits,  it's not strong enough for large traffic, usually DFS or other system with heavy TCP traffic will have another application level checksum. Adding to @rxin 's point, we did see this retry helped in production to work around temporary corrupt.

Tagar: Is netty/shuffle data being compressed using Snappy algorithm by default? If so, might be good to idea to enable checksum checking at Netty level too?

https://netty.io/4.0/api/io/netty/handler/codec/compression/SnappyFramedDecoder.html

> Note that by default, validation of the checksum header in each chunk is DISABLED for performance improvements. If performance is less of an issue, or if you would prefer the safety that checksum validation brings, please use the SnappyFramedDecoder(boolean) constructor with the argument set to true.

zsxwing: @Tagar Spark doesn't use Netty's Snappy compression.

PR #15925, opened 2016-11-18
[SPARK-18436][SQL]isin with a empty list throw exception

windpiger:
   when the list of `isin` is empty, the compileFilter will generate the sql string  using `IN()` .
And then connect to execute on sqlserver, it will throw exception:
```
Caused by: com.mysql.jdbc.exceptions.jdbc4.MySQLSyntaxErrorException: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '))' at line 1
```
It is a illeagal grammer for mysql.

   And execute `spark.sql("select * from test where key in()")`, the grammer is also illeagal and throw Exception:

```
mismatched input 'from' expecting {<EOF>, 'WHERE', 'GROUP', 'ORDER', 'HAVING', 'LIMIT', 'LATERAL', 'WINDOW', 'UNION', 'EXCEPT', 'MINUS', 'INTERSECT', 'SORT', 'CLUSTER', 'DISTRIBUTE'}(line 1, pos 9)

== SQL ==
select * from test where key in()
---------^^^

  at org.apache.spark.sql.catalyst.parser.ParseException.withCommand(ParseDriver.scala:197)
  at org.apache.spark.sql.catalyst.parser.AbstractSqlParser.parse(ParseDriver.scala:99)
```
   
   In this PR, add a list empty check in `case class In(value: Expression, list: Seq[Expression])`, so it can unify the action of the spark.sql() and the isin

## How was this patch tested?

**before**
>\>val filter = Array\[String\]()
>\>val jdbcDF = spark.read.format("jdbc").option("url", "jdbc:mysql://xxx").option("dbtable", "testdb.table").option("user", "xxx").option("password", "xxxx").option("driver","com.mysql.jdbc.Driver").load().filter($"id".isin(filter:_*))
```
16/11/18 09:31:25 ERROR executor.Executor: Exception in task 0.0 in stage 0.0 (TID 0)
com.mysql.jdbc.exceptions.jdbc4.MySQLSyntaxErrorException: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '))' at line 1
```
**after**
>\>val filter = Array\[String\]()
>\>val jdbcDF = spark.read.format("jdbc").option("url", "jdbc:mysql://xxx").option("dbtable", "testdb.table").option("user", "xxx").option("password", "xxxx").option("driver","com.mysql.jdbc.Driver").load().filter($"id".isin(filter:_*))
```
java.lang.IllegalArgumentException: requirement failed: list should not be empty
  at scala.Predef$.require(Predef.scala:224)
  at org.apache.spark.sql.catalyst.expressions.In.<init>(predicates.scala:122)
  at org.apache.spark.sql.Column.isin(Column.scala:778)
  ... 48 elided
```

> \>spark.sql("select * from src where key in()")
```
mismatched input 'from' expecting {<EOF>, 'WHERE', 'GROUP', 'ORDER', 'HAVING', 'LIMIT', 'LATERAL', 'WINDOW', 'UNION', 'EXCEPT', 'MINUS', 'INTERSECT', 'SORT', 'CLUSTER', 'DISTRIBUTE'}(line 1, pos 9)

== SQL ==
select * from src where key in()
---------^^^

  at org.apache.spark.sql.catalyst.parser.ParseException.withCommand(ParseDriver.scala:197)
  at org.apache.spark.sql.catalyst.parser.AbstractSqlParser.parse(ParseDriver.scala:99)
  at org.apache.spark.sql.execution.SparkSqlParser.parse(SparkSqlParser.scala:45)
  at org.apache.spark.sql.catalyst.parser.AbstractSqlParser.parsePlan(ParseDriver.scala:53)
  at org.apache.spark.sql.SparkSession.sql(SparkSession.scala:590)
  ... 48 elided
```

### DIFF
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/predicates.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/predicates.scala
@@ -119,6 +119,7 @@ case class In(value: Expression, list: Seq[Expression]) extends Predicate
     with ImplicitCastInputTypes {
 
   require(list != null, "list should not be null")
+  require(list.nonEmpty, "list should not be empty")
 
   override def inputTypes: Seq[AbstractDataType] = value.dataType +: list.map(_.dataType)
 


rxin:
Wouldn't this change the behavior for isin?

If there is a problem with jdbc we should just not generate isin in those.

windpiger:
not generate isin means that 
`select * from test where key in()` is equal to `select * from test` ?

I think this is not we expected behavior ,we should not cover this situation by ignore it.

And spark.sql("") also throw a exception .

gatorsmile:
When we use Spark SQL interface, we issue an error.
```
spark-sql> select * from t1 where c1 in ();
16/11/17 21:08:08 INFO SparkSqlParser: Parsing command: select * from t1 where c1 in ()

Error in query: 
mismatched input 'from' expecting {<EOF>, 'WHERE', 'GROUP', 'ORDER', 'HAVING', 'LIMIT', 'LATERAL', 'WINDOW', 'UNION', 'EXCEPT', 'MINUS', 'INTERSECT', 'SORT', 'CLUSTER', 'DISTRIBUTE'}(line 1, pos 9)
```

This is consistent with Hive.
```
hive> select * from t where col1 in ();
NoViableAltException(307@[()* loopback of 438:5: ( ( KW_NOT precedenceEqualNegatableOperator notExpr= precedenceBitwiseOrExpression ) -> ^( KW_NOT ^( precedenceEqualNegatableOperator $precedenceEqualExpression $notExpr) ) | ( precedenceEqualOperator equalExpr= precedenceBitwiseOrExpression ) -> ^( precedenceEqualOperator $precedenceEqualExpression $equalExpr) | ( KW_NOT KW_IN LPAREN KW_SELECT )=> ( KW_NOT KW_IN subQueryExpression ) -> ^( KW_NOT ^( TOK_SUBQUERY_EXPR ^( TOK_SUBQUERY_OP KW_IN ) subQueryExpression $precedenceEqualExpression) ) | ( KW_NOT KW_IN expressions ) -> ^( KW_NOT ^( TOK_FUNCTION KW_IN $precedenceEqualExpression expressions ) ) | ( KW_IN LPAREN KW_SELECT )=> ( KW_IN subQueryExpression ) -> ^( TOK_SUBQUERY_EXPR ^( TOK_SUBQUERY_OP KW_IN ) subQueryExpression $precedenceEqualExpression) | ( KW_IN expressions ) -> ^( TOK_FUNCTION KW_IN $precedenceEqualExpression expressions ) | ( KW_NOT KW_BETWEEN (min= precedenceBitwiseOrExpression ) KW_AND (max= precedenceBitwiseOrExpression ) ) -> ^( TOK_FUNCTION Identifier["between"] KW_TRUE $left $min $max) | ( KW_BETWEEN (min= precedenceBitwiseOrExpression ) KW_AND (max= precedenceBitwiseOrExpression ) ) -> ^( TOK_FUNCTION Identifier["between"] KW_FALSE $left $min $max) )*])
	at org.apache.hadoop.hive.ql.parse.HiveParser_IdentifiersParser$DFA43.specialStateTransition(HiveParser_IdentifiersParser.java)
```

However, when we use DataFrame, no error is reported. Instead, we return an empty result set. 
```Scala
    val df = Seq((1, "x"), (2, "y"), (3, "z")).toDF("a", "b")
    df.filter($"a".isin()).show()
```

@windpiger You want to make DataFrame APIs consistent with Spark SQL interface, right?

windpiger:
yes, I think it is better they are consistent.  
And exception should be throwed before connect to the sql server.

srowen:
The original issue was that the generated SQL code was, apparently, invalid for at least one JDBC driver. Semantically there's nothing wrong with "in (empty set)" right? it's false, always.

I had suggested, and I though @hvanhovell agreed, that it would be best to simply optimize away this condition early on, so that it's never even generated. How it's rendered in SQL is then a moot point.

Another solution, I thought, was to modify how the SQL is rendered for JDBC clients to match whatever is actually the right way to say this: `IN ('')`?

But this seems like something else. It's not quite relevant to compare `IN ()` in SQL dialects, which is the question here, with `isin()` in Scala, which seems syntactically and semantically just fine. Why make it change behavior to not work at all?

rxin:
+1

I don't think we really gain anything by restricting this use case that we allowed before.

rxin:
BTW optimizer technically shouldn't have anything to do with correctness, so it shouldn't be the optimizer's job to make sure jdbc can connect. I'd just generate the SQL to "false" (depending on how we define the behavior).

srowen:
Yes you're exactly right, what I meant was that if "IN (empty set)" is never generated in SQL, then there's no problem with how it's generated. I suppose it would still also be correct to make sure the the SQL generation code does generate "IN ('')" or whatever is correct for some/all dialects just in case this ever comes up again.

windpiger:
ok,thanks! 
another one is spark.sql("select * from test where key in()") throw a exception, it is not consistent with the isin of dataset, it is ok?

if is ok, I will try to optimize the in() condition, and make the SQL generation code is correct for mysql etc

srowen:
That isn't the same case. In one instance you are showing an example of SQL syntax, and in the other, usage of a programmatic Scala API. The SQL query is not semantically wrong, but syntactically wrong.

It would at least be OK to make sure that the "IN ()" case is generated in a way that works for all SQL engines. If it's always correct to write "IN ('')" then that's fine, but, doesn't that mean "a set containing the empty string" in some cases? I'm sort of surprised that's the way to express it in some SQL engines. Is it really empty?

windpiger:
write IN('') is not suitable.
if a query has a IN(), maybe we can return empty result without connect to the sql server.

rxin:
That's why I was suggesting just generate "false" for in(empty seq).

srowen:
@windpiger can you close this or, if desired, change it to work on an alternate approach?

PR #15926, opened 2016-11-18
[SPARK-16803] [SQL] SaveAsTable does not work when target table is a Hive serde table

gatorsmile:

In Spark 2.0, `SaveAsTable` does not work when the target table is a Hive serde table, but Spark 1.6 works. 

**Spark 1.6**

``` Scala
scala> sql("create table sample.sample stored as SEQUENCEFILE as select 1 as key, 'abc' as value")
res2: org.apache.spark.sql.DataFrame = []

scala> val df = sql("select key, value as value from sample.sample")
df: org.apache.spark.sql.DataFrame = [key: int, value: string]

scala> df.write.mode("append").saveAsTable("sample.sample")

scala> sql("select * from sample.sample").show()
+---+-----+
|key|value|
+---+-----+
|  1|  abc|
|  1|  abc|
+---+-----+
```

**Spark 2.0**

``` Scala
scala> df.write.mode("append").saveAsTable("sample.sample")
org.apache.spark.sql.AnalysisException: Saving data in MetastoreRelation sample, sample
 is not supported.;
```

So far, we do not plan to support it in Spark 2.1 due to the risk. Spark 1.6 works because it internally uses insertInto. But, if we change it back it will break the semantic of saveAsTable (this method uses by-name resolution instead of using by-position resolution used by insertInto). More extra changes are needed to support `hive` as a `format` in DataFrameWriter.

Instead, users should use insertInto API. This PR corrects the error messages. Users can understand how to bypass it before we support it in a separate PR.

Test cases are added

### DIFF
--- a/sql/hive/src/test/scala/org/apache/spark/sql/hive/MetastoreDataSourcesSuite.scala
+++ b/sql/hive/src/test/scala/org/apache/spark/sql/hive/MetastoreDataSourcesSuite.scala
@@ -416,14 +416,14 @@ class MetastoreDataSourcesSuite extends QueryTest with SQLTestUtils with TestHiv
   test("saveAsTable(CTAS) using append and insertInto when the target table is Hive serde") {
     val tableName = "tab1"
     withTable(tableName) {
-      sql(s"CREATE TABLE $tableName stored as SEQUENCEFILE as select 1 as key, 'abc' as value")
+      sql(s"CREATE TABLE $tableName STORED AS SEQUENCEFILE AS SELECT 1 AS key, 'abc' AS value")
 
-      val df = sql(s"select key, value from $tableName")
+      val df = sql(s"SELECT key, value FROM $tableName")
       val e = intercept[AnalysisException] {
         df.write.mode(SaveMode.Append).saveAsTable(tableName)
       }.getMessage
-      assert(e.contains("Saving data in the Hive serde table `default`.`tab1` is not supported. " +
-        "Instead, please use the insertInto() API"))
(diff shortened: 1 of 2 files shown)


gatorsmile:
cc @cloud-fan

cloud-fan:
LGTM

sql/hive/src/test/scala/org/apache/spark/sql/hive/MetastoreDataSourcesSuite.scala
```diff
@@ -413,6 +413,40 @@ class MetastoreDataSourcesSuite extends QueryTest with SQLTestUtils with TestHiv
...
+  test("saveAsTable(CTAS) using overwrite when the target table is Hive serde") {
+    val tableName = "tab1"
+    withTable(tableName) {
+      sql(s"CREATE TABLE $tableName STORED AS SEQUENCEFILE AS SELECT 1 AS key, 'abc' AS value")
+      val df = sql(s"SELECT key, value FROM $tableName")
+      df.write.mode(SaveMode.Overwrite).saveAsTable(tableName)
+      val tableMeta = spark.sessionState.catalog.getTableMetadata(TableIdentifier(tableName))
+      // When the mode is OVERWRITE, we drop the Hive serde tables and create a data source table
+      // TODO: Based on the definition of OVERWRITE, no change should be made on the table
+      // definition. When recreate the table, we need to create a Hive serde table.
```
gatorsmile: Sure. Will do it.

gatorsmile:
Thanks! Merging to master/2.1

PR #15932, opened 2016-11-18
[SPARK-18448][CORE] SparkSession should implement java.lang.AutoCloseable like JavaSparkContext

srowen:

Just adds `close()` + `Closeable` as a synonym for `stop()`. This makes it usable in Java in try-with-resources, as suggested by @ash211  (`Closeable` extends `AutoCloseable` BTW)

## How was this patch tested?

Existing tests

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/SparkSession.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/SparkSession.scala
@@ -18,6 +18,7 @@
 package org.apache.spark.sql
 
 import java.beans.Introspector
+import java.io.Closeable
 import java.util.concurrent.atomic.AtomicReference
 
 import scala.collection.JavaConverters._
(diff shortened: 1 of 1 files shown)


ash211:
Yep that's precisely what I was envisioning.  Thanks @srowen !

rxin:
Nice - given it's a tiny change I'd put this in 2.1, unless somebody else objects.

@srowen ?

srowen:
Sure I'll put it in first thing tomorrow if there are no objections.

felixcheung:
Great!

srowen:
Merged to master/2.1

sql/core/src/main/scala/org/apache/spark/sql/SparkSession.scala
```diff
@@ -648,6 +649,13 @@ class SparkSession private(
   }
 
   /**
+   * Synonym for `stop()`.
+   *
+   * @since 2.2.0
```
rxin: need to update this

srowen: Yeah fixed it in a follow up.

PR #15934, opened 2016-11-18
[SPARK-18497][SS]Make ForeachSink support watermark

zsxwing:

The issue in ForeachSink is the new created DataSet still uses the old QueryExecution. When `foreachPartition` is called, `QueryExecution.toString` will be called and then fail because it doesn't know how to plan EventTimeWatermark.

This PR just replaces the QueryExecution with IncrementalExecution to fix the issue.

## How was this patch tested?

`test("foreach with watermark")`.


zsxwing:
cc @marmbrus

marmbrus:
LGTM

tdas:
Merging to master, 2.1 and 2.0

tdas:
Seems like this is a correctness bug as aggregation will give wrong results when using aggregations.

PR #15936, opened 2016-11-18
[SPARK-18504][SQL] Scalar subquery with extra group by columns returning incorrect result

nsyca:

This PR blocks an incorrect result scenario in scalar subquery where there are GROUP BY column(s)
that are not part of the correlated predicate(s).

Example:
// Incorrect result
Seq(1).toDF("c1").createOrReplaceTempView("t1")
Seq((1,1),(1,2)).toDF("c1","c2").createOrReplaceTempView("t2")
sql("select (select sum(-1) from t2 where t1.c1=t2.c1 group by t2.c2) from t1").show

// How can selecting a scalar subquery from a 1-row table return 2 rows?

## How was this patch tested?
sql/test, catalyst/test
new test case covering the reported problem is added to SubquerySuite.scala


nsyca:
I think I have contaminated the PR with some of my old code. I will push a new PR soon.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
```diff
@@ -1244,6 +1244,23 @@ class Analyzer(
         case s @ ScalarSubquery(sub, conditions, exprId)
             if sub.resolved && conditions.isEmpty && sub.output.size != 1 =>
           failAnalysis(s"Scalar subquery must return only one column, but got ${sub.output.size}")
+        // SPARK-1854: block cases where GROUP BY columns
+        // are not part of the correlated columns
+        case s @ ScalarSubquery(sub, conditions, exprId)
```
hvanhovell: We also do a part of the scalar subquery checks in CheckAnalysis. Lets move this code there.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
```diff
@@ -1244,6 +1244,23 @@ class Analyzer(
...
+            if sub.resolved && sub.isInstanceOf[Aggregate] =>
```
hvanhovell: Nit (for a next time): It is easier to do this in the pattern match. For example:
`case s @ ScalarSubquery(agg: Aggregate, conditions, exprId) if agg.resolved => `

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
hvanhovell: Or move that code here - what ever is more concise

hvanhovell: BTW: we should not assume that an aggregate is always the top-level node for a correlated scalar subquery. We add a projection to the subquery plan when it shares attributes with the outer plan. This is solved in CheckAnalysis.

nsyca:
The latest push (3181a08) of the PR has the following changes from its previous version:

1. Consolidate existing checking from Analyzer.scala to CheckAnalysis.scala
2. Make use of `ExpressionSet` class to compare the canonical form of expressions between two ExpressionSet
3. Revise the test case to be more concise (using only one table) and, more importantly, exercise the `ExpressionSet` code in the comparison of the columns in the GROUP BY clause and the columns in the WHERE clause at line 143 of the file CheckAnalysis.scala

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/CheckAnalysis.scala
```diff
@@ -117,19 +117,36 @@ trait CheckAnalysis extends PredicateHelper {
...
+              // SPARK-18504: block cases where GROUP BY columns
+              // are not part of the correlated columns
+              val groupByCols = ExpressionSet.apply(agg.groupingExpressions.flatMap(_.references))
+              val predicateCols = ExpressionSet.apply(conditions.flatMap(_.references))
+              val invalidCols = groupByCols.diff(predicateCols)
+              // GROUP BY columns must be a subset of columns in the predicates
+              if (invalidCols.nonEmpty) {
+                failAnalysis(s"""
+                  |GROUP BY column(s) in scalar subquery must exist in the WHERE clause:
+                  |${invalidCols.toString}""".stripMargin.replaceAll("\n", " "))
```
hvanhovell: BTW if you use regular strings and regular concatenation instead of raw strings, you don't have to use `replaceAll("\n", " ")`.

nsyca: Right. Thank you.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/CheckAnalysis.scala
```diff
@@ -117,19 +117,36 @@ trait CheckAnalysis extends PredicateHelper {
...
+
+              // SPARK-18504: block cases where GROUP BY columns
+              // are not part of the correlated columns
+              val groupByCols = ExpressionSet.apply(agg.groupingExpressions.flatMap(_.references))
+              val predicateCols = ExpressionSet.apply(conditions.flatMap(_.references))
+              val invalidCols = groupByCols.diff(predicateCols)
+              // GROUP BY columns must be a subset of columns in the predicates
+              if (invalidCols.nonEmpty) {
+                failAnalysis(
+                  "GROUP BY column(s) in scalar subquery must exist in the WHERE clause: " +
```
hvanhovell: This might be a little hard to understand for someone who is not fully emerged in the world of subqueries. How about: `a GROUP BY clause in a scalar correlated subquery cannot contain non-correlated columns: `

Is this better?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/CheckAnalysis.scala
```diff
@@ -117,19 +117,36 @@ trait CheckAnalysis extends PredicateHelper {
...
+                  s"${invalidCols.toString}")
```
hvanhovell: `invalidCols.mkString(",")`? (No need for string interpolation)

hvanhovell (approved):
Two minor comments, otherwise LGTM

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/CheckAnalysis.scala
nsyca: It is.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/CheckAnalysis.scala
nsyca: Oop! missed this comment. Will work on it.

nsyca:
Should I say `retest this please` to kick off the regression test?

hvanhovell:
It should pick it up automatically.

hvanhovell:
ok to test

hvanhovell:
add to whitelist

hvanhovell:
LGTM. Merging to master/2.1/2.0.

ericl:
This seems to be causing https://issues.apache.org/jira/browse/SPARK-18814

PR #15938, opened 2016-11-19
[SPARK-18448][CORE] Fix @since 2.1.0 on new SparkSession.close() method

srowen:

Fix @since 2.1.0 on new SparkSession.close() method. I goofed in https://github.com/apache/spark/pull/15932 because it was back-ported to 2.1 instead of just master as originally planned.


srowen:
Merged to master/2.1

PR #15889, opened 2016-11-15
[SPARK-18445][BUILD][DOCS] Fix the markdown for `Note:`/`NOTE:`/`Note that`/`'''Note:'''` across Scala/Java API documentation

HyukjinKwon:
## What changes were proposed in this pull request?

It seems in Scala/Java,

- `Note:`
- `NOTE:`
- `Note that`
- `'''Note:'''`
- `@note`

This PR proposes to fix those to `@note` to be consistent.

**Before**

- Scala
  

- Java
  

**After**

- Scala
  

- Java
  

## How was this patch tested?

The notes were found via

```bash
grep -r "NOTE: " . | \ # Note:|NOTE:|Note that|'''Note:'''
grep -v "// NOTE: " | \  # starting with // does not appear in API documentation.
grep -E '.scala|.java' | \ # java/scala files
grep -v Suite | \ # exclude tests
grep -v Test | \ # exclude tests
grep -e 'org.apache.spark.api.java' \ # packages appear in API documenation
-e 'org.apache.spark.api.java.function' \ # note that this is a regular expression. So actual matches were mostly `org/apache/spark/api/java/functions ...`
-e 'org.apache.spark.api.r' \
...
```

```bash
grep -r "Note that " . | \ # Note:|NOTE:|Note that|'''Note:'''
grep -v "// Note that " | \  # starting with // does not appear in API documentation.
grep -E '.scala|.java' | \ # java/scala files
grep -v Suite | \ # exclude tests
grep -v Test | \ # exclude tests
grep -e 'org.apache.spark.api.java' \ # packages appear in API documenation
-e 'org.apache.spark.api.java.function' \
-e 'org.apache.spark.api.r' \
...
```

```bash
grep -r "Note: " . | \ # Note:|NOTE:|Note that|'''Note:'''
grep -v "// Note: " | \  # starting with // does not appear in API documentation.
grep -E '.scala|.java' | \ # java/scala files
grep -v Suite | \ # exclude tests
grep -v Test | \ # exclude tests
grep -e 'org.apache.spark.api.java' \ # packages appear in API documenation
-e 'org.apache.spark.api.java.function' \
-e 'org.apache.spark.api.r' \
...
```

```bash
grep -r "'''Note:'''" . | \ # Note:|NOTE:|Note that|'''Note:'''
grep -v "// '''Note:''' " | \  # starting with // does not appear in API documentation.
grep -E '.scala|.java' | \ # java/scala files
grep -v Suite | \ # exclude tests
[...]

PR #15940, opened 2016-11-19
[SPARK-18447][BUILD][DOCS]Fix `Note:`/`NOTE:`/`Note that` across Pyth…

aditya1702:

It seems that just like in Scala/Java there is inconsistency in the Python API too at some places,

Note:
NOTE:
Note that
'''Note:'''
@note
This PR proposes to fix those to @note to be consistent.

## How was this patch tested?

This was tested by searching in the editor for versions of "NOTE" in the python APIs and then fixing them to become consistence.

This PR is a sub-extension of #15889

### DIFF
--- a/core/src/main/scala/org/apache/spark/api/python/PythonHadoopUtil.scala
+++ b/core/src/main/scala/org/apache/spark/api/python/PythonHadoopUtil.scala
@@ -106,7 +106,7 @@ private[python] class WritableToJavaConverter(
 }
 
 /**
- * A converter that converts common types to [[org.apache.hadoop.io.Writable]]. Note that array
+ * A converter that converts common types to [[org.apache.hadoop.io.Writable]]. @note Array
  * types are not supported since the user needs to subclass [[org.apache.hadoop.io.ArrayWritable]]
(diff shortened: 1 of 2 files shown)


aditya1702:
@HyukjinKwon Could you review?

rxin:
I don't think any of these actually matter, because these functions are not public documentations.

HyukjinKwon:
Yes, you should check if they are exposed in the API documentation. I intentionally updated only those ones before.

HyukjinKwon:
Ah, @aditya1702, I assume, from the title, you meant to do this for Python documentation. Actually, the PR you pointed out deals with all of Scala/Java documentation if I haven’t missed some. Maybe, `.. note:` in Python codes should be handled for Python API documentation including some images from manually built API documentation in the PR. If you are not sure of what to fix, I can take over this and cc you in the PR. I guess you could close this PR meanwhile.

aditya1702:
@HyukjinKwon @rxin Yes I think I got confused as pointed out by Hyukjin above 😅 . @HyukjinKwon you can take this one over and cc me. I will work on another issue. Sorry for the troubles

PR #15942, opened 2016-11-20
[SPARK-18407] Inferred partition columns cause assertion error in StructuredStreaming

brkyvz:

It turns out we are a bit enthusiastic when providing users partition columns when they read the data even if they didn't specify it in their schema. This causes an assertion error in Streaming jobs, because the `Attribute`s of a given trigger don't match the `Attribute`s returned by the DataSource. The DataSource returns additional partition columns all the time.

While this is weird behavior for batch as well IMHO, because someone asked for a specific schema, but we returned them something else, apparently this behavior existed since Spark 1.6. I didn't try older versions. Anyway, I tried fixing this by not enforcing a strict size check, but by picking out the columns that we want from the batch DataSource.

## How was this patch tested?

Regression test


brkyvz:
Closing this in favor of #15951

PR #15943, opened 2016-11-20
[SPARK-18511] Added an api for join operations just on column name and the join type

shiv4nsh:
Added an api for join operations just on column name and the join type. An Api in which we do not have to provide the column sequences.

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/Dataset.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/Dataset.scala
@@ -733,6 +733,30 @@ class Dataset[T] private[sql](
   def join(right: Dataset[_], usingColumns: Seq[String]): DataFrame = {
     join(right, usingColumns, "inner")
   }
+  
+  /**
+    * Equi-join with another [[DataFrame]] using the given column.
+    *
+    * Different from other join functions, the join column will only appear once in the output,
+    * i.e. similar to SQL's `JOIN USING` syntax.
+    *
+    * {{{
+    *   // Joining df1 and df2 using the column "user_id"
+    *   df1.join(df2, "user_id","left")
+    * }}}
+    *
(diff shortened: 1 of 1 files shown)


sql/core/src/main/scala/org/apache/spark/sql/Dataset.scala
```diff
@@ -705,12 +701,10 @@ class Dataset[T] private[sql](
...
-   *
-   * @note If you perform a self-join using this function without aliasing the input
-   * [[DataFrame]]s, you will NOT be able to reference any columns after the join, since
-   * there is no way to disambiguate which side of the join you would like to reference.
-   *
-   * @group untypedrel
+    * @note If you perform a self-join using this function without aliasing the input
+   *        [[DataFrame]]s, you will NOT be able to reference any columns after the join, since
+   *        there is no way to disambiguate which side of the join you would like to reference.
+    * @group untypedrel
```
HyukjinKwon: Can we exclude unrelated changes here? Also, `@note` has the consistent indentation across documentation. I guess it is not encouraged to change some of them inconsistently without a concrete reason.

HyukjinKwon: Oh, no. I mean some changes here such as

```diff
-   *
-   * @group action
+    * @group action
```

seems not related with what the PR tries to proposes.

HyukjinKwon: It seems my IntelliJ does not. Whether it is a problem or not, reformatting seems not the goal of this PR and I guess it seems reformatting is arguable.

shiv4nsh:
@HyukjinKwon : I have Closed this PR and added a new PR #15944 with the main changes that were suggested in this PR .

PR #15944, opened 2016-11-20
[SPARK-18511] Added the Api for join with only one column name and join type

shiv4nsh:
Added an api for join operations just on column name and the join type. An Api in which we do not have to provide the column sequences.

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/Dataset.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/Dataset.scala
@@ -743,6 +743,30 @@ class Dataset[T] private[sql](
   }
 
   /**
+   * Equi-join with another [[DataFrame]] using the given column.
+   *
+   * Different from other join functions, the join column will only appear once in the output,
+   * i.e. similar to SQL's `JOIN USING` syntax.
+   *
+   * {{{
+   *   // Joining df1 and df2 using the column "user_id"
+   *   df1.join(df2, "user_id","left")
+   * }}}
+   *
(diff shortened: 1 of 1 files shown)


srowen:
I don't think this buys much of anything over the method below. We don't need every possible overload here.

shiv4nsh:
@srowen : This would help us write a much cleaner code , making sequence each and everytime makes our life difficult a little bit. This would allow us to write code in a more cleaner way. For example if we need to do left join on two tables using one field then we do not need to put it inside a sequence and then pass , now we can just pass directly.
Like this : `df1.join(df2,"commonFeild","left").show`

srowen:
`Seq("commonField")` is much less clean that `"commonField"`?

shiv4nsh:
Actually I was looking into the code and hence found Seq("commonFeild") in the code , but it was not making much sense as to why one should make it a sequence if only one column name is needed hence I thought this alias can be added. In my opinion , in a larger code base `"commonField"` can be taken as a little more cleaner than `Seq("commonField")` and hence the improvement !

HyukjinKwon:
> We don't need every possible overload here.

+1

srowen:
An API that takes 1 or more things takes a Seq of things, which possibly has 1 element. That doesn't seem confusing to me at all, or difficult. I would not like to make this change, please.

shiv4nsh:
Okay @srowen , if you think so ! please go ahead and close this PR !

srowen:
@shiv4nsh you will have to close this PR

PR #14909, opened 2016-09-01
revert PR#10896 and PR#14865

cloud-fan:
## What changes were proposed in this pull request?

according to the discussion in the original PR #10896 and the new approach PR #14876 , we decided to revert these 2 PRs and go with the new approach.
## How was this patch tested?

N/A

PR #15945, opened 2016-11-20
[SPARK-12978][SQL] Merge unnecessary partial aggregates

maropu:
This pr is to merge unnecessary partial aggregates if the inputs of aggregates satisfy the distribution requirement of these partial aggregates. This pr is rework based on the @cloud-fan 's suggestion in #14909.

## How was this patch tested?
Add tests in `PlannerSuite` to check if these partial aggregates are removed by catalyst.


maropu:
@hvanhovell @cloud-fan I think this target might be 2.2.0, so could you check this after 2.1 is cut. Thanks!

maropu:
@hvanhovell @cloud-fan Could you check this? Thanks!

sql/core/src/main/scala/org/apache/spark/sql/execution/QueryExecution.scala
```diff
@@ -90,18 +92,27 @@ class QueryExecution(val sparkSession: SparkSession, val logical: LogicalPlan) {
...
 
   /** A sequence of rules that will be applied in order to the physical plan before execution. */
-  protected def preparations: Seq[Rule[SparkPlan]] = Seq(
-    python.ExtractPythonUDFs,
-    PlanSubqueries(sparkSession),
-    EnsureRequirements(sparkSession.sessionState.conf),
-    CollapseCodegenStages(sparkSession.sessionState.conf),
-    ReuseExchange(sparkSession.sessionState.conf),
-    ReuseSubquery(sparkSession.sessionState.conf))
+  override protected def batches: Seq[Batch] = Seq(
```
cloud-fan: How about we add a `PhysicalOptimizer` to do these things? then we can simply write `lazy val executedPlan: SparkPlan = physicalOptimizer.execute(sparkPlan))` instead of `lazy val executedPlan: SparkPlan = prepareForExecution(sparkPlan)`

maropu: Aha, good idea, so I try to do so.

maropu:
@cloud-fan How about this fix?

maropu:
@cloud-fan ping

sql/core/src/main/scala/org/apache/spark/sql/execution/PhysicalOptimizer.scala
```diff
@@ -0,0 +1,48 @@
...
+
+import org.apache.spark.sql.SparkSession
+import org.apache.spark.sql.catalyst.rules.RuleExecutor
+import org.apache.spark.sql.execution.aggregate.MergePartialAggregate
+import org.apache.spark.sql.execution.exchange.{EnsureRequirements, ReuseExchange}
+
+/**
+ * A sequence of rules that will be applied in order to the physical plan before execution.
+ */
+class PhysicalOptimizer(sparkSession: SparkSession) extends RuleExecutor[SparkPlan] {
```
cloud-fan: let's think about a better name, it does more than only optimization

sql/core/src/main/scala/org/apache/spark/sql/execution/QueryExecution.scala
```diff
@@ -79,34 +78,21 @@ class QueryExecution(val sparkSession: SparkSession, val logical: LogicalPlan) {
     planner.plan(ReturnAnswer(optimizedPlan)).next()
   }
 
+  /**
+   * Physical query plan optimizer.
+   */
+  lazy val optimizer: RuleExecutor[SparkPlan] = new PhysicalOptimizer(sparkSession)
```
cloud-fan: shall we put it in `SessionState` like analyzer and optimizer?

maropu: yea, SGTM. I'll try to fix

sql/core/src/main/scala/org/apache/spark/sql/execution/PhysicalOptimizer.scala
maropu: Oh, yea. So, how about `PhysicalPlanRewriter`?

maropu:
Jenkins, retest this please.

maropu:
@cloud-fan ping

sql/core/src/main/scala/org/apache/spark/sql/execution/PhysicalOptimizer.scala
cloud-fan: SGTM

sql/core/src/main/scala/org/apache/spark/sql/execution/QueryExecution.scala
```diff
@@ -82,32 +81,14 @@ class QueryExecution(val sparkSession: SparkSession, val logical: LogicalPlan) {
...
 
   /** Internal version of the RDD. Avoids copies and has no schema */
   lazy val toRdd: RDD[InternalRow] = executedPlan.execute()
 
-  /**
-   * Prepares a planned [[SparkPlan]] for execution by inserting shuffle operations and internal
-   * row format conversions as needed.
-   */
-  protected def prepareForExecution(plan: SparkPlan): SparkPlan = {
-    preparations.foldLeft(plan) { case (sp, rule) => rule.apply(sp) }
```
cloud-fan: how about we return an anonymous `RuleExecutor[SparkPlan]` here? then we don't need to bother the name

sql/core/src/main/scala/org/apache/spark/sql/execution/aggregate/AggregateExec.scala
```diff
@@ -0,0 +1,68 @@
...
+import org.apache.spark.sql.catalyst.expressions.aggregate.AggregateExpression
+import org.apache.spark.sql.catalyst.plans.physical._
+import org.apache.spark.sql.execution.SparkPlan
+import org.apache.spark.sql.execution.UnaryExecNode
+
+/**
+ * A base class for aggregate implementation.
+ */
+abstract class AggregateExec extends UnaryExecNode {
+  self: SparkPlan =>
```
cloud-fan: is it needed? `UnaryExecNode` already extends `SparkPlan`

sql/core/src/main/scala/org/apache/spark/sql/execution/aggregate/MergePartialAggregate.scala
```diff
@@ -0,0 +1,96 @@
...
+ */
+object AggregatePair {
+
+  def unapply(plan: SparkPlan): Option[(AggregateExec, AggregateExec)] = plan match {
+    case outer: AggregateExec if outer.child.isInstanceOf[AggregateExec] =>
+      val inner = outer.child.asInstanceOf[AggregateExec]
+
+      // Check if classes and grouping keys are the same with each other to make sure the two
+      // aggregates are for the same group by a GROUP-BY clause.
+      if (outer.getClass.isAssignableFrom(inner.getClass) &&
```
cloud-fan: why not `outer.getClass == inner.getClass`?

sql/core/src/main/scala/org/apache/spark/sql/execution/QueryExecution.scala
maropu: okay, I'll try to fix.

sql/core/src/main/scala/org/apache/spark/sql/execution/aggregate/AggregateExec.scala
maropu: oh, you're right and this is meaningless. I'll remove this.

sql/core/src/main/scala/org/apache/spark/sql/execution/aggregate/MergePartialAggregate.scala
maropu: fixed

sql/core/src/main/scala/org/apache/spark/sql/execution/aggregate/AggregateExec.scala
maropu: removed

sql/core/src/main/scala/org/apache/spark/sql/execution/QueryExecution.scala
maropu: I think, if we use anonymous classes here, it seems we cannot avoid duplicate rule entries in `IncrementalExecution`: https://github.com/apache/spark/pull/15945/commits/4f1240d90b6e30947bbb7eae6bb5efcd425cb8d4#diff-13a3f1b22cd7c812e433f771d39eec97R103.
I keep looking for other approaches to avoid this though, I would appreciate your more suggestions.

maropu:
I'll close for now.

PR #15946, opened 2016-11-20
[SPARK-18513][Structured Streaming] Record and recover watermark

lw-lin:

In order to log source offsets, watermark and other possible states, this patch adds a `StreamExecutionProgress` case class to hold them.

This patch records `StreamExecutionProgress`'s json representation into logs and restores the watermark upon recovery.

`StreamExecutionProgress`'s json representation:
```
v1
{
  "offsets" : [ "1", "-", "-", "2" ],
  "watermark" : 200
}
```

## Known issue

Today for watermark to be collected correctly, a sink should trigger the real execution the dataset it received in every batch. However, during the recovery process, a sink might skip a batch (such as [MemorySink#addBatch()](https://github.com/apache/spark/blob/v2.0.2/sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/memory.scala#L202-L204)), and the watermark just goes wrong. This is a separate issue we should fix.

## How was this patch tested?

- added `StreamExecutionProgressLogSuite`
- enabled `test("recovery")` in `WatermarkSuite`


sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/memory.scala
```diff
@@ -202,6 +202,8 @@ class MemorySink(val schema: StructType, outputMode: OutputMode) extends Sink wi
             s"Output mode $outputMode is not supported by MemorySink")
       }
     } else {
+      // TODO: THIS IS NOT RIGHT!
+      data.collect()
```
lw-lin: this added line is to demonstrate the related issue
this should really be reverted

lw-lin:
@marmbrus @zsxwing could you take a look? Thanks!

zsxwing:
@lw-lin thanks for doing this. However, #15949 also includes the fix for this issue.

lw-lin:
Ah, I'll keep an eye on that! @zsxwing thanks for the notification.

zsxwing:
@lw-lin could you close this one please? Thanks!

lw-lin:
sure! closing this.

PR #15949, opened 2016-11-20
[SPARK-18339] [SPARK-18513] [SQL] Don't push down current_timestamp for filters in StructuredStreaming and persist batch and watermark timestamps to offset log.

ghost:

For the following workflow:
1. I have a column called time which is at minute level precision in a Streaming DataFrame
2. I want to perform groupBy time, count
3. Then I want my MemorySink to only have the last 30 minutes of counts and I perform this by
.where('time >= current_timestamp().cast("long") - 30 * 60)
what happens is that the `filter` gets pushed down before the aggregation, and the filter happens on the source data for the aggregation instead of the result of the aggregation (where I actually want to filter).
I guess the main issue here is that `current_timestamp` is non-deterministic in the streaming context and shouldn't be pushed down the filter.
Does this require us to store the `current_timestamp` for each trigger of the streaming job, that is something to discuss.

Furthermore, we want to persist current batch timestamp and watermark timestamp to the offset log so that these values are consistent across multiple executions of the same batch.

@brkyvz @zsxwing @tdas 

## How was this patch tested?

A test was added to StreamingAggregationSuite ensuring the above use case is handled. The test injects a stream of time values (in seconds) to a query that runs in complete mode and only outputs the (count) aggregation results for the past 10 seconds.


rxin:
@tcondie can you use all upper case for the JIRA ticket in the future, e.g. SPARK-18339?

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamExecution.scala
```diff
@@ -422,6 +432,7 @@ class StreamExecution(
     val replacementMap = AttributeMap(replacements)
     val triggerLogicalPlan = withNewSources transformAllExpressions {
       case a: Attribute if replacementMap.contains(a) => replacementMap(a)
+      case t: CurrentTimestamp => CurrentBatchTimestamp(currentBatchTimestamp)
```
hvanhovell: Maybe an incredibly stupid comment: Why not use a Literal instead of this dedicated expression? You are going to replace it with a Literal anyway.

tcondie: Good question. The purpose of the dedicated expression is to avoid having it pushed down below a stateful operator (e.g., aggregation). during optimization.

hvanhovell: Why would that be a problem? I thought the problem was that the `current_timestamp` was set to the current system time, instead of the timestamp of the `current` batch.

brkyvz: @hvanhovell Timestamp is "non-deterministic" when it comes to streaming aggregations. Therefore it shouldn't be pushed down stateful aggregations. Otherwise the resulting table you get is different than what you asked for.
Example: 
```
df.groupBy('time).agg(count("*")).where('time >= current_timestamp().cast("long") - 30 * 60)
```
What you get:
All aggregates from all of time, but the aggregates older than 30 minutes don't get updated. Those get filtered
What I wanted:
Just return me the last 30 minutes of aggregates. I don't care about anything else older.

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamExecution.scala
```diff
@@ -95,6 +95,9 @@ class StreamExecution(
   /** The current eventTime watermark, used to bound the lateness of data that will processed. */
   private var currentEventTimeWatermark: Long = 0
 
+  /** The current batch processing timestamp */
+  private var currentBatchTimestamp: Long = 0
```
brkyvz: A single thread will change these values. There doesn't need to be locking AFAIK

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamExecution.scala
hvanhovell: Nevermind, I just read the JIRA, and I think I understand what is going on.

To me the `time` column in `df.groupBy('time)` is different from the `time` column which is  (apparently) in the source. I personally think it is better to have the statefull aggregate produce a new time column, then to restrict push down.

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/IncrementalExecution.scala
```diff
@@ -50,6 +52,19 @@ class IncrementalExecution(
       stateStrategy)
 
   /**
+   * See [SPARK-18339]
+   * Walk the optimized logical plan and replace CurrentBatchTimestamp
+   * with the desired literal
+   */
+  override lazy val optimizedPlan: LogicalPlan = {
+    sparkSession.sessionState.optimizer.execute(withCachedData) transformAllExpressions {
```
zsxwing: Any reason to not define this in `org.apache.spark.sql.execution.SparkStrategies.StreamingRelationStrategy`?

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamExecution.scala
```diff
@@ -344,8 +351,11 @@ class StreamExecution(
       }
     }
     if (hasNewData) {
+      // Current batch timestamp in seconds
+      currentBatchTimestamp = triggerClock.getTimeMillis() / 1000L
       reportTimeTaken(OFFSET_WAL_WRITE_LATENCY) {
-        assert(offsetLog.add(currentBatchId, availableOffsets.toOffsetSeq(sources)),
+        assert(offsetLog.add(currentBatchId,
+          availableOffsets.toOffsetSeq(sources, Some(currentBatchTimestamp.toString))),
```
zsxwing: It's better to have a new class like `StreamExecutionMetadata` to store all fields that need to be checkpointed, e.g., `currentBatchTimestamp`, `currentEventTimeWatermark`, and others that we may add in future.

zsxwing (changes requested):
My major comment is it's better to define a new class to store fields that need to be checkpointed. Otherwise, looks pretty good.

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/IncrementalExecution.scala
tcondie: Would doing so risk having the optimizer push the filter on literal below the stateful aggregation?

zsxwing: Never mind. Since we cannot convert expressions in StreamingRelationStrategy, it's fine to add codes here.

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamExecution.scala
```diff
@@ -734,6 +766,13 @@ class StreamExecution(
   case object TERMINATED extends State
 }
 
+object StreamExecutionMetadata {
+  private implicit val formats = Serialization.formats(NoTypeHints)
+
+  def apply(json: String): StreamExecutionMetadata =
```
zsxwing: Could you add some unit tests for this? E.g.,
```
    assert(StreamExecutionMetadata(0, 0) ===
      StreamExecutionMetadata("""{}"""))
    assert(StreamExecutionMetadata(1, 0) ===
      StreamExecutionMetadata("""{"currentEventTimeWatermark":1}"""))
    assert(StreamExecutionMetadata(0, 2) ===
      StreamExecutionMetadata("""{"currentBatchTimestamp":2}"""))
    assert(StreamExecutionMetadata(1, 2) ===
      StreamExecutionMetadata("""{"currentEventTimeWatermark":1,"currentBatchTimestamp":2}"""))
```
Then it's easy for us to find any incompatible changes in future.

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamExecution.scala
```diff
@@ -288,7 +310,11 @@ class StreamExecution(
         logInfo(s"Resuming streaming query, starting with batch $batchId")
         currentBatchId = batchId
         availableOffsets = nextOffsets.toStreamProgress(sources)
-        logDebug(s"Found possibly uncommitted offsets $availableOffsets")
+        streamExecutionMetadata = StreamExecutionMetadata(nextOffsets.metadata.getOrElse(
+          throw new IllegalStateException("OffsetLog does not contain current batch timestamp!")
```
zsxwing: Instead of throwing an exception, we can just use `{}` for backward compatibility.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/datetimeExpressions.scala
```diff
@@ -72,6 +72,28 @@ case class CurrentTimestamp() extends LeafExpression with CodegenFallback {
 }
 
 /**
+ * Expression representing the current batch time, which is used by StreamExecution to
+ * 1. prevent optimizer from pushing this expression below a stateful operator
+ * 2. allow IncrementalExecution to substitute this expression with a Literal(timestamp)
+ *
+ * There is no code generation since this expression should be replaced with a literal.
+ */
+@ExpressionDescription(
```
brkyvz: nit: This class won't be user facing, won't have SQL APIs, etc, therefore we can remove this I think.

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamExecution.scala
```diff
@@ -422,6 +451,8 @@ class StreamExecution(
     val replacementMap = AttributeMap(replacements)
     val triggerLogicalPlan = withNewSources transformAllExpressions {
       case a: Attribute if replacementMap.contains(a) => replacementMap(a)
+      case t: CurrentTimestamp =>
```
brkyvz: Shall we also support `CurrentDate`?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/datetimeExpressions.scala
```diff
@@ -72,6 +72,28 @@ case class CurrentTimestamp() extends LeafExpression with CodegenFallback {
...
+  usage = "_FUNC_() - Returns the current timestamp at the start of batch evaluation.")
+case class CurrentBatchTimestamp(timestamp: Long) extends LeafExpression
```
zsxwing: nit: `Long` -> `SQLTimestamp`

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamExecution.scala
```diff
@@ -38,6 +40,26 @@ import org.apache.spark.sql.streaming._
...
+ * persisted to the offset log via the OffsetSeq metadata field. Current
+ * information contained in this object includes:
+ *
+ * 1. currentEventTimeWatermark: The current eventTime watermark, used to
+ * bound the lateness of data that will processed.
+ * 2. currentBatchTimestamp: The current batch processing timestamp
+ */
+case class StreamExecutionMetadata(
+    var currentEventTimeWatermark: Long = 0,
+    var currentBatchTimestamp: Long = 0) {
```
brkyvz: Shall we name this as `currentBatchTimestampSecs` to clarify that this is in seconds?

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamExecution.scala
```diff
@@ -38,6 +40,26 @@ import org.apache.spark.sql.streaming._
...
+ * Contains metadata associated with a stream execution. This information is
+ * persisted to the offset log via the OffsetSeq metadata field. Current
+ * information contained in this object includes:
+ *
+ * 1. currentEventTimeWatermark: The current eventTime watermark, used to
+ * bound the lateness of data that will processed.
+ * 2. currentBatchTimestamp: The current batch processing timestamp
+ */
+case class StreamExecutionMetadata(
+    var currentEventTimeWatermark: Long = 0,
```
zsxwing: nit: Could you document the time units for these two fields?

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamExecution.scala
tcondie: Ryan indicated we use SQLTimestamp to indicate microseconds. I changed the name of the watermark to currentEventTimeWatermarkMillis based on your suggestion.

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamExecution.scala
```diff
@@ -344,8 +370,11 @@ class StreamExecution(
       }
     }
     if (hasNewData) {
+      // Current batch timestamp in seconds
```
brkyvz: will need to update the comment to `microseconds`

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/datetimeExpressions.scala
```diff
@@ -72,6 +72,26 @@ case class CurrentTimestamp() extends LeafExpression with CodegenFallback {
 }
 
 /**
+ * Expression representing the current batch time, which is used by StreamExecution to
+ * 1. prevent optimizer from pushing this expression below a stateful operator
+ * 2. allow IncrementalExecution to substitute this expression with a Literal(timestamp)
+ *
+ * There is no code generation since this expression should be replaced with a literal.
+ */
+case class CurrentBatchTimestamp(timestamp: SQLTimestamp) extends LeafExpression
```
brkyvz: since this class will never be evaluated, but rewritten we can also extend `Unevaluable`

brkyvz: Feel free to push back on this...
Since we're going to support both `CurrentTimestamp` and `CurrentDate`, I would recommend the following signature:
```scala
case class CurrentBatchTimestamp(timestampMs: Long, dataType: DataType) {

  def toLiteral: Literal = dataType match {
    case _: TimestampType => Literal(timestampMs * 1000, TimestampType)
    case _: DateType => Literal(DateTimeUtils.millisToDays(timestampMs), DateType)
  }
```

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/IncrementalExecution.scala
```diff
@@ -50,6 +52,19 @@ class IncrementalExecution(
...
+      case CurrentBatchTimestamp(timestamp) =>
```
brkyvz: then following my comment above on the `CurrentBatchTimestamp` class, I would rewrite this as:
```scala
case ts @ CurrentBatchTimestamp(timestamp) =>
  logInfo(s"Current batch timestamp = $timestamp")
  ts.toLiteral
```

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamExecution.scala
brkyvz: or I would actually prefer using milliseconds all the time

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamExecution.scala
```diff
@@ -25,19 +25,43 @@ import scala.collection.mutable.ArrayBuffer
...
+ * information contained in this object includes:
+ *
+ * 1. currentEventTimeWatermark: The current eventTime watermark, used to
+ * bound the lateness of data that will processed. Time unit: milliseconds
+ * 2. currentBatchTimestamp: The current batch processing timestamp.
+ * Time unit: microseconds
+ */
+case class StreamExecutionMetadata(
+    var currentEventTimeWatermarkMillis: Long = 0,
+    var currentBatchTimestamp: SQLTimestamp = 0) {
```
brkyvz: @zsxwing Since we're going to handle both `currentDate` and `currentTimestamp` I think milliseconds is better here

brkyvz:
Left several comments that all are related to each other. @zsxwing I would like your feedback on those as well in order to not make @tcondie make too many changes flipping timestamp precision.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/datetimeExpressions.scala
```diff
@@ -72,6 +71,30 @@ case class CurrentTimestamp() extends LeafExpression with CodegenFallback {
...
+case class CurrentBatchTimestamp(timestampMs: Long, dataType: DataType)
+  extends LeafExpression with Nondeterministic with CodegenFallback {
+
+  override def nullable: Boolean = false
+
+  override def prettyName: String = "current_batch_timestamp"
+
+  override protected def initializeInternal(partitionIndex: Int): Unit = {}
+
+  override protected def evalInternal(input: InternalRow): Any = timestampMs
```
zsxwing: Could you throw an exception like `throw new UnsupportedOperationException(s"Cannot evaluate expression: $this")` here?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/datetimeExpressions.scala
```diff
@@ -72,6 +71,30 @@ case class CurrentTimestamp() extends LeafExpression with CodegenFallback {
...
+
+  def toLiteral: Literal = dataType match {
+    case _: TimestampType => Literal(timestampMs * 1000L, TimestampType)
```
zsxwing: nit: Use `Literal(DateTimeUtils.fromJavaTimestamp(new java.sql.Timestamp(timestampMs)), TimestampType)` instead. This is not a big deal anyway since `timestampMs` is always positive here.

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamExecution.scala
```diff
@@ -25,19 +25,43 @@ import scala.collection.mutable.ArrayBuffer
...
 import org.apache.spark.util.{Clock, UninterruptibleThread, Utils}
 
 /**
+ * Contains metadata associated with a stream execution. This information is
+ * persisted to the offset log via the OffsetSeq metadata field. Current
+ * information contained in this object includes:
+ *
+ * 1. currentEventTimeWatermark: The current eventTime watermark, used to
+ * bound the lateness of data that will processed. Time unit: milliseconds
+ * 2. currentBatchTimestamp: The current batch processing timestamp.
```
zsxwing: nit: Use Scala doc: `@param currentEventTimeWatermarkMillis  ...`

zsxwing (changes requested):
@tcondie could you add `[SPARK-18513]` to the title since you are also resolving this issue in this PR?

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamExecution.scala
```diff
@@ -92,8 +116,8 @@ class StreamExecution(
   /** The current batchId or -1 if execution has not yet been initialized. */
   private var currentBatchId: Long = -1
 
-  /** The current eventTime watermark, used to bound the lateness of data that will processed. */
-  private var currentEventTimeWatermark: Long = 0
+  /** stream execution metadata */
```
tdas: nit: s -> S

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamExecution.scala
```diff
@@ -422,6 +450,12 @@ class StreamExecution(
     val replacementMap = AttributeMap(replacements)
     val triggerLogicalPlan = withNewSources transformAllExpressions {
       case a: Attribute if replacementMap.contains(a) => replacementMap(a)
+      case ct: CurrentTimestamp =>
+        CurrentBatchTimestamp(streamExecutionMetadata.currentBatchTimestampMillis,
+          ct.dataType)
+      case cd: CurrentDate =>
+        CurrentBatchTimestamp(streamExecutionMetadata.currentBatchTimestampMillis,
```
tdas: Are you sure this is right? From what I see in tests about how CurrentDate is expected to be converted to literal, the millis should be exactly rounded to the day boundary. 

https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/datetimeExpressions.scala#L47

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamExecution.scala
```diff
@@ -38,6 +40,27 @@ import org.apache.spark.sql.streaming._
...
+ * persisted to the offset log via the OffsetSeq metadata field. Current
+ * information contained in this object includes:
+ *
+ * @param currentEventTimeWatermarkMillis: The current eventTime watermark, used to
+ * bound the lateness of data that will processed. Time unit: milliseconds
+ * @param currentBatchTimestampMillis: The current batch processing timestamp.
+ * Time unit: milliseconds
+ */
+case class StreamExecutionMetadata(
+    var currentEventTimeWatermarkMillis: Long = 0,
```
tdas: nit: this is pretty long! how about just `batchWatermarkMs`, and `batchTimestampMs`. Its not really current if a batch is being reprocessed.

sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamingAggregationSuite.scala
```diff
@@ -235,4 +236,49 @@ class StreamingAggregationSuite extends StreamTest with BeforeAndAfterAll {
...
+        .as[(Long, Long)]
+        .where('value >= current_timestamp().cast("long") - 10L)
+
+    testStream(aggregated, Complete)(
+      StartStream(ProcessingTime("10 seconds"), triggerClock = clock),
+
+      // advance clock to 10 seconds
+      AddData(inputData, 0L, 5L, 5L, 10L),
+      AdvanceManualClock(10 * 1000),
+      AssertOnQuery { _ => clock.getTimeMillis() === 10 * 1000 },
```
tdas: nit: unnecessary, AdvanceManualClock already tests this internally

sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamingAggregationSuite.scala
```diff
@@ -235,4 +236,49 @@ class StreamingAggregationSuite extends StreamTest with BeforeAndAfterAll {
...
+
+      // advance clock to 30 seconds
+      AddData(inputData, 0L),
+      AdvanceManualClock(10 * 1000),
+      CheckLastBatch((20L, 1)),
+
+      // advance clock to 40 seconds
+      AddData(inputData, 25L, 30L, 40L, 45L),
+      AdvanceManualClock(10 * 1000),
+      CheckLastBatch((30L, 1), (40L, 1), (45L, 1))
```
tdas: I dont see how this tests whether the watermark, and timestamp is recovered correctly from log. nothing is being stopped.

I think there should be another test that actually stops the stream, drops the complete mode table and restarts it from checkpoint, and see if the table gets recreated to the same value. 

something like 
```
val query = inputData.toDF().... queryName("aggregates").option("checkpointLocation", cpPath).start()
memoryStream.add(....)
query.processAllAvailable()
checkAnswer(table("aggregates"), expectedResults)
query.stop()

spark.drop("aggregates")
clock.add(1000000)

val query2 = inputData.toDF().... queryName("aggregates").option("checkpointLocation", cpPath).start()
query.processAllAvailable()
checkAnswer(table("aggregates"), expectedResults)  // verify that increasing the clock while query was down does not change result of the last batch

```
See this for more inspiration - https://github.com/apache/spark/blob/master/sql/core/src/test/scala/org/apache/spark/sql/streaming/test/DataStreamReaderWriterSuite.scala#L472

sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamingAggregationSuite.scala
```diff
@@ -235,4 +236,49 @@ class StreamingAggregationSuite extends StreamTest with BeforeAndAfterAll {
...
+    testStream(aggregated, Complete)(
+      StartStream(ProcessingTime("10 seconds"), triggerClock = clock),
+
+      // advance clock to 10 seconds
+      AddData(inputData, 0L, 5L, 5L, 10L),
+      AdvanceManualClock(10 * 1000),
+      AssertOnQuery { _ => clock.getTimeMillis() === 10 * 1000 },
+      CheckLastBatch((0L, 1), (5L, 2), (10L, 1)),
+
+      // advance clock to 20 seconds
```
tdas: nice to add `// advance clock to 20 seconds, should retain only keys >= 20`

tdas (changes requested):
Two major changes needed. 
- I think CurrentDate is wrong. millisecond should be converted to day boundaries.
- no test for recovery of watermark and timestamp from log.

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamExecution.scala
tdas: NVM!!!! I MISSED THE conversion inside `CurrentBatchTimeStamp.toLiteral`. MY BAD!!

But then there are not tests for current_date! So one of the branch in the above code is not being tested at all.

tdas:
Could you update the title and description to reflect that you are solving two issues related to watermark and current timestamp

lw-lin:
Hi @tcondie, just want to provide some information regarding watermark recovery:

After some investigation, I found that even if we do have correctly recovered the watermark from log, the test `WatermarkSuite#test("recovery")` still fails. It's due to a sink might just skip the real execution of a re-submitted batch, thus the watermark would not proceed correctly. I've opened [SPARK-18552](https://issues.apache.org/jira/browse/SPARK-18552) for this.

As a provisional fix to pass `WatermarkSuite#test("recovery")` and verify we can recover watermark from logs correctly, we can just do this to `MemorySink`:
```
if {
   ...
 } else {
  data.collect() // this provisional fix forces the execution of a re-submitted batch
  logDebug(s"Skipping already committed batch: $batchId")	
}
```

Hope this helps!

lw-lin:
Also cc @marmbrus @tdas @zsxwing; please be aware of [SPARK-18552](https://issues.apache.org/jira/browse/SPARK-18552), thanks.

sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamExecutionMetadataSuite.scala
```diff
@@ -0,0 +1,96 @@
...
+    import testImplicits._
+    val clock = new SystemClock()
+    val ms = new MemoryStream[Long](0, sqlContext)
+    val df = ms.toDF().toDF("a")
+    val checkpointLoc = newMetadataDir
+    val checkpointDir = new File(checkpointLoc, "complete")
+    checkpointDir.mkdirs()
+    assert(checkpointDir.exists())
+    val tableName = "test"
+    def startQuery: StreamingQuery = {
```
tdas: add some docs saying why this query was chosen

sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamExecutionMetadataSuite.scala
```diff
@@ -0,0 +1,96 @@
...
+    assert(StreamExecutionMetadata(1, 0) ===
+      StreamExecutionMetadata("""{"batchWatermarkMs":1}"""))
+    assert(StreamExecutionMetadata(0, 2) ===
+      StreamExecutionMetadata("""{"batchTimestampMs":2}"""))
+    assert(StreamExecutionMetadata(1, 2) ===
+      StreamExecutionMetadata(
+        """{"batchWatermarkMs":1,"batchTimestampMs":2}"""))
+  }
+
+  test("ensure consistent results across batch executions") {
```
tdas: can you be more specific about what this test. e.g. "metadata is recovered from log when query is restarted"

sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamExecutionMetadataSuite.scala
```diff
@@ -0,0 +1,96 @@
...
+        .where('a >= current_timestamp().cast("long"))
+        .writeStream
+        .format("memory")
+        .queryName(tableName)
+        .option("checkpointLocation", checkpointLoc)
+        .outputMode("complete")
+        .start()
+    }
+    // no exception here
+    val t1 = clock.getTimeMillis() + 5000L
```
tdas: increase the time added. this assumes that the query will be done before 5 to 10 seconds. that may not be true. probably add a minute.

tdas (commented):
Some comments

sql/core/src/test/scala/org/apache/spark/sql/streaming/WatermarkSuite.scala
```diff
@@ -96,28 +96,58 @@ class WatermarkSuite extends StreamTest with BeforeAndAfter with Logging {
...
+    def startQuery: StreamingQuery = {
+      ms.toDF()
         .withColumn("eventTime", $"value".cast("timestamp"))
         .withWatermark("eventTime", "10 seconds")
         .groupBy(window($"eventTime", "5 seconds") as 'window)
         .agg(count("*") as 'count)
         .select($"window".getField("start").cast("long").as[Long], $"count".as[Long])
+        .writeStream
+        .format("memory")
+        .queryName(tableName)
```
tdas: i think this is not testing it correctly! you are not even setting checkpoint directory. so offset log is not recovered.

sql/core/src/test/scala/org/apache/spark/sql/streaming/WatermarkSuite.scala
```diff
@@ -96,28 +96,58 @@ class WatermarkSuite extends StreamTest with BeforeAndAfter with Logging {
...
+    q = startQuery
+    q.processAllAvailable()
+    checkAnswer(
+      spark.table(tableName), Seq()
+    )
+
+    // Evict items less than previous watermark
+    ms.addData(25)
+    q.processAllAvailable()
+    checkAnswer(
```
tdas: why split into 3 lines?

sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamingAggregationSuite.scala
```diff
@@ -235,4 +239,85 @@ class StreamingAggregationSuite extends StreamTest with BeforeAndAfterAll {
...
+    val tz = TimeZone.getDefault.getID
+    val inputData = MemoryStream[Long]
+    val aggregated =
+      inputData.toDF()
+        .select(to_utc_timestamp(from_unixtime('value * DateTimeUtils.SECONDS_PER_DAY), tz))
+        .toDF("value")
+        .groupBy($"value")
+        .agg(count("*"))
+        // .select('value, date_sub(current_date(), 10).cast("timestamp").alias("t"))
+        // .select('value, 't, 'value >= 't)
```
tdas: please remove these lines

sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamingAggregationSuite.scala
```diff
@@ -235,4 +239,85 @@ class StreamingAggregationSuite extends StreamTest with BeforeAndAfterAll {
...
+      CheckLastBatch((0L, 1), (5L, 2), (10L, 1)),
+      // advance clock to 20 days, should retain keys >= 10
+      AddData(inputData, 15L, 15L, 20L),
+      AdvanceManualClock(DateTimeUtils.MILLIS_PER_DAY * 10),
+      CheckLastBatch((10L, 1), (15L, 2), (20L, 1)),
+      // advance clock to 30 days, should retain keys >= 20
+      AddData(inputData, 0L),
+      AdvanceManualClock(DateTimeUtils.MILLIS_PER_DAY * 10),
+      CheckLastBatch((20L, 1)),
+      // advance clock to 40 seconds, should retain keys >= 30
```
tdas: 40 seconds -> days

sql/core/src/test/scala/org/apache/spark/sql/streaming/WatermarkSuite.scala
```diff
@@ -96,27 +96,42 @@ class WatermarkSuite extends StreamTest with BeforeAndAfter with Logging {
...
+      .withColumn("eventTime", $"value".cast("timestamp"))
+      .withWatermark("eventTime", "10 seconds")
+      .groupBy(window($"eventTime", "5 seconds") as 'window)
+      .agg(count("*") as 'count)
+      .select($"window".getField("start").cast("long").as[Long], $"count".as[Long])
+    val outputMode = OutputMode.Append
+    val memorySink = new MemorySink(df.schema, outputMode)
+    testStream(df)(
       AddData(inputData, 10, 11, 12, 13, 14, 15),
       CheckAnswer(),
```
tdas: nit: Make this CheckAnswer -> CheckLastBatch for being consistent with rest of the checks in this test

sql/core/src/test/scala/org/apache/spark/sql/streaming/WatermarkSuite.scala
```diff
@@ -96,27 +96,42 @@ class WatermarkSuite extends StreamTest with BeforeAndAfter with Logging {
...
       AddData(inputData, 25), // Evict items less than previous watermark.
+      CheckLastBatch((10, 5)),
       StopStream,
+      AssertOnQuery { q => // clear the sink
+        q.sink.asInstanceOf[MemorySink].clear()
+        true
+      },
       StartStream(),
-      CheckAnswer((10, 5))
+      CheckLastBatch((10, 5)),
```
tdas: nit: add comment to explain  // Should recompute the last batch and re-evict timestamp 10

sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamingAggregationSuite.scala
```diff
@@ -235,4 +239,85 @@ class StreamingAggregationSuite extends StreamTest with BeforeAndAfterAll {
...
+      AdvanceManualClock(10 * 1000),
+      CheckLastBatch((20L, 1)),
+
+      // advance clock to 40 seconds, should retain keys >= 30
+      AddData(inputData, 25L, 30L, 40L, 45L),
+      AdvanceManualClock(10 * 1000),
+      CheckLastBatch((30L, 1), (40L, 1), (45L, 1))
+    )
+  }
+
```
tdas: nit: extra line

sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamExecutionMetadataSuite.scala
```diff
@@ -0,0 +1,99 @@
...
+    intercept[AnalysisException](spark.table(tableName).collect())
+    val q2 = startQuery
+    q2.processAllAvailable()
+    checkAnswer(
+      spark.table(tableName),
+      Seq(Row(t1, 1), Row(t2, 1))
+    )
+
+    q2.stop()
+
```
tdas: nit: extra line.

sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamExecutionMetadataSuite.scala
```diff
@@ -0,0 +1,99 @@
...
+    val df = ms.toDF().toDF("a")
+    val checkpointLoc = newMetadataDir
+    val checkpointDir = new File(checkpointLoc, "complete")
+    checkpointDir.mkdirs()
+    assert(checkpointDir.exists())
+    val tableName = "test"
+    // Query that prunes timestamps less than current_timestamp, making
+    // it easy to use for ensuring that a batch is re-processed with the
+    // timestamp used when it was first processed.
+    def startQuery: StreamingQuery = {
```
tdas: nit: functions that have sideeffects (like starting a thread), usually have `()` at the end, and is used with the `()`. for example, `val q = startQuery()`

sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamExecutionMetadataSuite.scala
```diff
@@ -0,0 +1,99 @@
...
+        .count()
+        .where('a >= current_timestamp().cast("long"))
+        .writeStream
+        .format("memory")
+        .queryName(tableName)
+        .option("checkpointLocation", checkpointLoc)
+        .outputMode("complete")
+        .start()
+    }
+    // no exception here
```
tdas: what does this comment mean?

sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamExecutionMetadataSuite.scala
```diff
@@ -0,0 +1,99 @@
...
+    val t1 = clock.getTimeMillis() + 60L * 1000L
+    val t2 = clock.getTimeMillis() + 60L * 1000L + 1000L
```
tdas: add a comment explaining how the test works.

sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamExecutionMetadataSuite.scala
```diff
@@ -0,0 +1,99 @@
...
+    ms.addData(t1, t2)
+    q.processAllAvailable()
+
+    checkAnswer(
+      spark.table(tableName),
+      Seq(Row(t1, 1), Row(t2, 1))
+    )
+
+    q.stop()
+    Thread.sleep(60L * 1000L + 5000L) // Expire t1 and t2
```
tdas: This test will now takes 60 seconds! I think I didnt quite understand the test earlier, but now I do. I think the earlier 5 second was closer to being fine. Okay, lets just use 10 seconds. And instead of sleep, use `eventually` to check the conditions `t2 < clock.getTimeMillis()`. this would make the system sleep no more than that is necessary.

sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamExecutionMetadataSuite.scala
```diff
@@ -0,0 +1,99 @@
...
+    assert(t1 < clock.getTimeMillis())
+    assert(t2 < clock.getTimeMillis())
+
+    spark.sql(s"drop table $tableName")
+
+    // verify table is dropped
+    intercept[AnalysisException](spark.table(tableName).collect())
```
tdas: i think you can use `spark.catalog,tableExists`

tdas (changes requested):
Almost LGTM. Address the comments on the tests.

sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamExecutionMetadataSuite.scala
zsxwing: Why not use `StreamManualClock` to avoid the indeterministic system clock.

tcondie: @zsxwing I don't see an obvious way to pass a StreamManualClock in DataStreamWriter.start(). Should I be taking an entirely different approach?

sql/core/src/test/scala/org/apache/spark/sql/streaming/WatermarkSuite.scala
```diff
@@ -96,27 +96,42 @@ class WatermarkSuite extends StreamTest with BeforeAndAfter with Logging {
...
-        .select($"window".getField("start").cast("long").as[Long], $"count".as[Long])
-
-    testStream(windowedAggregation)(
+    val df = inputData.toDF()
+      .withColumn("eventTime", $"value".cast("timestamp"))
+      .withWatermark("eventTime", "10 seconds")
+      .groupBy(window($"eventTime", "5 seconds") as 'window)
+      .agg(count("*") as 'count)
+      .select($"window".getField("start").cast("long").as[Long], $"count".as[Long])
+    val outputMode = OutputMode.Append
```
zsxwing: `outputMode` and `memorySink` are not used.

brkyvz:
The implementation LGTM! +1 to @tdas 's comments for tests

sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamingAggregationSuite.scala
```diff
@@ -235,4 +239,85 @@ class StreamingAggregationSuite extends StreamTest with BeforeAndAfterAll {
...
+      CheckLastBatch((20L, 1)),
+      // advance clock to 40 seconds, should retain keys >= 30
+      AddData(inputData, 25L, 30L, 40L, 45L),
+      AdvanceManualClock(DateTimeUtils.MILLIS_PER_DAY * 10),
+      CheckLastBatch((30L, 1), (40L, 1), (45L, 1))
+    )
+  }
+}
+
+object StreamingAggregationSuite {
```
zsxwing: You can remove this object. This is not used in any closure.

sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamExecutionMetadataSuite.scala
tcondie: I should also say that I'm not too concerned by the indeterministic system clock issues since the batchTimestamp is recorded prior to running the query. Therefore as long as the query gets planned within 10 seconds, we're good.

tdas:
LGTM.

sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamExecutionMetadataSuite.scala
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+package org.apache.spark.sql.streaming
+
+import org.apache.spark.sql.execution.streaming.StreamExecutionMetadata
+import org.apache.spark.util.Utils
+
+class StreamExecutionMetadataSuite extends StreamTest {
+
+  private def newMetadataDir =
```
zsxwing: nit: not used.

zsxwing:
retest this please

tdas:
retest this please.

tdas:
Merging this to master and 2.1

PR #15951, opened 2016-11-20
[SPARK-18510] Fix data corruption from inferred partition column dataTypes

brkyvz:
### The Issue

If I specify my schema when doing
```scala
spark.read
  .schema(someSchemaWherePartitionColumnsAreStrings)
```
but if the partition inference can infer it as IntegerType or I assume LongType or DoubleType (basically fixed size types), then once UnsafeRows are generated, your data will be corrupted.

### Proposed solution

The partition handling code path is kind of a mess. In my fix I'm probably adding to the mess, but at least trying to standardize the code path.

The real issue is that a user that uses the `spark.read` code path can never clearly specify what the partition columns are. If you try to specify the fields in `schema`, we practically ignore what the user provides, and fall back to our inferred data types. What happens in the end is data corruption.

My solution tries to fix this by always trying to infer partition columns the first time you specify the table. Once we find what the partition columns are, we try to find them in the user specified schema and use the dataType provided there, or fall back to the smallest common data type.

We will ALWAYS append partition columns to the user's schema, even if they didn't ask for it. We will only use the data type they provided if they specified it. While this is confusing, this has been the behavior since Spark 1.6, and I didn't want to change this behavior in the QA period of Spark 2.1. We may revisit this decision later.

A side effect of this PR is that we won't need https://github.com/apache/spark/pull/15942 if this PR goes in.

## How was this patch tested?

Regression tests

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
@@ -61,8 +61,12 @@ import org.apache.spark.util.Utils
  *              qualified. This option only works when reading from a [[FileFormat]].
  * @param userSpecifiedSchema An optional specification of the schema of the data. When present
  *                            we skip attempting to infer the schema.
- * @param partitionColumns A list of column names that the relation is partitioned by. When this
- *                         list is empty, the relation is unpartitioned.
+ * @param partitionColumns A list of column names that the relation is partitioned by. This list is
+ *                         generally empty during the read path, unless this DataSource is managed
+ *                         by Hive. In these cases, during `resolveRelation`, we will call
+ *                         `getOrInferFileFormatSchema` for file based DataSources to infer the
+ *                         partitioning. In other cases, if this list is empty, then this table
+ *                         is unpartitioned.
  * @param bucketSpec An optional specification for bucketing (hash-partitioning) of the data.
  * @param catalogTable Optional catalog table reference that can be used to push down operations
  *                     over the datasource to the catalog service.
@@ -357,9 +361,10 @@ case class DataSource(
       // This is a non-streaming file based datasource.
       case (format: FileFormat, _) =>
         val allPaths = caseInsensitiveOptions.get("path") ++ paths
+        val hadoopConf = sparkSession.sessionState.newHadoopConf()
         val globbedPaths = allPaths.flatMap { path =>
           val hdfsPath = new Path(path)
-          val fs = hdfsPath.getFileSystem(sparkSession.sessionState.newHadoopConf())
+          val fs = hdfsPath.getFileSystem(hadoopConf)
           val qualified = hdfsPath.makeQualified(fs.getUri, fs.getWorkingDirectory)
           val globPath = SparkHadoopUtil.get.globPathIfNecessary(qualified)
 
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
@@ -84,9 +84,32 @@ case class DataSource(
   private val caseInsensitiveOptions = new CaseInsensitiveMap(options)
 
   /**
-   * Infer the schema of the given FileFormat, returns a pair of schema and partition column names.
+   * Get the schema of the given FileFormat, if provided by `userSpecifiedSchema`, or try to infer
+   * it. In the read path, only Hive managed tables provide the partition columns properly when
+   * initializing this class. All other file based data sources will try to infer the partitioning,
+   * and then cast the inferred types to user specified dataTypes if the partition columns exist
+   * inside `userSpecifiedSchema`, otherwise we can hit data corruption bugs like SPARK-18510.
+   * This method will try to do the least amount of work given whether `userSpecifiedSchema` and
+   * `partitionColumns` are provided. Here are some code paths that use this method:
+   *   1. `spark.read` (no schema): Most amount of work. Infer both schema and partitioning columns
+   *   2. `spark.read.schema(userSpecifiedSchema)`: Parse partitioning columns, cast them to the
+   *     dataTypes provided in `userSpecifiedSchema` if they exist or fallback to inferred
+   *     dataType if they don't.
+   *   3. `spark.readStream.schema(userSpecifiedSchema)`: For streaming use cases, users have to
+   *     provide the schema. Here, we also perform partition inference like 2, and try to use
+   *     dataTypes in `userSpecifiedSchema`. All subsequent triggers for this stream will re-use
+   *     this information, therefore calls to this method should be very cheap, i.e. there won't
+   *     be any further inference in any triggers.
+   *   4. `df.saveAsTable(tableThatExisted)`: In this case, we call this method to resolve the
+   *     existing table's partitioning scheme. This is achieved by not providing
+   *     `userSpecifiedSchema`. For this case, we add the boolean `justPartitioning` for an early
+   *     exit, if we don't care about the schema of the original table.
+   *
+   * Returns a pair of schema and partition column names.
    */
-  private def inferFileFormatSchema(format: FileFormat): (StructType, Seq[String]) = {
+  private def getOrInferFileFormatSchema(
+      format: FileFormat,
+      justPartitioning: Boolean = false): (StructType, Seq[String]) = {
     lazy val tempFileCatalog = {
       val allPaths = caseInsensitiveOptions.get("path") ++ paths
       val globbedPaths = allPaths.toSeq.flatMap { path =>
@@ -128,6 +151,9 @@ case class DataSource(
         }
         StructType(partitionFields)
       }
+    }
+    if (justPartitioning) {
+      partitionSchema -> partitionSchema.map(_.name)
     }
     val tableSchema = userSpecifiedSchema.map { schema =>
       val equality = sparkSession.sessionState.conf.resolver
@@ -178,7 +204,7 @@ case class DataSource(
               "you may be able to create a static DataFrame on that directory with " +
               "'spark.read.load(directory)' and infer schema from it.")
         }
-        val (schema, partCols) = inferFileFormatSchema(format)
+        val (schema, partCols) = getOrInferFileFormatSchema(format)
         SourceInfo(s"FileSource[$path]", schema, partCols)
 
       case _ =>
@@ -269,9 +295,9 @@ case class DataSource(
    *                      1. We already know that the files exist, therefore we don't need to check
    *                    once again.
    *                      2. The user already specified the schema. We still fall back to
-   *                    `inferFileFormatSchema` to maintain similar semantics to batch data, but
-   *                    we won't have to perform any expensive operations there, because everything
-   *                    should already be provided
+   *                    `getOrInferFileFormatSchema` to maintain similar semantics to batch data,
+   *                    but we won't have to perform any expensive operations there, because
+   *                    everything should already be provided
    */
   def resolveRelation(isStreaming: Boolean = false): BaseRelation = {
     val relation = (providingClass.newInstance(), userSpecifiedSchema) match {
@@ -317,12 +343,6 @@ case class DataSource(
 
       // This is a non-streaming file based datasource.
       case (format: FileFormat, _) =>
-        val (schema, inferredPartitionColumns) = inferFileFormatSchema(format)
-        val dataSchema = if (isStreaming) {
-          schema
-        } else {
-          StructType(schema.dropRight(inferredPartitionColumns.length))
-        }
         val allPaths = caseInsensitiveOptions.get("path") ++ paths
         val globbedPaths = allPaths.flatMap { path =>
           val hdfsPath = new Path(path)
@@ -340,6 +360,13 @@ case class DataSource(
           }
           globPath
         }.toArray
+
+        val (schema, inferredPartitionColumns) = getOrInferFileFormatSchema(format)
+        val dataSchema = if (isStreaming) {
+          schema
+        } else {
+          StructType(schema.dropRight(inferredPartitionColumns.length))
+        }
 
         val partitionSchema = if (inferredPartitionColumns.isEmpty) {
           None
@@ -407,11 +434,7 @@ case class DataSource(
         // up.  If we fail to load the table for whatever reason, ignore the check.
         if (mode == SaveMode.Append) {
           val existingPartitionColumns = Try {
-            resolveRelation()
-              .asInstanceOf[HadoopFsRelation]
-              .partitionSchema
-              .fieldNames
-              .toSeq
+            getOrInferFileFormatSchema(format, justPartitioning = true)._2
           }.getOrElse(Seq.empty[String])
           // TODO: Case sensitivity.
           val sameColumns =
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
@@ -107,6 +107,8 @@ case class DataSource(
           partitionField)
       }
       StructType(resolved)
+    } else if (catalogTable.nonEmpty) {
+      catalogTable.get.partitionSchema
     } else {
       // in streaming mode, we have already inferred and registered partition columns, we will
       // never have to re-use this
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
@@ -105,11 +105,15 @@ case class DataSource(
    *     `userSpecifiedSchema`. For this case, we add the boolean `justPartitioning` for an early
    *     exit, if we don't care about the schema of the original table.
    *
-   * Returns a pair of schema and partition column names.
+   * Returns a pair of the data schema (excluding partition columns) and the schema of the partition
+   * columns.
+   *
+   * @param justPartitioning Whether to exit early and provide just the schema partitioning.
    */
   private def getOrInferFileFormatSchema(
       format: FileFormat,
-      justPartitioning: Boolean = false): (StructType, Seq[String]) = {
+      justPartitioning: Boolean = false): (StructType, StructType) = {
+    // the operations below are expensive therefore try not to do them if we don't need to
     lazy val tempFileCatalog = {
       val allPaths = caseInsensitiveOptions.get("path") ++ paths
       val globbedPaths = allPaths.toSeq.flatMap { path =>
@@ -153,13 +157,9 @@ case class DataSource(
       }
     }
     if (justPartitioning) {
-      return partitionSchema -> partitionSchema.map(_.name)
+      return new StructType() -> partitionSchema
     }
-    if (catalogTable.isDefined && userSpecifiedSchema.isDefined) {
-      // backwards compatibility before SPARK-18510. Return the schema of catalog tables as is
-      return userSpecifiedSchema.get -> partitionSchema.map(_.name)
-    }
-    val tableSchema = userSpecifiedSchema.map { schema =>
+    val dataSchema = userSpecifiedSchema.map { schema =>
       val equality = sparkSession.sessionState.conf.resolver
       StructType(schema.filterNot(f => partitionSchema.exists(p => equality(p.name, f.name))))
     }.orElse {
@@ -168,12 +168,10 @@ case class DataSource(
         caseInsensitiveOptions,
         tempFileCatalog.allFiles())
     }.getOrElse {
-      val exampleFiles = tempFileCatalog.allFiles().take(2).mkString(",")
       throw new AnalysisException(
-        s"Unable to infer schema for $format at $exampleFiles. It must be specified manually.")
+        s"Unable to infer schema for $format. It must be specified manually.")
     }
-    // always append partitions to the end
-    StructType(tableSchema ++ partitionSchema) -> partitionSchema.map(_.name)
+    dataSchema -> partitionSchema
   }
 
   /** Returns the name and schema of the source that can be used to continually read data. */
@@ -211,7 +209,7 @@ case class DataSource(
               "'spark.read.load(directory)' and infer schema from it.")
         }
         val (schema, partCols) = getOrInferFileFormatSchema(format)
-        SourceInfo(s"FileSource[$path]", schema, partCols)
+        SourceInfo(s"FileSource[$path]", StructType(schema ++ partCols), partCols.fieldNames)
 
       case _ =>
         throw new UnsupportedOperationException(
@@ -296,16 +294,13 @@ case class DataSource(
    * Create a resolved [[BaseRelation]] that can be used to read data from or write data into this
    * [[DataSource]]
    *
-   * @param isStreaming Whether a Streaming Source is asking us for resolution. In this case, we
-   *                    have some special cases where we don't want to duplicate work. These are:
-   *                      1. We already know that the files exist, therefore we don't need to check
-   *                    once again.
-   *                      2. The user already specified the schema. We still fall back to
-   *                    `getOrInferFileFormatSchema` to maintain similar semantics to batch data,
-   *                    but we won't have to perform any expensive operations there, because
-   *                    everything should already be provided
+   * @param checkFilesExist Whether to confirm that the files exist when generating the
+   *                        non-streaming file based datasource. StructuredStreaming jobs already
+   *                        list file existence, and when generating incremental jobs, the batch
+   *                        is considered as a non-streaming file based data source. Since we know
+   *                        that files already exist, we don't need to check them again.
    */
-  def resolveRelation(isStreaming: Boolean = false): BaseRelation = {
+  def resolveRelation(checkFilesExist: Boolean = true): BaseRelation = {
     val relation = (providingClass.newInstance(), userSpecifiedSchema) match {
       // TODO: Throw when too much is given.
       case (dataSource: SchemaRelationProvider, Some(schema)) =>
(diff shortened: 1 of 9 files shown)


brkyvz:
cc @rxin @marmbrus  Don't know who's the best person to look at this, but git blame sais I mainly changed your code :)

sql/core/src/test/scala/org/apache/spark/sql/test/DataFrameReaderWriterSuite.scala
```diff
@@ -573,4 +573,39 @@ class DataFrameReaderWriterSuite extends QueryTest with SharedSQLContext with Be
...
+    import testImplicits._
+    withTempDir { src =>
+      val createArray = udf { (length: Long) =>
+        for (i <- 1 to length.toInt) yield i.toString
+      }
+      spark.range(4).select(createArray('id + 1) as 'ex, 'id, 'id % 4 as 'part).coalesce(1).write
+        .partitionBy("part", "id")
+        .mode("overwrite")
+        .parquet(src.toString)
+      // make sure to specify the schema in the wrong order. Partition column in the middle, etc.
```
gatorsmile: Without the fix, the order of columns does not matter, right? As long as the types are right, it should work?

gatorsmile:
When users specify schemas, we do not want to infer the schemas due to its potentially expensive cost. Based on my understanding, data corruption issues are common when user-specified schemas do not provide correct types.

gatorsmile:
> The real issue is that a user that uses the spark.read code path can never clearly specify what the partition columns are. If you try to specify the fields in schema, we practically ignore what the user provides, and fall back to our inferred data types. What happens in the end is data corruption.

For data source tables, the partition columns are part of data schema. Users do not need to know which columns are used for partitioning. If they can provide the right types, they should be able to see the expected data. 

In the test cases, we can get the correct result with the following changes:
```
      spark.range(4).select(createArray('id + 1) as 'ex, 'id, 'id % 4 as 'part).coalesce(1)
      val schema = new StructType()
        .add("part", LongType)
        .add("ex", ArrayType(StringType))
        .add("id", LongType)
      spark.read
        .schema(schema)
        .format("parquet")
        .load(src.toString).show()
```

brkyvz:
True. But there's no reason "part" and "id" can't be strings right?

gatorsmile:
Your concern is right. I just did another try. Using the string-type columns as partition columns. See the following code:

```Scala
val rowRdd: RDD[Row] = sparkContext.parallelize(1 to 10).map(i => Row(i, i.toString))
val inputSchema = StructType(Seq(
  StructField("intCol", IntegerType),
  StructField("stringCol", StringType)
))
spark.createDataFrame(rowRdd, inputSchema)
  .write.partitionBy("stringCol").mode("overwrite").parquet(src.toString)
val schema = new StructType()
  .add("intCol", IntegerType)
  .add("stringCol", IntegerType)
spark.read
  .schema(schema)
  .format("parquet")
  .load(src.toString).show()
```
Users have to use `IntegerType` for these partition columns even if the original data type is `StringType`. This looks werid. Otherwise, they will hit the following error:
```
Job aborted due to stage failure: Task 0 in stage 1.0 failed 1 times, most recent failure: Lost task 0.0 in stage 1.0 (TID 2, localhost, executor driver): java.lang.NullPointerException
	at org.apache.spark.sql.execution.vectorized.OnHeapColumnVector.getArrayLength(OnHeapColumnVector.java:375)
```

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
```diff
@@ -316,27 +354,12 @@ case class DataSource(
...
-        }.getOrElse {
-          throw new AnalysisException(
-            s"Unable to infer schema for $format at ${allPaths.take(2).mkString(",")}. " +
-              "It must be specified manually")
+          new InMemoryFileIndex(sparkSession, globbedPaths, options, partitionSchema)
         }
 
         HadoopFsRelation(
           fileCatalog,
-          partitionSchema = fileCatalog.partitionSchema,
```
gatorsmile: This line causes the problem, right? We always ignore the user-specified column types and use the inferred partition columns.

sql/core/src/test/scala/org/apache/spark/sql/test/DataFrameReaderWriterSuite.scala
brkyvz: with the fix, it still does not matter. The comment is outdated, I thought something else was the problem. In terms of schema though, the output from Spark is always consistent, i.e. partition columns go last.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
brkyvz: exactly

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
```diff
@@ -84,30 +84,90 @@ case class DataSource(
...
+              s"$partitionColumn. It must be specified manually.")
+          }
+        }
+        StructType(partitionFields)
+      }
+    }
+    if (justPartitioning) {
+      partitionSchema -> partitionSchema.map(_.name)
+    }
+    val tableSchema = userSpecifiedSchema.map { schema =>
```
gatorsmile: Are you missing `else` here?

brkyvz: good catch. missed the `return`

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
```diff
@@ -84,30 +84,96 @@ case class DataSource(
...
+      }
+    }
+    if (justPartitioning) {
+      return partitionSchema -> partitionSchema.map(_.name)
+    }
+    if (catalogTable.isDefined && userSpecifiedSchema.isDefined) {
+      // backwards compatibility before SPARK-18510. Return the schema of catalog tables as is
+      return userSpecifiedSchema.get -> partitionSchema.map(_.name)
+    }
+    val tableSchema = userSpecifiedSchema.map { schema =>
```
gatorsmile: `tableSchema` -> `dataSchema`

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
```diff
@@ -291,22 +360,24 @@ case class DataSource(
...
-              // TODO: Case sensitivity.
-              schema
-                  .find(_.name.toLowerCase() == c.toLowerCase())
-                  .getOrElse(throw new AnalysisException(s"Invalid partition column '$c'"))
-            })
+        val (schema, inferredPartitionColumns) = getOrInferFileFormatSchema(format)
+        val dataSchema = if (isStreaming) {
+          schema
+        } else {
+          StructType(schema.dropRight(inferredPartitionColumns.length))
```
gatorsmile: This is little hacky.  How about changing the return of `getOrInferFileFormatSchema` to (`dataSchema`, `partitionSchema`)?

gatorsmile:
Also cc @ericl @cloud-fan who just recently change the related codes.

tdas (commented):
This is round 1 of reviewing. I am still looking to understand the code.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
```diff
@@ -84,30 +84,96 @@ case class DataSource(
...
         caseInsensitiveOptions,
-        fileCatalog.allFiles())
-
-      inferred.map { inferredSchema =>
-        StructType(inferredSchema ++ partitionSchema) -> partitionSchema.map(_.name)
-      }
+        tempFileCatalog.allFiles())
     }.getOrElse {
-      throw new AnalysisException("Unable to infer schema. It must be specified manually.")
+      val exampleFiles = tempFileCatalog.allFiles().take(2).mkString(",")
```
tdas: Why will it generate the files? at max it may print few file names that the json format actually ignores while inferring schema.

tdas: None the less, to avoid this, could you not just print the `allPaths`?

brkyvz: I tried to keep it consistent with before. I can remove it.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
tdas: nvm. with justPartitioning param, its more confusing as it is right now. so I like this suggestion

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
```diff
@@ -291,22 +360,24 @@ case class DataSource(
...
-                  .getOrElse(throw new AnalysisException(s"Invalid partition column '$c'"))
-            })
+        val (schema, inferredPartitionColumns) = getOrInferFileFormatSchema(format)
+        val dataSchema = if (isStreaming) {
+          schema
+        } else {
+          StructType(schema.dropRight(inferredPartitionColumns.length))
+        }
+
+        val partitionSchema = if (inferredPartitionColumns.isEmpty) {
```
tdas: This is the one i dont understand. why should the behavior be different if this non-streaming source (HadoopFSRelation) is being create by user directly or by the file streaming source.

tdas (commented):
Round 2

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
brkyvz: good idea

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
```diff
@@ -84,30 +84,96 @@ case class DataSource(
...
 
   /**
-   * Infer the schema of the given FileFormat, returns a pair of schema and partition column names.
+   * Get the schema of the given FileFormat, if provided by `userSpecifiedSchema`, or try to infer
+   * it. In the read path, only Hive managed tables provide the partition columns properly when
+   * initializing this class. All other file based data sources will try to infer the partitioning,
+   * and then cast the inferred types to user specified dataTypes if the partition columns exist
+   * inside `userSpecifiedSchema`, otherwise we can hit data corruption bugs like SPARK-18510.
+   * This method will try to do the least amount of work given whether `userSpecifiedSchema` and
+   * `partitionColumns` are provided. Here are some code paths that use this method:
```
tdas: Can you document what "least amount of work" is? That is, it will skip file scanning if .....

sql/core/src/test/scala/org/apache/spark/sql/streaming/test/DataStreamReaderWriterSuite.scala
```diff
@@ -532,4 +532,50 @@ class DataStreamReaderWriterSuite extends StreamTest with BeforeAndAfter {
...
+          .start()
+        sq.processAllAvailable()
+        checkAnswer(
+          spark.table("corruption_test"),
+          // notice how `part` is ordered before `id`
+          Row(Array("1"), 0, "0") :: Row(Array("1", "2"), 1, "1") ::
+            Row(Array("1", "2", "3"), 2, "2") :: Row(Array("1", "2", "3", "4"), 3, "3") :: Nil
+        )
+      } finally {
+        spark.streams.active.foreach(_.stop())
```
tdas: nit: you can just to sq.stop() instead of try...catch, because `after { }`of this testsuite already has `spark.streams.active.foreach(_.stop())`

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
```diff
@@ -84,30 +84,95 @@ case class DataSource(
...
+      val resolved = tempFileCatalog.partitionSchema.map { partitionField =>
+        val equality = sparkSession.sessionState.conf.resolver
+        // SPARK-18510: try to get schema from userSpecifiedSchema, otherwise fallback to inferred
+        userSpecifiedSchema.flatMap(_.find(f => equality(f.name, partitionField.name))).getOrElse(
+          partitionField)
+      }
+      StructType(resolved)
+    } else {
+      // in streaming mode, we have already inferred and registered partition columns, we will
+      // never have to re-use this
```
tejasapatil: Feels like you wanted to say `we will never have to re-compute this`. Typo ?

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
```diff
@@ -84,30 +84,95 @@ case class DataSource(
...
+      lazy val inferredPartitions = tempFileCatalog.partitionSchema
```
tejasapatil: Every call of this method will re-compute `tempFileCatalog` which as per your comment is an expensive op. This kinda contradicts 3rd point in the method doc which says that subsequent invocations of this method would be cheap. Did I miss anything ?

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
```diff
@@ -84,30 +84,95 @@ case class DataSource(
...
-      val allPaths = caseInsensitiveOptions.get("path")
+  private def getOrInferFileFormatSchema(
+      format: FileFormat,
+      justPartitioning: Boolean = false): (StructType, StructType) = {
+    // the operations below are expensive therefore try not to do them if we don't need to
+    lazy val tempFileCatalog = {
+      val allPaths = caseInsensitiveOptions.get("path") ++ paths
       val globbedPaths = allPaths.toSeq.flatMap { path =>
         val hdfsPath = new Path(path)
         val fs = hdfsPath.getFileSystem(sparkSession.sessionState.newHadoopConf())
```
tejasapatil: Creating new Hadoop config object for every path seems expensive to me. Can it be created once outside the `flatMap` and re-used inside ?

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
```diff
@@ -84,30 +84,95 @@ case class DataSource(
...
+      // never have to re-use this
+      lazy val inferredPartitions = tempFileCatalog.partitionSchema
+      // maintain old behavior before SPARK-18510. If userSpecifiedSchema is empty used inferred
+      // partitioning
+      if (userSpecifiedSchema.isEmpty) {
+        inferredPartitions
+      } else {
+        val partitionFields = partitionColumns.map { partitionColumn =>
+          userSpecifiedSchema.flatMap(_.find(_.name == partitionColumn)).orElse {
+            logDebug(s"Schema of partition column: $partitionColumn not found in specified " +
```
tejasapatil: - can you include the name of the table / relation ? the computation can refer to multiple tables which might have same `partitionColumn`
- can you also print the specified schema ? seems like would be informative for debugging
- It would be good to include the inferred datatype used instead of just mentioning that we would fallback

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
```diff
@@ -84,30 +84,95 @@ case class DataSource(
...
+              "schema. Falling back to inferred dataType if it exists.")
+            inferredPartitions.find(_.name == partitionColumn)
+          }.getOrElse {
+            throw new AnalysisException("Failed to resolve the schema for the partition column: " +
```
tejasapatil: can you include the name of the table / relation ? the computation can refer to multiple tables which might have same `partitionColumn`

ericl:
This might change the behavior, but how about just raising an error if the partition types differ from those provided by the user, or the user failed to provide a partitioning schema? It seems confusing to partially infer a schema when the user does not provide it.

brkyvz:
@ericl I feel that would probably break 90% of production Spark jobs out there, therefore am a bit scared of something radical. I agree, it's confusing and annoying

brkyvz:
Thanks @tejasapatil for the review. Addressed your comments

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
```diff
@@ -84,30 +84,95 @@ case class DataSource(
...
+              s"$partitionColumn. It must be specified manually.")
+          }
+        }
+        StructType(partitionFields)
+      }
+    }
+    if (justPartitioning) {
+      return new StructType() -> partitionSchema
```
brkyvz: done

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
brkyvz: I included `format`. Not sure how to access the name

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
brkyvz: Since everything is already defined in `userSpecifiedSchema` once we create the `FileStreamSource`, we will never have to materialize this variable, because we will not use it.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
brkyvz: reworded to `materialize this`

cloud-fan:
```
spark.read
  .schema(someSchemaWherePartitionColumnsAreStrings)
```
I don't think this is valid use case, `DataFrameReader` can't specify partition columns, so we will always infer partitions.

I think the real problem is `HadoopFsRelation.schema`:
```
val schema: StructType = {
  val dataSchemaColumnNames = dataSchema.map(_.name.toLowerCase).toSet
  StructType(dataSchema ++ partitionSchema.filterNot { column =>
    dataSchemaColumnNames.contains(column.name.toLowerCase)
  })
}
```
It sliently drops the partition schema if the partition column names are duplicated in data schema.

I think the best solution is to add `partitionBy` in `DataFrameReader` so that we can skip inferring partitions really. But this maybe too late for 2.1, we should define a better semantic for the current "broken" API.

> Once we find what the partition columns are, we try to find them in the user specified schema and use the dataType provided there, or fall back to the smallest common data type.

This LGTM

cloud-fan:
shall we update the document of `DataSource.partitionColumns`? `When this ist is empty, the relation is unpartitioned.` this is wrong now, the relation can be partitioned even if `partitionColumns` is empty.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
```diff
@@ -84,30 +88,106 @@ case class DataSource(
...
+      }
+    }
+    if (justPartitioning) {
+      return (null, partitionSchema)
+    }
+    val dataSchema = userSpecifiedSchema.map { schema =>
+      val equality = sparkSession.sessionState.conf.resolver
+      StructType(schema.filterNot(f => partitionSchema.exists(p => equality(p.name, f.name))))
+    }.orElse {
+      format.inferSchema(
```
tdas: just to confirm, inferschema returns schema without the partition columns?

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
```diff
@@ -144,8 +224,8 @@ case class DataSource(
               "you may be able to create a static DataFrame on that directory with " +
               "'spark.read.load(directory)' and infer schema from it.")
         }
-        val (schema, partCols) = inferFileFormatSchema(format)
-        SourceInfo(s"FileSource[$path]", schema, partCols)
+        val (schema, partCols) = getOrInferFileFormatSchema(format)
```
tdas: rename to dataSchema and partitionSchema

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
```diff
@@ -281,33 +361,25 @@ case class DataSource(
...
-        val partitionSchema = userSpecifiedSchema.map { schema =>
-          StructType(
-            partitionColumns.map { c =>
-              // TODO: Case sensitivity.
-              schema
-                  .find(_.name.toLowerCase() == c.toLowerCase())
-                  .getOrElse(throw new AnalysisException(s"Invalid partition column '$c'"))
-            })
-        }
+        val (dataSchema, inferredPartitionSchema) = getOrInferFileFormatSchema(format)
```
tdas: nit: this may not be inferred right? so just partitionSchema would be a better name.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
```diff
@@ -281,33 +361,25 @@ case class DataSource(
       // This is a non-streaming file based datasource.
       case (format: FileFormat, _) =>
         val allPaths = caseInsensitiveOptions.get("path") ++ paths
+        val hadoopConf = sparkSession.sessionState.newHadoopConf()
         val globbedPaths = allPaths.flatMap { path =>
           val hdfsPath = new Path(path)
-          val fs = hdfsPath.getFileSystem(sparkSession.sessionState.newHadoopConf())
+          val fs = hdfsPath.getFileSystem(hadoopConf)
```
tdas: any need for this change? hadoopConf is not reused any where else

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
```diff
@@ -281,33 +361,25 @@ case class DataSource(
       // This is a non-streaming file based datasource.
       case (format: FileFormat, _) =>
         val allPaths = caseInsensitiveOptions.get("path") ++ paths
+        val hadoopConf = sparkSession.sessionState.newHadoopConf()
         val globbedPaths = allPaths.flatMap { path =>
```
tdas: Just to confirm, for paths with globs, this would expand here ONCE, and then expand them AGAIN in `getOrInferFileFormatSchema`, right? 

If so, we dont have to fix it in this PR, but we should document this in a JIRA or something for fixing later.

tdas (commented):
LGTM. Just a few questions that needs clarification.

tdas:
Would be good if @rxin can take a look.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
```diff
@@ -84,30 +88,106 @@ case class DataSource(
...
+   *         `null`.
    */
-  private def inferFileFormatSchema(format: FileFormat): (StructType, Seq[String]) = {
-    userSpecifiedSchema.map(_ -> partitionColumns).orElse {
-      val allPaths = caseInsensitiveOptions.get("path")
+  private def getOrInferFileFormatSchema(
+      format: FileFormat,
+      justPartitioning: Boolean = false): (StructType, StructType) = {
+    // the operations below are expensive therefore try not to do them if we don't need to
+    lazy val tempFileCatalog = {
```
liancheng: Nit: `tempFileIndex`

sql/core/src/test/scala/org/apache/spark/sql/execution/command/DDLSuite.scala
```diff
@@ -274,7 +274,7 @@ class DDLSuite extends QueryTest with SharedSQLContext with BeforeAndAfterEach {
           pathToPartitionedTable,
           userSpecifiedSchema = Option("num int, str string"),
           userSpecifiedPartitionCols = partitionCols,
-          expectedSchema = new StructType().add("num", IntegerType).add("str", StringType),
+          expectedSchema = new StructType().add("str", StringType).add("num", IntegerType),
```
liancheng: I believe the original test case was incorrect. Although the schema check passes, if you really read rows out of the Dataset, you'll hit an exception, as shown in the following Spark shell session:

```scala

val df0 = spark.range(10).select(
  ('id % 4) cast StringType as "part",
  'id cast StringType as "data"
)

val path = "/tmp/part.parquet"
df0.write.mode("overwrite").partitionBy("part").parquet(path)

val df1 = spark.read.schema(
  new StructType()
    .add("part", StringType, nullable = true)
    .add("data", StringType, nullable = true)
).parquet(path)

df1.printSchema()
// root
//  |-- part: string (nullable = true)
//  |-- data: string (nullable = true)

df1.show()
// 16/11/22 22:52:21 ERROR Executor: Exception in task 0.0 in stage 10.0 (TID 34)
// java.lang.NullPointerException
//         at org.apache.spark.sql.execution.vectorized.OnHeapColumnVector.getArrayLength(OnHeapColumnVector.java:375)
//         at org.apache.spark.sql.execution.vectorized.ColumnVector.getArray(ColumnVector.java:554)
//         at org.apache.spark.sql.execution.vectorized.ColumnVector.getByteArray(ColumnVector.java:576)
//         [...]
```

liancheng (commented):
Mostly LGTM except for a few minor issues.

tdas:
Thanks @liancheng  for your comments. Since these are mostly nits, I am going to merge this PR (since it fixes critical bug for 2.1) and address the final comments in a separate PR.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
zsxwing: > any need for this change? hadoopConf is not reused any where else

this is to avoid creating a new hadoopConf for each path.

PR #15953, opened 2016-11-21
[SPARK-18517][SQL] DROP TABLE IF EXISTS should not warn for non-existing tables

dongjoon-hyun:

Currently, `DROP TABLE IF EXISTS` shows warning for non-existing tables. However, it had better be quiet for this case by definition of the command.

**BEFORE**
```scala
scala> sql("DROP TABLE IF EXISTS nonexist")
16/11/20 20:48:26 WARN DropTableCommand: org.apache.spark.sql.catalyst.analysis.NoSuchTableException: Table or view 'nonexist' not found in database 'default';
```

**AFTER**
```scala
scala> sql("DROP TABLE IF EXISTS nonexist")
```

## How was this patch tested?

Manual because this is related to the warning messages instead of exceptions.

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
@@ -28,7 +28,7 @@ import org.apache.hadoop.mapred.{FileInputFormat, JobConf}
 
 import org.apache.spark.sql.{AnalysisException, Row, SparkSession}
 import org.apache.spark.sql.catalyst.TableIdentifier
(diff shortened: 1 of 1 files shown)


dongjoon-hyun:
Hi, @gatorsmile .
This is a related case with #15682 (Dropping View) and the `uncache` issue.
But, this case is different from the above case because this PR has the obvious solution.
How do you think about this bug and PR?

andrewor14:
Makes sense. This LGTM merging into master and 2.0. Thanks.

dongjoon-hyun:
Thank you so much, @andrewor14 ! :)

PR #15955, opened 2016-11-21
[SPARK-18434][ML][FOLLOWUP] Add checking for setSolver in GLM

zhengruifeng:
add checking in GLM.setSolver to forbidden unsupported solvers:
`val glm = new GeneralizedLinearRegression().setSolver("123")`

## How was this patch tested?
existing tests

cc @yanboliang

### DIFF
--- a/mllib/src/main/scala/org/apache/spark/ml/regression/GeneralizedLinearRegression.scala
+++ b/mllib/src/main/scala/org/apache/spark/ml/regression/GeneralizedLinearRegression.scala
@@ -227,11 +227,11 @@ class GeneralizedLinearRegression @Since("2.0.0") (@Since("2.0.0") override val 
    */
   @Since("2.0.0")
   def setSolver(value: String): this.type = {
-    require("irls" == value,
-      s"Solver $value was not supported. Supported options: irls")
+    require(supportedSolvers.contains(value),
+      s"Solver $value was not supported. Supported options: ${supportedSolvers.mkString(", ")}")
     set(solver, value)
   }
-  setDefault(solver -> "irls")
+  setDefault(solver -> IRLS)
 
   /**
    * Sets the link prediction (linear predictor) column name.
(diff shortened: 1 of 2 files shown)


zhengruifeng:
`HasSolver` now don't support value validation, so it's not easy to use:
GLR and LiR inherit `HasSolver`, but need to do param validation explicitly like this PR
MLP dont inherit `HasSolver`, and create param `solver` with `supportedSolvers`
It may be reasonable to modify `HasSolver` after 2.1

yanboliang:
@zhengruifeng It's better we can define a variable ```supportedSolvers``` in object ```GeneralizedLinearRegression``` and check it contains users' input value. I think it also applies to ```LinearRegression```, could you also change that? Thanks.

zhengruifeng:
@yanboliang I updated this pr according to your comments. And now it's `solver` looks like `MPLC`

mllib/src/main/scala/org/apache/spark/ml/regression/GeneralizedLinearRegression.scala
```diff
@@ -297,6 +301,13 @@ object GeneralizedLinearRegression extends DefaultParamsReadable[GeneralizedLine
   @Since("2.0.0")
   override def load(path: String): GeneralizedLinearRegression = super.load(path)
 
+  /** String name for "irls" solver. */
+  private[regression] val IRLS = "irls"
```
yanboliang: Could we make it more private (i.e. ```private[GeneralizedLinearRegression]```)?

mllib/src/main/scala/org/apache/spark/ml/regression/LinearRegression.scala
```diff
@@ -410,6 +412,18 @@ class LinearRegression @Since("1.3.0") (@Since("1.3.0") override val uid: String
 @Since("1.6.0")
 object LinearRegression extends DefaultParamsReadable[LinearRegression] {
 
+  /** String name for "auto" solver. */
+  private[regression] val AUTO = "auto"
```
yanboliang: Ditto.

yanboliang (commented):
One minor comment, otherwise, LGTM. Thanks.

zhengruifeng:
@yanboliang I add the `private[regression]` limitation refering to `MultilayerPerceptronClassifier` implementation: https://github.com/apache/spark/blob/master/mllib/src/main/scala/org/apache/spark/ml/classification/MultilayerPerceptronClassifier.scala#L272

zhengruifeng:
After offline discussion with @yanboliang , I will update this PR to make solvers more private and include MLPC in this PR.

yanboliang:
After offline discussion with @zhengruifeng , we think the validation in the ```Param``` itself should be better, so other ways (which not calling setter method) of setting params can also benefit from the validation. We open [SPARK-18518](https://issues.apache.org/jira/browse/SPARK-18518) to a more perfect solution.

PR #15956, opened 2016-11-21
[SPARK-18519][SQL] map type can not be used in EqualTo

cloud-fan:

Technically map type is not orderable, but can be used in equality comparison. However, due to the limitation of the current implementation, map type can't be used in equality comparison so that it can't be join key or grouping key.

This PR makes this limitation explicit, to avoid wrong result.

## How was this patch tested?

updated tests.

### DIFF
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/CheckAnalysis.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/CheckAnalysis.scala
@@ -183,21 +183,6 @@ trait CheckAnalysis extends PredicateHelper {
               s"join condition '${condition.sql}' " +
                 s"of type ${condition.dataType.simpleString} is not a boolean.")
 
-          case j @ Join(_, _, _, Some(condition)) =>
-            def checkValidJoinConditionExprs(expr: Expression): Unit = expr match {
-              case p: Predicate =>
-                p.asInstanceOf[Expression].children.foreach(checkValidJoinConditionExprs)
-              case e if e.dataType.isInstanceOf[BinaryType] =>
-                failAnalysis(s"binary type expression ${e.sql} cannot be used " +
-                  "in join conditions")
-              case e if e.dataType.isInstanceOf[MapType] =>
-                failAnalysis(s"map type expression ${e.sql} cannot be used " +
-                  "in join conditions")
-              case _ => // OK
-            }
-
-            checkValidJoinConditionExprs(condition)
-
           case Aggregate(groupingExprs, aggregateExprs, child) =>
             def checkValidAggregateExpression(expr: Expression): Unit = expr match {
               case aggExpr: AggregateExpression =>
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/predicates.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/predicates.scala
@@ -412,6 +412,29 @@ case class EqualTo(left: Expression, right: Expression)
 
   override def inputType: AbstractDataType = AnyDataType
 
+  override def checkInputDataTypes(): TypeCheckResult = {
+    super.checkInputDataTypes() match {
+      case TypeCheckResult.TypeCheckSuccess =>
+        // TODO: although map type is not orderable, technically map type should be able to be used
+        // in equality comparison, remove this type check once we support it.
+        if (hasMapType(left.dataType)) {
+          TypeCheckResult.TypeCheckFailure("Cannot use map type in EqualTo, but the actual " +
+            s"input type is ${left.dataType.catalogString}.")
+        } else {
+          TypeCheckResult.TypeCheckSuccess
+        }
+      case failure => failure
+    }
+  }
+
+  private def hasMapType(dt: DataType): Boolean = dt match {
+    case _: MapType => true
+    case st: StructType => st.map(_.dataType).exists(hasMapType)
+    case a: ArrayType => hasMapType(a.elementType)
+    case udt: UserDefinedType[_] => hasMapType(udt.sqlType)
+    case _ => false
+  }
+
   override def symbol: String = "="
 
   protected override def nullSafeEval(input1: Any, input2: Any): Any = {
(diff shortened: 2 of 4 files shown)


sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/CheckAnalysis.scala
```diff
@@ -183,21 +183,6 @@ trait CheckAnalysis extends PredicateHelper {
               s"join condition '${condition.sql}' " +
                 s"of type ${condition.dataType.simpleString} is not a boolean.")
 
-          case j @ Join(_, _, _, Some(condition)) =>
-            def checkValidJoinConditionExprs(expr: Expression): Unit = expr match {
-              case p: Predicate =>
-                p.asInstanceOf[Expression].children.foreach(checkValidJoinConditionExprs)
-              case e if e.dataType.isInstanceOf[BinaryType] =>
```
cloud-fan: This doesn't make sense, `BinaryType` is comparable, we should accept it as join key.

cloud-fan:
cc @yhuai @hvanhovell  @gatorsmile

rxin:
So technically I think map type should be able to be used in equality comparison, but not orderable, so I'm not sure if this is correct.

Spark SQL currently has a restriction that we just don't allow map types to be used in equality comparison at all, but I don't want to structure code in a way that would make it harder to support that in the future.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/predicates.scala
```diff
@@ -388,6 +388,21 @@ abstract class BinaryComparison extends BinaryOperator with Predicate {
       defineCodeGen(ctx, ev, (c1, c2) => s"${ctx.genComp(left.dataType, c1, c2)} $symbol 0")
     }
   }
+
+  override def inputType: AbstractDataType = AnyDataType
```
rxin: isn't this "wrong"? map type is not supported here is it?

cloud-fan: We need this to satisfy `BinaryOperator`, the type checking logic is defined by both `BinaryOperator` and `checkInputDataTypes` here. Shall we make `BinaryComparison` not extends `BinaryOperator` and duplicate its type checking logic here?

cloud-fan:
> Spark SQL currently has a restriction that we just don't allow map types to be used in equality comparison at all, but I don't want to structure code in a way that would make it harder to support that in the future.

If we wanna support equality comparison for map type in the future, we can just update the type checking logic for `BinaryComparison`. For now I think it's better to forbid users to do this, to avoid confusing behaviors, e.g. `SELECT MAP(1,2) = MAP(1,2)` returns false.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/predicates.scala
```diff
@@ -388,6 +388,26 @@ abstract class BinaryComparison extends BinaryOperator with Predicate {
...
+  // AnyDataType here to bypass this check, and implement the actual type checking logic in
+  // `checkInputDataTypes`
+  override def inputType: AbstractDataType = AnyDataType
+
+  // TODO: although map type is not orderable, technically map type should be able to be used
+  // in equality comparison, update this type check once we support it.
+  override def checkInputDataTypes(): TypeCheckResult = {
+    super.checkInputDataTypes() match {
+      case TypeCheckResult.TypeCheckSuccess =>
+        if (RowOrdering.isOrderable(left.dataType)) {
```
gatorsmile: Semantically, this is different from the previous lilmits:
```
override def inputType: AbstractDataType = TypeCollection.Ordered
```

We add the support of `StructType`, `ArrayType` and `UDT`. It sounds like no test case coverage. Could you please add them?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/predicates.scala
```diff
@@ -388,6 +388,26 @@ abstract class BinaryComparison extends BinaryOperator with Predicate {
       defineCodeGen(ctx, ev, (c1, c2) => s"${ctx.genComp(left.dataType, c1, c2)} $symbol 0")
     }
   }
+
+  // The type checking logic of BinaryComparison can't be expressed by `inputType`, so we use
+  // AnyDataType here to bypass this check, and implement the actual type checking logic in
+  // `checkInputDataTypes`
+  override def inputType: AbstractDataType = AnyDataType
```
cloud-fan: hmm, Hive only supports atomic type, shall we follow it?

cloud-fan: FYI, postgres supports binary comparison on array type.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/predicates.scala
cloud-fan: let me minimize the size of this PR, I'll keep `GreaterThan`, `LessThan`, etc. unchanged, and only update `EqualTo`, `EqualNullSafe` to forbid map types.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/CheckAnalysis.scala
hvanhovell: We should add a few tests with binary keys, that can also be done in a follow-up. The parser supports binary literals, so we can use `SQLQueryTestSuite` for this.

hvanhovell (commented):
A few minor things, otherwise LGTM

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/CheckAnalysis.scala
cloud-fan: I already added one: https://github.com/apache/spark/pull/15956/files#diff-999f50ed13c690eaec243e3b3446af08R468 , yea we should add more end-to-end tests in `SQLQueryTestSuite` in follow-up

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/predicates.scala
```diff
@@ -440,6 +463,29 @@ case class EqualNullSafe(left: Expression, right: Expression) extends BinaryComp
 
   override def inputType: AbstractDataType = AnyDataType
 
+  override def checkInputDataTypes(): TypeCheckResult = {
```
cloud-fan: well it's only duplicated twice and will be removed after https://github.com/apache/spark/pull/15970 , I think it should be fine.

hvanhovell:
LGTM pending Jenkins

hvanhovell:
LGTM - merging to master/2.0. Thanks!

hvanhovell:
@cloud-fan I cannot merge this to 2.0. Can you make a backport if we need one?

cloud-fan:
Yea, I'll backport

PR #15957, opened 2016-11-21
[SPARK-18520][ML] Add missing setXXXCol methods for BisectingKMeansModel and GaussianMixtureModel

zhengruifeng:
add `setFeaturesCol` and `setPredictionCol` for BiKModel and GMModel
add `setProbabilityCol` for GMModel
## How was this patch tested?
existing tests


srowen:
CC @jkbradley per your message today in case this should really be in for 2.1

holdenk:
These look like really basic additions of things we simply forgot to expose so I think we can include it between RCs. Also LGTM.

yanboliang:
LGTM, just like @holdenk said, this is very basic change. So I'll merge it into master and branch-2.1. Thanks!

PR #15958, opened 2016-11-21
[SPARK-17932][SQL] Support SHOW TABLES EXTENDED LIKE 'identifier_with_wildcards' statement

jiangxb1987:

Currently we haven't implemented `SHOW TABLE EXTENDED` in Spark 2.0. This PR is to implement the statement.
Goals:
1. Support `SHOW TABLES EXTENDED LIKE 'identifier_with_wildcards'`;
2. Explicitly output an unsupported error message for `SHOW TABLES [EXTENDED] ... PARTITION` statement;
3. Improve test cases for `SHOW TABLES` statement.

## How was this patch tested?
1. Add new test cases in file `show-tables.sql`.
2. Modify tests for `SHOW TABLES` in `DDLSuite`.


jiangxb1987:
retest this please.

jiangxb1987:
cc @hvanhovell @gatorsmile

gatorsmile:
cc @dilipbiswal The original code was wrote by you. Could you review it? Thanks!

dilipbiswal:
@gatorsmile Thank you for cc'ing me. Overall it looks good to me. 
I have a few observations/comments that i would list below which hopefully should help us
decide.

1) Hive actually has two commands i) show tables ii) show table and the extended
   switch is applicable for the "show table" command. Have we decided to merge
   the semantics of these two commands in an effort to similify ?

2) As a side effect, we actually print the database name and table name two times. 
   Is that ok ?

3) The format of hive seems better as there is a newline between two tables. I created
   a regular table and a temporary view with same prefix and tried the command. Its very
   difficult to distinguish the output between table and view. here is how it looks.

```SQL
spark-sql> create table realtable(c1 int) using parquet;
Time taken: 0.467 seconds
spark-sql> create temporary view realview (c1 int, c2 int) using parquet;
Time taken: 0.053 seconds
spark-sql> show tables extended like 'real*';
default	realtable	false	CatalogTable(
	Table: `default`.`realtable`
	Owner: cloudera
	Created: Tue Nov 22 17:41:56 PST 2016
	Last Access: Wed Dec 31 16:00:00 PST 1969
	Type: MANAGED
	Schema: [StructField(c1,IntegerType,true)]
	Provider: parquet
	Properties: [transient_lastDdlTime=1479865316]
	Storage(Location: file:/user/hive/warehouse/realtable, InputFormat: org.apache.hadoop.hive.ql.io.parquet.MapredParquetInputFormat, OutputFormat: org.apache.hadoop.hive.ql.io.parquet.MapredParquetOutputFormat, Serde: org.apache.hadoop.hive.ql.io.parquet.serde.ParquetHiveSerDe, Properties: [serialization.format=1]))
	realview	true	CatalogTable(
	Table: `realview`
	Created: Tue Nov 22 17:42:28 PST 2016
	Last Access: Wed Dec 31 15:59:59 PST 1969
	Type: VIEW
	Schema: [StructField(c1,IntegerType,true), StructField(c2,IntegerType,true)]
	Storage())
Time taken: 0.104 seconds, Fetched 2 row(s)

```

jiangxb1987:
@dilipbiswal Thank you for reviewing this PR! I want to share my thoughts about the questions:
1. Perhaps using the same semantics is better IMO, because that'll be less comfusing. I've often mistype between `show table` and `show tables` in HIVE;
2. I'm neutral to the issue whether we should still print the database name and the table name when we have printed the extended information;
3. +1 for that the format should be improved, perhaps add one empty line after each row?

gatorsmile:
Could you update the PR title? The current one is confusing. This PR is not a bug fix.

jiangxb1987:
@gatorsmile I've updated the title, thank you!

gatorsmile:
The `Non-Goals` in the PR description is also confusing. Basically, what this PR does is to explicitly output an unsupported error message if users specify PARTITION. Could you also update that? Thanks!

sql/core/src/main/scala/org/apache/spark/sql/execution/command/tables.scala
```diff
@@ -589,30 +589,49 @@ case class DescribeTableCommand(
...
     AttributeReference("database", StringType, nullable = false)() ::
       AttributeReference("tableName", StringType, nullable = false)() ::
-      AttributeReference("isTemporary", BooleanType, nullable = false)() :: Nil
+      AttributeReference("isTemporary", BooleanType, nullable = false)() :: tableExtendedInfo
   }
 
   override def run(sparkSession: SparkSession): Seq[Row] = {
+    if (isExtended && !tableIdentifierPattern.isDefined) {
+      throw new AnalysisException(
+        s"SHOW TABLES EXTENDED must have identifier_with_wildcards specified.")
```
hvanhovell: Throw this exception when creating the class. This is to late.

jiangxb1987:
retest this please.

hvanhovell:

sql/catalyst/src/main/antlr4/org/apache/spark/sql/catalyst/parser/SqlBase.g4
```diff
@@ -124,8 +124,8 @@ statement
         (USING resource (',' resource)*)?                              #createFunction
     | DROP TEMPORARY? FUNCTION (IF EXISTS)? qualifiedName              #dropFunction
     | EXPLAIN (LOGICAL | FORMATTED | EXTENDED | CODEGEN)? statement    #explain
-    | SHOW TABLES ((FROM | IN) db=identifier)?
-        (LIKE? pattern=STRING)?                                        #showTables
+    | SHOW TABLES EXTENDED? ((FROM | IN) db=identifier)?
+        (LIKE? pattern=STRING)? partitionSpec?                         #showTables
```
gatorsmile: This Syntax is different from Hive. 
```
SHOW TABLE EXTENDED [IN|FROM database_name] LIKE 'identifier_with_wildcards' [PARTITION(partition_spec)];
```

gatorsmile:
@jiangxb1987  Could you submit a follow-up PR to correct it? Thanks!

jiangxb1987:
@gatorsmile Could you explicitly point out which part should be modified? IMO we don't follow the HIVE syntax unconditionally. Thanks!

gatorsmile:
If needed, we should not break the consistency. For example, `SHOW TABLE` and `SHOW TABLES` are different.

PR #15959, opened 2016-11-21
[SPARK-18522][SQL] Explicit contract for column stats serialization

rxin:
The current implementation of column stats uses the base64 encoding of the internal UnsafeRow format to persist statistics (in table properties in Hive metastore). This is an internal format that is not stable across different versions of Spark and should NOT be used for persistence. In addition, it would be better if statistics stored in the catalog is human readable.

This pull request introduces the following changes:

1. Created a single ColumnStat class to for all data types. All data types track the same set of statistics.
2. Updated the implementation for stats collection to get rid of the dependency on internal data structures (e.g. InternalRow, or storing DateType as an int32). For example, previously dates were stored as a single integer, but are now stored as java.sql.Date. When we implement the next steps of CBO, we can add code to convert those back into internal types again.
3. Documented clearly what JVM data types are being used to store what data.
4. Defined a simple Map[String, String] interface for serializing and deserializing column stats into/from the catalog.
5. Rearranged the method/function structure so it is more clear what the supported data types are, and also moved how stats are generated into ColumnStat class so they are easy to find.

## How was this patch tested?
Removed most of the original test cases created for column statistics, and added three very simple ones to cover all the cases. The three test cases validate:
1. Roundtrip serialization works.
2. Behavior when analyzing non-existent column or unsupported data type column.
3. Result for stats collection for all valid data types.

Also moved parser related tests into a parser test suite and added an explicit serialization test for the Hive external catalog.


rxin:
cc @srinathshankar @cloud-fan @wzhfy

sql/core/src/main/scala/org/apache/spark/sql/execution/command/AnalyzeColumnCommand.scala
```diff
@@ -100,99 +100,30 @@ object AnalyzeColumnCommand extends Logging {
...
     val rowCount = statsRow.getLong(0)
     val columnStats = attributesToAnalyze.zipWithIndex.map { case (expr, i) =>
-      val numFields = AnalyzeColumnCommand.numStatFields(expr.dataType)
-      (expr.name, ColumnStat(statsRow.getStruct(i + 1, numFields)))
+      (expr.name, ColumnStat.rowToColumnStat(statsRow.getStruct(i + 1)))
     }.toMap
     (rowCount, columnStats)
   }
-
-  private val zero = Literal(0, LongType)
```
rxin: all of these are now defined in a single function: ColumnStat.statExprs

sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveExternalCatalog.scala
```diff
@@ -627,26 +629,43 @@ private[spark] class HiveExternalCatalog(conf: SparkConf, hadoopConf: Configurat
...
+
+    if (statsProps.nonEmpty) {
+      val colStats = new scala.collection.mutable.HashMap[String, ColumnStat]
+
+      // For each column, recover its column stats. Note that this is currently a O(n^2) operation,
+      // but given the number of columns it usually not enormous, this is probably OK as a start.
+      // If we want to map this a linear operation, we'd need a stronger contract between the
+      // naming convention used for serialization.
+      tableWithSchema.schema.foreach { field =>
+        // TODO: Do we need to consider case sensivitiy here?
```
rxin: cc @cloud-fan any idea on this TODO?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/Statistics.scala
```diff
@@ -58,60 +61,127 @@ case class Statistics(
...
  */
-case class ColumnStat(statRow: InternalRow) {
+// TODO: decide if we want to use bigint to represent ndv and numNulls.
+case class ColumnStat(
+    ndv: Long,
+    min: Any,
+    max: Any,
+    numNulls: Long,
+    avgLen: Long,
+    maxLen: Long) {
```
wzhfy: I think we can remove maxLen too. We will use avgLen to estimate column length.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/Statistics.scala
```diff
@@ -58,60 +61,127 @@ case class Statistics(
...
+
+object ColumnStat extends Logging {
 
-  override def toString: String = {
-    // use Base64 for encoding
-    Base64.encodeBase64String(statRow.asInstanceOf[UnsafeRow].getBytes)
+  /** Returns true iff the we support gathering column statistics on column of the given type. */
+  def supportsType(dataType: DataType): Boolean = dataType match {
+    case _: NumericType | TimestampType | DateType | BooleanType => true
+    case StringType | BinaryType => true
```
cloud-fan: `AtomicType`?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/Statistics.scala
```diff
@@ -58,60 +61,127 @@ case class Statistics(
...
+      case _ => identity
+    }
+
+    try {
+      Some(ColumnStat(
+        ndv = map("ndv").toLong,
+        min = str2val(map.get("min").orNull),
+        max = str2val(map.get("max").orNull),
+        numNulls = map("numNulls").toLong,
+        avgLen = map.getOrElse("avgLen", "1").toLong,
```
cloud-fan: shall we make `avgLen` a `Option[Long]`?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/Statistics.scala
```diff
@@ -58,60 +61,127 @@ case class Statistics(
...
-  val numNulls: Long = statRow.getLong(0)
-  val avgColLen: Double = statRow.getDouble(1)
-  val maxColLen: Long = statRow.getInt(2)
-}
+    col.dataType match {
+      case _: NumericType | TimestampType | DateType | BooleanType =>
+        struct(ndv, Min(col), Max(col), numNulls, one, one)
+
+      case StringType | BinaryType =>
+        val emptyStr = Literal("")
```
cloud-fan: shall we use null?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/Statistics.scala
rxin: What about AtomicType?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/Statistics.scala
srinathshankar: For avgLen I think we should either have them be correct for fixed length types, or as cloud-fan suggests have it be an option.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/Statistics.scala
```diff
@@ -58,60 +61,127 @@ case class Statistics(
...
 
-case class StringColumnStat(statRow: InternalRow) {
-  // The indices here must be consistent with `ColumnStatStruct.stringColumnStat`.
-  val numNulls: Long = statRow.getLong(0)
-  val avgColLen: Double = statRow.getDouble(1)
-  val maxColLen: Long = statRow.getInt(2)
-  val ndv: Long = statRow.getLong(3)
-}
+    // the approximate ndv should never be larger than the number of rows
+    val ndv = Least(Seq(HyperLogLogPlusPlus(col, relativeSD), Count(one)))
```
srinathshankar: I'm assuming `HyperLogLogPlusPlus` ignores like other aggregates. In that case shouldn't you use `Count(col)` instead of `Count(one)` as an upper bound ?

srinathshankar (commented):
I also agree with cloud-fan's comments on null instead of emptyStr and removing maxLen

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/Statistics.scala
rxin: For the purpose of statistics I'd say there is no difference between the two, and it would be computationally cheaper to use count(1), since all columns will share one. Any objections?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/Statistics.scala
rxin: Yea I agree. I will use datatype.defaultSize.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/Statistics.scala
rxin: maxLen is actually useful if we want to check whether a column is a dimension column.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/Statistics.scala
rxin: good point fixed

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/Statistics.scala
rxin: yup

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/Statistics.scala
cloud-fan: we can just write `case _: AtomicType => true`

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/Statistics.scala
```diff
@@ -58,60 +61,127 @@ case class Statistics(
...
+ * @param ndv number of distinct values
+ * @param min minimum value
+ * @param max maximum value
+ * @param numNulls number of nulls
+ * @param avgLen average length of the values. For fixed-length types, this should be a constant.
+ * @param maxLen maximum length of the values. For fixed-length types, this should be a constant.
  */
-case class ColumnStat(statRow: InternalRow) {
+// TODO: decide if we want to use bigint to represent ndv and numNulls.
+case class ColumnStat(
```
tejasapatil: can you add some basic sanity checks ? eg. 
- `max >= min`
- `maxLen  >= avgLen` 
- `if (ndv == 1) then min == max`

Floats / decimals might behave badly but its good to check before anyone consumes these stats.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/Statistics.scala
rxin: That's a bad idea, unless we have an explicit contract that all atomic types (including future ones) support stats.

sql/core/src/main/scala/org/apache/spark/sql/execution/command/AnalyzeColumnCommand.scala
```diff
@@ -100,99 +100,30 @@ object AnalyzeColumnCommand extends Logging {
       exprOption.getOrElse(throw new AnalysisException(s"Invalid column name: $col."))
     }).toSeq
 
+    // Make sure the column types are supported for stats gathering.
+    attributesToAnalyze.foreach { attr =>
+      if (!ColumnStat.supportsType(attr.dataType)) {
+        throw new AnalysisException(
+          s"Data type ${attr.dataType.simpleString} for column ${attr.name} is not supported " +
```
tejasapatil: Yes. The command can be ran in a context of a pipeline with other commands. Say a job is writing to `n` tables and then populating stats for each of them. In case one of the queries to populate the stats fails, user would have to dig through the logs to find out which table caused the problem. If the exception message has ample information, one can simply look at the failure message of the container where the driver ran without needing to scroll through logs.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/Statistics.scala
```diff
@@ -58,60 +61,127 @@ case class Statistics(
...
-  def forBinary: BinaryColumnStat = BinaryColumnStat(statRow)
-  def forBoolean: BooleanColumnStat = BooleanColumnStat(statRow)
+  /**
+   * Returns a map from string to string that can be used to serialize the column stats.
+   * The key is the name of the field (e.g. "ndv" or "min"), and the value is the string
+   * representation for the value. The deserialization side is defined in [[ColumnStat.fromMap]].
+   *
+   * As part of the protocol, the returned map always contains a key called "version".
+   */
+  def toMap: Map[String, String] = Map(
```
rxin: yea locally i already used Option actually for min/max

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/Statistics.scala
```diff
@@ -58,60 +61,127 @@ case class Statistics(
...
+    val str2val: (String => Any) = dataType match {
+      case _: IntegralType => _.toLong
+      case _: DecimalType => Decimal(_)
+      case DoubleType | FloatType => _.toDouble
+      case BooleanType => _.toBoolean
+      case _ => identity
+    }
+
+    try {
+      Some(ColumnStat(
```
rxin: we haven't released anything in a non-backward compatible way. I think versioning is best dealt with by the serialization, and not by the engine.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/Statistics.scala
```diff
@@ -58,60 +61,127 @@ case class Statistics(
...
+        ndv = map("ndv").toLong,
+        min = str2val(map.get("min").orNull),
+        max = str2val(map.get("max").orNull),
+        numNulls = map("numNulls").toLong,
+        avgLen = map.getOrElse("avgLen", "1").toLong,
+        maxLen = map.getOrElse("maxLen", "1").toLong
+      ))
+    } catch {
+      case NonFatal(e) =>
+        logWarning("Failed to parse column statistics", e)
```
rxin: done

rxin:
OK I've pushed a new version that added test cases and also addressed almost all the comments.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/Statistics.scala
```diff
@@ -58,60 +61,164 @@ case class Statistics(
...
+  def fromMap(table: String, field: StructField, map: Map[String, String])
+    : Option[ColumnStat] = {
+    val str2val: (String => Any) = field.dataType match {
+      case _: IntegralType => _.toLong
+      case _: DecimalType => new java.math.BigDecimal(_)
+      case DoubleType | FloatType => _.toDouble
+      case BooleanType => _.toBoolean
+      case DateType => java.sql.Date.valueOf
+      case TimestampType => java.sql.Timestamp.valueOf
+      case _ => (v: String) => if (v == NULL_STRING) null else v
```
rxin: probably

rxin:
cc @cloud-fan

sql/hive/src/test/scala/org/apache/spark/sql/hive/StatisticsSuite.scala
```diff
@@ -319,7 +319,8 @@ class StatisticsSuite extends QueryTest with TestHiveSingleton with SQLTestUtils
           TableIdentifier(parquetTable))
         assert(DDLUtils.isDatasourceTable(catalogTable))
 
-        sql(s"INSERT INTO TABLE $parquetTable SELECT * FROM src")
+        // Add a filter to avoid creating too many partitions
+        sql(s"INSERT INTO TABLE $parquetTable SELECT * FROM src WHERE key < 10")
```
rxin: cc @wzhfy @cloud-fan 

This test case used to take 10 + secs on my laptop. We should avoid introducing test cases that take a long time to run.

wzhfy: got it

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/Statistics.scala
```diff
@@ -58,60 +61,170 @@ case class Statistics(
...
 
-  def forNumeric[T <: AtomicType](dataType: T): NumericColumnStat[T] = {
-    NumericColumnStat(statRow, dataType)
-  }
-  def forString: StringColumnStat = StringColumnStat(statRow)
-  def forBinary: BinaryColumnStat = BinaryColumnStat(statRow)
-  def forBoolean: BooleanColumnStat = BooleanColumnStat(statRow)
+  /**
+   * Returns a map from string to string that can be used to serialize the column stats.
+   * The key is the name of the field (e.g. "ndv" or "min"), and the value is the string
```
cloud-fan: nit: `ndv` -> `distinctCount`

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/Statistics.scala
```diff
@@ -58,60 +61,170 @@ case class Statistics(
...
+   * representation for the value. The deserialization side is defined in [[ColumnStat.fromMap]].
+   *
+   * As part of the protocol, the returned map always contains a key called "version".
+   * In the case min/max values are null (None), they will be stored as "<null>".
```
cloud-fan: I may miss the discussion, why not just remove the `max`, `min` entries in the map if they are null?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/Statistics.scala
```diff
@@ -58,60 +61,170 @@ case class Statistics(
...
+  def fromMap(table: String, field: StructField, map: Map[String, String])
+    : Option[ColumnStat] = {
+    val str2val: (String => Any) = field.dataType match {
+      case _: IntegralType => _.toLong
+      case _: DecimalType => new java.math.BigDecimal(_)
+      case DoubleType | FloatType => _.toDouble
+      case BooleanType => _.toBoolean
+      case DateType => java.sql.Date.valueOf
+      case TimestampType => java.sql.Timestamp.valueOf
+      case BinaryType | StringType => (v: String) => if (v == NULL_STRING) null else v
```
cloud-fan: so we use String as the external type for binary column? maybe we should mention it in the doc: https://github.com/apache/spark/pull/15959/files#diff-a4113ed6a89e8d19a39f5c27ce95658bR69

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/Statistics.scala
```diff
@@ -58,60 +61,170 @@ case class Statistics(
...
+        throw new AnalysisException("Column statistics deserialization is not supported for " +
+          s"column ${field.name} of data type: ${field.dataType}.")
+    }
+
+    try {
+      Some(ColumnStat(
+        distinctCount = BigInt(map(KEY_DISTINCT_COUNT).toLong),
+        // Note Option(..).flatMap(Option.apply) replaces Option(null) with None
+        min = map.get(KEY_MIN_VALUE).map(str2val).flatMap(Option.apply),
+        max = map.get(KEY_MAX_VALUE).map(str2val).flatMap(Option.apply),
```
cloud-fan: looks like here we assume the `max`, `min` entries may not in the map.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/Statistics.scala
```diff
@@ -58,60 +61,170 @@ case class Statistics(
...
-  val numNulls: Long = statRow.getLong(0)
-  val max: T#InternalType = statRow.get(1, dataType).asInstanceOf[T#InternalType]
-  val min: T#InternalType = statRow.get(2, dataType).asInstanceOf[T#InternalType]
-  val ndv: Long = statRow.getLong(3)
-}
+  /**
+   * Constructs an expression to compute column statistics for a given column.
+   *
+   * The expression should create a single struct column with the following schema:
+   * ndv: Long, min: T, max: T, numNulls: Long, avgLen: Long, maxLen: Long
```
cloud-fan: nit: `ndv` -> `distinctCount`, `numNulla` -> `nullCount`

sql/hive/src/test/scala/org/apache/spark/sql/hive/StatisticsSuite.scala
rxin: yea good point.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/Statistics.scala
rxin: yea good point.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/Statistics.scala
```diff
@@ -58,60 +61,170 @@ case class Statistics(
...
-  // The indices here must be consistent with `ColumnStatStruct.stringColumnStat`.
-  val numNulls: Long = statRow.getLong(0)
-  val avgColLen: Double = statRow.getDouble(1)
-  val maxColLen: Long = statRow.getInt(2)
-  val ndv: Long = statRow.getLong(3)
-}
+    // the approximate ndv should never be larger than the number of rows
+    val numNonNulls = if (col.nullable) Count(col) else Count(one)
+    val ndv = Least(Seq(HyperLogLogPlusPlus(col, relativeSD), numNonNulls))
+    val numNulls = Subtract(Count(one), numNonNulls)
```
rxin: That's just an optimizer question though.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/Statistics.scala
rxin: I updated it to be more future proof - so even if we serialize bytes using future version of spark, using the current version won't crash the system.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/Statistics.scala
```diff
@@ -58,60 +61,176 @@ case class Statistics(
...
+    : Option[ColumnStat] = {
+    val str2val: (String => Any) = field.dataType match {
+      case _: IntegralType => _.toLong
+      case _: DecimalType => new java.math.BigDecimal(_)
+      case DoubleType | FloatType => _.toDouble
+      case BooleanType => _.toBoolean
+      case DateType => java.sql.Date.valueOf
+      case TimestampType => java.sql.Timestamp.valueOf
+      case StringType => identity
+      // This version of Spark does not use min/max for binary columns so we ignore it.
```
cloud-fan: actually it's also true for string type: https://github.com/apache/spark/pull/15959/files#diff-a4113ed6a89e8d19a39f5c27ce95658bR213 , shall we put the `StringType` and `BinaryType` in the same case statement?

cloud-fan:
LGTM except one minor comment

sql/hive/src/test/scala/org/apache/spark/sql/hive/StatisticsSuite.scala
```diff
@@ -310,6 +270,110 @@ class StatisticsSuite extends QueryTest with TestHiveSingleton with SQLTestUtils
...
+
+      // Collect statistics
+      sql(s"analyze table $tableName compute STATISTICS FOR COLUMNS " + stats.keys.mkString(", "))
+
+      // Validate statistics
+      val hiveClient = spark.sharedState.externalCatalog.asInstanceOf[HiveExternalCatalog].client
+      val table = hiveClient.getTable("default", tableName)
+
+      val props = table.properties.filterKeys(_.startsWith("spark.sql.statistics.colStats"))
+      assert(props == Map(
```
rxin: cc @cloud-fan / @wzhfy 

This is a very explicit test for the serialization protocol.

sql/core/src/test/scala/org/apache/spark/sql/StatisticsCollectionSuite.scala
```diff
@@ -0,0 +1,218 @@
...
+      val df1 = spark.table("test")
+      val df2 = spark.table("test").limit(0)
+      val df = df1.join(df2, Seq("k"), "left")
+
+      val sizes = df.queryExecution.analyzed.collect { case g: Join =>
+        g.statistics.sizeInBytes
+      }
+
+      assert(sizes.size === 1, s"number of Join nodes is wrong:\n ${df.queryExecution}")
+      assert(sizes.head === BigInt(96),
```
cloud-fan: nit: `(20 + 4) * 4` may looks clearer than 96

sql/hive/src/test/scala/org/apache/spark/sql/hive/StatisticsSuite.scala
```diff
@@ -310,6 +270,110 @@ class StatisticsSuite extends QueryTest with TestHiveSingleton with SQLTestUtils
...
+      df.write.saveAsTable(tableName)
+
+      // Collect statistics
+      sql(s"analyze table $tableName compute STATISTICS FOR COLUMNS " + stats.keys.mkString(", "))
+
+      // Validate statistics
+      val hiveClient = spark.sharedState.externalCatalog.asInstanceOf[HiveExternalCatalog].client
+      val table = hiveClient.getTable("default", tableName)
+
+      val props = table.properties.filterKeys(_.startsWith("spark.sql.statistics.colStats"))
```
cloud-fan: use `HiveExternalCatalog.STATISTICS_COL_STATS_PREFIX`?

cloud-fan:
LGTM

sql/core/src/test/scala/org/apache/spark/sql/StatisticsCollectionSuite.scala
rxin: This is just code move around - it was the old test.

cloud-fan:
merging to master/2.1!

PR #15960, opened 2016-11-21
[SPARK-18521] Add `NoRedundantStringInterpolator` Scala rule

weiqingy:
This PR is to add a new scala style rule 'NoRedundantStringInterpolator' to prevent unnecessary string interpolators.

- Add the `NoRedundantStringInterpolator` rule in `scalastyle-config.xml`.
```
  <check customId="NoRedundantStringInterpolator" level="error" class="org.scalastyle.file.RegexChecker" enabled="true">
    <parameters><parameter name="regex">(?&lt;=[\s(])s"[^\$\n"]*"(?![\w"])</parameter></parameters>
    <customMessage>Do not use s string interpolator when no embed variable reference
    is in the processed string literals.</customMessage>
  </check>
```
- Fix the current cases which can not pass the`NoRedundantStringInterpolator` rule.

## How was this patch tested?
Pass the Jenkins tests and Scala style checking (`./dev/lint-scala`).


rxin:
This is going to make backports pretty painful. What's the concrete gain from this?

srowen:
Of course it's better to not use interpolation where there is no interpolation, but yes I wonder how much benefit there is to changing every single instance, because of the back-porting issue.

This pointless interpretation does have a runtime impact, so I could see making this change in non-test code in any section where performance is important.

I do like consistent style, good practices, and enforcing them automatically. We can't enforce this unless all the code complies unfortunately. 

I would support fixing this in any important section of code and am neutral about changing it all.

rxin:
So if we had started from clean slate I'd say it might be good to do. But given the size of the pull request it will create a lot of conflicts in backporting patches, and I doubt any of the locations are performance sensitive (unlike Java, our logging only generates the strings if the logging functionality for that level is turned on). If we can find a location that is actually perf sensitive then it's definitely worth fixing that one.

weiqingy:
Hi, @rxin @srowen Thanks for the prompt feedbacks and suggestions. Yes, I understand and agree with your concerns. The motivation of creating this PR is I think the redundant string interpolators are unnecessary and misleading, so it’s good to clean up the code and make complication a little faster.

How about making this PR in a long-term step-by-step manner in terms of modules? (the rule of `NoRedundantStringInterpolator` Scala style will be added in the final step)

Spark 2.1(~ Nov 2016): Mesos (2 files), Yarn (4 files), External (4 files), GraphX (5 files)
Spark 2.2(~ Mar 2017): Hive-Thrift-Server (4 files), Streaming (8 files), Example (18 files)
Spark 2.3(~ Jul 2017): Hive (28 files), Catalyst(19 files)
Spark 2.4(~ Nov 2017): SQL Core(47 files)
Spark 2.5(~ 2018): Core(54 files), MLLib (69 files)

If it is ok for you, I can make PRs based on the schedule above (Yes, I need to fix the failures of Spark unit tests). If not, I can just close this PR and SPARK-18521. 😊😊 Thanks.

srowen:
No, doing it in stages over years doesn't change the drawback, and just makes this more confusing. I agree with fixing this where performance may matter, or, fixing these issues when code is being changed otherwise.

weiqingy:
@srowen Ok. Thanks for the reply.

holdenk:
Arguing on the for-side: if we do want to add a rule like this adding it closer to the start of the 2.X series is probably the time to do it - any backports will hopefully not need to go back very far. (Otherwise we could leave it for the start of the 3.X since backports across major versions are already normally a bit painful).

Maybe it would make sense to run spark-perf & spark-sql-perf on this branch and see if there is any meaningful performance improvement?

srowen:
I skimmed all the instances here, and only a handful look like they might make a noticeable impact on performance. Almost all are tests, log messages, exception messages. Still, might be a tiny win in here.

I am still tempted because we are able to enforce this with checkstyle and it is a little ugly. (I'm especially amused by the `toString` that uses a StringBuilder, but feeds it strings concatenated with "+", and uses interpolation, but it isn't even needed!). 

What if we did the work to back-port this to 2.0.x? that would significantly reduce the back-porting conflict problem.

rxin:
It doesn't really hurt my eyes, but I can see why it hurts yours. GitHub doesn't even show the diff on all the files, and a large change like this is really easy to miss something wrong (e.g. one string interpolation really requires string interpolation but the "s" was removed and then the message shows up incorrectly). If we want to do this, we need somebody who is extremely detailed oriented to go through every single line of change carefully, and then backport them to all the old branches. If you have time here, I'd really recommend putting it into better use, e.g. help review other prs that actually have user facing impact or clean up real tech debt.

srowen:
Let's close this for now and consider changing these on a case-by-case basis if it appears to be a real performance bottleneck or while the code is being changed otherwise.

weiqingy:
Thanks @rxin @srowen @holdenk.

PR #15961, opened 2016-11-21
[SPARK-18523][PySpark]Make SparkContext.stop more reliable

kxepal:

This PR fixes SparkContext broken state in which it may fall if spark driver get crashed or killed by OOM.

## How was this patch tested?

1. Start SparkContext;
2. Find Spark driver process and `kill -9` it;
3. Call `sc.stop()`;
4. Create new SparkContext after that;

Without this patch you will crash on step 3 and won't be able to do step 4 without manual reset private attibutes or IPython notebook / shell restart.

### DIFF
--- a/python/pyspark/context.py
+++ b/python/pyspark/context.py
@@ -26,6 +26,8 @@ import warnings
 from threading import RLock
 from tempfile import NamedTemporaryFile
 
+from py4j.protocol import Py4JError
+
 from pyspark import accumulators
 from pyspark.accumulators import Accumulator
 from pyspark.broadcast import Broadcast
@@ -373,8 +375,19 @@ class SparkContext(object):
         Shut down the SparkContext.
         """
         if getattr(self, "_jsc", None):
-            self._jsc.stop()
-            self._jsc = None
+            try:
+                self._jsc.stop()
+            except Py4JError:
+                # Case: SPARK-18523
+                warnings.warn(
+                    'Unable to cleanly shutdown Spark JVM process.'
+                    ' It is possible that the process has crashed,'
+                    ' been killed or may also be in a zombie state.',
+                    RuntimeWarning
+                )
+                pass
+            finally:
+                self._jsc = None
         if getattr(self, "_accumulatorServer", None):
             self._accumulatorServer.shutdown()
             self._accumulatorServer = None


rxin:
cc @joshrosen @zsxwing

holdenk:
I'm not sure we want to silently swallow all Py4J errors on `sc.stop()` - at least we should probably warn about the situation (since it seems like it might be possible the JVM process is stuck or another issue has broken the py4j bridge).

srowen:
I don't see why you wouldn't warn()? otherwise there's no indication that stop() failed.

kxepal:
@srowen 
I was thought I pretty well described that above. TL;DR this information is not useful and you cannot do anything with it, but ignore, imho. 

But if you insist, I'll add a warning, that's not a problem. Just want to make sure that this is really reasonable to do.

srowen:
I can't think of any code I've seen that would fail to log a failure like this. Why do you say it's so completely unactionable that nobody should know about it? what's the downside to giving this information versus making it impossible to tell that the JVM failed to shut down?

kxepal:
@srowen 

> Why do you say it's so completely unactionable that nobody should know about it? 

That's just from my experience: in all the cases when drivers dies, it dies in the middle of something, when you do some computations. In this case you already get the Py4J exception on the next operation that triggers communication with JVM process. You already knows about the problem and you're going to fix anyway. So additional reminder on `sc.stop()` was looked redundant for me.

> what's the downside to giving this information versus making it impossible to tell that the JVM failed to shut down?

Perfect question! Well, I don't have any strong argument here. It seems like a matter of taste about logging information usefulness.

Ok, I'll add a warning. Does the following looks good for you?:
```python
warning.warn(RuntimeWarning, 
             'Unable to stop JVM process. Probably, it was crashed or externally killed')
```

holdenk:
I think that warning gives a false sense of security - as I indicated above this swallows all of the Py4J errors and there are a host of things which could cause the Py4J bridge to break down. The warning text I would choose would be more along the lines of: "Unable to cleanly shutdown Spark JVM process. It is possible that the process has crashed or been killed, but may also be in a zombie state."

This way the end user know to take a look and verify that the previous Spark JVM did indeed exit.

I'm still not sure that we want to do this though - if the Spark JVM has failed stopping and restarting the Spark context is going to have all of the previous references to Java objects invalid (not just the RDDs/DataFrames which happens on a stop/start of the Scala Spark context normally). It seems like the correct action for the user to take when the Py4J bridge breaks is starting over from scratch, either by exiting and re-running their notebook or otherwise re-submitting there job.

kxepal:
@holdenk 
Thanks for a warning message text. Nice one!

> I indicated above this swallows all of the Py4J errors and there are a host of things which could cause the Py4J bridge to break down. 

Suddenly, as you may see in [issue's traceback](https://issues.apache.org/jira/browse/SPARK-18523), it's py4j who raises too general exception on such kind of problem. I was too expected to see there Py4JNetworkError since it's network communication issue, but this didn't happen. The really good exceptions are get swallowed somewhere in the middle and get just printed to stderr via logging, but I'm not sure how to reraise them and how much things this will break.

> It seems like the correct action for the user to take when the Py4J bridge breaks is starting over from scratch, either by exiting and re-running their notebook or otherwise re-submitting there job.

Yes, that's what happens now: in case of failure we have to shutdown notebook, start it and re-run all the cells again. If we're not running in notebook - crash whole the script. Here comes two issues:

1. Usability. If you made some mistake or Spark job eventually fails, you wouldn't restart whole the notebook, but run cell with `sc.stop` and else cleanup stuff and re-run your Spark cells. That's simple procedure. But when Spark context stop fails, you have to follow plan B: restart all the things and re-run all the cells. That's could be quite boring and actually it is.

2. Correctness. SparkContext is a global shared mutable object and if we cannot correctly reset it state to default to start over that feels like something really wrong here. Should we run all the code that uses SparkContext in subprocesses just to be able to implement retry logic otherwise?

kxepal:
@holdenk @srowen 
Added a warning message, please take a second look.

srowen (approved):
The error handling seems prudent to me, but will wait for an OK from @holdenk

python/pyspark/context.py
```diff
@@ -373,8 +375,19 @@ def stop(self):
...
-            self._jsc.stop()
-            self._jsc = None
+            try:
+                self._jsc.stop()
+            except Py4JError:
+                # Case: SPARK-18523
+                warnings.warn(
+                    'Unable to cleanly shutdown Spark JVM process.'
+                    ' It is possible that the process has crashed'
+                    ' or been killed, but may also be in a zombie state.',
```
rxin: but -> or?

holdenk:
This looks good to me pending @rxins proposed wording change. I think restarting a stopped SparkContext which has been killed by an OOM killer or other issue is probably not a good thing to generally do and the warning text makes it clear enough that the machine may be in a bad state so the user can investigate if necessary.

While I'm unlikely to use this, it sounds like it could make life easier for some notebook users to avoid having to restart their notebook kernel. Thanks for taking the time on this PR @kxepal :)

Note: that if you do have automatic retry logic that likely should not be implemented this way - there is a chance this might lead to a very bad state (hence the warning message). [Just mentioning this since it was mentioned as one of the possible uses in our discussion].

python/pyspark/context.py
kxepal: Fixed, thanks!

kxepal:
@holdenk 
Agree with you here. The message is fixed, PR rebased.

holdenk:
Great, can you maybe rebase one more time? For some reason in the PR dashboard its showing up as not mergeable even though github thinks it is.

kxepal:
@holdenk 
Sure, done.

rxin:
hm you didn't need to do that. It was just a transient issue with github's own api.

rxin:
Merging in master/branch-2.1.

rxin:
Thanks!

kxepal:
Hooray! 🎉 Thank you all for help here! (:

PR #15962, opened 2016-11-21
[SPARK-18526][SQL][KAFKA] Allow users to configure max.poll.records.

ScrapCodes:

In Kafka source for structured streaming the value of the property `max.poll.records` is set to a default of 1, ~this is very slow rate of reading~. This should be set to a higher value or at least user should be able to configure it.

## How was this patch tested?

Existing tests and by deploying to a cluster.


marmbrus:
Do you have any performance comparisons to show that we need to do this?

On the driver, I think we want it to be as small as possible, because we don't want to pull any data down (it would be even better if we could just issue a raw ListOffsets request).  On the executors, we should know exactly how many offsets we need to pull down, so I'm not sure it make sense to allow users to configure this.

ScrapCodes:
You are right @marmbrus , I am sorry for not actually doing the proper comparison.

 Only after seeing the code I realized, we use `consumer.poll(0)` not for pulling records but to update consumer about the newly available offsets. And during my run, something else improved my throughput and not this property.

PR #15963, opened 2016-11-21

AnthonyTruchet:

block in-place.

## How was this patch tested?

### DIFF
--- a/mllib/src/main/scala/org/apache/spark/mllib/optimization/LBFGS.scala
+++ b/mllib/src/main/scala/org/apache/spark/mllib/optimization/LBFGS.scala
@@ -241,16 +241,28 @@ object LBFGS extends Logging {
       val bcW = data.context.broadcast(w)
       val localGradient = gradient
 
-      val (gradientSum, lossSum) = data.treeAggregate((Vectors.zeros(n), 0.0))(
-          seqOp = (c, v) => (c, v) match { case ((grad, loss), (label, features)) =>
-            val l = localGradient.compute(
-              features, label, bcW.value, grad)
-            (grad, loss + l)
-          },
-          combOp = (c1, c2) => (c1, c2) match { case ((grad1, loss1), (grad2, loss2)) =>
-            axpy(1.0, grad2, grad1)
-            (grad1, loss1 + loss2)
-          })
+      /** Given (current accumulated gradient, current loss) and (label, features)
+       * tuples, updates the current gradient and current loss
+       */
+      val seqOp = (c: (Vector, Double), v: (Double, Vector)) => {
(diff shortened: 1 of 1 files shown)


AnthonyTruchet:
@srowen Here is at last the real PR for SPARK-18471. 

Sorry for the noise due to GitHub fiddling...

mllib/src/main/scala/org/apache/spark/mllib/optimization/LBFGS.scala
```diff
@@ -241,16 +241,27 @@ object LBFGS extends Logging {
       val bcW = data.context.broadcast(w)
       val localGradient = gradient
 
-      val (gradientSum, lossSum) = data.treeAggregate((Vectors.zeros(n), 0.0))(
-          seqOp = (c, v) => (c, v) match { case ((grad, loss), (label, features)) =>
-            val l = localGradient.compute(
-              features, label, bcW.value, grad)
+      /** Given (current accumulated gradient, current loss) and (label, features)
```
srowen: Nit: just use `//` for two lines of comment. Really `/**` (vs `/*`) is for javadoc.

mllib/src/main/scala/org/apache/spark/mllib/optimization/LBFGS.scala
```diff
@@ -241,16 +241,27 @@ object LBFGS extends Logging {
...
+
+      // Adds two (gradient, loss) tuples
+      val combOp = (c1: (Vector, Double), c2: (Vector, Double)) =>
+        (c1, c2) match { case ((grad1, loss1), (grad2, loss2)) =>
             axpy(1.0, grad2, grad1)
             (grad1, loss1 + loss2)
-          })
+       }
+
+      val (gradientSum, lossSum) = data.mapPartitions { it => {
```
srowen: Nit: the second brace and its match are redundant

srowen (changes requested):
I think it's still worth looking into making a similar change elsewhere that seqOp and combOp are used this way, and which might also accidentally take something into the closure. But (modulo minor changes) I think this is an OK change.

srowen:
Ping @AnthonyTruchet

AnthonyTruchet:
Hello @srowen , sorry not to have updated this lately, I've been taken by other emergencies at work. I'll update this on Monday. Actually I'll submit a variant of treeAggregate in core that we will be able to use for other similar use case in ML(lib). 

I appreciate your care about those patches and will attempt to reach a reasonable reactivity too :-)

AnthonyTruchet:
Once more (last time hopefully) I mistakenly fiddled with PR. Closing this one and replace it with #16037.
Code style review above taken into account in new PR.

PR #15966, opened 2016-11-21
[SPARK-18413][SQL][FOLLOW-UP] Use `numPartitions` instead of `maxConnections`

dongjoon-hyun:

This is a follow-up PR of #15868 to merge `maxConnections` option into `numPartitions` options.

## How was this patch tested?

Pass the existing tests.

### DIFF
--- a/docs/sql-programming-guide.md
+++ b/docs/sql-programming-guide.md
@@ -1077,7 +1077,8 @@ the following case-sensitive options:
      <td>
        The number of partitions that can be used, if set. It works by limiting both read and write
        operations' parallelism. If the number of partitions to write exceeds this limit, the
-       operation will coalesce the data set to fewer partitions before writing.
+       operation will coalesce the data set to fewer partitions before writing. In other words,
+       this determines the maximum number of concurrent JDBC connections.
      </td>
   </tr>
 
--- a/docs/sql-programming-guide.md
+++ b/docs/sql-programming-guide.md
@@ -1061,7 +1061,7 @@ the following case-sensitive options:
   </tr>
 
   <tr>
-    <td><code>partitionColumn, lowerBound, upperBound, numPartitions</code></td>
+    <td><code>partitionColumn, lowerBound, upperBound</code></td>
     <td>
       These options must all be specified if any of them is specified. They describe how to
       partition the table when reading in parallel from multiple workers.
@@ -1070,6 +1070,15 @@ the following case-sensitive options:
       partition stride, not for filtering the rows in table. So all rows in the table will be
       partitioned and returned. This option applies only to reading.
     </td>
+  </tr>
+
+  <tr>
+     <td><code>numPartitions</code></td>
+     <td>
+       The number of partitions that can be used, if set. It works by limiting both read and write
+       operations' parallelism. If the number of partitions to write exceeds this limit, the
+       operation will coalesce the data set to fewer partitions before writing.
+     </td>
   </tr>
 
   <tr>
@@ -1083,13 +1092,6 @@ the following case-sensitive options:
      <td><code>batchsize</code></td>
      <td>
        The JDBC batch size, which determines how many rows to insert per round trip. This can help performance on JDBC drivers. This option applies only to writing. It defaults to <code>1000</code>.
-     </td>
-  </tr>
-
-  <tr>
-     <td><code>maxConnections</code></td>
-     <td>
-       The maximum number of concurrent JDBC connections that can be used, if set. Only applies when writing. It works by limiting the operation's parallelism, which depends on the input's partition count. If its partition count exceeds this limit, the operation will coalesce the input to fewer partitions before writing.
      </td>
   </tr>
 
--- a/docs/sql-programming-guide.md
+++ b/docs/sql-programming-guide.md
@@ -1076,11 +1076,10 @@ the following case-sensitive options:
   <tr>
      <td><code>numPartitions</code></td>
      <td>
-       The number of partitions that can be used, if set. It works by limiting both read and write
-       operations' parallelism. In other words, this determines the maximum number of concurrent
-       JDBC connections. For reading, it will make partitions less than or equal to this maximum.
-       For writing, if the number of partitions to write exceeds this limit, the operation will
-       coalesce the data set with this maximum before writing.
+       The maximum number of partitions that can be used for parallelism in table reading and
+       writing. This also determines the maximum number of concurrent JDBC connections.
+       If the number of partitions to write exceeds this limit, we decrease it to this limit by
+       calling <code>coalesce(numPartitions)</code> before writing.
      </td>
   </tr>
 
--- a/docs/sql-programming-guide.md
+++ b/docs/sql-programming-guide.md
@@ -1063,8 +1063,9 @@ the following case-sensitive options:
   <tr>
     <td><code>partitionColumn, lowerBound, upperBound</code></td>
     <td>
-      These options must all be specified if any of them is specified. They describe how to
-      partition the table when reading in parallel from multiple workers.
+      These options must all be specified if any of them is specified. In addition,
+      <code>numPartitions</code> must be specified. They describe how to partition the table when
+      reading in parallel from multiple workers.
       <code>partitionColumn</code> must be a numeric column from the table in question. Notice
       that <code>lowerBound</code> and <code>upperBound</code> are just used to decide the
       partition stride, not for filtering the rows in table. So all rows in the table will be
@@ -1076,9 +1077,10 @@ the following case-sensitive options:
      <td><code>numPartitions</code></td>
      <td>
        The number of partitions that can be used, if set. It works by limiting both read and write
-       operations' parallelism. If the number of partitions to write exceeds this limit, the
-       operation will coalesce the data set to fewer partitions before writing. In other words,
-       this determines the maximum number of concurrent JDBC connections.
+       operations' parallelism. In other words, this determines the maximum number of concurrent
+       JDBC connections. For reading, it will make partitions less than or equal to this maximum.
+       For writing, if the number of partitions to write exceeds this limit, the operation will
+       coalesce the data set with this maximum before writing.
      </td>
   </tr>
 
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JDBCOptions.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JDBCOptions.scala
@@ -74,6 +74,12 @@ class JDBCOptions(
     }
   }
 
+  // the number of partitions
+  val numPartitions = parameters.get(JDBC_NUM_PARTITIONS).map(_.toInt)
+  require(numPartitions.isEmpty || numPartitions.get > 0,
+    s"Invalid value `${numPartitions.get}` for parameter `$JDBC_NUM_PARTITIONS`. " +
+      "The minimum value is 1.")
+
   // ------------------------------------------------------------
   // Optional parameters only for reading
   // ------------------------------------------------------------
@@ -83,10 +89,8 @@ class JDBCOptions(
   val lowerBound = parameters.getOrElse(JDBC_LOWER_BOUND, null)
   // the upper bound of the partition column
   val upperBound = parameters.getOrElse(JDBC_UPPER_BOUND, null)
-  // the number of partitions
-  val numPartitions = parameters.getOrElse(JDBC_NUM_PARTITIONS, null)
   require(partitionColumn == null ||
-    (lowerBound != null && upperBound != null && numPartitions != null),
+    (lowerBound != null && upperBound != null && numPartitions.isDefined),
     s"If '$JDBC_PARTITION_COLUMN' is specified then '$JDBC_LOWER_BOUND', '$JDBC_UPPER_BOUND'," +
       s" and '$JDBC_NUM_PARTITIONS' are required.")
   val fetchSize = {
@@ -122,11 +126,6 @@ class JDBCOptions(
       case "REPEATABLE_READ" => Connection.TRANSACTION_REPEATABLE_READ
       case "SERIALIZABLE" => Connection.TRANSACTION_SERIALIZABLE
     }
-  // the maximum number of connections
-  val maxConnections = parameters.get(JDBC_MAX_CONNECTIONS).map(_.toInt)
-  require(maxConnections.isEmpty || maxConnections.get > 0,
-    s"Invalid value `${maxConnections.get}` for parameter `$JDBC_MAX_CONNECTIONS`. " +
-      "The minimum value is 1.")
 }
 
 object JDBCOptions {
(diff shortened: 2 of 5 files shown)


docs/sql-programming-guide.md
```diff
@@ -1073,6 +1073,15 @@ the following case-sensitive options:
   </tr>
 
   <tr>
+     <td><code>numPartitions</code></td>
+     <td>
+       The number of partitions that can be used, if set. It works by limiting both read and write
+       operations' parallelism. If the number of partitions to write exceeds this limit, the
+       operation will coalesce the data set to fewer partitions before writing.
```
gatorsmile: Could you also document the impact on the number of JDBC connections here?

dongjoon-hyun: Sure. I'll add that, too.

docs/sql-programming-guide.md
```diff
@@ -1073,6 +1073,16 @@ the following case-sensitive options:
   </tr>
 
   <tr>
+     <td><code>numPartitions</code></td>
+     <td>
+       The number of partitions that can be used, if set. It works by limiting both read and write
+       operations' parallelism. If the number of partitions to write exceeds this limit, the
+       operation will coalesce the data set to fewer partitions before writing. In other words,
```
dongjoon-hyun: Thank you for review again. Yep. I'll revise it. :)

dongjoon-hyun: > The number of partitions that can be used, if set. It works by limiting both read and write operations' parallelism. If the number of partitions to write exceeds this limit, the operation will coalesce the data set **with this value** before writing. In other words, this determines the maximum number of concurrent JDBC connections.

docs/sql-programming-guide.md
```diff
@@ -1061,7 +1061,7 @@ the following case-sensitive options:
   </tr>
 
   <tr>
-    <td><code>partitionColumn, lowerBound, upperBound, numPartitions</code></td>
```
gatorsmile: We should revert it back. These four parameters are related. 

`These options must all be specified if any of them is specified`

dongjoon-hyun: That is incorrect now. `numPartitions` can be used alone.

gatorsmile: In the [read path](https://github.com/dongjoon-hyun/spark/blob/48b6d2513ddebcf365999c8567a7e6dbcc5a3bb2/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JDBCOptions.scala#L92-L93), they are still related.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JDBCOptions.scala
```diff
@@ -83,10 +89,8 @@ class JDBCOptions(
   val lowerBound = parameters.getOrElse(JDBC_LOWER_BOUND, null)
   // the upper bound of the partition column
   val upperBound = parameters.getOrElse(JDBC_UPPER_BOUND, null)
-  // the number of partitions
-  val numPartitions = parameters.getOrElse(JDBC_NUM_PARTITIONS, null)
   require(partitionColumn == null ||
-    (lowerBound != null && upperBound != null && numPartitions != null),
+    (lowerBound != null && upperBound != null && numPartitions.isDefined),
     s"If '$JDBC_PARTITION_COLUMN' is specified then '$JDBC_LOWER_BOUND', '$JDBC_UPPER_BOUND'," +
       s" and '$JDBC_NUM_PARTITIONS' are required.")
```
gatorsmile: We need to update this error message too.

docs/sql-programming-guide.md
dongjoon-hyun: For that, I'll add some other description mentioning the relations.

docs/sql-programming-guide.md
gatorsmile: In the read path, we might not generate the exact number of partitions.

Found a bug in the read path.

dongjoon-hyun: Oh really? The number of `JDBCPartition` is different?

gatorsmile: Conceptually, if the column values are exactly identical, only one partition will be generated. 

The bug in https://github.com/apache/spark/pull/15499 is not related the above fact.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JDBCOptions.scala
gatorsmile: I did not try it. What happened if we input `JDBC_PARTITION_COLUMN ` when writing the JDBC tables?

docs/sql-programming-guide.md
gatorsmile: Maybe we need to rephrase it and document both behaviors (read and write paths). Also emphasize this is the max number.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JDBCOptions.scala
dongjoon-hyun: It will raise exception due to the `require`. I think it's the correct previous behavior.

docs/sql-programming-guide.md
dongjoon-hyun: Like this?
> The number of partitions that can be used, if set. It works by limiting both read and write operations' parallelism. In other words, this determines the maximum number of concurrent JDBC connections. For reading, it will make partitions less than or equal to this maximum. For writing, if the number of partitions to write exceeds this limit, the operation will coalesce the data set with this value before writing.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JDBCOptions.scala
dongjoon-hyun: Since it's used in declaration of view, users cannot go to writing path.
```sql
CREATE OR REPLACE TEMPORARY VIEW t1 USING org.apache.spark.sql.jdbc OPTIONS (url 'jdbc:mysql://localhost:3306/t', dbtable 't1', user 'root', password '', maxConnections '2', partitionColumn 'value')
java.lang.IllegalArgumentException: requirement failed: If 'partitionColumn' is specified then 'lowerBound', 'upperBound', and 'numPartitions' are required.
```

dongjoon-hyun:
To further review, I updated the doc. We can proceed on the updated content.

docs/sql-programming-guide.md
HyukjinKwon: (I am sorry I goofed in the PR.)

cloud-fan:
Is it possible that users read a table through JDBC and then write data to a table through JDBC and want the read and write have different parallelism?

dongjoon-hyun:
Thank you for review, @cloud-fan . 

With the same parameter name `numPartitions` for read/write, we will use the same parallelism by default. It's easy to use. 

To use different parallelisms for read/write, we are able to do that with different view names. In the following example, t1 is `numPartitions=1` and t2 is `numPartitions=2`. It's an example for writing, but I think the situation is the same with read operation.

```scala
sql("CREATE OR REPLACE TEMPORARY VIEW data USING org.apache.spark.sql.jdbc OPTIONS (url 'jdbc:mysql://localhost:3306/t', dbtable 'data', user 'root', password '')")
sql("CREATE OR REPLACE TEMPORARY VIEW t1 USING org.apache.spark.sql.jdbc OPTIONS (url 'jdbc:mysql://localhost:3306/t', dbtable 't1', user 'root', password '', numPartitions '1')")
sql("CREATE OR REPLACE TEMPORARY VIEW t2 USING org.apache.spark.sql.jdbc OPTIONS (url 'jdbc:mysql://localhost:3306/t', dbtable 't1', user 'root', password '', numPartitions '2')")
sql("INSERT OVERWRITE TABLE t1 SELECT a FROM data GROUP BY a")
sql("INSERT OVERWRITE TABLE t2 SELECT a FROM data GROUP BY a")
```

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JDBCOptions.scala
gatorsmile: Have you tried the `DataFrameWriter`'s JDBC/write() APIs?

gatorsmile:
I found a way to verify the coalesce logics of JDBC writing. See my PR: https://github.com/apache/spark/pull/15975 It added `numPartition` into `JDBCRelation`. 

With minor code changes, you can see the adjusted `numPartition` in the output of `EXPLAIN`. 
```Scala
sql("INSERT INTO TABLE PEOPLE1 SELECT * FROM PEOPLE").explain(true)
```

docs/sql-programming-guide.md
```diff
@@ -1073,6 +1074,17 @@ the following case-sensitive options:
   </tr>
 
   <tr>
+     <td><code>numPartitions</code></td>
+     <td>
+       The number of partitions that can be used, if set. It works by limiting both read and write
+       operations' parallelism. In other words, this determines the maximum number of concurrent
+       JDBC connections. For reading, it will make partitions less than or equal to this maximum.
+       For writing, if the number of partitions to write exceeds this limit, the operation will
+       coalesce the data set with this maximum before writing.
```
gatorsmile: Note, I am not good at writing tech document. Below is my revision.
> The maximum number of partitions that can be used for parallelism in table reading and writing. This also determines the maximum number of concurrent JDBC connections. If the number of partitions to write exceeds this limit, we decrease it to this limit by calling <code>coalesce(numPartitions)</code> before writing.

dongjoon-hyun:
Great! Thank you for #15975.

gatorsmile:
LGTM pending test. 

Also cc @rxin

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcRelationProvider.scala
```diff
@@ -39,7 +39,7 @@ class JdbcRelationProvider extends CreatableRelationProvider
       null
     } else {
       JDBCPartitioningInfo(
-        partitionColumn, lowerBound.toLong, upperBound.toLong, numPartitions.toInt)
+        partitionColumn, lowerBound.toLong, upperBound.toLong, numPartitions.get)
```
rxin: is this safe to call get on? calling get on an option is always dangerous and not future proof.

dongjoon-hyun: Thank you for review, @rxin . Yes. It implicitly depends on the partitionColumn requires all others columns.
I'll change like this. Is it better?
```
-    val partitionInfo = if (partitionColumn == null) {
+    val partitionInfo = if (partitionColumn == null || jdbcOptions == null || upperBound == null ||
+        numPartitions.isEmpty) {
```

dongjoon-hyun:
The only one failure seems to be irrelevant to this.
```
- SPARK-8020: set sql conf in spark conf *** FAILED *** (17 seconds, 602 milliseconds)
```

dongjoon-hyun:
Hi, @rxin .
Could you review this again?

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcRelationProvider.scala
rxin: If you are doing this, how about changing partitionColumn, lowerBound, upperBound to Option types as well?

dongjoon-hyun:
Hi, @rxin , @cloud-fan , @gatorsmile .
Please let me know if there is something to do more.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcRelationProvider.scala
dongjoon-hyun: Sure! No problem. I'll update them tonight.

dongjoon-hyun:
The only one failure is irrelevant to this PR.
```
[info] KafkaSourceStressForDontFailOnDataLossSuite:
[info] - stress test for failOnDataLoss=false *** FAILED *** (1 minute, 58 seconds)
```

dongjoon-hyun:
Retest this please.

dongjoon-hyun:
Maybe something internal errors.
```
    main()
KeyError: -9
```

dongjoon-hyun:
Retest this please.

cloud-fan:
LGTM

dongjoon-hyun:
Thank you, @cloud-fan !

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcRelationProvider.scala
```diff
@@ -35,11 +35,12 @@ class JdbcRelationProvider extends CreatableRelationProvider
     val upperBound = jdbcOptions.upperBound
     val numPartitions = jdbcOptions.numPartitions
 
-    val partitionInfo = if (partitionColumn == null) {
+    val partitionInfo = if (partitionColumn.isEmpty || lowerBound.isEmpty || upperBound.isEmpty ||
```
rxin: I'd change this to 

```
if (partitionColumn == null) {
  assert(lowerBound.isEmpty && upperBound.isEmpty && numPartitions.isEmpty)
  null
} else {
  ...
}
```

to be future proof.

rxin:
LGTM other than that one change.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcRelationProvider.scala
dongjoon-hyun: Thanks. I'll update soon.

dongjoon-hyun: I assumed the following.
```
if (partitionColumn.isEmpty) {
  assert(lowerBound.isEmpty && upperBound.isEmpty && numPartitions.isEmpty)
  null
} else {
  ...
}
```

dongjoon-hyun: Ur, @rxin .
We are using `numPartition` for both writing and reading, and `numPartition` can be used alone.

dongjoon-hyun: I'll use the following.
```scala
    val partitionInfo = if (partitionColumn.isEmpty) {
      assert(lowerBound.isEmpty && upperBound.isEmpty)
      null
    } else {
      assert(lowerBound.nonEmpty && upperBound.nonEmpty && numPartitions.nonEmpty)
      JDBCPartitioningInfo(
        partitionColumn.get, lowerBound.get, upperBound.get, numPartitions.get)
    }
```

rxin:
Thanks - merging in master.

dongjoon-hyun:
Thank you for review and merging, @rxin , @gatorsmile , @cloud-fan !

PR #15969, opened 2016-11-21
[SPARK-18530][SS][KAFKA]Change Kafka timestamp column type to TimestampType

zsxwing:

Changed Kafka timestamp column type to TimestampType.

## How was this patch tested?

`test("Kafka column types")`.

### DIFF
--- a/external/kafka-0-10-sql/src/test/scala/org/apache/spark/sql/kafka010/KafkaSourceSuite.scala
+++ b/external/kafka-0-10-sql/src/test/scala/org/apache/spark/sql/kafka010/KafkaSourceSuite.scala
@@ -29,6 +29,7 @@ import org.scalatest.concurrent.PatienceConfiguration.Timeout
 import org.scalatest.time.SpanSugar._
 
 import org.apache.spark.sql.execution.streaming._
+import org.apache.spark.sql.functions.{count, window}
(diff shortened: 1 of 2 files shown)


zsxwing:
cc @marmbrus @tdas

external/kafka-0-10-sql/src/test/scala/org/apache/spark/sql/kafka010/KafkaSourceSuite.scala
```diff
@@ -413,6 +414,43 @@ class KafkaSourceSuite extends KafkaSourceTest {
...
+
+    val reader = spark
+      .readStream
+      .format("kafka")
+      .option("kafka.bootstrap.servers", testUtils.brokerAddress)
+      .option("kafka.metadata.max.age.ms", "1")
+      .option("startingOffsets", s"earliest")
+      .option("subscribe", topic)
+
+    val kafka = reader.load()
```
tdas: nit: why have separate var for reader

tdas:
I wonder whether it makes sense to actually test with a query that has window and withWatermark on the timestamp column. because that is what we want to enable.

tdas:
LGTM, pending tests.

zsxwing:
@tdas the test already finished. Thanks for reviewing. Merging to master and 2.1.

PR #15970, opened 2016-11-21
[SPARK-18134][SQL] Comparable MapTypes [POC]

hvanhovell:
This is a small POC to see if we can make MapType orderable, and thus usable in aggregates and joins. The key idea in this PR is that there is a difference between an unordered and an ordered map (an ordered map is can be compared), and that `ordered` is a property of `MapType`.

A map can be converted from an unordered map to an ordered map by injecting a `SortMap` expression. The analyzer will inject `SortMap` expressions whenever we use a map in a binary comparison and when we use it in an aggregate. Note that the `SortMap` expression is far from optimized, it should however perform reasonable.

## How was this patch tested?
No tests yet. This will probably fail tests.


hvanhovell:
cc @cloud-fan we discussed something like this earlier today

albertferras:
Any update on this?

gatorsmile:
cc @hvanhovell

janewangfb:
Is there any updates on this PR? This feature is quite useful to us.

jinxing64:
@hvanhovell  Are you still working on this? I think this is feature is useful :)

PR #15971, opened 2016-11-22
[SPARK-18535][UI][YARN] Redact sensitive information from Spark logs and UI

markgrover:

This patch adds a new property called `spark.secret.redactionPattern` that
allows users to specify a scala regex to decide which Spark configuration
properties and environment variables in driver and executor environments
contain sensitive information. When this regex matches the property or
environment variable name, its value is redacted from the environment UI and
various logs like YARN and event logs.

This change uses this property to redact information from event logs and YARN
logs. It also, updates the UI code to adhere to this property instead of
hardcoding the logic to decipher which properties are sensitive.

Here's an image of the UI post-redaction:
image

Here's the text in the YARN logs, post-redaction:
``HADOOP_CREDSTORE_PASSWORD -> *********(redacted)``

Here's the text in the event logs, post-redaction:
``...,"spark.executorEnv.HADOOP_CREDSTORE_PASSWORD":"*********(redacted)","spark.yarn.appMasterEnv.HADOOP_CREDSTORE_PASSWORD":"*********(redacted)",...``

## How was this patch tested?
1. Unit tests are added to ensure that redaction works.
2. A YARN job reading data off of S3 with confidential information
(hadoop credential provider password) being provided in the environment
variables of driver and executor. And, afterwards, logs were grepped to make
sure that no mention of secret password was present. It was also ensure that
the job was able to read the data off of S3 correctly, thereby ensuring that
the sensitive information was being trickled down to the right places to read
the data.
3. The event logs were checked to make sure no mention of secret password was
present.
4. UI environment tab was checked to make sure there was no secret information
being displayed.

### DIFF
--- a/core/src/main/scala/org/apache/spark/internal/config/package.scala
+++ b/core/src/main/scala/org/apache/spark/internal/config/package.scala
@@ -223,4 +223,13 @@ package object config {
       " bigger files.")
     .longConf
     .createWithDefault(4 * 1024 * 1024)
+
+  private[spark] val SECRET_REDACTION_PATTERN =
+    ConfigBuilder("spark.secret.redactionPattern")
+      .doc("Scala regex(case-sensitive) to decide which Spark configuration properties and " +
+        "environment variables in driver and executor environments contain sensitive information." +
+        " When this regex matches the property or environment variable name, its value is " +
+        "redacted from the environment UI and various logs like YARN and event logs")
+      .stringConf
+      .createWithDefault("secret|password|SECRET|PASSWORD")
 }
--- a/core/src/main/scala/org/apache/spark/internal/config/package.scala
+++ b/core/src/main/scala/org/apache/spark/internal/config/package.scala
@@ -225,11 +225,11 @@ package object config {
     .createWithDefault(4 * 1024 * 1024)
 
   private[spark] val SECRET_REDACTION_PATTERN =
-    ConfigBuilder("spark.secret.redactionPattern")
-      .doc("Scala regex(case-sensitive) to decide which Spark configuration properties and " +
-        "environment variables in driver and executor environments contain sensitive information." +
-        " When this regex matches the property or environment variable name, its value is " +
-        "redacted from the environment UI and various logs like YARN and event logs")
+    ConfigBuilder("spark.redacton.regex")
+      .doc("Regex to decide which Spark configuration properties and environment variables in " +
+        "driver and executor environments contain sensitive information. When this regex matches " +
+        "a property , its value is redacted from the environment UI and various logs like YARN " +
+        "and event logs")
       .stringConf
-      .createWithDefault("secret|password|SECRET|PASSWORD")
+      .createWithDefault("(?i)secret|password")
 }
--- a/core/src/main/scala/org/apache/spark/internal/config/package.scala
+++ b/core/src/main/scala/org/apache/spark/internal/config/package.scala
@@ -228,8 +228,8 @@ package object config {
     ConfigBuilder("spark.redaction.regex")
       .doc("Regex to decide which Spark configuration properties and environment variables in " +
         "driver and executor environments contain sensitive information. When this regex matches " +
-        "a property , its value is redacted from the environment UI and various logs like YARN " +
-        "and event logs")
+        "a property, its value is redacted from the environment UI and various logs like YARN " +
+        "and event logs.")
       .stringConf
       .createWithDefault("(?i)secret|password")
 }
--- a/core/src/main/scala/org/apache/spark/internal/config/package.scala
+++ b/core/src/main/scala/org/apache/spark/internal/config/package.scala
@@ -225,7 +225,7 @@ package object config {
     .createWithDefault(4 * 1024 * 1024)
 
   private[spark] val SECRET_REDACTION_PATTERN =
-    ConfigBuilder("spark.redacton.regex")
+    ConfigBuilder("spark.redaction.regex")
       .doc("Regex to decide which Spark configuration properties and environment variables in " +
         "driver and executor environments contain sensitive information. When this regex matches " +
         "a property , its value is redacted from the environment UI and various logs like YARN " +
--- a/core/src/main/scala/org/apache/spark/scheduler/EventLoggingListener.scala
+++ b/core/src/main/scala/org/apache/spark/scheduler/EventLoggingListener.scala
@@ -233,13 +233,10 @@ private[spark] class EventLoggingListener(
     }
   }
 
-
   private def redactEvent(event: SparkListenerEnvironmentUpdate): SparkListenerEnvironmentUpdate = {
     // "Spark Properties" entry will always exist because the map is always populated with it.
     val props = event
-      .environmentDetails
-      .get("Spark Properties")
-      .get
+      .environmentDetails("Spark Properties")
     val redactedProps = Utils.redact(sparkConf, props)
     val redactedEnvironmentDetails = event.environmentDetails +
       ("Spark Properties" -> redactedProps)
--- a/core/src/main/scala/org/apache/spark/scheduler/EventLoggingListener.scala
+++ b/core/src/main/scala/org/apache/spark/scheduler/EventLoggingListener.scala
@@ -233,7 +233,8 @@ private[spark] class EventLoggingListener(
     }
   }
 
-  private def redactEvent(event: SparkListenerEnvironmentUpdate): SparkListenerEnvironmentUpdate = {
+  private[spark] def redactEvent(event: SparkListenerEnvironmentUpdate):
+  SparkListenerEnvironmentUpdate = {
     // "Spark Properties" entry will always exist because the map is always populated with it.
     val props = event
       .environmentDetails("Spark Properties")
--- a/core/src/main/scala/org/apache/spark/scheduler/EventLoggingListener.scala
+++ b/core/src/main/scala/org/apache/spark/scheduler/EventLoggingListener.scala
@@ -153,7 +153,9 @@ private[spark] class EventLoggingListener(
 
   override def onTaskEnd(event: SparkListenerTaskEnd): Unit = logEvent(event)
 
-  override def onEnvironmentUpdate(event: SparkListenerEnvironmentUpdate): Unit = logEvent(event)
+  override def onEnvironmentUpdate(event: SparkListenerEnvironmentUpdate): Unit = {
+    logEvent(redactEvent(event))
+  }
 
   // Events that trigger a flush
   override def onStageCompleted(event: SparkListenerStageCompleted): Unit = {
@@ -229,6 +231,19 @@ private[spark] class EventLoggingListener(
     } catch {
       case e: Exception => logDebug(s"failed to set time of $target", e)
     }
+  }
+
+
+  private def redactEvent(event: SparkListenerEnvironmentUpdate): SparkListenerEnvironmentUpdate = {
+    // "Spark Properties" entry will always exist because the map is always populated with it.
+    val redactedProps = event
+      .environmentDetails
+      .get("Spark Properties")
+      .get
+      .map(Utils.redact(sparkConf))
+    val redactedEnvironmentDetails = event.environmentDetails +
+      ("Spark Properties" -> redactedProps)
+    SparkListenerEnvironmentUpdate(redactedEnvironmentDetails)
   }
 
 }
--- a/core/src/main/scala/org/apache/spark/scheduler/EventLoggingListener.scala
+++ b/core/src/main/scala/org/apache/spark/scheduler/EventLoggingListener.scala
@@ -236,11 +236,11 @@ private[spark] class EventLoggingListener(
 
   private def redactEvent(event: SparkListenerEnvironmentUpdate): SparkListenerEnvironmentUpdate = {
     // "Spark Properties" entry will always exist because the map is always populated with it.
-    val redactedProps = event
+    val props = event
       .environmentDetails
       .get("Spark Properties")
       .get
-      .map(Utils.redact(sparkConf))
+    val redactedProps = Utils.redact(sparkConf, props)
     val redactedEnvironmentDetails = event.environmentDetails +
       ("Spark Properties" -> redactedProps)
     SparkListenerEnvironmentUpdate(redactedEnvironmentDetails)
--- a/core/src/main/scala/org/apache/spark/scheduler/EventLoggingListener.scala
+++ b/core/src/main/scala/org/apache/spark/scheduler/EventLoggingListener.scala
@@ -234,7 +234,7 @@ private[spark] class EventLoggingListener(
   }
 
   private[spark] def redactEvent(
-    event: SparkListenerEnvironmentUpdate): SparkListenerEnvironmentUpdate = {
+      event: SparkListenerEnvironmentUpdate): SparkListenerEnvironmentUpdate = {
     // "Spark Properties" entry will always exist because the map is always populated with it.
     val redactedProps = Utils.redact(sparkConf, event.environmentDetails("Spark Properties"))
     val redactedEnvironmentDetails = event.environmentDetails +
--- a/core/src/main/scala/org/apache/spark/scheduler/EventLoggingListener.scala
+++ b/core/src/main/scala/org/apache/spark/scheduler/EventLoggingListener.scala
@@ -233,12 +233,10 @@ private[spark] class EventLoggingListener(
     }
   }
 
-  private[spark] def redactEvent(event: SparkListenerEnvironmentUpdate):
-  SparkListenerEnvironmentUpdate = {
+  private[spark] def redactEvent(
+    event: SparkListenerEnvironmentUpdate): SparkListenerEnvironmentUpdate = {
     // "Spark Properties" entry will always exist because the map is always populated with it.
-    val props = event
-      .environmentDetails("Spark Properties")
-    val redactedProps = Utils.redact(sparkConf, props)
+    val redactedProps = Utils.redact(sparkConf, event.environmentDetails("Spark Properties"))
     val redactedEnvironmentDetails = event.environmentDetails +
       ("Spark Properties" -> redactedProps)
     SparkListenerEnvironmentUpdate(redactedEnvironmentDetails)
--- a/core/src/main/scala/org/apache/spark/util/Utils.scala
+++ b/core/src/main/scala/org/apache/spark/util/Utils.scala
@@ -2558,11 +2558,14 @@ private[spark] object Utils extends Logging {
 
   private[util] val REDACTION_REPLACEMENT_TEXT = "*********(redacted)"
 
-  def redact(conf: SparkConf)(kv: (String, String)): (String, String) = {
+  def redact(conf: SparkConf, kvs: Seq[(String, String)]): Seq[(String, String)] = {
     val redactionPattern = conf.get(SECRET_REDACTION_PATTERN).r
-    if (redactionPattern.findFirstIn(kv._1).isDefined) {
-      (kv._1, REDACTION_REPLACEMENT_TEXT)
-    } else kv
+    kvs.map { kv =>
+      if (redactionPattern.findFirstIn(kv._1).isDefined) {
+        (kv._1, REDACTION_REPLACEMENT_TEXT)
+      }
+      else kv
+    }
   }
 
 }
--- a/core/src/main/scala/org/apache/spark/util/Utils.scala
+++ b/core/src/main/scala/org/apache/spark/util/Utils.scala
@@ -55,7 +55,7 @@ import org.slf4j.Logger
 import org.apache.spark._
 import org.apache.spark.deploy.SparkHadoopUtil
 import org.apache.spark.internal.Logging
-import org.apache.spark.internal.config.{DYN_ALLOCATION_INITIAL_EXECUTORS, DYN_ALLOCATION_MIN_EXECUTORS, EXECUTOR_INSTANCES}
+import org.apache.spark.internal.config._
 import org.apache.spark.network.util.JavaUtils
 import org.apache.spark.serializer.{DeserializationStream, SerializationStream, SerializerInstance}
 import org.apache.spark.util.logging.RollingFileAppender
@@ -2555,6 +2555,15 @@ private[spark] object Utils extends Logging {
       sparkJars.map(_.split(",")).map(_.filter(_.nonEmpty)).toSeq.flatten
     }
   }
+
+  private[util] val REDACTION_REPLACEMENT_TEXT = "*********(redacted)"
+  def redact(conf: SparkConf)(kv: (String, String)): (String, String) = {
+    val redactionPattern = conf.get(SECRET_REDACTION_PATTERN).r
+    if (redactionPattern.findFirstIn(kv._1).isDefined) {
+      (kv._1, REDACTION_REPLACEMENT_TEXT)
+    } else kv
+  }
+
 }
 
 private[util] object CallerContext extends Logging {
--- a/core/src/main/scala/org/apache/spark/util/Utils.scala
+++ b/core/src/main/scala/org/apache/spark/util/Utils.scala
@@ -2562,9 +2562,9 @@ private[spark] object Utils extends Logging {
     val redactionPattern = conf.get(SECRET_REDACTION_PATTERN).r
     kvs.map { kv =>
       redactionPattern.findFirstIn(kv._1)
-        .map{ ignore => (kv._1, REDACTION_REPLACEMENT_TEXT) }
+        .map { ignore => (kv._1, REDACTION_REPLACEMENT_TEXT) }
         .getOrElse(kv)
-      }
+    }
   }
 
 }
--- a/core/src/main/scala/org/apache/spark/util/Utils.scala
+++ b/core/src/main/scala/org/apache/spark/util/Utils.scala
@@ -2557,6 +2557,7 @@ private[spark] object Utils extends Logging {
   }
 
   private[util] val REDACTION_REPLACEMENT_TEXT = "*********(redacted)"
+
   def redact(conf: SparkConf)(kv: (String, String)): (String, String) = {
     val redactionPattern = conf.get(SECRET_REDACTION_PATTERN).r
     if (redactionPattern.findFirstIn(kv._1).isDefined) {
--- a/core/src/main/scala/org/apache/spark/util/Utils.scala
+++ b/core/src/main/scala/org/apache/spark/util/Utils.scala
@@ -2561,11 +2561,10 @@ private[spark] object Utils extends Logging {
   def redact(conf: SparkConf, kvs: Seq[(String, String)]): Seq[(String, String)] = {
     val redactionPattern = conf.get(SECRET_REDACTION_PATTERN).r
     kvs.map { kv =>
-      if (redactionPattern.findFirstIn(kv._1).isDefined) {
-        (kv._1, REDACTION_REPLACEMENT_TEXT)
+      redactionPattern.findFirstIn(kv._1)
+        .map{ ignore => (kv._1, REDACTION_REPLACEMENT_TEXT) }
+        .getOrElse(kv)
       }
-      else kv
-    }
   }
 
 }
--- a/core/src/test/scala/org/apache/spark/scheduler/EventLoggingListenerSuite.scala
+++ b/core/src/test/scala/org/apache/spark/scheduler/EventLoggingListenerSuite.scala
@@ -95,6 +95,30 @@ class EventLoggingListenerSuite extends SparkFunSuite with LocalSparkContext wit
     }
   }
 
+  test("Event logging with password redaction") {
+    val secretPassword = "secret_password"
+    val conf = getLoggingConf(testDirPath, None).set("spark.executorEnv.HADOOP_CREDSTORE_PASSWORD",
+      secretPassword)
+    sc = new SparkContext("local-cluster[2,2,1024]", "test", conf)
+    assert(sc.eventLogger.isDefined)
+    val eventLogger = sc.eventLogger.get
+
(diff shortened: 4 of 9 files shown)


tejasapatil:
Does this protect against doing `sc.getConf.getAll.foreach(println)` in spark shell ?

markgrover:
> Does this protect against doing sc.getConf.getAll.foreach(println) in spark shell ?
No, it doesn't. The goal of the JIRA is the first step - to stop printing sensitive information all over the place.

If a user has access to spark-shell (say, in a cluster secured through kerberos), it may be reasonable for them to see the sensitive information.

This, however, prevents anyone from going to the unauthenticated Spark UI, and having all the creds to S3.

core/src/main/scala/org/apache/spark/internal/config/package.scala
```diff
@@ -223,4 +223,13 @@ package object config {
       " bigger files.")
     .longConf
     .createWithDefault(4 * 1024 * 1024)
+
+  private[spark] val SECRET_REDACTION_PATTERN =
+    ConfigBuilder("spark.secret.redactionPattern")
```
ajbozarth: wouldn't this mean this pattern would get redacted since it contains secret?

core/src/main/scala/org/apache/spark/internal/config/package.scala
```diff
@@ -223,4 +223,13 @@ package object config {
...
+      .doc("Scala regex(case-sensitive) to decide which Spark configuration properties and " +
+        "environment variables in driver and executor environments contain sensitive information." +
+        " When this regex matches the property or environment variable name, its value is " +
+        "redacted from the environment UI and various logs like YARN and event logs")
+      .stringConf
+      .createWithDefault("secret|password|SECRET|PASSWORD")
```
ajbozarth: would a case-insensitive version be better?

ajbozarth (commented):
This is a good upgrade over the last fix. I made a couple notes but overall I think this looks pretty good.

core/src/test/scala/org/apache/spark/util/UtilsSuite.scala
```diff
@@ -974,4 +974,18 @@ class UtilsSuite extends SparkFunSuite with ResetSystemProperties with Logging {
 
     assert(pValue > threshold)
   }
+
+  test("redact sensitive information") {
+    val sparkConf = new SparkConf
+    sparkConf.set("spark.executorEnv.HADOOP_CREDSTORE_PASSWORD", "secret_password")
```
vanzin: Much cleaner if you do something like:

```
val keys = Seq("spark.executorEnv.HADOOP_CREDSTORE_PASSWORD", "spark.my.password", ...)
keys.foreach { key =>
  // test redaction for that key
}
```

core/src/test/scala/org/apache/spark/scheduler/EventLoggingListenerSuite.scala
```diff
@@ -95,6 +95,30 @@ class EventLoggingListenerSuite extends SparkFunSuite with LocalSparkContext wit
...
+
+    sc.parallelize(1 to 10000).count()
+    sc.stop()
+
+    val logData = EventLoggingListener.openEventLog(new Path(eventLogger.logPath), fileSystem)
+    val eventLog = Source.fromInputStream(logData).mkString
+    // Make sure nothing secret shows up anywhere
+    assert(!eventLog.contains(secretPassword), s"Secret password ($secretPassword) not redacted " +
+      s"from event logs:\n $eventLog")
+    val expected = """"spark.executorEnv.HADOOP_CREDSTORE_PASSWORD":"*********(redacted)""""
```
vanzin: This is pretty hacky. If makes assumptions about how things are formatted in the event log.

The previous assert should be enough (ignoring my previous comment about changing this test).

core/src/test/scala/org/apache/spark/scheduler/EventLoggingListenerSuite.scala
```diff
@@ -95,6 +95,30 @@ class EventLoggingListenerSuite extends SparkFunSuite with LocalSparkContext wit
...
+    // Make sure every occurrence of the property is accompanied by a redaction text.
+    val regex = """"spark.executorEnv.HADOOP_CREDSTORE_PASSWORD":"([^"]*)"""".r
+    val matches = regex.findAllIn(eventLog)
+    assert(matches.nonEmpty)
+    matches.foreach(matched => assert(matched.equals(expected)))
```
vanzin: Ignoring my previous comments, this should be `.foreach { matched => ... }`

core/src/main/scala/org/apache/spark/internal/config/package.scala
vanzin: +1

core/src/test/scala/org/apache/spark/scheduler/EventLoggingListenerSuite.scala
```diff
@@ -95,6 +95,30 @@ class EventLoggingListenerSuite extends SparkFunSuite with LocalSparkContext wit
     }
   }
 
+  test("Event logging with password redaction") {
+    val secretPassword = "secret_password"
+    val conf = getLoggingConf(testDirPath, None).set("spark.executorEnv.HADOOP_CREDSTORE_PASSWORD",
+      secretPassword)
+    sc = new SparkContext("local-cluster[2,2,1024]", "test", conf)
```
vanzin: This is a pretty expensive way of testing this. Why not just call the redaction method and make sure it's doing the right thing?

core/src/main/scala/org/apache/spark/internal/config/package.scala
markgrover: Good point! I think it's good for the actual regex to not be redacted. So, I will rename the property to be clearer anyways (and not have secret in the name) to `spark.redaction.regex`.

core/src/main/scala/org/apache/spark/internal/config/package.scala
```diff
@@ -223,4 +223,13 @@ package object config {
       " bigger files.")
     .longConf
     .createWithDefault(4 * 1024 * 1024)
+
+  private[spark] val SECRET_REDACTION_PATTERN =
+    ConfigBuilder("spark.secret.redactionPattern")
+      .doc("Scala regex(case-sensitive) to decide which Spark configuration properties and " +
```
markgrover: Fixed.

core/src/main/scala/org/apache/spark/internal/config/package.scala
```diff
@@ -223,4 +223,13 @@ package object config {
...
+        "environment variables in driver and executor environments contain sensitive information." +
+        " When this regex matches the property or environment variable name, its value is " +
```
markgrover: Fixed.

core/src/main/scala/org/apache/spark/internal/config/package.scala
markgrover: Fixed.

core/src/main/scala/org/apache/spark/scheduler/EventLoggingListener.scala
```diff
@@ -231,6 +233,19 @@ private[spark] class EventLoggingListener(
     }
   }
 
+
+  private def redactEvent(event: SparkListenerEnvironmentUpdate): SparkListenerEnvironmentUpdate = {
```
markgrover: Good question. I thought about this quite a bit when making this change. I should have posted that decision in the PR description, apologies for not providing that context. The way I see it - there are three places where redaction can be done:
1. Right at the source of SparkListenerEnvironmentUpdate ([here, in SparkContext.scala](https://github.com/apache/spark/blob/39a1d30636857715247c82d551b200e1c331ad69/core/src/main/scala/org/apache/spark/SparkContext.scala#L2214)).
2. [In JsonProtocol.scala](https://github.com/apache/spark/blob/39a1d30636857715247c82d551b200e1c331ad69/core/src/main/scala/org/apache/spark/util/JsonProtocol.scala#L167) when converting the event to JSON.
3. In [EventLogginListener.scala](https://github.com/apache/spark/blob/39a1d30636857715247c82d551b200e1c331ad69/core/src/main/scala/org/apache/spark/util/JsonProtocol.scala#L167) (where it is right now), when being persisted to disk.

A user could write a custom listener that listened to the environment updates and did something useful with them. And, I didn't want to impose redaction on them. They could be using it to create a clone of their environment, for example and may need to the same sensitive properties. So, I ruled out 1.

And, JsonProtocol seemed like a generic utility to convert events to JSON. While I could be selective about only redacting events of `SparkListenerEnvironmentUpdate` type, I didn't want to assume that everyone translating the environment update to JSON should only be seeing redacted configuration. So, that made me rule out 2.

I decided that it was best redact "closest to disk", which made me put the redaction code where I did - in EventLoggingListener. Hope that makes sense, happy to hear your thoughts if you think otherwise.

core/src/main/scala/org/apache/spark/util/Utils.scala
```diff
@@ -2555,6 +2555,15 @@ private[spark] object Utils extends Logging {
       sparkJars.map(_.split(",")).map(_.filter(_.nonEmpty)).toSeq.flatten
     }
   }
+
+  private[util] val REDACTION_REPLACEMENT_TEXT = "*********(redacted)"
+  def redact(conf: SparkConf)(kv: (String, String)): (String, String) = {
```
markgrover: Fixed the empty line. So, have the redact method take in a conf as a parameter and return a method?
I see them as equivalent, but don't feel strongly about it, so I will change it.

core/src/main/scala/org/apache/spark/util/Utils.scala
```diff
@@ -2555,6 +2555,15 @@ private[spark] object Utils extends Logging {
...
+    val redactionPattern = conf.get(SECRET_REDACTION_PATTERN).r
```
markgrover: What part do you think is expensive? Going through all the configuration properties and matching them with the regex?
If so, I agree. However, that has to be done somewhere. All the callers of this function have a `SparkConf` that they want stuff redacted from. So, if this function accepts a list of tuples, they have to run the regex check to find that list first before sending it to `redact()`. So, overall, unless I am missing something, I don't think we can avoid the expense.

core/src/test/scala/org/apache/spark/scheduler/EventLoggingListenerSuite.scala
markgrover: Fixed.

core/src/test/scala/org/apache/spark/util/UtilsSuite.scala
markgrover: Fixed.

core/src/test/scala/org/apache/spark/scheduler/EventLoggingListenerSuite.scala
vanzin: Both are not doing the exact same thing. There's logic in the `EventLoggingListener` code that also needs to be tested. But you can have a more targeted unit test instead of running a distributed Spark application.

core/src/main/scala/org/apache/spark/util/Utils.scala
vanzin: Compiling the regex once for every item in the list being redacted, instead of doing it once for the whole list.

core/src/main/scala/org/apache/spark/util/Utils.scala
vanzin: I mean the `redact` method should take 2 parameters, the config and the list of things to be redacted, instead of using currying for the second parameter.

core/src/main/scala/org/apache/spark/util/Utils.scala
markgrover: Ah, good point. Let me fix this.

markgrover:
Thanks for your feedback @vanzin and @ajbozarth.
I have updated the PR with your suggestions. The only one that's pending is @vanzin's suggestion of improving the EventLoggingSuiteTest. Let me think about that one for a bit before I update it.

core/src/main/scala/org/apache/spark/scheduler/EventLoggingListener.scala
vanzin: The downside is that with this choice you have more complicated tests, and you have to do redaction at every place where this information might be written (which at the moment is just two - the event logger and the UI).

JsonProtocol is not really a choice because it doesn't cover the UI.

> A user could write a custom listener that listened to the environment updates and did something useful with them. And, I didn't want to impose redaction on them.

That's the only argument I can buy, and said user has different ways to get that information.

core/src/main/scala/org/apache/spark/util/Utils.scala
```diff
@@ -2555,6 +2555,19 @@ private[spark] object Utils extends Logging {
...
+
+  private[util] val REDACTION_REPLACEMENT_TEXT = "*********(redacted)"
+
+  def redact(conf: SparkConf, kvs: Seq[(String, String)]): Seq[(String, String)] = {
+    val redactionPattern = conf.get(SECRET_REDACTION_PATTERN).r
+    kvs.map { kv =>
+      if (redactionPattern.findFirstIn(kv._1).isDefined) {
+        (kv._1, REDACTION_REPLACEMENT_TEXT)
+      }
+      else kv
```
vanzin: Style here is wrong. Better to use the `Option` api:

```
regex.findFirstIn(...).map(...).getOrElse(...)
```

core/src/test/scala/org/apache/spark/util/UtilsSuite.scala
```diff
@@ -974,4 +974,28 @@ class UtilsSuite extends SparkFunSuite with ResetSystemProperties with Logging {
 
     assert(pValue > threshold)
   }
+
+  test("redact sensitive information") {
+    val sparkConf = new SparkConf
+
+    // Set some secret keys
+    val secretKeys = Seq("" +
```
vanzin: What is `"" +` accomplishing here?

core/src/test/scala/org/apache/spark/util/UtilsSuite.scala
```diff
@@ -974,4 +974,28 @@ class UtilsSuite extends SparkFunSuite with ResetSystemProperties with Logging {
...
+    sparkConf.set("spark.regular.property", "not_a_secret")
+
+    // Redact sensitive information
+    val redactedConf = Utils.redact(sparkConf, sparkConf.getAll).toMap
+
+    // Assert that secret information got redacted while the regular property remained the same
+    secretKeys.foreach { key =>
+      assert(redactedConf.get(key).get == Utils.REDACTION_REPLACEMENT_TEXT)
+    }
+    assert(redactedConf.get("spark.regular.property").get == "not_a_secret")
```
vanzin: Same as above.

core/src/main/scala/org/apache/spark/scheduler/EventLoggingListener.scala
```diff
@@ -231,6 +233,19 @@ private[spark] class EventLoggingListener(
...
+    // "Spark Properties" entry will always exist because the map is always populated with it.
+    val props = event
+      .environmentDetails
+      .get("Spark Properties")
+      .get
```
vanzin: Instead of using `map.get(key).get`, just use `map(key)`.

core/src/main/scala/org/apache/spark/internal/config/package.scala
```diff
@@ -223,4 +223,13 @@ package object config {
       " bigger files.")
     .longConf
     .createWithDefault(4 * 1024 * 1024)
+
+  private[spark] val SECRET_REDACTION_PATTERN =
+    ConfigBuilder("spark.redaction.regex")
+      .doc("Regex to decide which Spark configuration properties and environment variables in " +
+        "driver and executor environments contain sensitive information. When this regex matches " +
+        "a property , its value is redacted from the environment UI and various logs like YARN " +
```
vanzin: nit: no space before comma.

core/src/main/scala/org/apache/spark/util/Utils.scala
markgrover: That's much better, thanks.

core/src/test/scala/org/apache/spark/util/UtilsSuite.scala
markgrover: Me trying to format things in my editor led to this, apologies. Fixed it.

core/src/test/scala/org/apache/spark/util/UtilsSuite.scala
```diff
@@ -974,4 +974,28 @@ class UtilsSuite extends SparkFunSuite with ResetSystemProperties with Logging {
...
+    }
+    // Set a non-secret key
+    sparkConf.set("spark.regular.property", "not_a_secret")
+
+    // Redact sensitive information
+    val redactedConf = Utils.redact(sparkConf, sparkConf.getAll).toMap
+
+    // Assert that secret information got redacted while the regular property remained the same
+    secretKeys.foreach { key =>
+      assert(redactedConf.get(key).get == Utils.REDACTION_REPLACEMENT_TEXT)
```
markgrover: Fixed.

core/src/test/scala/org/apache/spark/util/UtilsSuite.scala
markgrover: Fixed.

core/src/main/scala/org/apache/spark/scheduler/EventLoggingListener.scala
markgrover: Fixed.

core/src/main/scala/org/apache/spark/internal/config/package.scala
markgrover: Fixed.

core/src/main/scala/org/apache/spark/internal/config/package.scala
```diff
@@ -223,4 +223,13 @@ package object config {
...
+        "and event logs")
```
markgrover: Fixed.

core/src/main/scala/org/apache/spark/scheduler/EventLoggingListener.scala
```diff
@@ -231,6 +233,19 @@ private[spark] class EventLoggingListener(
     }
   }
 
+
```
markgrover: Fixed.

core/src/main/scala/org/apache/spark/scheduler/EventLoggingListener.scala
markgrover: >That's the only argument I can buy, and said user has different ways to get that information.

True, `sparkConf.getAll` is always available.

I still think it's better to put the redaction closer to the "sinks" for now. The good thing though is that if we see the number of "sinks" increasing, and everyone wanting redaction, we can take redaction more upstream. For now, 2 sinks seem manageable and it's hard to guess if future sinks are going to want redaction or not. So, unless you strongly object, I'd like to keep the redaction closer to the "sinks" now.

core/src/test/scala/org/apache/spark/scheduler/EventLoggingListenerSuite.scala
markgrover: Ok, I have updated this test to be less bloated. Thanks!

markgrover:
Thanks for your review, @vanzin I have incorporated all suggestions, I'd appreciate if you could take another look.

And, the test failures in the last run seem unrelated.

core/src/main/scala/org/apache/spark/scheduler/EventLoggingListener.scala
```diff
@@ -231,6 +233,17 @@ private[spark] class EventLoggingListener(
     }
   }
 
+  private[spark] def redactEvent(event: SparkListenerEnvironmentUpdate):
+  SparkListenerEnvironmentUpdate = {
```
vanzin: This is kind of an edge case, but this line needs indentation. I recommend:

```
private[spark] def redactEvent(
    event: SparkListenerEnvironmentUpdate): SparkListenerEnvironmentUpdate = {
```

core/src/main/scala/org/apache/spark/scheduler/EventLoggingListener.scala
```diff
@@ -231,6 +233,17 @@ private[spark] class EventLoggingListener(
...
+    // "Spark Properties" entry will always exist because the map is always populated with it.
+    val props = event
+      .environmentDetails("Spark Properties")
```
vanzin: nit: fits in previous line. You can even merge it with the next statement.

core/src/main/scala/org/apache/spark/util/Utils.scala
```diff
@@ -2555,6 +2555,18 @@ private[spark] object Utils extends Logging {
...
     }
   }
+
+  private[util] val REDACTION_REPLACEMENT_TEXT = "*********(redacted)"
+
+  def redact(conf: SparkConf, kvs: Seq[(String, String)]): Seq[(String, String)] = {
+    val redactionPattern = conf.get(SECRET_REDACTION_PATTERN).r
+    kvs.map { kv =>
+      redactionPattern.findFirstIn(kv._1)
+        .map{ ignore => (kv._1, REDACTION_REPLACEMENT_TEXT) }
```
vanzin: nit: space after `map`

core/src/main/scala/org/apache/spark/util/Utils.scala
```diff
@@ -2555,6 +2555,18 @@ private[spark] object Utils extends Logging {
...
+        .getOrElse(kv)
+      }
```
vanzin: nit: indented too far

core/src/test/scala/org/apache/spark/scheduler/EventLoggingListenerSuite.scala
```diff
@@ -95,6 +95,18 @@ class EventLoggingListenerSuite extends SparkFunSuite with LocalSparkContext wit
...
 
+  test("Event logging with password redaction") {
+    val key = "spark.executorEnv.HADOOP_CREDSTORE_PASSWORD"
+    val secretPassword = "secret_password"
+    val conf = getLoggingConf(testDirPath, None)
+      .set(key, secretPassword)
+    val eventLogger = new EventLoggingListener("test", None, testDirPath.toUri(), conf)
+    val envDetails = SparkEnv.environmentDetails(conf, "FIFO", Seq.empty, Seq.empty)
+    val event = SparkListenerEnvironmentUpdate(envDetails)
+    val redactedProps = eventLogger.redactEvent(event).environmentDetails("Spark Properties").toMap
```
vanzin: "Spark Properties" is begging to be turned into a constant somewhere...

core/src/test/scala/org/apache/spark/util/UtilsSuite.scala
```diff
@@ -974,4 +974,24 @@ class UtilsSuite extends SparkFunSuite with ResetSystemProperties with Logging {
...
+      "spark.my.sECreT")
+    secretKeys.foreach { key => sparkConf.set(key, "secret_password") }
+    // Set a non-secret key
+    sparkConf.set("spark.regular.property", "not_a_secret")
+
+    // Redact sensitive information
+    val redactedConf = Utils.redact(sparkConf, sparkConf.getAll).toMap
+
+    // Assert that secret information got redacted while the regular property remained the same
+    secretKeys.foreach { key => assert(redactedConf(key) == Utils.REDACTION_REPLACEMENT_TEXT) }
```
vanzin: `===`

core/src/test/scala/org/apache/spark/util/UtilsSuite.scala
```diff
@@ -974,4 +974,24 @@ class UtilsSuite extends SparkFunSuite with ResetSystemProperties with Logging {
...
+    assert(redactedConf("spark.regular.property") == "not_a_secret")
```
vanzin: `===`

vanzin (commented):
A few minor things to take care of.

core/src/main/scala/org/apache/spark/scheduler/EventLoggingListener.scala
```diff
@@ -231,6 +233,15 @@ private[spark] class EventLoggingListener(
     }
   }
 
+  private[spark] def redactEvent(
+    event: SparkListenerEnvironmentUpdate): SparkListenerEnvironmentUpdate = {
```
vanzin: nit: this needs to be indented one more level...

markgrover: Thanks, I'll go through Spark style guide so I don't cause as much trouble next time.

Thanks for reviewing.

vanzin:
LGTM pending tests.

markgrover:
Thanks for reviewing, Marcelo.

Hmm, the failures are still unrelated:-(
```
[info] Exception encountered when attempting to run a suite with class name: org.apache.spark.sql.jdbc.JDBCWriteSuite *** ABORTED *** (1 second, 791 milliseconds)
[info]   org.h2.jdbc.JdbcSQLException: Schema "TEST" already exists; SQL statement:
[info] create schema test [90078-183]
[info]   at org.h2.message.DbException.getJdbcSQLException(DbException.java:345)
[info]   at org.h2.message.DbException.get(DbException.java:179)
[info]   at org.h2.message.DbException.get(DbException.java:155)
[info]   at org.h2.command.ddl.CreateSchema.update(CreateSchema.java:48)
[info]   at org.h2.command.CommandContainer.update(CommandContainer.java:78)
[info]   at org.h2.command.Command.executeUpdate(Command.java:254)
[info]   at org.h2.jdbc.JdbcPreparedStatement.executeUpdateInternal(JdbcPreparedStatement.java:157)
[info]   at org.h2.jdbc.JdbcPreparedStatement.executeUpdate(JdbcPreparedStatement.java:143)
[info]   at org.apache.spark.sql.jdbc.JDBCWriteSuite$$anonfun$29.apply(JDBCWriteSuite.scala:56)
[info]   at org.apache.spark.sql.jdbc.JDBCWriteSuite$$anonfun$29.apply(JDBCWriteSuite.scala:53)
[info]   at org.scalatest.BeforeAndAfter$class.runTest(BeforeAndAfter.scala:195)
[info]   at org.apache.spark.sql.jdbc.JDBCWriteSuite.runTest(JDBCWriteSuite.scala:34)
[info]   at org.scalatest.FunSuiteLike$$anonfun$runTests$1.apply(FunSuiteLike.scala:208)
[info]   at org.scalatest.FunSuiteLike$$anonfun$runTests$1.apply(FunSuiteLike.scala:208)
[info]   at org.scalatest.SuperEngine$$anonfun$traverseSubNodes$1$1.apply(Engine.scala:413)
[info]   at org.scalatest.SuperEngine$$anonfun$traverseSubNodes$1$1.apply(Engine.scala:401)
[info]   at scala.collection.immutable.List.foreach(List.scala:381)
[info]   at org.scalatest.SuperEngine.traverseSubNodes$1(Engine.scala:401)
[info]   at org.scalatest.SuperEngine.org$scalatest$SuperEngine$$runTestsInBranch(Engine.scala:396)
[info]   at org.scalatest.SuperEngine.runTestsImpl(Engine.scala:483)
[info]   at org.scalatest.FunSuiteLike$class.runTests(FunSuiteLike.scala:208)
[info]   at org.scalatest.FunSuite.runTests(FunSuite.scala:1555)
[info]   at org.scalatest.Suite$class.run(Suite.scala:1424)
[info]   at org.scalatest.FunSuite.org$scalatest$FunSuiteLike$$super$run(FunSuite.scala:1555)
[info]   at org.scalatest.FunSuiteLike$$anonfun$run$1.apply(FunSuiteLike.scala:212)
[info]   at org.scalatest.FunSuiteLike$$anonfun$run$1.apply(FunSuiteLike.scala:212)
[info]   at org.scalatest.SuperEngine.runImpl(Engine.scala:545)
[info]   at org.scalatest.FunSuiteLike$class.run(FunSuiteLike.scala:212)
[info]   at org.apache.spark.SparkFunSuite.org$scalatest$BeforeAndAfterAll$$super$run(SparkFunSuite.scala:31)
[info]   at org.scalatest.BeforeAndAfterAll$class.liftedTree1$1(BeforeAndAfterAll.scala:257)
[info]   at org.scalatest.BeforeAndAfterAll$class.run(BeforeAndAfterAll.scala:256)
[info]   at org.apache.spark.sql.jdbc.JDBCWriteSuite.org$scalatest$BeforeAndAfter$$super$run(JDBCWriteSuite.scala:34)
[info]   at org.scalatest.BeforeAndAfter$class.run(BeforeAndAfter.scala:241)
[info]   at org.apache.spark.sql.jdbc.JDBCWriteSuite.run(JDBCWriteSuite.scala:34)
[info]   at org.scalatest.tools.Framework.org$scalatest$tools$Framework$$runSuite(Framework.scala:357)
[info]   at org.scalatest.tools.Framework$ScalaTestTask.execute(Framework.scala:502)
[info]   at sbt.ForkMain$Run$2.call(ForkMain.java:296)
[info]   at sbt.ForkMain$Run$2.call(ForkMain.java:286)
[info]   at java.util.concurrent.FutureTask.run(FutureTask.java:266)
[info]   at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
[info]   at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
[info]   at java.lang.Thread.run(Thread.java:745)
```
I won't be available for the next few days. If you deem fit to you, I'd appreciate if you could commit this after the test runs stabilize. Thanks!

markgrover:
Jenkins, retest this please.

markgrover:
Ok, looks like all is good now!

vanzin:
Merging to master.

PR #15973, opened 2016-11-22
[do-not-submit] Initial Kubernetes cluster manager implementation.

mccheah:
Includes the following initial feature set:
- Cluster mode with only Scala/Java jobs
- Spark-submit support
- Dynamic allocation

Does not include, most notably:
- Client mode support
- Proper testing on both the unit and integration level; integration
tests are flaky


mccheah:
Sorry, wrong base for the fork here

PR #15974, opened 2016-11-22
[SPARK-18537] [Web UI]Add a REST api to spark streaming 

ChorPangChan:

1. implement a package(org.apache.spark.streaming.status.api.v1) that serve the same purpose as org.apache.spark.status.api.v1
1. register the api path through StreamingPage
1. retrive the streaming informateion through StreamingJobProgressListener

this api should cover exceptly the same amount of information as you can get from the web interface
the implementation is base on the current REST implementation of spark-core
and will be available for running applications only

https://issues.apache.org/jira/browse/SPARK-18537


ChorPangChan:
hi @tdas
pre our conversation in spark-dev channel, I have done the implementation.
can you please have a look on this please?

ajbozarth:
I haven't had a chance to look through your code, but how does this related to #15904 ? From your descriptions it looks like you both opened a JIRA and a PR for the same task. Also is there a reason this is opened for merging into 1.6?

ChorPangChan:
I think what we trying to do is duplicated. I do believe create a new package under streaming is better then modify the spark-core package.
the reason for merging this to 1.6 is just because I'm using 1.6, my implementation is base on 1.6 and it require manual merge for master.

vanzin:
This will never be merged into 1.6, so you'll have to send a new PR against master. Please close this one.

ChorPangChan:
It require manual merge if I do it against master.
should I just PR with conflicts or rebase to master before PR?
btw those conflicts are just versions from the pom file.

uncleGen:
I think there is no need to open another reduplicate PR. Do your mind closing this PR, and let work on #15904 ?

uncleGen:
I think the base solutions are same, expect some other information which I am working to add.

ChorPangChan:
I stay adding a new package to streaming is a better structure then modify the spark-core.
can we make the decision for which implementation to use first

ChorPangChan:
hey @uncleGen 
I think we need to settle this down in order to make progress.

I had a quick look in your code, as i said above,
modifying spark-core is no a good idea to implement the streaming REST api
because user can use spark without streaming

at least as a spark user, I don't want something not related to spark-core inside it.

if you already come up a even better design, then it will be another story,
but for now, I think my implementation is the way to go

how do you think?

uncleGen:
OK

vanzin:
@ChorPangChan we will not merge new features into a maintenance branch. You need to submit a new PR against master. You can't change this one to be against master, you have to open a new one. Please close this.

ChorPangChan:
@uncleGen Thankyou
@vanzin OK

ShubhamGupta29:
From where can I get the data for executors as /api/v1 end point for /allexecutors is not updating , that means , after first fetch of data , the data remains same forever.Why is it so?

PR #15977, opened 2016-11-22
[SPARK-18436][SQL] isin causing SQL syntax error with JDBC

jiangxb1987:

The expression `in(empty seq)` is invalid in some data source. Since `in(empty seq)` is always false, we should generate `in(empty seq)` to false literal in optimizer.
The sql `SELECT * FROM t WHERE a IN ()` throws a `ParseException` which is consistent with Hive, don't need to change that behavior.

## How was this patch tested?
Add new test case in `OptimizeInSuite`.


jiangxb1987:
cc @hvanhovell @rxin @srowen @gatorsmile @windpiger

hvanhovell:
Could you also fix (the actual problem): https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JDBCRDD.scala#L123

jiangxb1987:
@hvanhovell Thanks for that catch! I'll fix that too.

jiangxb1987:
The failed unit test seems not related to our change here.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JDBCRDD.scala
```diff
@@ -120,7 +120,13 @@ object JDBCRDD extends Logging {
       case StringStartsWith(attr, value) => s"${attr} LIKE '${value}%'"
       case StringEndsWith(attr, value) => s"${attr} LIKE '%${value}'"
       case StringContains(attr, value) => s"${attr} LIKE '%${value}%'"
-      case In(attr, value) => s"$attr IN (${compileValue(value)})"
+      case In(attr, value) =>
+        if (value.nonEmpty) {
+          s"$attr IN (${compileValue(value)})"
+        } else {
+          // Return false literal when value is empty.
+          s"false"
```
srowen: Nit: remove interpolation

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JDBCRDD.scala
```diff
@@ -120,7 +120,13 @@ object JDBCRDD extends Logging {
       case StringStartsWith(attr, value) => s"${attr} LIKE '${value}%'"
       case StringEndsWith(attr, value) => s"${attr} LIKE '%${value}'"
       case StringContains(attr, value) => s"${attr} LIKE '%${value}%'"
-      case In(attr, value) => s"$attr IN (${compileValue(value)})"
+      case In(attr, value) =>
```
hvanhovell: NIT: It is shorter to put both options in different case statements:
```scala
...
case In(attr, values) if value.nonEmpty => s"$attr IN (${compileValue(value)})"
case In(_, _) => "false"
...
```

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/expressions.scala
```diff
@@ -114,10 +114,14 @@ object ReorderAssociativeOperator extends Rule[LogicalPlan] {
  * 1. Removes literal repetitions.
  * 2. Replaces [[In (value, seq[Literal])]] with optimized version
  *    [[InSet (value, HashSet[Literal])]] which is much faster.
+ * 3. Replaces [[In (value, Seq.empty)]] with false literal.
  */
 case class OptimizeIn(conf: CatalystConf) extends Rule[LogicalPlan] {
   def apply(plan: LogicalPlan): LogicalPlan = plan transform {
     case q: LogicalPlan => q transformExpressionsDown {
+      case expr @ In(v, list) if list.isEmpty =>
```
rxin: what's the current behavior for (null in ()) ? We want to make sure the implementation of In returns the same result (False here).

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JDBCRDD.scala
```diff
@@ -120,7 +120,13 @@ object JDBCRDD extends Logging {
...
+        if (value.nonEmpty) {
+          s"$attr IN (${compileValue(value)})"
+        } else {
+          // Return false literal when value is empty.
```
gatorsmile: This is dead code, right? 

If Optimizer already replace the empty value by a constant `false`, how to reach this branch?

rxin: We shouldn't rely on the optimizer for correctness.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/expressions.scala
jiangxb1987: How about we add a new case to handle null literal in value? Like the following:
```
      case expr @ In(v @ Literal(null, _), list) =>
        v
        
      case expr @ In(v, list) if list.isEmpty =>
        FalseLiteral
```

jiangxb1987:
retest this please.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JDBCRDD.scala
```diff
@@ -120,7 +120,9 @@ object JDBCRDD extends Logging {
       case StringStartsWith(attr, value) => s"${attr} LIKE '${value}%'"
       case StringEndsWith(attr, value) => s"${attr} LIKE '%${value}'"
       case StringContains(attr, value) => s"${attr} LIKE '%${value}%'"
-      case In(attr, value) => s"$attr IN (${compileValue(value)})"
+      case In(null, value) => "NULL"
```
srowen: Dumb question but does this replacement make sense? can NULL be a predicate? Something that might otherwise render as `SELECT * FROM foo WHERE NULL in ()` now becomes `SELECT * FROM foo WHERE NULL`?

rxin:
I believe the correct replacement for "x in ()" is "if (isnull(x)) null else false"

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JDBCRDD.scala
gatorsmile: Yeah. `NULL` in predicates means `UNKNOWN`. 

FYI, 
- our source code for the `IN` operator: https://github.com/apache/spark/blob/84284e8c82542d80dad94e458a0c0210bf803db3/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/predicates.scala#L141-L142

Oracle Docs of `NULL`: https://docs.oracle.com/cd/B19306_01/server.102/b14200/sql_elements005.htm

gatorsmile:
How about adding the following two lines into the test case in [`PredicateSuite.scala`](https://github.com/apache/spark/blob/84284e8c82542d80dad94e458a0c0210bf803db3/sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/PredicateSuite.scala#L122)? 
```Scala
    checkEvaluation(In(Literal.create(null, IntegerType), Seq.empty), null)
    checkEvaluation(In(Literal(1), Seq.empty), false)
```

rxin:
Also use NonFoldableLiteral to do the test. Make sure you create a null nonfoldable literal to verify the result is null.

jiangxb1987:
Thank you for comment@rxin@srowen@gatorsmile! I'll update that later today!

jiangxb1987:
retest this please.

jiangxb1987:
retest this please. - The failure seems not related to our change in this PR.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/expressions.scala
hvanhovell: You can remove this change. `IN()` is very efficient in these cases.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JDBCRDD.scala
```diff
@@ -120,6 +120,7 @@ object JDBCRDD extends Logging {
       case StringStartsWith(attr, value) => s"${attr} LIKE '${value}%'"
       case StringEndsWith(attr, value) => s"${attr} LIKE '%${value}'"
       case StringContains(attr, value) => s"${attr} LIKE '%${value}%'"
+      case In(attr, value) if value.isEmpty => s"IF(${attr} IS NULL, NULL, false)"
```
hvanhovell: `IF` is not a part of the SQL standard, use: `CASE WHEN $attr IS NULL THEN NULL ELSE FALSE END`

hvanhovell:

PR #15978, opened 2016-11-22
[SPARK-18507][SQL] HiveExternalCatalog.listPartitions should only call getTable once

cloud-fan:

HiveExternalCatalog.listPartitions should only call `getTable` once, instead of calling it for every partitions.

## How was this patch tested?

N/A


cloud-fan:
cc @yhuai @mallman @ericl

mallman:
I'll give this patch a try. Thanks.

mallman:
This patch resolves the performance problem I was seeing. Thank you!

ericl:
Lgtm. Should we also add a metric on the number if hive calls and check that in the table perf stats suite to prevent future regression?

andrewor14:
Oops, that was my fault. Thanks merging into master 2.1

andrewor14:
@cloud-fan can you make a patch for 2.0?

andrewor14:
(Oops never mind, not my fault! :p)

PR #13469, opened 2016-06-02
[SPARK-15140][SQL] make the semantics of null input object for encoder clear

cloud-fan:
## What changes were proposed in this pull request?

For input object of non-flat type, we can't encode it to row if it's null, as Spark SQL doesn't allow row to be null, only its columns can be null.

This PR explicitly add this constraint and throw exception if users break it.
## How was this patch tested?

several new tests

PR #15979, opened 2016-11-22
[SPARK-18251][SQL] the type of Dataset can't be Option of non-flat type

cloud-fan:

For input object of non-flat type, we can't encode it to row if it's null, as Spark SQL doesn't allow the entire row to be null, only its columns can be null. That's the reason we forbid users to use top level null objects in https://github.com/apache/spark/pull/13469

However, if users wrap non-flat type with `Option`, then we may still encoder top level null object to row, which is not allowed.

This PR fixes this case, and suggests users to wrap their type with `Tuple1` if they do wanna top level null objects.

## How was this patch tested?

new test

### DIFF
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/encoders/ExpressionEncoder.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/encoders/ExpressionEncoder.scala
@@ -52,7 +52,8 @@ object ExpressionEncoder {
       throw new UnsupportedOperationException(
         "Cannot create encoder for Option of non-flat type, as non-flat type is represented " +
           "as a row, and the entire row can not be null in Spark SQL like normal databases. " +
-          "You can wrap your type with Tuple1 if you do want top level null objects.")
+          "You can wrap your type with Tuple1 if you do want top level null objects, e.g. " +
+          "val ds: Dataset[Tuple1[MyClass]] = Seq(Tuple1(MyClass(...)), Tuple1(null)).toDS")
     }
 
     val cls = mirror.runtimeClass(tpe)
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/encoders/ExpressionEncoder.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/encoders/ExpressionEncoder.scala
@@ -47,6 +47,14 @@ object ExpressionEncoder {
     // We convert the not-serializable TypeTag into StructType and ClassTag.
     val mirror = typeTag[T].mirror
     val tpe = typeTag[T].tpe
+
+    if (ScalaReflection.optionOfNonFlatType(tpe)) {
+      throw new UnsupportedOperationException(
+        "Cannot create encoder for Option of non-flat type, as non-flat type is represented " +
+          "as a row, and the entire row can not be null in Spark SQL like normal databases. " +
+          "You can wrap your type with Tuple1 if you do want top level null objects.")
+    }
+
     val cls = mirror.runtimeClass(tpe)
     val flat = !ScalaReflection.definedByConstructorParams(tpe)
 
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/encoders/ExpressionEncoder.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/encoders/ExpressionEncoder.scala
@@ -48,12 +48,13 @@ object ExpressionEncoder {
     val mirror = typeTag[T].mirror
     val tpe = typeTag[T].tpe
 
-    if (ScalaReflection.optionOfNonFlatType(tpe)) {
+    if (ScalaReflection.optionOfProductType(tpe)) {
       throw new UnsupportedOperationException(
-        "Cannot create encoder for Option of non-flat type, as non-flat type is represented " +
+        "Cannot create encoder for Option of Product type, because Product type is represented " +
           "as a row, and the entire row can not be null in Spark SQL like normal databases. " +
-          "You can wrap your type with Tuple1 if you do want top level null objects, e.g. " +
-          "val ds: Dataset[Tuple1[MyClass]] = Seq(Tuple1(MyClass(...)), Tuple1(null)).toDS")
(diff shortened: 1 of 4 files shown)


cloud-fan:
cc @yhuai @liancheng

rxin:
What does "non-flat type" mean?

cloud-fan:
"non-flat type" means "complex type", i.e. array, seq, map, product, etc.

cloud-fan:
retest it please

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/encoders/ExpressionEncoder.scala
```diff
@@ -47,6 +47,14 @@ object ExpressionEncoder {
     // We convert the not-serializable TypeTag into StructType and ClassTag.
     val mirror = typeTag[T].mirror
     val tpe = typeTag[T].tpe
+
+    if (ScalaReflection.optionOfNonFlatType(tpe)) {
+      throw new UnsupportedOperationException(
+        "Cannot create encoder for Option of non-flat type, as non-flat type is represented " +
+          "as a row, and the entire row can not be null in Spark SQL like normal databases. " +
+          "You can wrap your type with Tuple1 if you do want top level null objects.")
```
yhuai: Let's provide an example in the error message to help users understand how to handle this case.

yhuai:
looks good. @liancheng want to double check?

cloud-fan:
retest this please

liancheng:
LGTM, merging to master. Thanks!

liancheng:
My only concern is that "non-flat" type is neither intuitive nor a well-known term. In fact, this PR only prevents `Option[T <: Product]` to be top-level Dataset types. How about just call them "Product" types?

Otherwise LGTM.

rxin:
FWIW I don't think we should call it nonflat.

cloud-fan:
retest this please

liancheng:
Good to merge pending Jenkins. Thanks!

cloud-fan:
retest this please

liancheng:
Merging to master. Thanks!

liancheng:
@rxin Shall we backport this to branch-2.1? I think it's relatively safe.

rxin:
Sounds good.

liancheng:
Also backported to branch-2.1.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/encoders/ExpressionEncoder.scala
```diff
@@ -47,16 +47,26 @@ object ExpressionEncoder {
     // We convert the not-serializable TypeTag into StructType and ClassTag.
     val mirror = typeTag[T].mirror
     val tpe = typeTag[T].tpe
+
+    if (ScalaReflection.optionOfProductType(tpe)) {
+      throw new UnsupportedOperationException(
+        "Cannot create encoder for Option of Product type, because Product type is represented " +
```
koertkuipers: this also means an Aggregator cannot use an Option of Product Type for its intermediate type. e.g.
Aggregator[Int, Option[(Int, Int)], Int] is now invalid. but i see no good reason why such an Aggregator wouldnt exist?

koertkuipers: this strikes me more as a limitation on Dataset[X] then on Encoder[X]

koertkuipers: and now that i think about it more, i also think Dataset[Option[(Int, Int)]] should be valid too if possible. 

it should not be represented by a top level Row object, so the schema should be
```StructType(StructField("_1", StructType(StructField("_1", IntegerType, false), StructField("_2", IntegerType, false)), true))```

basically it should be treated as flat. we do this trick for flat types all the time (nest them inside a row with one element), so why not for an Option[X <: Product]?

**Comment by koertkuipers** [comment]
this means anything that uses an encoder can no longer use Option[_ <: Product].
encoders are not just used for the top level Dataset creation.

Dataset.groupByKey[K] requires an encoder for K.
KeyValueGroupedDataset.mapValues[W] requires an encoder for V
Aggregator[A, B, C] requires encoders for B and C

none of these always create top level row objects (for which this pullreq creates the restriction that they cannot be null).

for an aggregator it is sometimes the case. ```dataset.select(aggregator)``` does create a top level row object, but ```dataset.groupByKey(...).agg(aggregator)``` does not.

so i am not sure it makes sense to put this restriction on the encoder. it seems to belong on the dataset.

another example of something that won't work anymore:
```
val x: Dataset[String, Option[(String, String)]] = ...
x.groupByKey(_._1).mapValues(_._2).agg(someAgg)
```
in this case the mapValues requires``` Encoder[Option[(String, String)]]```

**Comment by cloud-fan** [comment]
```
val x: Dataset[String, Option[(String, String)]] = ...
x.groupByKey(_._1).mapValues(_._2).agg(someAgg)
```
Does it work before?

Please see the discussion in the JIRA: https://issues.apache.org/jira/browse/SPARK-18251
Ideally we have a map between type T and catalyst schema, and `Option[T]` maps to the same catalyst schema with `T`, with additional null handling. We shouldn't change this mapping, which means we can't use a single field struct type to represent `Option[T]`.

It's still possible to support `Option[T]` completely(without breaking backward compatibility), but that may need a lof of hacky code and special handling, I don't think it worth, as we can easy work around it, by `Tuple1`.

koertkuipers:
Yes it worked before

koertkuipers:
spark 2.0.x does not have mapValues. but this works:

scala> Seq(("a", Some((1, 1))), ("a",
None)).toDS.groupByKey(_._2).count.show
+-----------+--------+
|        key|count(1)|
+-----------+--------+
|[null,null]|       1|
|      [1,1]|       1|
+-----------+--------+

koertkuipers:
admittedly the result looks weird. it really should be:

+-----------+--------+
|        key|count(1)|
+-----------+--------+
|       null|       1|
|      [1,1]|       1|
+-----------+--------+

is that a separate bug or related? i remember running into this before,
because serializing and then deserializing None comes back out as
Some((null, null)), which causes NPE in codegen. i ran into this with
Aggregator buffers.

PR #15980, opened 2016-11-22
[SPARK-18528][SQL] Fix a bug to initialise an iterator of aggregation buffer 

maropu:
This pr is to fix an `NullPointerException` issue caused by a following `limit + aggregate` query;
```
scala> val df = Seq(("a", 1), ("b", 2), ("c", 1), ("d", 5)).toDF("id", "value")
scala> df.limit(2).groupBy("id").count().show
WARN TaskSetManager: Lost task 0.0 in stage 9.0 (TID 8204, lvsp20hdn012.stubprod.com): java.lang.NullPointerException
at org.apache.spark.sql.catalyst.expressions.GeneratedClass$GeneratedIterator.agg_doAggregateWithKeys$(Unknown Source)
```
The root culprit is that [`$doAgg()`](https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/aggregate/HashAggregateExec.scala#L596) skips an initialization of [the buffer iterator](https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/aggregate/HashAggregateExec.scala#L603); `BaseLimitExec` sets `stopEarly=true` and `$doAgg()` exits in the middle without the initialization.

## How was this patch tested?
Added a test to check if no exception happens for limit + aggregates in `DataFrameAggregateSuite.scala`.

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/limit.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/limit.scala
@@ -67,6 +67,15 @@ trait BaseLimitExec extends UnaryExecNode with CodegenSupport {
   }
 
   override def doConsume(ctx: CodegenContext, input: Seq[ExprCode], row: ExprCode): String = {
+    val stopEarly = ctx.freshName("stopEarly")
+    ctx.addMutableState("boolean", stopEarly, s"$stopEarly = false;")
+
+    ctx.addNewFunction("stopEarly", s"""
+      @Override
+      protected boolean stopEarly() {
+        return $stopEarly;
+      }
+    """)
     val countTerm = ctx.freshName("count")
     ctx.addMutableState("int", countTerm, s"$countTerm = 0;")
     s"""
@@ -74,7 +83,7 @@ trait BaseLimitExec extends UnaryExecNode with CodegenSupport {
        |   $countTerm += 1;
        |   ${consume(ctx, input)}
        | } else {
-       |   // Do nothing
+       |   $stopEarly = true;
        | }
      """.stripMargin
   }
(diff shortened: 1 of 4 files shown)


hvanhovell:
cc @davies

sql/core/src/main/scala/org/apache/spark/sql/execution/limit.scala
```diff
@@ -67,23 +67,14 @@ trait BaseLimitExec extends UnaryExecNode with CodegenSupport {
...
-    """)
     val countTerm = ctx.freshName("count")
     ctx.addMutableState("int", countTerm, s"$countTerm = 0;")
     s"""
        | if ($countTerm < $limit) {
        |   $countTerm += 1;
        |   ${consume(ctx, input)}
        | } else {
-       |   $stopEarly = true;
+       |   // Do nothing
```
maropu: yea, I thinks so... I'm currently looking for other approaches.

maropu:
@davies How about this fix? I think it seems to be difficult to put the `earlyStop` flag in `shouldStop()` because `shouldStop()` is used for a caller (an uppper `SparkPlan`) to exit a loop for fetchting output row-by-row in `processNext()`. If a LIMIT plan sets true at `earlyStop`, `shouldStop` always returns true and
pre-aggregation wrongly breaks in the middle of `$doAgg`.

davies:
lgtm

hvanhovell:
retest this please

hvanhovell:
I'll merge this after another test run. Ping me if you have objections.

hvanhovell:

PR #15981, opened 2016-11-22
[SPARK-18547][core] Propagate I/O encryption key when executors register.

vanzin:
This change modifies the method used to propagate encryption keys used during
shuffle. Instead of relying on YARN's UserGroupInformation credential propagation,
this change explicitly distributes the key using the messages exchanged between
driver and executor during registration. When RPC encryption is enabled, this means
key propagation is also secure.

This allows shuffle encryption to work in non-YARN mode, which means that it's
easier to write unit tests for areas of the code that are affected by the feature.

The key is stored in the SecurityManager; because there are many instances of
that class used in the code, the key is only guaranteed to exist in the instance
managed by the SparkEnv. This path was chosen to avoid storing the key in the
SparkConf, which would risk having the key being written to disk as part of the
configuration (as, for example, is done when starting YARN applications).

Tested by new and existing unit tests (which were moved from the YARN module to
core), and by running apps with shuffle encryption enabled.


core/src/main/scala/org/apache/spark/SparkEnv.scala
```diff
@@ -213,18 +220,20 @@ object SparkEnv extends Logging {
       bindAddress: String,
       advertiseAddress: String,
       port: Int,
-      isDriver: Boolean,
```
vanzin: I had to modify this to avoid going over the 10 argument limit the style checker imposes. Which is also a reason why the key is not stored in the SparkEnv itself.

vanzin:
Ping.

core/src/main/scala/org/apache/spark/serializer/SerializerManager.scala
```diff
@@ -33,7 +33,12 @@ import org.apache.spark.util.io.{ChunkedByteBuffer, ChunkedByteBufferOutputStrea
  * Component which configures serialization, compression and encryption for various Spark
  * components, including automatic selection of which [[Serializer]] to use for shuffles.
  */
-private[spark] class SerializerManager(defaultSerializer: Serializer, conf: SparkConf) {
+private[spark] class SerializerManager(
+    defaultSerializer: Serializer,
+    conf: SparkConf,
+    encryptionKey: Option[Array[Byte]]) {
+
+  def this(defaultSerializer: Serializer, conf: SparkConf) = this(defaultSerializer, conf, None)
```
zsxwing: nit: why not use the parameter default value?

mesos/src/main/scala/org/apache/spark/executor/MesosExecutorBackend.scala
```diff
@@ -75,7 +75,7 @@ private[spark] class MesosExecutorBackend
     val conf = new SparkConf(loadDefaults = true).setAll(properties)
     val port = conf.getInt("spark.executor.port", 0)
     val env = SparkEnv.createExecutorEnv(
-      conf, executorId, slaveInfo.getHostname, port, cpusPerTask, isLocal = false)
+      conf, executorId, slaveInfo.getHostname, port, cpusPerTask, None, isLocal = false)
```
zsxwing: This means mesos is not suppported. Right? Could you throw an exception if IO encryption is enabled for mesos?

zsxwing (changes requested):
Since it sends the key over the wire, I think it's better to throw an exception if `securityManager.isSaslEncryptionEnabled()` is false.

vanzin:
> Since it sends the key over the wire, I think it's better to throw an exception if securityManager.isSaslEncryptionEnabled() is false.

We don't do that now, and there's a whole lot of things that need to configured correctly for things to be really secure (like HDFS and YARN encryption). It would also make unit tests slightly more complicated (or at least their set up). I think it's fine to just have a warning here.

core/src/main/scala/org/apache/spark/serializer/SerializerManager.scala
vanzin: Actually, I'm calling this from `UnsafeShuffleWriterSuite.java` in the accompanying change, and default values make it hard to call the code from Java.

zsxwing:
> I think it's fine to just have a warning here.

Sounds good. Just make sure you mention it when updating the doc.

core/src/main/scala/org/apache/spark/serializer/SerializerManager.scala
zsxwing: I see. Thanks for your clarifying.

zsxwing:
> I think it's fine to just have a warning here.

I cannot find the warning log in your latest commit. Otherwise, LGTM.

vanzin:
Looks like I removed the warning from SparkEnv at some point, let me add it back.

zsxwing:
LGTM. The build for the latest commit passed. Merging to master and 2.1.

PR #15982, opened 2016-11-22
[SPARK-18546][core] Fix merging shuffle spills when using encryption.

vanzin:
The problem exists because it's not possible to just concatenate encrypted
partition data from different spill files; currently each partition would
have its own initial vector to set up encryption, and the final merged file
should contain a single initial vector for each merged partiton, otherwise
iterating over each record becomes really hard.

To fix that, UnsafeShuffleWriter now decrypts the partitions when merging,
so that the merged file contains a single initial vector at the start of
the partition data.

Because it's not possible to do that using the fast transferTo path, when
encryption is enabled UnsafeShuffleWriter will revert back to using file
streams when merging. It may be possible to use a hybrid approach when
using encryption, using an intermediate direct buffer when reading from
files and encrypting the data, but that's better left for a separate patch.

As part of the change I made DiskBlockObjectWriter take a SerializerManager
instead of a "wrap stream" closure, since that makes it easier to test the
code without having to mock SerializerManager functionality.

Tested with newly added unit tests (UnsafeShuffleWriterSuite for the write
side and ExternalAppendOnlyMapSuite for integration), and by running some
apps that failed without the fix.

### DIFF
--- a/core/src/main/java/org/apache/spark/shuffle/sort/UnsafeShuffleWriter.java
+++ b/core/src/main/java/org/apache/spark/shuffle/sort/UnsafeShuffleWriter.java
@@ -40,6 +40,8 @@ import org.apache.spark.annotation.Private;
 import org.apache.spark.executor.ShuffleWriteMetrics;
 import org.apache.spark.io.CompressionCodec;
 import org.apache.spark.io.CompressionCodec$;
+import org.apache.commons.io.output.CloseShieldOutputStream;
+import org.apache.commons.io.output.CountingOutputStream;
 import org.apache.spark.memory.TaskMemoryManager;
 import org.apache.spark.network.util.LimitedInputStream;
 import org.apache.spark.scheduler.MapStatus;
@@ -264,6 +266,7 @@ public class UnsafeShuffleWriter<K, V> extends ShuffleWriter<K, V> {
       sparkConf.getBoolean("spark.shuffle.unsafe.fastMergeEnabled", true);
     final boolean fastMergeIsSupported = !compressionEnabled ||
       CompressionCodec$.MODULE$.supportsConcatenationOfSerializedStreams(compressionCodec);
+    final boolean encryptionEnabled = blockManager.serializerManager().encryptionKey().isDefined();
     try {
       if (spills.length == 0) {
         new FileOutputStream(outputFile).close(); // Create an empty file
@@ -289,7 +292,7 @@ public class UnsafeShuffleWriter<K, V> extends ShuffleWriter<K, V> {
           // Compression is disabled or we are using an IO compression codec that supports
           // decompression of concatenated compressed streams, so we can perform a fast spill merge
           // that doesn't need to interpret the spilled bytes.
-          if (transferToEnabled) {
+          if (transferToEnabled && !encryptionEnabled) {
             logger.debug("Using transferTo-based fast merge");
             partitionLengths = mergeSpillsWithTransferTo(spills, outputFile);
           } else {
@@ -337,7 +340,8 @@ public class UnsafeShuffleWriter<K, V> extends ShuffleWriter<K, V> {
     final int numPartitions = partitioner.numPartitions();
     final long[] partitionLengths = new long[numPartitions];
     final InputStream[] spillInputStreams = new FileInputStream[spills.length];
-    OutputStream mergedFileOutputStream = null;
+    final CountingOutputStream mergedFileOutputStream = new CountingOutputStream(
+      new FileOutputStream(outputFile));
 
     boolean threwException = true;
     try {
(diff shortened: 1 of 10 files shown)


vanzin:
This also contains the changes in #15981 since it's built on top of those, but they should be merged separately. The fix for this bug is in the second commit.

vanzin:
retest this please

vanzin:
Unrelated failure. retest this please

vanzin:
Ping.

core/src/main/java/org/apache/spark/shuffle/sort/UnsafeShuffleWriter.java
```diff
@@ -337,42 +340,38 @@ void forceSorterToSpill() throws IOException {
     final int numPartitions = partitioner.numPartitions();
     final long[] partitionLengths = new long[numPartitions];
     final InputStream[] spillInputStreams = new FileInputStream[spills.length];
-    OutputStream mergedFileOutputStream = null;
+    final CountingOutputStream mergedFileOutputStream = new CountingOutputStream(
```
zsxwing: Could you add a comment about why need to use `CountingOutputStream` + `CloseShieldOutputStream`? It took me a while to figure out the optimization you did.

core/src/main/scala/org/apache/spark/serializer/SerializerManager.scala
```diff
@@ -144,14 +144,14 @@ private[spark] class SerializerManager(
   /**
    * Wrap an output stream for compression if block compression is enabled for its block type
    */
-  private[this] def wrapForCompression(blockId: BlockId, s: OutputStream): OutputStream = {
+  def wrapForCompression(blockId: BlockId, s: OutputStream): OutputStream = {
```
zsxwing: nit: not need to open this method

core/src/main/scala/org/apache/spark/serializer/SerializerManager.scala
```diff
@@ -144,14 +144,14 @@ private[spark] class SerializerManager(
...
     if (shouldCompress(blockId)) compressionCodec.compressedOutputStream(s) else s
   }
 
   /**
    * Wrap an input stream for compression if block compression is enabled for its block type
    */
-  private[this] def wrapForCompression(blockId: BlockId, s: InputStream): InputStream = {
+  def wrapForCompression(blockId: BlockId, s: InputStream): InputStream = {
```
zsxwing: nit: not need to open this method

core/src/main/java/org/apache/spark/shuffle/sort/UnsafeShuffleWriter.java
```diff
@@ -337,42 +340,38 @@ void forceSorterToSpill() throws IOException {
...
-              if (compressionCodec != null) {
-                partitionInputStream = compressionCodec.compressedInputStream(partitionInputStream);
-              }
-              ByteStreams.copy(partitionInputStream, mergedFileOutputStream);
-              innerThrewException = false;
-            } finally {
-              Closeables.close(partitionInputStream, innerThrewException);
+            InputStream partitionInputStream = new LimitedInputStream(spillInputStreams[i],
+              partitionLengthInSpill, false);
+            partitionInputStream = blockManager.serializerManager().wrapForEncryption(
```
zsxwing: `partitionInputStream` is not closed

core/src/test/scala/org/apache/spark/util/collection/ExternalAppendOnlyMapSuite.scala
```diff
@@ -17,9 +17,13 @@
 
 package org.apache.spark.util.collection
 
+import java.security.PrivilegedExceptionAction
```
zsxwing: nit: unused import

core/src/test/scala/org/apache/spark/util/collection/ExternalAppendOnlyMapSuite.scala
```diff
@@ -17,9 +17,13 @@
...
+
 import scala.collection.mutable.ArrayBuffer
 
 import org.apache.spark._
+import org.apache.spark.deploy.SparkHadoopUtil
```
zsxwing: nit: unused import

core/src/test/java/org/apache/spark/shuffle/sort/UnsafeShuffleWriterSuite.java
```diff
@@ -338,42 +354,60 @@ private void testMergingSpills(
...
+    testMergingSpills(true, LZ4CompressionCodec.class.getName(), true);
+  }
+
+  @Test
+  public void mergeSpillsWithFileStreamAndCompressionAndEncryption() throws Exception {
+    testMergingSpills(false, LZ4CompressionCodec.class.getName(), true);
+  }
+
+  @Test
+  public void mergeSpillsWithCompressionAndEncryptionSlowPath() throws Exception {
```
zsxwing: We should also test `testMergingSpills(false, null, true);` and `testMergingSpills(true, null, true)`.

zsxwing (changes requested):
Overall looks good. Just left some minor comments

core/src/test/java/org/apache/spark/shuffle/sort/UnsafeShuffleWriterSuite.java
```diff
@@ -40,9 +41,11 @@
 import org.mockito.stubbing.Answer;
 
 import org.apache.spark.HashPartitioner;
+import org.apache.spark.SecurityManager;
 import org.apache.spark.ShuffleDependency;
 import org.apache.spark.SparkConf;
 import org.apache.spark.TaskContext;
+import org.apache.spark.deploy.SparkHadoopUtil;
```
squito: other than `CryptoStreamUtils`, the other added imports look unused.  Also looks like you can eliminate `AbstractFunction1` and `ByteStreams` since you are no longer using them.

core/src/test/java/org/apache/spark/unsafe/map/AbstractBytesToBytesMapSuite.java
```diff
@@ -75,13 +75,6 @@
   @Mock(answer = RETURNS_SMART_NULLS) BlockManager blockManager;
   @Mock(answer = RETURNS_SMART_NULLS) DiskBlockManager diskBlockManager;
 
-  private static final class WrapStream extends AbstractFunction1<OutputStream, OutputStream> {
```
squito: you can eliminate the imports of `AbstractFunction1` and `OutputStream` after this

core/src/test/java/org/apache/spark/util/collection/unsafe/sort/UnsafeExternalSorterSuite.java
```diff
@@ -86,14 +88,7 @@ public int compare(
 
   protected boolean shouldUseRadixSort() { return false; }
 
-  private final long pageSizeBytes = new SparkConf().getSizeAsBytes("spark.buffer.pageSize", "4m");
-
-  private static final class WrapStream extends AbstractFunction1<OutputStream, OutputStream> {
```
squito: same on trimming imports

core/src/main/java/org/apache/spark/shuffle/sort/UnsafeShuffleWriter.java
```diff
@@ -337,42 +340,47 @@ void forceSorterToSpill() throws IOException {
...
+        // Shield the underlying output stream from close() calls, so that we can close the higher
+        // level streams to make sure all data is really flushed and internal state is cleaned.
+        OutputStream partitionOutput = new CloseShieldOutputStream(mergedFileOutputStream);
+        partitionOutput = blockManager.serializerManager().wrapForEncryption(partitionOutput);
         if (compressionCodec != null) {
-          mergedFileOutputStream = compressionCodec.compressedOutputStream(mergedFileOutputStream);
+          partitionOutput = compressionCodec.compressedOutputStream(partitionOutput);
         }
-
+        partitionOutput = new TimeTrackingOutputStream(writeMetrics, partitionOutput);
```
vanzin: Hmm... let me revert this and open a bug. `DiskBlockObjectWriter` doesn't count the time for compression / encryption, so this should behave the same. Both should be fixed together.

vanzin:
retest this please

squito:
I think the comment on `#mergeSpillsWithFileStreams` needs to be updated slightly to include encryption, but other than that lgtm.

zsxwing:
LGTM

vanzin:
Merging to master / 2.1.

PR #15983, opened 2016-11-23
[SPARK-18544] [SQL] Append with df.saveAsTable writes data to wrong location

ericl:

We failed to properly propagate table metadata for existing tables for the saveAsTable command. This caused a downstream component to think the table was MANAGED, writing data to the wrong location.

## How was this patch tested?

Unit test that fails before the patch.

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/DataFrameWriter.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/DataFrameWriter.scala
@@ -372,6 +372,11 @@ final class DataFrameWriter[T] private[sql](ds: Dataset[T]) {
       case (true, SaveMode.ErrorIfExists) =>
         throw new AnalysisException(s"Table $tableIdent already exists.")
 
+      case (true, _) =>
+        val tableDesc = df.sparkSession.sessionState.catalog.getTableMetadata(tableIdent)
+        df.sparkSession.sessionState.executePlan(
+          CreateTable(tableDesc, mode, Some(df.logicalPlan))).toRdd
+
       case _ =>
         val storage = DataSource.buildStorageFormatFromOptions(extraOptions.toMap)
         val tableType = if (storage.locationUri.isDefined) {
@@ -391,12 +396,6 @@ final class DataFrameWriter[T] private[sql](ds: Dataset[T]) {
         )
         df.sparkSession.sessionState.executePlan(
           CreateTable(tableDesc, mode, Some(df.logicalPlan))).toRdd
-        if (tableDesc.partitionColumnNames.nonEmpty &&
-            df.sparkSession.sqlContext.conf.manageFilesourcePartitions) {
-          // Need to recover partitions into the metastore so our saved data is visible.
-          df.sparkSession.sessionState.executePlan(
-            AlterTableRecoverPartitionsCommand(tableDesc.identifier)).toRdd
-        }
     }
   }
 
--- a/sql/core/src/main/scala/org/apache/spark/sql/DataFrameWriter.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/DataFrameWriter.scala
@@ -372,14 +372,20 @@ final class DataFrameWriter[T] private[sql](ds: Dataset[T]) {
       case (true, SaveMode.ErrorIfExists) =>
         throw new AnalysisException(s"Table $tableIdent already exists.")
 
-      case (true, _) =>
-        val tableDesc = df.sparkSession.sessionState.catalog.getTableMetadata(tableIdent)
-        df.sparkSession.sessionState.executePlan(
-          CreateTable(tableDesc, mode, Some(df.logicalPlan))).toRdd
-
       case _ =>
-        val storage = DataSource.buildStorageFormatFromOptions(extraOptions.toMap)
-        val tableType = if (storage.locationUri.isDefined) {
+        val existingTable = if (tableExists) {
+          Some(df.sparkSession.sessionState.catalog.getTableMetadata(tableIdent))
+        } else {
+          None
+        }
+        val storage = if (tableExists) {
+          existingTable.get.storage
+        } else {
+          DataSource.buildStorageFormatFromOptions(extraOptions.toMap)
+        }
+        val tableType = if (tableExists) {
+          existingTable.get.tableType
+        } else if (storage.locationUri.isDefined) {
           CatalogTableType.EXTERNAL
         } else {
           CatalogTableType.MANAGED
--- a/sql/hive/src/test/scala/org/apache/spark/sql/hive/PartitionProviderCompatibilitySuite.scala
+++ b/sql/hive/src/test/scala/org/apache/spark/sql/hive/PartitionProviderCompatibilitySuite.scala
@@ -188,6 +188,25 @@ class PartitionProviderCompatibilitySuite
     }
   }
 
+  for (enabled <- Seq(true, false)) {
+    test(s"SPARK-18544 append with saveAsTable - partition management $enabled") {
+      withSQLConf(SQLConf.HIVE_MANAGE_FILESOURCE_PARTITIONS.key -> enabled.toString) {
+        withTable("test") {
+          withTempDir { dir =>
+            setupPartitionedDatasourceTable("test", dir)
+            if (enabled) {
+              spark.sql("msck repair table test")
+            }
+            assert(spark.sql("select * from test").count() == 5)
+            spark.range(10).selectExpr("id as fieldOne", "id as partCol")
+              .write.partitionBy("partCol").mode("append").saveAsTable("test")
+            assert(spark.sql("select * from test").count() == 15)
+          }
+        }
+      }
+    }
+  }
+
   /**
    * Runs a test against a multi-level partitioned table, then validates that the custom locations
    * were respected by the output writer.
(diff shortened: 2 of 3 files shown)


sql/core/src/main/scala/org/apache/spark/sql/DataFrameWriter.scala
```diff
@@ -391,12 +396,6 @@ final class DataFrameWriter[T] private[sql](ds: Dataset[T]) {
         )
         df.sparkSession.sessionState.executePlan(
           CreateTable(tableDesc, mode, Some(df.logicalPlan))).toRdd
-        if (tableDesc.partitionColumnNames.nonEmpty &&
```
ericl: I think this is unnecessary since the create / insert commands already recover partitions themselves. We'll see if any tests fail.

ericl:
cc @yhuai @cloud-fan

sql/core/src/main/scala/org/apache/spark/sql/DataFrameWriter.scala
```diff
@@ -373,8 +373,19 @@ final class DataFrameWriter[T] private[sql](ds: Dataset[T]) {
         throw new AnalysisException(s"Table $tableIdent already exists.")
 
       case _ =>
-        val storage = DataSource.buildStorageFormatFromOptions(extraOptions.toMap)
-        val tableType = if (storage.locationUri.isDefined) {
+        val existingTable = if (tableExists) {
```
cloud-fan: shall we move this logic in `CreateDataSourceTableAsSelectCommand`?

cloud-fan:
retest this please

sql/core/src/main/scala/org/apache/spark/sql/DataFrameWriter.scala
cloud-fan: nvm, I think it's fine here

cloud-fan:
retest this please

rxin:
Merging in master/branch-2.1.

sql/core/src/main/scala/org/apache/spark/sql/DataFrameWriter.scala
gatorsmile: This is still needed.

sql/hive/src/test/scala/org/apache/spark/sql/hive/PartitionProviderCompatibilitySuite.scala
```diff
@@ -188,6 +188,25 @@ class PartitionProviderCompatibilitySuite
...
   }
 
+  for (enabled <- Seq(true, false)) {
+    test(s"SPARK-18544 append with saveAsTable - partition management $enabled") {
+      withSQLConf(SQLConf.HIVE_MANAGE_FILESOURCE_PARTITIONS.key -> enabled.toString) {
+        withTable("test") {
+          withTempDir { dir =>
+            setupPartitionedDatasourceTable("test", dir)
+            if (enabled) {
+              spark.sql("msck repair table test")
```
gatorsmile: If `HIVE_MANAGE_FILESOURCE_PARTITIONS` is on, we have to repair the table; otherwise the table is empty to the external users. This looks weird to me. It is also inconsistent with the behavior when `HIVE_MANAGE_FILESOURCE_PARTITIONS` is off. I think we should repair the table after we create the table. Let me submit a PR and cc you then.

ericl:
It's the hive behavior to not repair the table. Otherwise, create table can
have an unbounded cost if there are many partitions.

gatorsmile:
Yeah, table repair is expensive, but this causes an external behavior change. I tried it in 2.0. It can show the whole data source table without repairing the table. In 2.1, it returns empty unless we repair the table. 

```Scala
scala> spark.range(5).selectExpr("id as fieldOne", "id as partCol").write.partitionBy("partCol").mode("overwrite").saveAsTable("test")
[Stage 0:======================>                                    (3 + 5) / 8]SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
16/12/17 17:41:20 WARN CreateDataSourceTableUtils: Persisting partitioned data source relation `test` into Hive metastore in Spark SQL specific format, which is NOT compatible with Hive. Input path(s): 
file:/Users/xiaoli/sparkBin/spark-2.0.2-bin-hadoop2.7/bin/spark-warehouse/test

scala> spark.sql("select * from test").show()
+--------+-------+
|fieldOne|partCol|
+--------+-------+
|       2|      2|
|       1|      1|
|       3|      3|
|       0|      0|
|       4|      4|
+--------+-------+

scala> spark.sql("desc formatted test").show(50, false)
+----------------------------+------------------------------------------------------------------------------+-------+
|col_name                    |data_type                                                                     |comment|
+----------------------------+------------------------------------------------------------------------------+-------+
...
|  path                      |file:/Users/xiaoli/sparkBin/spark-2.0.2-bin-hadoop2.7/bin/spark-warehouse/test|       |
+----------------------------+------------------------------------------------------------------------------+-------+


scala> spark.sql(s"create table newTab (fieldOne long, partCol int) using parquet options (path '/Users/xiaoli/sparkBin/spark-2.0.2-bin-hadoop2.7/bin/spark-warehouse/test') partitioned by (partCol)")
16/12/17 17:43:24 WARN CreateDataSourceTableUtils: Persisting partitioned data source relation `newTab` into Hive metastore in Spark SQL specific format, which is NOT compatible with Hive. Input path(s): 
file:/Users/xiaoli/sparkBin/spark-2.0.2-bin-hadoop2.7/bin/spark-warehouse/test
res3: org.apache.spark.sql.DataFrame = []

scala> spark.table("newTab").show()
+--------+-------+
|fieldOne|partCol|
+--------+-------+
|       2|      2|
|       1|      1|
|       3|      3|
|       0|      0|
|       4|      4|
+--------+-------+
```

gatorsmile:
Regarding the concern of the repair cost, I think we still face the same issue. Each time when we append an extra row, we also repair the table, right?  That is still expensive.

cloud-fan:
@gatorsmile I think this behavior change is necessary for the new "Scalable Partition Handling" feature. BTW, the table created by Spark 2.0 can still be read by Spark 2.1 without `REPAIR TABLE`, and users can fall back to the old behavior by turn off `spark.sql.hive.manageFilesourcePartitions`. So the backward compatibility is still good.

gatorsmile:
@cloud-fan 

First, if this behavior change is required, we need to document it. I think this is not clear to external users when they do not realize the underlying change since `spark.sql.hive.manageFilesourcePartitions` is set to true by default. 

Second, let us discuss the existing user interface when they creating data source tables. 
- **CREATE EXTERNAL  data source TABLE**: not allowed. 
- **CREATE MANAGED non-partitioned data source table without specifying the path in option**: an empty table
- **CREATE MANAGED non-partitioned data source table with specifying the path in option**: the data in the path is visible. No need to repair the table.
- **CREATE MANAGED partitioned data source table without specifying the path in option**: an empty table
- **CREATE MANAGED partitioned data source table with specifying the path in option**: 
  -- If `spark.sql.hive.manageFilesourcePartitions` is set to false, the data in the path is visible. 
  -- If `spark.sql.hive.manageFilesourcePartitions` is set to true, the data in the path is not visible. If they repair the table, they can see all the partitions. Without the repair, if they append the new rows to that partition, the existing/old rows in that partition will be visiable with the newly inserted row.
- **CREATE MANAGED partitioned hive serde table with specifying the LOCATION**: automatically converted to an **EXTERNAL** table in Spark 2.0+, but allowed by Hive. 

It is pretty complex to document/rememeber the above behaviors.
- Should we support `EXTERNAL` for data source tables? Like what we did for Hive serde tables, convert it to `EXTERNAL` table when users specify the `path` in `option`?
- If we decide to change the behaviors, should we follow Hive's MANAGED table when users specify a location? Ignore the existing data in the specified path/location?

gatorsmile:
We are facing another behavior inconsistency between `CTAS + Append` and `Insert Into`, because CTAS always repairt the table but InsertInto does not do it. Below is an example. 

```Scala
  test("append with saveAsTable + partition management") {
    withSQLConf(SQLConf.HIVE_MANAGE_FILESOURCE_PARTITIONS.key -> "true") {
      withTable("test") {
        withTempDir { dir =>
          setupPartitionedDatasourceTable("test", dir)
          spark.range(2).selectExpr("id as fieldOne", "id as partCol")
            .write.partitionBy("partCol").mode("append").saveAsTable("test")
          // all the partitions are visible
          assert(spark.sql("select * from test").count() == 7)
        }
      }
    }
  }


  test("insertInto + partition management") {
    withSQLConf(SQLConf.HIVE_MANAGE_FILESOURCE_PARTITIONS.key -> "true") {
      withTable("test") {
        withTempDir { dir =>
          setupPartitionedDatasourceTable("test", dir)
          spark.range(2).selectExpr("id as fieldOne", "id as partCol")
            .write.insertInto("test")
          // only the involved partitions are visible
          assert(spark.sql("select * from test").count() == 4)
        }
      }
    }
  }
```

cloud-fan:
> Should we support EXTERNAL for data source tables?

I'd like to hide the `EXTERNAL` concept to users, and only expose the `path` option or `LOCATION` syntax, for both data source and hive serde tables. It's pretty weird to have a managed table with a custom location, or an external table without custom location, and I don't think Hive's behavior makes sense when we rename, drop, alterLocation, ... these tables.
After we unify the CREATE TABLE syntax, we will only document the new syntax(which doesn't have the `EXTERNAL` concept), and leave the legacy hive syntax undocumented(but still supported).

> CTAS always repairs the table but InsertInto does not do it
As hive can't append a table with CTAS, we can define our own semantic here. InsertInto only update the involved partitions, I think it makes more sense to follow it for CTAS, which will be done in https://github.com/apache/spark/pull/15996

gatorsmile:
I see the plan, but the behavior difference will still be affected by the value of `spark.sql.hive.manageFilesourcePartitions`, right? 

I might need more time to chew over it to find out the potential impacts.

PR #15984, opened 2016-11-23
[SPARK-18551] [Web UI] [Core] [WIP] Add functionality to delete event logs from the History Server UI

ajbozarth:

Sometimes a Spark user will only have access to a History Server to interact with their (past) applications. But without access to the server they can only delete applications through use of the FS Cleaner feature, which itself can only clean logs older than a set date. 

I've added functionality to delete individual event logs from the history server, added a request handler to call said function and added a button to the UI to call it. The code is based on the current cleanLogs and kill jobs/stages code. I've also set this functionality to off by default.

I've left this as WIP because I still need to add tests and documentation for the new code and have design questions that I need input on (commented inline below)

## How was this patch tested?

Manual testing and dev/run-tests

spark18551

### DIFF
--- a/core/src/main/resources/org/apache/spark/ui/static/historypage-template.html
+++ b/core/src/main/resources/org/apache/spark/ui/static/historypage-template.html
@@ -77,7 +77,12 @@
       <td><span title="{{duration}}" class="durationClass">{{duration}}</span></td>
       <td>{{sparkUser}}</td>
       <td>{{lastUpdated}}</td>
-      <td><a href="{{uiroot}}/api/v1/applications/{{id}}/{{num}}/logs" class="btn btn-info btn-mini">Download</a></td>
+      <td>
+        <a href="{{uiroot}}/api/v1/applications/{{id}}/{{num}}/logs" class="btn btn-info btn-mini">Download</a>
+        <a href="{{uiroot}}/logs/delete/?appId={{id}}&attemptId={{attemptId}}"
+           onclick="if (window.confirm('Are you sure you want to delete the Event Log for {{id}} from the server?\n\nThis can not be undone!')) { this.parentNode.submit(); return true; } else { return false; }"
+           class="btn btn-danger btn-mini delete-app">Delete</a>
+      </td>
       {{/attempts}}
     </tr>
   {{/applications}}
--- a/core/src/main/resources/org/apache/spark/ui/static/historypage.js
+++ b/core/src/main/resources/org/apache/spark/ui/static/historypage.js
@@ -21,6 +21,12 @@ function setAppLimit(val) {
     appLimit = val;
 }
 
+var deleteLogsEnabled = false;
+
+function setDeleteLogsEnabled(val) {
+    deleteLogsEnabled = val;
+}
+
 function makeIdNumeric(id) {
   var strs = id.split("_");
   if (strs.length < 3) {
@@ -180,6 +186,10 @@ $(document).ready(function() {
 
         $(selector).DataTable(conf);
         $('#hisotry-summary [data-toggle="tooltip"]').tooltip();
+
+        if (deleteLogsEnabled) {
+          $('.delete-app').css('display', 'inline-block');
+        }
       });
     });
 });
--- a/core/src/main/scala/org/apache/spark/deploy/history/FsHistoryProvider.scala
+++ b/core/src/main/scala/org/apache/spark/deploy/history/FsHistoryProvider.scala
@@ -535,6 +535,25 @@ private[history] class FsHistoryProvider(conf: SparkConf, clock: Clock)
     }
   }
 
+  override def deleteLog(appId: String, attemptId: Option[String]): Unit = {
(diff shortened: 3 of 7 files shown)


core/src/main/resources/org/apache/spark/ui/static/historypage-template.html
```diff
@@ -77,7 +77,12 @@
       <td><span title="{{duration}}" class="durationClass">{{duration}}</span></td>
       <td>{{sparkUser}}</td>
       <td>{{lastUpdated}}</td>
-      <td><a href="{{uiroot}}/api/v1/applications/{{id}}/{{num}}/logs" class="btn btn-info btn-mini">Download</a></td>
+      <td>
+        <a href="{{uiroot}}/api/v1/applications/{{id}}/{{num}}/logs" class="btn btn-info btn-mini">Download</a>
+        <a href="{{uiroot}}/logs/delete/?appId={{id}}&attemptId={{attemptId}}"
+           onclick="if (window.confirm('Are you sure you want to delete the Event Log for {{id}} from the server?\n\nThis can not be undone!')) { this.parentNode.submit(); return true; } else { return false; }"
```
ajbozarth: I had trouble moving this js code into a separate file and still have it function as expected. If leaving it here is a problem I can try replicate its functionality by other methods. This was based on the kill job/stages code.

core/src/main/scala/org/apache/spark/deploy/history/FsHistoryProvider.scala
```diff
@@ -535,6 +535,26 @@ private[history] class FsHistoryProvider(conf: SparkConf, clock: Clock)
...
+            fs.delete(new Path(logDir, attempt.logPath), true)
+            applications.remove(appId)
+          } catch {
+            case e: AccessControlException =>
+              logInfo(s"No permission to delete ${attempt.logPath}, ignoring.")
+            case t: IOException =>
+              logError(s"IOException in cleaning ${attempt.logPath}", t)
+          }
+        }
+      case None => logError(s"Can't delete, no application ${appId} found")
```
ajbozarth: I'm not sure if the way I'm handling the logging/exceptions here is the best possible. I'm open to input, this logic was based on logic in `writeEventLogs` and `cleanLogs` in this file.

core/src/main/resources/org/apache/spark/ui/static/historypage.js
```diff
@@ -21,6 +21,12 @@ function setAppLimit(val) {
     appLimit = val;
 }
 
+var deleteLogsEnabled = false;
```
ajbozarth: Should I put this in `history-common.js` instead?

core/src/main/scala/org/apache/spark/deploy/history/HistoryServer.scala
```diff
@@ -131,6 +139,16 @@ class HistoryServer(
     attachHandler(contextHandler)
   }
 
+  def handleDeleteLogRequest(request: HttpServletRequest): Unit = {
+    if (deleteLogsEnabled && securityManager.checkModifyPermissions(request.getRemoteUser)) {
```
ajbozarth: I'm not sure that modify permissions is correct here, should we restrict this further to admin?

ajbozarth (commented):
I've included comments below on specific code questions I'd like input on.

core/src/main/scala/org/apache/spark/deploy/history/FsHistoryProvider.scala
steveloughran: from a testability perspective, returning true/false allows the HistoryServerSuite to see what happens in the cases of: no matching app; no matching attempt

core/src/main/scala/org/apache/spark/deploy/history/HistoryServer.scala
```diff
@@ -121,6 +123,12 @@ class HistoryServer(
   def initialize() {
     attachPage(new HistoryPage(this))
 
+    attachHandler(createRedirectHandler(
+      "/logs/delete",
+      "/",
+      handleDeleteLogRequest,
+      httpMethods = Set("GET", "POST")))
+
```
steveloughran: it shouldn't be a GET for something side effecting; breaks the whole HTTP "no side effects on a GET" rule. Better: support DELETE for high-rest; POST if you want to support HTML <= 4.

steveloughran:
Like you note, tests will be good here. Don't forget the corner cases: unknown app, duplicate POSTs, known app but unknown attempt. I'm also curious about what the policy would be if there is >1 attempt in the history, but no attempt Id in the delete.

Also: would it ever be the case that a history provider didn't support deletion? I can see where that would arise, but suspect that in such a case, you'd just deploy it with the deletion option disabled

ajbozarth:
Note: Pausing work on this until JIRA discussion is addressed.

ajbozarth:
Closing this based on security and complexity issues raised by @vanzin on the JIRA

PR #15985, opened 2016-11-23
[SPARK-18545] [SQL] Verify number of hive client RPCs in PartitionedTablePerfStatsSuite

ericl:

This would help catch accidental O(n) calls to the hive client as in https://issues.apache.org/jira/browse/SPARK-18507

## How was this patch tested?

Checked that the test fails before https://issues.apache.org/jira/browse/SPARK-18507 was patched. cc @cloud-fan


cloud-fan:
LGTM, merging to master/2.1

PR #11888, opened 2016-03-22
[SPARK-14069][SQL] Improve SparkStatusTracker to also track executor information

cloud-fan:
## What changes were proposed in this pull request?

Track executor information like host and port, cache size, running tasks.

## How was this patch tested?

manual test

PR #15986, opened 2016-11-23
[SPARK-18553][CORE][branch-2.0] Fix leak of TaskSetManager following executor loss

JoshRosen:

This patch fixes a critical resource leak in the TaskScheduler which could cause RDDs and ShuffleDependencies to be kept alive indefinitely if an executor with running tasks is permanently lost and the associated stage fails.

This problem was originally identified by analyzing the heap dump of a driver belonging to a cluster that had run out of shuffle space. This dump contained several `ShuffleDependency` instances that were retained by `TaskSetManager`s inside the scheduler but were not otherwise referenced. Each of these `TaskSetManager`s was considered a "zombie" but had no running tasks and therefore should have been cleaned up. However, these zombie task sets were still referenced by the `TaskSchedulerImpl.taskIdToTaskSetManager` map.

Entries are added to the `taskIdToTaskSetManager` map when tasks are launched and are removed inside of `TaskScheduler.statusUpdate()`, which is invoked by the scheduler backend while processing `StatusUpdate` messages from executors. The problem with this design is that a completely dead executor will never send a `StatusUpdate`. There is [some code](https://github.com/apache/spark/blob/072f4c518cdc57d705beec6bcc3113d9a6740819/core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala#L338) in `statusUpdate` which handles tasks that exit with the `TaskState.LOST` state (which is supposed to correspond to a task failure triggered by total executor loss), but this state only seems to be used in Mesos fine-grained mode. There doesn't seem to be any code which performs per-task state cleanup for tasks that were running on an executor that completely disappears without sending any sort of final death message. The `executorLost` and [`removeExecutor`](https://github.com/apache/spark/blob/072f4c518cdc57d705beec6bcc3113d9a6740819/core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala#L527) methods don't appear to perform any cleanup of the `taskId -> *` mappings, causing the leaks observed here.

This patch's fix is to maintain a `executorId -> running task id` mapping so that these `taskId -> *` maps can be properly cleaned up following an executor loss.

There are some potential corner-case interactions that I'm concerned about here, especially some details in [the comment](https://github.com/apache/spark/blob/072f4c518cdc57d705beec6bcc3113d9a6740819/core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala#L523) in `removeExecutor`, so I'd appreciate a very careful review of these changes.

This PR is opened against branch-2.0, where I first observed this problem, but will also need to be fixed in master, branch-2.1, and branch-1.6 (which I'll do in followup PRs after this fix is reviewed and merged).

## How was this patch tested?

I added a new unit test to `TaskSchedulerImplSuite`. You can check out this PR as of 25e455e711b978cd331ee0f484f70fde31307634 to see the failing test.

cc @kayousterhout, @markhamstra, @rxin for review.

### DIFF
--- a/core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
+++ b/core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
@@ -337,18 +337,18 @@ private[spark] class TaskSchedulerImpl(
     var reason: Option[ExecutorLossReason] = None
     synchronized {
       try {
-        if (state == TaskState.LOST && taskIdToExecutorId.contains(tid)) {
-          // We lost this entire executor, so remember that it's gone
-          val execId = taskIdToExecutorId(tid)
-
-          if (executorIdToRunningTaskIds.contains(execId)) {
+        taskIdToTaskSetManager.get(tid) match {
+          case Some(taskSet) if state == TaskState.LOST =>
+            // TaskState.LOST is only used by the deprecated Mesos fine-grained scheduling mode,
+            // where each executor corresponds to a single task, so mark the executor as failed.
+            val execId = taskIdToExecutorId.getOrElse(tid, throw new IllegalStateException(
+              "taskIdToTaskSetManager.contains(tid) <=> taskIdToExecutorId.contains(tid)"))
             reason = Some(
               SlaveLost(s"Task $tid was lost, so marking the executor as lost as well."))
             removeExecutor(execId, reason.get)
             failedExecutor = Some(execId)
-          }
-        }
-        taskIdToTaskSetManager.get(tid) match {
+            taskSet.removeRunningTask(tid)
+            taskResultGetter.enqueueFailedTask(taskSet, tid, state, serializedData)
           case Some(taskSet) =>
             if (TaskState.isFinished(state)) {
               taskIdToTaskSetManager.remove(tid)
@@ -361,7 +361,7 @@ private[spark] class TaskSchedulerImpl(
             if (state == TaskState.FINISHED) {
               taskSet.removeRunningTask(tid)
               taskResultGetter.enqueueSuccessfulTask(taskSet, tid, serializedData)
-            } else if (Set(TaskState.FAILED, TaskState.KILLED, TaskState.LOST).contains(state)) {
+            } else if (Set(TaskState.FAILED, TaskState.KILLED).contains(state)) {
               taskSet.removeRunningTask(tid)
               taskResultGetter.enqueueFailedTask(taskSet, tid, state, serializedData)
             }
--- a/core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
+++ b/core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
@@ -91,7 +91,9 @@ private[spark] class TaskSchedulerImpl(
   // IDs of the tasks running on each executor
   private val executorIdToRunningTaskIds = new HashMap[String, HashSet[Long]]
 
-  def runningTasksByExecutors: Map[String, Int] = executorIdToRunningTaskIds.toMap.mapValues(_.size)
+  def runningTasksByExecutors(): Map[String, Int] = {
+    executorIdToRunningTaskIds.toMap.mapValues(_.size)
+  }
 
   // The set of executors we have on each host; this is used to compute hostsAlive, which
   // in turn is used to decide when we can attain data locality on a given host
--- a/core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
+++ b/core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
@@ -344,10 +344,12 @@ private[spark] class TaskSchedulerImpl(
               // where each executor corresponds to a single task, so mark the executor as failed.
               val execId = taskIdToExecutorId.getOrElse(tid, throw new IllegalStateException(
                 "taskIdToTaskSetManager.contains(tid) <=> taskIdToExecutorId.contains(tid)"))
-              reason = Some(
-                SlaveLost(s"Task $tid was lost, so marking the executor as lost as well."))
-              removeExecutor(execId, reason.get)
-              failedExecutor = Some(execId)
+              if (executorIdToRunningTaskIds.contains(execId)) {
+                reason = Some(
+                  SlaveLost(s"Task $tid was lost, so marking the executor as lost as well."))
+                removeExecutor(execId, reason.get)
+                failedExecutor = Some(execId)
+              }
               taskSet.removeRunningTask(tid)
               taskResultGetter.enqueueFailedTask(taskSet, tid, state, serializedData)
             } else {
--- a/core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
+++ b/core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
@@ -528,6 +528,8 @@ private[spark] class TaskSchedulerImpl(
    */
   private def removeExecutor(executorId: String, reason: ExecutorLossReason) {
     executorIdToRunningTaskIds.remove(executorId).foreach { taskIds =>
+      logDebug("Cleaning up TaskScheduler state for tasks " +
+        s"${taskIds.mkString("[", ",", "]")} on failed executor $executorId")
       taskIds.foreach { tid =>
         taskIdToExecutorId.remove(tid)
         taskIdToTaskSetManager.remove(tid)
--- a/core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
+++ b/core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
@@ -89,9 +89,11 @@ private[spark] class TaskSchedulerImpl(
   val nextTaskId = new AtomicLong(0)
 
   // Number of tasks running on each executor
-  private val executorIdToTaskCount = new HashMap[String, Int]
+  private val executorIdToRunningTaskIds = new HashMap[String, HashSet[Long]]
 
-  def runningTasksByExecutors(): Map[String, Int] = executorIdToTaskCount.toMap
+  def runningTasksByExecutors(): Map[String, Int] = synchronized {
+    executorIdToRunningTaskIds.toMap.mapValues(_.size)
+  }
 
   // The set of executors we have on each host; this is used to compute hostsAlive, which
   // in turn is used to decide when we can attain data locality on a given host
@@ -259,7 +261,7 @@ private[spark] class TaskSchedulerImpl(
             val tid = task.taskId
             taskIdToTaskSetManager(tid) = taskSet
             taskIdToExecutorId(tid) = execId
-            executorIdToTaskCount(execId) += 1
+            executorIdToRunningTaskIds(execId).add(tid)
             executorsByHost(host) += execId
             availableCpus(i) -= CPUS_PER_TASK
             assert(availableCpus(i) >= 0)
@@ -288,7 +290,7 @@ private[spark] class TaskSchedulerImpl(
     var newExecAvail = false
     for (o <- offers) {
       executorIdToHost(o.executorId) = o.host
-      executorIdToTaskCount.getOrElseUpdate(o.executorId, 0)
+      executorIdToRunningTaskIds.getOrElseUpdate(o.executorId, HashSet[Long]())
       if (!executorsByHost.contains(o.host)) {
         executorsByHost(o.host) = new HashSet[String]()
         executorAdded(o.executorId, o.host)
@@ -339,7 +341,7 @@ private[spark] class TaskSchedulerImpl(
           // We lost this entire executor, so remember that it's gone
           val execId = taskIdToExecutorId(tid)
 
-          if (executorIdToTaskCount.contains(execId)) {
+          if (executorIdToRunningTaskIds.contains(execId)) {
             reason = Some(
               SlaveLost(s"Task $tid was lost, so marking the executor as lost as well."))
             removeExecutor(execId, reason.get)
@@ -351,9 +353,7 @@ private[spark] class TaskSchedulerImpl(
             if (TaskState.isFinished(state)) {
               taskIdToTaskSetManager.remove(tid)
               taskIdToExecutorId.remove(tid).foreach { execId =>
-                if (executorIdToTaskCount.contains(execId)) {
-                  executorIdToTaskCount(execId) -= 1
-                }
+                executorIdToRunningTaskIds.remove(execId)
               }
             }
             if (state == TaskState.FINISHED) {
@@ -477,7 +477,7 @@ private[spark] class TaskSchedulerImpl(
     var failedExecutor: Option[String] = None
 
     synchronized {
-      if (executorIdToTaskCount.contains(executorId)) {
+      if (executorIdToRunningTaskIds.contains(executorId)) {
         val hostPort = executorIdToHost(executorId)
         logExecutorLoss(executorId, hostPort, reason)
         removeExecutor(executorId, reason)
@@ -525,7 +525,12 @@ private[spark] class TaskSchedulerImpl(
    * of any running tasks, since the loss reason defines whether we'll fail those tasks.
    */
   private def removeExecutor(executorId: String, reason: ExecutorLossReason) {
-    executorIdToTaskCount -= executorId
+    executorIdToRunningTaskIds.remove(executorId).foreach { taskIds =>
+      taskIds.foreach { tid =>
+        taskIdToExecutorId.remove(tid)
+        taskIdToTaskSetManager.remove(tid)
+      }
+    }
 
     val host = executorIdToHost(executorId)
     val execs = executorsByHost.getOrElse(host, new HashSet)
@@ -563,11 +568,11 @@ private[spark] class TaskSchedulerImpl(
   }
 
   def isExecutorAlive(execId: String): Boolean = synchronized {
-    executorIdToTaskCount.contains(execId)
+    executorIdToRunningTaskIds.contains(execId)
   }
 
   def isExecutorBusy(execId: String): Boolean = synchronized {
-    executorIdToTaskCount.getOrElse(execId, -1) > 0
+    executorIdToRunningTaskIds.get(execId).exists(_.nonEmpty)
   }
 
   // By default, rack is unknown
--- a/core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
+++ b/core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
@@ -353,7 +353,9 @@ private[spark] class TaskSchedulerImpl(
             if (TaskState.isFinished(state)) {
               taskIdToTaskSetManager.remove(tid)
               taskIdToExecutorId.remove(tid).foreach { execId =>
-                executorIdToRunningTaskIds.remove(execId)
+                if (executorIdToRunningTaskIds.contains(execId)) {
+                  executorIdToRunningTaskIds(execId).remove(tid)
+                }
               }
             }
             if (state == TaskState.FINISHED) {
--- a/core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
+++ b/core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
@@ -354,12 +354,7 @@ private[spark] class TaskSchedulerImpl(
               taskResultGetter.enqueueFailedTask(taskSet, tid, state, serializedData)
             } else {
               if (TaskState.isFinished(state)) {
-                taskIdToTaskSetManager.remove(tid)
-                taskIdToExecutorId.remove(tid).foreach { execId =>
-                  if (executorIdToRunningTaskIds.contains(execId)) {
-                    executorIdToRunningTaskIds(execId).remove(tid)
-                  }
-                }
+                cleanupTaskState(tid)
               }
               if (state == TaskState.FINISHED) {
                 taskSet.removeRunningTask(tid)
@@ -526,18 +521,29 @@ private[spark] class TaskSchedulerImpl(
   }
 
   /**
+   * Cleans up the TaskScheduler's state for tracking the given task.
+   */
+  private def cleanupTaskState(tid: Long): Unit = {
+    taskIdToTaskSetManager.remove(tid)
+    taskIdToExecutorId.remove(tid).foreach { executorId =>
+      executorIdToRunningTaskIds.get(executorId).foreach { _.remove(tid) }
+    }
+  }
+
+  /**
    * Remove an executor from all our data structures and mark it as lost. If the executor's loss
    * reason is not yet known, do not yet remove its association with its host nor update the status
    * of any running tasks, since the loss reason defines whether we'll fail those tasks.
    */
   private def removeExecutor(executorId: String, reason: ExecutorLossReason) {
+    // The tasks on the lost executor may not send any more status updates (because the executor
+    // has been lost), so they should be cleaned up here.
     executorIdToRunningTaskIds.remove(executorId).foreach { taskIds =>
       logDebug("Cleaning up TaskScheduler state for tasks " +
         s"${taskIds.mkString("[", ",", "]")} on failed executor $executorId")
-      taskIds.foreach { tid =>
-        taskIdToExecutorId.remove(tid)
-        taskIdToTaskSetManager.remove(tid)
-      }
+      // We do not notify the TaskSetManager of the task failures because that will
+      // happen below in the rootPool.executorLost() call.
+      taskIds.foreach(cleanupTaskState)
     }
 
     val host = executorIdToHost(executorId)
--- a/core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
+++ b/core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
@@ -354,12 +354,12 @@ private[spark] class TaskSchedulerImpl(
             if (TaskState.isFinished(state)) {
               cleanupTaskState(tid)
               taskSet.removeRunningTask(tid)
+              if (state == TaskState.FINISHED) {
+                taskResultGetter.enqueueSuccessfulTask(taskSet, tid, serializedData)
+              } else if (Set(TaskState.FAILED, TaskState.KILLED, TaskState.LOST).contains(state)) {
+                taskResultGetter.enqueueFailedTask(taskSet, tid, state, serializedData)
+              }
             }
-            if (state == TaskState.FINISHED) {
-              taskResultGetter.enqueueSuccessfulTask(taskSet, tid, serializedData)
-            } else if (Set(TaskState.FAILED, TaskState.KILLED, TaskState.LOST).contains(state)) {
-              taskResultGetter.enqueueFailedTask(taskSet, tid, state, serializedData)
-          }
           case None =>
             logError(
               ("Ignoring update with state %s for TID %s because its task set is gone (this is " +
--- a/core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
+++ b/core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
@@ -91,9 +91,7 @@ private[spark] class TaskSchedulerImpl(
   // IDs of the tasks running on each executor
   private val executorIdToRunningTaskIds = new HashMap[String, HashSet[Long]]
 
-  def runningTasksByExecutors(): Map[String, Int] = synchronized {
-    executorIdToRunningTaskIds.toMap.mapValues(_.size)
-  }
+  def runningTasksByExecutors: Map[String, Int] = executorIdToRunningTaskIds.toMap.mapValues(_.size)
 
   // The set of executors we have on each host; this is used to compute hostsAlive, which
   // in turn is used to decide when we can attain data locality on a given host
--- a/core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
+++ b/core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
@@ -88,7 +88,7 @@ private[spark] class TaskSchedulerImpl(
   // Incrementing task IDs
   val nextTaskId = new AtomicLong(0)
 
-  // Number of tasks running on each executor
+  // IDs of the tasks running on each executor
   private val executorIdToRunningTaskIds = new HashMap[String, HashSet[Long]]
 
   def runningTasksByExecutors(): Map[String, Int] = synchronized {
--- a/core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
+++ b/core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
@@ -367,7 +367,8 @@ private[spark] class TaskSchedulerImpl(
           case None =>
             logError(
               ("Ignoring update with state %s for TID %s because its task set is gone (this is " +
-                "likely the result of receiving duplicate task finished status updates)")
+                "likely the result of receiving duplicate task finished status updates) or its " +
+                "executor has been marked as failed.")
                 .format(state, tid))
         }
       } catch {
--- a/core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
+++ b/core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
@@ -350,20 +350,16 @@ private[spark] class TaskSchedulerImpl(
                 removeExecutor(execId, reason.get)
                 failedExecutor = Some(execId)
               }
-              taskSet.removeRunningTask(tid)
-              taskResultGetter.enqueueFailedTask(taskSet, tid, state, serializedData)
-            } else {
-              if (TaskState.isFinished(state)) {
-                cleanupTaskState(tid)
-              }
-              if (state == TaskState.FINISHED) {
-                taskSet.removeRunningTask(tid)
-                taskResultGetter.enqueueSuccessfulTask(taskSet, tid, serializedData)
-              } else if (Set(TaskState.FAILED, TaskState.KILLED).contains(state)) {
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
```diff
@@ -89,9 +89,11 @@ private[spark] class TaskSchedulerImpl(
   val nextTaskId = new AtomicLong(0)
 
   // Number of tasks running on each executor
-  private val executorIdToTaskCount = new HashMap[String, Int]
+  private val executorIdToRunningTaskIds = new HashMap[String, HashSet[Long]]
```
JoshRosen: To pre-emptively address any concerns about the memory usage implications of this change, note that the hash set sizes should be bounded by the number of cores / task slots on the executor, so I don't think that there's much to be gained by using a more memory-efficient HashMap structure here. The only real optimization that I could think of would be to replace the map with a fixed-size array that we just linearly scan, but that seems like premature optimization and adds a lot of hard-to-reason-about complexity.

core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
```diff
@@ -89,9 +89,11 @@ private[spark] class TaskSchedulerImpl(
...
 
-  def runningTasksByExecutors(): Map[String, Int] = executorIdToTaskCount.toMap
+  def runningTasksByExecutors(): Map[String, Int] = synchronized {
```
JoshRosen: The `synchronized` here fixes a thread-safety bug added in #11888

core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
```diff
@@ -525,7 +525,12 @@ private[spark] class TaskSchedulerImpl(
    * of any running tasks, since the loss reason defines whether we'll fail those tasks.
```
JoshRosen: Based on the ["tasks are not re-scheduled while executor loss reason is pending" test](https://github.com/apache/spark/blob/072f4c518cdc57d705beec6bcc3113d9a6740819/core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala#L268) in `TaskSchedulerImplSuite`, it looks like the API contract here is that if `executorLost` is called with `LossReasonPending` then it will eventually be called with some other reason. This will cause it to [call](https://github.com/apache/spark/pull/15986/files#diff-d4000438827afe3a185ae75b24987a61R550) `rootPool.executorLost()` , which, in turn, will call `executorLost` for all TaskSetManagers, which will perform their own internal executorId to task id mapping to mark tasks as failed and inform the DAGScheduler. The `TaskSetManager` doesn't call back into the `TaskScheduler` to access any of the data in these mappings so I think it's safe to clean them up immediately at the top of `removeExecutor` rather than putting them behind the `reason != LossReasonPending` check.

Note that it's also not as simple as just putting those behind `reason != LossReasonPending` as a defensive measure because then we'd be changing the contract on when `runningTasksByExecutors()` is updated: previously, it would set a failed executor's running task count to zero as soon as the executor failed, whereas it would do it only after the reason was known should we move this update behind that check.

I think that these subtleties / distinctions are only relevant to YARN mode, so I'll loop in @vanzin to comment on them.

core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
```diff
@@ -339,7 +341,7 @@ private[spark] class TaskSchedulerImpl(
           // We lost this entire executor, so remember that it's gone
           val execId = taskIdToExecutorId(tid)
 
-          if (executorIdToTaskCount.contains(execId)) {
+          if (executorIdToRunningTaskIds.contains(execId)) {
             reason = Some(
               SlaveLost(s"Task $tid was lost, so marking the executor as lost as well."))
             removeExecutor(execId, reason.get)
```
JoshRosen: Following the call to `removeExecutor` here, the `taskIdToTaskSetManager.get(tid)` call immediately following this `if` will be guaranteed to return `None`. This is going to slightly change the behavior of how `TaskState.LOST` events are handled, which, given my earlier comments, I believe is only relevant to Mesos fine-grained mode (I think this is underscored by the error message `s"Task $tid was lost, so marking the executor as lost as well."`, since this only makes sense if you assume one task per executor).

Given this, I'm planning to refactor this block slightly so that the body if the `if` corresponds to the old code and so that the code following the `if` is only run for non `TaskState.LOST` end reasons.

JoshRosen: I added some logging in this `TaskState.LOST` branch and it looks like this case isn't hit at all in our existing scheduler tests.

rxin: cc @mgummelt

JoshRosen: More archaeology:

It looks like `TaskState.LOST` was introduced in Spark 0.6.0 as part of a refactoring to make the cluster scheduler pluggable: https://github.com/apache/spark/commit/e72afdb817bcc8388aeb8b8d31628fd5fd67acf1. That commit is from July 2012.

At the time, standalone mode didn't even exist and the schedulers were Mesos and local mode, and only Mesos fine-grained mode was supported. The only way to get a `TaskState.LOST` state was to convert the Mesos task loss state to it.

core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
```diff
@@ -274,4 +274,30 @@ class TaskSchedulerImplSuite extends SparkFunSuite with LocalSparkContext with L
...
+    // submit attempt 1, offer resources, task gets scheduled
+    taskScheduler.submitTasks(attempt1)
+    val taskDescriptions = taskScheduler.resourceOffers(e0Offers).flatten
+    assert(1 === taskDescriptions.length)
+
+    // mark executor0 as dead
+    taskScheduler.executorLost("executor0", SlaveLost())
+
+    // Check that state associated with the lost task attempt is cleaned up:
+    assert(taskScheduler.taskIdToExecutorId.isEmpty)
```
JoshRosen: I suppose that we should also strengthen the assertions in the existing tests to check that these maps are updated following task successes, but this may be tricky given that the existing tests aren't exercising the `statusUpdate` path. Rather, we may have to test this more end-to-end by asserting that these always become empty once all jobs and tasks are done.

core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
JoshRosen: To look at this another way, note that `TaskSchedulerImpl` is only called from three places:

- LocalSchedulerBackend, which won't use `TaskState.LOST`
- MesosFineGrainedSchedulerBackend, where `TaskState.LOST` means the total loss of an executor that corresponded to a single task (due to fine-grained mode)
- CoarseGrainedSchedulerBackend, where this is only called with a state that comes from a `StatusUpdate` message sent by an executor. This task state will never be `TaskState.LOST`.

Given all of this, I think that the right course of action here is to update the comments to clarify that `TaskState.LOST` is only relevant to fine-grained Mesos mode and to refactor this block to call

```
              taskSet.removeRunningTask(tid)
              taskResultGetter.enqueueFailedTask(taskSet, tid, state, serializedData)
```

after calling `removeExecutor` for the fine-grained task, then skipping the rest of the logic which only applies to local mode or coarse-grained schedulers.

rxin:
cc @tejasapatil

JoshRosen:
I managed to come up with a standalone end-to-end reproduction of the shuffle file leak, allowing me to validate this patch's fix.

Run

```
./bin/spark-shell --master=local-cluster[2,5,1024] --conf spark.task.maxFailures=1 --conf spark.local.dir=/tmp
```

to bring up a Spark shell with two executor JVMs. Then, execute the following:

```scala
sc.parallelize(1 to 10, 10).map { x => Thread.sleep(1000); (x, x) }.groupByKey(10).map { 
    case _ => Thread.sleep(120 * 1000); 1
}.union(sc.parallelize(1 to 10)).count()
```

(Note that the `.union()` here is critical for the reproduction; I explain this below).

The `Thread.sleep()` calls were strategically chosen so that we'll get the executor JVMs into a state where both executors have run shuffle map tasks and both are in the middle of running reduce / result tasks.

Next, kill one of the executor JVMs abruptly with `kill -9`. The worker JVM will immediately detect its executor JVM's death and will send messages to the master causing that executor's tasks to be marked as failed. Because of `spark.task.maxFailures=1` this will cause the job to immediately fail but there will still be five running zombie tasks on the executor that we didn't kill.

Wait until those zombie tasks have finished (which will happen within two minutes), then run `System.gc()`, then check the non-killed executor's block manager directories and observe that shuffle files have been leaked. This is due to the leak of the `ShuffleDependency`, which can be validated with `jmap -histo`:

```bash
$ jmap -histo 72081 | grep ShuffleDependency
2037:             1             56  org.apache.spark.ShuffleDependency
```

This is because the `TaskSetManager` was leaked:

```
 jmap -histo 72081 | grep 'org.apache.spark.scheduler.TaskSetManager$'
1252:             1            224  org.apache.spark.scheduler.TaskSetManager
```

Note that while executor death seems to always leak a `TaskSetManager`, this doesn't always result in a leaked `ShuffleDependency`; the reasons for this are slightly subtle and I can expand on them later, but to summarize in a nutshell: a `Task` whose partition is a `ShuffleRddPartition` won't actually contain a reference to the parent RDD; the parent RDD and `ShuffleDependency` will be kept alive in the scheduler via the parent stage and via inter-stage relationships, but there won't be a direct reference chain from the `Task` itself. On the other hand, some partition types such as `UnionRDDPartition` may have transient references to parent RDD objects, causing the driver-side `Task` to keep the whole RDD and ShuffleDependency lineage chain alive. This usually isn't a problem since `Task`s typically don't get leaked like this and the `@transient` fields prevent us from over-capturing during serialization, but it exacerbates the `TaskSetManager` leaks here.

After applying this PR's changes, you can re-run the same experiment and see that both the `TaskSetManager` and `ShuffleDependency` are properly cleaned up after the zombie tasks finish and GC has run to trigger the ContextCleaner.

core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
```diff
@@ -351,9 +353,7 @@ private[spark] class TaskSchedulerImpl(
             if (TaskState.isFinished(state)) {
               taskIdToTaskSetManager.remove(tid)
               taskIdToExecutorId.remove(tid).foreach { execId =>
-                if (executorIdToTaskCount.contains(execId)) {
-                  executorIdToTaskCount(execId) -= 1
-                }
+                executorIdToRunningTaskIds.remove(execId)
```
JoshRosen: Spotted a bug in this fix: this should actually be `executorIdToTaskCount(execId).remove(tid)`.

core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
```diff
@@ -335,31 +337,31 @@ private[spark] class TaskSchedulerImpl(
     var reason: Option[ExecutorLossReason] = None
     synchronized {
       try {
-        if (state == TaskState.LOST && taskIdToExecutorId.contains(tid)) {
```
JoshRosen: To understand this change, note that `taskIdToExecutorId` and `taskIdToTaskSetManager` should always have the same key set, so the `case Some(taskSet) if state == TaskState.LOST` is equivalent to this `if` statement.

core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
```diff
@@ -335,31 +337,31 @@ private[spark] class TaskSchedulerImpl(
...
-          // We lost this entire executor, so remember that it's gone
-          val execId = taskIdToExecutorId(tid)
-
-          if (executorIdToTaskCount.contains(execId)) {
+        taskIdToTaskSetManager.get(tid) match {
+          case Some(taskSet) if state == TaskState.LOST =>
+            // TaskState.LOST is only used by the deprecated Mesos fine-grained scheduling mode,
+            // where each executor corresponds to a single task, so mark the executor as failed.
+            val execId = taskIdToExecutorId.getOrElse(tid, throw new IllegalStateException(
```
JoshRosen: It would always be safe to just do a blind `get()` here, but I added the `getOrElse` and exception to explain why this is safe.

core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
```diff
@@ -335,31 +337,31 @@ private[spark] class TaskSchedulerImpl(
...
+              "taskIdToTaskSetManager.contains(tid) <=> taskIdToExecutorId.contains(tid)"))
             reason = Some(
               SlaveLost(s"Task $tid was lost, so marking the executor as lost as well."))
             removeExecutor(execId, reason.get)
             failedExecutor = Some(execId)
-          }
-        }
-        taskIdToTaskSetManager.get(tid) match {
+            taskSet.removeRunningTask(tid)
```
JoshRosen: Previously, these lines would be executed for the `TaskState.LOST` case by continuing onwards after the  `if (state == TaskState.LOST && taskIdToExecutorId.contains(tid))` block finished. The problem with doing that here is that my changes in `removeExecutor()` will have already removed this task from `taskIdToTaskSetManager` so you'd get an incorrect `Ignoring update with state ...` error. By moving the logic from that original `if` case into this `case Some(...) if ...` case it's easy to keep the shared logic for the "unknown task id" case while avoiding hitting that case spuriously.

core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
```diff
@@ -335,31 +337,31 @@ private[spark] class TaskSchedulerImpl(
...
-                  executorIdToTaskCount(execId) -= 1
+                if (executorIdToRunningTaskIds.contains(execId)) {
+                  executorIdToRunningTaskIds(execId).remove(tid)
                 }
               }
             }
             if (state == TaskState.FINISHED) {
               taskSet.removeRunningTask(tid)
               taskResultGetter.enqueueSuccessfulTask(taskSet, tid, serializedData)
-            } else if (Set(TaskState.FAILED, TaskState.KILLED, TaskState.LOST).contains(state)) {
```
JoshRosen: Here, `TaskState.LOST` will have already been handled by the new case above, so I removed it here because this case will never be hit.

core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
```diff
@@ -335,31 +337,31 @@ private[spark] class TaskSchedulerImpl(
...
-          if (executorIdToTaskCount.contains(execId)) {
+        taskIdToTaskSetManager.get(tid) match {
+          case Some(taskSet) if state == TaskState.LOST =>
+            // TaskState.LOST is only used by the deprecated Mesos fine-grained scheduling mode,
+            // where each executor corresponds to a single task, so mark the executor as failed.
+            val execId = taskIdToExecutorId.getOrElse(tid, throw new IllegalStateException(
+              "taskIdToTaskSetManager.contains(tid) <=> taskIdToExecutorId.contains(tid)"))
             reason = Some(
               SlaveLost(s"Task $tid was lost, so marking the executor as lost as well."))
             removeExecutor(execId, reason.get)
```
markhamstra: The handling of `var failedExecutor` and `var reason`, while not wholly new, appears odd.  They are only set within the `TaskState.LOST` case, and only used within that case or in the post-try block protected by `failedExecutor.isDefined`.  Seems like that post-try logic can be moved into the `TaskState.LOST` case and the handling of  `failedExecutor` and `reason` simplified.

core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
```diff
@@ -274,4 +274,30 @@ class TaskSchedulerImplSuite extends SparkFunSuite with LocalSparkContext with L
...
+    sc = new SparkContext("local", "TaskSchedulerImplSuite")
+    val taskScheduler = new TaskSchedulerImpl(sc)
+    taskScheduler.initialize(new FakeSchedulerBackend)
+    // Need to initialize a DAGScheduler for the taskScheduler to use for callbacks.
+    new DAGScheduler(sc, taskScheduler) {
+      override def taskStarted(task: Task[_], taskInfo: TaskInfo) {}
+      override def executorAdded(execId: String, host: String) {}
+    }
+
+    val e0Offers = Seq(new WorkerOffer("executor0", "host0", 1))
```
markhamstra: nit:  remove `new`

core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
```diff
@@ -88,10 +88,12 @@ private[spark] class TaskSchedulerImpl(
   // Incrementing task IDs
   val nextTaskId = new AtomicLong(0)
 
-  // Number of tasks running on each executor
-  private val executorIdToTaskCount = new HashMap[String, Int]
+  // IDs of the tasks running on each executor
+  private val executorIdToRunningTaskIds = new HashMap[String, HashSet[Long]]
 
-  def runningTasksByExecutors(): Map[String, Int] = executorIdToTaskCount.toMap
+  def runningTasksByExecutors(): Map[String, Int] = synchronized {
```
markhamstra: Is there a reason why this shouldn't be `def runningTasksByExecutors: Map[String, Int]` -- it's not a mutator.

markhamstra:
The code looks pretty good @JoshRosen , but I still want to spend some time looking at your standalone end-to-end reproduction to get more familiar with the details.

core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
```diff
@@ -525,7 +527,14 @@ private[spark] class TaskSchedulerImpl(
    * of any running tasks, since the loss reason defines whether we'll fail those tasks.
    */
   private def removeExecutor(executorId: String, reason: ExecutorLossReason) {
-    executorIdToTaskCount -= executorId
+    executorIdToRunningTaskIds.remove(executorId).foreach { taskIds =>
```
kayousterhout: Can you add a comment here -- something like 

The tasks on the lost executor may not send any more status updates (because the executor has been lost), so they should be cleaned up here.

core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
```diff
@@ -525,7 +527,14 @@ private[spark] class TaskSchedulerImpl(
...
+      logDebug("Cleaning up TaskScheduler state for tasks " +
+        s"${taskIds.mkString("[", ",", "]")} on failed executor $executorId")
+      taskIds.foreach { tid =>
+        taskIdToExecutorId.remove(tid)
+        taskIdToTaskSetManager.remove(tid)
```
kayousterhout: What about moving the two above lines to a "cleanupTaskState" method (or something like that) that can be called here and in statusUpdate? I wonder if that will help us avoid state tracking issues in the future -- because it's clear that the same cleanup should be happening in both places.

Can you also add a comment here saying that we only need to cleanup state because telling the TSM that the task has failed happens via the rootPool.executorLost call?

core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
kayousterhout: Josh, my understanding is consistent with yours -- that if executorLost() is called with LossReasonPending, it will eventually be called again with a real loss reason.  To elaborate on what Marcelo said, executorIdToHost is used at the "gatekeeper" for this: on [this line](https://github.com/JoshRosen/spark/blob/078ac0e6321aeb72c670a65ec90b9c20ef0a7788/core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala#L487), we use the existence of an entry in executorIdToHost to determine whether we need to call removeExecutor (possibly for a 2nd time).

My only concern with the current approach is the following: the task doesn't get marked as finished in the task set manager until rootPool.executorLost gets called (i.e., until the 2nd removeExecutor call).  Is it possible for anything to go awry for task updates that happen between the first removeExecutor call (at which point the TaskSchedulerImpl's state for the task will be removed) and when the task is marked as failed and removed from the TSM?  I looked over this and I think things should be fine -- we'll just drop the update -- but can you update the error message we log on 369 (since now it can happen when an update happens after an executor has been lost, in addition to when duplicate updates are received)?

core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
kayousterhout: Yeah is it prohibitively difficult to add a 2nd test for when the executorLost happens as a result of a status update call (e.g., have two tasks running on a particular executor, fail one of them via statusUpdate, and make sure the other one's state gets cleaned up?)?  The scheduler code is such a tangled web that it would be nice to test both code paths.

kayousterhout (commented):
Can you do the "synchronized" change in a separate PR? Worried about it getting lost here (/ being harder to track), and it seems easy to do separately after this is merged since it's just one line.

Otherwise, I looked over this carefully and it looks good.  Left a few suggestions to improve readability and try to avoid more issues in the future.

core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
JoshRosen: I agree. I originally held off on doing this because I wanted to minimize code movement, but at this phase I think it's worth doing.

core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
JoshRosen: No good reason; just a carryover from the old code.

core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
```diff
@@ -88,10 +88,12 @@ private[spark] class TaskSchedulerImpl(
...
   val nextTaskId = new AtomicLong(0)
 
-  // Number of tasks running on each executor
-  private val executorIdToTaskCount = new HashMap[String, Int]
+  // IDs of the tasks running on each executor
+  private val executorIdToRunningTaskIds = new HashMap[String, HashSet[Long]]
 
-  def runningTasksByExecutors(): Map[String, Int] = executorIdToTaskCount.toMap
+  def runningTasksByExecutors(): Map[String, Int] = synchronized {
+    executorIdToRunningTaskIds.toMap.mapValues(_.size)
```
JoshRosen: We need this to convert from a mutable to immutable map.

core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
vanzin: > Is it possible for anything to go awry for task updates that happen between the first removeExecutor call ... and when the task is marked as failed

If there's something really wrong with the YARN AM, it can take some time, but eventually the second call will come, either with the actual reason or with a generic "executor lost" reason. This is handled in `YarnSchedulerEndpoint.handleExecutorDisconnectedFromDriver`.

core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
kayousterhout: Sorry should have been more specific: why does the map need to be immutable?

core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
```diff
@@ -335,31 +337,31 @@ private[spark] class TaskSchedulerImpl(
     var reason: Option[ExecutorLossReason] = None
     synchronized {
       try {
-        if (state == TaskState.LOST && taskIdToExecutorId.contains(tid)) {
-          // We lost this entire executor, so remember that it's gone
-          val execId = taskIdToExecutorId(tid)
-
-          if (executorIdToTaskCount.contains(execId)) {
+        taskIdToTaskSetManager.get(tid) match {
+          case Some(taskSet) if state == TaskState.LOST =>
```
JoshRosen: Sure, can do. I held off on doing this originally because it would involve a lot more code movement and might have made the changes harder to review.

core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
JoshRosen: Yes. I think it's also needed for thread-safety. If we return a mutable map to the caller and they then go on to iterate over it then we're in trouble if that map isn't thread-safe.

core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
kayousterhout: Ah got it -- I appreciate that! But I do think that makes sense to go ahead and do.

core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
```diff
@@ -335,31 +337,31 @@ private[spark] class TaskSchedulerImpl(
     var reason: Option[ExecutorLossReason] = None
     synchronized {
       try {
-        if (state == TaskState.LOST && taskIdToExecutorId.contains(tid)) {
-          // We lost this entire executor, so remember that it's gone
-          val execId = taskIdToExecutorId(tid)
-
-          if (executorIdToTaskCount.contains(execId)) {
```
JoshRosen: Actually, we might want to keep this in case the Mesos scheduler sends us duplicate task failure events so that we don't try to remove an already-removed executor. I'll put this back in with an explanatory comment in my next commit.

core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
kayousterhout: But it doesn't matter because we're creating a new map with mapValues, right?

Anyway this seems not especially important to this PR so fine to leave as-is.

JoshRosen:
Thanks for the review pass. I'll split the thread-safety fix into a separate PR and will work on addressing your comments now.

I'll see about adding new unit tests to cover the `statusUpdate` case.

core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
JoshRosen: I don't think that case can occur because `TaskState.LOST` should only be sent in Mesos fine-grained mode where we'll only ever have a single task per executor (in other words, executorLoss can only happen via statusUpdate in situations where there would never be concurrent tasks on an executor).

JoshRosen:
I think that we can extend some of our existing tests in `TaskSchedulerImpl` suite to add the `statusUpdate()` calls that we might expect to see in the wild, but this will require a bit more mocking of various TaskScheduler components so things like `TaskGetter` aren't interacted with.

core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
markhamstra: Yup, good point.

core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
```diff
@@ -335,38 +337,38 @@ private[spark] class TaskSchedulerImpl(
...
-        }
         taskIdToTaskSetManager.get(tid) match {
           case Some(taskSet) =>
-            if (TaskState.isFinished(state)) {
-              taskIdToTaskSetManager.remove(tid)
-              taskIdToExecutorId.remove(tid).foreach { execId =>
-                if (executorIdToTaskCount.contains(execId)) {
-                  executorIdToTaskCount(execId) -= 1
-                }
+            if (state == TaskState.LOST) {
```
JoshRosen: I think this is technically okay, but the rationale is a little tricky:

If we receive a status update for a task with `TaskState.LOST` and the task corresponds to a task set which exists (i.e. we're taking this `if` branch), then there are two cases to consider:
  1. the task is running on an executor that we know about (i.e. `executorIdToRunningTaskIds.contains(execId) == true`), so we mark the executor as failed in the `removeExecutor()` call, which has the side-effect of calling `cleanupTaskState` for all tasks running on that executor (including this task).
  2. we don't know about the executor now, but we must have known about it earlier otherwise we wouldn't have scheduled a task there. Therefore, the executor was already removed earlier and that `removeExecutor()` call should have already called `cleanupTaskState()`. We shouldn't have to worry about races within `TaskSchedulerImpl` because all of the methods which interact with these data structures are synchronized.

That said, `cleanupTaskState()` is idempotent and there's not really any harm in putting in an extra call here just to make it absolutely clear that it's guaranteed to be cleaned up in all cases. I'll look into performing this restructuring.

core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
```diff
@@ -350,20 +350,16 @@ private[spark] class TaskSchedulerImpl(
                 removeExecutor(execId, reason.get)
                 failedExecutor = Some(execId)
               }
+            }
+            if (TaskState.isFinished(state)) {
+              cleanupTaskState(tid)
```
kayousterhout: ok last comment: do you think it's more readable to structure this code as:

`if (TaskState.isFinished(state) {
  if (state == TaskState.LOST) {
    <existing code>
    taskResultGetter.enqueueFailed
  } else if (SET(TaskState.FAILED, TaskState.KILLED).contains(state)) {
               taskResultGetter.enqueueFailedTask(taskSet, tid, state, serializedData)
  } else if (state == TaskState.FINISHED) {
       taskResultGetter.enqueueSuccessful(...)
  }

  cleanupTaskState
  taskSet.removeRunningTask
}`

core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
```diff
@@ -350,20 +350,16 @@ private[spark] class TaskSchedulerImpl(
...
-              } else if (Set(TaskState.FAILED, TaskState.KILLED).contains(state)) {
-                taskSet.removeRunningTask(tid)
-                taskResultGetter.enqueueFailedTask(taskSet, tid, state, serializedData)
-              }
             }
+            if (state == TaskState.FINISHED) {
+              taskResultGetter.enqueueSuccessfulTask(taskSet, tid, serializedData)
+            } else if (Set(TaskState.FAILED, TaskState.KILLED, TaskState.LOST).contains(state)) {
+              taskResultGetter.enqueueFailedTask(taskSet, tid, state, serializedData)
+          }
```
kayousterhout: if you don't take the idea above, fix indentation here

core/src/test/scala/org/apache/spark/scheduler/TaskSchedulerImplSuite.scala
```diff
@@ -274,4 +276,70 @@ class TaskSchedulerImplSuite extends SparkFunSuite with LocalSparkContext with L
...
+    assert(taskScheduler.getExecutorsAliveOnHost("host0").isEmpty)
+
+
+    // Check that state associated with the lost task attempt is cleaned up:
+    assert(taskScheduler.taskIdToExecutorId.isEmpty)
+    assert(taskScheduler.taskIdToTaskSetManager.isEmpty)
+    assert(taskScheduler.runningTasksByExecutors().get("executor0").isEmpty)
+  }
+
+  test("if a task finishes with TaskState.LOST then mark its executor as dead") {
```
kayousterhout: super nit but can you write this as "if a task finishes with TaskState.LOST its executor is marked as dead"

core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala
JoshRosen: I can do that, but I we might want to make sure that `taskSet.removeRunningTask` is called prior to the `taskResultGetter` call.

kayousterhout: I don't think that's necessary because all access to the TSM is gated on the TaskSchedulerImpl, so even though the TaskResultGetter might do some stuff, the TSM's state won't be accessed until the later handleSuccessfulTask call to the TaskSchedulerImpl

JoshRosen: Fair enough, although I ended up preserving the ordering just to reduce the set of changes that we need to reason about.

JoshRosen:
@kayousterhout, @markhamstra, I believe that I've addressed all of the outstanding review comments, so do you want to take a final look at this to see whether it's ready to go?

markhamstra:
If Kay is happy with the last couple of changes, then I'm fine with this, too.  The only tiny nit I've still got is a change from `runningTasksByExecutors()` to `runningTasksByExecutors`.  Outside of this PR, there's only a single call site in `SparkStatusTracker`, so the fix is pretty trivial -- but so is this issue itself, so I don't really care much if it stays as is.

kayousterhout (commented):
LGTM

kayousterhout:
Thanks for all of the work on this Josh! Happy to review the version for master if the merge isn't clean.

JoshRosen:
@markhamstra, I'll fix that in a followup patch to address the pre-existing synchronization / thread-safety issue impacting that method.

markhamstra:
Thanks, @JoshRosen

JoshRosen:
Alright, I'm going to merge this now and will open backport PRs against the other branches. Thanks for reviewing!

PR #15988, opened 2016-11-23
[SPARK-18519][SQL][BRANCH-2.0] map type can not be used in EqualTo

cloud-fan:

Technically map type is not orderable, but can be used in equality comparison. However, due to the limitation of the current implementation, map type can't be used in equality comparison so that it can't be join key or grouping key.

This PR makes this limitation explicit, to avoid wrong result.

backport https://github.com/apache/spark/pull/15956 to 2.0

## How was this patch tested?
updated tests


cloud-fan:
cc @hvanhovell

hvanhovell:
retest this please

hvanhovell:
LGTM. Merging to 2.0. Thanks!

Can you close?

PR #15990, opened 2016-11-23
[SPARK-18559] [SQL] Fix HLL++ with small relative error

wzhfy:

In `HyperLogLogPlusPlus`, if the relative error is so small that p >= 19, it will cause ArrayIndexOutOfBoundsException in `THRESHOLDS(p-4)` . We should check `p` and when p >= 19, regress to the original HLL result and use the small range correction they use.

The pr also fixes the upper bound in the log info in `require()`.
The upper bound is computed by:
```
val relativeSD = 1.106d / Math.pow(Math.E, p * Math.log(2.0d) / 2.0d)
```
which is derived from the equation for computing `p`:
```
val p = 2.0d * Math.log(1.106d / relativeSD) / Math.log(2.0d)
```

## How was this patch tested?

add test cases for:
1. checking validity of parameter relatvieSD
2. estimation with smaller relative error so that p >= 19

### DIFF
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/HyperLogLogPlusPlus.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/HyperLogLogPlusPlus.scala
@@ -307,6 +307,8 @@ case class HyperLogLogPlusPlus(
     val estimate = if (V > 0) {
       // Use linear counting for small cardinality estimates.
       val H = m * Math.log(m / V)
+      // HLL++ is defined only when p < 19, otherwise we need to fallback to HLL.
+      // The threshold `2.5 * m` is from the original HLL algorithm.
       if ((p < 19 && H <= THRESHOLDS(p - 4)) || E <= 2.5 * m) {
         H
       } else {
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/HyperLogLogPlusPlus.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/HyperLogLogPlusPlus.scala
@@ -92,11 +92,8 @@ case class HyperLogLogPlusPlus(
    */
   private[this] val p = Math.ceil(2.0d * Math.log(1.106d / relativeSD) / Math.log(2.0d)).toInt
 
-  // THRESHOLDS, RAW_ESTIMATE_DATA and BIAS_DATA all have the same length 15, and we probe these
-  // arrays by (p - 4), so we need to guarantee 0 <= p - 4 <= 14.
-  require(p >= 4 && p <= HyperLogLogPlusPlus.THRESHOLDS.length + 3,
-    s"HLL++ requires at least 4 bits and at most ${HyperLogLogPlusPlus.THRESHOLDS.length + 3} " +
-    "bits for addressing. The error should be in the range [0.22%, 39%].")
+  require(p >= 4, "HLL++ requires at least 4 bits for addressing. " +
+    "Use a lower error, at most 39%.")
 
   /**
    * Shift used to extract the index of the register from the hashed value.
@@ -299,8 +296,9 @@ case class HyperLogLogPlusPlus(
     // We integrate two steps from the paper:
     // val Z = 1.0d / zInverse
     // val E = alphaM2 * Z
+    val E = alphaM2 / zInverse
     @inline
-    def EBiasCorrected = alphaM2 / zInverse match {
+    def EBiasCorrected = E match {
       case e if p < 19 && e < 5.0d * m => e - estimateBias(e)
       case e => e
     }
@@ -309,7 +307,7 @@ case class HyperLogLogPlusPlus(
     val estimate = if (V > 0) {
       // Use linear counting for small cardinality estimates.
       val H = m * Math.log(m / V)
-      if (H <= THRESHOLDS(p - 4)) {
+      if ((p < 19 && H <= THRESHOLDS(p - 4)) || E <= 2.5 * m) {
         H
       } else {
         EBiasCorrected
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/HyperLogLogPlusPlus.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/HyperLogLogPlusPlus.scala
@@ -92,8 +92,11 @@ case class HyperLogLogPlusPlus(
    */
   private[this] val p = Math.ceil(2.0d * Math.log(1.106d / relativeSD) / Math.log(2.0d)).toInt
 
-  require(p >= 4, "HLL++ requires at least 4 bits for addressing. " +
-    "Use a lower error, at most 27%.")
+  // THRESHOLDS, RAW_ESTIMATE_DATA and BIAS_DATA all have the same length 15, and we probe these
+  // arrays by (p - 4), so we need to guarantee 0 <= p - 4 <= 14.
+  require(p >= 4 && p < HyperLogLogPlusPlus.THRESHOLDS.length + 4,
+    "HLL++ requires at least 4 bits and at most 18 bits for addressing. " +
+    "The error should be in the range [0.3%, 39%].")
 
   /**
    * Shift used to extract the index of the register from the hashed value.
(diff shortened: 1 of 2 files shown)


sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/HyperLogLogPlusPlus.scala
```diff
@@ -92,8 +92,11 @@ case class HyperLogLogPlusPlus(
    */
   private[this] val p = Math.ceil(2.0d * Math.log(1.106d / relativeSD) / Math.log(2.0d)).toInt
 
-  require(p >= 4, "HLL++ requires at least 4 bits for addressing. " +
-    "Use a lower error, at most 27%.")
+  // THRESHOLDS, RAW_ESTIMATE_DATA and BIAS_DATA all have the same length 15, and we probe these
+  // arrays by (p - 4), so we need to guarantee 0 <= p - 4 <= 14.
+  require(p >= 4 && p < HyperLogLogPlusPlus.THRESHOLDS.length + 4,
+    "HLL++ requires at least 4 bits and at most 18 bits for addressing. " +
```
srowen: Instead of hard-coding 18, this should be `${HyperLogLogPlusPlus.THRESHOLDS.length + 3}`. In which case it might be easier to make the condition above `<=` and `+ 3`. 

If you're going to 2 significant figures, I think 0.22% is a better lower bound.

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/aggregate/HyperLogLogPlusPlusSuite.scala
```diff
@@ -50,6 +50,16 @@ class HyperLogLogPlusPlusSuite extends SparkFunSuite {
     assert(error < hll.trueRsd * 3.0d, "Error should be within 3 std. errors.")
   }
 
+  test("test invalid parameter relativeSD") {
+    // `relativeSD` should be in the range [0.003, 0.39].
+    intercept[IllegalArgumentException] {
+      new HyperLogLogPlusPlus(new BoundReference(0, IntegerType, true), relativeSD = 0.002d)
```
srowen: I know the code around does this, but, writing `0.002d` isn't useful IMHO. `0.002` is clearer and already a double. I don't really care.

srowen (changes requested):

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/HyperLogLogPlusPlus.scala
hvanhovell: This LGTM. The only question I have is: Should we allow a user to use a value higher then `p=18` and just apply a max when we go above it?

hvanhovell:
LGTM - pending jenkins.

hvanhovell:
retest this please

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/aggregate/HyperLogLogPlusPlusSuite.scala
hvanhovell: yeah, the whole 0.002d style would be my bad.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/HyperLogLogPlusPlus.scala
hvanhovell: +1

wzhfy: Yea good point, fixed

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/aggregate/HyperLogLogPlusPlusSuite.scala
wzhfy: fixed

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/HyperLogLogPlusPlus.scala
wzhfy: Since we can't satisfy user's accuracy requirement when p > 18, I think it's better to stop and notify user, and let user decide whether he sets a lower accuracy, or chooses to use count distinct instead of count approximate distinct.

hvanhovell:
@wzhfy I did a little of digging. I think we are fixing something that is not broken. The current implementation does not apply bias correction if you set the relative error lower than 0.22%. See this line the code: https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/HyperLogLogPlusPlus.scala#L301

So I am fine with improving the error message, but not with capping the 'p' value.

Also note that bias correction only applies to an intermediate range: all HLL++ registries have to be in use and the raw estimate is smaller then `num_registries * 5`. So this only applies to a subset of use cases.

wzhfy:
@hvanhovell If you set the relative error lower than 0.22%, say 0.0021, we will get ArrayIndexOutOfBoundsException in `THRESHOLDS` before `EBiasCorrected`, see this line:
https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/HyperLogLogPlusPlus.scala#L309

hvanhovell:
Then fix that line.

wzhfy:
Do you mean fixing L309 by capping the `THRESHOLD` index `p-4`?
I just did some tests after such fixing, and if I use some relative error lower than 0.0022, the error I got is NOT within 3 std. errors.

hvanhovell:
It should regress to the original HLL result and use the small range correction they use. Something like this:
```scala
// We integrate two steps from the paper:
    // val Z = 1.0d / zInverse
    // val E = alphaM2 * Z
    val E = alphaM2 / zInverse
    @inline
    def EBiasCorrected = E match {
      case e if p < 19 && e < 5.0d * m => e - estimateBias(e)
      case e => e
    }

    // Estimate the cardinality.
    val estimate = if (V > 0) {
      // Use linear counting for small cardinality estimates.
      val H = m * Math.log(m / V)
      if (p < 19 && H <= THRESHOLDS(p - 4)) {
        H
      } else if (E <= 2.5 * m) {
        H
      } else {
        EBiasCorrected
      }
    } else {
      EBiasCorrected
    }
```

I don't think we should start throwing errors for things that used to work.

wzhfy:
I see, thanks! Updated.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/HyperLogLogPlusPlus.scala
```diff
@@ -306,7 +307,7 @@ case class HyperLogLogPlusPlus(
     val estimate = if (V > 0) {
       // Use linear counting for small cardinality estimates.
       val H = m * Math.log(m / V)
-      if (H <= THRESHOLDS(p - 4)) {
+      if ((p < 19 && H <= THRESHOLDS(p - 4)) || E <= 2.5 * m) {
```
srowen: I know the 2.5m bit is from the Flajolet paper but was it just missing here before?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/HyperLogLogPlusPlus.scala
```diff
@@ -306,7 +307,9 @@ case class HyperLogLogPlusPlus(
     val estimate = if (V > 0) {
       // Use linear counting for small cardinality estimates.
       val H = m * Math.log(m / V)
-      if (H <= THRESHOLDS(p - 4)) {
+      // HLL++ is defined only when p < 19, otherwise we need to fallback to HLL.
+      // The threshold `2.5 * m` is from the original HLL algorithm.
+      if ((p < 19 && H <= THRESHOLDS(p - 4)) || E <= 2.5 * m) {
```
srowen: My question is a little different -- I know it's from the paper, all of this is. I guess I'm not sure why this term didn't show up before? was that a bug? This logic doesn't look quite like what's in the paper but it's close.

srowen (approved):
Ah OK. If it looks OK to you, looks OK to me.

hvanhovell:

PR #15991, opened 2016-11-23
[SPARK-17843][WEB UI] Indicate event logs pending for processing on h…

vijoshi:

Backport PR #15410 to branch-2.0

## How was this patch tested?

Existing unit tests. Screenshots for UI changes provided in PR #15410.

### DIFF
--- a/core/src/main/scala/org/apache/spark/deploy/history/HistoryPage.scala
+++ b/core/src/main/scala/org/apache/spark/deploy/history/HistoryPage.scala
@@ -33,13 +33,30 @@ private[history] class HistoryPage(parent: HistoryServer) extends WebUIPage("") 
       .filter(_.completed != requestedIncomplete)
     val allAppsSize = allApps.size
 
+    val eventLogsUnderProcessCount = parent.getEventLogsUnderProcess()
+    val lastUpdatedTime = parent.getLastUpdatedTime()
     val providerConfig = parent.getProviderConfig()
     val content =
+      <script src={UIUtils.prependBaseUri("/static/historypage-common.js")}></script>
       <div>
           <div class="span12">
             <ul class="unstyled">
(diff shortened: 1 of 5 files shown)


vijoshi:
@tgravescs - Looking to get this change early with 2.0.x. This merges cleanly - are you ok to let this into branch-2.0?

core/src/main/scala/org/apache/spark/deploy/history/HistoryPage.scala
```diff
@@ -33,21 +33,40 @@ private[history] class HistoryPage(parent: HistoryServer) extends WebUIPage("")
...
+            if (eventLogsUnderProcessCount > 0) {
+              <p>There are {eventLogsUnderProcessCount} event log(s) currently being
+                processed which may result in additional applications getting listed on this page.
+                Refresh the page to view updates. </p>
+            }
+            }
+
+            {
+            if (lastUpdatedTime > 0) {
+              <p>Last updated: <span id="last-updated">{lastUpdatedTime}</span></p>
```
steveloughran: Is this just going to display a Long to the user? That's not going to be very informative. What about a human time where <= 0 is mapped to "never"?

steveloughran: I see —thanks. That's all I was worried about.

vijoshi:
@tgravesc - are you ok to merge this backport to branch-2.0 ?

tgravescs:
The reason I hadn't pulled it back is its listed as a minor improvement and generally we only pull defects back into the maintenance releases.

vijoshi:
@tgravescs would like to have this in 2.0 along with the other improvement that got accepted for backport (spark-18010). so would you consider allowing this into 2.0 as well?

tgravescs:
since its just in the history server and pretty isolated seems ok. +1

PR #15994, opened 2016-11-23
[SPARK-18555][SQL]DataFrameNaFunctions.fill miss up original values in long integers

windpiger:
  
   DataSet.na.fill(0) used on a DataSet which has a long value column, it will change the original long value.
  
   The reason is that the type of the function fill's param is Double, and the numeric columns are always cast to double(`fillCol[Double](f, value)`) .
```
  def fill(value: Double, cols: Seq[String]): DataFrame = {
    val columnEquals = df.sparkSession.sessionState.analyzer.resolver
    val projections = df.schema.fields.map { f =>
      // Only fill if the column is part of the cols list.
      if (f.dataType.isInstanceOf[NumericType] && cols.exists(col => columnEquals(f.name, col))) {
        fillCol[Double](f, value)
      } else {
        df.col(f.name)
      }
    }
    df.select(projections : _*)
  }
```

 For example:
```  
scala> val df = Seq[(Long, Long)]((1, 2), (-1, -2), (9123146099426677101L, 9123146560113991650L)).toDF("a", "b")
df: org.apache.spark.sql.DataFrame = [a: bigint, b: bigint]

scala> df.show
+-------------------+-------------------+
|                  a|                  b|
+-------------------+-------------------+
|                  1|                  2|
|                 -1|                 -2|
|9123146099426677101|9123146560113991650|
+-------------------+-------------------+

scala> df.na.fill(0).show
+-------------------+-------------------+
|                  a|                  b|
+-------------------+-------------------+
|                  1|                  2|
|                 -1|                 -2|
|9123146099426676736|9123146560113991680|
+-------------------+-------------------+
 ```

the original values changed [which is not we expected result]:
```
 9123146099426677101 -> 9123146099426676736
 9123146560113991650 -> 9123146560113991680
```

## How was this patch tested?

unit test added.

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/DataFrameNaFunctions.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/DataFrameNaFunctions.scala
@@ -133,7 +133,7 @@ final class DataFrameNaFunctions private[sql](df: DataFrame) {
   def fill(value: Long): DataFrame = fill(value, df.columns)
 
   /**
-   * Returns a new [[DataFrame]] that replaces null or NaN values in numeric columns with `value`.
+   * Returns a new `DataFrame` that replaces null or NaN values in numeric columns with `value`.
    * @since 1.3.1
    */
   def fill(value: Double): DataFrame = fill(value, df.columns)
@@ -154,7 +154,7 @@ final class DataFrameNaFunctions private[sql](df: DataFrame) {
   def fill(value: Long, cols: Array[String]): DataFrame = fill(value, cols.toSeq)
 
   /**
-   * Returns a new [[DataFrame]] that replaces null or NaN values in specified numeric columns.
+   * Returns a new `DataFrame` that replaces null or NaN values in specified numeric columns.
    * If a specified column is not a numeric column, it is ignored.
    *
    * @since 1.3.1
@@ -170,7 +170,7 @@ final class DataFrameNaFunctions private[sql](df: DataFrame) {
   def fill(value: Long, cols: Seq[String]): DataFrame = fillValue(value, cols)
 
   /**
-   * (Scala-specific) Returns a new [[DataFrame]] that replaces null or NaN values in specified
+   * (Scala-specific) Returns a new `DataFrame` that replaces null or NaN values in specified
    * numeric columns. If a specified column is not a numeric column, it is ignored.
    *
    * @since 1.3.1
--- a/sql/core/src/main/scala/org/apache/spark/sql/DataFrameNaFunctions.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/DataFrameNaFunctions.scala
@@ -128,6 +128,12 @@ final class DataFrameNaFunctions private[sql](df: DataFrame) {
   /**
    * Returns a new `DataFrame` that replaces null or NaN values in numeric columns with `value`.
    *
+   * @since 2.2.0
+   */
+  def fill(value: Long): DataFrame = fill(value, df.columns)
+
+  /**
+   * Returns a new [[DataFrame]] that replaces null or NaN values in numeric columns with `value`.
    * @since 1.3.1
    */
   def fill(value: Double): DataFrame = fill(value, df.columns)
@@ -143,6 +149,14 @@ final class DataFrameNaFunctions private[sql](df: DataFrame) {
    * Returns a new `DataFrame` that replaces null or NaN values in specified numeric columns.
    * If a specified column is not a numeric column, it is ignored.
    *
+   * @since 2.2.0
+   */
+  def fill(value: Long, cols: Array[String]): DataFrame = fill(value, cols.toSeq)
+
+  /**
+   * Returns a new [[DataFrame]] that replaces null or NaN values in specified numeric columns.
+   * If a specified column is not a numeric column, it is ignored.
+   *
    * @since 1.3.1
    */
   def fill(value: Double, cols: Array[String]): DataFrame = fill(value, cols.toSeq)
@@ -151,20 +165,18 @@ final class DataFrameNaFunctions private[sql](df: DataFrame) {
    * (Scala-specific) Returns a new `DataFrame` that replaces null or NaN values in specified
    * numeric columns. If a specified column is not a numeric column, it is ignored.
    *
+   * @since 2.2.0
+   */
+  def fill(value: Long, cols: Seq[String]): DataFrame = fillValue(value, cols)
+
+  /**
+   * (Scala-specific) Returns a new [[DataFrame]] that replaces null or NaN values in specified
+   * numeric columns. If a specified column is not a numeric column, it is ignored.
+   *
    * @since 1.3.1
    */
-  def fill(value: Double, cols: Seq[String]): DataFrame = {
-    val columnEquals = df.sparkSession.sessionState.analyzer.resolver
-    val projections = df.schema.fields.map { f =>
-      // Only fill if the column is part of the cols list.
-      if (f.dataType.isInstanceOf[NumericType] && cols.exists(col => columnEquals(f.name, col))) {
-        fillCol[Double](f, value)
-      } else {
-        df.col(f.name)
-      }
-    }
-    df.select(projections : _*)
-  }
+  def fill(value: Double, cols: Seq[String]): DataFrame = fillValue(value, cols)
+
 
   /**
    * Returns a new `DataFrame` that replaces null values in specified string columns.
@@ -180,18 +192,7 @@ final class DataFrameNaFunctions private[sql](df: DataFrame) {
    *
    * @since 1.3.1
    */
-  def fill(value: String, cols: Seq[String]): DataFrame = {
-    val columnEquals = df.sparkSession.sessionState.analyzer.resolver
-    val projections = df.schema.fields.map { f =>
-      // Only fill if the column is part of the cols list.
-      if (f.dataType.isInstanceOf[StringType] && cols.exists(col => columnEquals(f.name, col))) {
-        fillCol[String](f, value)
-      } else {
-        df.col(f.name)
-      }
-    }
-    df.select(projections : _*)
-  }
+  def fill(value: String, cols: Seq[String]): DataFrame = fillValue(value, cols)
 
   /**
    * Returns a new `DataFrame` that replaces null values.
@@ -210,7 +211,7 @@ final class DataFrameNaFunctions private[sql](df: DataFrame) {
    *
    * @since 1.3.1
    */
-  def fill(valueMap: java.util.Map[String, Any]): DataFrame = fill0(valueMap.asScala.toSeq)
+  def fill(valueMap: java.util.Map[String, Any]): DataFrame = fillMap(valueMap.asScala.toSeq)
 
   /**
    * (Scala-specific) Returns a new `DataFrame` that replaces null values.
@@ -230,7 +231,7 @@ final class DataFrameNaFunctions private[sql](df: DataFrame) {
    *
    * @since 1.3.1
    */
-  def fill(valueMap: Map[String, Any]): DataFrame = fill0(valueMap.toSeq)
+  def fill(valueMap: Map[String, Any]): DataFrame = fillMap(valueMap.toSeq)
 
   /**
    * Replaces values matching keys in `replacement` map with the corresponding values.
@@ -368,7 +369,7 @@ final class DataFrameNaFunctions private[sql](df: DataFrame) {
     df.select(projections : _*)
   }
 
-  private def fill0(values: Seq[(String, Any)]): DataFrame = {
+  private def fillMap(values: Seq[(String, Any)]): DataFrame = {
     // Error handling
     values.foreach { case (colName, replaceValue) =>
       // Check column name exists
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: sql/core/src/main/scala/org/apache/spark/sql/DataFrameNaFunctions.scala [hunk 1]
```diff
@@ -132,62 +132,42 @@ final class DataFrameNaFunctions private[sql](df: DataFrame) {
...
    * @since 1.3.1
    */
-  def fill(value: Double): DataFrame = fill(value, df.columns)
-
-  /**
-   * Returns a new [[DataFrame]] that replaces null values in string columns with `value`.
-   *
-   * @since 1.3.1
-   */
-  def fill(value: String): DataFrame = fill(value, df.columns)
```
srowen: You won't be able to remove public API methods like this. Why is this necessary if the idea is to handle Long differently?

windpiger: I thought fill[T](value: T) can handle the String Type, then I remove it.....

srowen: Doesn't it change the binary signature of the API? That is the problem.

windpiger:
@srowen I exclude the fill function from mima, is it ok?

windpiger:
cc @cloud-fan

project/MimaExcludes.scala
```diff
@@ -529,6 +529,7 @@ object MimaExcludes {
       ProblemFilters.exclude[IncompatibleMethTypeProblem]("org.apache.spark.mllib.evaluation.MulticlassMetrics.this"),
       ProblemFilters.exclude[IncompatibleMethTypeProblem]("org.apache.spark.mllib.evaluation.RegressionMetrics.this"),
       ProblemFilters.exclude[IncompatibleMethTypeProblem]("org.apache.spark.sql.DataFrameNaFunctions.this"),
+      ProblemFilters.exclude[IncompatibleMethTypeProblem]("org.apache.spark.sql.DataFrameNaFunctions.fill"),
```
windpiger: But although it is removed, the fill[T](value T) can handle this. It is backward compatibility. should we must keep the fill(value String) and fill(value Double) function?

rxin: The thing is they are not backward compatible at bytecode level, so applications will break if they are not rebuilt.

sql/core/src/main/scala/org/apache/spark/sql/DataFrameNaFunctions.scala
```diff
@@ -128,66 +128,49 @@ final class DataFrameNaFunctions private[sql](df: DataFrame) {
...
 
   /**
-   * Returns a new [[DataFrame]] that replaces null or NaN values in numeric columns with `value`.
+   * Returns a new [[DataFrame]] that replaces null or NaN values
+   * in numeric, boolean, string columns with `value`.
    *
    * @since 1.3.1
    */
-  def fill(value: Double): DataFrame = fill(value, df.columns)
+  def fill[T](value: T): DataFrame = fill(value, df.columns)
```
cloud-fan: instead of introducing `fill[T]`, which breaks backward compatibility, can we just add a new version for long type? i.e. `def fill(value: Long)`

windpiger: add a long type function is ok. I thought the string/long/double logic is the same ,so I want to merge them by a template function.

windpiger: this is  ok~
https://github.com/windpiger/spark/commit/360eafe7b7dc5135371c03b9493cad3af62b92ce

cloud-fan: I think we can have a `private def fill[T]`, but the public API can't be changed.

windpiger: ok, thanks a lot ! I have put it as a private. @cloud-fan

sql/core/src/main/scala/org/apache/spark/sql/DataFrameNaFunctions.scala
```diff
@@ -153,19 +168,20 @@ final class DataFrameNaFunctions private[sql](df: DataFrame) {
    * (Scala-specific) Returns a new [[DataFrame]] that replaces null or NaN values in specified
    * numeric columns. If a specified column is not a numeric column, it is ignored.
    *
+   * @since 2.1.0
+   */
+  def fill(value: Long, cols: Seq[String]): DataFrame = {
+    fill1[Long](value, cols)
```
cloud-fan: nit: `fill1(value, cols)` should work, scala has type inference.

sql/core/src/main/scala/org/apache/spark/sql/DataFrameNaFunctions.scala
```diff
@@ -437,4 +444,38 @@ final class DataFrameNaFunctions private[sql](df: DataFrame) {
...
+   */
+  private def fill1[T](value: T, cols: Seq[String]): DataFrame = {
+    value match {
+      case _: jl.Double | _: jl.Integer | _: jl.Float | _: jl.Boolean | _: jl.Long | _: String =>
+      case _ =>
+        throw new IllegalArgumentException(
+          s"Unsupported value type ${value.getClass.getName} ($value).")
+    }
+
+    val targetColumnType = value match {
```
cloud-fan: nit: we can combine the check here:
```
val targetType = value match {
  case _: Long => LongType
  case _: Double => DoubleType
  case _: String => StringType
  case _ => throw ...
}
```

sql/core/src/main/scala/org/apache/spark/sql/DataFrameNaFunctions.scala
```diff
@@ -437,4 +444,38 @@ final class DataFrameNaFunctions private[sql](df: DataFrame) {
...
+      case _: jl.Double | _: jl.Integer | _: jl.Float | _: jl.Long => NumericType
+      case _: jl.Boolean => BooleanType
+      case _: String => StringType
+    }
+
+    val columnEquals = df.sparkSession.sessionState.analyzer.resolver
+    val projections = df.schema.fields.map { f =>
+      // Only fill if the column is part of the cols list.
+      if (((f.dataType.isInstanceOf[NumericType] && targetColumnType == NumericType)
```
windpiger: Thanks!
 I have modified except one:
If T is a double type , this should be apply to all Numeric columns(include LongType/IntegerType), or just apply to FractionType?
The fill(value Double) apply to all Numeric columns, and I think fill(value Long) also keep the logic.

sql/core/src/main/scala/org/apache/spark/sql/DataFrameNaFunctions.scala
```diff
@@ -153,19 +168,20 @@ final class DataFrameNaFunctions private[sql](df: DataFrame) {
    * (Scala-specific) Returns a new [[DataFrame]] that replaces null or NaN values in specified
    * numeric columns. If a specified column is not a numeric column, it is ignored.
    *
+   * @since 2.1.0
+   */
+  def fill(value: Long, cols: Seq[String]): DataFrame = {
+    fill1(value, cols)
```
cloud-fan: nit: put it in one line? i.e. `def fill... = fill1...`

sql/core/src/main/scala/org/apache/spark/sql/DataFrameNaFunctions.scala
```diff
@@ -437,4 +444,38 @@ final class DataFrameNaFunctions private[sql](df: DataFrame) {
...
       s"Unsupported value type ${v.getClass.getName} ($v).")
   }
+
+  /**
+   * Returns a new [[DataFrame]] that replaces null or NaN values in specified
+   * numeric, string columns. If a specified column is not a numeric, string column,
+   * it is ignored.
+   */
+  private def fill1[T](value: T, cols: Seq[String]): DataFrame = {
+    // the fill[T] which T is  Long/Integer/Float/Double,
```
cloud-fan: why the T can be `Integer` and `Float`?

sql/core/src/main/scala/org/apache/spark/sql/DataFrameNaFunctions.scala
```diff
@@ -437,4 +444,38 @@ final class DataFrameNaFunctions private[sql](df: DataFrame) {
...
+    // should apply on all the NumericType Column, for example:
+    // val input = Seq[(java.lang.Integer, java.lang.Double)]((null, 164.3)).toDF("a","b")
+    // input.na.fill(3.1)
+    // the result is (3,164.3), not (null, 164.3)
+    val targetType = value match {
+      case _: jl.Double | _: jl.Integer | _: jl.Float | _: jl.Long => NumericType
```
cloud-fan: why we match `jd.Double` here intead of scala `Double`?

sql/core/src/main/scala/org/apache/spark/sql/DataFrameNaFunctions.scala
```diff
@@ -130,6 +130,13 @@ final class DataFrameNaFunctions private[sql](df: DataFrame) {
   /**
    * Returns a new [[DataFrame]] that replaces null or NaN values in numeric columns with `value`.
    *
+   * @since 2.1.0
+   */
+  def fill(value: Long): DataFrame = fill(value, df.columns)
+
+  /**
+   * Returns a new [[DataFrame]] that replaces null or NaN values in numeric columns with `value`.
```
HyukjinKwon: Could I ask change `[[DataFrame]]` to `` `DataFrame` ``?  It seems the `DataFrame` is unrecognisable via unidoc/genjavadoc (see https://github.com/apache/spark/pull/16013) which ends up with documentation build failure with Java 8.

sql/core/src/main/scala/org/apache/spark/sql/DataFrameNaFunctions.scala
windpiger: remove them is ok

sql/core/src/main/scala/org/apache/spark/sql/DataFrameNaFunctions.scala
HyukjinKwon: What if that one is merged first?

HyukjinKwon: nvm, I will let you know if that one is merged first. If this one is merged first, I will rebase and fix them together.

sql/core/src/main/scala/org/apache/spark/sql/DataFrameNaFunctions.scala
windpiger: fixed it

sql/core/src/main/scala/org/apache/spark/sql/DataFrameNaFunctions.scala
srowen: @HyukjinKwon yeah the bad news is that I'm sure the javadoc generation is going to re-break periodically. we can try to catch it with reviews and your work at least gets it to a working state. But we'll clean it up again before releases regularly.

sql/core/src/main/scala/org/apache/spark/sql/DataFrameNaFunctions.scala
```diff
@@ -437,4 +438,38 @@ final class DataFrameNaFunctions private[sql](df: DataFrame) {
     case v => throw new IllegalArgumentException(
       s"Unsupported value type ${v.getClass.getName} ($v).")
   }
+
+  /**
+   * Returns a new `DataFrame` that replaces null or NaN values in specified
+   * numeric, string columns. If a specified column is not a numeric, string column,
+   * it is ignored.
+   */
+  private def fill1[T](value: T, cols: Seq[String]): DataFrame = {
```
srowen: "fill1" isn't a good name. "fillInternal" maybe?

sql/core/src/main/scala/org/apache/spark/sql/DataFrameNaFunctions.scala
```diff
@@ -145,6 +151,14 @@ final class DataFrameNaFunctions private[sql](df: DataFrame) {
    * Returns a new [[DataFrame]] that replaces null or NaN values in specified numeric columns.
    * If a specified column is not a numeric column, it is ignored.
    *
+   * @since 2.1.0
```
srowen: I think this may be in 2.2.0 in the end

srowen (changes requested):

sql/core/src/main/scala/org/apache/spark/sql/DataFrameNaFunctions.scala
windpiger: ok,thanks

sql/core/src/main/scala/org/apache/spark/sql/DataFrameNaFunctions.scala
srowen: Yeah, that's a poor choice. I'd even suggest renaming it, while you rename this new method too.

windpiger: what about change `fill1` to `fillValue` ,and change `private def fill0(values: Seq[(String, Any)])` to `private def fillMap(values: Seq[(String, Any)])` ?

windpiger: @srowen

HyukjinKwon:
Hi @windpiger it seems something gone wrong. Would you try to rebase this please?

sql/core/src/main/scala/org/apache/spark/sql/DataFrameNaFunctions.scala
srowen: @windpiger I don't think this was quite resolved -- you need to back-tick DataFrame unfortunately to get it to work with javadoc 8.

srowen (changes requested):

windpiger:
@srowen fixed it.

srowen (approved):
That looks reasonable to me though I'd prefer to have @rxin give a final OK

rxin:
Thanks - merging in master.

rxin:
@windpiger you should use a proper email on your github commit next time. It is currently shown as:

```
Enter primary author in the format of "name <email>" [root <root@iZbp1gsnrlfzjxh82cz80vZ.(none)>]: 
```

windpiger:
ok ,thanks!

PR #15995, opened 2016-11-23
[SPARK-18566][SQL] remove OverwriteOptions

cloud-fan:

`OverwriteOptions` was introduced in https://github.com/apache/spark/pull/15705, to carry the information of static partitions. However, after further refactor, this information becomes duplicated and we can remove `OverwriteOptions`.

## How was this patch tested?

N/A


cloud-fan:
cc @yhuai @ericl

ericl:
Iirc there was an issue where `convertStaticPartitions` erased the partitioning information during analysis. Was this fixed by the refactoring?

cloud-fan:
@ericl you are right, I pushed a new commit to do `convertStaticPartitions` right before we convert `InsertIntoTable` to `InsertIntoHadoopFsRelation`, so the partitioning information won't be erased.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSourceStrategy.scala
```diff
@@ -129,65 +129,67 @@ case class DataSourceAnalysis(conf: CatalystConf) extends Rule[LogicalPlan] {
...
+        query
+      }
 
-    case i @ logical.InsertIntoTable(
-           l @ LogicalRelation(t: HadoopFsRelation, _, table), part, query, overwrite, false)
-        if query.resolved && t.schema.asNullable == query.schema.asNullable =>
+      // Sanity check, this should be guaranteed by `PreprocessTableInsertion`
+      if (t.schema.asNullable != actualQuery.schema.asNullable) {
+        throw new AnalysisException(
+          s"Can not insert into ${l.simpleString} because of schema mismatch.")
```
ericl: This seems to be triggered in the tests. Is it because the process rule no longer gets a chance to run now that the static resolution is combined with this rule?

cloud-fan: I'm going to submit another PR to tweak the extended analyzer rules execution order, currently these rules like `PreProcessCreateTable`, `PreWriteCheck`, `DataSourceAnalysis`, etc. may have some dependencies and may worth to put them in different batches.

BTW I have retargeted this JIRA ticket to 2.2

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSourceStrategy.scala
```diff
@@ -192,19 +200,13 @@ case class DataSourceAnalysis(conf: CatalystConf) extends Rule[LogicalPlan] {
...
       var customPartitionLocations: Map[TablePartitionSpec, String] = Map.empty
 
-      val staticPartitionKeys: TablePartitionSpec = if (overwrite.enabled) {
-        overwrite.staticPartitionKeys.map { case (k, v) =>
-          (partitionSchema.map(_.name).find(_.equalsIgnoreCase(k)).get, v)
-        }
-      } else {
-        Map.empty
-      }
+      val staticPartitions = parts.filter(_._2.nonEmpty).map { case (k, v) => k -> v.get }
```
cloud-fan: The column names in partition spec are already normalized in `PreprocessTableInsertion` rule, we don't need to consider case sensitivity here. And the `if-else` is not needed, because:
1. `staticPartitions` is used to get `matchingPartitions` in [this line](https://github.com/apache/spark/pull/15995/files#diff-d99813bd5bbc18277e4090475e4944cfR208), and the `matchingPartitions` is used to decided which partitions need to be [added to metastore](https://github.com/apache/spark/pull/15995/files#diff-d99813bd5bbc18277e4090475e4944cfR219). Previously if `overwrite` is false, we will get all partitions as `matchingPartitions`, and issue a lot of unnecessary `ADD PARTITION` calls. After removing the `if-else`, it's fixed.
2. After we pass `staticPartitions` to `InsertIntoHadoopFsRelationCommand`, it will be used only with `OverWriter` mode, so the `if-else` is unnecessary.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSourceStrategy.scala
```diff
@@ -128,61 +128,69 @@ case class DataSourceAnalysis(conf: CatalystConf) extends Rule[LogicalPlan] {
     projectList
   }
 
+  private def hasBeenPreprocessed(
```
ericl: Add a comment about who preprocesses this?

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSourceStrategy.scala
```diff
@@ -128,61 +128,69 @@ case class DataSourceAnalysis(conf: CatalystConf) extends Rule[LogicalPlan] {
...
+      tableOutput: Seq[Attribute],
+      partSchema: StructType,
+      partSpec: Map[String, Option[String]],
+      query: LogicalPlan): Boolean = {
+    val partColNames = partSchema.map(_.name).toSet
+    query.resolved && partSpec.keys.forall(partColNames.contains) && {
```
ericl: Is it necessary to check that the keys are all valid columns?

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSourceStrategy.scala
```diff
@@ -128,61 +128,69 @@ case class DataSourceAnalysis(conf: CatalystConf) extends Rule[LogicalPlan] {
...
+      val staticPartCols = partSpec.filter(_._2.isDefined).keySet
+      val expectedColumns = tableOutput.filterNot(a => staticPartCols.contains(a.name))
+      expectedColumns.toStructType.sameType(query.schema)
```
ericl: similar question, when is this false?

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSourceStrategy.scala
ericl: Ah, is the issue to avoid this running before PreprocessTableInsertion?

cloud-fan: yup

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSourceStrategy.scala
cloud-fan: this is to follow the previous condition: https://github.com/apache/spark/pull/15995/files#diff-d99813bd5bbc18277e4090475e4944cfL166

This can be caused if users issue an invalid command, e.g. `INSERT INTO src SELECT 1,2` while table `src` has 3 columns.

ericl:
Cool, this looks good then

cloud-fan:

PR #15996, opened 2016-11-23
[SPARK-18567][SQL] Simplify CreateDataSourceTableAsSelectCommand

cloud-fan:

The `CreateDataSourceTableAsSelectCommand` is quite complex now, as it has a lot of work to do if the table already exists:

1. throw exception if we don't want to ignore it.
2. do some check and adjust the schema if we want to append data.
3. drop the table and create it again if we want to overwrite.

The work 2 and 3 should be done by analyzer, so that we can also apply it to hive tables.

## How was this patch tested?

existing tests.

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/DataFrameWriter.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/DataFrameWriter.scala
@@ -23,11 +23,12 @@ import scala.collection.JavaConverters._
 
 import org.apache.spark.annotation.InterfaceStability
 import org.apache.spark.sql.catalyst.TableIdentifier
-import org.apache.spark.sql.catalyst.analysis.UnresolvedRelation
+import org.apache.spark.sql.catalyst.analysis.{EliminateSubqueryAliases, UnresolvedRelation}
 import org.apache.spark.sql.catalyst.catalog.{BucketSpec, CatalogTable, CatalogTableType}
 import org.apache.spark.sql.catalyst.plans.logical.InsertIntoTable
 import org.apache.spark.sql.execution.command.DDLUtils
-import org.apache.spark.sql.execution.datasources.{CreateTable, DataSource}
+import org.apache.spark.sql.execution.datasources.{CreateTable, DataSource, LogicalRelation}
+import org.apache.spark.sql.sources.BaseRelation
 import org.apache.spark.sql.types.StructType
 
 /**
@@ -364,7 +365,11 @@ final class DataFrameWriter[T] private[sql](ds: Dataset[T]) {
       throw new AnalysisException("Cannot create hive serde table with saveAsTable API")
     }
 
-    val tableExists = df.sparkSession.sessionState.catalog.tableExists(tableIdent)
+    val catalog = df.sparkSession.sessionState.catalog
+    val tableExists = catalog.tableExists(tableIdent)
+    val db = tableIdent.database.getOrElse(catalog.getCurrentDatabase)
+    val tableIdentWithDB = tableIdent.copy(database = Some(db))
+    val tableName = tableIdentWithDB.unquotedString
 
     (tableExists, mode) match {
       case (true, SaveMode.Ignore) =>
@@ -373,26 +378,48 @@ final class DataFrameWriter[T] private[sql](ds: Dataset[T]) {
       case (true, SaveMode.ErrorIfExists) =>
         throw new AnalysisException(s"Table $tableIdent already exists.")
 
-      case _ =>
-        val storage = DataSource.buildStorageFormatFromOptions(extraOptions.toMap)
-        val tableType = if (storage.locationUri.isDefined) {
-          CatalogTableType.EXTERNAL
-        } else {
-          CatalogTableType.MANAGED
+      case (true, SaveMode.Overwrite) =>
+        // Get all input data source relations of the query.
+        val srcRelations = df.logicalPlan.collect {
+          case LogicalRelation(src: BaseRelation, _, _) => src
+        }
+        EliminateSubqueryAliases(catalog.lookupRelation(tableIdentWithDB)) match {
+          // Only do the check if the table is a data source table (the relation is a BaseRelation).
+          case LogicalRelation(dest: BaseRelation, _, _) =>
+            if (srcRelations.contains(dest)) {
+              throw new AnalysisException(
+                s"Cannot overwrite table $tableName that is also being read from")
+            }
+          case _ => // OK
         }
 
-        val tableDesc = CatalogTable(
-          identifier = tableIdent,
-          tableType = tableType,
-          storage = storage,
-          schema = new StructType,
-          provider = Some(source),
-          partitionColumnNames = partitioningColumns.getOrElse(Nil),
-          bucketSpec = getBucketSpec
-        )
-        df.sparkSession.sessionState.executePlan(
-          CreateTable(tableDesc, mode, Some(df.logicalPlan))).toRdd
+        // Drop the existing table
+        catalog.dropTable(tableIdentWithDB, ignoreIfNotExists = true, purge = false)
+        createTable(tableIdent)
+
+      case _ => createTable(tableIdent)
     }
+  }
+
+  private def createTable(tableIdent: TableIdentifier): Unit = {
+    val storage = DataSource.buildStorageFormatFromOptions(extraOptions.toMap)
+    val tableType = if (storage.locationUri.isDefined) {
+      CatalogTableType.EXTERNAL
+    } else {
+      CatalogTableType.MANAGED
+    }
+
+    val tableDesc = CatalogTable(
+      identifier = tableIdent,
+      tableType = tableType,
+      storage = storage,
+      schema = new StructType,
+      provider = Some(source),
+      partitionColumnNames = partitioningColumns.getOrElse(Nil),
+      bucketSpec = getBucketSpec
+    )
+    df.sparkSession.sessionState.executePlan(
+      CreateTable(tableDesc, mode, Some(df.logicalPlan))).toRdd
   }
 
   /**
--- a/sql/core/src/main/scala/org/apache/spark/sql/DataFrameWriter.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/DataFrameWriter.scala
@@ -374,19 +374,8 @@ final class DataFrameWriter[T] private[sql](ds: Dataset[T]) {
         throw new AnalysisException(s"Table $tableIdent already exists.")
 
       case _ =>
-        val existingTable = if (tableExists) {
-          Some(df.sparkSession.sessionState.catalog.getTableMetadata(tableIdent))
-        } else {
-          None
-        }
-        val storage = if (tableExists) {
-          existingTable.get.storage
-        } else {
-          DataSource.buildStorageFormatFromOptions(extraOptions.toMap)
-        }
-        val tableType = if (tableExists) {
-          existingTable.get.tableType
-        } else if (storage.locationUri.isDefined) {
+        val storage = DataSource.buildStorageFormatFromOptions(extraOptions.toMap)
+        val tableType = if (storage.locationUri.isDefined) {
           CatalogTableType.EXTERNAL
         } else {
           CatalogTableType.MANAGED
@@ -441,7 +430,7 @@ final class DataFrameWriter[T] private[sql](ds: Dataset[T]) {
     assertNotPartitioned("jdbc")
     assertNotBucketed("jdbc")
     // connectionProperties should override settings in extraOptions.
-    this.extraOptions = this.extraOptions ++ connectionProperties.asScala
+    this.extraOptions ++= connectionProperties.asScala
     // explicit url and dbtable should override all
     this.extraOptions += ("url" -> url, "dbtable" -> table)
     format("jdbc").save()
@@ -588,7 +577,7 @@ final class DataFrameWriter[T] private[sql](ds: Dataset[T]) {
 
   private var mode: SaveMode = SaveMode.ErrorIfExists
 
-  private var extraOptions = new scala.collection.mutable.HashMap[String, String]
+  private val extraOptions = new scala.collection.mutable.HashMap[String, String]
 
   private var partitioningColumns: Option[Seq[String]] = None
 
--- a/sql/core/src/main/scala/org/apache/spark/sql/DataFrameWriter.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/DataFrameWriter.scala
@@ -385,11 +385,9 @@ final class DataFrameWriter[T] private[sql](ds: Dataset[T]) {
         }
         EliminateSubqueryAliases(catalog.lookupRelation(tableIdentWithDB)) match {
           // Only do the check if the table is a data source table (the relation is a BaseRelation).
-          case LogicalRelation(dest: BaseRelation, _, _) =>
-            if (srcRelations.contains(dest)) {
-              throw new AnalysisException(
-                s"Cannot overwrite table $tableName that is also being read from")
-            }
+          case LogicalRelation(dest: BaseRelation, _, _) if srcRelations.contains(dest) =>
+            throw new AnalysisException(
+              s"Cannot overwrite table $tableName that is also being read from")
           case _ => // OK
         }
 
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/command/createDataSourceTables.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/command/createDataSourceTables.scala
@@ -19,7 +19,6 @@ package org.apache.spark.sql.execution.command
 
 import org.apache.spark.sql._
 import org.apache.spark.sql.catalyst.catalog._
-import org.apache.spark.sql.catalyst.expressions.NamedExpression
 import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan
 import org.apache.spark.sql.execution.datasources._
 import org.apache.spark.sql.sources.BaseRelation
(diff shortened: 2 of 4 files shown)


cloud-fan:
cc @yhuai @gatorsmile 

The first commit is from another PR and you can ignore it. Do you think we should target this ticket to 2.1? It's kind of a refactor but do fix some problems.

sql/core/src/main/scala/org/apache/spark/sql/execution/command/createDataSourceTables.scala
```diff
@@ -117,7 +117,7 @@ case class CreateDataSourceTableCommand(table: CatalogTable, ignoreIfExists: Boo
  */
 case class CreateDataSourceTableAsSelectCommand(
     table: CatalogTable,
-    mode: SaveMode,
+    ignoreIfExists: Boolean,
```
gatorsmile: This is a good change, I like it! Now it is identical to the interface `CreateHiveTableAsSelectCommand`. Maybe we can copy the params here.

sql/core/src/main/scala/org/apache/spark/sql/execution/command/createDataSourceTables.scala
```diff
@@ -134,114 +134,57 @@ case class CreateDataSourceTableAsSelectCommand(
...
-            s"the existing data. " +
-            s"Or, if you are using SQL CREATE TABLE, you need to drop $tableName first.")
-        case SaveMode.Ignore =>
-          // Since the table already exists and the save mode is Ignore, we will just return.
-          return Seq.empty[Row]
-        case SaveMode.Append =>
-          // Check if the specified data source match the data source of the existing table.
-          val existingProvider = DataSource.lookupDataSource(provider)
-          // TODO: Check that options from the resolved relation match the relation that we are
-          // inserting into (i.e. using the same compression).
```
gatorsmile: This todo is still valid?

gatorsmile:
A general issue: after we moving the execution of `Append` and `Overwrite` into `DataFrameWriter`, the verification in `AnalyzeCreateTable` is not called. Some logics are still required.

cloud-fan:
@gatorsmile , can you point out which verification we need to add back from `AnalyzeCreateTable`?

cloud-fan:
retest this please

sql/core/src/main/scala/org/apache/spark/sql/DataFrameWriter.scala
```diff
@@ -363,48 +364,120 @@ final class DataFrameWriter[T] private[sql](ds: Dataset[T]) {
...
+              case o =>
+                throw new AnalysisException(s"Saving data in ${o.toString} is not supported.")
+            }
+
+            val existingSchema = tableMeta.schema
+            if (df.logicalPlan.schema.size != existingSchema.size) {
+              throw new AnalysisException(
+                s"The column number of the existing schema[$existingSchema] " +
+                  s"doesn't match the data schema[${df.logicalPlan.schema}]")
+            }
```
gatorsmile: uh, this fixes a bug. Before this PR, we only check the size when the target is the `LogicalRelation`.

sql/core/src/main/scala/org/apache/spark/sql/DataFrameWriter.scala
```diff
@@ -363,48 +364,120 @@ final class DataFrameWriter[T] private[sql](ds: Dataset[T]) {
...
+            if (specifiedPartCols.map(_.toLowerCase) != existingPartCols.map(_.toLowerCase)) {
+              throw new AnalysisException("The partition columns of the existing table " +
+                s"$tableIdent are: [${existingPartCols.mkString(", ")}]. It doesn't match the " +
+                s"specified partition columns: [${specifiedPartCols.mkString(", ")}]")
+            }
+
+            // Because we are inserting into an existing table, we should respect the existing
+            // schema and adjust columns order of the given dataframe according to it.
+            df.select(existingSchema.map(f => Column(f.name)): _*)
+              .write.insertInto(tableIdentWithDB)
```
gatorsmile: We will analyze the dataframe twice. It will cause a bug. 
```Scala
    withTable("bar") {
      withTempView("tab1") {
        spark.range(99, 101).createOrReplaceTempView("tab1")
        sql("SELECT id, cast(id as long) * cast('1.0' as decimal(38, 18)) as num FROM tab1")
          .write.format("parquet").saveAsTable("bar")

        sql("SELECT id, cast(id as long) * cast('1.0' as decimal(38, 18)) as num FROM tab1")
          .write.format("parquet").insertInto("bar")
        spark.table("bar").show()
      }
    }
```

The result will be
```
+---+--------------------+
| id|                 num|
+---+--------------------+
| 99|99.00000000000000...|
| 99|99.00000000000000...|
|100|100.0000000000000...|
|100|100.0000000000000...|
+---+--------------------+
```

cloud-fan: I thought it's ok to analyze twice, but not analyze an optimized plan, let me look into it.

gatorsmile: sorry, my mistake. Let me delete it.

gatorsmile: Sorry, I made the mistake here. I deleted the comment after I realized it. 

`assertNotBucketed("insertInto")` is missing here. This is an existing bug, right?

sql/core/src/main/scala/org/apache/spark/sql/DataFrameWriter.scala
```diff
@@ -363,48 +365,125 @@ final class DataFrameWriter[T] private[sql](ds: Dataset[T]) {
...
+                  s"(${df.logicalPlan.schema.catalogString})")
+            }
+
+            if (partitioningColumns.isDefined) {
+              logWarning("append to an existing table, the specified partition columns " +
+                s"[${partitioningColumns.get.mkString(", ")}] will be ignored.")
+            }
+
+            val specifiedBucketSpec = getBucketSpec
+            if (specifiedBucketSpec.isDefined) {
```
cloud-fan: These 2 checkings are newly added, previously we ignore the user specified partition columns and bucket silently, now we will log a warning message.

sql/core/src/main/scala/org/apache/spark/sql/DataFrameWriter.scala
```diff
@@ -363,48 +365,125 @@ final class DataFrameWriter[T] private[sql](ds: Dataset[T]) {
...
+                s"(${specifiedBucketSpec.get}) will be ignored.")
+            }
+
+            // Because we are inserting into an existing table, we should respect the existing
+            // schema and adjust columns order of the given dataframe according to it.
+            val adjustedColumns = existingSchema.map(f => Column(df.resolve(f.name)))
+            df.select(adjustedColumns: _*).write.insertInto(tableIdentWithDB)
+
+          case SaveMode.Overwrite =>
+            EliminateSubqueryAliases(catalog.lookupRelation(tableIdentWithDB)) match {
```
cloud-fan: This is mostly moved from https://github.com/apache/spark/pull/15996/files#diff-73bd90660f41c12a87ee9fe8d35d856aL314

sql/core/src/main/scala/org/apache/spark/sql/DataFrameWriter.scala
```diff
@@ -363,48 +365,125 @@ final class DataFrameWriter[T] private[sql](ds: Dataset[T]) {
...
-        df.sparkSession.sessionState.executePlan(
-          CreateTable(tableDesc, mode, Some(df.logicalPlan))).toRdd
+      // If the table doesn't exist...
+      case None => createTable(tableIdentWithDB)
     }
   }
 
+  private def createTable(tableIdent: TableIdentifier): Unit = {
+    val storage = DataSource.buildStorageFormatFromOptions(extraOptions.toMap)
+    val tableType = if (storage.locationUri.isDefined) {
```
cloud-fan: I reverted https://github.com/apache/spark/pull/15983 here because it's not needed anymore after this refactor.

gatorsmile:
Could we update the PR description and add the test case in `PartitionProviderCompatibilitySuite.scala` to reflect the external behavior changes of CTAS on partitioned data source tables?

sql/core/src/main/scala/org/apache/spark/sql/DataFrameWriter.scala
```diff
@@ -364,48 +365,149 @@ final class DataFrameWriter[T] private[sql](ds: Dataset[T]) {
...
+              "SaveMode.Overwrite to overwrite the existing data.")
+
+          case SaveMode.Append =>
+            if (existingTable.tableType == CatalogTableType.VIEW) {
+              throw new AnalysisException("Saving data into a view is not allowed.")
+            }
+
+            if (existingTable.provider.get == DDLUtils.HIVE_PROVIDER) {
+              throw new AnalysisException(s"Saving data in the Hive serde table $tableName is " +
+                s"not supported yet. Please use the insertInto() API as an alternative.")
```
gatorsmile: Nit: this string interpolation is not needed.

gatorsmile:
LGTM

sql/core/src/main/scala/org/apache/spark/sql/DataFrameWriter.scala
```diff
@@ -364,48 +366,157 @@ final class DataFrameWriter[T] private[sql](ds: Dataset[T]) {
...
+                   |Existing bucketing: $existingBucketString
+                """.stripMargin)
+            }
+
+            if (classOf[CreatableRelationProvider].isAssignableFrom(existingProvider)) {
+              // For data source that implementing `CreatableRelationProvider`, we should call its
+              // `createRelation` to do the data appending. `CreatableRelationProvider` may not
+              // provide an `InsertableRelation`, so we can't use `DataFrameWriter.insertInto` here.
+              existingProvider.newInstance().asInstanceOf[CreatableRelationProvider].createRelation(
+                df.sparkSession.sqlContext, mode, new CaseInsensitiveMap(extraOptions.toMap), df)
```
gatorsmile: In this case, does the order of columns still matter?

cloud-fan: good catch!

gatorsmile:
retest this please

sql/core/src/test/scala/org/apache/spark/sql/test/DataFrameReaderWriterSuite.scala
```diff
@@ -635,4 +638,13 @@ class DataFrameReaderWriterSuite extends QueryTest with SharedSQLContext with Be
       checkAnswer(spark.table("t"), Row(1, "a") :: Row(2, "b") :: Nil)
     }
   }
+
+  test("use saveAsTable to append to a data source table implementing CreatableRelationProvider") {
+    withTable("t") {
+      val provider = "org.apache.spark.sql.test.DefaultSource"
```
yhuai: Is the relation created by this guy an InsertableRelation?

yhuai: I am thinking that we probably should create a test data source that explicitly mentions its relation is not an InsertableRelation.

sql/core/src/main/scala/org/apache/spark/sql/DataFrameWriter.scala
```diff
@@ -364,48 +366,162 @@ final class DataFrameWriter[T] private[sql](ds: Dataset[T]) {
...
+            // As we are inserting into an existing table, we should respect the existing schema and
+            // adjust the column order of the given dataframe according to it, or throw exception
+            // if the column names do not match.
+            val adjustedColumns = tableCols.map { col =>
+              df.queryExecution.analyzed.resolve(Seq(col), resolver).getOrElse {
+                val inputColumns = df.schema.map(_.name).mkString(", ")
+                throw new AnalysisException(
+                  s"cannot resolve '$col' given input columns: [$inputColumns]")
+              }
+            }
```
yhuai: Do we still need the comment at https://github.com/apache/spark/pull/15996/files#diff-945e51801b84b92da242fcb42f83f5f5L195 ?

sql/core/src/main/scala/org/apache/spark/sql/DataFrameWriter.scala
```diff
@@ -364,48 +366,162 @@ final class DataFrameWriter[T] private[sql](ds: Dataset[T]) {
...
+              // For data source that implementing `CreatableRelationProvider`, we should call its
+              // `createRelation` to do the data appending. `CreatableRelationProvider` may not
+              // provide an `InsertableRelation`, so we can't use `DataFrameWriter.insertInto` here.
+              existingProvider.newInstance().asInstanceOf[CreatableRelationProvider].createRelation(
+                sqlContext = df.sparkSession.sqlContext,
+                mode = mode,
+                parameters = new CaseInsensitiveMap(extraOptions.toMap),
+                data = adjustedDataFrame)
+            } else {
+              adjustedDataFrame.write.insertInto(tableIdentWithDB)
```
yhuai: Before this change, we always go to `createRelation`, right?

sql/core/src/main/scala/org/apache/spark/sql/DataFrameWriter.scala
yhuai: Which part of that pr is reverted?

sql/core/src/main/scala/org/apache/spark/sql/execution/command/createDataSourceTables.scala
```diff
@@ -140,153 +140,55 @@ case class CreateDataSourceTableAsSelectCommand(
...
+      val dataSource = DataSource(
+        sparkSession,
+        className = provider,
+        partitionColumns = table.partitionColumnNames,
+        bucketSpec = table.bucketSpec,
+        options = table.storage.properties ++ pathOption,
+        catalogTable = Some(table))
+
+      val result = try {
+        dataSource.write(SaveMode.Overwrite, Dataset.ofRows(sparkSession, query))
```
yhuai: Can you explain why we always use `SaveMode.Overwrite` at here?

sql/hive/src/test/scala/org/apache/spark/sql/hive/PartitionProviderCompatibilitySuite.scala
```diff
@@ -195,12 +195,25 @@ class PartitionProviderCompatibilitySuite
           withTempDir { dir =>
             setupPartitionedDatasourceTable("test", dir)
             if (enabled) {
-              spark.sql("msck repair table test")
```
yhuai: Why we do not need this anymore?

yhuai: oh, we are checking the number of rows before the msck, right?

sql/hive/src/test/scala/org/apache/spark/sql/hive/PartitionProviderCompatibilitySuite.scala
```diff
@@ -195,12 +195,25 @@ class PartitionProviderCompatibilitySuite
...
             }
-            assert(spark.sql("select * from test").count() == 5)
-            spark.range(10).selectExpr("id as fieldOne", "id as partCol")
+
+            spark.range(3, 13).selectExpr("id as fieldOne", "id as partCol")
               .write.partitionBy("partCol").mode("append").saveAsTable("test")
-            assert(spark.sql("select * from test").count() == 15)
+
+            if (enabled) {
+              // Only the newly written partitions are visible, which means the partitions
```
yhuai: Let's also explain why we only see newly written partitions.

sql/hive/src/test/scala/org/apache/spark/sql/hive/PartitionProviderCompatibilitySuite.scala
```diff
@@ -195,12 +195,25 @@ class PartitionProviderCompatibilitySuite
...
             if (enabled) {
-              spark.sql("msck repair table test")
+              assert(spark.table("test").count() == 0)
+            } else {
+              assert(spark.table("test").count() == 5)
             }
-            assert(spark.sql("select * from test").count() == 5)
-            spark.range(10).selectExpr("id as fieldOne", "id as partCol")
+
+            spark.range(3, 13).selectExpr("id as fieldOne", "id as partCol")
```
yhuai: It will be good to also explain the reason that we use (3, 13) in the comment.

sql/core/src/main/scala/org/apache/spark/sql/DataFrameWriter.scala
cloud-fan: no we don't. We will use `ImsertIntoTable` to do the appending.

sql/core/src/main/scala/org/apache/spark/sql/DataFrameWriter.scala
cloud-fan: Currently only 2 data sources accept `saveAsTable` with append mode: `CreatableRelationProvider` and `FileFormat`. For `CreatableRelationProvider`, we always go to `createRelation`, for `FileFormat`, we go to `InsertIntoHadoopFsRelation`, which is same with `InsertIntoTable`. That's why I add the if-else here.

sql/core/src/main/scala/org/apache/spark/sql/DataFrameWriter.scala
cloud-fan: all of it, except the test

sql/core/src/main/scala/org/apache/spark/sql/execution/command/createDataSourceTables.scala
cloud-fan: We are creating a new table, the data dir is empty, ideally we can use whatever mode. Maybe use `ErrorIfExists` is safer?

sql/hive/src/test/scala/org/apache/spark/sql/hive/PartitionProviderCompatibilitySuite.scala
cloud-fan: yep

sql/core/src/test/scala/org/apache/spark/sql/test/DataFrameReaderWriterSuite.scala
cloud-fan: The data source is defined in this file: https://github.com/apache/spark/pull/15996/files#diff-b9ddfbc9be8d83ecf100b3b8ff9610b9R48

I think it's easy to tell that it extends `CreatableRelationProvider` but not return a `InsertableRelation`

sql/hive/src/test/scala/org/apache/spark/sql/hive/PartitionProviderCompatibilitySuite.scala
cloud-fan: to be consistent with the behavior of `InsertItoTable`. I'll add that.

sql/core/src/test/scala/org/apache/spark/sql/test/DataFrameReaderWriterSuite.scala
```diff
@@ -643,6 +644,14 @@ class DataFrameReaderWriterSuite extends QueryTest with SharedSQLContext with Be
     withTable("t") {
       val provider = "org.apache.spark.sql.test.DefaultSource"
       sql(s"CREATE TABLE t USING $provider")
+
+      // make sure the data source doesn't provide `InsertableRelation`, so that we can only append
+      // data to it with `CreatableRelationProvider.createRelation`
```
yhuai: One last comment. Let's explicitly say that we want to test the case that a data source is a CreatableRelationProvider but its relation does not implement InsertableRelation.

yhuai:
LGTM. Can you update the comment to address my last comment (https://github.com/apache/spark/pull/15996#discussion_r93730700)?

yhuai:
ah https://github.com/apache/spark/commit/9a1ad71db44558bb6eb380dc23a1a1abbc2f3e98 failed.

sql/core/src/main/scala/org/apache/spark/sql/execution/command/createDataSourceTables.scala
```diff
@@ -289,4 +180,29 @@ case class CreateDataSourceTableAsSelectCommand(
...
     Seq.empty[Row]
   }
+
+  private def saveDataIntoTable(
+      session: SparkSession,
+      table: CatalogTable,
+      tableLocation: Option[String],
+      data: LogicalPlan,
+      mode: SaveMode): BaseRelation = {
+    // Create the relation based on the data of df.
```
gatorsmile: Nit: Need an update in the comment.

sql/core/src/main/scala/org/apache/spark/sql/DataFrameWriter.scala
```diff
@@ -373,39 +378,50 @@ final class DataFrameWriter[T] private[sql](ds: Dataset[T]) {
...
-        val tableType = if (tableExists) {
-          existingTable.get.tableType
-        } else if (storage.locationUri.isDefined) {
-          CatalogTableType.EXTERNAL
-        } else {
-          CatalogTableType.MANAGED
+        EliminateSubqueryAliases(catalog.lookupRelation(tableIdentWithDB)) match {
+          // Only do the check if the table is a data source table (the relation is a BaseRelation).
+          case LogicalRelation(dest: BaseRelation, _, _) =>
+            if (srcRelations.contains(dest)) {
```
gatorsmile: Nit:
```Scala
          case LogicalRelation(dest: BaseRelation, _, _) if (srcRelations.contains(dest)) =>
            throw new AnalysisException(...
```

sql/core/src/main/scala/org/apache/spark/sql/execution/command/createDataSourceTables.scala
```diff
@@ -134,146 +133,38 @@ case class CreateDataSourceTableAsSelectCommand(
...
-            s"insert data into the table or set SaveMode to SaveMode.Overwrite to overwrite" +
-            s"the existing data. " +
-            s"Or, if you are using SQL CREATE TABLE, you need to drop $tableName first.")
-        case SaveMode.Ignore =>
-          // Since the table already exists and the save mode is Ignore, we will just return.
-          return Seq.empty[Row]
-        case SaveMode.Append =>
-          val existingTable = sessionState.catalog.getTableMetadata(tableIdentWithDB)
+    val result = if (sessionState.catalog.tableExists(tableIdentWithDB)) {
+      assert(mode != SaveMode.Overwrite, "analyzer will drop the table to overwrite it.")
```
gatorsmile: How about `s"Expect the table $tableName has been dropped when the save mode is Overwrite"`?

gatorsmile:
LGTM except three minor comments

gatorsmile:
LGTM

yhuai:
LGTM. Merging to master.

PR #15998, opened 2016-11-23
[SPARK-18572][SQL] Add a method `listPartitionNames` to `ExternalCatalog`

mallman:
(Link to Jira issue: https://issues.apache.org/jira/browse/SPARK-18572)

Currently Spark answers the `SHOW PARTITIONS` command by fetching all of the table's partition metadata from the external catalog and constructing partition names therefrom. The Hive client has a `getPartitionNames` method which is many times faster for this purpose, with the performance improvement scaling with the number of partitions in a table.

To test the performance impact of this PR, I ran the `SHOW PARTITIONS` command on two Hive tables with large numbers of partitions. One table has ~17,800 partitions, and the other has ~95,000 partitions. For the purposes of this PR, I'll call the former table `table1` and the latter table `table2`. I ran 5 trials for each table with before-and-after versions of this PR. The results are as follows:

Spark at bdc8153, `SHOW PARTITIONS table1`, times in seconds:
7.901
3.983
4.018
4.331
4.261

Spark at bdc8153, `SHOW PARTITIONS table2`
(Timed out after 10 minutes with a `SocketTimeoutException`.)

Spark at this PR, `SHOW PARTITIONS table1`, times in seconds:
3.801
0.449
0.395
0.348
0.336

Spark at this PR, `SHOW PARTITIONS table2`, times in seconds:
5.184
1.63
1.474
1.519
1.41

Taking the best times from each trial, we get a 12x performance improvement for a table with ~17,800 partitions and at least a 426x improvement for a table with ~95,000 partitions. More significantly, the latter command doesn't even complete with the current code in master.

This is actually a patch we've been using in-house at VideoAmp since Spark 1.1. It's made all the difference in the practical usability of our largest tables. Even with tables with about 1,000 partitions there's a performance improvement of about 2-3x.

## How was this patch tested?

I added a unit test to `VersionsSuite` which tests that the Hive client's `getPartitionNames` method returns the correct number of partitions.

### DIFF
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/ExternalCatalog.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/ExternalCatalog.scala
@@ -190,6 +190,38 @@ abstract class ExternalCatalog {
       spec: TablePartitionSpec): Option[CatalogTablePartition]
 
   /**
+   * List the names of all partitions that belong to the specified table, assuming it exists.
+   *
+   * A partial partition spec may optionally be provided to filter the partitions returned.
+   * For instance, if there exist partitions (a='1', b='2'), (a='1', b='3') and (a='2', b='4'),
+   * then a partial spec of (a='1') will return the first two only.
+   *
+   * We provide a default implementation here which simply delegates to the `listPartitions`
+   * method. For efficiency's sake, overriding this method is recommended for external catalogs
+   * that can list partition names directly.
+   * @param db database name
+   * @param table table name
+   * @param partialSpec  partition spec
+   */
+  def listPartitionNames(
+      db: String,
+      table: String,
+      partialSpec: Option[TablePartitionSpec] = None): Seq[String] = {
+    def getPartName(spec: TablePartitionSpec, partColNames: Seq[String]): String = {
+      partColNames.map { name =>
+        ExternalCatalogUtils.escapePathName(name) + "=" +
+          ExternalCatalogUtils.escapePathName(spec(name))
+      }.mkString("/")
+    }
+
+    val catalogTable = getTable(db, table)
+
+    listPartitions(db, table, partialSpec).map { p =>
+      getPartName(p.spec, catalogTable.partitionColumnNames)
+    }
+  }
+
+  /**
    * List the metadata of all partitions that belong to the specified table, assuming it exists.
    *
    * A partial partition spec may optionally be provided to filter the partitions returned.
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/ExternalCatalog.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/ExternalCatalog.scala
@@ -192,9 +192,12 @@ abstract class ExternalCatalog {
   /**
    * List the names of all partitions that belong to the specified table, assuming it exists.
    *
-   * A partial partition spec may optionally be provided to filter the partitions returned.
-   * For instance, if there exist partitions (a='1', b='2'), (a='1', b='3') and (a='2', b='4'),
-   * then a partial spec of (a='1') will return the first two only.
+   * For a table with partition columns p1, p2, p3, each partition name is formatted as
+   * `p1=v1/p2=v2/p3=v3`. Each partition column name and value is an escaped path name, and can be
+   * decoded with the `ExternalCatalogUtils.unescapePathName` method.
+   *
+   * A partial partition spec may optionally be provided to filter the partitions returned, as
+   * described in the `listPartitions` method.
    *
    * @param db database name
    * @param table table name
@@ -227,7 +230,7 @@ abstract class ExternalCatalog {
    *
    * @param db database name
    * @param table table name
-   * @param predicates  partition-pruning predicates
+   * @param predicates partition-pruning predicates
    */
   def listPartitionsByFilter(
       db: String,
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/ExternalCatalog.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/ExternalCatalog.scala
@@ -198,7 +198,7 @@ abstract class ExternalCatalog {
    *
    * @param db database name
    * @param table table name
-   * @param partialSpec  partition spec
+   * @param partialSpec partition spec
    */
   def listPartitionNames(
       db: String,
@@ -214,7 +214,7 @@ abstract class ExternalCatalog {
    *
    * @param db database name
    * @param table table name
-   * @param partialSpec  partition spec
+   * @param partialSpec partition spec
    */
   def listPartitions(
       db: String,
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/ExternalCatalog.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/ExternalCatalog.scala
@@ -195,10 +195,6 @@ abstract class ExternalCatalog {
    * A partial partition spec may optionally be provided to filter the partitions returned.
    * For instance, if there exist partitions (a='1', b='2'), (a='1', b='3') and (a='2', b='4'),
    * then a partial spec of (a='1') will return the first two only.
-   *
-   * We provide a default implementation here which simply delegates to the `listPartitions`
-   * method. For efficiency's sake, overriding this method is recommended for external catalogs
-   * that can list partition names directly.
    * @param db database name
    * @param table table name
    * @param partialSpec  partition spec
@@ -206,20 +202,7 @@ abstract class ExternalCatalog {
   def listPartitionNames(
       db: String,
       table: String,
-      partialSpec: Option[TablePartitionSpec] = None): Seq[String] = {
-    def getPartName(spec: TablePartitionSpec, partColNames: Seq[String]): String = {
-      partColNames.map { name =>
-        ExternalCatalogUtils.escapePathName(name) + "=" +
-          ExternalCatalogUtils.escapePathName(spec(name))
-      }.mkString("/")
-    }
-
-    val catalogTable = getTable(db, table)
-
-    listPartitions(db, table, partialSpec).map { p =>
-      getPartName(p.spec, catalogTable.partitionColumnNames)
-    }
-  }
+      partialSpec: Option[TablePartitionSpec] = None): Seq[String]
 
   /**
    * List the metadata of all partitions that belong to the specified table, assuming it exists.
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/ExternalCatalog.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/ExternalCatalog.scala
@@ -196,6 +196,8 @@ abstract class ExternalCatalog {
    * `p1=v1/p2=v2/p3=v3`. Each partition column name and value is an escaped path name, and can be
    * decoded with the `ExternalCatalogUtils.unescapePathName` method.
    *
+   * The returned sequence is sorted as strings.
+   *
    * A partial partition spec may optionally be provided to filter the partitions returned, as
    * described in the `listPartitions` method.
    *
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/ExternalCatalog.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/ExternalCatalog.scala
@@ -195,6 +195,7 @@ abstract class ExternalCatalog {
    * A partial partition spec may optionally be provided to filter the partitions returned.
    * For instance, if there exist partitions (a='1', b='2'), (a='1', b='3') and (a='2', b='4'),
    * then a partial spec of (a='1') will return the first two only.
+   *
    * @param db database name
    * @param table table name
    * @param partialSpec  partition spec
@@ -210,6 +211,7 @@ abstract class ExternalCatalog {
    * A partial partition spec may optionally be provided to filter the partitions returned.
    * For instance, if there exist partitions (a='1', b='2'), (a='1', b='3') and (a='2', b='4'),
    * then a partial spec of (a='1') will return the first two only.
+   *
    * @param db database name
    * @param table table name
    * @param partialSpec  partition spec
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/InMemoryCatalog.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/InMemoryCatalog.scala
@@ -28,6 +28,7 @@ import org.apache.spark.{SparkConf, SparkException}
 import org.apache.spark.sql.AnalysisException
 import org.apache.spark.sql.catalyst.{FunctionIdentifier, TableIdentifier}
 import org.apache.spark.sql.catalyst.analysis._
+import org.apache.spark.sql.catalyst.catalog.ExternalCatalogUtils.escapePathName
 import org.apache.spark.sql.catalyst.expressions.Expression
 import org.apache.spark.sql.catalyst.util.StringUtils
 
@@ -496,7 +497,7 @@ class InMemoryCatalog(
 
     listPartitions(db, table, partialSpec).map { partition =>
       partitionColumnNames.map { name =>
-        name + "=" + partition.spec(name)
+        escapePathName(name) + "=" + escapePathName(partition.spec(name))
       }.mkString("/")
     }
   }
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/InMemoryCatalog.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/InMemoryCatalog.scala
@@ -488,6 +488,19 @@ class InMemoryCatalog(
     }
   }
 
+  override def listPartitionNames(
+      db: String,
+      table: String,
+      partialSpec: Option[TablePartitionSpec] = None): Seq[String] = synchronized {
+    val partitionColumnNames = getTable(db, table).partitionColumnNames
+
+    listPartitions(db, table, partialSpec).map { partition =>
+      partitionColumnNames.map { name =>
+        name + "=" + partition.spec(name)
+      }.mkString("/")
+    }
+  }
+
   override def listPartitions(
       db: String,
       table: String,
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/InMemoryCatalog.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/InMemoryCatalog.scala
@@ -499,7 +499,7 @@ class InMemoryCatalog(
       partitionColumnNames.map { name =>
         escapePathName(name) + "=" + escapePathName(partition.spec(name))
       }.mkString("/")
-    }
+    }.sorted
   }
 
   override def listPartitions(
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/SessionCatalog.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/SessionCatalog.scala
@@ -749,6 +749,23 @@ class SessionCatalog(
   }
 
   /**
+   * List the names of all partitions that belong to the specified table, assuming it exists.
+   *
+   * A partial partition spec may optionally be provided to filter the partitions returned.
+   * For instance, if there exist partitions (a='1', b='2'), (a='1', b='3') and (a='2', b='4'),
+   * then a partial spec of (a='1') will return the first two only.
+   */
+  def listPartitionNames(
+      tableName: TableIdentifier,
+      partialSpec: Option[TablePartitionSpec] = None): Seq[String] = {
+    val db = formatDatabaseName(tableName.database.getOrElse(getCurrentDatabase))
+    val table = formatTableName(tableName.table)
+    requireDbExists(db)
+    requireTableExists(TableIdentifier(table, Option(db)))
+    externalCatalog.listPartitionNames(db, table, partialSpec)
+  }
+
+  /**
    * List the metadata of all partitions that belong to the specified table, assuming it exists.
    *
    * A partial partition spec may optionally be provided to filter the partitions returned.
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/SessionCatalog.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/SessionCatalog.scala
@@ -762,6 +762,9 @@ class SessionCatalog(
     val table = formatTableName(tableName.table)
     requireDbExists(db)
     requireTableExists(TableIdentifier(table, Option(db)))
+    partialSpec.foreach { spec =>
+      requirePartialMatchedPartitionSpec(Seq(spec), getTableMetadata(tableName))
+    }
     externalCatalog.listPartitionNames(db, table, partialSpec)
   }
 
@@ -779,6 +782,9 @@ class SessionCatalog(
     val table = formatTableName(tableName.table)
     requireDbExists(db)
     requireTableExists(TableIdentifier(table, Option(db)))
+    partialSpec.foreach { spec =>
+      requirePartialMatchedPartitionSpec(Seq(spec), getTableMetadata(tableName))
+    }
     externalCatalog.listPartitions(db, table, partialSpec)
   }
 
--- a/sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/catalog/ExternalCatalogSuite.scala
+++ b/sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/catalog/ExternalCatalogSuite.scala
@@ -346,6 +346,31 @@ abstract class ExternalCatalogSuite extends SparkFunSuite with BeforeAndAfterEac
     assert(new Path(partitionLocation) == defaultPartitionLocation)
   }
 
+  test("list partition names") {
+    val catalog = newBasicCatalog()
+    val newPart = CatalogTablePartition(Map("a" -> "1", "b" -> "%="), storageFormat)
+    catalog.createPartitions("db2", "tbl2", Seq(newPart), ignoreIfExists = false)
+
+    val partitionNames = catalog.listPartitionNames("db2", "tbl2")
+    assert(partitionNames == Seq("a=1/b=%25%3D", "a=1/b=2", "a=3/b=4"))
+  }
+
+  test("list partition names with partial partition spec") {
+    val catalog = newBasicCatalog()
+    val newPart = CatalogTablePartition(Map("a" -> "1", "b" -> "%="), storageFormat)
+    catalog.createPartitions("db2", "tbl2", Seq(newPart), ignoreIfExists = false)
+
+    val partitionNames1 = catalog.listPartitionNames("db2", "tbl2", Some(Map("a" -> "1")))
+    assert(partitionNames1 == Seq("a=1/b=%25%3D", "a=1/b=2"))
+
+    // Partial partition specs including "weird" partition values should use the unescaped values
+    val partitionNames2 = catalog.listPartitionNames("db2", "tbl2", Some(Map("b" -> "%=")))
+    assert(partitionNames2 == Seq("a=1/b=%25%3D"))
+
+    val partitionNames3 = catalog.listPartitionNames("db2", "tbl2", Some(Map("b" -> "%25%3D")))
+    assert(partitionNames3.isEmpty)
+  }
+
   test("list partitions with partial partition spec") {
     val catalog = newBasicCatalog()
     val parts = catalog.listPartitions("db2", "tbl2", Some(Map("a" -> "1")))
--- a/sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveExternalCatalog.scala
+++ b/sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveExternalCatalog.scala
@@ -814,9 +814,21 @@ private[spark] class HiveExternalCatalog(conf: SparkConf, hadoopConf: Configurat
     spec.map { case (k, v) => k.toLowerCase -> v }
   }
 
+  // Build a map from lower-cased partition column names to exact column names for a given table
+  private def buildLowerCasePartColNameMap(table: CatalogTable): Map[String, String] = {
+    val actualPartColNames = table.partitionColumnNames
+    actualPartColNames.map(colName => (colName.toLowerCase, colName)).toMap
+  }
+
   // Hive metastore is not case preserving and the column names of the partition specification we
   // get from the metastore are always lower cased. We should restore them w.r.t. the actual table
   // partition columns.
+  private def restorePartitionSpec(
+      spec: TablePartitionSpec,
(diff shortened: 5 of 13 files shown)


sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/ExternalCatalog.scala
```diff
@@ -189,6 +189,38 @@ abstract class ExternalCatalog {
...
+   * method. For efficiency's sake, overriding this method is recommended for external catalogs
+   * that can list partition names directly.
+   * @param db database name
+   * @param table table name
+   * @param partialSpec  partition spec
+   */
+  def listPartitionNames(
+      db: String,
+      table: String,
+      partialSpec: Option[TablePartitionSpec] = None): Seq[String] = {
```
rxin: do we need the default implementation? maybe even just remove this since we only have two implementations for that.

mallman: I could move this implementation to `InMemoryCatalog`, but it feels more natural to me to keep it in `ExternalCatalog` because it has no implementation-specific code.

Alternatively, I could refactor `listPartitions` in `InMemoryCatalog` to accommodate a simple implementation of `listPartitionNames` there. The runtime behavior would be essentially the same.

rxin: My worry is that when we create a new implementation of the external catalog, we'd forget to implement this, leading to bad performance.

mallman:
CC @ericl @cloud-fan

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/InMemoryCatalog.scala
```diff
@@ -482,6 +482,19 @@ class InMemoryCatalog(
...
 
+  override def listPartitionNames(
+      db: String,
+      table: String,
+      partialSpec: Option[TablePartitionSpec] = None): Seq[String] = synchronized {
+    val partitionColumnNames = getTable(db, table).partitionColumnNames
+
+    listPartitions(db, table, partialSpec).map { partition =>
+      partitionColumnNames.map { name =>
+        name + "=" + partition.spec(name)
```
ericl: Does this need escaping, as provided by escapePathName?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/ExternalCatalog.scala
```diff
@@ -189,6 +189,21 @@ abstract class ExternalCatalog {
       spec: TablePartitionSpec): Option[CatalogTablePartition]
 
   /**
+   * List the names of all partitions that belong to the specified table, assuming it exists.
+   *
+   * A partial partition spec may optionally be provided to filter the partitions returned.
+   * For instance, if there exist partitions (a='1', b='2'), (a='1', b='3') and (a='2', b='4'),
+   * then a partial spec of (a='1') will return the first two only.
```
ericl: nit: newline here

sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveExternalCatalog.scala
```diff
@@ -922,6 +923,29 @@ private[spark] class HiveExternalCatalog(conf: SparkConf, hadoopConf: Configurat
...
+      db: String,
+      table: String,
+      partialSpec: Option[TablePartitionSpec] = None): Seq[String] = withClient {
+    val actualPartColNames = getTable(db, table).partitionColumnNames
+    val clientPartitionNames =
+      client.getPartitionNames(db, table, partialSpec.map(lowerCasePartitionSpec))
+
+    if (actualPartColNames.exists(partColName => partColName != partColName.toLowerCase)) {
+      clientPartitionNames.map { partName =>
+        val partSpec = PartitioningUtils.parsePathFragmentAsSeq(partName)
```
ericl: Is the (un)escaping here correct? It would be nice to have a unit test to verify these edge cases.

sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveExternalCatalog.scala
```diff
@@ -922,6 +923,29 @@ private[spark] class HiveExternalCatalog(conf: SparkConf, hadoopConf: Configurat
...
+        partSpec.map { case (partName, partValue) =>
+          actualPartColNames.find(_.equalsIgnoreCase(partName)).get + "=" + partValue
+        }.mkString("/")
+      }
+    } else {
+      clientPartitionNames
```
ericl: Consider not having this optimization to avoid two different code paths here.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/InMemoryCatalog.scala
mallman: I think there should be column name and value escaping as these are column names and values. What's confusing to me is that the previous code uses a method called `escapePathName` which suggests something related to a filesystem path. Indeed the source code includes a comment stating that the aforementioned method comes from a Hive class called `FileUtils`.

I'll look into how Hive escapes partition names and values in its `getPartitionNames` method and report back.

sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveExternalCatalog.scala
mallman: I'll try a build with and without this code branch and see if there's a noticeable performance difference.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/InMemoryCatalog.scala
mallman: I looked into how Hive escapes partition names. It does not appear to do any escaping at all. The partition names are in a string column in a table in the metastore database. Hive simply queries that partition name column on this table, filtering by Hive database and table, sorts the results and returns them.

sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveExternalCatalog.scala
mallman: I ran `show partitions` on a table with about 99k partitions using a build with and without this short circuit. I didn't keep count, but I must have run the command at least 10 times in each session. The best time I got for listing with the short circuit was 1.496 seconds. The best time I got for listing without the short circuit was 1.671.

I can't say I feel strongly one way or the other. I guess my vote is to keep the short circuit.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/ExternalCatalog.scala
mallman: Newline added here and in comment for `listPartitions`.

sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveExternalCatalog.scala
mallman: I added coverage for partition column names with uppercase characters to `HiveCommandSuite`.

mallman: Sorry, I just realized that isn't exactly what you were asking about.

I think adding coverage for mixed case partition column names is a good thing, but I'll address path (un)escaping as well.

mallman: I don't think this is correct, and based on what I've found in the Hive source code (and reading the DDL grammar on the Hive wiki) I'm going to push a change with another proposal for partition column name/value handling.

mallman: I've run some tests to compare behavior between Hive and Spark in handling gnarly partition column names, and I found some disparities. We've spent a considerable amount of time wrangling with partition column name handling recently, and I'm not sure what semantics we've decided on. To ensure the behavior I'm seeing is what we're expecting, I want to describe a scenario I ran.

In my test scenario, I created a table named `test` with the stock Hive 2.1.0 distribution. (I simply downloaded it from its download page and initialized an empty Derby schema store.) The exact DDL I used to create this table is as follows:

```create table test(a string) partitioned by (`P``Дr t` int);```

When I do a `describe test` with `hive` it shows the column name as ``p`дr t``. It appears to lowercase the P and the cyrillic Д before storing the table schema it in the metastore. I then run

```alter table test add partition(`P``Дr t`=0);```

When I run `show partitions test` in `hive` it gives me ``p`дr t=0``. Additionally, when I list the contents of the `test` table's base directory in HDFS, the partition directory entry is

```/user/hive/warehouse/test/p`дr t=0```

If I drop the table, create it with `spark-sql` using the same DDL as before and do a `describe test`, the partition column is given as ``P`Дr t``. Spark has preserved the case of the partition column name. If I then do

```alter table test add partition(`P``Дr t`=0);```

in `spark-sql` and `show partitions test` I get ``P`Дr t=0``. When I list the directory contents in HDFS, I get

```/user/hive/warehouse/test/P`Дr t=0```

The upshot is Hive is lowercasing the partition column name and Spark is leaving it unaltered. Is this correct?

File: sql/hive/src/main/scala/org/apache/spark/sql/hive/client/HiveClientImpl.scala [hunk 6]
```diff
@@ -518,6 +518,20 @@ private[hive] class HiveClientImpl(
...
 
+  /**
+   * Returns the partition names for the given table that match the supplied partition spec.
+   * If no partition spec is specified, all partitions are returned.
+   */
+  override def getPartitionNames(
+      table: CatalogTable,
+      partialSpec: Option[TablePartitionSpec] = None): Seq[String] = withHiveState {
+    partialSpec match {
+      case None => client.getPartitionNames(table.database, table.identifier.table, -1).asScala
```
gatorsmile: Could you add a comment to explain what is `-1`? 

After reading `Hive`, that means the maximum number of partition names to return.

sql/hive/src/test/scala/org/apache/spark/sql/hive/execution/HiveCommandSuite.scala
```diff
@@ -53,25 +53,28 @@ class HiveCommandSuite extends QueryTest with SQLTestUtils with TestHiveSingleto
...
     sql("CREATE TABLE parquet_tab3(col1 int, `col 2` int)")
-    sql("CREATE TABLE parquet_tab4 (price int, qty int) partitioned by (year int, month int)")
-    sql("INSERT INTO parquet_tab4 PARTITION(year = 2015, month = 1) SELECT 1, 1")
-    sql("INSERT INTO parquet_tab4 PARTITION(year = 2015, month = 2) SELECT 2, 2")
-    sql("INSERT INTO parquet_tab4 PARTITION(year = 2016, month = 2) SELECT 3, 3")
-    sql("INSERT INTO parquet_tab4 PARTITION(year = 2016, month = 3) SELECT 3, 3")
+
+    // NB: some table partition column names in this test suite have upper-case characters to test
+    // column name case preservation. Do not lowercase these partition names without good reason.
+    sql("CREATE TABLE parquet_tab4 (price int, qty int) partitioned by (Year int, Month int)")
```
gatorsmile: The code changes are not related to this PR. How about submitting a separate PR for this purpose? These changes only cover the Hive serde table. We should create dedicated test cases, if we do not have such test cases.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/SessionCatalog.scala
```diff
@@ -730,6 +730,23 @@ class SessionCatalog(
...
+   * For instance, if there exist partitions (a='1', b='2'), (a='1', b='3') and (a='2', b='4'),
+   * then a partial spec of (a='1') will return the first two only.
+   */
+  def listPartitionNames(
+      tableName: TableIdentifier,
+      partialSpec: Option[TablePartitionSpec] = None): Seq[String] = {
+    val db = formatDatabaseName(tableName.database.getOrElse(getCurrentDatabase))
+    val table = formatTableName(tableName.table)
+    requireDbExists(db)
+    requireTableExists(TableIdentifier(table, Option(db)))
```
gatorsmile: Do we need to call `requirePartialMatchedPartitionSpec` here?

cloud-fan:
where is the speed-up come from? Is it because the hive API  `getPartitionNames` is faster than `getPartitions`? Or is it because we generate the partition string(`a=1/b=2/c=3`) at hive side and it's faster than Spark?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/SessionCatalog.scala
mallman: I don't know. The only method that calls that method is `dropPartitions`.

mallman:
> where is the speed-up come from? Is it because the hive API getPartitionNames is faster than getPartitions? Or is it because we generate the partition string(a=1/b=2/c=3) at hive side and it's faster than Spark?

To answer `getPartitionNames`, Hive does a simple `SELECT` on a table in the metastore. It doesn't not transform the results—the partition names are in a column in that table. That's where the speed up comes from.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/SessionCatalog.scala
gatorsmile: Without `requirePartialMatchedPartitionSpec`, do we detect these negative cases?

mallman: Got it. I will add that in my next push.

mallman:
Hi Guys,

Repeating my comment/query for @ericl. I'm hoping someone can provide affirmation/refutation to my question before I proceed with new unit tests.

I've run some tests to compare behavior between Hive and Spark in handling gnarly partition column names, and I found some disparities. We've spent a considerable amount of time wrangling with partition column name handling recently, and I'm not sure what semantics we've decided on. To ensure the behavior I'm seeing is what we're expecting, I want to describe a scenario I ran.

In my test scenario, I created a table named `test` with the stock Hive 2.1.0 distribution. (I simply downloaded it from its download page and initialized an empty Derby schema store.) The exact DDL I used to create this table is as follows:

```create table test(a string) partitioned by (`P``Дr t` int);```

When I do a `describe test` with `hive` it shows the column name as ``p`дr t``. It appears to lowercase the P and the cyrillic Д before storing the table schema it in the metastore. I then run

```alter table test add partition(`P``Дr t`=0);```

When I run `show partitions test` in `hive` it gives me ``p`дr t=0``. Additionally, when I list the contents of the `test` table's base directory in HDFS, the partition directory entry is

```/user/hive/warehouse/test/p`дr t=0```

If I drop the table, create it with `spark-sql` using the same DDL as before and do a `describe test`, the partition column is given as ``P`Дr t``. Spark has preserved the case of the partition column name. If I then do

```alter table test add partition(`P``Дr t`=0);```

in `spark-sql` and `show partitions test` I get ``P`Дr t=0``. When I list the directory contents in HDFS, I get

```/user/hive/warehouse/test/P`Дr t=0```

The upshot is Hive is lowercasing the partition column name and Spark is leaving it unaltered. Is this correct?

**Comment by ericl** [comment]
@mallman I'll take a look today

File: sql/hive/src/test/scala/org/apache/spark/sql/hive/execution/HiveCommandSuite.scala [hunk 9]
```diff
@@ -408,14 +411,18 @@ class HiveCommandSuite extends QueryTest with SQLTestUtils with TestHiveSingleto
...
+        .partitionBy("A")
         .format("parquet")
         .mode(SaveMode.Overwrite)
         .saveAsTable("part_datasrc")
 
-      assert(sql("SHOW PARTITIONS part_datasrc").count() == 3)
+      checkAnswer(
+        sql("SHOW PARTITIONS part_datasrc"),
+        Row("A=1") ::
+          Row("A=2") ::
```
ericl: nit: indentation seems a bit funny here, could inline the rows

ericl (commented):
Overall this looks good. I think there are some issues with escaping in other parts of Spark, but they do not seem related to this PR, so I'll file a separate ticket for those.


 For example, I tried `spark.sqlContext.range(10).selectExpr("id", "id as A", "'A$\\=%' as B").write.partitionBy("A", "B").mode("overwrite").saveAsTable("testy")`, and this fails to write out a correctly formed table.

ericl:
* looks good once InMemoryCatalog is fixed

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/ExternalCatalog.scala
```diff
@@ -189,11 +189,28 @@ abstract class ExternalCatalog {
       spec: TablePartitionSpec): Option[CatalogTablePartition]
 
   /**
+   * List the names of all partitions that belong to the specified table, assuming it exists.
```
mallman: `PartitioningUtils` is in SQL core, not catalyst. I decided to use the `escapePathName` and `unescapePathName` methods directly.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/InMemoryCatalog.scala
```diff
@@ -482,6 +483,19 @@ class InMemoryCatalog(
...
 
+  override def listPartitionNames(
+      db: String,
+      table: String,
+      partialSpec: Option[TablePartitionSpec] = None): Seq[String] = synchronized {
+    val partitionColumnNames = getTable(db, table).partitionColumnNames
+
+    listPartitions(db, table, partialSpec).map { partition =>
+      partitionColumnNames.map { name =>
+        escapePathName(name) + "=" + escapePathName(partition.spec(name))
```
mallman: As mentioned above, `PartitioningUtils` is not available here so I went ahead with `escapePathName` itself.

mallman:
I will work on additional unit test coverage tomorrow.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/ExternalCatalog.scala
```diff
@@ -189,14 +189,34 @@ abstract class ExternalCatalog {
...
+   *
+   * For a table with partition columns p1, p2, p3, each partition name is formatted as
+   * `p1=v1/p2=v2/p3=v3`. Each partition column name and value is an escaped path name, and can be
+   * decoded with the `ExternalCatalogUtils.unescapePathName` method.
+   *
+   * A partial partition spec may optionally be provided to filter the partitions returned, as
```
cloud-fan: can we just copy the document for the partial partition spec from `listPartitions` to here?

sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveExternalCatalog.scala
```diff
@@ -922,6 +924,25 @@ private[spark] class HiveExternalCatalog(conf: SparkConf, hadoopConf: Configurat
...
+      db: String,
+      table: String,
+      partialSpec: Option[TablePartitionSpec] = None): Seq[String] = withClient {
+    val actualPartColNames = getTable(db, table).partitionColumnNames
+    val clientPartitionNames =
+      client.getPartitionNames(db, table, partialSpec.map(lowerCasePartitionSpec))
+    clientPartitionNames.map { partName =>
+      val partSpec = PartitioningUtils.parsePathFragmentAsSeq(partName)
+      partSpec.map { case (partName, partValue) =>
+        escapePathName(actualPartColNames.find(_.equalsIgnoreCase(partName)).get) + "=" +
```
cloud-fan: can we just build a map with lower cased partition column name as key, and path-escaped actual partition column name as value? then we can do a simple map lookup here, i.e. `map.get(partName.toLowerCase)`

sql/hive/src/main/scala/org/apache/spark/sql/hive/client/HiveClient.scala
```diff
@@ -155,6 +155,25 @@ private[hive] trait HiveClient {
...
+      table: String,
+      partialSpec: Option[TablePartitionSpec]): Seq[String] = {
+    getPartitionNames(getTable(db, table), partialSpec)
+  }
+
+  /**
+   * Returns the partition names for the given table that match the supplied partition spec.
+   * If no partition spec is specified, all partitions are returned.
+   */
+  def getPartitionNames(
```
cloud-fan: why we create 2 APIs? who will call this method directly?

cloud-fan:
I think this PR is for performance improvement, why we update the `HiveCommandSuite` here? Otherwise LGTM

mallman:
Added a couple of unit tests and rebased.

sql/hive/src/main/scala/org/apache/spark/sql/hive/client/HiveClient.scala
mallman: Good question. I guess I'm just following the example of the two `getPartitions` methods.

mallman:
LMK if there's anything else you'd like me to address, otherwise—assuming the tests pass—please merge to master. Also, it would be great if we can back port this into 2.1 as well.

mallman:
The test that failed is definitely related to this PR, however it passes for me locally. I'll investigate...

gatorsmile:
Like the other partition ExternalCatalog APIs, could you also add the negative test cases to `ExternalCatalogSuite.scala`?

mallman:
> Like the other partition ExternalCatalog APIs, could you also add the negative test cases to ExternalCatalogSuite.scala?

I'm sorry, I don't understand what you're asking for. Can you be more specific?

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/catalog/ExternalCatalogSuite.scala
```diff
@@ -346,6 +346,31 @@ abstract class ExternalCatalogSuite extends SparkFunSuite with BeforeAndAfterEac
     assert(new Path(partitionLocation) == defaultPartitionLocation)
   }
 
+  test("list partition names") {
+    val catalog = newBasicCatalog()
+    val newPart = CatalogTablePartition(Map("a" -> "1", "b" -> "%="), storageFormat)
+    catalog.createPartitions("db2", "tbl2", Seq(newPart), ignoreIfExists = false)
+
+    val partitionNames = catalog.listPartitionNames("db2", "tbl2")
+    assert(partitionNames == Seq("a=1/b=%25%3D", "a=1/b=2", "a=3/b=4"))
```
cloud-fan: is it the same behavior as before? i.e. `SHOW PARTITIONS` will output URI encoded partition value.

sql/hive/src/main/scala/org/apache/spark/sql/hive/client/HiveClientImpl.scala
```diff
@@ -519,6 +519,26 @@ private[hive] class HiveClientImpl(
     client.alterPartitions(table, newParts.map { p => toHivePartition(p, hiveTable) }.asJava)
   }
 
+  /**
+   * Returns the partition names for the given table that match the supplied partition spec.
+   * If no partition spec is specified, all partitions are returned.
+   *
+   * The returned sequence is sorted as strings.
```
cloud-fan: why do we need to sort?

cloud-fan:
can you also address this comment? https://github.com/apache/spark/pull/15998#discussion_r90181774

LGTM except a few comments.

gatorsmile:
What is the expected output? In the ExternalCatalogSuite, it sounds like we do not issue any error. 
```
catalog.listPartitionNames("db2", "tbl2", Some(Map("unknown" -> "unknown")))
```

In the `SessionCatalogSuite`, we also should add the test cases like [`listPartitions`](https://github.com/apache/spark/blob/e64a2047eaf02d65dcf98b6e0710e10196aa74b1/sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/catalog/SessionCatalogSuite.scala#L872-L889) and also the negative cases like what I mentioned above.

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/catalog/ExternalCatalogSuite.scala
mallman: Yes

sql/hive/src/main/scala/org/apache/spark/sql/hive/client/HiveClientImpl.scala
mallman: According to the Hive wiki, these values are supposed to be sorted. However, a test build failed because the output of this method without explicit sorting was not sorted. I don't know why. I'm including ".sorted" to explicitly ensure sorting.

mallman:
> can you also address this comment? #15998 (comment)

Addressed.

mallman:
@gatorsmile I enhanced the coverage of `SessionCatalog.listPartitions` and `SessionCatalog.listPartitionNames` to include tests for invalid partial partition specs.

mallman:
@ericl @gatorsmile Please see test failure here: https://amplab.cs.berkeley.edu/jenkins/job/SparkPullRequestBuilder/69581/testReport/junit/org.apache.spark.sql.hive/PartitionProviderCompatibilitySuite/SPARK_18659_insert_overwrite_table_with_lowercase___partition_management_true/

Are partial partition specs supposed to be case insensitive? Should I just alter the `SessionCatalog.requirePartialMatchedPartitionSpec` method to do a case-insensitive comparison instead of an exact match?

Thanks.

gatorsmile:
This depends on `spark.sql.caseSensitive`, right? cc @ericl @cloud-fan

ericl:
Spark SQL is case insensitive by default. As @gatorsmile mentioned it might be good to check that flag, though I don't know how well case sensitive mode works when enabled.

sql/hive/src/main/scala/org/apache/spark/sql/hive/client/HiveClientImpl.scala
cloud-fan: but according to how we use this API in Spark SQL, we don't need to sort right? maybe we can just do the sorting in test.

mallman:
I've modified the behavior of the partition spec checking methods in `SessionCatalog` to test for case-sensitive analysis.

cloud-fan:
@mallman do you know which tests fail the partition spec checking? It looks to me that before we call partition related API in `SessionCatalog`, the partition column names should be normalized already.

mallman:
> @mallman do you know which tests fail the partition spec checking? It looks to me that before we call partition related API in SessionCatalog, the partition column names should be normalized already.

https://amplab.cs.berkeley.edu/jenkins/job/SparkPullRequestBuilder/69581/testReport/junit/org.apache.spark.sql.hive/PartitionProviderCompatibilitySuite/SPARK_18659_insert_overwrite_table_with_lowercase___partition_management_true/

cloud-fan:
@mallman ah it's a mistake, check https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSourceStrategy.scala#L228 , we need to move this to place before https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSourceStrategy.scala#L198, so that the static partitions are normalized.

mallman:
@cloud-fan That's unfortunate if it's going to block this PR. How do we proceed?

cloud-fan:
I think we can just fix it in this PR, it's only several lines change.

mallman:
@cloud-fan I'm not familiar enough with that code to be comfortable making that change. Can you submit a PR against `VideoAmp:spark-18572-list_partition_names` with the necessary changes to this file?

gatorsmile:
https://github.com/apache/spark/pull/15998#issuecomment-264761754 found a bug. If this PR will not be merged to Spark 2.1 branch, I think we need to submit a separate PR for resolving the bug.

gatorsmile:
@mallman ,  https://github.com/apache/spark/commit/f4c48e1d61dc4d1fc581b78dd9e5c948965acad5 might be what @cloud-fan wants.

mallman:
@gatorsmile I can't find your commit:

```
[msa@ip-10-0-8-34 spark-master]$ git fetch origin
remote: Counting objects: 114, done.
remote: Compressing objects: 100% (53/53), done.
remote: Total 114 (delta 41), reused 22 (delta 22), pack-reused 15
Receiving objects: 100% (114/114), 51.13 KiB | 0 bytes/s, done.
Resolving deltas: 100% (41/41), completed with 22 local objects.
From github.com:apache/spark
   1821cbe..e23c8cf  branch-2.1 -> origin/branch-2.1
   eb8dd68..01a7d33  master     -> origin/master
[msa@ip-10-0-8-34 spark-master]$ git show f4c48e1d61dc4d1fc581b78dd9e5c948965acad5
fatal: bad object f4c48e1d61dc4d1fc581b78dd9e5c948965acad5
```

mallman:
> #15998 (comment) found a bug. If this PR will not be merged to Spark 2.1 branch, I think we need to submit a separate PR for resolving the bug.

I would like to get this patch into Spark 2.1 as it's a scalability issue for partitioned tables.

mallman:
@gatorsmile I've applied your patch and reverted the change I made in the previous commit to workaround that defect. The failed test now passes for me. Let's see what Jenkins says.

mallman:
I suspect this is a spurious, unrelated test failure. Can we get a rebuild, please?

gatorsmile:
retest this please

cloud-fan:
LGTM, @rxin shall we backport this? Although it's not a bug, it's a scalability issue.

rxin:
Yea this one seems like directly related to the work we did in 2.1 for large table handling.

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/catalog/ExternalCatalogSuite.scala
```diff
@@ -346,6 +346,31 @@ abstract class ExternalCatalogSuite extends SparkFunSuite with BeforeAndAfterEac
...
+  }
+
+  test("list partition names with partial partition spec") {
+    val catalog = newBasicCatalog()
+    val newPart = CatalogTablePartition(Map("a" -> "1", "b" -> "%="), storageFormat)
+    catalog.createPartitions("db2", "tbl2", Seq(newPart), ignoreIfExists = false)
+
+    val partitionNames1 = catalog.listPartitionNames("db2", "tbl2", Some(Map("a" -> "1")))
+    assert(partitionNames1 == Seq("a=1/b=%25%3D", "a=1/b=2"))
```
cloud-fan: cc @gatorsmile @ericl is it same with Hive? It looks weird to me that we return the "weird" value to users.

gatorsmile: Yeah, I tried Hive 1.2. It actually returns the weird value. 
```
hive> create table partTab (col1 int, col2 int) partitioned by (pcol1 String, pcol2 String);
OK
hive> insert into table partTab partition(pcol1='1', pcol2='2') select 3, 4 from dummy;
OK
hive> insert into table partTab partition(pcol1='1', pcol2='%=') select 3, 4 from dummy;
OK
hive> show partitions partTab;
OK
pcol1=1/pcol2=%25%3D
pcol1=1/pcol2=2
hive> show partitions partTab PARTITION(pcol1=1);
OK
pcol1=1/pcol2=2
pcol1=1/pcol2=%25%3D
```

cloud-fan:

PR #16007, opened 2016-11-25
[SPARK-18583][SQL] Fix nullability of InputFileName.

ueshin:

The nullability of `InputFileName` should be `false`.

## How was this patch tested?

Existing tests.

### DIFF
--- a/core/src/main/scala/org/apache/spark/rdd/InputFileNameHolder.scala
+++ b/core/src/main/scala/org/apache/spark/rdd/InputFileNameHolder.scala
@@ -22,6 +22,8 @@ import org.apache.spark.unsafe.types.UTF8String
 /**
  * This holds file names of the current Spark task. This is used in HadoopRDD,
  * FileScanRDD, NewHadoopRDD and InputFileName function in Spark SQL.
+ *
+ * The returned value should never be null but empty string if it is unknown.
  */
 private[spark] object InputFileNameHolder {
   /**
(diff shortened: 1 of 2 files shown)


srowen (approved):
CC @JDrit but yes it seems like it shouldn't be nullable

rxin:
Actually this can be nullable if input file name is not set, isn't it?

ueshin:
@rxin The default value is `""` (`UTF8String.fromString("")`) if input file name is not set for now.
https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/rdd/InputFileNameHolder.scala#L32
Should we change the default value to `null`?

rxin:
I wouldn't change the default as it might break compatibility. That said, I don't think it is safe to just set this to non-nullable because it is a very implicit assumption, and setting it to be nullable is never "wrong". I'd add some comment explaining why it is nullable (e.g. "It depends on the semantics of the caller for InputFileNameHolder, and there is no guarantee that it won't be null")

ueshin:
I see, I'll revert this and add the comment. Thanks.

ueshin:
@rxin And also we should modify the generated code to check the value is null or not, shouldn't we?

rxin:
Alright I looked more into this -- I think your approach might be better actually.

Can you add an require in `InputFileNameHolder.setInputFileName` to verify the input is not null, and then document in InputFileNameHolder to say the returned value should never be null, and empty string if it is unknown?

Then we can change the nullable value to true for this expression.

ueshin:
@rxin Sorry but finally we can change the nullable value to `false`?

rxin:
Yes! That's what I meant -- change it false and add some documentation and one require to force that contract.

ueshin:
I see, thanks!

rxin:
Thanks - LGTM.

rxin:

PR #16008, opened 2016-11-25
[SPARK-18585][SQL] Use `ev.isNull = "false"` if possible for Janino to have a chance to optimize.

ueshin:

Janino can optimize `true ? a : b` into `a` or `false ? a : b` into `b`, or if/else with literal condition, so we should use literal as `ev.isNull` if possible.

## How was this patch tested?

Existing tests.

### DIFF
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/complexTypeCreator.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/complexTypeCreator.scala
@@ -61,7 +61,6 @@ case class CreateArray(children: Seq[Expression]) extends Expression {
     ctx.addMutableState("Object[]", values, s"this.$values = null;")
 
     ev.copy(code = s"""
-      final boolean ${ev.isNull} = false;
       this.$values = new Object[${children.size}];""" +
       ctx.splitExpressions(
         ctx.INPUT_ROW,
@@ -78,7 +77,7 @@ case class CreateArray(children: Seq[Expression]) extends Expression {
       s"""
         final ArrayData ${ev.value} = new $arrayClass($values);
         this.$values = null;
-      """)
+      """, isNull = "false")
   }
 
   override def prettyName: String = "array"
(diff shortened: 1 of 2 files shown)


sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/complexTypeCreator.scala
```diff
@@ -61,7 +61,6 @@ case class CreateArray(children: Seq[Expression]) extends Expression {
     ctx.addMutableState("Object[]", values, s"this.$values = null;")
 
     ev.copy(code = s"""
-      final boolean ${ev.isNull} = false;
```
rxin: can you explain how this change improves the code? I'd think it is no-op but maybe it's not the case?

ueshin: I removed the line because the variable will never be used.
The generated code of the caller of this expression will use literal `false` and Janino might be able to optimize the conditional expressions or if/else statements for null checking.

rxin: Did you look into janino's source code to verify this?

ueshin: Yes, the related codes are:

- conditional expressions
https://github.com/janino-compiler/janino/blob/22a7787e62037049e337cb1ce4064c29a4856022/janino/src/main/java/org/codehaus/janino/UnitCompiler.java#L4247
- if/else statement
https://github.com/janino-compiler/janino/blob/22a7787e62037049e337cb1ce4064c29a4856022/janino/src/main/java/org/codehaus/janino/UnitCompiler.java#L2277

And `UnitCompiler.getConstantValue()` seems to optimize `!true` to `false` or `!false` to `true` [here](https://github.com/janino-compiler/janino/blob/22a7787e62037049e337cb1ce4064c29a4856022/janino/src/main/java/org/codehaus/janino/UnitCompiler.java#L5216), or `true && a` to `a`, `false && a` to `false`, `true && a` to `true`, `false && a` to `a` [here](https://github.com/janino-compiler/janino/blob/22a7787e62037049e337cb1ce4064c29a4856022/janino/src/main/java/org/codehaus/janino/UnitCompiler.java#L5397).

rxin:
Thanks. The change here looks good. How hard would it be if we submit a patch to Janino to improve this case? It will help to cut down bytecode size and speed up before JIT kicks in.

```
boolean isNull = false;
if (isNull) {
  // ...
}
```

ueshin:
Hmm, it is a great idea, but I think it would be very hard to submit such a patch (at least for me now).

rxin:
Thanks - merging in master/branch-2.1. We should look into the Janino change in the future.

PR #16009, opened 2016-11-25
[SPARK-18318][ML] ML, Graph 2.1 QA: API: New Scala APIs, docs

yanboliang:
API review for 2.1, except ```LSH``` related classes which are still under development.

## How was this patch tested?
Only doc changes, no new tests.

### DIFF
--- a/docs/ml-features.md
+++ b/docs/ml-features.md
@@ -1189,8 +1189,8 @@ that the number of buckets used will be smaller than this value, for example, if
 distinct values of the input to create enough distinct quantiles.
 
 NaN values:
-NaN values will be removed from the column when `QuantileDiscretizer` fitting. This will produce
-a `Bucketizer` model for making prediction and transformation. During the transformation, `Bucketizer`
+NaN values will be removed from the column during `QuantileDiscretizer` fitting. This will produce
+a `Bucketizer` model for making predictions. During the transformation, `Bucketizer`
 will raise an error when it finds NaN values in the dataset, but the user can also choose to either
 keep or remove NaN values within the dataset by setting `handleInvalid`. If the user chooses to keep
 NaN values, they will be handled specially and placed into their own bucket, for example, if 4 buckets
--- a/docs/ml-features.md
+++ b/docs/ml-features.md
@@ -1188,7 +1188,9 @@ categorical features. The number of bins is set by the `numBuckets` parameter. I
 that the number of buckets used will be smaller than this value, for example, if there are too few
 distinct values of the input to create enough distinct quantiles.
 
-NaN values: Note also that QuantileDiscretizer
+NaN values:
+NaN values will be removed from the column when `QuantileDiscretizer` fitting. This will produce
+a `Bucketizer` model for making prediction and transformation. During the transformation, `Bucketizer`
 will raise an error when it finds NaN values in the dataset, but the user can also choose to either
 keep or remove NaN values within the dataset by setting `handleInvalid`. If the user chooses to keep
 NaN values, they will be handled specially and placed into their own bucket, for example, if 4 buckets
--- a/mllib/src/main/scala/org/apache/spark/ml/feature/ChiSqSelector.scala
+++ b/mllib/src/main/scala/org/apache/spark/ml/feature/ChiSqSelector.scala
@@ -49,7 +49,6 @@ private[feature] trait ChiSqSelectorParams extends Params
    *
    * @group param
    */
-  @Since("1.6.0")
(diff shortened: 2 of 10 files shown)


mllib/src/main/scala/org/apache/spark/ml/feature/ChiSqSelector.scala
```diff
@@ -49,15 +49,13 @@ private[feature] trait ChiSqSelectorParams extends Params
    *
    * @group param
    */
-  @Since("1.6.0")
```
felixcheung: why are the`@since` removed, btw?

yanboliang: Usually we don't add ```since``` tag to variables and functions in traits, since they may be inherited by new child classes later on and the tag is incorrect for them.

sethah: This happens in several other places though, are we going to remove them all?

mllib/src/main/scala/org/apache/spark/ml/feature/QuantileDiscretizer.scala
```diff
@@ -70,17 +70,16 @@ private[feature] trait QuantileDiscretizerBase extends Params
    * invalid values), error (throw an error), or keep (keep invalid values in a special additional
    * bucket).
    * Default: "error"
+   * TODO: Reuse handleInvalid in HasHandleInvalid.
```
sethah: Is this just a note to add "HasHandleInvalid" as a shared param? I was a bit confused by it at first. Maybe we can reference a jira number for it?

mllib/src/main/scala/org/apache/spark/ml/feature/ChiSqSelector.scala
yanboliang: Yeah, theoretically we should do that, but I'm not very confidence whether this change is appropriate. If we meet an agreement on how to deal with this issue, we can address other places in this PR or follow-up work. cc @jkbradley

mllib/src/main/scala/org/apache/spark/ml/feature/QuantileDiscretizer.scala
yanboliang: Yeah, updated.

mllib/src/main/scala/org/apache/spark/ml/feature/ChiSqSelector.scala
jkbradley: In this case, I think it's OK to have Since annotations in the trait since it is private and should never be used beyond ChiSqSelector and ChiSqSelectorModel.  That seems pretty safe to me.

docs/ml-features.md
```diff
@@ -1188,7 +1188,9 @@ categorical features. The number of bins is set by the `numBuckets` parameter. I
 that the number of buckets used will be smaller than this value, for example, if there are too few
 distinct values of the input to create enough distinct quantiles.
 
-NaN values: Note also that QuantileDiscretizer
+NaN values:
+NaN values will be removed from the column when `QuantileDiscretizer` fitting. This will produce
```
sethah: "from the column when" -> "from the column during"
"model for making prediction and transformation" -> "model for making predictions"

docs/ml-features.md
```diff
@@ -1188,7 +1188,9 @@ categorical features. The number of bins is set by the `numBuckets` parameter. I
 that the number of buckets used will be smaller than this value, for example, if there are too few
 distinct values of the input to create enough distinct quantiles.
 
-NaN values: Note also that QuantileDiscretizer
+NaN values:
+NaN values will be removed from the column during `QuantileDiscretizer` fitting. This will produce
+a `Bucketizer` model for making predictions. During the transformation, `Bucketizer`
```
sethah: Actually, what does this mean exactly? When I train a QuantileDiscretizer to handle NaN values, I get back a Bucketizer that also handles them. The bucketizer does not raise an error when encountering NaNs. That seems contradictory to this statement.

jkbradley: QuantileDiscretizer always drops NaNs during fitting, so it will not throw an error for a dataset with NaNs even if handleInvalid = "error"

jkbradley:
LGTM
Thank you!

docs/ml-features.md
sethah: Thanks for clarifying that!

PR #16011, opened 2016-11-25
[SPARK-18587][ML] Remove handleInvalid from QuantileDiscretizer

yanboliang:
HandleInvalid only happens when ```Bucketizer``` transforming a dataset which contains NaN, however, when the training dataset containing NaN, ```QuantileDiscretizer``` will always ignore them. So we should keep ```handleInvalid``` in ```Bucketizer``` and remove it from ```QuantileDiscretizer```.

For example, before this PR, the default behavior of ```QuantileDiscretizer``` to handle invalid value (i.e. NaN) is throwing error. So we expect errors if running against the following code.
```
val data = Array(-0.9, -0.5, -0.3, 0.0, 0.2, 0.5, 0.9, Double.NaN, Double.NaN, Double.NaN)
val df = data.toSeq.toDF("input")

val discretizer = new QuantileDiscretizer()
      .setInputCol("input")
      .setOutputCol("result")
      .setNumBuckets(3)

val model = discretizer.fit(df)
println(model.getSplits.mkString(","))
```
However, the above code can work well w/o any error thrown. This is because ```QuantileDiscretizer``` which use ```DataFrameStatFunctions.approxQuantile``` will always ignore NaN values when computing quantiles.

## How was this patch tested?
Existing tests.

### DIFF
--- a/docs/ml-features.md
+++ b/docs/ml-features.md
@@ -979,6 +979,12 @@ Note that if you have no idea of the upper and lower bounds of the targeted colu
 
 Note also that the splits that you provided have to be in strictly increasing order, i.e. `s0 < s1 < s2 < ... < sn`.
 
+NaN values: Note also that `Bucketizer`
+will raise an error when it finds NaN values in the dataset by default, but the user can also choose to either
+keep or remove NaN values within the dataset by setting `handleInvalid`. If the user chooses to keep
+NaN values, they will be handled specially and placed into their own bucket, for example, if 4 buckets
+are used, then non-NaN data will be put into buckets[0-3], but NaNs will be counted in a special bucket[4].
+
 More details can be found in the API docs for [Bucketizer](api/scala/index.html#org.apache.spark.ml.feature.Bucketizer).
 
 **Examples**
@@ -1187,12 +1193,6 @@ for more details on the API.
 categorical features. The number of bins is set by the `numBuckets` parameter. It is possible
 that the number of buckets used will be smaller than this value, for example, if there are too few
 distinct values of the input to create enough distinct quantiles.
-
-NaN values: Note also that QuantileDiscretizer
-will raise an error when it finds NaN values in the dataset, but the user can also choose to either
-keep or remove NaN values within the dataset by setting `handleInvalid`. If the user chooses to keep
-NaN values, they will be handled specially and placed into their own bucket, for example, if 4 buckets
-are used, then non-NaN data will be put into buckets[0-3], but NaNs will be counted in a special bucket[4].
 
 Algorithm: The bin ranges are chosen using an approximate algorithm (see the documentation for
 [approxQuantile](api/scala/index.html#org.apache.spark.sql.DataFrameStatFunctions) for a
--- a/mllib/src/main/scala/org/apache/spark/ml/feature/Bucketizer.scala
+++ b/mllib/src/main/scala/org/apache/spark/ml/feature/Bucketizer.scala
@@ -33,6 +33,13 @@ import org.apache.spark.sql.types.{DoubleType, StructField, StructType}
 
 /**
  * `Bucketizer` maps a column of continuous features to a column of feature buckets.
+ *
+ * NaN handling:
+ * `Bucketizer` will raise an error when it finds NaN values in the dataset by default,
+ * but the user can also choose to either keep or remove NaN values within the dataset
+ * by setting `handleInvalid`. If the user chooses to keep NaN values, they will be handled
+ * specially and placed into their own bucket, for example, if 4 buckets are used, then
+ * non-NaN data will be put into buckets[0-3], but NaNs will be counted in a special bucket[4].
  */
 @Since("1.4.0")
 final class Bucketizer @Since("1.4.0") (@Since("1.4.0") override val uid: String)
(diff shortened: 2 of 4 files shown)


yanboliang:
cc @jkbradley @srowen @VinceShieh @sethah

MLnick:
I don't think this is correct. The idea is `QuantileDiscretizer` skips NaN when creating the buckets, but there will still be an error thrown during `transform` (rather than `fit`) if NaNs are in data. Hence the test cases intercepting the error. I think the test cases show it works, no?

e.g.

`model.transform(df).show` throws:

```
...
Caused by: org.apache.spark.SparkException: Bucketizer encountered NaN value. To handle or skip NaNs, try setting Bucketizer.handleInvalid.
  at org.apache.spark.ml.feature.Bucketizer$.binarySearchForBuckets(Bucketizer.scala:188)
  at org.apache.spark.ml.feature.Bucketizer$$anonfun$1.apply$mcDD$sp(Bucketizer.scala:115)
  at org.apache.spark.ml.feature.Bucketizer$$anonfun$1.apply(Bucketizer.scala:114)
  at org.apache.spark.ml.feature.Bucketizer$$anonfun$1.apply(Bucketizer.scala:114)
```

Setting the param works:

```
scala> model.setHandleInvalid("skip").transform(df).show
+-----+------+
|input|result|
+-----+------+
| -0.9|   0.0|
| -0.5|   0.0|
| -0.3|   1.0|
|  0.0|   1.0|
|  0.2|   2.0|
|  0.5|   2.0|
|  0.9|   2.0|
+-----+------+


scala> model.setHandleInvalid("keep").transform(df).show
+-----+------+
|input|result|
+-----+------+
| -0.9|   0.0|
| -0.5|   0.0|
| -0.3|   1.0|
|  0.0|   1.0|
|  0.2|   2.0|
|  0.5|   2.0|
|  0.9|   2.0|
|  NaN|   3.0|
|  NaN|   3.0|
|  NaN|   3.0|
+-----+------+
```

yanboliang:
@MLnick Your description is totally correct. However, the ```model``` you used in your example is type of ```Bucketizer```. I will keep ```handleInvalid``` in ```Bucketizer```. In the current ML code, ```QuantileDiscretizer```(which is an estimator) will produce ```Bucketizer```(which is a model), and the invalid data handling only happens when you use ```Bucketizer``` to transform some data(may be not the same data used for training). When you run ```QuantileDiscretizer```, all NaN values will be ignored and no any error raised. That is the cause why I propose to remove ```handleInvalid``` from ```QuantileDiscretizer```.

MLnick:
Typically the estimator Params are copied to the model though. How do you
propose to set the handle invalid param in say a pipeline?

yanboliang:
@MLnick Yeah, I think this is the most common case that copying Params from estimators to models. However, I also found some algorithms do not comply this rule, such as ```ALS``` which has ```ALSParams``` and ```ALSModelParams``` for estimator and model separately.

I think we can set params to models not via estimator, for example:
```
val discretizer = new QuantileDiscretizer()
val pipeline = new Pipeline().setStages(Array(discretizer))
val model = pipeline.fit(df)
model.stages(0).asInstanceOf[Bucketizer].setHandleInvalid("skip")
```
I know this way is a little tricky, a better way may be we can have ```QuantileDiscretizerModel``` which is produced by ```QuantileDiscretizer```.
Think more about it, ```Bucketizer``` is a separate transformer which mainly has two params(```splits``` and ```handleInvalid```) can be set. Users can provides candidates for these two params when doing cross validation to select the best model. But if we constrict it must be produced by ```QuantileDiscretizer```, the ```splits``` would be a member variable of the model rather than a param. From this perspective, it's more make sense to see ```Bucketizer``` as a separate transformer.

MLnick:
As far as I recall, the idea is that the `Bucketizer` can be used standalone, and because the `QuantileDiscretizer` itself produced the same thing as a bucketizer, it was used as the model rather than having a dedicated `QuantileDiscretizerModel`.

`Bucketizer` is already a separate transformer (it is not required to be produced by a `QuantileDiscretizer`), since it's a `Model` and the constructor is public (by design). So it by itself can be used in a pipeline, and the `splits` param could be selected via cross-validation (for example).

What you propose here makes using `QuantileDiscretizer` and a non-default `handleInvalid` param together with cross-validation impossible. In addition, as you've pointed out in your code example above, this would force a pretty clunky "workaround" to set the `handleInvalid` param in a pipeline.

Why do this? What is the actual problem with what exists currently? To me it seems better the way it is. Also, I don't see any major benefit to adding a new `QuantileDiscretizerModel`.

yanboliang:
When I did QA work for 2.1, I found the parameter ``` handleInvalid``` is disorienting. For example, the default behavior of ```QuantileDiscretizer``` to handle invalid value (i.e. NaN) is throwing error. So users expect errors if running against the following code.
```
val data = Array(-0.9, -0.5, -0.3, 0.0, 0.2, 0.5, 0.9, Double.NaN, Double.NaN, Double.NaN)
val df = data.toSeq.toDF("input")

val discretizer = new QuantileDiscretizer()
      .setInputCol("input")
      .setOutputCol("result")
      .setNumBuckets(3)

val model = discretizer.fit(df)
println(model.getSplits.mkString(","))
```
However, the above code can work well w/o any error thrown. 
I'm OK with not make this change, but we should clarify it more clear in document. Thanks.

jkbradley:
I agree with @MLnick that we should not make this change.  We need users to be able to set all Params in Estimators, including the Params of the Models they produce.  If this is confusing for QuantileDiscretizer, then let's improve the docs.  Thanks!

yanboliang:
That's OK. I will improve the docs in another 2.1 QA PR and close this one. Thanks for all your clarification. @MLnick @jkbradley

PR #16012, opened 2016-11-25
[SPARK-17251][SQL] Support `OuterReference` in projection list of IN correlated subqueries

dongjoon-hyun:

Currently, correlated subqueries do not allow `OuterReference` columns in projection lists of IN correlated subqueries. This PR aims to support that by making `OuterReference` as `NamedExpression` and extending `pullOutCorrelatedPredicates` into `pullOutCorrelatedProjectionAndPredicates`.

```scala
scala> sql("CREATE TEMPORARY VIEW t1 AS SELECT * FROM VALUES 1, 2 AS t1(a)")
scala> sql("CREATE TEMPORARY VIEW t2 AS SELECT * FROM VALUES 1 AS t2(b)")
scala> sql("SELECT a FROM t1 WHERE a IN (SELECT a FROM t2)").show
java.lang.ClassCastException: org.apache.spark.sql.catalyst.expressions.OuterReference cannot be cast to org.apache.spark.sql.catalyst.expressions.NamedExpression
```

## How was this patch tested?

Pass the Jenkins test with new test cases.


sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/namedExpressions.scala
```diff
@@ -356,10 +356,17 @@ case class PrettyAttribute(
...
  * plan. This is used for correlated subqueries.
  */
-case class OuterReference(e: NamedExpression) extends LeafExpression with Unevaluable {
+case class OuterReference(e: NamedExpression)
+  extends LeafExpression with NamedExpression with Unevaluable {
   override def dataType: DataType = e.dataType
   override def nullable: Boolean = e.nullable
   override def prettyName: String = "outer"
+
+  override def name: String = throw new UnsupportedOperationException
```
hvanhovell: Implement as much of these as possible (delegate to the NamedExpression). It is a resolved expression after all.

sql/core/src/test/resources/sql-tests/inputs/correlated.sql
```diff
@@ -0,0 +1,18 @@
+CREATE TEMPORARY VIEW t1 AS SELECT * FROM VALUES 1, 2 AS t1(a);
```
hvanhovell: Naming: lets call this `subqueries.sql`. Correlation has a different meaning in different contexts

dongjoon-hyun:
Thank you, @hvanhovell ! I'll fix like that!

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/namedExpressions.scala
```diff
@@ -356,10 +356,17 @@ case class PrettyAttribute(
  * A place holder used to hold a reference that has been resolved to a field outside of the current
  * plan. This is used for correlated subqueries.
  */
-case class OuterReference(e: NamedExpression) extends LeafExpression with Unevaluable {
+case class OuterReference(e: NamedExpression)(
+  val exprId: ExprId = NamedExpression.newExprId)
```
hvanhovell: Use the `exprId` of the `NamedExpression`.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/namedExpressions.scala
```diff
@@ -356,10 +356,17 @@ case class PrettyAttribute(
...
+  extends LeafExpression with NamedExpression with Unevaluable {
   override def dataType: DataType = e.dataType
   override def nullable: Boolean = e.nullable
   override def prettyName: String = "outer"
+
+  override def name: String = e.name
+  override def qualifier: Option[String] = e.qualifier
+  override def toAttribute: Attribute = e.toAttribute
+  override def newInstance(): NamedExpression = OuterReference(e)()
```
hvanhovell: `OuterReference(e.newInstance())()`?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
```diff
@@ -989,7 +989,7 @@ class Analyzer(
               withPosition(u) {
                 try {
                   outer.resolve(nameParts, resolver) match {
-                    case Some(outerAttr) => OuterReference(outerAttr)
+                    case Some(outerAttr) => OuterReference(outerAttr)()
```
hvanhovell: I am not sure the analyzer change has the desired effect. This just remove the outer reference from the tree, and this won't work if we use the attribute anywhere in the tree. For example:
```sql
select *
from   tbl_a
where id in (select x
             from (select tbl_b.id,
                          tbl_a.id + 1 as x,
                          tbl_a.id + tbl_b.id as y
                   from   tbl_b)
              where y > 0)
```

I think we need to break this down into two steps:
1. Do not support this for now and just fix the named expression. That would be my goal for 2.1.
2. Try to see if we can rewrite the tree in such a way that we can extract the value. That would be my goal for 2.2. I am not sure how well we can make this work. In the end I think we need a dedicated subquery operator.

cc @nsyca what do you think.

hvanhovell (changes requested):

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/namedExpressions.scala
dongjoon-hyun: Sure.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
nsyca: Another interesting case to consider:
````sql
select ...
from   t1
where  t1.c1 in (select sum(t1.c2) from t2)
````

If we support correlated columns in SELECT clause, do we build the Aggregate on T2 or T1?

dongjoon-hyun:
Thank you for review, @hvanhovell and @nsyca .
I agree with you. We need enough time for this.
So, the option one for 2.1 is spun off into #16015.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
hvanhovell: The `LEFT ANTI` join should produce the correct result. Unfortunately we push down the `tbl_a.c1 = tbl_a.c2` expression into the `tbl_a` side of the plan. So we need to fix this. I have created https://issues.apache.org/jira/browse/SPARK-18597 to track this.

dongjoon-hyun: Thank you @hvanhovell .
BTW, may I rebase this PR and try to the second plan for 2.2 here?
> 2. Try to see if we can rewrite the tree in such a way that we can extract the value. That would be my goal for 2.2. I am not sure how well we can make this work. In the end I think we need a dedicated subquery operator.

dongjoon-hyun:
Hi, @hvanhovell and @nsyca .

I created a JIRA issue for Option 2 with simple and complex example (@hvanhovell 's).

* https://issues.apache.org/jira/browse/SPARK-18603

Now, I'm closing this PR. Thank you for reviews, @hvanhovell and @nsyca .

PR #16014, opened 2016-11-25
[SPARK-18590][SPARKR] build R source package when making distribution

felixcheung:

This PR has 2 key changes. One, we are building source package (aka bundle package) for SparkR which could be released on CRAN. Two, we should include in the official Spark binary distributions SparkR installed from this source package instead (which would have help/vignettes rds needed for those to work when the SparkR package is loaded in R, whereas earlier approach with devtools does not)

But, because of various differences in how R performs different tasks, this PR is a fair bit more complicated. More details below.

This PR also includes a few minor fixes.

### more details

These are the additional steps in make-distribution; please see [here](https://github.com/apache/spark/blob/master/R/CRAN_RELEASE.md) on what's going to a CRAN release, which is now run during make-distribution.sh.
1. package needs to be installed because the first code block in vignettes is `library(SparkR)` without lib path
2. `R CMD build` will build vignettes (this process runs Spark/SparkR code and captures outputs into pdf documentation)
3. `R CMD check` on the source package will install package and build vignettes again (this time from source packaged) - this is a key step required to release R package on CRAN
 (will skip tests here but tests will need to pass for CRAN release process to success - ideally, during release signoff we should install from the R source package and run tests)
4. `R CMD Install` on the source package (this is the only way to generate doc/vignettes rds files correctly, not in step # 1)
 (the output of this step is what we package into Spark dist and sparkr.zip)

Alternatively, 
   R CMD build should already be installing the package in a temp directory though it might just be finding this location and set it to lib.loc parameter; another approach is perhaps we could try calling `R CMD INSTALL --build pkg` instead.
 But in any case, despite installing the package multiple times this is relatively fast. 
Building vignettes takes a while though.

## How was this patch tested?

Manually, CI.

### DIFF
--- a/R/check-cran.sh
+++ b/R/check-cran.sh
@@ -34,8 +34,9 @@ if [ ! -z "$R_HOME" ]
     fi
     R_SCRIPT_PATH="$(dirname $(which R))"
 fi
-echo "USING R_HOME = $R_HOME"
+echo "Using R_SCRIPT_PATH = ${R_SCRIPT_PATH}"
 
+# Install the package (this is required for code in vignettes to run when building it later)
 # Build the latest docs, but not vignettes, which is built with the package next
 $FWDIR/create-docs.sh
 
@@ -82,4 +83,20 @@ else
   # This will run tests and/or build vignettes, and require SPARK_HOME
   SPARK_HOME="${SPARK_HOME}" "$R_SCRIPT_PATH/"R CMD check $CRAN_CHECK_OPTIONS SparkR_"$VERSION".tar.gz
 fi
+
+# Install source package to get it to generate vignettes rds files, etc.
+if [ -n "$CLEAN_INSTALL" ]
+then
+  echo "Removing lib path and installing from source package"
+  LIB_DIR="$FWDIR/lib"
+  rm -rf $LIB_DIR
+  mkdir -p $LIB_DIR
+  "$R_SCRIPT_PATH/"R CMD INSTALL SparkR_"$VERSION".tar.gz --library=$LIB_DIR
+
+  # Zip the SparkR package so that it can be distributed to worker nodes on YARN
+  pushd $LIB_DIR > /dev/null
+  jar cfM "$LIB_DIR/sparkr.zip" SparkR
+  popd > /dev/null
+fi
+
 popd > /dev/null
--- a/R/pkg/DESCRIPTION
+++ b/R/pkg/DESCRIPTION
@@ -1,8 +1,8 @@
 Package: SparkR
 Type: Package
-Title: R Frontend for Apache Spark
 Version: 2.1.0
-Date: 2016-11-06
+Title: R Frontend for Apache Spark
+Description: The SparkR package provides an R Frontend for Apache Spark.
 Authors@R: c(person("Shivaram", "Venkataraman", role = c("aut", "cre"),
                     email = "shivaram@cs.berkeley.edu"),
              person("Xiangrui", "Meng", role = "aut",
@@ -10,19 +10,18 @@ Authors@R: c(person("Shivaram", "Venkataraman", role = c("aut", "cre"),
              person("Felix", "Cheung", role = "aut",
                     email = "felixcheung@apache.org"),
              person(family = "The Apache Software Foundation", role = c("aut", "cph")))
+License: Apache License (== 2.0)
 URL: http://www.apache.org/ http://spark.apache.org/
 BugReports: http://spark.apache.org/contributing.html
 Depends:
     R (>= 3.0),
     methods
 Suggests:
+    knitr,
+    rmarkdown,
     testthat,
     e1071,
-    survival,
-    knitr,
-    rmarkdown
-Description: The SparkR package provides an R frontend for Apache Spark.
-License: Apache License (== 2.0)
+    survival
 Collate:
     'schema.R'
     'generics.R'
--- a/R/pkg/NAMESPACE
+++ b/R/pkg/NAMESPACE
@@ -3,7 +3,7 @@
 importFrom("methods", "setGeneric", "setMethod", "setOldClass")
 importFrom("methods", "is", "new", "signature", "show")
 importFrom("stats", "gaussian", "setNames")
-importFrom("utils", "download.file", "object.size", "packageVersion", "untar")
+importFrom("utils", "download.file", "object.size", "packageVersion", "tail", "untar")
 
 # Disable native libraries till we figure out how to package it
 # See SPARKR-7839
--- a/dev/create-release/release-build.sh
+++ b/dev/create-release/release-build.sh
@@ -160,11 +160,6 @@ if [[ "$1" == "package" ]]; then
       ./dev/change-scala-version.sh 2.10
     fi
 
-    # Build R only once
-    if [[ "${NAME}" == "without-hadoop" ]]; then
-      FLAGS="${FLAGS} --r"
-    fi
-
     export ZINC_PORT=$ZINC_PORT
     echo "Creating distribution: $NAME ($FLAGS)"
 
@@ -226,13 +221,14 @@ if [[ "$1" == "package" ]]; then
 
   # We increment the Zinc port each time to avoid OOM's and other craziness if multiple builds
   # share the same Zinc server.
+  # Make R source package only once. (--r)
   FLAGS="-Psparkr -Phive -Phive-thriftserver -Pyarn -Pmesos"
   make_binary_release "hadoop2.3" "-Phadoop-2.3 $FLAGS" "3033" &
   make_binary_release "hadoop2.4" "-Phadoop-2.4 $FLAGS" "3034" &
   make_binary_release "hadoop2.6" "-Phadoop-2.6 $FLAGS" "3035" &
   make_binary_release "hadoop2.7" "-Phadoop-2.7 $FLAGS" "3036" "withpip" &
   make_binary_release "hadoop2.4-without-hive" "-Psparkr -Phadoop-2.4 -Pyarn -Pmesos" "3037" &
-  make_binary_release "without-hadoop" "-Psparkr -Phadoop-provided -Pyarn -Pmesos" "3038" &
+  make_binary_release "without-hadoop" "--r -Psparkr -Phadoop-provided -Pyarn -Pmesos" "3038" &
   wait
   rm -rf spark-$SPARK_VERSION-bin-*/
 
--- a/dev/create-release/release-build.sh
+++ b/dev/create-release/release-build.sh
@@ -198,6 +198,9 @@ if [[ "$1" == "package" ]]; then
         $PYTHON_DIST_NAME.sha
     fi
 
+    echo "Copying R source package"
+    cp spark-$SPARK_VERSION-bin-$NAME/R/SparkR_$SPARK_VERSION.tar.gz .
+
     echo "Copying and signing regular binary distribution"
     cp spark-$SPARK_VERSION-bin-$NAME/spark-$SPARK_VERSION-bin-$NAME.tgz .
     echo $GPG_PASSPHRASE | $GPG --passphrase-fd 0 --armour \
--- a/dev/create-release/release-build.sh
+++ b/dev/create-release/release-build.sh
@@ -150,7 +150,7 @@ if [[ "$1" == "package" ]]; then
     NAME=$1
     FLAGS=$2
     ZINC_PORT=$3
-    BUILD_PIP_PACKAGE=$4
+    BUILD_PACKAGE=$4
     cp -r spark spark-$SPARK_VERSION-bin-$NAME
 
     cd spark-$SPARK_VERSION-bin-$NAME
@@ -172,11 +172,30 @@ if [[ "$1" == "package" ]]; then
     MVN_HOME=`$MVN -version 2>&1 | grep 'Maven home' | awk '{print $NF}'`
 
 
-    if [ -z "$BUILD_PIP_PACKAGE" ]; then
-      echo "Creating distribution without PIP package"
+    if [ -z "$BUILD_PACKAGE" ]; then
+      echo "Creating distribution without PIP/R package"
       ./dev/make-distribution.sh --name $NAME --mvn $MVN_HOME/bin/mvn --tgz $FLAGS \
         -DzincPort=$ZINC_PORT 2>&1 >  ../binary-release-$NAME.log
       cd ..
+    elif [[ "$BUILD_PACKAGE" == "withr" ]]; then
+      echo "Creating distribution with R package"
+      ./dev/make-distribution.sh --name $NAME --mvn $MVN_HOME/bin/mvn --tgz --r $FLAGS \
+        -DzincPort=$ZINC_PORT 2>&1 >  ../binary-release-$NAME.log
+      cd ..
+
+      echo "Copying and signing R source package"
+      R_DIST_NAME=SparkR_$SPARK_VERSION.tar.gz
+      cp spark-$SPARK_VERSION-bin-$NAME/R/$R_DIST_NAME .
+
+      echo $GPG_PASSPHRASE | $GPG --passphrase-fd 0 --armour \
+        --output $R_DIST_NAME.asc \
+        --detach-sig $R_DIST_NAME
+      echo $GPG_PASSPHRASE | $GPG --passphrase-fd 0 --print-md \
+        MD5 $R_DIST_NAME > \
+        $R_DIST_NAME.md5
+      echo $GPG_PASSPHRASE | $GPG --passphrase-fd 0 --print-md \
+        SHA512 $R_DIST_NAME > \
+        $R_DIST_NAME.sha
     else
       echo "Creating distribution with PIP package"
       ./dev/make-distribution.sh --name $NAME --mvn $MVN_HOME/bin/mvn --tgz --pip $FLAGS \
@@ -198,11 +217,6 @@ if [[ "$1" == "package" ]]; then
         $PYTHON_DIST_NAME.sha
     fi
 
-    if [ -f spark-$SPARK_VERSION-bin-$NAME/R/SparkR_$SPARK_VERSION.tar.gz ]; then
-      echo "Copying R source package"
-      cp spark-$SPARK_VERSION-bin-$NAME/R/SparkR_$SPARK_VERSION.tar.gz .
-    fi
-
     echo "Copying and signing regular binary distribution"
     cp spark-$SPARK_VERSION-bin-$NAME/spark-$SPARK_VERSION-bin-$NAME.tgz .
     echo $GPG_PASSPHRASE | $GPG --passphrase-fd 0 --armour \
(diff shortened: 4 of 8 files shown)


R/pkg/DESCRIPTION
```diff
@@ -1,28 +1,27 @@
 Package: SparkR
 Type: Package
-Title: R Frontend for Apache Spark
 Version: 2.1.0
-Date: 2016-11-06
```
felixcheung: this is removed - I tried but haven't found a way to update this automatically, (I guess this could be in the [release-tag](https://github.com/apache/spark/blob/master/dev/create-release/release-tag.sh) script though)
But more importantly, seems like many (most?) packages do not have this in their DESCRIPTION.

In any case, release date are stamped when releasing to CRAN.

R/pkg/NAMESPACE
```diff
@@ -3,7 +3,7 @@
 importFrom("methods", "setGeneric", "setMethod", "setOldClass")
 importFrom("methods", "is", "new", "signature", "show")
 importFrom("stats", "gaussian", "setNames")
-importFrom("utils", "download.file", "object.size", "packageVersion", "untar")
+importFrom("utils", "download.file", "object.size", "packageVersion", "tail", "untar")
```
felixcheung: This was regressed from a recent commit. check-cran.sh actually is flagging this in an existing NOTE but we only check for # of NOTE (which is still 1), and so this went in undetected.

dev/create-release/release-build.sh
```diff
@@ -189,6 +189,9 @@ if [[ "$1" == "package" ]]; then
       SHA512 $PYTHON_DIST_NAME > \
       $PYTHON_DIST_NAME.sha
 
+    echo "Copying R source package"
+    cp spark-$SPARK_VERSION-bin-$NAME/R/SparkR_$SPARK_VERSION.tar.gz .
```
felixcheung: this is the source package we should release to CRAN

felixcheung:
@shivaram

R/check-cran.sh
```diff
@@ -82,4 +83,20 @@ else
   # This will run tests and/or build vignettes, and require SPARK_HOME
   SPARK_HOME="${SPARK_HOME}" "$R_SCRIPT_PATH/"R CMD check $CRAN_CHECK_OPTIONS SparkR_"$VERSION".tar.gz
 fi
+
+# Install source package to get it to generate vignettes rds files, etc.
+if [ -n "$CLEAN_INSTALL" ]
```
shivaram: Isn't this already done by `install-dev.sh` ? I'm a bit confused as to why we need to call install again.

dev/make-distribution.sh
```diff
@@ -208,11 +212,24 @@ cp -r "$SPARK_HOME/data" "$DISTDIR"
...
+fi
+
+# Make R package - this is used for both CRAN release and packing R layout into distribution
+if [ "$MAKE_R" == "true" ]; then
+  echo "Building R source package"
+  pushd "$SPARK_HOME/R" > /dev/null
+  # Build source package and run full checks
+  # Install source package to get it to generate vignettes, etc.
+  # Do not source the check-cran.sh - it should be run from where it is for it to set SPARK_HOME
+  NO_TESTS=1 CLEAN_INSTALL=1 "$SPARK_HOME/"R/check-cran.sh
```
shivaram: Its a little awkward that we use `check-cran.sh` to build, install the package. I think it points to the fact that we can refactor the scripts more. But that can be done in a future PR

shivaram (commented):
Thanks @felixcheung - Change mostly looks good. The one thing I want to do is check if this works locally (we should also check this on Jenkins ideally to avoid any surprises while cutting the release cc @rxin)

Also do you think there are any changes in terms of what goes into the actual release zip / tar.gz built ?

felixcheung:
I thought we don't run `make-distribution.sh` on Jenkins, so I have been testing it manually/locally.
As per [here](https://github.com/apache/spark/pull/16014#discussion_r89912061) the output of install is different - I'll come back to list that in a bit.

shivaram:
Just to clarify the Jenkins comment, we use Jenkins to test PRs and that doesn't run make-distribution.sh -- But the main release builds (say 2.0.2[2]) are also built on Jenkins by a separate job that calls just `release-build.sh package`[1] with the appropriate environment variables. 

[1] https://github.com/apache/spark/blob/master/dev/create-release/release-build.sh#L149
[2]https://amplab.cs.berkeley.edu/jenkins/view/Spark%20Release/job/spark-release-package/57/consoleFull

rxin:
Why is this a blocker thing?

felixcheung:
@shivaram Cool, I did know about release-build but I didn't know it's running on Jenkins. I *think* we should be ok but might want to check Jenkins has "e1071" and "survival" which are optional for compatibility tests but `R CMD check` is enforcing/requiring it. If you recall, [this](https://github.com/apache/spark/pull/15790#issuecomment-259780799) is this conversation that prompted this change.

@rxin This PR updates what goes into the Spark binary release to match what we (intend to) release on CRAN for the R package

As for the diff, this is the delta between this PR and Spark 2.0.2 under the R/lib/SparkR directory. It turns out `R CMD check` also depends on Rd file generation in install-dev.sh (ie. `devtools::document(pkg="./pkg", roclets=c("rd")) }`).. this is going to take more time to untangle this in a follow up.

_what's additional_
```
SparkR/
-rw-r--r--   INDEX
drwxr-xr-x   doc

SparkR/Meta/
-rw-r--r--   vignette.rds

SparkR/doc/
-rw-r--r--   sparkr-vignettes.Rmd
-rw-r--r--   sparkr-vignettes.R
-rw-r--r--   sparkr-vignettes.html
-rw-r--r--   index.html
```

_what's omitted_
```
SparkR/html/
-rw-r--r--  1 root root  1319 Nov 29 08:05 R.css
-rw-r--r--  1 root root 81153 Nov 29 08:05 00Index.html
```

What it used to have `year.html` `write.parquet.html` `sparkR.session.html`, the html directory now only has 2 files. My understanding is these knitr html output are actually not used at runtime. I checked that `?sparkR.session` in the `sparkR` is still working correctly.

felixcheung:
Any more thought on this? Without this we don't really have a signed tarball in the official release to release to CRAN...

shivaram:
Sorry I was caught up with some other stuff today. Will take a final look tomm morning.

R/check-cran.sh
shivaram: So it looks like we lost the knitted HTML files in the SparkR package with this change. FWIW this may not be bad as the html files are not usually used locally and only used for the website and I think the docs creation part of the build should pick that up. (Verifying that now)

dev/make-distribution.sh
```diff
@@ -71,6 +72,9 @@ while (( "$#" )); do
     --pip)
       MAKE_PIP=true
       ;;
+    --r)
+      MAKE_R=true
```
shivaram: FWIW if you want this to get picked up by the official release building procedure we also need to edit release-build.sh [1]. Can you coordinate this with @rxin ?

[1] https://github.com/apache/spark/blob/d9eb4c7215f26dd05527c0b9980af35087ab9d64/dev/create-release/release-build.sh#L220

shivaram (commented):
Sorry for the delay @felixcheung - There was a bunch of other stuff I got caught up with. I think this change itself looks good. The only minor things I found were the issue with the `html` files and the need to change release-build.sh. 

I can merge this to master and branch-2.1  unless you think we should fix those above issues in this change.

felixcheung:
@shivaram yes, that is correct and what I've found [here](https://github.com/apache/spark/pull/16014#issuecomment-263505953)

Basically I don't think it is the R source package process to include the html files (they are still here getting created), and as explained and you pointed out, they are not actually used by R.

felixcheung:
As for release-build.sh - I had the change in there but I've changed it to make it more clear.

shivaram:
Ah got it - I didn't notice that. BTW the way @rxin handled this for pip is a bit different in https://github.com/apache/spark/commit/37e52f8793bff306a7ae5a9aecc16f28333b70e3

Might be good to merge with upstream and handle pip, R similarly ?

felixcheung:
thanks, updated. @rxin could you please review release-build.sh?

dev/create-release/release-build.sh
```diff
@@ -221,14 +235,13 @@ if [[ "$1" == "package" ]]; then
...
   # share the same Zinc server.
-  # Make R source package only once. (--r)
   FLAGS="-Psparkr -Phive -Phive-thriftserver -Pyarn -Pmesos"
   make_binary_release "hadoop2.3" "-Phadoop-2.3 $FLAGS" "3033" &
   make_binary_release "hadoop2.4" "-Phadoop-2.4 $FLAGS" "3034" &
   make_binary_release "hadoop2.6" "-Phadoop-2.6 $FLAGS" "3035" &
   make_binary_release "hadoop2.7" "-Phadoop-2.7 $FLAGS" "3036" "withpip" &
   make_binary_release "hadoop2.4-without-hive" "-Psparkr -Phadoop-2.4 -Pyarn -Pmesos" "3037" &
-  make_binary_release "without-hadoop" "--r -Psparkr -Phadoop-provided -Pyarn -Pmesos" "3038" &
+  make_binary_release "without-hadoop" "-Psparkr -Phadoop-provided -Pyarn -Pmesos" "3038" "withr" &
```
shivaram: Any specific reason to use the `without-hadoop` build for the R package ? Just wondering if this will affect the users in any fashion

felixcheung: It was mostly to use a "separate profile" from "withpip"

Running this `R CMD build` would run some Spark code (mainly in vignettes since tests are not run in `R CMD check`), but nothing that depends on the file system etc.

Also the Spark jar, while loaded and called into during the process, will not be packaged into the resulting R source package, so I thought it didn't matter which build profile we would run this in.

felixcheung: @shivaram

felixcheung: @shivaram what do you think about [this](https://github.com/apache/spark/pull/16014#discussion_r91196396)

I'd like to merge this to branch-2.1 to see if we could make it to 2.1.0 if at all possible.

shivaram: I think it sounds fine. I was waiting to see if @rxin (or @Joshrosen ?) would take a look because I have not reviewed changes to this file before. Let me take another closer look and then we can merge it to branch-2.1 -- We'll see what happens to the RC process after that

dev/create-release/release-build.sh
```diff
@@ -172,11 +172,30 @@ if [[ "$1" == "package" ]]; then
...
         -DzincPort=$ZINC_PORT 2>&1 >  ../binary-release-$NAME.log
       cd ..
+    elif [[ "$BUILD_PACKAGE" == "withr" ]]; then
+      echo "Creating distribution with R package"
+      ./dev/make-distribution.sh --name $NAME --mvn $MVN_HOME/bin/mvn --tgz --r $FLAGS \
+        -DzincPort=$ZINC_PORT 2>&1 >  ../binary-release-$NAME.log
+      cd ..
+
+      echo "Copying and signing R source package"
+      R_DIST_NAME=SparkR_$SPARK_VERSION.tar.gz
```
shivaram: Just to clarify this is the tgz that we will upload to CRAN right ?

shivaram:
LGTM. I took another look at `release-build.sh` and I think it looks fine. Merging this into master, branch-2.1. I'll also see if I can test this out somehow on jenkins

shivaram:
@felixcheung I triggered a nightly build on branch-2.1 and this doesn't work correctly in the no-hadoop build. While building the vignettes we run into an error -- I'm going to change this to use another of the hadoop builds

```
~/workspace/spark-branch-2.1-package/spark-2.1.1-SNAPSHOT-bin-without-hadoop/R
* checking for file '/home/jenkins/workspace/spark-branch-2.1-package/spark-2.1.1-SNAPSHOT-bin-without-hadoop/R/pkg/DESCRIPTION' ... OK
* preparing 'SparkR':
* checking DESCRIPTION meta-information ... OK
* installing the package to build vignettes
* creating vignettes ... ERROR

Attaching package: 'SparkR'

The following objects are masked from 'package:stats':

The following objects are masked from 'package:base':

    as.data.frame, colnames, colnames<-, drop, intersect, rank,
    rbind, sample, subset, summary, transform, union

Spark package found in SPARK_HOME: /home/jenkins/workspace/spark-branch-2.1-package/spark-2.1.1-SNAPSHOT-bin-without-hadoop
Exception in thread "main" java.lang.NoClassDefFoundError: org/apache/hadoop/fs/FSDataInputStream
        at org.apache.spark.deploy.SparkSubmitArguments$$anonfun$mergeDefaultSparkProperties$1.apply(SparkSubmitArguments.scala:118)
        at org.apache.spark.deploy.SparkSubmitArguments$$anonfun$mergeDefaultSparkProperties$1.apply(SparkSubmitArguments.scala:118)
        at scala.Option.getOrElse(Option.scala:121)
        at org.apache.spark.deploy.SparkSubmitArguments.mergeDefaultSparkProperties(SparkSubmitArguments.scala:118)
        at org.apache.spark.deploy.SparkSubmitArguments.<init>(SparkSubmitArguments.scala:104)
        at org.apache.spark.deploy.SparkSubmit$.main(SparkSubmit.scala:119)
        at org.apache.spark.deploy.SparkSubmit.main(SparkSubmit.scala)
Caused by: java.lang.ClassNotFoundException: org.apache.hadoop.fs.FSDataInputStream
        at java.net.URLClassLoader$1.run(URLClassLoader.java:366)
        at java.net.URLClassLoader$1.run(URLClassLoader.java:355)
        at java.security.AccessController.doPrivileged(Native Method)
        at java.net.URLClassLoader.findClass(URLClassLoader.java:354)
        at java.lang.ClassLoader.loadClass(ClassLoader.java:425)
        at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:308)
        at java.lang.ClassLoader.loadClass(ClassLoader.java:358)
        ... 7 more
Quitting from lines 31-32 (sparkr-vignettes.Rmd)
Error: processing vignette 'sparkr-vignettes.Rmd' failed with diagnostics:
JVM is not ready after 10 seconds
Execution halted
```

shivaram:
Hmm still doesn't work 

```
* DONE (SparkR)
+ popd
+ mkdir /home/jenkins/workspace/spark-branch-2.1-package/spark-2.1.1-SNAPSHOT-bin-hadoop2.6/dist/conf
+ cp /home/jenkins/workspace/spark-branch-2.1-package/spark-2.1.1-SNAPSHOT-bin-hadoop2.6/conf/docker.properties.template /home/jenkins/workspace/spark-branch-2.1-package/spark-2.1.1-SNAPSHOT-bin-hadoop2.6/conf/fairscheduler.xml.template /home/jenkins/workspace/spark-branch-2.1-package/spark-2.1.1-SNAPSHOT-bin-hadoop2.6/conf/log4j.properties.template /home/jenkins/workspace/spark-branch-2.1-package/spark-2.1.1-SNAPSHOT-bin-hadoop2.6/conf/metrics.properties.template /home/jenkins/workspace/spark-branch-2.1-package/spark-2.1.1-SNAPSHOT-bin-hadoop2.6/conf/slaves.template /home/jenkins/workspace/spark-branch-2.1-package/spark-2.1.1-SNAPSHOT-bin-hadoop2.6/conf/spark-defaults.conf.template /home/jenkins/workspace/spark-branch-2.1-package/spark-2.1.1-SNAPSHOT-bin-hadoop2.6/conf/spark-env.sh.template /home/jenkins/workspace/spark-branch-2.1-package/spark-2.1.1-SNAPSHOT-bin-hadoop2.6/dist/conf
+ cp /home/jenkins/workspace/spark-branch-2.1-package/spark-2.1.1-SNAPSHOT-bin-hadoop2.6/README.md /home/jenkins/workspace/spark-branch-2.1-package/spark-2.1.1-SNAPSHOT-bin-hadoop2.6/dist
+ cp -r /home/jenkins/workspace/spark-branch-2.1-package/spark-2.1.1-SNAPSHOT-bin-hadoop2.6/bin /home/jenkins/workspace/spark-branch-2.1-package/spark-2.1.1-SNAPSHOT-bin-hadoop2.6/dist
+ cp -r /home/jenkins/workspace/spark-branch-2.1-package/spark-2.1.1-SNAPSHOT-bin-hadoop2.6/python /home/jenkins/workspace/spark-branch-2.1-package/spark-2.1.1-SNAPSHOT-bin-hadoop2.6/dist
+ cp -r /home/jenkins/workspace/spark-branch-2.1-package/spark-2.1.1-SNAPSHOT-bin-hadoop2.6/sbin /home/jenkins/workspace/spark-branch-2.1-package/spark-2.1.1-SNAPSHOT-bin-hadoop2.6/dist
+ '[' -d /home/jenkins/workspace/spark-branch-2.1-package/spark-2.1.1-SNAPSHOT-bin-hadoop2.6/R/lib/SparkR ']'
+ mkdir -p /home/jenkins/workspace/spark-branch-2.1-package/spark-2.1.1-SNAPSHOT-bin-hadoop2.6/dist/R/lib
+ cp -r /home/jenkins/workspace/spark-branch-2.1-package/spark-2.1.1-SNAPSHOT-bin-hadoop2.6/R/lib/SparkR /home/jenkins/workspace/spark-branch-2.1-package/spark-2.1.1-SNAPSHOT-bin-hadoop2.6/dist/R/lib
+ cp /home/jenkins/workspace/spark-branch-2.1-package/spark-2.1.1-SNAPSHOT-bin-hadoop2.6/R/lib/sparkr.zip /home/jenkins/workspace/spark-branch-2.1-package/spark-2.1.1-SNAPSHOT-bin-hadoop2.6/dist/R/lib
+ '[' true == true ']'
+ TARDIR_NAME=spark-2.1.1-SNAPSHOT-bin-hadoop2.6
+ TARDIR=/home/jenkins/workspace/spark-branch-2.1-package/spark-2.1.1-SNAPSHOT-bin-hadoop2.6/spark-2.1.1-SNAPSHOT-bin-hadoop2.6
+ rm -rf /home/jenkins/workspace/spark-branch-2.1-package/spark-2.1.1-SNAPSHOT-bin-hadoop2.6/spark-2.1.1-SNAPSHOT-bin-hadoop2.6
+ cp -r /home/jenkins/workspace/spark-branch-2.1-package/spark-2.1.1-SNAPSHOT-bin-hadoop2.6/dist /home/jenkins/workspace/spark-branch-2.1-package/spark-2.1.1-SNAPSHOT-bin-hadoop2.6/spark-2.1.1-SNAPSHOT-bin-hadoop2.6
+ tar czf spark-2.1.1-SNAPSHOT-bin-hadoop2.6.tgz -C /home/jenkins/workspace/spark-branch-2.1-package/spark-2.1.1-SNAPSHOT-bin-hadoop2.6 spark-2.1.1-SNAPSHOT-bin-hadoop2.6
+ rm -rf /home/jenkins/workspace/spark-branch-2.1-package/spark-2.1.1-SNAPSHOT-bin-hadoop2.6/spark-2.1.1-SNAPSHOT-bin-hadoop2.6
Copying and signing R source package
cp: cannot stat `spark-2.1.1-SNAPSHOT-bin-hadoop2.6/R/SparkR_2.1.1-SNAPSHOT.tar.gz': No such file or directory
```

shivaram:
Hmm - The problem is that the directory we want to get the file from gets deleted in the previous step ? I'm going to debug this locally now.

felixcheung:
This just means the source tgz file is not created?

Copying and signing R source package
cp: cannot stat `spark-2.1.1-SNAPSHOT-bin-hadoop2.6/R/SparkR_2.1.1-SNAPSHOT.tar.gz': No such file or directory

felixcheung:
I'm guessing maybe the path needs to be a full path?

shivaram:
No I found the problem - its two fold
- The tgz is not copied into the `dist` directory in `make-distribution.sh` [1]. This is relatively easy to fix. It only works in Python because the entire python directory is copied into the dist directory.

- The second problem is one of naming - Our jar file is called SparkR_2.1.0.tar.gz as that is the version in DESCRIPTION file in R. I think we need to do a copy at some point to create SparkR_2.1.1-SNAPSHOT.tar.gz   

[1] https://github.com/apache/spark/blob/202fcd21ce01393fa6dfaa1c2126e18e9b85ee96/dev/make-distribution.sh#L244

felixcheung:
For the 2nd point, I thought make-tag.sh should update the version in DESCRIPTION file. Was that not the case when we make preparations for 2.1.1?

felixcheung:
https://github.com/apache/spark/blob/202fcd21ce01393fa6dfaa1c2126e18e9b85ee96/dev/create-release/release-tag.sh

I mean.

shivaram:
I think release-tag.sh is only run while building RCs and final releases - I dont think we run it for nightly builds - So thats not getting run as a part of my test.

shivaram:
Hmm looking more closely I dont think the directory being deleted / being in `dist/` is a real issue. This is because in `release-build.sh` we do `cd ..`[1] before trying to look for the file. I think just fixing the file naming thing should be sufficient. I'm going to try that first. 
[1] https://github.com/apache/spark/blob/202fcd21ce01393fa6dfaa1c2126e18e9b85ee96/dev/create-release/release-build.sh#L184

shivaram:
New PR in #16221

PR #16013, opened 2016-11-25
[SPARK-3359][DOCS] Make javadoc8 working for unidoc/genjavadoc compatibility in Java API documentation

HyukjinKwon:
## What changes were proposed in this pull request?

This PR make `sbt unidoc` complete with Java 8.

This PR roughly includes several fixes as below:

- Fix unrecognisable class and method links in javadoc by changing it from `[[..]]` to `` `...` `` 

  ```diff
  - * A column that will be computed based on the data in a [[DataFrame]].
  + * A column that will be computed based on the data in a `DataFrame`.
  ```

- Fix throws annotations so that they are recognisable in javadoc

- Fix URL links to `<a href="http..."></a>`.

  ```diff
  - * [[http://en.wikipedia.org/wiki/Decision_tree_learning Decision tree]] model for regression.
  + * <a href="http://en.wikipedia.org/wiki/Decision_tree_learning">
  + * Decision tree (Wikipedia)</a> model for regression.
  ```

  ```diff
  -   * @see http://en.wikipedia.org/wiki/Receiver_operating_characteristic
  +   * @see <a href="http://en.wikipedia.org/wiki/Receiver_operating_characteristic">
  +   * Receiver operating characteristic (Wikipedia)</a>
  ```

- Fix < to > to

  - `greater than`/`greater than or equal to` or `less than`/`less than or equal to` where applicable.

  - Wrap it with `{{{...}}}` to print them in javadoc or use `{@code ...}` or `{@literal ..}`. Please refer https://github.com/apache/spark/pull/16013#discussion_r89665558

- Fix `</p>` complaint

## How was this patch tested?

Manually tested by `jekyll build` with Java 7 and 8

```
java version "1.7.0_80"
Java(TM) SE Runtime Environment (build 1.7.0_80-b15)
Java HotSpot(TM) 64-Bit Server VM (build 24.80-b11, mixed mode)
```

```
java version "1.8.0_45"
Java(TM) SE Runtime Environment (build 1.8.0_45-b14)
Java HotSpot(TM) 64-Bit Server VM (build 25.45-b02, mixed mode)
```

PR #13855, opened 2016-06-22
[SPARK-16153] [MLLIB] switch to multi-line doc to avoid a genjavadoc bug

mengxr:
## What changes were proposed in this pull request?

We recently deprecated setLabelCol in ChiSqSelectorModel (#13823):

``` scala
  /** @group setParam */
  @Since("1.6.0")
  @deprecated("labelCol is not used by ChiSqSelectorModel.", "2.0.0")
  def setLabelCol(value: String): this.type = set(labelCol, value)
```

This unfortunately hit a genjavadoc bug and broken doc generation. This is the generated Java code:

``` java
  /** @group setParam */
  public  org.apache.spark.ml.feature.ChiSqSelectorModel setOutputCol (java.lang.String value)  { throw new RuntimeException(); }
   *
   * @deprecated labelCol is not used by ChiSqSelectorModel. Since 2.0.0.
  */
  public  org.apache.spark.ml.feature.ChiSqSelectorModel setLabelCol (java.lang.String value)  { throw new RuntimeException(); }
```

Switching to multiline is a workaround.

PR #16017, opened 2016-11-26
[SPARK-18592][ML] Move DT/RF/GBT Param setter methods to subclasses

yanboliang:
Mainly two changes:
* Move DT/RF/GBT Param setter methods to subclasses.
* Deprecate corresponding setter methods in the model classes.

See discussion here https://github.com/apache/spark/pull/15913#discussion_r89662469.

## How was this patch tested?
Existing tests.

### DIFF
--- a/mllib/src/main/scala/org/apache/spark/ml/classification/DecisionTreeClassifier.scala
+++ b/mllib/src/main/scala/org/apache/spark/ml/classification/DecisionTreeClassifier.scala
@@ -52,33 +52,49 @@ class DecisionTreeClassifier @Since("1.4.0") (
 
   // Override parameter setters from parent trait for Java API compatibility.
 
+  /** @group setParam */
   @Since("1.4.0")
-  override def setMaxDepth(value: Int): this.type = super.setMaxDepth(value)
+  override def setMaxDepth(value: Int): this.type = set(maxDepth, value)
 
+  /** @group setParam */
   @Since("1.4.0")
-  override def setMaxBins(value: Int): this.type = super.setMaxBins(value)
+  override def setMaxBins(value: Int): this.type = set(maxBins, value)
 
+  /** @group setParam */
   @Since("1.4.0")
-  override def setMinInstancesPerNode(value: Int): this.type =
-    super.setMinInstancesPerNode(value)
+  override def setMinInstancesPerNode(value: Int): this.type = set(minInstancesPerNode, value)
 
+  /** @group setParam */
   @Since("1.4.0")
-  override def setMinInfoGain(value: Double): this.type = super.setMinInfoGain(value)
+  override def setMinInfoGain(value: Double): this.type = set(minInfoGain, value)
 
+  /** @group expertSetParam */
   @Since("1.4.0")
-  override def setMaxMemoryInMB(value: Int): this.type = super.setMaxMemoryInMB(value)
+  override def setMaxMemoryInMB(value: Int): this.type = set(maxMemoryInMB, value)
 
+  /** @group expertSetParam */
   @Since("1.4.0")
-  override def setCacheNodeIds(value: Boolean): this.type = super.setCacheNodeIds(value)
+  override def setCacheNodeIds(value: Boolean): this.type = set(cacheNodeIds, value)
 
+  /**
+   * Specifies how often to checkpoint the cached node IDs.
+   * E.g. 10 means that the cache will get checkpointed every 10 iterations.
+   * This is only used if cacheNodeIds is true and if the checkpoint directory is set in
+   * [[org.apache.spark.SparkContext]].
+   * Must be >= 1.
+   * (default = 10)
+   * @group setParam
+   */
   @Since("1.4.0")
-  override def setCheckpointInterval(value: Int): this.type = super.setCheckpointInterval(value)
+  override def setCheckpointInterval(value: Int): this.type = set(checkpointInterval, value)
 
+  /** @group setParam */
   @Since("1.4.0")
-  override def setImpurity(value: String): this.type = super.setImpurity(value)
+  override def setImpurity(value: String): this.type = set(impurity, value)
 
+  /** @group setParam */
   @Since("1.6.0")
-  override def setSeed(value: Long): this.type = super.setSeed(value)
+  override def setSeed(value: Long): this.type = set(seed, value)
 
   override protected def train(dataset: Dataset[_]): DecisionTreeClassificationModel = {
     val categoricalFeatures: Map[Int, Int] =
--- a/mllib/src/main/scala/org/apache/spark/ml/tree/treeParams.scala
+++ b/mllib/src/main/scala/org/apache/spark/ml/tree/treeParams.scala
@@ -107,25 +107,41 @@ private[ml] trait DecisionTreeParams extends PredictorParams
   setDefault(maxDepth -> 5, maxBins -> 32, minInstancesPerNode -> 1, minInfoGain -> 0.0,
     maxMemoryInMB -> 256, cacheNodeIds -> false, checkpointInterval -> 10)
 
-  /** @group setParam */
+  /**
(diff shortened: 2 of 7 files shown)


yanboliang:
cc @jkbradley

mllib/src/main/scala/org/apache/spark/ml/classification/DecisionTreeClassifier.scala
```diff
@@ -52,33 +52,49 @@ class DecisionTreeClassifier @Since("1.4.0") (
...
+  override def setCacheNodeIds(value: Boolean): this.type = set(cacheNodeIds, value)
 
+  /**
+   * Specifies how often to checkpoint the cached node IDs.
+   * E.g. 10 means that the cache will get checkpointed every 10 iterations.
+   * This is only used if cacheNodeIds is true and if the checkpoint directory is set in
+   * [[org.apache.spark.SparkContext]].
+   * Must be >= 1.
+   * (default = 10)
+   * @group setParam
```
felixcheung: just curious, repeating this is kinda tedious - I thought it inherits the doc from the super, does it not?

mllib/src/main/scala/org/apache/spark/ml/classification/DecisionTreeClassifier.scala
```diff
@@ -52,33 +52,49 @@ class DecisionTreeClassifier @Since("1.4.0") (
 
   // Override parameter setters from parent trait for Java API compatibility.
 
+  /** @group setParam */
   @Since("1.4.0")
-  override def setMaxDepth(value: Int): this.type = super.setMaxDepth(value)
+  override def setMaxDepth(value: Int): this.type = set(maxDepth, value)
```
jkbradley: Do you need to change this?  I'd prefer the former, in case some special logic is ever added to super.setMaxDepth.

Same elsewhere

mllib/src/main/scala/org/apache/spark/ml/classification/DecisionTreeClassifier.scala
yanboliang: The cause of this change was suggested at https://github.com/apache/spark/pull/15913#discussion_r89662469 , since Param setter methods in traits used to have the wrong type in Java. We would like to remove the setter method from the trait since it does not make sense to have it in the Model classes. We could put the setter method in each subclass and then deprecate the method in the Model classes.

So if we remove the setter method from the traits, we can not inherit docs from them. BTW, the current change is consistent with other ML algorithms which inherit traits.

jkbradley: Right, that would make the setter appear in models.  Your way is better.

mllib/src/main/scala/org/apache/spark/ml/tree/treeParams.scala
```diff
@@ -107,25 +107,41 @@ private[ml] trait DecisionTreeParams extends PredictorParams
   setDefault(maxDepth -> 5, maxBins -> 32, minInstancesPerNode -> 1, minInfoGain -> 0.0,
     maxMemoryInMB -> 256, cacheNodeIds -> false, checkpointInterval -> 10)
 
-  /** @group setParam */
+  /**
+   * @deprecated This method is deprecated and will be removed in 2.2.0.
+   * @group setParam
+   */
+  @deprecated("This method is deprecated and will be removed in 2.2.0.", "2.1.0")
```
jkbradley: Does this appear as deprecated in the Estimator classes?  If so, then can you update the comment to make it clear it is only deprecated for Models?

jkbradley:
LGTM pending a check on whether the deprecations affect Estimators in addition to Models
Thanks for the follow-up!

mllib/src/main/scala/org/apache/spark/ml/tree/treeParams.scala
```diff
@@ -134,27 +150,31 @@ private[ml] trait DecisionTreeParams extends PredictorParams
   /** @group setParam */
   def setSeed(value: Long): this.type = set(seed, value)
```
jkbradley: Are you intentionally keeping setSeed in Models?

mllib/src/main/scala/org/apache/spark/ml/tree/treeParams.scala
jkbradley: Actually, if this is a problem, we could mark the setters in Models as deprecated and leave the treeParams alone since the treeParams traits are private.

yanboliang: This will not appear in estimator classes, only in models. This can be confirmed by the following ways:
The IDE will prompt deprecated methods only for models: 
image
We can also confirm it in Scala API docs:
```DecisionTreeClassifier```
image
```DecisionTreeClassificationModel```
image

File: mllib/src/main/scala/org/apache/spark/ml/tree/treeParams.scala (hunk 4 above)
Comment (yanboliang) [inline mllib/src/main/scala/org/apache/spark/ml/tree/treeParams.scala]: Oops, I missed it and will add it. Thanks.

**Comment by yanboliang** [comment]
Jenkins, test this please.

File: mllib/src/main/scala/org/apache/spark/ml/tree/treeParams.scala (hunk 3 above)
Comment (jkbradley) [inline mllib/src/main/scala/org/apache/spark/ml/tree/treeParams.scala]: Great, thank you for checking!

**Comment by jkbradley** [comment]
LGTM
Thanks!

File: mllib/src/main/scala/org/apache/spark/ml/classification/DecisionTreeClassifier.scala [hunk 5]
```diff
@@ -52,33 +52,49 @@ class DecisionTreeClassifier @Since("1.4.0") (
...
   @Since("1.4.0")
-  override def setCacheNodeIds(value: Boolean): this.type = super.setCacheNodeIds(value)
+  override def setCacheNodeIds(value: Boolean): this.type = set(cacheNodeIds, value)
 
+  /**
+   * Specifies how often to checkpoint the cached node IDs.
+   * E.g. 10 means that the cache will get checkpointed every 10 iterations.
+   * This is only used if cacheNodeIds is true and if the checkpoint directory is set in
+   * [[org.apache.spark.SparkContext]].
+   * Must be >= 1.
```
HyukjinKwon: Hi all, this might be too trivial but I just want to let you know we probably should write `<` or `>` as other ones such as `{@literal <}` or `{@literal >}`. Please refer https://github.com/apache/spark/pull/16013#discussion_r89666914.

mllib/src/main/scala/org/apache/spark/ml/classification/DecisionTreeClassifier.scala
```diff
@@ -52,33 +52,49 @@ class DecisionTreeClassifier @Since("1.4.0") (
 
   // Override parameter setters from parent trait for Java API compatibility.
 
+  /** @group setParam */
```
HyukjinKwon: I am sorry but please allow me leave another trivial comment about documentation. I haven't checked the built documentation but it seems we should use multiple-lines in this case (see https://github.com/apache/spark/pull/13855).

yanboliang: I checked the generated documentation and found it works well. Could you let me know what you found and how to reproduce it? Thanks.

HyukjinKwon: Oh, I should have built and checked it first closely by myself. Thank you for correcting me.

jkbradley: Btw, I created http://issues.apache.org/jira/browse/SPARK-18692 for monitoring Java 8 unidoc in the future

PR #16019, opened 2016-11-27
[SPARK-18595] [SQL] Handling ignoreIfExists in HiveExternalCatalog createTable API

gatorsmile:
Currently, `HiveExternalCatalog`'s `createTable` issues the exception if `ignoreIfExists` is true and the table exists. However, this is just a partial handling of `ignoreIfExists`. We can cover the other cases too. 

Instead of passing `ignoreIfExists` of CREATE TABLE to Hive metastore, we can handle it in `HiveExternalCatalog`'s `createTable`. 

Covered by the existing test case.

### DIFF
--- a/sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveExternalCatalog.scala
+++ b/sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveExternalCatalog.scala
@@ -191,42 +191,44 @@ private[spark] class HiveExternalCatalog(conf: SparkConf, hadoopConf: Configurat
     requireDbExists(db)
     verifyTableProperties(tableDefinition)
 
-    if (tableExists(db, table) && !ignoreIfExists) {
-      throw new TableAlreadyExistsException(db = db, table = table)
-    }
-
-    if (tableDefinition.tableType == VIEW) {
-      client.createTable(tableDefinition, ignoreIfExists)
-    } else {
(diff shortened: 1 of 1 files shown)


sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveExternalCatalog.scala
```diff
@@ -191,42 +191,44 @@ private[spark] class HiveExternalCatalog(conf: SparkConf, hadoopConf: Configurat
     requireDbExists(db)
     verifyTableProperties(tableDefinition)
 
-    if (tableExists(db, table) && !ignoreIfExists) {
-      throw new TableAlreadyExistsException(db = db, table = table)
-    }
-
-    if (tableDefinition.tableType == VIEW) {
```
gatorsmile: To reviewers: the actual code changes are very small. No code changes after this line.

gatorsmile:
cc @cloud-fan @hvanhovell

cloud-fan:
so the actual difference is, if the table exists and `ignoreIfExists` is true, previously we will still try to create this table, but logically we should do nothing.

Since we are revisiting it, I have some more thoughts. If the table already exists and `ignoreIfExists` is false, we will throw exception at spark side. However, if the table doesn't exist at the beginning, and get created right after the if check in `HiveExternalCatalog.createTable`, then we will throw exception at hive side. Shall we unify this and let hive metastore to handle `ignoreIfExists`?

PR #16020, opened 2016-11-27
[SPARK-18596][ML] add checking and caching to bisecting kmeans

hhbyyh:
jira: https://issues.apache.org/jira/browse/SPARK-18596 
This is a follow up for https://issues.apache.org/jira/browse/SPARK-18356.

Check if the DataFrame sent to BisectingKMeans is cached, if not, we need to cache the converted RDD to ensure performance for BisectingKMeans.

## How was this patch tested?

existing unit tests.


mllib/src/main/scala/org/apache/spark/ml/clustering/BisectingKMeans.scala
```diff
@@ -255,10 +256,19 @@ class BisectingKMeans @Since("2.0.0") (
 
   @Since("2.0.0")
   override def fit(dataset: Dataset[_]): BisectingKMeansModel = {
+    val handlePersistence = dataset.rdd.getStorageLevel == StorageLevel.NONE
```
MLnick: By the way, I've been meaning to log a ticket for this issue, but have been tied up.

This will actually never work. `dataset.rdd` will always have storage level `NONE`. To see this:

```
scala> import org.apache.spark.storage.StorageLevel
import org.apache.spark.storage.StorageLevel

scala> val df = spark.range(10).toDF("num")
df: org.apache.spark.sql.DataFrame = [num: bigint]

scala> df.storageLevel == StorageLevel.NONE
res0: Boolean = true

scala> df.persist
res1: df.type = [num: bigint]

scala> df.storageLevel == StorageLevel.MEMORY_AND_DISK
res2: Boolean = true

scala> df.rdd.getStorageLevel == StorageLevel.MEMORY_AND_DISK
res3: Boolean = false

scala> df.rdd.getStorageLevel == StorageLevel.NONE
res4: Boolean = true
```

So in fact all the algorithms that are checking for storage level using `dataset.rdd` are actually double-caching the data if the input DataFrame is actually cached, because the RDD will not appear to be cached.

So we should migrate all the checks to use `dataset.storageLevel` which was added in https://github.com/apache/spark/pull/13780

mllib/src/main/scala/org/apache/spark/ml/clustering/KMeans.scala
```diff
@@ -334,10 +334,8 @@ class KMeans @Since("1.5.0") (
     val summary = new KMeansSummary(
       model.transform(dataset), $(predictionCol), $(featuresCol), $(k))
     model.setSummary(Some(summary))
+    if (handlePersistence) instances.unpersist()
     instr.logSuccess(model)
-    if (handlePersistence) {
```
MLnick: prefer to keep this form according to style guide.

mllib/src/main/scala/org/apache/spark/ml/clustering/BisectingKMeans.scala
```diff
@@ -273,6 +283,7 @@ class BisectingKMeans @Since("2.0.0") (
     val summary = new BisectingKMeansSummary(
       model.transform(dataset), $(predictionCol), $(featuresCol), $(k))
     model.setSummary(Some(summary))
+    if (handlePersistence) rdd.unpersist()
```
MLnick: Prefer 

```
if (handlePersistence) {
  rdd.unpersist()
}
```

MLnick (changes requested):

mllib/src/main/scala/org/apache/spark/ml/clustering/KMeans.scala
```diff
@@ -334,10 +334,10 @@ class KMeans @Since("1.5.0") (
     val summary = new KMeansSummary(
       model.transform(dataset), $(predictionCol), $(featuresCol), $(k))
     model.setSummary(Some(summary))
-    instr.logSuccess(model)
     if (handlePersistence) {
       instances.unpersist()
     }
+    instr.logSuccess(model)
```
MLnick: The `handlePersistence` check in `KMeans` at L309 should also be updated to use `dataset.storageLevel`. Since we're touching KMeans here anyway we may as well do it now.

hhbyyh:
Sure we can change KMeans as well. And perhaps we can also add some unit tests. I'll send update.

MLnick:
Yes unit tests would be good to add.

Tests may require using event listeners to check the caching of the
intermediate dataset with/without cached initial data. Or at least that is
the only way I thought of so far ...

hhbyyh:
I just added a straight forward check for the data frame cache level.

hhbyyh:
@MLnick Do you think we still need the event listener unit test?

hhbyyh:
Close this as it's better resolved in https://issues.apache.org/jira/browse/SPARK-18608.
Thanks for the comments and discussion.

PR #16022, opened 2016-11-27
[SPARK-18598][SQL] Fixed inconsistent bean property inference

hamishmorgan:

`JavaTypeInference#inferDataType` used a different bean property filter from `serializerFor` and `deserializerFor`. While `inferDataType` required properties to simple not be named "class", the others require properties to have associated accessor and mutator. This caused an inconsistency in the `ExpressionEncoder`, resulting in an `AssertionError` when Datasets are operated on. (see https://issues.apache.org/jira/browse/SPARK-18598)

The PR patches `JavaTypeInference` to _always_ require properties to have an associated accessor and mutator.

## How was this patch tested?

Added a test case for the change. 

The larger, upstream issue (of AssertionError being thrown) has been tested with an integration test (as seen in the JIRA description) but I haven't submitted that.

### DIFF
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/JavaTypeInference.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/JavaTypeInference.scala
@@ -108,10 +108,7 @@ object JavaTypeInference {
         (MapType(keyDataType, valueDataType, nullable), true)
 
       case _ =>
-        // TODO: we should only collect properties that have getter and setter. However, some tests
-        // pass in scala case class as java bean class which doesn't have getter and setter.
-        val beanInfo = Introspector.getBeanInfo(typeToken.getRawType)
-        val properties = beanInfo.getPropertyDescriptors.filterNot(_.getName == "class")
+        val properties = getJavaBeanProperties(typeToken.getRawType)
         val fields = properties.map { property =>
           val returnType = typeToken.method(property.getReadMethod).getReturnType
           val (dataType, nullable) = inferDataType(returnType)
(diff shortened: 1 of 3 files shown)


srowen:
Given the comment it sounds like this change could cause tests to fail. We'll find out shortly. It seems like a good change because indeed bean properties are defined by getters and/or setters so it seems like the same definition should be used here.

@cloud-fan will know more.

srowen:
Jenkins test this please

hamishmorgan:
Thanks @srowen - I see you were right. Not sure how proceed. Some options:

 * Update failing tests to load their data a different way. 
 * Update the relevant beans (used in the tests) to include accessors and mutators.
 * Define a bean as _any_ property, irrespective of accessors. This gives up on the bean contract, which maybe doesn't make sense in Scala anyway. 
 * Hack something into the ExpressionEncoder to make the resultant Datasets less broken, but that's not really dealing with the root cause.
 
Either way, I think this becomes a breaking change. Thoughts?

srowen:
Yes, all but one of the failures arise because `LabeledPoint` is used as a bean class but it's a case class. It would be straightforward to add `def getFeatures = features` and `def getLabel = label` to sort that out. (The other failure is a test-only case class.) At least, those are the only failures that turned up so far here.

I suppose the only catch is that's kind of a behavior change, if case classes stop working, even though they weren't intended to work. Or is it a change -- I don't quite see that they ever fully worked? they just happened to because the tests didn't cause a case class to be used with the serialize/deserialize path?

rxin:
We can't do this since it breaks compatibility now.

hamishmorgan:
Okay. If it can't be done I'm happy to close the PR. I'll  leave the JIRA ticket open though, since it's still a bug IMO. Thanks for the help guys.

rxin:
Can we fix that specific bug than taking away the ability to infer just case classes?

hamishmorgan:
Probably. Worst case, we should be able to at least fail faster with more helpful message. And maybe there's something better. I'll have a poke around and see.

PR #16026, opened 2016-11-28
[SPARK-18597][SQL] Do not push-down join conditions to the right side of a LEFT ANTI join

hvanhovell:
We currently push down join conditions of a Left Anti join to both sides of the join. This is similar to Inner, Left Semi and Existence (a specialized left semi) join. The problem is that this changes the semantics of the join; a left anti join filters out rows that matches the join condition.

This PR fixes this by only pushing down conditions to the left hand side of the join. This is similar to the behavior of left outer join.

## How was this patch tested?
Added tests to `FilterPushdownSuite.scala` and created a SQLQueryTestSuite file for left anti joins with a regression test.


hvanhovell:
cc @dongjoon-hyun could you take a look at this?

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/optimizer/FilterPushdownSuite.scala
```diff
@@ -514,6 +514,39 @@ class FilterPushdownSuite extends PlanTest {
...
+        .analyze
+    val optimized = Optimize.execute(originalQuery)
+    val correctAnswer =
+      x.where("x.a".attr > 10)
+        .join(y, LeftAnti, Some("x.b".attr === "y.b".attr))
+        .analyze
+    comparePlans(optimized, analysis.EliminateSubqueryAliases(correctAnswer))
+  }
+
+  test("joins: only push down to the right of a left anti join") {
```
gatorsmile: How about `joins: only push down join conditions to the right of a left anti join`?

gatorsmile:
The current PR title is a little bit confusing. Could you please update the PR title from `push-down predicates` to `push-down join conditions`?

dongjoon-hyun:
Thank you for pinging me, @hvanhovell . I'll take a look, too.

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/optimizer/FilterPushdownSuite.scala
dongjoon-hyun: Do we need a JIRA issue number here?

gatorsmile:
LGTM except a minor comment.

dongjoon-hyun:
LGTM, @hvanhovell .

hvanhovell:
I am merging this to master/2.1. Thanks for the reviews!

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/Optimizer.scala
```diff
@@ -932,7 +932,7 @@ object PushPredicateThroughJoin extends Rule[LogicalPlan] with PredicateHelper {
         split(joinCondition.map(splitConjunctivePredicates).getOrElse(Nil), left, right)
 
       joinType match {
-        case _: InnerLike | LeftExistence(_) =>
+        case _: InnerLike |  LeftSemi | ExistenceJoin(_) =>
```
nsyca: The semantics of `ExistenceJoin` says we need to preserve all the rows from the left table through the join operation as if it is a regular `LeftOuter` join. The `ExistenceJoin` augments the `LeftOuter` operation with a new column called exists, set to true when the join condition in the ON clause is true and false otherwise. The filter of any rows will happen in the Filter operation above the `ExistenceJoin`.

Example:

A(c1, c2):
{ (1, 1), (1, 2) }

B(c1):
{ (NULL) }

// can be any value as it is irrelevant in this example
````
select A.*
from   A
where  exists (select 1 from B where A.c1 = A.c2)
       or A.c2=2
````
In this example, the correct result is all the rows from A. If the pattern `ExistenceJoin` at line 935 in `Optimizer.scala` added by the PR of this JIRA is indeed active, the code will push down the predicate `A.c1 = A.c2` to be a Filter on relation A, which will filter the row (1,2) from A.

gatorsmile: @nsyca Have you tried the above example using the latest master?

nsyca: I should have included the full comment from the JIRA. The keyword is "this is not currently exposed". Here is the first part of my comment:

"`ExistenceJoin` should be treated the same as `LeftOuter` and `LeftAnti`, not `InnerLike` and `LeftSemi`. This is not currently exposed because the rewrite of `[NOT] EXISTS OR ...` to `ExistenceJoin` happens in rule `RewritePredicateSubquery`, which is in a separate rule set and placed after the rule `PushPredicateThroughJoin`. During the transformation in the rule `PushPredicateThroughJoin`, an `ExistenceJoin` never exists."

PR #16029, opened 2016-11-28
[MINOR][ML] Fix wrong @since version in KMeans

zhengruifeng:
fix wrong version in `KMeans.fit`
there were two `import GeneralizedLinearRegression._` in trait GLR.GeneralizedLinearRegressionBase
## How was this patch tested?
existing tests

### DIFF
--- a/mllib/src/main/scala/org/apache/spark/ml/clustering/KMeans.scala
+++ b/mllib/src/main/scala/org/apache/spark/ml/clustering/KMeans.scala
@@ -310,7 +310,7 @@ class KMeans @Since("1.5.0") (
     fit(dataset, handlePersistence)
   }
 
-  @Since("2.2.0")
+  @Since("2.1.0")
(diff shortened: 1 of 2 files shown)


srowen:
This is too trivial to bother with.

zhengruifeng:
@srowen I found a small bug, and merge it in.

mllib/src/main/scala/org/apache/spark/ml/clustering/KMeans.scala
```diff
@@ -310,7 +310,7 @@ class KMeans @Since("1.5.0") (
     fit(dataset, handlePersistence)
   }
 
-  @Since("2.2.0")
+  @Since("2.1.0")
```
srowen: This method doesn't appear to be in 2.1 though:
https://github.com/apache/spark/blob/branch-2.1/mllib/src/main/scala/org/apache/spark/ml/clustering/KMeans.scala

let's close this please

PR #16030, opened 2016-11-28
[SPARK-18108][SQL] Fix a schema inconsistent bug that makes a parquet reader fail to read data

maropu:
A vectorized parquet reader fails to read column data if data schema and partition schema overlap with each other and inferred types in the partition schema differ from ones in the data schema. An example code to reproduce this bug is as follows;

```
scala> case class A(a: Long, b: Int)
scala> val as = Seq(A(1, 2))
scala> spark.createDataFrame(as).write.parquet("/data/a=1/")
scala> val df = spark.read.parquet("/data/")
scala> df.printSchema
root
 |-- a: long (nullable = true)
 |-- b: integer (nullable = true)
scala> df.collect
java.lang.NullPointerException
        at org.apache.spark.sql.execution.vectorized.OnHeapColumnVector.getLong(OnHeapColumnVector.java:283)
        at org.apache.spark.sql.execution.vectorized.ColumnarBatch$Row.getLong(ColumnarBatch.java:191)
        at org.apache.spark.sql.catalyst.expressions.GeneratedClass$SpecificUnsafeProjection.apply(Unknown Source)
        at org.apache.spark.sql.catalyst.expressions.GeneratedClass$SpecificUnsafeProjection.apply(Unknown Source)
        at scala.collection.Iterator$$anon$11.next(Iterator.scala:409)
        at scala.collection.Iterator$$anon$11.next(Iterator.scala:409)
```
The root cause is that a logical layer (`HadoopFsRelation`) and a physical layer (`VectorizedParquetRecordReader`) have a different assumption on partition schema; the logical layer trusts the data schema to infer the type the overlapped partition columns, and, on the other hand, the physical layer trusts partition schema which is inferred from path string. To fix this bug, this pr simply updates `HadoopFsRelation.schema` to respect the partition columns position in data schema and respect the partition columns type in partition schema.

## How was this patch tested?
Add tests in `ParquetPartitionDiscoverySuite`

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
@@ -132,16 +132,14 @@ case class DataSource(
       }.toArray
       new InMemoryFileIndex(sparkSession, globbedPaths, options, None)
     }
-
-    val fieldEquality = sparkSession.sessionState.conf.resolver
-
     val partitionSchema = if (partitionColumns.isEmpty) {
       // Try to infer partitioning, because no DataSource in the read path provides the partitioning
       // columns properly unless it is a Hive DataSource
       val resolved = tempFileIndex.partitionSchema.map { partitionField =>
+        val equality = sparkSession.sessionState.conf.resolver
         // SPARK-18510: try to get schema from userSpecifiedSchema, otherwise fallback to inferred
-        userSpecifiedSchema.flatMap(_.find(f => fieldEquality(f.name, partitionField.name)))
-          .getOrElse(partitionField)
+        userSpecifiedSchema.flatMap(_.find(f => equality(f.name, partitionField.name))).getOrElse(
+          partitionField)
       }
       StructType(resolved)
     } else {
@@ -152,9 +150,10 @@ case class DataSource(
         inferredPartitions
       } else {
         val partitionFields = partitionColumns.map { partitionColumn =>
-          userSpecifiedSchema.flatMap(_.find(c => fieldEquality(c.name, partitionColumn))).orElse {
+          val equality = sparkSession.sessionState.conf.resolver
+          userSpecifiedSchema.flatMap(_.find(c => equality(c.name, partitionColumn))).orElse {
             val inferredPartitions = tempFileIndex.partitionSchema
-            val inferredOpt = inferredPartitions.find(p => fieldEquality(p.name, partitionColumn))
+            val inferredOpt = inferredPartitions.find(p => equality(p.name, partitionColumn))
             if (inferredOpt.isDefined) {
               logDebug(
                 s"""Type of partition column: $partitionColumn not found in specified schema
@@ -175,12 +174,13 @@ case class DataSource(
         StructType(partitionFields)
       }
     }
-
     if (justPartitioning) {
       return (null, partitionSchema)
     }
-
-    val dataSchema = userSpecifiedSchema.orElse {
+    val dataSchema = userSpecifiedSchema.map { schema =>
+      val equality = sparkSession.sessionState.conf.resolver
+      StructType(schema.filterNot(f => partitionSchema.exists(p => equality(p.name, f.name))))
+    }.orElse {
       format.inferSchema(
         sparkSession,
         caseInsensitiveOptions,
@@ -189,11 +189,7 @@ case class DataSource(
       throw new AnalysisException(
         s"Unable to infer schema for $format. It must be specified manually.")
     }
-
-    val dataWithoutPartitions = dataSchema.filterNot { field =>
-      partitionSchema.exists(p => fieldEquality(p.name, field.name))
-    }
-    (StructType(dataWithoutPartitions), partitionSchema)
+    (dataSchema, partitionSchema)
   }
 
   /** Returns the name and schema of the source that can be used to continually read data. */
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
@@ -32,6 +32,7 @@ import org.apache.spark.sql._
 import org.apache.spark.sql.catalyst.catalog.{BucketSpec, CatalogStorageFormat, CatalogTable}
 import org.apache.spark.sql.catalyst.expressions.Attribute
 import org.apache.spark.sql.catalyst.util.CaseInsensitiveMap
+import org.apache.spark.sql.execution.datasources.PartitioningUtils._
 import org.apache.spark.sql.execution.datasources.csv.CSVFileFormat
 import org.apache.spark.sql.execution.datasources.jdbc.JdbcRelationProvider
 import org.apache.spark.sql.execution.datasources.json.JsonFileFormat
@@ -432,7 +433,19 @@ case class DataSource(
         val outputPath = if (allPaths.length == 1) {
           val path = new Path(allPaths.head)
           val fs = path.getFileSystem(sparkSession.sessionState.newHadoopConf())
-          path.makeQualified(fs.getUri, fs.getWorkingDirectory)
+          val qualifiedPath = path.makeQualified(fs.getUri, fs.getWorkingDirectory)
+          // Check if a data schema does not include partition columns explicitly defined in path
+          val equality = sparkSession.sessionState.conf.resolver
+          parsePartition(qualifiedPath, false, Set.empty[Path])._1
+              .map { case PartitionValues(columnNames, _) =>
+            columnNames.foreach { columnName =>
+              if (data.schema.exists(f => equality(f.name, columnName))) {
+                throw new IllegalArgumentException(s"Cannot use the columns of the data schema " +
+                  s"${data.schema.simpleString} as partition ones, but got $path")
+              }
+            }
+          }
+          qualifiedPath
         } else {
           throw new IllegalArgumentException("Expected exactly one path to be specified, but " +
             s"got: ${allPaths.mkString(", ")}")
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
@@ -135,7 +135,7 @@ case class DataSource(
 
     val fieldEquality = sparkSession.sessionState.conf.resolver
 
-    val partitionSchema = if (partitionColumns.isEmpty && catalogTable.isEmpty) {
+    val partitionSchema = if (partitionColumns.isEmpty) {
       // Try to infer partitioning, because no DataSource in the read path provides the partitioning
       // columns properly unless it is a Hive DataSource
       val resolved = tempFileIndex.partitionSchema.map { partitionField =>
@@ -190,14 +190,10 @@ case class DataSource(
         s"Unable to infer schema for $format. It must be specified manually.")
     }
 
-    // Override the fields of the partition schema if the data schema has the same field
-    val resolvedPartitionSchema = partitionSchema.map { partitionField =>
-      dataSchema.find(f => fieldEquality(f.name, partitionField.name)).getOrElse(partitionField)
-    }
     val dataWithoutPartitions = dataSchema.filterNot { field =>
       partitionSchema.exists(p => fieldEquality(p.name, field.name))
     }
-    (StructType(dataWithoutPartitions), StructType(resolvedPartitionSchema))
+    (StructType(dataWithoutPartitions), partitionSchema)
   }
 
   /** Returns the name and schema of the source that can be used to continually read data. */
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
@@ -133,14 +133,11 @@ case class DataSource(
       new InMemoryFileIndex(sparkSession, globbedPaths, options, None)
     }
 
-    val dataSchema = userSpecifiedSchema.map { schema =>
-      StructType(schema)
-    }.orElse {
-      val schema = format.inferSchema(
+    val dataSchema = userSpecifiedSchema.orElse {
+      format.inferSchema(
         sparkSession,
         caseInsensitiveOptions,
         tempFileIndex.allFiles())
-      schema
     }.getOrElse {
       throw new AnalysisException(
         s"Unable to infer schema for $format. It must be specified manually.")
@@ -192,9 +189,11 @@ case class DataSource(
     if (justPartitioning) {
       (null, partitionSchema)
     } else if (userSpecifiedSchema.isDefined) {
-      val equality = sparkSession.sessionState.conf.resolver
-      (StructType(dataSchema.filterNot(f => partitionSchema.exists(p => equality(p.name, f.name)))),
-        partitionSchema)
+      val dataWithoutPartitions = dataSchema.filterNot { field =>
+        val equality = sparkSession.sessionState.conf.resolver
+        partitionSchema.exists(p => equality(p.name, field.name))
+      }
+      (StructType(dataWithoutPartitions), partitionSchema)
     } else {
       (dataSchema, partitionSchema)
     }
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
@@ -132,13 +132,27 @@ case class DataSource(
       }.toArray
       new InMemoryFileIndex(sparkSession, globbedPaths, options, None)
     }
-    val partitionSchema = if (partitionColumns.isEmpty) {
+
+    val dataSchema = userSpecifiedSchema.map { schema =>
+      StructType(schema)
+    }.orElse {
+      val schema = format.inferSchema(
+        sparkSession,
+        caseInsensitiveOptions,
+        tempFileIndex.allFiles())
+      schema
+    }.getOrElse {
+      throw new AnalysisException(
+        s"Unable to infer schema for $format. It must be specified manually.")
+    }
+
+    val partitionSchema = if (partitionColumns.isEmpty && catalogTable.isEmpty) {
       // Try to infer partitioning, because no DataSource in the read path provides the partitioning
       // columns properly unless it is a Hive DataSource
       val resolved = tempFileIndex.partitionSchema.map { partitionField =>
         val equality = sparkSession.sessionState.conf.resolver
-        // SPARK-18510: try to get schema from userSpecifiedSchema, otherwise fallback to inferred
-        userSpecifiedSchema.flatMap(_.find(f => equality(f.name, partitionField.name))).getOrElse(
+        // SPARK-18510: try to get partition schema from data schema, otherwise fallback to inferred
+        dataSchema.find(f => equality(f.name, partitionField.name)).getOrElse(
           partitionField)
       }
       StructType(resolved)
@@ -174,22 +188,16 @@ case class DataSource(
         StructType(partitionFields)
       }
     }
+
     if (justPartitioning) {
-      return (null, partitionSchema)
-    }
-    val dataSchema = userSpecifiedSchema.map { schema =>
+      (null, partitionSchema)
+    } else if (userSpecifiedSchema.isDefined) {
       val equality = sparkSession.sessionState.conf.resolver
-      StructType(schema.filterNot(f => partitionSchema.exists(p => equality(p.name, f.name))))
-    }.orElse {
-      format.inferSchema(
-        sparkSession,
-        caseInsensitiveOptions,
-        tempFileIndex.allFiles())
-    }.getOrElse {
-      throw new AnalysisException(
-        s"Unable to infer schema for $format. It must be specified manually.")
+      (StructType(dataSchema.filterNot(f => partitionSchema.exists(p => equality(p.name, f.name)))),
+        partitionSchema)
+    } else {
+      (dataSchema, partitionSchema)
     }
-    (dataSchema, partitionSchema)
   }
 
   /** Returns the name and schema of the source that can be used to continually read data. */
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
@@ -133,24 +133,15 @@ case class DataSource(
       new InMemoryFileIndex(sparkSession, globbedPaths, options, None)
     }
 
-    val dataSchema = userSpecifiedSchema.orElse {
-      format.inferSchema(
-        sparkSession,
-        caseInsensitiveOptions,
-        tempFileIndex.allFiles())
-    }.getOrElse {
-      throw new AnalysisException(
-        s"Unable to infer schema for $format. It must be specified manually.")
-    }
+    val fieldEquality = sparkSession.sessionState.conf.resolver
 
     val partitionSchema = if (partitionColumns.isEmpty && catalogTable.isEmpty) {
       // Try to infer partitioning, because no DataSource in the read path provides the partitioning
       // columns properly unless it is a Hive DataSource
       val resolved = tempFileIndex.partitionSchema.map { partitionField =>
-        val equality = sparkSession.sessionState.conf.resolver
-        // SPARK-18510: try to get partition schema from data schema, otherwise fallback to inferred
-        dataSchema.find(f => equality(f.name, partitionField.name)).getOrElse(
-          partitionField)
+        // SPARK-18510: try to get schema from userSpecifiedSchema, otherwise fallback to inferred
+        userSpecifiedSchema.flatMap(_.find(f => fieldEquality(f.name, partitionField.name)))
+          .getOrElse(partitionField)
       }
       StructType(resolved)
     } else {
@@ -161,10 +152,9 @@ case class DataSource(
         inferredPartitions
       } else {
         val partitionFields = partitionColumns.map { partitionColumn =>
-          val equality = sparkSession.sessionState.conf.resolver
-          userSpecifiedSchema.flatMap(_.find(c => equality(c.name, partitionColumn))).orElse {
+          userSpecifiedSchema.flatMap(_.find(c => fieldEquality(c.name, partitionColumn))).orElse {
             val inferredPartitions = tempFileIndex.partitionSchema
-            val inferredOpt = inferredPartitions.find(p => equality(p.name, partitionColumn))
+            val inferredOpt = inferredPartitions.find(p => fieldEquality(p.name, partitionColumn))
             if (inferredOpt.isDefined) {
               logDebug(
                 s"""Type of partition column: $partitionColumn not found in specified schema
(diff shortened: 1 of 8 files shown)


maropu:
This query passed in the released spark-2.0.2, so it seems this regression is involved with SPARK-18510.

maropu:
@brkyvz @tdas Could you check this?

sql/core/src/test/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetPartitionDiscoverySuite.scala
```diff
@@ -969,4 +969,15 @@ class ParquetPartitionDiscoverySuite extends QueryTest with ParquetTest with Sha
...
     }
   }
+
+  test("SPARK-18108 Partition discovery fails with explicitly written long partitions") {
+    withTempPath { dir =>
+      val path = dir.getCanonicalPath
+
+      val df = Seq((1L, 2.0)).toDF("a", "b")
+      // partition explicitly written
+      df.write.parquet(s"$path/a=1")
```
brkyvz: @maropu I think you're doing the wrong thing. You're having the value `a=1` as part of your data schema and your partition directory, which is not allowed. I would say the correct behavior here would be to fail the reading case, because the value is both in the partition schema and the data schema, and it's possible that they may not be equal.
For example:
```scala
val df = Seq((1L, 2.0)).toDF("a", "b")
df.write.parquet(s"$path/a=2")
// this should fail because `a` is both part of the partitioning and schema
checkAnswer(spark.read.parquet(s"$path"), Seq(Row(1L, 2.0)))
```

brkyvz:
@maropu I wouldn't say this is a regression. I would say that this working for 2.0.2 was a bug in 2.0.2. If you want the column `a` to be interpreted as a `LongType` instead of `IntegerType`, you should provide the schema with `df.read.schema(schemaWhereAIsLong).parquet("path")`

brkyvz:
Or the thing that we should fix here is that if a partition column is found also as part of the dataSchema, to throw an exception.

maropu:
@brkyvz Thanks for your comment!  okay, I'll fix in a that way.

maropu:
Jenkins, retest this please.

maropu:
I'm looking into the failures.

maropu:
@brkyvz How about this fix?

liancheng:
@brkyvz @maropu Actually, we do allow users to create partitioned tables that allow data schema to contains (part of) the partition columns, and there are [test][1] [cases][2] for this use case.

This use case is mostly useful when you are trying to reorganize an existing dataset into a partitioned form. Say you have a JSON dataset containing all the tweets in 2016 and you'd like to partition it by date. By allowing the data schema to contain partitioned columns, you may simply put JSON files of the same date into the same directory. Otherwise, you'll have to run an ETL job to erase the date column from the dataset, which can be time-consuming.

As for the query @maropu mentioned in the PR description, the query itself is problematic, because it lacks a user-specified schema to override the data type of the partitioned column `a`. Ideally, partition discovery should be able to fill in the correct data type `LongType`, but it's impossible since the directory path doesn't really expose that information. That's why a user-specified schema is necessary.

This query works in 2.0.2 because of the bug @brkyvz fixed: Spark ignores the data types of partitioned columns specified in the user-specified schema. Now the bug is fixed and this query doesn't work as expected.

In short:

1. This isn't a regression, the original query itself is problematic.
2. For this PR, we can either just close it or try to provide a better error message in the read path (ask the user to provide a user-specified schema) when:

   - A partitioned columns `p` also appears in the data schema, and
   - The discovered data type of `p` is different from the data type specified in the data schema.

   An alternative is to override the discovered partition column data type using the one in the data schema if any. But I'd say this change is probably too risky at this moment for 2.1.

[1]: https://github.com/apache/spark/blob/c51c7725944d60738e2bac3e11f6aea74812905c/sql/hive/src/test/scala/org/apache/spark/sql/sources/ParquetHadoopFsRelationSuite.scala#L44-L66
[2]: https://github.com/apache/spark/blob/c51c7725944d60738e2bac3e11f6aea74812905c/sql/hive/src/test/scala/org/apache/spark/sql/hive/orc/OrcHadoopFsRelationSuite.scala#L43-L66

liancheng:
@maropu I tried your snippet (with minor modifications). It works for 1.6.0 instead of 2.0.2:

```scala
case class A(a: Long, b: Int)
val as = Seq(A(1, 2))
val path = "/tmp/part"
sqlContext.createDataFrame(as).write.mode("overwrite").parquet(s"$path/a=1/")
val df = sqlContext.read.parquet(path)
df.printSchema()
df.collect()
```

brkyvz:
I also made this query work on 2.1 branch by configuring `spark.sql.parquet.enableVectorizedReader = false`. It seems to be an issue with the VectorizedParquetReader.

liancheng:
My hunch is that we somehow passed a wrong requested schema containing the partition column down to the vectorized Parquet reader. IIRC, we prune partition columns from the data schema when generating the requested schema for the underlying reader since partition values are directly available in the directory path, there's no need to read and decode them from the physical file.

maropu:
This is not a bug in `VectorizedParquetReader ` as @liancheng  said, and the root cause is that wrongly inferred types are passed into the reader in `VectorizedParquetReader#initBatch` https://github.com/apache/spark/blob/master/sql/core/src/main/java/org/apache/spark/sql/execution/datasources/parquet/VectorizedParquetRecordReader.java#L176

yea, I know this functionality is helpful for skilful users, and, on the other hand, newbies easily could break query results via this interface as @brkyvz said. Therefore, in case of that we allow data schema to contain (part of) the partition columns, IMO it'd be better to alert that risk to users via `logWarning` or something.

BTW this is the original fix of this bug (Sorry, but I wrongly overrode and remove this commit ): https://github.com/apache/spark/compare/master...maropu:SPARK-18108-2. The fix of this commit is to fill in correct data types when data schema contains (part of) the partition columns.

brkyvz:
@maropu The reason I call it a bug with the VectorizedParquetReader is that all other data sources always replace the value read from the data with the partition value. The VectorizedReader also doesn't throw an exception if you try the following example:
```scala
case class A(a: Long, b: Int)
val as = Seq(A(1, 2))
val path = "/tmp/part"
sqlContext.createDataFrame(as).write.mode("overwrite").parquet(s"$path/a=1480617712537/")
val df = sqlContext.read.parquet(path)
df.printSchema()
df.collect()
```
returns
```
root
 |-- a: long (nullable = true)
 |-- b: integer (nullable = true)

df: org.apache.spark.sql.DataFrame = [a: bigint, b: int]
res7: Array[org.apache.spark.sql.Row] = Array([1480617712537,2])
```

brkyvz:
@maropu I think I found the simplest fix!
Would you mind changing:
https://github.com/apache/spark/blob/2ab8551e79e1655c406c358b21c0a1e719f498be/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala#L180
to something like:
```scala
val dataSchema = userSpecifiedSchema.orElse {
    format.inferSchema(
      sparkSession,
      caseInsensitiveOptions
      tempFileIndex.allFiles())
}.getOrElse {
  throw new AnalysisException(
    s"Unable to infer schema for $format. It must be specified manually.")
}
val dataWithoutPartitions = dataSchema.filterNot { field =>
  partitionSchema.exists(p => equality(p.name, field.name))
}
```
What we were missing was removing the partition columns from the data schema when we infer the format

maropu:
@brkyvz Thanks! Does the latest fix apply your suggestion?

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
```diff
@@ -132,13 +132,24 @@ case class DataSource(
       }.toArray
       new InMemoryFileIndex(sparkSession, globbedPaths, options, None)
     }
+
+    val dataSchema = userSpecifiedSchema.orElse {
```
brkyvz: I would still keep this below the `if (justPartitioning)` area, because otherwise everytime someone performs a `df.mode("append").saveAsTable()` we will perform schema inference.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
```diff
@@ -174,22 +185,18 @@ case class DataSource(
         StructType(partitionFields)
       }
     }
+
     if (justPartitioning) {
-      return (null, partitionSchema)
-    }
-    val dataSchema = userSpecifiedSchema.map { schema =>
```
brkyvz: I would keep the code here like I mentioned above, but keep your changes.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
```diff
@@ -174,22 +185,18 @@ case class DataSource(
...
-    }.orElse {
-      format.inferSchema(
-        sparkSession,
-        caseInsensitiveOptions,
-        tempFileIndex.allFiles())
-    }.getOrElse {
-      throw new AnalysisException(
-        s"Unable to infer schema for $format. It must be specified manually.")
+      (null, partitionSchema)
+    } else if (userSpecifiedSchema.isDefined) {
```
brkyvz: no need for this `else if`

sql/core/src/test/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetPartitionDiscoverySuite.scala
```diff
@@ -969,4 +969,15 @@ class ParquetPartitionDiscoverySuite extends QueryTest with ParquetTest with Sha
       ))
     }
   }
+
+  test("SPARK-18108 Partition discovery fails with explicitly written long partitions") {
+    withTempPath { dir =>
```
brkyvz: since this test is broken only when the vectorizedReader is on, I would also keep the conf here as enabled, just in case anyone changes the default implementation later.
so I would also wrap this as:
```scala
withSQLConf(SQLConf.PARQUET_VECTORIZED_READER_ENABLED.key -> "true") {
  withTempPath { dir =>
    ...
  }
}
```

brkyvz:
@maropu I would still keep the changes I proposed below L180 like I commented before. We don't need to use the inferred data type as the partition type

brkyvz:
It would be great to get this into 2.1!

maropu:
Thanks your review, @brkyvz ! I'm checking your comments now.

sql/core/src/test/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetPartitionDiscoverySuite.scala
maropu: fixed.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
maropu: yea, removed this.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
```diff
@@ -132,13 +132,24 @@ case class DataSource(
...
+
     val partitionSchema = if (partitionColumns.isEmpty && catalogTable.isEmpty) {
       // Try to infer partitioning, because no DataSource in the read path provides the partitioning
       // columns properly unless it is a Hive DataSource
       val resolved = tempFileIndex.partitionSchema.map { partitionField =>
         val equality = sparkSession.sessionState.conf.resolver
-        // SPARK-18510: try to get schema from userSpecifiedSchema, otherwise fallback to inferred
-        userSpecifiedSchema.flatMap(_.find(f => equality(f.name, partitionField.name))).getOrElse(
+        // SPARK-18510: try to get partition schema from data schema, otherwise fallback to inferred
+        dataSchema.find(f => equality(f.name, partitionField.name)).getOrElse(
```
maropu: reverted

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
maropu: reverted

maropu:
@brkyvz How should we handle this query? I think we have some overheads to check the constant value (`1480617712537`)  in path matches all the data in DataFrame and to throw an exception if illegal values detected.
```
case class A(a: Long, b: Int)
val as = Seq(A(1, 2))
val path = "/tmp/part"
sqlContext.createDataFrame(as).write.mode("overwrite").parquet(s"$path/a=1480617712537/")
val df = sqlContext.read.parquet(path)
df.printSchema()
df.collect()
```

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
```diff
@@ -132,14 +132,16 @@ case class DataSource(
       }.toArray
       new InMemoryFileIndex(sparkSession, globbedPaths, options, None)
     }
+
+    val fieldEquality = sparkSession.sessionState.conf.resolver
```
brkyvz: awesome! thanks!

sql/hive/src/test/scala/org/apache/spark/sql/hive/orc/OrcPartitionDiscoverySuite.scala
```diff
@@ -146,7 +146,7 @@ class OrcPartitionDiscoverySuite extends QueryTest with TestHiveSingleton with B
             i <- 1 to 10
             pi <- Seq(1, 2)
             ps <- Seq("foo", "bar")
-          } yield Row(i, pi, i.toString, ps))
+          } yield Row(i, i.toString, pi, ps))
```
brkyvz: Uh-oh, did we change some behavior here?

brkyvz:
@maropu Thank you so much. This LGTM! I'm just worried about the behavior changes in the ParquetPartition Suites. I'm going to take a quick look on why that may be happening.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
```diff
@@ -189,7 +189,15 @@ case class DataSource(
       throw new AnalysisException(
         s"Unable to infer schema for $format. It must be specified manually.")
     }
-    (dataSchema, partitionSchema)
+
+    // Override the fields of the partition schema if the data schema has the same field
+    val resolvedPartitionSchema = partitionSchema.map { partitionField =>
```
brkyvz: I don't think we need this. Otherwise, the value you return when `if (justPartitioning)` is inconsistent.
In a real world setting, as you provided in your example, for:
`case class A(a: Long, b: Int)`
if `a` is in fact a `Long` and is also the partitioning column, there will be examples of `a` being a `Long` among the partition columns, therefore, things should work. No-one should ever have more than 2 billion partition columns anyway. Spark wouldn't be able to resolve that many columns all in-memory right now anyway

brkyvz:
I would remove the part at: https://github.com/apache/spark/pull/16030/files#diff-7a6cb188d2ae31eb3347b5629a679cecR194

brkyvz:
@liancheng I have one more comment and this LGTM then. Do you have any feedback? What do you think about the behavior change in the Parquet schema reporting, i.e. when the data is both in the partition column and as part of the data Schema. The ordering in the row changes.

sql/hive/src/test/scala/org/apache/spark/sql/hive/orc/OrcPartitionDiscoverySuite.scala
maropu: It seems moving partition columns into the end is a correct behaviour by referring comments here: https://github.com/apache/spark/blame/master/sql/core/src/test/scala/org/apache/spark/sql/test/DataFrameReaderWriterSuite.scala#L590. Anything I miss?

brkyvz:
That is something I tried to standardize. I think we should standardize it,
otherwise it is very confusing. I think it's the right thing to do. I just
don't want people's production jobs failing in weird ways.

maropu:
@brkyvz yea, I think so, but the issue of column ordering seems to be out of scope in this original ticket. So, why don't we open a new jira and discuss there?

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
maropu: yea, removed

brkyvz:
Looks great! Thanks so much @maropu . @tdas or @liancheng would you mind taking a last look and possibly merging to 2.1?

maropu:
Jenkins, retest this please.

maropu:
The failure seems to be not related to this pr?

maropu:
Jenkins, retest this please.

brkyvz:
@maropu There seems to be a merge conflict. Should be pretty trivial to resolve. Mind taking a look when you can?

maropu:
okay, I'll do

liancheng:
@brkyvz I also worry about the behavior change. Let me check whether the original behavior is by design or by accident. If it is a bug from the very beginning, then we should just fix it in this PR.

brkyvz:
Even if it is a bug or not, I think we should have a consistent story for all of these. It's weird that the behavior is different for CatalogTables, for Vectorized and non-vectorized parquet reader, etc.

liancheng:
@brkyvz I agree that always moving all partitioned columns to the end of the schema is more consistent and intuitive. However, users may have ordinal-dependent code like this:

```scala
df.rdd.map { case Row(name: String, age: Int) =>
  ...
}
```

If we silently change the column order, these applications may break.

brkyvz:
yeah, that's what worries me. but does that merit keeping bad inconsistent behavior forever? Maybe a dev list question?

liancheng:
I'm checking whether the original behavior is consistent (do we always respect data schema column order when partition columns are included in data schema?). If not, I call it a bug and we just fix it. If yes, we may want to come up with a fix for this PR that preserves this old behavior and add a feature flag to deprecate it and enable the more preferrable behavior.

cloud-fan:
the new behavior LGTM, but I'm not sure if we still need to keep the old behavior

maropu:
What's the status of this pr? If I've something to do, plz let me know.

liancheng:
@maropu @brkyvz Sorry for the delay, I was blocked by some other tasks during the last a few days.

@cloud-fan and I just did some investigation and we think we came up with a minimal fix to this issue that also preserves the old behavior. Although we do agree that the new behavior makes more sense (moving all partition columns to the end of the output schema), it is a breaking change and we'd like to minimize risks for 2.1 release.

@cloud-fan will give a summary of the new fix and reasons behind it soon.

cloud-fan:
After an offline discussion with @liancheng , here is the result:

**Why does the test fail?**
1. We write a parquet file with schema `[a: long, b: int]` to path `/data/a=1`.
2. When read it back, we will infer the data schema as `[a: long, b: int]` and partition schema as `[a: int]`. 
3. In [`HadoopFsRelation.schema`](https://github.com/apache/spark/blob/branch-2.1/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/HadoopFsRelation.scala#L51-L56), we merge data schema and partition schema, and announce to users that the output schema will be `[a: long, b: int]`
4. In [`FileSourceScanExec`](https://github.com/apache/spark/blob/branch-2.1/sql/core/src/main/scala/org/apache/spark/sql/execution/DataSourceScanExec.scala#L252-L260), we build a parquet reader and tell it that the data schema is `[a: long, b: int]`, the required schema is `[b: int]`, the partition schema is `[a: int]`.
5. In [vectorized parquet read](https://github.com/apache/spark/blob/branch-2.1/sql/core/src/main/java/org/apache/spark/sql/execution/datasources/parquet/VectorizedParquetRecordReader.java#L181-L188), we read the data from parquet files according to the required schema: `[b: int]`, and append partition values according to the partition schema: `[a: int]`, so the schema of the physical row data is: `[b: int, a: int]`
6. In [`FileSourceStrategy`](https://github.com/apache/spark/blob/branch-2.1/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/FileSourceStrategy.scala#L74-L76), we mistakenly think the parquet scan will output data of schema `[b: int, a: long]`, and read the second column as long type. The vectorized parquet read can NOT read an int column as long and throw NPE.

**How to fix?**
The root cause is that, when data schema includes partition columns, how to determine the type of partition columns? Currently, at physical layer(the reader), we trust the partition schema, which is inferred from directory strings. At logical layer(`HadoopFsRelation`), we trust the partition columns inside of data schema. This inconsistency caused the bug.

w.r.t. the fact that we use partition values extracted from directory strings and ignore the partition columns inside physical data files, we think it's more reasonable to trust partition schema.

So the fix is simple, update `HadoopFsRelation.schema`, to respect the partition columns position in data schema, but also respect the partition columns type in partition schema.

maropu:
okay, I'll try to fix in that way, thanks!

liancheng:
Another thing that @cloud-fan and I agreed upon is that our current `DataFrameReader.schema()` interface method is insufficient. We may want to add

- `DataFrameReader.dataSchema()`, and
- `DataFrameReader.partitoinSchema()`

in the future to capture cases where data schema and partition schema overlap with each other.

The reason is that `FileFormat.buildReader()` requires a `dataSchema` argument, which should be the full schema of the underlying data files and may contain one or more partition columns. The reason is that many data sources (e.g. CSV and LibSVM) simply don't support column pruning, and you have to read out all physical columns first and then project out unnecessary ones.

Say we have a table with the following schema information:

- Data schema: `[x, y, p1]`
- Partition schema: `[p1, p2]`

Say a user tries to read this table with a user-specified schema `[x, y, p1, p2]`. Now we've no idea whether `p1` is part of the data schema or not since we skip schema inference when a user-specified schema is provided.

Although I haven't tested it extensively, I'd say it would be problematic if we try to use user-specified schema together with data sources without column pruning support (e.g. CSV) when the data schema overlaps with partition schema.

sql/core/src/test/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetPartitionDiscoverySuite.scala
```diff
@@ -969,4 +969,17 @@ class ParquetPartitionDiscoverySuite extends QueryTest with ParquetTest with Sha
       ))
     }
   }
+
+  test("SPARK-18108 Partition discovery fails with explicitly written long partitions") {
```
maropu: yea, thanks. I'm now working on this and I'll update soon.

maropu:
@cloud-fan Does the latest fix satisfy what you suggested?

maropu:
@liancheng As for `DataFrameReader.dataSchema()` and `DataFrameReader.partitoinSchema()`, did you mean we add new interfaces there for users to set user-defined data and partition schema, respectively?

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/HadoopFsRelation.scala
```diff
@@ -49,9 +49,12 @@ case class HadoopFsRelation(
   override def sqlContext: SQLContext = sparkSession.sqlContext
 
   val schema: StructType = {
-    val dataSchemaColumnNames = dataSchema.map(_.name.toLowerCase).toSet
-    StructType(dataSchema ++ partitionSchema.filterNot { column =>
-      dataSchemaColumnNames.contains(column.name.toLowerCase)
+    val equality = sparkSession.sessionState.conf.resolver
+    val overriddenDataSchema = dataSchema.map { dataField =>
```
cloud-fan: how about
```
val getColName: (StructField => String) = if (conf.caseSensitive) _.name else _.name.toLowerCase
val overlappedPartCols = mutable.Map.empty[String, StructField]
for {
  dataField <- dataSchema
  partitionField <- partitionSchema
  if getColName(dataField) == getColName(partitionField)
} overlappedPartCols += getColName(partitionField) -> partitionField

StructType(dataSchema.map(f => overlappedPartCols.getOrElse(getColName(f), f)) ++
  partitionSchema.filterNot(f => overlappedPartCols.contains(getColName(f))))
```

maropu: Why didn't you use `sparkSession.sessionState.conf.resolver`? Any reason I missed?
I just wrote this code with the same style with `DataSource#getOrInferFileFormatSchema` and is this a bad idea? Anyway, since the two patterns have the same output, both is okay to me.

cloud-fan: Because the current code will iterate `dataSchema` many times(depend on the number of partition columns), while my proposal only iterate it 2 times.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/HadoopFsRelation.scala
```diff
@@ -49,10 +51,16 @@ case class HadoopFsRelation(
...
   val schema: StructType = {
-    val dataSchemaColumnNames = dataSchema.map(_.name.toLowerCase).toSet
-    StructType(dataSchema ++ partitionSchema.filterNot { column =>
-      dataSchemaColumnNames.contains(column.name.toLowerCase)
-    })
+    val getColName: (StructField => String) =
+      if (sparkSession.sessionState.conf.caseSensitiveAnalysis) _.name else _.name.toLowerCase
+    val overlappedPartCols = mutable.Map.empty[String, StructField]
+    partitionSchema.foreach { partitionField =>
+      dataSchema.find(getColName(_) == getColName(partitionField)).map { overlappedCol =>
```
cloud-fan: a bit clearer:
```
if (dataSchema.find(getColName(_) == getColName(partitionField))) {
  overlappedPartCols += getColName(partitionField) -> partitionField
}
```

cloud-fan:
LGTM, pending jenkins. Can you also update the PR title and description? thanks!

maropu:
okay

maropu:
@cloud-fan okay, I updated the desc.

cloud-fan:
Can you also update the title? And the description has a mistake: the logical layer trusts the data schema to infer the type the overlapped partition columns, and, on the other hand, the physical layer trusts partition schema which is inferred from path string.

maropu:
oh, I wrongly wrote in an opposite way..., okay, fixed cc: @cloud-fan

cloud-fan:

maropu:
Thanks! I found another wired behaviour related to this issue. Is this expected? or, should we fix? 

```
scala> import org.apache.spark.sql.types._
scala> val schema = new StructType().add("a", LongType).add("b", IntegerType)
scala> case class A(a: Long, b: Int)
scala> val as = Seq(A(1, 2))
scala> spark.createDataFrame(as).write.parquet("/Users/maropu/Desktop/data/a=1/")
scala> val df = spark.read.schema(schema).parquet("/Users/maropu/Desktop/data/")
scala> df.printSchema
root
 |-- b: integer (nullable = true)
 |-- a: long (nullable = true)

scala> val df = spark.read.parquet("/Users/maropu/Desktop/data/")
scala> df.printSchema
root
 |-- a: integer (nullable = true)
 |-- b: integer (nullable = true)
```

cloud-fan:
This was the behavior change your PR proposed before, I think it makes sense, you can send a PR to fix it in Spark 2.2

maropu:
okay! I'll make a JIRA later and thanks!

PR #16031, opened 2016-11-28
[SPARK-18606][HISTORYSERVER]remove useless elements while searching

WangTaoTheTonic:

When we search applications in HistoryServer, it will include all contents between <td> tag, which including useless elemtns like "<span title...", "a href" and making results confused.
We should remove those to make it clear.

## How was this patch tested?

manual tests.

Before:
before

After:
after

### DIFF
--- a/core/src/main/resources/org/apache/spark/ui/static/historypage.js
+++ b/core/src/main/resources/org/apache/spark/ui/static/historypage.js
@@ -78,6 +78,12 @@ jQuery.extend( jQuery.fn.dataTableExt.oSort, {
     }
 } );
 
+jQuery.extend( jQuery.fn.dataTableExt.ofnSearch, {
+    "appid-numeric": function ( a ) {
+        return a.replace(/[\r\n]/g, " ").replace(/<.*?>/g, "");
+    }
+} );
+
 $(document).ajaxStop($.unblockUI);
 $(document).ajaxStart(function(){
     $.blockUI({ message: '<h3>Loading history summary...</h3>'});


core/src/main/resources/org/apache/spark/ui/static/historypage.js
```diff
@@ -78,6 +78,12 @@ jQuery.extend( jQuery.fn.dataTableExt.oSort, {
     }
 } );
 
+jQuery.extend( jQuery.fn.dataTableExt.ofnSearch, {
+    "appid-numeric": function ( a ) {
+        return a.replace(/[\r\n]/g, " ").replace(/<.*?>/g, "");
```
srowen: CC @ajbozarth for questions of JS style. There isn't a nice built-in way to strip HTML? maybe not. I imagine this does work correctly for well-formed HTML

WangTaoTheTonic: Refer to `jquery.dataTables.1.10.4.min.js`. I'd like to change it to better style if there's any :)

srowen: @ajbozarth are you saying this is an altnerative / better way to accomplish this? sorry I don't have much context to understand the comment. Or is the current approach OK?

ajbozarth: Sorry for the confusion, yes I was suggesting what I think is a better approach using jQuery DataTables confs

WangTaoTheTonic:
@srowen @ajbozarth Sorry for the delay.
I've tried the solution, but found it didn't work, as we already defined the type(`appid-numeric`) which will override value of `sType`.

I checked the implementation of `html` sType, it is same with what I proposed in this patch. Here I pasted snapshot:
226c5a24-d87f-480d-ad00-6c4a4b30560f
924f3273-ae3e-403d-ad9b-4cf00a9d50fc

ajbozarth:
Since you've checked and it doesn't work then this generally **LGTM**, but I still find it odd that jQuery can't do this for us.

Also, it's not important now, but we may want to eventually pull the actual `replace` function into a common file since other pages might need to use it as well when we switch them to DataTables.

WangTaoTheTonic:
I agree with you @ajbozarth. Since only one column uses `replace` function we can keep it same as now. If there is more data using this function in the future we will extract it to a simple file or something :)

srowen:
Merged to master

PR #16033, opened 2016-11-28
SPARK-18607 get a result on a percent of the tasks succeed

Ru-Xiang:

In this patch, we modify the codes corresponding to runApproximateJob so that we can get a result when the specified percent of tasks succeed.
In a production environment, 'long tail' is a common urgent problem. In practice, as long as we can get a specified percent of tasks' results, we can guarantee the final results. And this is a common requirement in the practice of machine learning algorithms.
## How was this patch tested?

We compile the codes by dev/make-distribution.sh, and deploy it on a cluster. and run a test codes reduce on the cluster, and we get the desired results.

https://issues.apache.org/jira/browse/SPARK-18607

### DIFF
--- a/core/src/main/scala/org/apache/spark/SparkContext.scala
+++ b/core/src/main/scala/org/apache/spark/SparkContext.scala
@@ -2003,6 +2003,27 @@ class SparkContext(config: SparkConf) extends Logging {
   }
 
   /**
+    * :: DeveloperApi ::
+    * Run a job that can return a partial result that get from parts of the tasks.
+    */
+  @DeveloperApi
+  def runPartialJob[T, U, R](
+      rdd: RDD[T],
+      func: (TaskContext, Iterator[T]) => U,
+      evaluator: ApproximateEvaluator[U, R],
+      percent: Double): PartialResult[R] = {
+    assertNotStopped()
+    val callSite = getCallSite
+    logInfo("Starting job: " + callSite.shortForm)
+    val start = System.nanoTime
+    val cleanedFunc = clean(func)
+    val result = dagScheduler.runPartialJob(rdd, cleanedFunc, evaluator, callSite, percent,
+      localProperties.get)
+    logInfo(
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/apache/spark/partial/ApproximateActionListener.scala
```diff
@@ -34,11 +34,13 @@ private[spark] class ApproximateActionListener[T, U, R](
     rdd: RDD[T],
     func: (TaskContext, Iterator[T]) => U,
     evaluator: ApproximateEvaluator[U, R],
-    timeout: Long)
+    timeout: Long = 1000*60*60*24*30*12,
```
srowen: You won't be able to do this because it changes the binary API. This is also mixing up two different semantics: wait for an amount of time, or wait for an amount of completion. I get the use case for both, but if this is intended to deal with skew, I think it's probably not the right solution in general. You need to deal with the skew more directly. It's not clear that ignoring 1 partition is the right thing to do, especially when it contains a lot of the data.

Ru-Xiang: This submission stems from our needs that drop out the tasks which are delayed by the hardware, scheduler or the networks but not  the skew data. In a production environment,  different machines may be congested at different times. While in some machine learning algorithms like SGD, we do not  need to wait all the tasks completion in each iteration for we can guarantee the convergence as long as the slow nodes are not always slow nodes.

rxin: @Ru-Xiang that's orthogonal to what @srowen was suggesting. The issue here is that the change is breaking binary compatibility for Spark applications.

Ru-Xiang:
@srowen @rxin  May you give some advice on this commit?

srowen:
Hm, I am just concerned about all the questions this raises. These jobs wouldn't be deterministic anymore, and this isn't really a way to deal with skew. I don't think this is something we shoudl add.

PR #14136, opened 2016-07-11
[SPARK-16282][SQL] Implement percentile SQL function.

jiangxb1987:
## What changes were proposed in this pull request?

Implement percentile SQL function. It computes the exact percentile(s) of expr at pc with range in [0, 1].

## How was this patch tested?

Add a new testsuite `PercentileSuite` to test percentile directly.
Updated related testcases in `ExpressionToSQLSuite`.

PR #13930, opened 2016-06-27
[SPARK-16228][SQL] HiveSessionCatalog should return `double`-param functions for decimal param lookups

dongjoon-hyun:
## What changes were proposed in this pull request?

This PR supports a fallback lookup by casting `DecimalType` into `DoubleType` for the external functions with `double`-type parameter.

**Reported Error Scenarios**

``` scala
scala> sql("select percentile(value, 0.5) from values 1,2,3 T(value)")
org.apache.spark.sql.AnalysisException: ... No matching method for class org.apache.hadoop.hive.ql.udf.UDAFPercentile with (int, decimal(38,18)). Possible choices: _FUNC_(bigint, array<double>)  _FUNC_(bigint, double)  ; line 1 pos 7

scala> sql("select percentile_approx(value, 0.5) from values 1.0,2.0,3.0 T(value)")
org.apache.spark.sql.AnalysisException: ... Only a float/double or float/double array argument is accepted as parameter 2, but decimal(38,18) was passed instead.; line 1 pos 7
```
## How was this patch tested?

Pass the Jenkins tests (including a new testcase).

PR #16034, opened 2016-11-28
[SPARK-18527][SQL] Convert decimal array to double array double for Hive UDAFPercentile

hvanhovell:
Hive UDAFPercentile takes a double or an array of doubles as its second argument. Spark typically passes a decimal or an array of decimals. This will causes a failure during function lookup for the percentile function.

We added a stop-gap measure in https://github.com/apache/spark/pull/13930 for doubles, by retrying the lookup with all decimal arguments casted to doubles. This PR extends this approach by adding the same logic for arrays of decimals.

Note that this is a temporary measure until https://github.com/apache/spark/pull/14136 is merged.

## How was this patch tested?
Added a test to `HiveUDFSuite`


hvanhovell:
We have merged the percentile PR. So this does not need to be added to branch-2.1. branch-2.0 still has this problems, should we add it there? cc @rxin @yhuai

rxin:
I'm ok not fixing this for branch-2.0.

hvanhovell:
closing :)

PR #16035, opened 2016-11-28
[SQL][minor] DESC should use 'Catalog' as partition provider

cloud-fan:

`CatalogTable` has a parameter named `tracksPartitionsInCatalog`, and in `CatalogTable.toString` we use `"Partition Provider: Catalog"` to represent it. This PR fixes `DESC TABLE` to make it consistent with `CatalogTable.toString`.

## How was this patch tested?

N/A


cloud-fan:
cc @ericl @yhuai @gatorsmile

rxin:
Merging in master/branch-2.1.

PR #16036, opened 2016-11-28
[SPARK-17732][SQL] Revert ALTER TABLE DROP PARTITION should support comparators

cloud-fan:

https://github.com/apache/spark/pull/15704 will fail if we use int literal in `DROP PARTITION`, and we have reverted it in branch-2.1.

This PR reverts it in master branch, and add a regression test for it, to make sure the master branch is healthy.

## How was this patch tested?

new regression test

### DIFF
--- a/sql/catalyst/src/main/antlr4/org/apache/spark/sql/catalyst/parser/SqlBase.g4
+++ b/sql/catalyst/src/main/antlr4/org/apache/spark/sql/catalyst/parser/SqlBase.g4
@@ -235,7 +235,11 @@ partitionSpecLocation
     ;
 
 partitionSpec
-    : PARTITION '(' expression (',' expression)* ')'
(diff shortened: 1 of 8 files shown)


cloud-fan:
cc @hvanhovell @dongjoon-hyun

hvanhovell:
LGTM pending jenkins.

@dongjoon-hyun lets merge the implementation of `ALTER TABLE DROP PARTITION` with your fix for it in https://github.com/apache/spark/pull/15987.

dongjoon-hyun:
I see. @cloud-fan and @hvanhovell .

hvanhovell:
LGTM. Merging to master (2.1 was already reverted). Thanks!

dongjoon-hyun:
@hvanhovell , could you give me some direction on #15987 before updating? 
The as-is direction of #15987 seems to be not proper.

PR #16037, opened 2016-11-28

AnthonyTruchet:

block in-place.

## How was this patch tested?

### DIFF
--- a/mllib/src/main/scala/org/apache/spark/mllib/optimization/LBFGS.scala
+++ b/mllib/src/main/scala/org/apache/spark/mllib/optimization/LBFGS.scala
@@ -241,16 +241,25 @@ object LBFGS extends Logging {
       val bcW = data.context.broadcast(w)
       val localGradient = gradient
 
-      val (gradientSum, lossSum) = data.treeAggregate((Vectors.zeros(n), 0.0))(
-          seqOp = (c, v) => (c, v) match { case ((grad, loss), (label, features)) =>
-            val l = localGradient.compute(
-              features, label, bcW.value, grad)
-            (grad, loss + l)
-          },
-          combOp = (c1, c2) => (c1, c2) match { case ((grad1, loss1), (grad2, loss2)) =>
+      // Given (current accumulated gradient, current loss) and (label, features)
+      // tuples, updates the current gradient and current loss
+      val seqOp = (c: (Vector, Double), v: (Double, Vector)) =>
+        (c, v) match {
+          case ((grad, loss), (label, features)) =>
+            val denseGrad = grad.toDense
+            val l = localGradient.compute(features, label, bcW.value, denseGrad)
+            (denseGrad, loss + l)
+        }
+
+      // Adds two (gradient, loss) tuples
+      val combOp = (c1: (Vector, Double), c2: (Vector, Double)) =>
+        (c1, c2) match { case ((grad1, loss1), (grad2, loss2)) =>
             axpy(1.0, grad2, grad1)
             (grad1, loss1 + loss2)
-          })
+       }
+
+      val zeroSparseVector = Vectors.sparse(n, Seq())
+      val (gradientSum, lossSum) = data.treeAggregate(zeroSparseVector, 0.0)(seqOp, combOp)
 
       /**
        * regVal is sum of weight squares if it's L2 updater;
--- a/mllib/src/main/scala/org/apache/spark/mllib/optimization/LBFGS.scala
+++ b/mllib/src/main/scala/org/apache/spark/mllib/optimization/LBFGS.scala
@@ -241,8 +241,6 @@ object LBFGS extends Logging {
       val bcW = data.context.broadcast(w)
       val localGradient = gradient
 
-      // Given (current accumulated gradient, current loss) and (label, features)
-      // tuples, updates the current gradient and current loss
       val seqOp = (c: (Vector, Double), v: (Double, Vector)) =>
         (c, v) match {
           case ((grad, loss), (label, features)) =>
@@ -251,7 +249,6 @@ object LBFGS extends Logging {
             (denseGrad, loss + l)
         }
 
-      // Adds two (gradient, loss) tuples
       val combOp = (c1: (Vector, Double), c2: (Vector, Double)) =>
         (c1, c2) match { case ((grad1, loss1), (grad2, loss2)) =>
           val denseGrad1 = grad1.toDense
@@ -261,7 +258,7 @@ object LBFGS extends Logging {
        }
 
       val zeroSparseVector = Vectors.sparse(n, Seq())
-      val (gradientSum, lossSum) = data.treeAggregate(zeroSparseVector, 0.0)(seqOp, combOp)
+      val (gradientSum, lossSum) = data.treeAggregate((zeroSparseVector, 0.0))(seqOp, combOp)
 
       /**
        * regVal is sum of weight squares if it's L2 updater;
--- a/mllib/src/main/scala/org/apache/spark/mllib/optimization/LBFGS.scala
+++ b/mllib/src/main/scala/org/apache/spark/mllib/optimization/LBFGS.scala
@@ -254,8 +254,10 @@ object LBFGS extends Logging {
       // Adds two (gradient, loss) tuples
       val combOp = (c1: (Vector, Double), c2: (Vector, Double)) =>
         (c1, c2) match { case ((grad1, loss1), (grad2, loss2)) =>
-            axpy(1.0, grad2, grad1)
-            (grad1, loss1 + loss2)
+          val denseGrad1 = grad1.toDense
+          val denseGrad2 = grad2.toDense
+          axpy(1.0, denseGrad2, denseGrad1)
+          (denseGrad1, loss1 + loss2)
        }
 
       val zeroSparseVector = Vectors.sparse(n, Seq())
--- a/mllib/src/test/scala/org/apache/spark/mllib/optimization/LBFGSSuite.scala
+++ b/mllib/src/test/scala/org/apache/spark/mllib/optimization/LBFGSSuite.scala
@@ -230,6 +230,25 @@ class LBFGSSuite extends SparkFunSuite with MLlibTestSparkContext with Matchers 
       (weightLBFGS(0) ~= weightGD(0) relTol 0.02) && (weightLBFGS(1) ~= weightGD(1) relTol 0.02),
       "The weight differences between LBFGS and GD should be within 2%.")
   }
+
+  test("SPARK-18471: LBFGS aggregator on empty partitions") {
+    val regParam = 0
+
+    val initialWeightsWithIntercept = Vectors.dense(0.0)
+    val convergenceTol = 1e-12
+    val numIterations = 1
+    val dataWithEmptyPartitions = sc.parallelize(Seq((1.0, Vectors.dense(2.0))), 2)
+
+    LBFGS.runLBFGS(
+      dataWithEmptyPartitions,
+      gradient,
+      simpleUpdater,
+      numCorrections,
+      convergenceTol,
+      numIterations,
+      regParam,
+      initialWeightsWithIntercept)
+  }
 }
 
 class LBFGSClusterSuite extends SparkFunSuite with LocalClusterSparkContext {


AnthonyTruchet:
See #15963 for first feed back on those changes.

srowen:
OK, this is the fourth pull request though (not counting a not-quite-related 5th). You don't need to open a new PR to push more changes and it adds to the difficulty in reviewing. This still doesn't incorporate suggestions from the last review.

srowen:
Following https://github.com/apache/spark/pull/16038 I suggest this proceed by making the zero value a sparse vector, and then making it dense in the seqOp immediately.

MLnick:
This is all a bit confusing - can we highlight which PR is actually to be reviewed?

srowen:
This should be the main PR @MLnick

MLnick:
Right ok. So I think the approach of making the zero vector sparse then calling `toDense` in `seqOp` as @srowen suggested makes most sense.

Currently the gradient vector *must* be dense in MLlib since both `axpy` and the logic for multinomial logreg requires it. So the thing that is initially serialized with the task should be tiny, and the call `toDense` for the first instance in each partition will essentially generate the dense zero vector. Thereafter it should be a no-op as the vector will be dense and `toDense` will just be a ref to the values array.

Can we see if this works:
```scala
      val zeroVector = Vectors.sparse(n, Seq())
      val (gradientSum, lossSum) = data.treeAggregate((zeroVector, 0.0))(
          seqOp = (c, v) => (c, v) match { case ((grad, loss), (label, features)) =>
            val denseGrad = grad.toDense
            val l = localGradient.compute(
              features, label, bcW.value, denseGrad)
            (denseGrad, loss + l)
          },
          combOp = (c1, c2) => (c1, c2) match { case ((grad1, loss1), (grad2, loss2)) =>
            axpy(1.0, grad2, grad1)
            (grad1, loss1 + loss2)
          })
```

MLnick:
What worries me more actually is that the initial vector when sent in the closure should be compressed. So why is this issue occurring? Is it a problem with serialization / compression? OR even after compression it is still too large? Would be good to understand that.

AnthonyTruchet:
Hello @MLnick,

Sorry that my push of a new version just crossed with your suggestion.

I'll test you suggestion. Thanks to for having written it, I didn't get that Sean's hint was this.

AnthonyTruchet:
@MLnick I have to add a transtyping denseGrad in the expression returned by seqOp from DenseVector to Vector. The only way I could find to do it with the API are: `Vectors.fromBreeze(denseGrad.asBreeze)` or `Vectors.dense(denseGrad.values)`: I'm a biut concerned with the potential recopy of a 128MB piece of memory...

MLnick:
ok to test

AnthonyTruchet:
Scala style checks fixed, and spurious cast removed.

srowen (approved):
Yes this LGTM. Let's make sure it passes tests. CC maybe @dbtsai FYI

AnthonyTruchet:
This seems very related to the change. I know nothing wrt the way PySpark interfaces Python and Scala and I'm surprised that changing an internal f the Scala lib causes this ? Ready to learn and help though :-)

```
File "/home/jenkins/workspace/SparkPullRequestBuilder/python/pyspark/mllib/classification.py", line 155, in __main__.LogisticRegressionModel
Failed example:
    mcm = LogisticRegressionWithLBFGS.train(data, iterations=10, numClasses=3)
Exception raised:
      File "/usr/lib64/python2.6/doctest.py", line 1253, in __run
        compileflags, 1) in test.globs
      File "<doctest __main__.LogisticRegressionModel[23]>", line 1, in <module>
        mcm = LogisticRegressionWithLBFGS.train(data, iterations=10, numClasses=3)
      [...]
    Py4JJavaError: An error occurred while calling o173.trainLogisticRegressionModelWithLBFGS.
     [...]
Caused by: java.lang.IllegalArgumentException: axpy only supports adding to a dense vector but got type class org.apache.spark.mllib.linalg.SparseVector.
    	at org.apache.spark.mllib.linalg.BLAS$.axpy(BLAS.scala:58)
    	at org.apache.spark.mllib.optimization.LBFGS$CostFun$$anonfun$2.apply(LBFGS.scala:257)
    	at org.apache.spark.mllib.optimization.LBFGS$CostFun$$anonfun$2.apply(LBFGS.scala:255)
```

mllib/src/main/scala/org/apache/spark/mllib/optimization/LBFGS.scala
```diff
@@ -241,16 +241,27 @@ object LBFGS extends Logging {
...
+            (denseGrad, loss + l)
+        }
+
+      // Adds two (gradient, loss) tuples
+      val combOp = (c1: (Vector, Double), c2: (Vector, Double)) =>
+        (c1, c2) match { case ((grad1, loss1), (grad2, loss2)) =>
+          val denseGrad1 = grad1.toDense
+          val denseGrad2 = grad2.toDense
+          axpy(1.0, denseGrad2, denseGrad1)
+          (grad1, loss1 + loss2)
```
MLnick: I think `grad1` here will need to be `denseGrad1`?

AnthonyTruchet: done, thanks for spotting.

AnthonyTruchet:
@MLnick there it is all tests passing :-)

srowen (approved):

MLnick:
By the way this same issue may also impact the `ml` optimizers that use L-BFGS. We should check the various gradient aggregators for `LogisticRegression`, `LinearRegression`, `MLP` etc. cc @sethah

MLnick:
I'm sure this will be net positive, and _shouldn't_ cause any regression. Still, we must be certain. @AnthonyTruchet can you provide for posterity the detailed test results for the vector sizes you mentioned? And perhaps also some results for some smaller sizes (since I imagine the benefit of this change for that scenario is quite small and we should just check there is no unexpected overhead or regression we've somehow missed from the `toDense` calls though I can't see that happening).

AnthonyTruchet:
L-BFGS is the only optimizer I've used so far. I'm not sure how much time I can free to take care of the other ones, but I'll try :-)

Regarding the bench, I'll check if we have archived the results, otherwise I'll relaunch it next week

srowen:
Yes I'm pretty OK with merging this. If you can dig up any results, that's all the better. Will check in with you next week.

mllib/src/main/scala/org/apache/spark/mllib/optimization/LBFGS.scala
```diff
@@ -241,16 +241,27 @@ object LBFGS extends Logging {
...
+      val combOp = (c1: (Vector, Double), c2: (Vector, Double)) =>
+        (c1, c2) match { case ((grad1, loss1), (grad2, loss2)) =>
+          val denseGrad1 = grad1.toDense
+          val denseGrad2 = grad2.toDense
+          axpy(1.0, denseGrad2, denseGrad1)
+          (denseGrad1, loss1 + loss2)
+       }
+
+      val zeroSparseVector = Vectors.sparse(n, Seq())
+      val (gradientSum, lossSum) = data.treeAggregate(zeroSparseVector, 0.0)(seqOp, combOp)
```
sethah: I'm slightly in favor of keeping the parentheses for the zero values. If you don't know the signature of `treeAggregate` and that the scala compiler evidently packs these into a tuple, you may be confused about what the second argument is. Not a strong preference.

mllib/src/main/scala/org/apache/spark/mllib/optimization/LBFGS.scala
```diff
@@ -241,16 +241,27 @@ object LBFGS extends Logging {
...
+          case ((grad, loss), (label, features)) =>
+            val denseGrad = grad.toDense
+            val l = localGradient.compute(features, label, bcW.value, denseGrad)
+            (denseGrad, loss + l)
+        }
+
+      // Adds two (gradient, loss) tuples
+      val combOp = (c1: (Vector, Double), c2: (Vector, Double)) =>
+        (c1, c2) match { case ((grad1, loss1), (grad2, loss2)) =>
+          val denseGrad1 = grad1.toDense
```
sethah: This is necessary because we may have empty partitions, right? Might be nice to add a test explicitly for this (it seems it failed in pyspark without it, but that was just a coincidence?) so someone doesn't remove these lines in the future.

mllib/src/main/scala/org/apache/spark/mllib/optimization/LBFGS.scala
```diff
@@ -241,16 +241,27 @@ object LBFGS extends Logging {
...
+      // tuples, updates the current gradient and current loss
+      val seqOp = (c: (Vector, Double), v: (Double, Vector)) =>
+        (c, v) match {
+          case ((grad, loss), (label, features)) =>
+            val denseGrad = grad.toDense
+            val l = localGradient.compute(features, label, bcW.value, denseGrad)
+            (denseGrad, loss + l)
+        }
+
+      // Adds two (gradient, loss) tuples
```
sethah: I don't think this comment is necessary. The semantics of a combine op should be well understood.

sethah:
@MLnick Yeah, this is likely a problem with all the ML aggregators as well. We can probably take care of it using lazy evaluation.

mllib/src/main/scala/org/apache/spark/mllib/optimization/LBFGS.scala
srowen: Meaning, when would the args ever not be dense? I agree, shouldn't be sparse at this stage, but doing this defensively seems fine since it's a no-op for dense.

srowen (changes requested):
Agree with @sethat's two comments regarding the comment and the arg to treeAggregate (that generates a warning I think). Then I think it's good to go.

mllib/src/main/scala/org/apache/spark/mllib/optimization/LBFGS.scala
AnthonyTruchet: OK

mllib/src/main/scala/org/apache/spark/mllib/optimization/LBFGS.scala
AnthonyTruchet: Yes. actually missing the handling of dense vector was the cause for the PySpark UTest failure we observed.

mllib/src/main/scala/org/apache/spark/mllib/optimization/LBFGS.scala
AnthonyTruchet: You mean `data.treeAggregate((zeroSparseVector, 0.0))`, don't you ?

AnthonyTruchet:
@sethah Agreed, that's why I uggested to add a dedicated treeAggregate wrapper to MLlin Utils which would take care of that without fiddling with sparsity for each seqOP and comboOp. See closed  #16078 for the idea...

mllib/src/main/scala/org/apache/spark/mllib/optimization/LBFGS.scala
sethah: Yes :)

srowen (approved):

mllib/src/main/scala/org/apache/spark/mllib/optimization/LBFGS.scala
```diff
@@ -241,16 +241,24 @@ object LBFGS extends Logging {
...
+        (c, v) match {
+          case ((grad, loss), (label, features)) =>
+            val denseGrad = grad.toDense
+            val l = localGradient.compute(features, label, bcW.value, denseGrad)
+            (denseGrad, loss + l)
+        }
+
+      val combOp = (c1: (Vector, Double), c2: (Vector, Double)) =>
+        (c1, c2) match { case ((grad1, loss1), (grad2, loss2)) =>
+          val denseGrad1 = grad1.toDense
```
sethah: I am still pretty strongly in favor of adding a test case explicitly for this. Just make an RDD with at least one empty partition, and be sure that LBFGS will run on it.

AnthonyTruchet: I'm sorry, I have no clue how to generate a non-empty RDD with an empty partition. Can you please hint me at some entry points so that I can contribute the UTest you request ?

srowen: What about just parallelizing n elements into more than n partitions? at least one partition must be empty.

srowen: @AnthonyTruchet does that let you add a quick test case? then I think this can be merged.

AnthonyTruchet: I'll try to work on this in the next two days. I will not be able to relaunch the actual benchmark we did weeks ago, our internal codebase has changed too much.

AnthonyTruchet: I did thanks and I merged it. I was just comming back to this contrib when I saw it.

srowen (approved):

sethah:
LGTM

sethah (approved):

srowen:
Merged to master

AnthonyTruchet:
Thanks for keeping up with this merge request, I've learned a lot wrt the contribution process and good practice, and next contrib will hopefully be much more straightforward. Thanks to the Spark commiter team for this great piece of software :-)

dbtsai:
Sorry for late review. Just come back to US. LGTM too! Thanks.

PR #16040, opened 2016-11-28
[SPARK-18612][MLLIB] Delete broadcasted variable in LBFGS CostFun

AnthonyTruchet:

Fix a broadcasted variable leak occurring at each invocation of CostFun in L-BFGS.

## How was this patch tested?

UTests + check that fixed fatal memory consumption on Criteo's use cases.

This contribution is made on behalf of Criteo S.A.
(http://labs.criteo.com/) under the terms of the Apache v2 License.

### DIFF
--- a/mllib/src/main/scala/org/apache/spark/mllib/optimization/LBFGS.scala
+++ b/mllib/src/main/scala/org/apache/spark/mllib/optimization/LBFGS.scala
@@ -253,6 +253,9 @@ object LBFGS extends Logging {
             (grad1, loss1 + loss2)
           })
 
+      // broadcasted model is not needed anymore
+      bcW.destroy()
+
       /**
        * regVal is sum of weight squares if it's L2 updater;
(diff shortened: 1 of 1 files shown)


srowen (approved):
OK, would be good if you can look for similar instances and fix them in one pass here.

srowen:
Jenkins test this please

AnthonyTruchet:
Agreed, and we might also consider making the broadcasted variable compatible with some Automatic Resource Management to avoid such leak (like https://github.com/jsuereth/scala-arm).

srowen:
If that's just a Scala version of try-with-resources, it is unfortunately insufficient. It's not clear when the broadcast is actually no longer used because it may be used in computations that are computed lazily, or recomputed when something fails. Here the scope of the broadcast is clear though.

srowen:
Merged to master/2.1

PR #16044, opened 2016-11-28
[Spark-18614][SQL] Incorrect predicate pushdown from ExistenceJoin

nsyca:

ExistenceJoin should be treated the same as LeftOuter and LeftAnti, not InnerLike and LeftSemi. This is not currently exposed because the rewrite of [NOT] EXISTS OR ... to ExistenceJoin happens in rule RewritePredicateSubquery, which is in a separate rule set and placed after the rule PushPredicateThroughJoin. During the transformation in the rule PushPredicateThroughJoin, an ExistenceJoin never exists.

The semantics of ExistenceJoin says we need to preserve all the rows from the left table through the join operation as if it is a regular LeftOuter join. The ExistenceJoin augments the LeftOuter operation with a new column called exists, set to true when the join condition in the ON clause is true and false otherwise. The filter of any rows will happen in the Filter operation above the ExistenceJoin.

Example:

A(c1, c2): { (1, 1), (1, 2) }
// B can be any value as it is irrelevant in this example
B(c1): { (NULL) }

select A.*
from   A
where  exists (select 1 from B where A.c1 = A.c2)
       or A.c2=2

In this example, the correct result is all the rows from A. If the pattern ExistenceJoin around line 935 in Optimizer.scala is indeed active, the code will push down the predicate A.c1 = A.c2 to be a Filter on relation A, which will incorrectly filter the row (1,2) from A.

## How was this patch tested?

Since this is not an exposed case, no new test cases is added. The scenario is discovered via a code review of another PR and confirmed to be valid with peer.


hvanhovell:
@nsyca this was not uncovered because we didn't add tests, could you add one? `org.apache.spark.sql.catalyst.optimizer.FilterPushdownSuite` seems a good place (they could be similar to the ones for Left Anti).

nsyca:
@hvanhovell I can add a new test case but I want to reiterate that this is not an exposed case because `ExistenceJoin` does not exist in the course of the `PushPredicateThroughJoin` rule. The `ExistenceJoin` is rewritten at the `RewritePredicateSubquery` rule, which is currently placed at the very last batch of the Optimizer phase, after the batch that have `PushPredicateThroughJoin` rule.

hvanhovell:
I know it is currently not exposed due to the way we optimize plans. However this may change, and it is better to be safe than sorry. Also note that we typically only test the rule at hand in optimizer unit tests, so it should be relatively straightforward to create such a test case, by explicitly creating a join using `ExistenceJoin`.

nsyca:
Hmm. `ExistenceJoin` is not exposed via `DataFrame` APIs. Let me study the way `org.apache.spark.sql.execution.joins.ExistenceJoinSuite` was written then I will add a new test case.

gatorsmile:
You just need to add test cases to `FilterPushdownSuite`, like what @hvanhovell did: https://github.com/apache/spark/pull/16039/files#diff-acd51a263163be1e40af0b8b41f2b073

hvanhovell:
You can just write an optimizer test case. You can for example add this to the `FilterPushdownSuite`:
```scala
  test("joins: only push down join conditions to the right of an existence join") {
    val x = testRelation.subquery('x)
    val y = testRelation.subquery('y)
    val filerVal = 'val.boolean
    val originalQuery =
      x.join(y,
        ExistenceJoin(filerVal),
        Some("x.b".attr === "y.b".attr && "y.a".attr > 10 && "x.a".attr > 10)).analyze
    val optimized = Optimize.execute(originalQuery)
    val correctAnswer =
      x.join(
        y.where("y.a".attr > 10),
        ExistenceJoin(filerVal),
        Some("x.b".attr === "y.b".attr && "x.a".attr > 10))
        .analyze
    comparePlans(optimized, analysis.EliminateSubqueryAliases(correctAnswer))
  }
```
This probably passes.

nsyca:
@hvanhovell Thank you for the detailed instruction on the test case.
@gatorsmile Also thanks for the directive.

In addition to the white box testing on the Optimized Logical Plan, I also added two black box test cases in `org.apache.spark.sql.JoinSuite`. Hopefully it's a right place to put.

This is just to share my experience: The good thing about writing white box test cases, such as doing plan comparison as in `org.apache.spark.sql.catalyst.optimizer._`, are good in verifying internal behaviour, such as whether an optimization does happen or not. It may also catch problems that are not exposed externally. However, the down side is writing white box test cases are more time-consuming and more sensitive to future code changes. It could create a huge cost of test cases maintenance in the future.

sql/core/src/test/scala/org/apache/spark/sql/JoinSuite.scala
```diff
@@ -575,6 +575,24 @@ class JoinSuite extends QueryTest with SharedSQLContext {
         Row(3, 2) :: Nil)
   }
 
+  test("predicate push down in left anti join") {
+    val df = sql("SELECT l.a, l.b FROM testData2 l LEFT ANTI JOIN testData r ON l.a = l.b")
```
gatorsmile: This can be added to `anti-join.sql`

hvanhovell: +1

nsyca: Ah. This is the place I was looking for.. putting everything in SQL. Since this is the topic of predicate pushdown, I'd like to propose to put both of my test cases here and change the file name to `pred-pushdown.sql`. How can I run this collection of tests and verify the results?

dongjoon-hyun: If you add that file into `spark/sql/core/src/test/resources/sql-tests/inputs`, the result can be generated by the following. Please refer `SQLQueryTestSuite.scala`.
```
SPARK_GENERATE_GOLDEN_FILES=1 build/sbt "sql/test-only *SQLQueryTestSuite"
```
And you can run the test by the following.
```
build/sbt "sql/test-only *SQLQueryTestSuite"
```

hvanhovell: Nice, I was about to type that.

nsyca: @dongjoon-hyun @hvanhovell Thanks!

hvanhovell:
LGTM - pending jenkins

gatorsmile:
LGTM - pending jenkins

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/Optimizer.scala
```diff
@@ -932,7 +932,7 @@ object PushPredicateThroughJoin extends Rule[LogicalPlan] with PredicateHelper {
         split(joinCondition.map(splitConjunctivePredicates).getOrElse(Nil), left, right)
 
       joinType match {
-        case _: InnerLike |  LeftSemi | ExistenceJoin(_) =>
+        case _: InnerLike |  LeftSemi =>
```
gatorsmile: Nit: `|  LeftSemi` -> `| LeftSemi`. There is an extra space we should remove.

nsyca: I will make that change as part of my next PR.

hvanhovell:

PR #16046, opened 2016-11-28
[SPARK-18582][SQL] Whitelist LogicalPlan operators allowed in correlated subqueries

nsyca:

This fix puts an explicit list of operators that Spark supports for correlated subqueries.

## How was this patch tested?

Run sql/test, catalyst/test and add a new test case on Generate.


nsyca:
We are reviewing existing test cases on subqueries, especially correlated subqueries, and will open a new JIRA, as part of the umbrella JIRA SPARK-18455, to add new test cases to extend the coverage in this area.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
```diff
@@ -1077,10 +1077,54 @@ class Analyzer(
...
+        // Any operators that are not in the above list are allowed
+        // in a correlated subquery only if they are not on a correlation path.
+        // In other word, these operators are allowed only under a correlation point.
+        //
+        // A correlation path is defined as the sub-tree of all the operators that
+        // are on the path from the operator hosting the correlated expressions
+        // up to the operator producing the correlated values.
+
+        // Category 1:
+        // Leaf node can be anywhere in a correlated subquery.
```
hvanhovell: Nit: `LeafNode` can only be at the leaves of a tree. So `... can be anywhere...` is stretching it.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
```diff
@@ -1120,47 +1173,54 @@ class Analyzer(
...
-          failOnOuterReference(j)
-          failOnOuterReferenceInSubTree(right, "a LEFT (OUTER) JOIN")
+
+        // Join can host correlated expressions.
+        case j @ Join(left, right, joinType, _) =>
+          joinType match {
+            // Inner join, like Filter, can be anywhere.
+            // LeftSemi is a special case of Inner join which returns
+            // only the first matched row to the right table.
+            case _: InnerLike | LeftSemi =>
```
hvanhovell: We cannot pull-out a predicate which is defined in the right side of a `LeftSemi` join.  `LeftSemi` join should be treated like all a `LeftOuter` join.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
```diff
@@ -1120,47 +1173,54 @@ class Analyzer(
...
-          failOnOuterReferenceInSubTree(s, "a TABLESAMPLE")
-          s
+
+        // Generator with join=true, i.e., expressed with
+        // LATERAL VIEW [OUTER], similar to inner join,
+        // allows to have correlation under it
+        // but must not host any outer references.
+        // Note:
+        // Generator with join=false is treated as Category 4.
+        case p @ Generate(generator, join, _, _, _, _) if (join) =>
```
hvanhovell: Pattern match directly on `join=true`, in this case: `case p @ Generate(generator, true, _, _, _, _) =>`

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
```diff
@@ -1120,47 +1173,54 @@ class Analyzer(
...
+          if (containsOuter(generator)) {
```
hvanhovell: I am not sure what is going on here. Why only check all the expressions in the operator, when the generator contains an outer reference? Generate only has one expression, the generator, so I think you can safely call `failOnOuterReference(p)` directly.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
hvanhovell: Is generate part of group 2 or group 3? I would say group 2?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
```diff
@@ -1011,24 +1011,24 @@ class Analyzer(
     private def pullOutCorrelatedPredicates(sub: LogicalPlan): (LogicalPlan, Seq[Expression]) = {
       val predicateMap = scala.collection.mutable.Map.empty[LogicalPlan, Seq[Expression]]
 
-      /** Make sure a plans' subtree does not contain a tagged predicate. */
-      def failOnOuterReferenceInSubTree(p: LogicalPlan, msg: String): Unit = {
+      // Make sure a plan's subtree does not contain outer references
+      def failOnOuterReferenceInSubTree(p: LogicalPlan): Unit = {
         if (p.collect(predicateMap).nonEmpty) {
```
hvanhovell: Lets change this line into `p.collectFirst(predicateMap).nonEmpty` that is a little more efficient.

hvanhovell (changes requested):
This looks pretty good overall. I left a few minor comments.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
nsyca: Agreed. I will make the change in the next PR.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
```diff
@@ -1077,10 +1077,54 @@ class Analyzer(
...
+
+        // Category 1:
+        // Leaf node can be anywhere in a correlated subquery.
+        case n: LeafNode =>
+          n
+        // Category 2:
+        // These operators can be anywhere in a correlated subquery.
+        // so long as they do not host outer references in the operators.
+        // SubqueryAlias can be anywhere in a correlated subquery.
+        case p: SubqueryAlias =>
```
nsyca: I will put the code for these 4 operators in a comment to indicate we have thought through. Note Distinct, in the future, could be on a subset of the columns of its child's output and might contain expressions. An example is if we have (C1, C2) as output of Distinct's child but through some analysis we know that the relationship of C1 -> C2 is N : 1, we can optimize the Distinct operation to hash/sort only on C1.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
```diff
@@ -1110,6 +1157,12 @@ class Analyzer(
           } else {
             p
           }
+
+        // Aggregate cannot host any correlated expressions
+        // It can be on a correlation path if the correlation has
+        // only equality correlated predicates.
+        // It cannot be on a correlation path if the correlation has
```
nsyca: Will change.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
nsyca: You were doing too many code reviews yesterday/today. ;-) In your PR for SPARK-18597, you placed `LeftSemi` in the same category as `InnerJoin`.

The Left in `LeftSemi` does not mean it's a left outer join. It means a "semi" match from the "left" table of an inner join, returning only the first match of the right table (and ignoring subsequent matches). This is why I do not like the terminology. It's confusing. I prefer to call it a (left) early-out (inner) join, or simply early-out join.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
nsyca: Right, the `if (join)` is too verbose. I still need to sharpen my Scala coding skill. More concise.

nsyca: `Generate` with join=true is group 2. The reason I placed in group 3 is because the code handles two different cases when join=true and join=false.

By the way, how to construct an SQL with join=false? I could not find an example in the existing test cases.

nsyca: I interpret `Generate` with join=true just like an inner join with the child as the left table and the `generator` as the right table, which may host outer references from the child of `Generate`. This is sometimes called the side-way information passing. Currently `generator` is an expression but it could really be modelled as a (sub)plan, which can contain deep correlation -- where correlated point is at a descendant operator of the root operator of the subplan. In fact, `Generate` with join=true is just a syntactic sugar of a `Join`.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
nsyca: Right. I was debating on that. I was thinking of

````
if (containsOuter(generator)) {
  failAnalysis( ... )
}
````

I will make the change.

File: sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala (hunk 1 above)
Comment (nsyca) [inline sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala]: With the addition of `SubqueryAlias, Distinct, Repartition` and `BroadcastHint` in group 1 (see the next comment), I will rephrase the comment here.

File: sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala (hunk 2 above)
Comment (hvanhovell) [inline sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala]: I have to admit that have been reviewing a lot of PRs. However I am quite sure that you cannot define a correlated predicate in the plan on the  right hand side of a `LEFT SEMI`/`early-out join` because we only output the column of the plan on the left hand side. For example:
```sql
select *
from   tbl_a
where  exists (select 1
               from tbl_b
               left semi join( select id
                               from tbl_c
                               where tbl_c.id = tbl_a.id) c
                on c.id = tbl_b.id)
```
In this example we could not move the correlated predicate ` tbl_c.id = tbl_a.id` because the Left Semi join does not output `cid`. BTW: In this case it would actually be OK to convert the Left Semi join into an Inner join.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
hvanhovell: If you only use the columns generated by generator, the optimizer will set `join=false`. There is no way to do this in SQL.

hvanhovell: Generate is much like a join. You are right to envisage a scenario in which we would allow a correlated predicate inside a `Generate` expression, and I would be happy to add such cases to the white list when that time comes :)... but for now I think we should not allow them just yet.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
nsyca: The reason I write this lengthy response here is to convince you that we should leave `LeftSemi` in the same group as `InnerJoin`. Please bear with me.

The example you gave here demonstrates a limitation of the subquery supported in Spark today. We should plan to be able to handle this case of deep correlation in the future:

````sql
select  *
from    t
where   exists (select 1
                from   t2
                where  t2.c1=t1.c1
                and    exists (select 1
                               from   t3
                               where  t3.c2=t1.c1))
````

And if we do, then we will need to allow LeftSemi to output the columns from the right table.

One way to imagine a use case of LeftSemi is if we have a look up join where the join predicate forms a N:1 relationship, just like between a foreign key and its primary key. The join is effectively a LeftSemi that is guaranteed we only need to find the first matched row and move on to the next row of the left table (just like a hash join that we need to probe the first matched and stop early seeking the next matched in the hash chain). From a run-time viewpoint, a LeftSemi is (almost, more on this later) better than a regular InnerJoin that it does not need to probe for the next matched rows regardless of the chosen join methods: nested-loop join, sort-merge join, or hash join. A LeftSemi, however, dictates which tables can be the left and the right. As the name implies, the left table of a LeftSemi needs to be the N-side of the N:1 join. So in the case that N-side is the smaller table, it could be better to do the regular inner join with 1-side as the left table then perform a compensation on top of the join to remove the duplicate matched rows. Having said that, we can also implement a RightSemi join in the runtime layer so that we can pick any join, LeftSemi, RightSemi, or InnerJoin with a compensation, based on the cost.

Until we wade into the CBO, the whole planning business would be an interesting area, isn't it?

nsyca:
This latest push `1d32958` addresses all the review comments except the placement of `LeftSemi`, and includes the removal of an extra space in `Optimizer.scala`.

nsyca:
@hvanhovell On the placement of `LeftSemi`, have you got a chance to read my response? Please share your thought. How are we going to implement the support of deep correlation like the examples you and I raised?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
hvanhovell: It makes sense to extract predicates from the right hand side of a `LeftSemi` join. My problem with this is far more practical. `Join` with a `LeftSemi` join type does not output any right hand side attributes (see: https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/basicLogicalOperators.scala#L273), the plan **breaks** are as soon as you extract a correlated predicate from the right hand side of the join and try to rewrite the tree using that predicate. That is all.

It would - fortunately - be quite simple to support this. Just rewrite every LeftSemi join with underlying predicates into an Inner join. I am not entirely sure if we should support this, technically the query is incorrect. Lets defer this to a follow-up PR.

The example you give is different. I do think we should support that. Please note that the example you give will not have any left semi joins during analysis, the left semi joins are introduced during optimization; this makes it relatively straight forward to detect such a nested case.

nsyca: Got your point now. As it is today, pulling up the correlated pred from the right operand of `LeftSemi` will break.

hvanhovell:
LGTM - pending Jenkins.

hvanhovell:
retest this please

nsyca:
Thank you. I was reading the test log just to see if my code might cause the failure. It does not look like it does.

hvanhovell:
Merging to master/2.1/2.0. Thanks!

PR #16049, opened 2016-11-29
[SPARK-16282][SQL] Follow-up: remove "percentile" from temp function detection after implementing it natively

lins05:

In #15764 we added a mechanism to detect if a function is temporary or not. Hive functions are treated as non-temporary. Of the three hive functions, now "percentile" has been implemented natively, and "hash" has been removed. So we should update the list.

## How was this patch tested?

Unit tests.


rxin:
LGTM pending Jenkins.

gatorsmile:
Merging to master/2.1! Thanks!

PR #16050, opened 2016-11-29
[SPARK-18615][DOCS] Switch to multi-line doc to avoid a genjavadoc bug for backticks

HyukjinKwon:

Currently, single line comment does not mark down backticks to `<code>..</code>` but prints as they are (`` `..` ``). For example, the line below:

```scala
/** Return an RDD with the pairs from `this` whose keys are not in `other`. */
```

So, we could work around this as below:

```scala
/**
 * Return an RDD with the pairs from `this` whose keys are not in `other`.
 */
```

- javadoc

  - **Before**
    

  - **After**
    

- scaladoc (this one looks fine either way)

  - **Before**
    

  - **After**
    

I suspect this is related with SPARK-16153 and genjavadoc issue in ` typesafehub/genjavadoc#85`.

## How was this patch tested?

I found them via 

```
grep -r "\/\*\*.*\`" . | grep .scala
```` 

and then checked if each is in the public API documentation with manually built docs (`jekyll build`) with Java 7.

### DIFF
--- a/core/src/main/scala/org/apache/spark/SparkConf.scala
+++ b/core/src/main/scala/org/apache/spark/SparkConf.scala
@@ -378,7 +378,9 @@ class SparkConf(loadDefaults: Boolean) extends Cloneable with Logging with Seria
     settings.entrySet().asScala.map(x => (x.getKey, x.getValue)).toArray
   }
 
-  /** Get all parameters that start with `prefix` */
+  /**
+   * Get all parameters that start with `prefix`
+   */
   def getAllWithPrefix(prefix: String): Array[(String, String)] = {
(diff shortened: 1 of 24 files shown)


HyukjinKwon:
I guess this might not cause conflicts with https://github.com/apache/spark/pull/16013.

HyukjinKwon:
cc @srowen do you mind if I ask to take a look please?

srowen (approved):
Weird, OK. Looks fine. Let's also get this into 2.1 along with your other big doc cleanups.

srowen:
Merged to master/2.1

PR #16052, opened 2016-11-29
[SPARK-18617][CORE][STREAMING] Close "kryo auto pick" feature for Spark Streaming

uncleGen:

#15992 provided a solution to fix the bug, i.e. **receiver data can not be deserialized properly**. As @zsxwing said, it is a critical bug, but we should not break APIs between maintenance releases. It may be a rational choice to close auto pick kryo serializer for Spark Streaming in the first step. I will continue #15992 to optimize the solution.

## How was this patch tested?

existing ut

### DIFF
--- a/core/src/main/scala/org/apache/spark/serializer/SerializerManager.scala
+++ b/core/src/main/scala/org/apache/spark/serializer/SerializerManager.scala
@@ -77,8 +77,8 @@ private[spark] class SerializerManager(defaultSerializer: Serializer, conf: Spar
     primitiveAndPrimitiveArrayClassTags.contains(ct) || ct == stringClassTag
   }
 
-  def getSerializer(ct: ClassTag[_]): Serializer = {
-    if (canUseKryo(ct)) {
+  def getSerializer(ct: ClassTag[_], autoPick: Boolean): Serializer = {
+    if (autoPick && canUseKryo(ct)) {
       kryoSerializer
     } else {
       defaultSerializer
@@ -155,7 +155,12 @@ private[spark] class SerializerManager(defaultSerializer: Serializer, conf: Spar
       outputStream: OutputStream,
       values: Iterator[T]): Unit = {
     val byteStream = new BufferedOutputStream(outputStream)
-    val ser = getSerializer(implicitly[ClassTag[T]]).newInstance()
+    val ser = blockId match {
+      case a: StreamBlockId =>
+        getSerializer(implicitly[ClassTag[T]], autoPick = false).newInstance()
+      case _ =>
+        getSerializer(implicitly[ClassTag[T]], autoPick = true).newInstance()
+    }
     ser.serializeStream(wrapStream(blockId, byteStream)).writeAll(values).close()
   }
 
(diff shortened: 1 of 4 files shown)


uncleGen:
cc @zsxwing

core/src/main/scala/org/apache/spark/serializer/SerializerManager.scala
```diff
@@ -77,8 +77,8 @@ private[spark] class SerializerManager(defaultSerializer: Serializer, conf: Spar
     primitiveAndPrimitiveArrayClassTags.contains(ct) || ct == stringClassTag
   }
 
-  def getSerializer(ct: ClassTag[_]): Serializer = {
-    if (canUseKryo(ct)) {
+  def getSerializer(ct: ClassTag[_], autoPick: Boolean): Serializer = {
```
rxin: document what this means

core/src/main/scala/org/apache/spark/serializer/SerializerManager.scala
```diff
@@ -155,7 +155,12 @@ private[spark] class SerializerManager(defaultSerializer: Serializer, conf: Spar
       outputStream: OutputStream,
       values: Iterator[T]): Unit = {
     val byteStream = new BufferedOutputStream(outputStream)
-    val ser = getSerializer(implicitly[ClassTag[T]]).newInstance()
+    val ser = blockId match {
```
srowen: Is this simpler as ...

```
val autoPick = blockId.isInstanceOf[StreamBlockId]
val ser = getSerializer(implicitly[ClassTag[T]], autoPick).newInstance()
```

to avoid duplication?

uncleGen: Ah, yours is better.

zsxwing:
@uncleGen Could you add the following test to StreamingContextSuite? Otherwise, LGTM.
```Scala
  test("SPARK-18560 Receiver data should be deserialized properly.") {
    // Start a two nodes cluster, so receiver will use one node, and Spark jobs will use the
    // other one. Then Spark jobs need to fetch remote blocks and it will trigger SPARK-18560.
    val conf = new SparkConf().setMaster("local-cluster[2,1,1024]").setAppName(appName)
    sc = new SparkContext(conf)
    ssc = new StreamingContext(sc, Milliseconds(100))
    val input = ssc.receiverStream(new TestReceiver)
    input.count().foreachRDD { rdd =>
      // Make sure we can read from BlockRDD
      if (rdd.collect().headOption.getOrElse(0L) > 0) {
        // Stop StreamingContext to unblock "awaitTerminationOrTimeout"
        new Thread() {
          setDaemon(true)
          override def run(): Unit = {
            ssc.stop(stopSparkContext = true, stopGracefully = true)
          }
        }.start()
      }
    }
    ssc.start()
    ssc.awaitTerminationOrTimeout(60000)
  }
```

uncleGen:
OK

rxin:
Merging in master/branch-2.1.

rxin:
This doesn't merge cleanly into branch-2.0. @uncleGen can you submit a pull request for branch-2.0?

streaming/src/test/scala/org/apache/spark/streaming/StreamingContextSuite.scala
```diff
@@ -869,6 +891,31 @@ object TestReceiver {
   val counter = new AtomicInteger(1)
 }
 
+class FakeByteArrayReceiver extends Receiver[Array[Byte]](StorageLevel.MEMORY_ONLY) with Logging {
```
zsxwing: nit: Why create a new class? Is there any concern to just use TestReceiver?

uncleGen:
@rxin OK, I will backport it to branch-2.0

PR #15505, opened 2016-10-16
[SPARK-18890][CORE] Move task serialization from the TaskSetManager to the CoarseGrainedSchedulerBackend

witgo:
## What changes were proposed in this pull request?
Performance Testing:

The code: 
``` scala

val rdd = sc.parallelize(0 until 100).repartition(100000)
rdd.localCheckpoint().count()
rdd.sum()
(1 to 10).foreach{ i=>
  val serializeStart = System.currentTimeMillis()
  rdd.sum()
  val serializeFinish = System.currentTimeMillis()
  println(f"Test $i: ${(serializeFinish - serializeStart) / 1000D}%1.2f")
}

```

and `spark-defaults.conf` file:

```
spark.master                                      yarn-client
spark.executor.instances                          20
spark.driver.memory                               64g
spark.executor.memory                             30g
spark.executor.cores                              5
spark.default.parallelism                         100 
spark.sql.shuffle.partitions                      100
spark.serializer                                  org.apache.spark.serializer.KryoSerializer
spark.driver.maxResultSize                        0
spark.ui.enabled                                  false 
spark.driver.extraJavaOptions                     -XX:+UseG1GC -XX:+UseStringDeduplication -XX:G1HeapRegionSize=16M -XX:MetaspaceSize=512M 
spark.executor.extraJavaOptions                   -XX:+UseG1GC -XX:+UseStringDeduplication -XX:G1HeapRegionSize=16M -XX:MetaspaceSize=256M 
spark.cleaner.referenceTracking.blocking          true
spark.cleaner.referenceTracking.blocking.shuffle  true

```

The test results are as follows:

| [SPARK-17931](https://github.com/witgo/spark/tree/SPARK-17931) |https://github.com/apache/spark/commit/db0ddce523bb823cba996e92ef36ceca31492d2c|
| --- | --- |
| 9.427 s | 9.566 s |

## How was this patch tested?

Existing tests.

PR #16053, opened 2016-11-29
[SPARK-17931] Eliminate unncessary task (de) serialization

kayousterhout:

In the existing code, there are three layers of serialization
    involved in sending a task from the scheduler to an executor:
        - A Task object is serialized
        - The Task object is copied to a byte buffer that also
          contains serialized information about any additional JARs,
          files, and Properties needed for the task to execute. This
          byte buffer is stored as the member variable serializedTask
          in the TaskDescription class.
        - The TaskDescription is serialized (in addition to the serialized
          task + JARs, the TaskDescription class contains the task ID and
          other metadata) and sent in a LaunchTask message.
    
While it *is* necessary to have two layers of serialization, so that
the JAR, file, and Property info can be deserialized prior to
deserializing the Task object, the third layer of deserialization is
unnecessary.  This commit eliminates a layer of serialization by moving
the JARs, files, and Properties into the TaskDescription class.

This commit also serializes the Properties manually (by traversing the map),
as is done with the JARs and files, which reduces the final serialized size.

## How was this patch tested?

Unit tests

## Performance benefit and comparison with #15505 

This is a simpler alternative to the approach proposed in #15505.

@shivaram and I did some benchmarking of this and #15505 on a 20-machine m2.4xlarge EC2 machines (160 cores). We ran ~30 trials of code [1] (a very simple job with 10K tasks per stage) and measured the average time per stage:

Before this change: 2490ms
With this change: 2345 ms (so ~6% improvement over the baseline)
With @witgo's approach in #15505: 2046 ms (~18% improvement over baseline)

The reason that #15505 has a more significant improvement is that it also moves the serialization from the TaskSchedulerImpl thread to the CoarseGrainedSchedulerBackend thread. I added that functionality on top of this change, and got almost the same improvement [1] as #15505 (average of 2103ms). I think we should decouple these two changes, both so we have some record of the improvement form each individual improvement, and because this change is more about simplifying the code base (the improvement is negligible) while the other is about performance improvement.  The plan, currently, is to merge this PR and then merge the remaining part of #15505 that moves serialization.

[1] The reason the improvement wasn't quite as good as with #15505 when we ran the benchmarks is almost certainly because, at the point when we ran the benchmarks, I hadn't updated the code to manually serialize the Properties (instead the code was using Java's default serialization for the Properties object, whereas #15505 manually serialized the Properties).  This PR has since been updated to manually serialize the Properties, just like the other maps.


core/src/main/scala/org/apache/spark/scheduler/TaskDescription.scala
```diff
@@ -17,27 +17,119 @@
...
+ *     (1) When a TaskDescription is received by an Executor, the Executor needs to first get the
+ *         list of JARs and files and add these to the classpath, and set the properties, before
+ *         deserializing the Task object (serializedTask). This is why the Properties are included
+ *         in the TaskDescription, even though they're also in the serialized task.
+ *     (2) Because a TaskDescription is serialized and sent to an executor for each task, efficient
+ *         serialization (both in terms of serialization time and serialized buffer size) is
+ *         important. For this reason, we serialize TaskDescriptions ourselves with the
+ *         TaskDescription.encode and TaskDescription.decode methods.  This results in a smaller
+ *         serialized size because it avoids serializing unnecessary fields in the Map objects
+ *         (which can introduce significant overhead when the maps are small).
```
squito: I realize you are just adding comments to explain the old behavior -- if this explanation about serializing maps is right, then shouldn't you do the same for the `properties`?

squito (commented):
just one minor comment -- I will leave my more general comments about comparison of approaches on https://github.com/apache/spark/pull/15505

kayousterhout:
Thanks for the review @squito.  I got sidetracked from this at the end of last week and forgot to post the results of some benchmarks @shivaram and I did on a 20-machine m2.4xlarge EC2 machines (160 cores).  We ran ~30 trials of code [1] (a very simple job with 10K tasks per stage) and measured the average time per stage:

Before this change: 2490ms
With this change: 2345 ms (so ~6% improvement over the baseline)
With @witgo's approach in #15505: 2046 ms (~18% improvement over baseline)

The reason that #15505 has a more significant improvement is that it also moves the serialization from the TaskSchedulerImpl thread to the CoarseGrainedSchedulerBackend thread.  I added that functionality on top of this change, and got almost the same improvement as #15505 (average of 2103ms).  I think we should decouple these two changes, both so we have some record of the improvement form each individual improvement, and because this change is more about simplifying the code base (the improvement is negligible) while the other is about performance improvement.  I filed a separate JIRA for that issue here: https://issues.apache.org/jira/browse/SPARK-18890

kayousterhout:
Also I need to fix up the Mesos functionality here.  I deliberately posted a broken Mesos version to verify my hypothesis that there are no end-to-end Mesos tests (this code would have failed any such tests), and need to fix it up and then ask one of the Mesos folks to take a look to double check things, since the last of end-to-end tests is a bit scary.

core/src/main/scala/org/apache/spark/scheduler/TaskDescription.scala
kayousterhout: Oh this is a great point, and I realize this is likely also the explanation for the ~2% remaining performance disparity between this change and #15505 that I mentioned in my comment.  Thanks for noticing this oversight -- I'll fix this!

kayousterhout:
I fixed all of the issues mentioned with the commits I pushed a few days ago, in case anyone was waiting for that and didn't notice!

shivaram:
@kayousterhout it shows there are some conflicts ?

shivaram:
I think the code change in core/ looks pretty good to me. Is there someone who can look at the mesos changes ?

core/src/main/scala/org/apache/spark/scheduler/TaskDescription.scala
```diff
@@ -17,27 +17,123 @@
...
-import org.apache.spark.util.SerializableBuffer
+import scala.collection.JavaConverters._
+import scala.collection.mutable.{HashMap, Map}
+
+import org.apache.spark.util.{ByteBufferInputStream, ByteBufferOutputStream, Utils}
 
 /**
  * Description of a task that gets passed onto executors to be executed, usually created by
- * `TaskSetManager.resourceOffer`.
+  * `TaskSetManager.resourceOffer`.
```
mridulm: nit: indent

core/src/main/scala/org/apache/spark/scheduler/TaskDescription.scala
```diff
@@ -17,27 +17,123 @@
...
+      dataOut.writeUTF(name)
+      dataOut.writeLong(timestamp)
+    }
+
+    // Write jars.
+    dataOut.writeInt(taskDescription.addedJars.size)
+    for ((name, timestamp) <- taskDescription.addedJars) {
+      dataOut.writeUTF(name)
+      dataOut.writeLong(timestamp)
+    }
```
mridulm: nit: Looks like repeated code which could be refactored out as a single method ?

```
def writeStringLongMap(map: Map[String, Long]): Unit = {
  dataOut.writeInt(map.size)
  for ((k, v) <- map) {
    dataOut.writeUTF(k)
    dataOut.writeLong(v)
  }
}

writeStringLongMap(taskDescription.addedFiles)
writeStringLongMap(taskDescription. addedJars)

```

Similar applies to decode.
The reason I mention is that both properties have very similar names, and it is easy to misread the code as it evolves.

core/src/main/scala/org/apache/spark/scheduler/TaskDescription.scala
```diff
@@ -17,27 +17,123 @@
...
+
+    // Write properties.
+    dataOut.writeInt(taskDescription.properties.size())
+    taskDescription.properties.stringPropertyNames.asScala.foreach { name =>
+      dataOut.writeUTF(name)
+      dataOut.writeUTF(taskDescription.properties.getProperty(name))
+    }
```
mridulm: Why not :
```
taskDescription.properties.asScala.foreach { case (key, value) => {
    dataOut.writeUTF(key)
    dataOut.writeUTF(value)
  }
}
```

Avoid probing the map again.

core/src/main/scala/org/apache/spark/scheduler/TaskDescription.scala
```diff
@@ -17,27 +17,123 @@
...
+
+    // Write the task. The task is already serialized, so write it directly to the byte buffer
+    // (this requires first flushing the data output stream, so that all of the data has been
+    // written from the data output stream so the underlying ByteBufferOutputStream before
+    // we write the task).
+    dataOut.flush()
```
mridulm: This is a noop and can be removed.

core/src/main/scala/org/apache/spark/scheduler/TaskDescription.scala
```diff
@@ -17,27 +17,123 @@
...
+    Utils.writeByteBuffer(taskDescription.serializedTask, bytesOut)
```
mridulm: Just noticed that `Utils.writeByteBuffer` actually can have different state for serializedTask on return.
If it has array, `remaining` remains unchanged - if not, `remaining` will be 0 : was this on purpose ?

Will this behavior have issues in our codepath here ? (needs to be fixed in Utils.writeByteBuffer - but unrelated to this pr ?)
Or perhaps send `taskDescription.serializedTask.duplicate()` over to the method ?

mridulm (commented):
Added a few comments.

core/src/main/scala/org/apache/spark/scheduler/TaskDescription.scala
kayousterhout: Just opened #16462 to fix this.  I don't think this has any issues for this code path, because this is exactly the same as the old code that was previously located in Task.scala (which was moved here in this PR).

core/src/main/scala/org/apache/spark/scheduler/TaskDescription.scala
kayousterhout: Done

kayousterhout (commented):
@squito @mridulm @shivaram Thanks for looking at this! I've addressed all of the comments and opened a separate PR for one of the issues you mentioned Mridul.

@mgummelt would you mind taking a look at the Mesos changes here?  I've checked them over but there aren't any unit tests for this part of the code, so it would be great to have a second set of eyes (or, even better, if you happen to have a Mesos cluster up, would be awesome to quickly test that tasks can still be launched properly in coarse grained mode).

kayousterhout:
Jenkins, retest this please

mgummelt:
@kayousterhout It looks like you only touched deprecated Mesos code (fine-grained mode), so LGTM

kayousterhout:
Cool thanks for looking at the Mesos changes here @mgummelt!

squito:
Jenkins, retest this please

squito:
lgtm, assuming tests pass

kayousterhout:
Cool thanks for the review @squito and for re-testing (I tracked down the flaky test that caused this but then forgot to re-test!)

squito:
merged to master

PR #16056, opened 2016-11-29
[SPARK-18623][SQL] Add `returnNullable` to `StaticInvoke` and modify it to handle properly.

ueshin:

Add `returnNullable` to `StaticInvoke` the same as #15780 is trying to add to `Invoke` and modify to handle properly.

## How was this patch tested?

Existing tests.


ueshin:
I will update the comment again after #15780 is merged.

gatorsmile:
cc @ueshin

kiszk:
@ueshin should we close this? This is because the current [`StaticInvoke`](github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala#L122-L127) does not have `returnNullable` argument.

ueshin:
I'd close this for now.

cloud-fan:
shall we reopen? I think it's a valid fix

ueshin:
Yes, I'll reopen, and update if needed.

kiszk:
Oh, I am sorry for my misunderstanding. This PR also add an argument ` returnNullable` to `StaticInvoke()`. I overlooked it.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/JavaTypeInference.scala
```diff
@@ -216,23 +216,26 @@ object JavaTypeInference {
           ObjectType(c),
           "valueOf",
           getPath :: Nil,
-          propagateNull = true)
+          propagateNull = true,
```
cloud-fan: nit: we can remove `propagateNull = true` as it's true by default

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/encoders/RowEncoder.scala
```diff
@@ -96,28 +96,32 @@ object RowEncoder {
...
+        returnNullable = false)
 
     case StringType =>
       StaticInvoke(
         classOf[UTF8String],
         StringType,
         "fromString",
-        inputObject :: Nil)
+        inputObject :: Nil,
+        returnNullable = true)
```
cloud-fan: `UTF8String.fromString` only returns null if input string is null, and here `propogateNull` is true, which means `retutnNullabe` can be false, because when we invoke this method, the input string must be non-null.

ueshin:
#18524 could fix the test failure.

ueshin:
Jenkins, retest this please.

cloud-fan:
LGTM, pending test

cloud-fan:
thanks, merging to master!

PR #16057, opened 2016-11-29
[SPARK-18624][SQL] Implicit cast ArrayType(InternalType)

jiangxb1987:

Currently `ImplicitTypeCasts` doesn't handle casts between `ArrayType`s, this is not convenient, we should add a rule to enable casting from `ArrayType(InternalType)` to `ArrayType(newInternalType)`.

Goals:
1. Add a rule to `ImplicitTypeCasts` to enable casting between `ArrayType`s;
2. Simplify `Percentile` and `ApproximatePercentile`.

## How was this patch tested?

Updated test cases in `TypeCoercionSuite`.

### DIFF
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/TypeCoercion.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/TypeCoercion.scala
@@ -673,59 +673,54 @@ object TypeCoercion {
      * If the expression has an incompatible type that cannot be implicitly cast, return None.
      */
     def implicitCast(e: Expression, expectedType: AbstractDataType): Option[Expression] = {
-      val inType = e.dataType
+      implicitCast(e.dataType, expectedType).map { dt =>
+        if (dt == e.dataType) e else Cast(e, dt)
+      }
+    }
 
+    private def implicitCast(inType: DataType, expectedType: AbstractDataType): Option[DataType] = {
       // Note that ret is nullable to avoid typing a lot of Some(...) in this local scope.
       // We wrap immediately an Option after this.
-      @Nullable val ret: Expression = (inType, expectedType) match {
-
+      @Nullable val ret: DataType = (inType, expectedType) match {
         // If the expected type is already a parent of the input type, no need to cast.
-        case _ if expectedType.acceptsType(inType) => e
+        case _ if expectedType.acceptsType(inType) => inType
 
         // Cast null type (usually from null literals) into target types
-        case (NullType, target) => Cast(e, target.defaultConcreteType)
+        case (NullType, target) => target.defaultConcreteType
 
         // If the function accepts any numeric type and the input is a string, we follow the hive
         // convention and cast that input into a double
-        case (StringType, NumericType) => Cast(e, NumericType.defaultConcreteType)
+        case (StringType, NumericType) => NumericType.defaultConcreteType
 
         // Implicit cast among numeric types. When we reach here, input type is not acceptable.
 
         // If input is a numeric type but not decimal, and we expect a decimal type,
         // cast the input to decimal.
-        case (d: NumericType, DecimalType) => Cast(e, DecimalType.forType(d))
+        case (d: NumericType, DecimalType) => DecimalType.forType(d)
         // For any other numeric types, implicitly cast to each other, e.g. long -> int, int -> long
-        case (_: NumericType, target: NumericType) => Cast(e, target)
+        case (_: NumericType, target: NumericType) => target
 
         // Implicit cast between date time types
-        case (DateType, TimestampType) => Cast(e, TimestampType)
-        case (TimestampType, DateType) => Cast(e, DateType)
+        case (DateType, TimestampType) => TimestampType
+        case (TimestampType, DateType) => DateType
 
         // Implicit cast from/to string
-        case (StringType, DecimalType) => Cast(e, DecimalType.SYSTEM_DEFAULT)
-        case (StringType, target: NumericType) => Cast(e, target)
-        case (StringType, DateType) => Cast(e, DateType)
-        case (StringType, TimestampType) => Cast(e, TimestampType)
-        case (StringType, BinaryType) => Cast(e, BinaryType)
+        case (StringType, DecimalType) => DecimalType.SYSTEM_DEFAULT
+        case (StringType, target: NumericType) => target
+        case (StringType, DateType) => DateType
+        case (StringType, TimestampType) => TimestampType
+        case (StringType, BinaryType) => BinaryType
         // Cast any atomic type to string.
-        case (any: AtomicType, StringType) if any != StringType => Cast(e, StringType)
+        case (any: AtomicType, StringType) if any != StringType => StringType
 
         // When we reach here, input type is not acceptable for any types in this type collection,
         // try to find the first one we can implicitly cast.
-        case (_, TypeCollection(types)) => types.flatMap(implicitCast(e, _)).headOption.orNull
+        case (_, TypeCollection(types)) =>
+          types.flatMap(implicitCast(inType, _)).headOption.orNull
 
-        case (ArrayType(_, nullable), ArrayType(internalType: DataType, expectedNullable))
-          if (expectedNullable || nullable == expectedNullable) =>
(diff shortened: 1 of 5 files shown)


sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/TypeCoercion.scala
```diff
@@ -714,6 +714,17 @@ object TypeCoercion {
         // try to find the first one we can implicitly cast.
         case (_, TypeCollection(types)) => types.flatMap(implicitCast(e, _)).headOption.orNull
 
+        case (ArrayType(_, nullable), ArrayType(internalType: DataType, expectedNullable))
+          if (expectedNullable || nullable == expectedNullable) =>
+          val newChildren = e.children.map { expr =>
```
hvanhovell: Why is this needed? We can cast the array directly.

hvanhovell (changes requested):

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/TypeCoercion.scala
```diff
@@ -714,6 +714,17 @@ object TypeCoercion {
         // try to find the first one we can implicitly cast.
         case (_, TypeCollection(types)) => types.flatMap(implicitCast(e, _)).headOption.orNull
 
+        case (ArrayType(_, nullable), ArrayType(internalType: DataType, expectedNullable))
+          if (expectedNullable || nullable == expectedNullable) =>
```
HyukjinKwon: Ah, actually @hvanhovell, in case of `case _ if expectedType.acceptsType(inType) => inType` would this return `None` instead of the same input type? 

I see this is being used by `implicitCast(in, expected).getOrElse(in)` so it would be fine to work but I am worried of this (e.g. for unit test and etc).

viirya:
If this is going to support ArrayType only, can you update the pr title to reflect it?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/TypeCoercion.scala
viirya: We should also consider the case forcing the casting result is null. Please see https://github.com/apache/spark/blob/7eb2ca8e338e04034a662920261e028f56b07395/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/Cast.scala#L94

hvanhovell: @HyukjinKwon the idea is that returning `None` means that we should not cast at all. Returning the same type would mean that we could cast. This is only relevant for `ArrayType` case, because we could only change the nullability of the column, e.g.: `Cast(a, ArrayType(DoubleType, false), ArrayType(DoubleType, true))`

Replacing the `collect { ... }` by a `map { ... }` changes the semantic of the `implicitCast(...)` function, and will probably break the [`ResolveTableValuedFunctions`](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/ResolveTableValuedFunctions.scala#L39-L49) rule.

hvanhovell: @viirya yeah, we should consider the nullability.

HyukjinKwon: Thank you for your detailed explanation and yeap that makes sense. I see it is up to what `None` means from the function. I was just worried of changing the existing behaviour and wanted to note just in case. I agree with either way if it sounds consistent:).

HyukjinKwon: I just ran `TypeCoercionSuite` with the change and it seems several tests are failed when the source and target types are the same.

hvanhovell: @HyukjinKwon it is a suggestion, I did not test it. I am curious to see what is failing.

hvanhovell: Running it now....:)

HyukjinKwon: Thank you!!

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/TypeCoercion.scala
```diff
@@ -673,48 +673,55 @@ object TypeCoercion {
...
+        case (any: AtomicType, StringType) if any != StringType => StringType
 
         // When we reach here, input type is not acceptable for any types in this type collection,
         // try to find the first one we can implicitly cast.
-        case (_, TypeCollection(types)) => types.flatMap(implicitCast(e, _)).headOption.orNull
+        case (_, TypeCollection(types)) =>
+          types.flatMap(implicitCast(inType, _)).headOption.orNull
+
+        case (ArrayType(fromType, fn), ArrayType(toType: DataType, tn))
+            if !(fn || Cast.forceNullable(fromType, toType)) || tn =>
```
hvanhovell: I had to look quite long at this one, so:

- `toNull` => Always allow
- `fromNull && !toNull` => Never allow
- `!fromNull && !toNull` => Only allow when `!Cast.forceNullable(fromType, toType)`

What you have written is just a very terse representation of that.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/ApproximatePercentile.scala
```diff
@@ -85,23 +85,16 @@ case class ApproximatePercentile(
   private lazy val accuracy: Int = accuracyExpression.eval().asInstanceOf[Int]
 
   override def inputTypes: Seq[AbstractDataType] = {
-    Seq(DoubleType, TypeCollection(DoubleType, ArrayType), IntegerType)
+    Seq(DoubleType, TypeCollection(DoubleType, ArrayType(DoubleType)), IntegerType)
```
hvanhovell: `ArrayType(DoubleType, false)`?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/Percentile.scala
```diff
@@ -98,7 +92,7 @@ case class Percentile(
   }
 
   override def inputTypes: Seq[AbstractDataType] = percentageExpression.dataType match {
-    case _: ArrayType => Seq(NumericType, ArrayType)
+    case _: ArrayType => Seq(NumericType, ArrayType(DoubleType))
```
hvanhovell: `ArrayType(DoubleType, false)`?

jiangxb1987: If the input array's widest type is a `DecimalType`, `CreateArray()` will create a `ArrayType(DecimalType, containsNull = true)`, because the function `Cast. forceNullable(fromType, toType)` always returns true if `toType` is a `DecimalType`.
Then, if we are expecting a inputType as `ArrayType(DoubleType, containsNull = false)`, we fail to cast because the nullability is not compatible.
For example, the `array(0.0, 1.0)` will be casted to `ArrayType(DecimalType(2, 1), containsNull = true)`, and this is a valid input for `Percentile` function.
Thus, I decide to keep the `inputType` to be `ArrayType(DoubleType, containsNull = true)`, to be compatible with the above case.

hvanhovell: Ok, that is reasonable.

hvanhovell:

ashokblend:
Hi Guys

ashokblend:
Any Reason why its not merged in branch2.1

PR #16059, opened 2016-11-29
[SPARK-18625][ML] OneVsRestModel should support setFeaturesCol and setPredictionCol

zhengruifeng:
add `setFeaturesCol` and `setPredictionCol` for `OneVsRestModel`

## How was this patch tested?
added tests

### DIFF
--- a/mllib/src/main/scala/org/apache/spark/ml/classification/OneVsRest.scala
+++ b/mllib/src/main/scala/org/apache/spark/ml/classification/OneVsRest.scala
@@ -140,6 +140,20 @@ final class OneVsRestModel private[ml] (
     this(uid, Metadata.empty, models.asScala.toArray)
   }
 
+  /** @group setParam */
+  @Since("2.2.0")
+  def setFeaturesCol(value: String): this.type = {
+    models.foreach(_.setFeaturesCol(value))
+    set(featuresCol, value)
+  }
+
+  /** @group setParam */
+  @Since("2.2.0")
+  def setPredictionCol(value: String): this.type = {
+    models.foreach(_.setPredictionCol(value))
+    set(predictionCol, value)
+  }
+
   @Since("1.4.0")
   override def transformSchema(schema: StructType): StructType = {
     validateAndTransformSchema(schema, fitting = false, getClassifier.featuresDataType)
--- a/mllib/src/main/scala/org/apache/spark/ml/classification/OneVsRest.scala
+++ b/mllib/src/main/scala/org/apache/spark/ml/classification/OneVsRest.scala
@@ -142,17 +142,11 @@ final class OneVsRestModel private[ml] (
 
   /** @group setParam */
   @Since("2.2.0")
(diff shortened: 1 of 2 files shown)


srowen:
Is this not the same as https://github.com/apache/spark/pull/15957 but for a different class?

zhengruifeng:
@srowen Excepet that OvR is a meta algorithm, and we need to update the behavior of base models.

srowen:
OK, should it not have been added in the other PR? I want to push back on adding logical changes piecemeal. I understand sometimes it only becomes clear later that further changes are needed, but if it happens regularly or seems avoidable it becomes a problem.

zhengruifeng:
I understand your opinion that we should make same update in a batch. But I just find this issue this night...

srowen:
CC @yanboliang ?

yanboliang:
@srowen Thanks for cc me, I'm having a look now.

mllib/src/main/scala/org/apache/spark/ml/classification/OneVsRest.scala
```diff
@@ -175,6 +183,8 @@ final class OneVsRestModel private[ml] (
         val updateUDF = udf { (predictions: Map[Int, Double], prediction: Vector) =>
           predictions + ((index, prediction(1)))
         }
+        model.setFeaturesCol($(featuresCol))
+        model.setPredictionCol($(predictionCol))
```
yanboliang: It's not necessary to ```setPredictionCol``` here, since the prediction column of each base binary classifier is ignored during the transformation. The ```predictionCol``` of ```OneVsRestModel``` was set at L198.

mllib/src/test/scala/org/apache/spark/ml/classification/OneVsRestSuite.scala
```diff
@@ -136,6 +137,19 @@ class OneVsRestSuite extends SparkFunSuite with MLlibTestSparkContext with Defau
...
+    val ova = new OneVsRest().setClassifier(new LogisticRegression)
+    val ovaModel = ova.fit(dataset)
+    val dataset2 = dataset.select(col("label").as("y"), col("features").as("fea"))
+    ovaModel.setFeaturesCol("fea")
+    ovaModel.setPredictionCol("pred")
+    val transformedDataset = ovaModel.transform(dataset2)
+    val outputFields = transformedDataset.schema.fieldNames.toSet
+    assert(outputFields.contains("y"))
+    assert(outputFields.contains("fea"))
+    assert(outputFields.contains("pred"))
```
yanboliang: Minor: ```assert(outputFields === Set("y", "fea", "pred"))```

yanboliang (commented):
Looks good overall, except some minor comments. Thanks!

zhengruifeng:
@yanboliang Updated. Thanks for reviewing.

yanboliang:
LGTM, merged into master and branch-2.1. Thanks for all.

PR #14363, opened 2016-07-26
[SPARK-16731][SQL] use StructType in CatalogTable and remove CatalogColumn

cloud-fan:
## What changes were proposed in this pull request?

`StructField` has very similar semantic with `CatalogColumn`, except that `CatalogColumn` use string to express data type. I think it's reasonable to use `StructType` as the `CatalogTable.schema` and remove `CatalogColumn`.
## How was this patch tested?

existing tests.

PR #16060, opened 2016-11-29
[SPARK-18220][SQL] read Hive orc table with varchar column should not fail

cloud-fan:

Spark SQL only has `StringType`, when reading hive table with varchar column, we will read that column as `StringType`. However, we still need to use varchar `ObjectInspector` to read varchar column in hive table, which means we need to know the actual column type at hive side.

In Spark 2.1, after https://github.com/apache/spark/pull/14363 , we parse hive type string to catalyst type, which means the actual column type at hive side is erased. Then we may use string `ObjectInspector` to read varchar column and fail.

This PR keeps the original hive column type string in the metadata of `StructField`, and use it when we convert it to a hive column.

## How was this patch tested?

newly added regression test

### DIFF
--- a/sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveUtils.scala
+++ b/sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveUtils.scala
@@ -51,8 +51,11 @@ private[spark] object HiveUtils extends Logging {
     sc
   }
 
-  /** The version of hive used internally by Spark SQL. */
+  // The version of hive used internally by Spark SQL.
   val hiveExecutionVersion: String = "1.2.1"
+
+  // The property key that is used to store the raw hive type string in the metadata of StructField.
+  val hiveTypeString: String = "HIVE_TYPE_STRING"
 
   val HIVE_METASTORE_VERSION = SQLConfigBuilder("spark.sql.hive.metastore.version")
     .doc("Version of the Hive metastore. Available options are " +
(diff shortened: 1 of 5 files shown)


cloud-fan:
cc @yhuai @gatorsmile

sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveUtils.scala
```diff
@@ -51,9 +51,12 @@ private[spark] object HiveUtils extends Logging {
     sc
   }
 
-  /** The version of hive used internally by Spark SQL. */
+  // The version of hive used internally by Spark SQL.
```
rxin: why did you change the comment style?

sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveUtils.scala
```diff
@@ -51,9 +51,12 @@ private[spark] object HiveUtils extends Logging {
...
   val hiveExecutionVersion: String = "1.2.1"
 
+  // The property key that is used to store the raw hive type string in the metadata of StructField.
```
rxin: I'd add a bit more color here, e.g. by adding an example: "For example, in the case where the Hive type is varchar, the type gets mapped to a string type in Spark SQL, but we need to preserve the original type in order to invoke the correct object inspector in Hive"

rxin:
Merging in master/branch-2.1.

PR #16061, opened 2016-11-29
[SPARK-18278] [Scheduler] Support native submission of spark jobs to a kubernetes cluster

erikerlandson:
Add support for native submission of spark jobs to a kubernetes cluster, as a first-class scheduler back-end.

## How was this patch tested?
Currently, testing has been mostly informal.  Integration tests will be added before completion.

## Notes
The initial branch on this PR is intended to serve as an "MVP" base for integrating various contributions to a "k8s-native" effort being conducted by several interested community members, including:

@foxish, @erikerlandson, @iyanuobidele
https://github.com/foxish/spark/tree/k8s-support/kubernetes

@mccheah 
https://github.com/foxish/spark/pull/7

Goals for integrating features include:

- build of docker image (or images) for spark
- integration tests
- publishing of third-party resources
- resource constraints
- external shuffle service via daemon-sets
- possible hooks for control of executor scale by kubernetes control plane

## Instructions

Current instructions for building at the time of this submission are here:
https://github.com/foxish/spark/tree/k8s-support/kubernetes


kubernetes/src/main/scala/org/apache/spark/scheduler/cluster/kubernetes/KubernetesClusterSchedulerBackend.scala
```diff
@@ -0,0 +1,222 @@
...
+  // Dynamic allocation interfaces
+  override def doRequestTotalExecutors(requestedTotal: Int): Future[Boolean] = {
+    logInfo(s"Received doRequestTotalExecutors: $requestedTotal")
+    val n = executorToPod.size
+    val delta = requestedTotal - n
+    if (delta > 0) {
+      logInfo(s"Adding $delta new executors")
+      createExecutorPods(delta)
+    } else if (delta < 0) {
+      val d = -delta
```
tnachen: This shouldn't happen, assert instead

kubernetes/src/main/scala/org/apache/spark/scheduler/cluster/kubernetes/KubernetesClusterSchedulerBackend.scala
```diff
@@ -0,0 +1,222 @@
...
+import org.apache.spark.rpc.RpcEndpointAddress
+import org.apache.spark.scheduler.TaskSchedulerImpl
+import org.apache.spark.util.Utils
+
+import scala.collection.mutable
+import scala.util.Random
+import scala.concurrent.Future
+
+private[spark] class KubernetesClusterSchedulerBackend(
+                                                  scheduler: TaskSchedulerImpl,
```
tnachen: Fix the formatting to conform with Spark style

kubernetes/src/main/scala/org/apache/spark/scheduler/cluster/kubernetes/KubernetesClusterScheduler.scala
```diff
@@ -0,0 +1,236 @@
...
+      .endSpec()
+      .build()
+
+    client
+      .services()
+      .inNamespace(nameSpace)
+      .withName(svcName)
+      .create(svc)
+
+//    try {
```
tnachen: Remove

rxin:
This is pretty cool.

rxin:
One thing - can we submit a separate pr to move all resource managers into 

resource-managers/yarn
resource-managers/mesos

?

tnachen:
@rxin Makes sense, @srowen also talked about starting the discussion of having a better support for external cluster managers as well.

liancheng:
@erikerlandson For the RAT failure, you may either add Apache license header to newly added files or add the file to `dev/.rat-excludes`.

kubernetes/src/main/scala/org/apache/spark/scheduler/cluster/kubernetes/KubernetesClusterSchedulerBackend.scala
erikerlandson: Delta (as currently computed) can become negative, whenever the requested total starts to decrease.

However, this logic warrants a total overhaul, since it was designed around the assumption that `doRequestTotalExecutors` is symmetric, in the sense of having to handle a decreasing total analogously to an increasing total.  But responsibilities for executor shutdown vs spin-up aren't symmetric.  I might file a small doc PR to clarify that point.

erikerlandson:
@rxin, when you say "move all resource managers" does that mean "move scheduler back-ends for mesos, yarn, etc, into some `resource-managers` sub-project"  ?

ash211:
Another external scheduler backend I'm aware of is Two Sigma's scheduler backend for the system they've created called [Cook](https://github.com/twosigma/Cook).  See [CoarseCookSchedulerBackend.scala](https://github.com/twosigma/spark/blob/2.0.2-cook/core/src/main/scala/org/apache/spark/scheduler/CoarseCookSchedulerBackend.scala)

kubernetes/README.md
```diff
@@ -0,0 +1,21 @@
+# Pre-requisites
+* maven, JDK and all other pre-requisites for building Spark.
+
+# Steps to compile
+
+* Clone the fork of spark: https://github.com/foxish/spark/ and switch to the k8s-support branch.
```
viirya: I think this is not correct now?

kubernetes/README.md
```diff
@@ -0,0 +1,21 @@
...
+* Build the project
+    * ./build/mvn -Pkubernetes -Phadoop-2.4 -Dhadoop.version=2.4.0 -DskipTests package
+* Ensure that you are pointing to a k8s cluster (kubectl config current-context), which you want to use with spark.
+* Launch a spark-submit job:
+   * `./bin/spark-submit --deploy-mode cluster --class org.apache.spark.examples.SparkPi --master k8s://default --conf spark.executor.instances=5 --conf spark.kubernetes.sparkImage=manyangled/kube-spark:dynamic  http://storage.googleapis.com/foxish-spark-distro/original-spark-examples_2.11-2.1.0-SNAPSHOT.jar 10000`
```
erikerlandson: There will need to be some official Apache Spark repository for images, which I presume will be up to Apache Spark to create.   The exact nature of the images to be produced is still being discussed

I did create a "semi-official" org up on docker hub called "k8s4spark":
https://hub.docker.com/u/k8s4spark/dashboard/

I haven't actually pushed any images to it, but we could start using it as an interim repo if people think that is useful

dev/make-distribution.sh
```diff
@@ -154,7 +154,9 @@ export MAVEN_OPTS="${MAVEN_OPTS:--Xmx2g -XX:MaxPermSize=512M -XX:ReservedCodeCac
 # Store the command as an array because $MVN variable might have spaces in it.
 # Normal quoting tricks don't work.
 # See: http://mywiki.wooledge.org/BashFAQ/050
-BUILD_COMMAND=("$MVN" -T 1C clean package -DskipTests $@)
+# BUILD_COMMAND=("$MVN" -T 1C clean package -DskipTests $@)
+
+BUILD_COMMAND=("$MVN" -T 2C package -DskipTests $@)
```
foxish: This should be reverted.

foxish:
Cross posting link to [the proposal](https://issues.apache.org/jira/secure/attachment/12840988/SPARK-18278%20-%20Spark%20on%20Kubernetes%20Design%20Proposal.pdf) here.

kubernetes/src/main/scala/org/apache/spark/scheduler/cluster/kubernetes/KubernetesClusterSchedulerBackend.scala
```diff
@@ -0,0 +1,222 @@
...
+      .withNewMetadata()
+      .withLabels(labelMap.asJava)
+      .withName(podName)
+      .endMetadata()
+      .withNewSpec()
+      .withRestartPolicy("OnFailure")
+
+      .addNewContainer().withName("spark-executor").withImage(sparkImage)
+      .withImagePullPolicy("IfNotPresent")
+      .withCommand("/opt/executor.sh")
```
mccheah: Another approach that we could take is to put the command in the Dockerfile and supply the container with environment variables to configure the runtime behavior. A few benefits to this:
- Transparency: Instead of needing to inspect the shim scripts to discover the specific execution behavior, the behavior is well-defined in the Dockerfile.
- Immutability: The run command does not need to be assembled by the client code every time.
- Decoupling execution from logic: The Docker containers are only configured with properties, as opposed to needing to specify both logic (running the shim script) and configuration.

kubernetes/src/main/scala/org/apache/spark/scheduler/cluster/kubernetes/KubernetesClusterSchedulerBackend.scala
```diff
@@ -0,0 +1,222 @@
...
+
+    if (conf.getBoolean("spark.dynamicAllocation.enabled", false)) {
+      submitArgs ++= Vector(
+        "dynamic-executors")
+    }
+
+    submitArgs ++= Vector("org.apache.spark.executor.CoarseGrainedExecutorBackend",
+      "--driver-url", s"$driverURL",
+      "--executor-id", s"$executorNum",
+      "--hostname", "localhost",
```
mccheah: Should we use the HOSTNAME environment variable here?

dev/deps/spark-deps-hadoop-2.4
```diff
@@ -98,7 +98,7 @@ jersey-client-2.22.2.jar
 jersey-common-2.22.2.jar
 jersey-container-servlet-2.22.2.jar
 jersey-container-servlet-core-2.22.2.jar
-jersey-guava-2.22.2.jar
+jersey-guava-2.22.2.jarshaded-proto
```
mccheah: Any particular reason why this changed? Would be curious to trace how this dependency changed.

core/src/main/scala/org/apache/spark/deploy/SparkSubmit.scala
```diff
@@ -596,6 +599,26 @@ object SparkSubmit extends CommandLineUtils {
       }
     }
 
+    if (isKubernetesCluster) {
```
mccheah: We should throw an error for now if in client mode with Kubernetes. There's still open questions about how we can manage the networking there, so we can revisit the support later.

kubernetes/src/main/scala/org/apache/spark/scheduler/cluster/kubernetes/KubernetesClusterSchedulerBackend.scala
```diff
@@ -0,0 +1,222 @@
...
+  val sparkExecutorName = s"spark-executor-${Random.alphanumeric take 5 mkString("")}".toLowerCase()
+
+  // TODO: do these need mutex guarding?
+  // key is executor id, value is pod name
+  var executorToPod = mutable.Map.empty[String, String] // active executors
+  var shutdownToPod = mutable.Map.empty[String, String] // pending shutdown
+  var executorID = 0
+
+  val sparkImage = conf.get("spark.kubernetes.sparkImage")
+  val clientJarUri = conf.get("spark.executor.jar")
```
mccheah: Is this used anywhere in this class?

kubernetes/src/main/scala/org/apache/spark/scheduler/cluster/kubernetes/KubernetesClusterSchedulerBackend.scala
```diff
@@ -0,0 +1,222 @@
...
+    if (conf.getBoolean("spark.dynamicAllocation.enabled", false)) {
+      submitArgs ++= Vector(
+        "dynamic-executors")
+    }
+
+    submitArgs ++= Vector("org.apache.spark.executor.CoarseGrainedExecutorBackend",
+      "--driver-url", s"$driverURL",
+      "--executor-id", s"$executorNum",
+      "--hostname", "localhost",
+      "--app-id", "1", // TODO: change app-id per application and pass from driver.
```
mccheah: We can fill this in with the `applicationId()` method defined in `SchedulerBackend`.

kubernetes/src/main/scala/org/apache/spark/scheduler/cluster/kubernetes/KubernetesClusterScheduler.scala
```diff
@@ -0,0 +1,236 @@
...
+    "spark.executor.jar",
+    "spark.dynamicAllocation.enabled",
+    "spark.shuffle.service.enabled")
+
+  def start(args: ClientArguments): Unit = {
+    startDriver(client, args)
+  }
+
+  def stop(): Unit = {
+    client.pods().inNamespace(nameSpace).withName(driverName).delete()
```
mccheah: I think if we do this here in cluster mode we will shut ourselves down and risk not actually finishing the `stop()` method. It's also possible that the `SparkContext` object and its associated components like the scheduler can be stopped, but the JVM still is doing other things afterwards. Therefore I don't think we should be deleting the driver pod here.

kubernetes/src/main/scala/org/apache/spark/scheduler/cluster/kubernetes/KubernetesClusterSchedulerBackend.scala
erikerlandson: I think if we can just execute the container and pass config via something along the lines of `SPARK_MASTER_OPTS` (is there `SPARK_DRIVER_OPTS` ?), that could also implicitly handle the case of running a custom user container.  It might shift documentation of conventions to which env vars to expect, but if we can use standard spark env-vars that would be idiomatic.

mccheah: That makes sense - essentially the Dockerfile has this line:

`CMD exec ${JAVA_HOME}/bin/java -Xmx$SPARK_EXECUTOR_MEMORY org.apache.spark... --executor-id $SPARK_EXECUTOR_ID ...` etc.

And then we set these environment variables on the container.

kubernetes/src/main/scala/org/apache/spark/scheduler/cluster/kubernetes/KubernetesClusterSchedulerBackend.scala
erikerlandson: I had started in on keeping track of new pods until they stop saying "pending", however I think this may be a good use case for a `PodWatcher`

kubernetes/src/main/scala/org/apache/spark/scheduler/cluster/kubernetes/KubernetesClusterSchedulerBackend.scala
erikerlandson: IIUC, you don't even need to throw the `-Xmx` mem flags; the driver, executor, etc are supposed to honor those (and the `*_CORES`), aren't they?    We've been using `spark-class` to run the executor backends.   And I think just `spark-submit` (client mode) for the driver.   Although I think it all eventually bakes down to a jvm call.

mccheah: `spark-class` wasn't built to handle starting executors so it won't pick up memory flags for them. We have to execute the Java binary directly. `spark-class` I believe was built for running long-lived daemons like the history server and standalone components.

As for the driver - we can run `spark-submit` for now but when we move to supporting uploading files from the submitter's local disk we're going to need an in-between process in the pod to upload the local resources to it.

mccheah: Actually I think I was mistaken here - `spark-class` eventually calls to [this class](https://github.com/apache/spark/blob/master/launcher/src/main/java/org/apache/spark/launcher/SparkClassCommandBuilder.java#L65) which then forks the actual process running the passed-in class. We can probably then indeed take advantage of `SPARK_JAVA_OPTS` and `SPARK_EXECUTOR_MEMORY`, etc. Arguments like executor-id which are passed as command line arguments and not JVM options should probably still be set as environment variables on the container though and the Dockerfile handles formatting the command line arguments sent to `spark-class`.

erikerlandson:
Since we have at least four people with interest in working on the code, I have a workflow proposal, which is that we all submit candidate updates to #16061 in the form of PRs against [erikerlandson:spark-18278-k8s-native](https://github.com/erikerlandson/spark/tree/spark-18278-k8s-native) (recursive PRs against a PR!)

I did this with Anirudh against his `k8s-support` branch, and it seemed to work fine.  It should help keep asynchronous contributions sane, and provides the usual durable public forum for discussing the updates as they evolve.

erikerlandson:
At the risk of redundant announcements, I am going to present this spark-on-kube topic at the [Kubernetes SIG-Apps](https://github.com/kubernetes/community/tree/master/sig-apps) next Monday (Dec 5).  It will be similar to my earlier [OpenShift briefing](https://blog.openshift.com/openshift-commons-big-data-sig-2-running-apache-spark-natively-on-kubernetes-with-openshift/) but the focus will be on kube specifically, and I'll update it to cover all the latest developments.

iyanuobidele:
@erikerlandson are there any new commits to this branch ? Or it's at the same level as @foxish's k8s-support. In case I need to rebase....

foxish:
I would think that it would be easier if we continue to keep PRs and issues in foxish/spark since we have the previous PRs and issues over there. It should be easy enough to pick those commits up and update this branch once we review and merge them. It doesn't matter where it lives really, except that it seems like most of it is already on one place.

erikerlandson:
@iyanuobidele the current head of this branch should be equivalent, but it can't hurt to rebase just in case.

@foxish yesterday occurred to me that it would've been sensible set up a GH organization for this little consortium, and run the PR from there.  Presumably we still could.   The drawback to making any change to it now is losing the dialog that's happened here already.

The problem I'm interested in solving even more is ensuring that everybody who contributes to this PR gets logged somehow in the upstream commit history when this PR eventually gets merged.   I don't think that would happen in the case of a typical squash-merge, although I assume it _would_ using a rebase merge. @rxin do you have any thoughts on that?

rxin:
We should acknowledge people that have done this in the merged commit, but I don't think it'd make a lot of sense to do a merge rather than squash. The initial Spark SQL commit was a single commit even though it was a much larger project https://github.com/apache/spark/pull/146

core/src/main/scala/org/apache/spark/deploy/SparkSubmit.scala
mogthesprog: Hey Guys, really excited about this work by the way. :)

Just wondering if client mode from within a kubernetes cluster will be supported? Not looking to add work, just curious.

I'm wondering whether we could check if the client IP is within the kubernetes container CIDR (this info should be available via the kubernetes API), instead of just blocking client mode altogether. This would support Zeppelin/Jupyter instances  running within kube connecting to spark. Which is a big usecase in our case. 

We could take the [nodelist](https://kubernetes.io/docs/api-reference/v1/definitions/#_v1_nodelist) and check if the client IP is in any of the [v1.NodeSpec.PodCIDR](https://kubernetes.io/docs/api-reference/v1/definitions/#_v1_nodespec) of any node. Although as k8s clusters get large (nodes>1000), this may become a costly operation.

This would allow for client mode support, provided the client is on one of the kube nodes.

kubernetes/README.md
```diff
@@ -0,0 +1,21 @@
+# Pre-requisites
+* maven, JDK and all other pre-requisites for building Spark.
+
+# Steps to compile
+
+* Clone the fork of spark: https://github.com/foxish/spark/ and switch to the k8s-support branch.
+* Build the project
+    * ./build/mvn -Pkubernetes -Phadoop-2.4 -Dhadoop.version=2.4.0 -DskipTests package
```
jaceklaskowski: I think `hadoop-2.4` is gone in master.

kubernetes/pom.xml
```diff
@@ -0,0 +1,54 @@
...
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.apache.spark</groupId>
+    <artifactId>spark-parent_2.11</artifactId>
+    <version>2.1.0-SNAPSHOT</version>
```
jaceklaskowski: `2.1.0-SNAPSHOT`???

kubernetes/pom.xml
```diff
@@ -0,0 +1,54 @@
...
+    <relativePath>../pom.xml</relativePath>
+  </parent>
+
+  <artifactId>spark-kubernetes_2.11</artifactId>
```
jaceklaskowski: `${scala.binary.version}`

mccheah:
Work on this feature has moved to https://github.com/apache-spark-on-k8s/spark. The exact diff we are working with is this: https://github.com/apache-spark-on-k8s/spark/pull/1. Feel free to provide feedback on the fork's PR.

@erikerlandson can you close this PR?

foxish:
+1 I think we should close this PR to avoid confusion.

mccheah:
The diff showing our progress so far has moved to https://github.com/apache-spark-on-k8s/spark/pull/200 for anyone following along.

PR #16062, opened 2016-11-29
[SPARK-18629][SQL] Fix numPartition of JDBCSuite Testcase

weiqingy:
Fix numPartition of JDBCSuite Testcase. 

## How was this patch tested?
Before:
Run any one of the test cases in JDBCSuite, you will get the following warning.
```
10:34:26.389 WARN org.apache.spark.sql.execution.datasources.jdbc.JDBCRelation: The number of partitions is reduced because the specified number of partitions is less than the difference between upper bound and lower bound. Updated number of partitions: 3; Input number of partitions: 4; Lower bound: 1; Upper bound: 4.
```
After: Pass tests without the warning.


srowen (approved):
I think that's OK. CC @sureshthalamati

weiqingy:
@srowen Thanks for the review.

srowen:
Merged to master

PR #16063, opened 2016-11-29
[SPARK-18622][SQL] Fix the datatype of the Sum aggregate function

hvanhovell:
The result of a `sum` aggregate function is typically a Decimal, Double or a Long. Currently the output dataType is based on input's dataType.

The `FunctionArgumentConversion` rule will make sure that the input is promoted to the largest type, and that also ensures that the output uses a (hopefully) sufficiently large output dataType. The issue is that sum is in a resolved state when we cast the input type, this means that rules assuming that the dataType of the expression does not change anymore could have been applied in the mean time. This is what happens if we apply `WidenSetOperationTypes` before applying the casts, and this breaks analysis.

The most straight forward and future proof solution is to make `sum` always output the widest dataType in its class (Long for IntegralTypes, Decimal for DecimalTypes & Double for FloatType and DoubleType). This PR implements that solution.

We should move expression specific type casting rules into the given Expression at some point.

## How was this patch tested?
Added (regression) tests to SQLQueryTestSuite's `union.sql`.


hvanhovell:
cc @gatorsmile @cloud-fan could you take a look at this?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/TypeCoercion.scala
```diff
@@ -482,21 +482,6 @@ object TypeCoercion {
...
-
-      case s @ Average(e @ DecimalType()) => s // Decimal is already the biggest.
-      case Average(e @ IntegralType()) if e.dataType != LongType =>
-        Average(Cast(e, LongType))
-      case Average(e @ FractionalType()) if e.dataType != DoubleType =>
-        Average(Cast(e, DoubleType))
-
-      // Hive lets you do aggregation of timestamps... for some reason
-      case Sum(e @ TimestampType()) => Sum(Cast(e, DoubleType))
-      case Average(e @ TimestampType()) => Average(Cast(e, DoubleType))
```
dongjoon-hyun: Is this safe in terms of backward-compatibility for `Timestamp`?

hvanhovell: I was surprised to find out that we actually support this. For example:
```scala
val df = spark.range(1000).select((current_timestamp() + concat(lit("interval "), $"id", lit(" days")).cast(CalendarIntervalType)).as("time"))
df.groupBy().agg(sum($"time")).show()

+--------------------+
|           sum(time)|
+--------------------+
|1.523614355600014...|
+--------------------+
```
Back to the drawing board :)

gatorsmile:
LGTM cc @cloud-fan for double check

cloud-fan:
LGTM, merging to master/2.1!

kkyon:
I have  a similar error in spark 2.3.4  . the issue is not yet resolved . I open a new issue . can you look it  ?  @hvanhovell 
https://issues.apache.org/jira/browse/SPARK-32638

PR #16068, opened 2016-11-29
[SPARK-18637][SQL]Stateful UDF should be considered as nondeterministic

zhzhan:

Make stateful udf as nondeterministic

## How was this patch tested?
Add new test cases with both Stateful and Stateless UDF.
Without the patch, the test cases will throw exception:

1 did not equal 10
ScalaTestFailureLocation: org.apache.spark.sql.hive.execution.HiveUDFSuite$$anonfun$21 at (HiveUDFSuite.scala:501)
org.scalatest.exceptions.TestFailedException: 1 did not equal 10
        at org.scalatest.Assertions$class.newAssertionFailedException(Assertions.scala:500)
        at org.scalatest.FunSuite.newAssertionFailedException(FunSuite.scala:1555)
        ...

### DIFF
--- a/sql/hive/src/main/scala/org/apache/spark/sql/hive/hiveUDFs.scala
+++ b/sql/hive/src/main/scala/org/apache/spark/sql/hive/hiveUDFs.scala
@@ -61,7 +61,7 @@ private[hive] case class HiveSimpleUDF(
   @transient
   private lazy val isUDFDeterministic = {
     val udfType = function.getClass.getAnnotation(classOf[HiveUDFType])
-    udfType != null && udfType.deterministic()
+    udfType != null && udfType.deterministic() && !udfType.stateful()
   }
 
   override def foldable: Boolean = isUDFDeterministic && children.forall(_.foldable)
@@ -144,7 +144,7 @@ private[hive] case class HiveGenericUDF(
   @transient
   private lazy val isUDFDeterministic = {
     val udfType = function.getClass.getAnnotation(classOf[HiveUDFType])
-    udfType != null && udfType.deterministic()
+    udfType != null && udfType.deterministic() && !udfType.stateful()
   }
 
   @transient
--- a/sql/hive/src/test/scala/org/apache/spark/sql/hive/execution/HiveUDFSuite.scala
+++ b/sql/hive/src/test/scala/org/apache/spark/sql/hive/execution/HiveUDFSuite.scala
@@ -495,16 +495,14 @@ class HiveUDFSuite extends QueryTest with TestHiveSingleton with SQLTestUtils {
       sql(s"CREATE TEMPORARY FUNCTION statefulUDF AS '${classOf[StatefulUDF].getName}'")
       sql(s"CREATE TEMPORARY FUNCTION statelessUDF AS '${classOf[StatelessUDF].getName}'")
       val testData = spark.range(10).repartition(1)
-      val testData1 = spark.range(10).repartition(2)
 
       // Expected Max(s) is 10 as statefulUDF returns the sequence number starting from 1.
       checkAnswer(testData.selectExpr("statefulUDF() as s").agg(max($"s")), Row(10))
 
-      // Expected Max(s) is 10 as statefulUDF returns the sequence number starting from 1.
-      // Note that testData.union(testData) will have Row(20) because internally it is converted
-      // to PartitionerAwareUnionRDD with 1 partition as they are having same partitioning.
-      checkAnswer(testData.union(testData1)
-        .selectExpr("statefulUDF() as s").agg(max($"s")), Row(10))
+      // Expected Max(s) is 5 as statefulUDF returns the sequence number starting from 1,
+      // and the data is evenly distributed into 2 partitions.
+      checkAnswer(testData.repartition(2)
+        .selectExpr("statefulUDF() as s").agg(max($"s")), Row(5))
 
       // Expected Max(s) is 1, as stateless UDF is deterministic and foldable and replaced
       // by constant 1 by ConstantFolding optimizer.
--- a/sql/hive/src/test/scala/org/apache/spark/sql/hive/execution/HiveUDFSuite.scala
+++ b/sql/hive/src/test/scala/org/apache/spark/sql/hive/execution/HiveUDFSuite.scala
@@ -31,6 +31,7 @@ import org.apache.hadoop.hive.serde2.objectinspector.primitive.PrimitiveObjectIn
 import org.apache.hadoop.io.{LongWritable, Writable}
 
 import org.apache.spark.sql.{AnalysisException, QueryTest, Row}
+import org.apache.spark.sql.functions.max
 import org.apache.spark.sql.hive.test.TestHiveSingleton
 import org.apache.spark.sql.test.SQLTestUtils
 import org.apache.spark.util.Utils
@@ -493,45 +494,21 @@ class HiveUDFSuite extends QueryTest with TestHiveSingleton with SQLTestUtils {
     withUserDefinedFunction("statefulUDF" -> true, "statelessUDF" -> true) {
       sql(s"CREATE TEMPORARY FUNCTION statefulUDF AS '${classOf[StatefulUDF].getName}'")
       sql(s"CREATE TEMPORARY FUNCTION statelessUDF AS '${classOf[StatelessUDF].getName}'")
-      withTempView("inputTable") {
-        val testData = spark.sparkContext.parallelize(
-          (0 until 10) map (x => IntegerCaseClass(1)), 2).toDF()
-        testData.createOrReplaceTempView("inputTable")
-        // Distribute all rows to one partition (all rows have the same content),
-        // and expected Max(s) is 10 as statefulUDF returns the sequence number starting from 1.
-        checkAnswer(
-          sql(
-            """
-            |SELECT MAX(s) FROM
-            |  (SELECT statefulUDF() as s FROM
-            |    (SELECT i from inputTable DISTRIBUTE by i) a
-            |    ) b
-          """.stripMargin),
-          Row(10))
+      val testData = spark.range(10).repartition(1)
+      val testData1 = spark.range(10).repartition(2)
 
-        // Expected Max(s) is 5, as there are 2 partitions with 5 rows each, and statefulUDF
-        // returns the sequence number of the rows in the partition starting from 1.
-        checkAnswer(
-          sql(
-            """
-              |SELECT MAX(s) FROM
-              |  (SELECT statefulUDF() as s FROM
-              |    (SELECT i from inputTable) a
-              |    ) b
-            """.stripMargin),
-          Row(5))
+      // Expected Max(s) is 10 as statefulUDF returns the sequence number starting from 1.
+      checkAnswer(testData.selectExpr("statefulUDF() as s").agg(max($"s")), Row(10))
 
-        // Expected Max(s) is 1, as stateless UDF is deterministic and replaced by constant 1.
-        checkAnswer(
-          sql(
-            """
-              |SELECT MAX(s) FROM
-              |  (SELECT statelessUDF() as s FROM
-              |    (SELECT i from inputTable DISTRIBUTE by i) a
-              |    ) b
-            """.stripMargin),
-          Row(1))
-      }
+      // Expected Max(s) is 10 as statefulUDF returns the sequence number starting from 1.
+      // Note that testData.union(testData) will have Row(20) because internally it is converted
+      // to PartitionerAwareUnionRDD with 1 partition as they are having same partitioning.
+      checkAnswer(testData.union(testData1)
+        .selectExpr("statefulUDF() as s").agg(max($"s")), Row(10))
+
+      // Expected Max(s) is 1, as stateless UDF is deterministic and foldable and replaced
+      // by constant 1 by ConstantFolding optimizer.
+      checkAnswer(testData.selectExpr("statelessUDF() as s").agg(max($"s")), Row(1))
     }
   }
 }
--- a/sql/hive/src/test/scala/org/apache/spark/sql/hive/execution/HiveUDFSuite.scala
+++ b/sql/hive/src/test/scala/org/apache/spark/sql/hive/execution/HiveUDFSuite.scala
@@ -578,20 +578,18 @@ class PairUDF extends GenericUDF {
 
 @UDFType(stateful = true)
 class StatefulUDF extends UDF {
-  private val result = new LongWritable()
-  result.set(0)
+  private val result = new LongWritable(0)
 
-  def  evaluate(): LongWritable = {
+  def evaluate(): LongWritable = {
     result.set(result.get() + 1)
     result
   }
 }
 
 class StatelessUDF extends UDF {
-  private val result = new LongWritable()
-  result.set(0)
+  private val result = new LongWritable(0)
 
-  def  evaluate(): LongWritable = {
+  def evaluate(): LongWritable = {
     result.set(result.get() + 1)
     result
   }
--- a/sql/hive/src/test/scala/org/apache/spark/sql/hive/execution/HiveUDFSuite.scala
+++ b/sql/hive/src/test/scala/org/apache/spark/sql/hive/execution/HiveUDFSuite.scala
@@ -490,40 +490,42 @@ class HiveUDFSuite extends QueryTest with TestHiveSingleton with SQLTestUtils {
   }
 
   test("Hive Stateful UDF") {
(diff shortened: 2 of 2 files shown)


zhzhan:
retest it please

hvanhovell:
retest this please

hvanhovell:
@zhzhan you really should add a regression suite. Maybe HiveUDFSuite is a good place for this.

Are there any existing Hive UDFs that have this problem?

zhzhan:
@hvanhovell Thanks for looking at this. We have a big number of UDFs that have this issue. For example, the UDF gives different result with different partition/sort, but the UDF is pushdown before the partition/sort, resulting in unexpected behavior. I will working on finding some test cases for it.

zhzhan:
@hvanhovell Would you like take a look and let me know if you have any concern?

gatorsmile:
We have a rule [`PullOutNondeterministic`](https://github.com/apache/spark/blob/2f8776ccad532fbed17381ff97d302007918b8d8/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala#L1868
) to pull out the expressions that extend `Nondeterministic`. That does not include `HiveSimpleUDF`. Does this have an impact your case?

Also cc @hvanhovell @cloud-fan

sql/hive/src/test/scala/org/apache/spark/sql/hive/execution/HiveUDFSuite.scala
```diff
@@ -487,6 +488,29 @@ class HiveUDFSuite extends QueryTest with TestHiveSingleton with SQLTestUtils {
...
+  test("Hive Stateful UDF") {
+    sql(s"CREATE TEMPORARY FUNCTION statefulUDF AS '${classOf[StatefulUDF].getName}'")
+    sql(s"CREATE TEMPORARY FUNCTION statelessUDF AS '${classOf[StatelessUDF].getName}'")
+    val testData = spark.sparkContext.parallelize(
+      (0 until 10) map(x => IntegerCaseClass(1)), 2).toDF()
+    testData.createOrReplaceTempView("inputTable")
+    val max1 =
+      sql("SELECT MAX(s) FROM (" +
+        "SELECT statefulUDF() as s FROM (SELECT i from inputTable DISTRIBUTE by i) a" +
+        ") b").head().getLong(0)
```
gatorsmile: How about using multiline string syntax here?
```Scala
    sql(
      """
        |SELECT MAX(s) FROM
        |  (SELECT statefulUDF() as s FROM
        |    (SELECT i from inputTable DISTRIBUTE BY i) a
        |  ) b
      """.stripMargin)
```

sql/hive/src/main/scala/org/apache/spark/sql/hive/hiveUDFs.scala
```diff
@@ -144,7 +144,7 @@ private[hive] case class HiveGenericUDF(
   @transient
   private lazy val isUDFDeterministic = {
     val udfType = function.getClass.getAnnotation(classOf[HiveUDFType])
-    udfType != null && udfType.deterministic()
+    udfType != null && udfType.deterministic() && !udfType.stateful()
```
cloud-fan: an unrelated question, what's the difference between `udfType.deterministic` and `udfType.stateful`?

zhzhan:
My understanding is that the non-deterministic udf does not need to be stageful, but a stateful udf has to be non-deterministic. 

Here is the comments in hive regarding this property

/**
If a UDF stores state based on the sequence of records it has processed, it
is stateful. A stateful UDF cannot be used in certain expressions such as
case statement and certain optimizations such as AND/OR short circuiting
don't apply for such UDFs, as they need to be invoked for each record.
row_sequence is an example of stateful UDF. A stateful UDF is considered to
be non-deterministic, irrespective of what deterministic() returns.
*
@return true
*/

gatorsmile:
Could we directly use `@UDFType(deterministic = true/false)`?

gatorsmile:
Found the link: [HIVE-1994: Support new annotation @UDFType(stateful = true)](https://issues.apache.org/jira/browse/HIVE-1994 )

gatorsmile:
https://github.com/apache/hive/blob/master/ql/src/java/org/apache/hadoop/hive/ql/exec/FunctionRegistry.java#L1373-L1378

Copied the code from Hive `FunctionRegistry.java`:
```JAVA
  /**
   * Returns whether a GenericUDF is deterministic or not.
   */
  public static boolean isDeterministic(GenericUDF genericUDF) {
    if (isStateful(genericUDF)) {
      // stateful implies non-deterministic, regardless of whatever
      // the deterministic annotation declares
      return false;
    }
    ...

  }
```

zhzhan:
@gatorsmile  we cannot use deterministic = true/false, as there are existing udf with deterministic as true, but stateful as true as well.

sql/hive/src/test/scala/org/apache/spark/sql/hive/execution/HiveUDFSuite.scala
zhzhan: will rewrite it after gathering feedback from others.

sql/hive/src/test/scala/org/apache/spark/sql/hive/execution/HiveUDFSuite.scala
```diff
@@ -487,6 +488,43 @@ class HiveUDFSuite extends QueryTest with TestHiveSingleton with SQLTestUtils {
     assert(count4 == 1)
     sql("DROP TABLE parquet_tmp")
   }
+
+  test("Hive Stateful UDF") {
+    sql(s"CREATE TEMPORARY FUNCTION statefulUDF AS '${classOf[StatefulUDF].getName}'")
```
gatorsmile: Could you use `withUserDefinedFunction` for these two functions?

gatorsmile:
LGTM except one comment on test cases. 

cc @hvanhovell @cloud-fan

sql/hive/src/test/scala/org/apache/spark/sql/hive/execution/HiveUDFSuite.scala
```diff
@@ -487,6 +488,45 @@ class HiveUDFSuite extends QueryTest with TestHiveSingleton with SQLTestUtils {
...
     sql("DROP TABLE parquet_tmp")
   }
+
+  test("Hive Stateful UDF") {
+    withUserDefinedFunction("statefulUDF" -> true, "statelessUDF" -> true) {
+      sql(s"CREATE TEMPORARY FUNCTION statefulUDF AS '${classOf[StatefulUDF].getName}'")
+      sql(s"CREATE TEMPORARY FUNCTION statelessUDF AS '${classOf[StatelessUDF].getName}'")
+      val testData = spark.sparkContext.parallelize(
+        (0 until 10) map (x => IntegerCaseClass(1)), 2).toDF()
+      testData.createOrReplaceTempView("inputTable")
```
cloud-fan: this temp view need to be cleaned after test

sql/hive/src/test/scala/org/apache/spark/sql/hive/execution/HiveUDFSuite.scala
```diff
@@ -487,6 +488,45 @@ class HiveUDFSuite extends QueryTest with TestHiveSingleton with SQLTestUtils {
...
+      checkAnswer(
+        sql(
+          """
+            |SELECT MAX(s) FROM
+            |  (SELECT statefulUDF() as s FROM
+            |    (SELECT i from inputTable DISTRIBUTE by i) a
+            |    ) b
+          """.stripMargin),
+        Row(10))
```
cloud-fan: can you add some comments to explain this test? why the expected result is 10?

sql/hive/src/test/scala/org/apache/spark/sql/hive/execution/HiveUDFSuite.scala
```diff
@@ -487,6 +488,52 @@ class HiveUDFSuite extends QueryTest with TestHiveSingleton with SQLTestUtils {
...
+
+  test("Hive Stateful UDF") {
+    withUserDefinedFunction("statefulUDF" -> true, "statelessUDF" -> true) {
+      sql(s"CREATE TEMPORARY FUNCTION statefulUDF AS '${classOf[StatefulUDF].getName}'")
+      sql(s"CREATE TEMPORARY FUNCTION statelessUDF AS '${classOf[StatelessUDF].getName}'")
+      withTempView("inputTable") {
+        val testData = spark.sparkContext.parallelize(
+          (0 until 10) map (x => IntegerCaseClass(1)), 2).toDF()
+        testData.createOrReplaceTempView("inputTable")
+        // Distribute all rows to one partition (all rows have the same content),
```
cloud-fan: Why `DISTRIBUTE BY` can distribute all rows to one partition? It's implemented by `RepartitionByExpression` which doesn't always use one partition.

sql/hive/src/test/scala/org/apache/spark/sql/hive/execution/HiveUDFSuite.scala
```diff
@@ -487,6 +488,52 @@ class HiveUDFSuite extends QueryTest with TestHiveSingleton with SQLTestUtils {
...
+          sql(
+            """
+              |SELECT MAX(s) FROM
+              |  (SELECT statefulUDF() as s FROM
+              |    (SELECT i from inputTable) a
+              |    ) b
+            """.stripMargin),
+          Row(5))
+
+        // Expected Max(s) is 1, as stateless UDF is deterministic and replaced by constant 1.
```
cloud-fan: why we will replace it with constant 1?

zhzhan: StatelessUDF is foldable:   override def foldable: Boolean = isUDFDeterministic && children.forall(_.foldable)

ConstantFolding optimizer will replace it with constant:
      case e if e.foldable => Literal.create(e.eval(EmptyRow), e.dataType)

Here is the explain(true):
== Parsed Logical Plan ==
'Project [unresolvedalias('MAX('s), None)]
+- 'SubqueryAlias b
   +- 'Project ['statelessUDF() AS s#39]
      +- 'SubqueryAlias a
         +- 'RepartitionByExpression ['i]
            +- 'Project ['i]
               +- 'UnresolvedRelation `inputTable`

== Analyzed Logical Plan ==
max(s): bigint
Aggregate [max(s#39L) AS max(s)#46L]
+- SubqueryAlias b
   +- Project [HiveSimpleUDF#org.apache.spark.sql.hive.execution.StatelessUDF() AS s#39L]
      +- SubqueryAlias a
         +- RepartitionByExpression [i#4]
            +- Project [i#4]
               +- SubqueryAlias inputtable
                  +- SerializeFromObject [assertnotnull(assertnotnull(input[0, org.apache.spark.sql.hive.execution.IntegerCaseClass, true], top level Product input object), - root class: "org.apache.spark.sql.hive.execution.IntegerCaseClass").i AS i#4]
                     +- ExternalRDD [obj#3]

== Optimized Logical Plan ==
Aggregate [max(s#39L) AS max(s)#46L]
+- Project [1 AS s#39L]
   +- RepartitionByExpression [i#4]
      +- SerializeFromObject [assertnotnull(assertnotnull(input[0, org.apache.spark.sql.hive.execution.IntegerCaseClass, true], top level Product input object), - root class: "org.apache.spark.sql.hive.execution.IntegerCaseClass").i AS i#4]
         +- ExternalRDD [obj#3]

== Physical Plan ==
*HashAggregate(keys=[], functions=[max(s#39L)], output=[max(s)#46L])
+- Exchange SinglePartition
   +- *HashAggregate(keys=[], functions=[partial_max(s#39L)], output=[max#48L])
      +- *Project [1 AS s#39L]
         +- Exchange hashpartitioning(i#4, 5)
            +- *SerializeFromObject [assertnotnull(assertnotnull(input[0, org.apache.spark.sql.hive.execution.IntegerCaseClass, true], top level Product input object), - root class: "org.apache.spark.sql.hive.execution.IntegerCaseClass").i AS i#4]
               +- Scan ExternalRDDScan[obj#3]

sql/hive/src/test/scala/org/apache/spark/sql/hive/execution/HiveUDFSuite.scala
zhzhan: @cloud-fan  Thanks for the review. Because all rows only contains IntegerCaseClass(1), RepartitionByExpression will assign all rows to one partition, which has 10 records.

sql/hive/src/test/scala/org/apache/spark/sql/hive/execution/HiveUDFSuite.scala
```diff
@@ -487,6 +488,52 @@ class HiveUDFSuite extends QueryTest with TestHiveSingleton with SQLTestUtils {
     assert(count4 == 1)
     sql("DROP TABLE parquet_tmp")
   }
+
+  test("Hive Stateful UDF") {
+    withUserDefinedFunction("statefulUDF" -> true, "statelessUDF" -> true) {
+      sql(s"CREATE TEMPORARY FUNCTION statefulUDF AS '${classOf[StatefulUDF].getName}'")
+      sql(s"CREATE TEMPORARY FUNCTION statelessUDF AS '${classOf[StatelessUDF].getName}'")
+      withTempView("inputTable") {
```
cloud-fan: we can simplify this test a bit:
```
val testData = spark.range(10)
checkAnswer(testData.repartition(1).select("statefulUDF() as s").max($"s"), Row(10))
...
```

gatorsmile:
```Scala
        val testData = spark.range(10).repartition(1)
        testData.selectExpr("statefulUDF() as s")
```

Try this?

sql/hive/src/test/scala/org/apache/spark/sql/hive/execution/HiveUDFSuite.scala
```diff
@@ -487,6 +489,28 @@ class HiveUDFSuite extends QueryTest with TestHiveSingleton with SQLTestUtils {
...
+      sql(s"CREATE TEMPORARY FUNCTION statefulUDF AS '${classOf[StatefulUDF].getName}'")
+      sql(s"CREATE TEMPORARY FUNCTION statelessUDF AS '${classOf[StatelessUDF].getName}'")
+      val testData = spark.range(10).repartition(1)
+      val testData1 = spark.range(10).repartition(2)
+
+      // Expected Max(s) is 10 as statefulUDF returns the sequence number starting from 1.
+      checkAnswer(testData.selectExpr("statefulUDF() as s").agg(max($"s")), Row(10))
+
+      // Expected Max(s) is 10 as statefulUDF returns the sequence number starting from 1.
+      // Note that testData.union(testData) will have Row(20) because internally it is converted
```
cloud-fan: if you need to write comments to explain some technical problems for the test, it would be better to rewrite the test and avoid this problem. How about:
```
checkAnswer(spark.range(10).repartition(2, $"id" < 5).selectExpr("statefulUDF() as s").agg(max($"s")), Row(5))
```

sql/hive/src/test/scala/org/apache/spark/sql/hive/execution/HiveUDFSuite.scala
```diff
@@ -487,6 +489,26 @@ class HiveUDFSuite extends QueryTest with TestHiveSingleton with SQLTestUtils {
...
+    withUserDefinedFunction("statefulUDF" -> true, "statelessUDF" -> true) {
+      sql(s"CREATE TEMPORARY FUNCTION statefulUDF AS '${classOf[StatefulUDF].getName}'")
+      sql(s"CREATE TEMPORARY FUNCTION statelessUDF AS '${classOf[StatelessUDF].getName}'")
+      val testData = spark.range(10).repartition(1)
+
+      // Expected Max(s) is 10 as statefulUDF returns the sequence number starting from 1.
+      checkAnswer(testData.selectExpr("statefulUDF() as s").agg(max($"s")), Row(10))
+
+      // Expected Max(s) is 5 as statefulUDF returns the sequence number starting from 1,
+      // and the data is evenly distributed into 2 partitions.
```
cloud-fan: `repartition(2)` doesn't guarantee even distribution, it depends on the hash implementation used in shuffle, that's why I suggest `repartition(2, $"id" < 5)`.

cloud-fan: oh you are right, I misread the code.

cloud-fan:

PR #16071, opened 2016-11-30
[SPARK-18635] [SQL] Partition name/values not escaped correctly in some cases

ericl:

Due to confusion between URI vs paths, in certain cases we escape partition values too many times, which causes some Hive client operations to fail or write data to the wrong location. This PR fixes at least some of these cases.

To my understanding this is how values, filesystem paths, and URIs interact.
- Hive stores raw (unescaped) partition values that are returned to you directly when you call listPartitions.
- Internally, we convert these raw values to filesystem paths via `ExternalCatalogUtils.[un]escapePathName`.
- In some circumstances we store URIs instead of filesystem paths. When a path is converted to a URI via `path.toURI`, the escaped partition values are further URI-encoded. This means that to get a path back from a URI, you must call `new Path(new URI(uriTxt))` in order to decode the URI-encoded string.
- In `CatalogStorageFormat` we store URIs as strings. This makes it easy to forget to URI-decode the value before converting it into a path.
- Finally, the Hive client itself uses mostly Paths for representing locations, and only URIs occasionally.

In the future we should probably clean this up, perhaps by dropping use of URIs when unnecessary. We should also try fixing escaping for partition names as well as values, though names are unlikely to contain special characters.

cc @mallman @cloud-fan @yhuai 

## How was this patch tested?

Unit tests.

### DIFF
--- a/sql/hive/src/test/scala/org/apache/spark/sql/hive/PartitionProviderCompatibilitySuite.scala
+++ b/sql/hive/src/test/scala/org/apache/spark/sql/hive/PartitionProviderCompatibilitySuite.scala
@@ -231,30 +231,32 @@ class PartitionProviderCompatibilitySuite
         assert(parts.contains("A=0/B=%3D"))
         assert(parts.contains("A=0/B=%25%3D"))
 
-        // custom locations sanity check
+        // drop partition sanity check
+        spark.sql("alter table test drop partition (A=1, B='%')")
+        assert(spark.sql("select * from test").count() == 29)  // 1 file in dropped partition
+
         withTempDir { dir =>
+          // custom locations sanity check
           spark.sql(s"""
             |alter table test partition (A=0, B='%')
             |set location '${dir.getAbsolutePath}'""".stripMargin)
-          assert(spark.sql("select * from test").count() == 29)  // missing 1
+          assert(spark.sql("select * from test").count() == 28)  // moved to empty dir
+
+          // rename partition sanity check
+          spark.sql(s"""
+            |alter table test partition (A=5, B='%')
+            |rename to partition (A=100, B='%')""".stripMargin)
+          assert(spark.sql("select * from test where a = 5 and b = '%'").count() == 0)
+          assert(spark.sql("select * from test where a = 100 and b = '%'").count() == 1)
+
+          // try with A=0 which has a custom location
+          spark.sql("insert into test partition (A=0, B='%') select 1")
+          spark.sql(s"""
+            |alter table test partition (A=0, B='%')
+            |rename to partition (A=101, B='%')""".stripMargin)
+          assert(spark.sql("select * from test where a = 0 and b = '%'").count() == 0)
+          assert(spark.sql("select * from test where a = 101 and b = '%'").count() == 1)
         }
-
-        // drop partition sanity check
-        spark.sql("alter table test drop partition (A=1, B='%')")
-        assert(spark.sql("select * from test").count() == 28)
-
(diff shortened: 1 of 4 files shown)


sql/hive/src/test/scala/org/apache/spark/sql/hive/PartitionProviderCompatibilitySuite.scala
```diff
@@ -205,6 +205,58 @@ class PartitionProviderCompatibilitySuite
...
+        // drop partition sanity check
+        spark.sql("alter table test drop partition (A=1, B='%')")
+        assert(spark.sql("select * from test").count() == 28)
+
+        // TODO(ekl) fix rename partition
+//        withTempDir { dir =>
+//          spark.sql(s"""
+//            |alter table test partition (A=0, B='%')
+//            |rename to partition (A=100, B='%')""".stripMargin)
+//          assert(spark.sql("select * from test where a = 100").count() == 1)
```
ericl: @cloud-fan this seems to crash in some of the post-processing code in`HiveExternalCatalog:renamePartitions`. Any ideas there?

sql/hive/src/test/scala/org/apache/spark/sql/hive/PartitionProviderCompatibilitySuite.scala
```diff
@@ -205,6 +205,58 @@ class PartitionProviderCompatibilitySuite
...
+//        }
+
+        // TODO(ekl) fix overwrite table
+//        spark.sql("show partitions test").show(false)
+//        spark.sql("insert overwrite table test partition (a, b) select id, id, '%' from range(1)")
```
ericl: This crashes in the hive client. Not sure why, it might be a hive bug when processing partition values containing `=`.

sql/hive/src/test/scala/org/apache/spark/sql/hive/PartitionProviderCompatibilitySuite.scala
```diff
@@ -205,6 +205,58 @@ class PartitionProviderCompatibilitySuite
...
+        assert(parts.contains("A=0/B=%25"))
+        assert(parts.contains("A=0/B=%3D"))
+        assert(parts.contains("A=0/B=%25%3D"))
+
+        // custom locations sanity check
+        withTempDir { dir =>
+          spark.sql(s"""
+            |alter table test partition (A=0, B='%')
+            |set location '${dir.getAbsolutePath}'""".stripMargin)
+          assert(spark.sql("select * from test").count() == 29)  // missing 1
```
mallman: Why is this 29 instead of 30?

ericl: Since the partition was moved, there are no files in the new location. Hence -1 file

mallman: Ah. I see. So Hive/Spark doesn't move partition data when the partition location is changed. Might be clearer to call that out in your comment rather than just "missing 1".

mallman:
I can't vouch for how `Path` and `URI` work together to do the right thing, however the test coverage looks good. LGTM overall.

ericl:
Ok I've figured out issues with the other test cases, they seem to be due to a separate bug: https://issues.apache.org/jira/browse/SPARK-18659

cloud-fan:
LGTM, merging to master/2.1!

PR #16072, opened 2016-11-30
[SPARK-18639] Build only a single pip package

rxin:
We current build 5 separate pip binary tar balls, doubling the release script runtime. It'd be better to build one, especially for use cases that are just using Spark locally. In the long run, it would make more sense to have Hadoop support be pluggable.

## How was this patch tested?
N/A - this is a release build script that doesn't have any automated test coverage. We will know if it goes wrong when we prepare releases.

### DIFF
--- a/dev/create-release/release-build.sh
+++ b/dev/create-release/release-build.sh
@@ -150,6 +150,7 @@ if [[ "$1" == "package" ]]; then
     NAME=$1
     FLAGS=$2
     ZINC_PORT=$3
+    BUILD_PIP_PACKAGE=$4
     cp -r spark spark-$SPARK_VERSION-bin-$NAME
 
     cd spark-$SPARK_VERSION-bin-$NAME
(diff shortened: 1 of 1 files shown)


dev/create-release/release-build.sh
```diff
@@ -150,6 +150,7 @@ if [[ "$1" == "package" ]]; then
     NAME=$1
     FLAGS=$2
     ZINC_PORT=$3
+    BUILD_PIP_PACKAGE=$4
```
JoshRosen: This style of positional boolean flag won't really work if you ever need to have more than one of these arguments. That said, this is fine for now and if we're really going to add any more complexity here then we should probably just rewrite this whole script in Python.

JoshRosen (approved):
This looks fine to me, especially since the non-whitespace-change diff is really tiny: https://github.com/apache/spark/pull/16072/files?w=0

dongjoon-hyun:
Retest this please

rxin:
Merging in master/branch-2.1.

PR #16073, opened 2016-11-30
[SPARK-18640] Add synchronization to TaskScheduler.runningTasksByExecutors

JoshRosen:

The method `TaskSchedulerImpl.runningTasksByExecutors()` accesses the mutable `executorIdToRunningTaskIds` map without proper synchronization. In addition, as @markhamstra pointed out in #15986, the signature's use of parentheses is a little odd given that this is a pure getter method.

This patch fixes both issues.

## How was this patch tested?

Covered by existing tests.


andrewor14:
LGTM merging into master 2.1.

andrewor14:
and 2.0

PR #16076, opened 2016-11-30
[SPARK-18324][ML][DOC] Update ML programming and migration guide for 2.1 release

yanboliang:
Update ML programming and migration guide for 2.1 release.

## How was this patch tested?
Doc change, no test.

### DIFF
--- a/docs/ml-guide.md
+++ b/docs/ml-guide.md
@@ -60,152 +60,37 @@ MLlib is under active development.
 The APIs marked `Experimental`/`DeveloperApi` may change in future releases,
 and the migration guide below will explain all changes between releases.
 
-## From 1.6 to 2.0
+## From 2.0 to 2.1
 
 ### Breaking changes
 
-There were several breaking changes in Spark 2.0, which are outlined below.
-
-**Linear algebra classes for DataFrame-based APIs**
-
-Spark's linear algebra dependencies were moved to a new project, `mllib-local` 
-(see [SPARK-13944](https://issues.apache.org/jira/browse/SPARK-13944)). 
-As part of this change, the linear algebra classes were copied to a new package, `spark.ml.linalg`. 
-The DataFrame-based APIs in `spark.ml` now depend on the `spark.ml.linalg` classes, 
-leading to a few breaking changes, predominantly in various model classes 
-(see [SPARK-14810](https://issues.apache.org/jira/browse/SPARK-14810) for a full list).
-
-**Note:** the RDD-based APIs in `spark.mllib` continue to depend on the previous package `spark.mllib.linalg`.
-
-_Converting vectors and matrices_
-
-While most pipeline components support backward compatibility for loading, 
-some existing `DataFrames` and pipelines in Spark versions prior to 2.0, that contain vector or matrix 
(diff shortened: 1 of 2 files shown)


yanboliang:
cc @jkbradley @MLnick @sethah

jkbradley (commented):
Just a few comments.

One more: We need to note [https://issues.apache.org/jira/browse/SPARK-18291] or revert that change in order to avoid breaking APIs.  CC @mengxr

docs/ml-guide.md
```diff
@@ -60,152 +60,37 @@ MLlib is under active development.
...
-* [SPARK-13600](https://issues.apache.org/jira/browse/SPARK-13600):
- `QuantileDiscretizer` now uses `spark.sql.DataFrameStatFunctions.approxQuantile` to find splits (previously used custom sampling logic).
- The output buckets will differ for same input data and params.
+* [SPARK-17870](https://issues.apache.org/jira/browse/SPARK-17870):
+ Fix a bug of `ChiSqSelector` which will likely change its result.
+* [SPARK-3261](https://issues.apache.org/jira/browse/SPARK-3261):
+ `KMeans` returns potentially fewer than k cluster centers in cases where k distinct centroids aren't available or aren't selected.
+* [SPARK-17389](https://issues.apache.org/jira/browse/SPARK-17389):
+ `KMeans` reduces the default number of steps from 5 to 2 for the k-means|| initialization mode.
+* [SPARK-18481](https://issues.apache.org/jira/browse/SPARK-18481):
```
yanboliang: I moved it to breaking changes session. I think the Params setter methods are rarely used, since it take no effect and should not be exposed for model classes (This is also why we deprecate them from 2.1). Even they are used, it's no breaking changes for Scala users. And for Java users, they always should make forced type conversion, such as:
```
RandomForestClassificationModel model = (RandomForestClassificationModel)rf.fit(dataFrame).setFeatureSubsetStrategy("sqrt");
```
So I think it has little affect and actually no breaking happen, so I think it's OK. But I'm open to hear your thoughts.

yanboliang:
@jkbradley I addressed other comments except for [SPARK-18291](https://issues.apache.org/jira/browse/SPARK-18291), since I think it's SparkR related issue and should be listed at [SparkR session of migration guide](http://spark.apache.org/docs/latest/sparkr.html#migration-guide). I'm preparing a PR to update SparkR migration guide. Thanks.

docs/ml-guide.md
jkbradley: Ah, you're right.  I was thinking the "final" change was breaking for Scala, but it's not since the model constructor is private.  I'm OK with moving those back to changes of behavior, or just removing these items completely.

jkbradley:
Sounds good about SPARK-18291.

I responded inline above about SPARK-18481.  Apart from this update, this looks ready to me.  Thank you!

docs/ml-guide.md
yanboliang: OK, I remove these completely. Thanks.

jkbradley:
LGTM
I'll merge this with master and branch-2.1
Thank you!

PR #16077, opened 2016-11-30
[SPARK-18643][SPARKR] SparkR hangs at session start when installed as a package without Spark

felixcheung:

If SparkR is running as a package and it has previously downloaded Spark Jar it should be able to run as before without having to set SPARK_HOME. Basically with this bug the auto install Spark will only work in the first session.

This seems to be a regression on the earlier behavior.

Fix is to always try to install or check for the cached Spark if running in an interactive session.
As discussed before, we should probably only install Spark iff running in an interactive session (R shell, RStudio etc)

## How was this patch tested?

Manually

### DIFF
--- a/R/pkg/R/sparkR.R
+++ b/R/pkg/R/sparkR.R
@@ -565,7 +565,7 @@ sparkCheckInstall <- function(sparkHome, master, deployMode) {
       message(msg)
       NULL
     } else {
-      if (isMasterLocal(master)) {
+      if (interactive() || isMasterLocal(master)) {
         msg <- paste0("Spark not found in SPARK_HOME: ", sparkHome)
         message(msg)
         packageLocalDir <- install.spark()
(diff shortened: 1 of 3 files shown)


felixcheung:
@yanboliang @shivaram

yanboliang:
Is this means that it will install spark package each time when starting a new session in interactive mode? Thanks.

felixcheung:
install.spark() checks if SPARK_HOME is set or the Spark Jar can be found in the cache location.

yanboliang:
LGTM

shivaram:
Just to clarify this limits the auto-install feature to SparkR running on an interactive shell -- Is that right ? I think thats mostly fine as its the use-case we were targeting but it might good to check our documentation and update it appropriately.

felixcheung:
Right, I was just reviewing possible code paths for this in the last few days and I'm pretty confident that this change will not run install.spark in cluster modes (which would have Spark/JVM already running).

Also you are right, I just found that we didn't really talk about how install.spark would be called in sparkR.session() - I'll add that.

shivaram:
LGTM. Merging this to master, branch-2.1

PR #16080, opened 2016-11-30
[SPARK-18647][SQL] do not put provider in table properties for Hive serde table

cloud-fan:

In Spark 2.1, we make Hive serde tables case-preserving by putting the table metadata in table properties, like what we did for data source table. However, we should not put table provider, as it will break forward compatibility. e.g. if we create a Hive serde table with Spark 2.1, using `sql("create table test stored as parquet as select 1")`, we will fail to read it with Spark 2.0, as Spark 2.0 mistakenly treat it as data source table because there is a `provider` entry in table properties.

Logically Hive serde table's provider is always hive, we don't need to store it in table properties, this PR removes it.

## How was this patch tested?

manually test the forward compatibility issue.

### DIFF
--- a/sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveExternalCatalog.scala
+++ b/sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveExternalCatalog.scala
@@ -20,6 +20,7 @@ package org.apache.spark.sql.hive
 import java.io.IOException
 import java.util
 
+import scala.collection.mutable
 import scala.util.control.NonFatal
 
 import org.apache.hadoop.conf.Configuration
@@ -218,9 +219,9 @@ private[spark] class HiveExternalCatalog(conf: SparkConf, hadoopConf: Configurat
           // table location for tables in default database, while we expect to use the location of
           // default database.
           storage = tableDefinition.storage.copy(locationUri = tableLocation),
-          // Here we follow data source tables and put table metadata like provider, schema, etc. in
-          // table properties, so that we can work around the Hive metastore issue about not case
-          // preserving and make Hive serde table support mixed-case column names.
+          // Here we follow data source tables and put table metadata like table schema, partition
+          // columns etc. in table properties, so that we can work around the Hive metastore issue
+          // about not case preserving and make Hive serde table support mixed-case column names.
           properties = tableDefinition.properties ++ tableMetaToTableProps(tableDefinition))
         client.createTable(tableWithDataSourceProps, ignoreIfExists)
       } else {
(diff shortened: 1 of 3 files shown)


cloud-fan:
cc @rxin @yhuai @mallman

sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveExternalCatalog.scala
```diff
@@ -232,17 +233,26 @@ private[spark] class HiveExternalCatalog(conf: SparkConf, hadoopConf: Configurat
...
+    //  1. Put table metadata like table schema, partition columns, etc. in table properties.
     //  2. Check if this table is hive compatible.
     //    2.1  If it's not hive compatible, set location URI, schema, partition columns and bucket
     //         spec to empty and save table metadata to Hive.
     //    2.2  If it's hive compatible, set serde information in table metadata and try to save
     //         it to Hive. If it fails, treat it as not hive compatible and go back to 2.1
     val tableProperties = tableMetaToTableProps(table)
 
+    // put table provider and partition provider in table properties.
+    tableProperties.put(DATASOURCE_PROVIDER, provider)
```
mallman: Why are we putting the provider name in the table properties here?

mallman:
I built and tested this branch, and it resolves the issue I was having with reading Spark 2.1 tables in earlier versions of Spark. Thanks!

rxin:
We need a test case here.

gatorsmile:
We need to provide forward compatibility? That is pretty hard.

rxin:
In a lot of environments people run multiple Spark versions side by side. That's always been a big strength of Spark.

sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveExternalCatalog.scala
cloud-fan: Previously we store the provider in the code path for both data source and hive serde tables. Now I move it to the data source table only code path.

cloud-fan:
@rxin I'm afraid it's hard to write forward compatibility tests using unit test, we may need an external test infrastructure(python scripts) to do this.

rxin:
We can have a test to check the table properties don't contain the entry, can't we?

gatorsmile:
I see. Will be careful in the future to not break the forward compatibility.

sql/hive/src/test/scala/org/apache/spark/sql/hive/HiveExternalCatalogSuite.scala
```diff
@@ -52,4 +54,18 @@ class HiveExternalCatalogSuite extends ExternalCatalogSuite {
...
+    val catalog = newBasicCatalog()
+    val hiveTable = CatalogTable(
+      identifier = TableIdentifier("hive_tbl", Some("db1")),
+      tableType = CatalogTableType.MANAGED,
+      storage = storageFormat,
+      schema = new StructType().add("col1", "int").add("col2", "string"),
+      provider = Some("hive"))
+    catalog.createTable(hiveTable, ignoreIfExists = false)
+
+    val rawTable = externalCatalog.client.getTable("db1", "hive_tbl")
```
gatorsmile: Could we also add one more check for another API `externalCatalog.getTable("db1", "hive_tbl")`? The provider should contain `DDLUtils.HIVE_PROVIDER`

gatorsmile:
The [`alterTable`](https://github.com/cloud-fan/spark/blob/5ee6489cdd1c22a1071c4fb6c7e4c4af126c9d50/sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveExternalCatalog.scala#L533) API in HiveExternalCatalog is still based on the provider field. We need a change too.

cloud-fan:
@gatorsmile , `alterTable` don't need to change, the `provider` is not in table properties but still in the `CatalogTable.provider` field.

gatorsmile:
@cloud-fan True. That is not the `oldTableDef` fetched from the metastore. 

LGTM pending test

cloud-fan:
thanks for the review, merging to master/2.1!

PR #16083, opened 2016-11-30
[SPARK-18097][SQL] Add exception catch to handle corrupted metadata scenario while dropping the table

jayadevanmurali:
Add exception catch to handle corrupted meta data scenario while dropping the table

## How was this patch tested?
Unit tests


thomastechs:
This patch would be catching the exception happening while the metadatacheck(due to getting corrupted) and continue with a warning log and continue with the drop operation...

sql/core/src/main/scala/org/apache/spark/sql/execution/command/ddl.scala
```diff
@@ -189,14 +189,18 @@ case class DropTableCommand(
...
+          case CatalogTableType.VIEW if !isView =>
+            throw new AnalysisException(
+              "Cannot drop a view with DROP TABLE. Please use DROP VIEW instead")
+          case o if o != CatalogTableType.VIEW && isView =>
+            throw new AnalysisException(
+              s"Cannot drop a table with DROP VIEW. Please use DROP TABLE instead")
+          case _ =>
+        }
+      } catch {
+          case e: QueryExecutionException => log.warn(e.toString, e)
```
gatorsmile: How did you test this?

jayadevanmurali: Hi @gatorsmile, I tested this change against all existing test cases. Also @davies mentioned in the jira itself, we cannot purposefully corrupt the meta data. So I tested and verified, this change will not break any existing test cases. And we are expecting QueryExecutionException when the meta data is get corrupted in some way.

gatorsmile: The existing test cases do not cover this case. I am afraid this might not work as expected. We need to test it.

thomastechs: @gatorsmile  and @Davies  ; In that case, any suggestions to mock the metadata corrupt scenario?

gatorsmile: I do not know why we can hit `QueryExecutionException` in `getTableMetadata`.

gatorsmile: Maybe you can manually modify the metadata stored in the database and check whether it fails as expected.

HyukjinKwon:
Hi @jayadevanmurali, are you still working on this? It seems inactive for few months. Maybe, it might better be closed for now if you are currently not able to proceed further.

jayadevanmurali:
Hi @HyukjinKwon ,
I tried to corrupt the metadata by editing files under spark/warehouse directory. But it is not working. Do you have any idea to corrupt metadata manually.
Thanks

HyukjinKwon:
Actually, I don't know (or heard of) how the schema is corrupt. I guess any committer should verify this before merging it. How about asking this to dev/user mailing list if you are not sure? If it is expected to be open without the reproducible steps for a long time (like few weeks or months), I guess we maybe should better close this for now and then open it latter.

PR #16084, opened 2016-11-30
[SPARK-18654][SQL] Remove unreachable patterns in makeRootConverter

NathanHowell:

`makeRootConverter` is only called with a `StructType` value. By making this method less general we can remove pattern matches, which are never actually hit outside of the test suite.

## How was this patch tested?

The existing tests.

### DIFF
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/json/JacksonParser.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/json/JacksonParser.scala
@@ -119,7 +119,7 @@ class JacksonParser(
    * to a value according to a desired schema. This is a wrapper for the method
    * `makeConverter()` to handle a row wrapped with an array.
    */
-  def makeRootConverter(st: StructType): ValueConverter = {
+  private def makeRootConverter(st: StructType): ValueConverter = {
     val elementConverter = makeConverter(st)
     val fieldConverters = st.map(_.dataType).map(makeConverter)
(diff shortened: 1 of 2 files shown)


NathanHowell:
cc/ @HyukjinKwon

JoshRosen:
Jenkins, this is ok to test

HyukjinKwon:
Yes, I noticed it before when I made a last change (avoiding the type dispatch for each record) but I just made a safe choice.

It looks good to me as targeted because `JacksonParser` constructor actually takes `StructType` and `makeRootConverter` was exposed for testing purpose in `JsonSuite`.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/json/JacksonParser.scala
```diff
@@ -119,47 +119,33 @@ class JacksonParser(
...
-      (parser: JsonParser) => parseJsonToken(parser, dataType) {
-        // the business end of SPARK-3308:
-        // when an object is found but an array is requested just wrap it in a list.
-        // This is being wrapped in `JacksonParser.parse`.
-        case START_OBJECT => convertObject(parser, st, fieldConverters)
-        case START_ARRAY => convertArray(parser, elementConverter)
-      }
-
-    case _ => makeConverter(dataType)
+  def makeRootConverter(st: StructType): ValueConverter = {
```
NathanHowell: Yep, this can be private. Fixed in the most recent push.

rxin:
Merging in master.

PR #16086, opened 2016-11-30
[SPARK-18653][SQL] Fix incorrect space padding for unicode character at Dataset.show

kiszk:

This PR put correct space padding for unicode character at `Dataset.show()`.
The reason of putting incorrect padding is to count string width by string.length. This PR counds string width by using [East Asian Width](http://www.unicode.org/Public/UCD/latest/ucd/EastAsianWidth.txt).

Example program
````
case class UnicodeCaseClass(整数: Int, 実数: Double, s: String)
Seq(UnicodeCaseClass(1, 1.1, "文字列1")).toDS.show
````

Output without this PR
````
+---+---+----+
| 整数| 実数|   s|
+---+---+----+
|  1|1.1|文字列1|
+---+---+----+
````

Output with this PR
````
+----+----+-------+
|整数|実数|      s|
+----+----+-------+
|   1| 1.1|文字列1|
+----+----+-------+
````

## How was this patch tested?

Add a test suite

### DIFF
--- a/sql/core/pom.xml
+++ b/sql/core/pom.xml
@@ -92,6 +92,11 @@
       <version>${fasterxml.jackson.version}</version>
     </dependency>
     <dependency>
+      <groupId>com.ibm.icu</groupId>
+      <artifactId>icu4j</artifactId>
+      <version>${ibm.icu.version}</version>
+    </dependency>
+    <dependency>
       <groupId>org.scalacheck</groupId>
       <artifactId>scalacheck_${scala.binary.version}</artifactId>
       <scope>test</scope>
--- a/sql/core/src/main/scala/org/apache/spark/sql/Dataset.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/Dataset.scala
@@ -238,7 +238,7 @@ class Dataset[T] private[sql](
       }
   }
 
-  var EAST_ASIAN_LANGS = Seq("ja", "vi", "kr", "zh")
+  val EAST_ASIAN_LANGS = Seq("ja", "vi", "kr", "zh")
 
   private def unicodeWidth(str: String): Int = {
     val locale = Locale.getDefault()
@@ -249,24 +249,16 @@ class Dataset[T] private[sql](
     var len = 0
     for (i <- 0 until str.length) {
       val codePoint = str.codePointAt(i)
-      val value = UCharacter.getIntPropertyValue(codePoint, UProperty.EAST_ASIAN_WIDTH);
+      val value = UCharacter.getIntPropertyValue(codePoint, UProperty.EAST_ASIAN_WIDTH)
       len = len + (value match {
         case UCharacter.EastAsianWidth.NARROW | UCharacter.EastAsianWidth.NEUTRAL |
-              UCharacter.EastAsianWidth.HALFWIDTH => 1
+             UCharacter.EastAsianWidth.HALFWIDTH => 1
         case UCharacter.EastAsianWidth.FULLWIDTH | UCharacter.EastAsianWidth.WIDE => 2
         case UCharacter.EastAsianWidth.AMBIGUOUS => ambiguousLen
         case _ => 1
       })
     }
     len
-  }
-
-  private def repeatPadding(len: Int): StringBuilder = {
-    var str = new StringBuilder(len)
-    for (i <- 0 until len) {
-      str.append(' ')
-    }
-    str
   }
 
   /**
@@ -327,10 +319,11 @@ class Dataset[T] private[sql](
 
     // column names
     rows.head.zipWithIndex.map { case (cell, i) =>
+      val paddingLen = colMaxWidths(i) - colWidths(0)(i)
       if (truncate > 0) {
-        repeatPadding(colMaxWidths(i) - colWidths(0)(i)).append(cell)
+        new StringBuilder(cell.length, " " * paddingLen).append(cell)
       } else {
-        new StringBuffer(cell).append(repeatPadding(colMaxWidths(i) - colWidths(0)(i)))
+        new StringBuilder(paddingLen, cell).append(" " * paddingLen)
       }
     }.addString(sb, "|", "|", "|\n")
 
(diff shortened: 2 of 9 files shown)

### CODE REVIEW

**Comment by kiszk** [comment]
@gatorsmile would it be possible to review this? You would be familiar with Kanji?

File: sql/core/pom.xml [hunk 1]
```diff
@@ -92,6 +92,11 @@
       <version>${fasterxml.jackson.version}</version>
     </dependency>
     <dependency>
+      <groupId>com.ibm.icu</groupId>
```
srowen: I'm not against it, but I'm a little hesitant to bring in all this weight to fix a basically cosmetic problem. This may already be included transitively though. WOrth checking the a) license of this library and b) whether it's already in use in the transitive dependencies?

srowen: OK, yes that's a cat-A license so it's OK. With any dependency I'd also want to check whether it brings in anything else under a different license or whether it's particularly large, etc.

Disregard my other comment. I think I was thinking of the fact that Lucene already uses this.

sql/core/src/main/scala/org/apache/spark/sql/Dataset.scala
```diff
@@ -236,6 +238,37 @@ class Dataset[T] private[sql](
       }
   }
 
+  var EAST_ASIAN_LANGS = Seq("ja", "vi", "kr", "zh")
```
viirya: why var?

sql/core/src/main/scala/org/apache/spark/sql/Dataset.scala
```diff
@@ -236,6 +238,37 @@ class Dataset[T] private[sql](
...
+  private def unicodeWidth(str: String): Int = {
+    val locale = Locale.getDefault()
+    if (locale == null) {
+      throw new NullPointerException("locale is null")
+    }
+    val ambiguousLen = if (EAST_ASIAN_LANGS.contains(locale.getLanguage())) 2 else 1
+    var len = 0
+    for (i <- 0 until str.length) {
+      val codePoint = str.codePointAt(i)
+      val value = UCharacter.getIntPropertyValue(codePoint, UProperty.EAST_ASIAN_WIDTH);
```
viirya: There is a ";" at the end of this line.

sql/core/src/main/scala/org/apache/spark/sql/Dataset.scala
```diff
@@ -236,6 +238,37 @@ class Dataset[T] private[sql](
...
+      len = len + (value match {
+        case UCharacter.EastAsianWidth.NARROW | UCharacter.EastAsianWidth.NEUTRAL |
+              UCharacter.EastAsianWidth.HALFWIDTH => 1
```
viirya: ident looks weird.

sql/core/src/main/scala/org/apache/spark/sql/Dataset.scala
```diff
@@ -275,36 +308,43 @@ class Dataset[T] private[sql](
...
 
     // Create SeparateLine
-    val sep: String = colWidths.map("-" * _).addString(sb, "+", "+", "+\n").toString()
+    val sep: String = colMaxWidths.map("-" * _).addString(sb, "+", "+", "+\n").toString()
 
     // column names
     rows.head.zipWithIndex.map { case (cell, i) =>
       if (truncate > 0) {
-        StringUtils.leftPad(cell, colWidths(i))
+        repeatPadding(colMaxWidths(i) - colWidths(0)(i)).append(cell)
```
kiszk: `StringUtils.leftPad/rightPad` uses String.length. Since this usage causes the same problem, the new code does not use these methods.

viirya: oh. Got it.

For this purpose, current `repeatPadding` looks verbose. If you just want to create exact number of spaces, you can use `" " * n`.

kiszk: Good catch. However, I would like to get the number of spaces with `StringBuilder`. It seems to return `String`.

sql/core/src/main/scala/org/apache/spark/sql/Dataset.scala
kiszk: good catch, done

sql/core/pom.xml
gatorsmile: ICU is widely used, as shown in http://site.icu-project.org . A very useful package. Before, we used it for codepage conversion.

sql/core/src/main/scala/org/apache/spark/sql/Dataset.scala
```diff
@@ -236,6 +238,29 @@ class Dataset[T] private[sql](
...
   }
 
+  val EAST_ASIAN_LANGS = Seq("ja", "vi", "kr", "zh")
+
+  private def unicodeWidth(str: String): Int = {
+    val locale = Locale.getDefault()
+    if (locale == null) {
+      throw new NullPointerException("locale is null")
+    }
+    val ambiguousLen = if (EAST_ASIAN_LANGS.contains(locale.getLanguage())) 2 else 1
```
gatorsmile: How about creating a separate helper function for the default width?

sql/core/src/main/scala/org/apache/spark/sql/Dataset.scala
```diff
@@ -236,6 +238,29 @@ class Dataset[T] private[sql](
       }
   }
 
+  val EAST_ASIAN_LANGS = Seq("ja", "vi", "kr", "zh")
```
gatorsmile: Only these four?

gatorsmile:
Now, `showString` become more complex for handling the Unicode width. I am neutral on this fix.

viirya:
yeah, I would think this is a relatively rare use case. Need to consider if it is worth extra complexity.

rxin:
I agree - don't think this is worth the complexity.

srowen:
Unless someone vigorously objects, yes let's close this.

kiszk:
I am thinking about an simpler approach. However, it is fine to close for now.

PR #16087, opened 2016-11-30
[SPARK-18656][SQL] Memory efficient implementation of the org.apache.spark.sql.execution.stat.StatFunctions#multipleApproxQuantiles

sinasohangirsc:
@thunterdb 

Creating quantile summaries per partition and then merge together, to be more memory efficient. 

## How was this patch tested?

Passes: ./dev/run-tests
Also ran compared results between the old and new implementation.

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/stat/StatFunctions.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/stat/StatFunctions.scala
@@ -67,8 +67,6 @@ object StatFunctions extends Logging {
         " is not supported.")
       Column(Cast(Column(colName).expr, DoubleType))
     }
-    val emptySummaries = Array.fill(cols.size)(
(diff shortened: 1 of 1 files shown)


rxin:
We now have an aggregate function for this, so we can deprecate this usage now, can't we?

cc @thunterdb

sinasohangirsc:
Totally right. I jumped into conclusion too fast. Already put the JIRA issue for later. Forgot to close this.

barrybecker4:
@rxin, what is the new aggregate function that we should be using instead of multipleApproxQuantiles?

poplav:
@barrybecker4 Following the trail of this issue...What is the solution for many columns rather than the workaround for chunking?  Thanks!

barrybecker4:
I used chunking the columns to avoid running out of memory when using multipleApproxQuantiles. I don't know if there is a better way.

PR #16088, opened 2016-11-30
[SPARK-18659] [SQL] Incorrect behaviors in overwrite table for datasource tables

ericl:

Two bugs are addressed here
1. INSERT OVERWRITE TABLE sometime crashed when catalog partition management was enabled. This was because when dropping partitions after an overwrite operation, the Hive client will attempt to delete the partition files. If the entire partition directory was dropped, this would fail. The PR fixes this by adding a flag to control whether the Hive client should attempt to delete files.
2. The static partition spec for OVERWRITE TABLE was not correctly resolved to the case-sensitive original partition names. This resulted in the entire table being overwritten if you did not correctly capitalize your partition names.

cc @yhuai @cloud-fan 

## How was this patch tested?

Unit tests. Surprisingly, the existing overwrite table tests did not catch these edge cases.


sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/InMemoryCatalog.scala
```diff
@@ -399,7 +400,7 @@ class InMemoryCatalog(
     // TODO: we should follow hive to roll back if one partition path failed to delete, and support
     // partial partition spec.
     partSpecs.foreach { p =>
-      if (existingParts.contains(p) && shouldRemovePartitionLocation) {
+      if (existingParts.contains(p) && shouldRemovePartitionLocation && !retainData) {
```
cloud-fan: shall we put the `!retainData` in the definition of `shouldRemovePartitionLocation`?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/SessionCatalog.scala
```diff
@@ -687,13 +687,14 @@ class SessionCatalog(
       tableName: TableIdentifier,
       specs: Seq[TablePartitionSpec],
       ignoreIfNotExists: Boolean,
-      purge: Boolean): Unit = {
+      purge: Boolean,
+      retainData: Boolean): Unit = {
```
cloud-fan: shall we provide a default value? looks like most of the time we want it to be false

sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveExternalCatalog.scala
```diff
@@ -847,9 +847,11 @@ private[spark] class HiveExternalCatalog(conf: SparkConf, hadoopConf: Configurat
       table: String,
       parts: Seq[TablePartitionSpec],
       ignoreIfNotExists: Boolean,
-      purge: Boolean): Unit = withClient {
+      purge: Boolean,
+      deleteFiles: Boolean): Unit = withClient {
```
cloud-fan: why change the parameter name here? it's called `retainData` right?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/InMemoryCatalog.scala
ericl: Done

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/SessionCatalog.scala
ericl: It seemed a little safer to make it mandatory to avoid some component forgetting to propagate it.

sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveExternalCatalog.scala
ericl: Fixed

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSourceStrategy.scala
```diff
@@ -217,16 +217,25 @@ case class DataSourceAnalysis(conf: CatalystConf) extends Rule[LogicalPlan] {
             if (deletedPartitions.nonEmpty) {
               AlterTableDropPartitionCommand(
                 l.catalogTable.get.identifier, deletedPartitions.toSeq,
-                ifExists = true, purge = true).run(t.sparkSession)
+                ifExists = true, purge = true,
```
cloud-fan: an unrelated question, when dropping partitions here, do we have to set `purge = true`?

ericl: I think it doesn't matter since we don't delete files.

cloud-fan:
LGTM pending jenkins

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSourceStrategy.scala
cloud-fan: then can we set it to false? `DROP PARTITION ... PURGE` is not supported in hive 0.13, setting it to false can make the partition related functions still work in older hive versions.

ericl: Done

cloud-fan:
retest this please

cloud-fan:

PR #16089, opened 2016-11-30
[SPARK-18658][SQL] Write text records directly to a FileOutputStream

NathanHowell:

This replaces uses of `TextOutputFormat` with an `OutputStream`, which will either write directly to the filesystem or indirectly via a compressor (if so configured). This avoids intermediate buffering.

The inverse of this (reading directly from a stream) is necessary for streaming large JSON records (when `wholeFile` is enabled) so I wanted to keep the read and write paths symmetric.

## How was this patch tested?

Existing unit tests.

### DIFF
--- a/common/unsafe/src/main/java/org/apache/spark/unsafe/types/UTF8String.java
+++ b/common/unsafe/src/main/java/org/apache/spark/unsafe/types/UTF8String.java
@@ -147,6 +147,17 @@ public final class UTF8String implements Comparable<UTF8String>, Externalizable,
     buffer.position(pos + numBytes);
   }
 
+  public void writeTo(OutputStream out) throws IOException {
+    if (base instanceof byte[] &&
+        offset >= BYTE_ARRAY_OFFSET &&
+        offset < (long) Integer.MAX_VALUE) {
+      // the offset includes the object header... but it is only needed for unsafe copies
+      out.write((byte[]) base, (int) offset - BYTE_ARRAY_OFFSET, numBytes);
+    } else {
+      out.write(getBytes());
+    }
+  }
+
   /**
    * Returns the number of bytes for a code point with the first byte as `b`
    * @param b The first byte of a code point
--- a/common/unsafe/src/main/java/org/apache/spark/unsafe/types/UTF8String.java
+++ b/common/unsafe/src/main/java/org/apache/spark/unsafe/types/UTF8String.java
@@ -148,11 +148,19 @@ public final class UTF8String implements Comparable<UTF8String>, Externalizable,
   }
 
   public void writeTo(OutputStream out) throws IOException {
-    if (base instanceof byte[] &&
-        offset >= BYTE_ARRAY_OFFSET &&
-        offset < (long) Integer.MAX_VALUE) {
-      // the offset includes the object header... but it is only needed for unsafe copies
-      out.write((byte[]) base, (int) offset - BYTE_ARRAY_OFFSET, numBytes);
+    if (base instanceof byte[] && offset >= BYTE_ARRAY_OFFSET) {
+      final byte[] bytes = (byte[]) base;
+
+      // the offset includes an object header... this is only needed for unsafe copies
+      final long arrayOffset = offset - BYTE_ARRAY_OFFSET;
+
+      // verify that the offset and length points somewhere inside the byte array
+      // and that the offset can safely be truncated to a 32-bit integer
+      if ((long) bytes.length < arrayOffset + numBytes) {
+        throw new ArrayIndexOutOfBoundsException();
+      }
+
+      out.write(bytes, (int) arrayOffset, numBytes);
     } else {
       out.write(getBytes());
     }
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/json/JacksonGenerator.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/json/JacksonGenerator.scala
@@ -194,4 +194,8 @@ private[sql] class JacksonGenerator(
       writeFields(row, schema, rootFieldWriters)
     }
   }
+
+  private[sql] def writeLineEnding(): Unit = {
+    gen.writeRaw('\n')
+  }
 }
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/json/JacksonGenerator.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/json/JacksonGenerator.scala
@@ -195,7 +195,7 @@ private[sql] class JacksonGenerator(
     }
   }
 
-  private[sql] def writeLineEnding(): Unit = {
+  def writeLineEnding(): Unit = {
     gen.writeRaw('\n')
   }
 }
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/CodecStreams.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/CodecStreams.scala
@@ -1,0 +1,73 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.spark.sql.execution.datasources
+
+import java.io.{OutputStream, OutputStreamWriter}
+import java.nio.charset.{Charset, StandardCharsets}
+
+import org.apache.hadoop.fs.Path
+import org.apache.hadoop.io.compress._
+import org.apache.hadoop.mapreduce.JobContext
+import org.apache.hadoop.mapreduce.lib.output.FileOutputFormat
+import org.apache.hadoop.util.ReflectionUtils
+
+object CodecStreams {
+  private def getCompressionCodec(
+      context: JobContext,
+      file: Option[Path] = None): Option[CompressionCodec] = {
+    if (FileOutputFormat.getCompressOutput(context)) {
+      val compressorClass = FileOutputFormat.getOutputCompressorClass(
+        context,
+        classOf[GzipCodec])
+
+      Some(ReflectionUtils.newInstance(compressorClass, context.getConfiguration))
+    } else {
+      file.flatMap { path =>
+        val compressionCodecs = new CompressionCodecFactory(context.getConfiguration)
+        Option(compressionCodecs.getCodec(path))
+      }
+    }
+  }
+
+  /** Create a new file and open it for writing.
+   * If compression is enabled in the [[JobContext]] the stream will write compressed data to disk.
+   * An exception will be thrown if the file already exists.
+   */
+  def getOutputStream(context: JobContext, file: Path): OutputStream = {
+    val fs = file.getFileSystem(context.getConfiguration)
+    val outputStream: OutputStream = fs.create(file, false)
+
+    getCompressionCodec(context, Some(file)).fold(outputStream) { codec =>
+      codec.createOutputStream(outputStream)
+    }
+  }
+
+  def getOutputStreamWriter(
+      context: JobContext,
+      file: Path,
+      charset: Charset = StandardCharsets.UTF_8): OutputStreamWriter = {
+    new OutputStreamWriter(getOutputStream(context, file), charset)
+  }
+
+  /** Returns the compression codec extension to be used in a file name, e.g. ".gzip"). */
+  def getCompressionExtension(context: JobContext): String = {
+    getCompressionCodec(context).fold("") { code =>
+      code.getDefaultExtension
+    }
+  }
+}
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/CodecStreams.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/CodecStreams.scala
@@ -53,9 +53,9 @@ object CodecStreams {
     val fs = file.getFileSystem(context.getConfiguration)
     val outputStream: OutputStream = fs.create(file, false)
 
-    getCompressionCodec(context, Some(file)).fold(outputStream) { codec =>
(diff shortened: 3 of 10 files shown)


NathanHowell:
This touches a fair number of components. I also haven't done any performance testing to see what the impact of this is. Curious what your thoughts are?

cc/ @marmbrus @rxin @JoshRosen

rxin:
Does this work against file systems with HDFS API (not local posix)? If yes, sounds good!

NathanHowell:
Yep. It uses the Hadoop `FileSystem` class to open files, just like `TextOutputFormat` does.

rxin:
Yea then this is definitely fine.

JoshRosen:
Jenkins, this is ok to test

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/json/JacksonGenerator.scala
```diff
@@ -194,4 +194,8 @@ private[sql] class JacksonGenerator(
       writeFields(row, schema, rootFieldWriters)
     }
   }
+
+  private[sql] def writeLineEnding(): Unit = {
+    gen.writeRaw('\n')
```
HyukjinKwon: Hm, is this safe to guess the newline is always `\n`?

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/text/TextFileFormat.scala
```diff
@@ -132,39 +128,17 @@ class TextOutputWriter(
...
+  private val writer = CodecStreams.getOutputStream(context, new Path(path))
 
   override def write(row: Row): Unit = throw new UnsupportedOperationException("call writeInternal")
 
   override protected[sql] def writeInternal(row: InternalRow): Unit = {
     val utf8string = row.getUTF8String(0)
-    buffer.set(utf8string.getBytes)
-    recordWriter.write(NullWritable.get(), buffer)
+    writer.write(utf8string.getBytes)
+    writer.write('\n')
```
HyukjinKwon: Here too and for CSV as well.

NathanHowell: This mirrors what Hadoop code does, see https://github.com/apache/hadoop/blob/f67237cbe7bc48a1b9088e990800b37529f1db2a/hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-core/src/main/java/org/apache/hadoop/mapreduce/lib/output/TextOutputFormat.java#L48-L49

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/CodecStreams.scala
```diff
@@ -0,0 +1,73 @@
...
+import java.io.{OutputStream, OutputStreamWriter}
+import java.nio.charset.{Charset, StandardCharsets}
+
+import org.apache.hadoop.fs.Path
+import org.apache.hadoop.io.compress._
+import org.apache.hadoop.mapreduce.JobContext
+import org.apache.hadoop.mapreduce.lib.output.FileOutputFormat
+import org.apache.hadoop.util.ReflectionUtils
+
+private[spark] object CodecStreams {
```
HyukjinKwon: I guess we don't need `private[spark]` here assuming from the commit here (https://github.com/apache/spark/commit/511f52f8423e151b0d0133baf040d34a0af3d422)

NathanHowell:
Doh, forgot to run the Hive tests. Should be fixed now.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/CodecStreams.scala
NathanHowell: Looks that way, I've removed it.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/CodecStreams.scala
```diff
@@ -0,0 +1,73 @@
...
+      Some(ReflectionUtils.newInstance(compressorClass, context.getConfiguration))
+    } else {
+      file.flatMap { path =>
+        val compressionCodecs = new CompressionCodecFactory(context.getConfiguration)
+        Option(compressionCodecs.getCodec(path))
+      }
+    }
+  }
+
+  /** Create a new file and open it for writing.
```
rxin: can you fix the comment style here? we don't use scaladoc style comment in Spark.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/json/JacksonGenerator.scala
```diff
@@ -194,4 +194,8 @@ private[sql] class JacksonGenerator(
       writeFields(row, schema, rootFieldWriters)
     }
   }
+
+  private[sql] def writeLineEnding(): Unit = {
```
rxin: remove private[sql] here to be consistent with other methods

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/CodecStreams.scala
```diff
@@ -0,0 +1,73 @@
...
+   * If compression is enabled in the [[JobContext]] the stream will write compressed data to disk.
+   * An exception will be thrown if the file already exists.
+   */
+  def getOutputStream(context: JobContext, file: Path): OutputStream = {
+    val fs = file.getFileSystem(context.getConfiguration)
+    val outputStream: OutputStream = fs.create(file, false)
+
+    getCompressionCodec(context, Some(file)).fold(outputStream) { codec =>
```
rxin: I know you like Haskell but every time I see a fold I have to spend an extra 5 secs checking what it is doing :) Can we simplify this one?

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/CodecStreams.scala
```diff
@@ -0,0 +1,73 @@
...
+        Option(compressionCodecs.getCodec(path))
+      }
+    }
+  }
+
+  /** Create a new file and open it for writing.
+   * If compression is enabled in the [[JobContext]] the stream will write compressed data to disk.
+   * An exception will be thrown if the file already exists.
+   */
+  def getOutputStream(context: JobContext, file: Path): OutputStream = {
```
rxin: getOutputStream -> createOutputStream

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/CodecStreams.scala
```diff
@@ -0,0 +1,73 @@
...
+  def getOutputStreamWriter(
+      context: JobContext,
+      file: Path,
+      charset: Charset = StandardCharsets.UTF_8): OutputStreamWriter = {
+    new OutputStreamWriter(getOutputStream(context, file), charset)
+  }
+
+  /** Returns the compression codec extension to be used in a file name, e.g. ".gzip"). */
+  def getCompressionExtension(context: JobContext): String = {
+    getCompressionCodec(context).fold("") { code =>
```
rxin: maybe

```
getCompressionCodec(context).map(_. getDefaultExtension).getOrElse("")
```

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/text/TextFileFormat.scala
```diff
@@ -132,39 +128,17 @@ class TextOutputWriter(
...
-  }
+  private val writer = CodecStreams.getOutputStream(context, new Path(path))
 
   override def write(row: Row): Unit = throw new UnsupportedOperationException("call writeInternal")
 
   override protected[sql] def writeInternal(row: InternalRow): Unit = {
     val utf8string = row.getUTF8String(0)
-    buffer.set(utf8string.getBytes)
-    recordWriter.write(NullWritable.get(), buffer)
+    writer.write(utf8string.getBytes)
```
rxin: can you check UTF8String's implementation to make sure we are not creating a new array for each row?

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/csv/CSVRelation.scala
```diff
@@ -245,24 +230,12 @@ private[csv] class CsvOutputWriter(
...
-    }
-  }
-
-  private def flush(): Unit = {
-    val lines = csvWriter.flush()
-    if (lines.nonEmpty) {
-      text.set(lines)
-      recordWriter.write(NullWritable.get(), text)
-    }
+    csvWriter.writeRow(rowToString(row), printHeader)
```
rxin: we can probably optimize this as well - but not a huge deal in the first pr.

srowen:
I was going to say, hm, are we sure we want to reimplement / go around the Hadoop support for this? but in practice it looks like it actually simplifies some things. At the moment I can't think of any particular behaviors we're missing by avoiding the Input/OutputFormat.

But CC @vanzin @steveloughran for any comment.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/CodecStreams.scala
NathanHowell: Yah, it's a terrible name (and it's not a fold). I'll replace them with `.map(...).getOrElse`.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/text/TextFileFormat.scala
NathanHowell: It is creating a new array, I'll pass the internal one through instead.

NathanHowell: Done, but I'm not 100% sure about the implementation. Can you have someone more familiar with `UTF8String`'s internals double check it?

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/CodecStreams.scala
```diff
@@ -0,0 +1,74 @@
...
+        val compressionCodecs = new CompressionCodecFactory(context.getConfiguration)
+        Option(compressionCodecs.getCodec(path))
+      }
+    }
+  }
+
+  /**
+   * Create a new file and open it for writing.
+   * If compression is enabled in the [[JobContext]] the stream will write compressed data to disk.
+   * An exception will be thrown if the file already exists.
```
steveloughran: will "probably" be thrown; object stores have issues there

common/unsafe/src/main/java/org/apache/spark/unsafe/types/UTF8String.java
```diff
@@ -147,6 +147,17 @@ public void writeTo(ByteBuffer buffer) {
     buffer.position(pos + numBytes);
   }
 
+  public void writeTo(OutputStream out) throws IOException {
```
NathanHowell: Agreed.

steveloughran:
AFAIK, the big thing the FileOutputFormat really adds is not the compression, but the output committer and the stuff to go with that (working directories, paths, etc etc). If you aren't going near that, and just want a fast write of .csv and jackson with optional compression, well, I don't see anything in the code I'd run away from.

If you do want to think about how to write CSV files during the output of speculative work in the presence of failures, well, that's where the mapred.lib.output code really comes out to play.

Otherwise, in general PR review mode: Tests? What if the code asks for a committer that isn't there, passes in null sequences in rows to write, tries to hit the buffer corner cases. Hopefully those exist already, but if not, now is a good time to try to break things.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/json/JacksonGenerator.scala
```diff
@@ -194,4 +194,8 @@ private[sql] class JacksonGenerator(
       writeFields(row, schema, rootFieldWriters)
     }
   }
+
+  def writeLineEnding(): Unit = {
+    gen.writeRaw('\n')
```
steveloughran: Presumably just a convoluted way to create a byte array containing the byte `0x0a` then

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/CodecStreams.scala
NathanHowell: Is this a problem with Hadoop in general? The FileSystem docs also specify this behavior:

```
  /**
   * Create an FSDataOutputStream at the indicated Path.
   * @param f the file to create
   * @param overwrite if a file with this name already exists, then if true,
   *   the file will be overwritten, and if false an exception will be thrown.
   */
```

NathanHowell:
@steveloughran Spark is handling the output committing somewhere further up the stack. The path being passed in to `OutputWriterFactory.newInstance` is to a temporary file, such as `/private/var/folders/sq/vmncyd7506q_ch43llrwr8sn6zfknl/T/spark-3db2844b-1f3c-45c2-8bf4-8a3c81440e38/_temporary/0/_temporary/attempt_20161201081833_0000_m_000000_0/part-00000-8dd44cea-c01e-4bfe-ab03-641ebce18afb.txt`.

I'll make a pass through the existing tests to see if anything obvious is missing.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/json/JacksonGenerator.scala
NathanHowell: That is my assumption. I'm also assuming that writing a single byte is slightly more efficient than writing an array of a single byte.

steveloughran:
I ask about committers as I'm staring at the V1 and V2 committer APIs right now related to S3 destinations; not directly related to this though.

rxin:
@srowen yea the hadoop format api is pretty awkward to use, and actually makes everything more complicated than needed.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/csv/CSVRelation.scala
NathanHowell: The uniVocity CSV writer converts every column to a `String` before writing so it's (probably?) not possible to further optimize this without doing a whole bunch of work. I only did a quick scan through their code though.

common/unsafe/src/main/java/org/apache/spark/unsafe/types/UTF8String.java
NathanHowell: I've added a few tests for this method.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/csv/CSVRelation.scala
rxin: Yup I asked this question but never had time to follow-up: https://github.com/uniVocity/univocity-parsers/issues/99

rxin:
Merging in master. Thanks.

PR #16090, opened 2016-11-30
[SPARK-18661] [SQL] Creating a partitioned datasource table should not scan all files for table

ericl:

Even though in 2.1 creating a partitioned datasource table will not populate the partition data by default (until the user issues MSCK REPAIR TABLE), it seems we still scan the filesystem for no good reason.

We should avoid doing this when the user specifies a schema.

## How was this patch tested?

Perf stat tests.


sql/core/src/main/scala/org/apache/spark/sql/execution/command/createDataSourceTables.scala
```diff
@@ -58,13 +58,20 @@ case class CreateDataSourceTableCommand(table: CatalogTable, ignoreIfExists: Boo
     // Create the relation to validate the arguments before writing the metadata to the metastore,
     // and infer the table schema and partition if users didn't specify schema in CREATE TABLE.
     val pathOption = table.storage.locationUri.map("path" -> _)
+    val uncreatedTable = table.copy(
```
ericl: @cloud-fan is there a better way to do this?

ericl:
@rxin

sql/core/src/main/scala/org/apache/spark/sql/execution/command/createDataSourceTables.scala
cloud-fan: can we do this before we pass the `CatalogTable`? e.g. in the parser and `DataFrameWriter`

ericl: Hm, that seems more brittle since you'd have to duplicate the logic. I added a comment describing why we need to do this here.

sql/core/src/test/scala/org/apache/spark/sql/execution/command/DDLSuite.scala
```diff
@@ -565,7 +571,8 @@ class DDLSuite extends QueryTest with SharedSQLContext with BeforeAndAfterEach {
       val table = catalog.getTableMetadata(TableIdentifier("tbl"))
       assert(table.tableType == CatalogTableType.MANAGED)
       assert(table.provider == Some("parquet"))
-      assert(table.schema == new StructType().add("a", IntegerType).add("b", IntegerType))
+      // a is ordered last since it is a user-specified partitioning column
+      assert(table.schema == new StructType().add("b", IntegerType).add("a", IntegerType))
```
ericl: @yhuai this is the minor behavior change we discussed about create table in 2.1

sql/core/src/main/scala/org/apache/spark/sql/execution/command/createDataSourceTables.scala
```diff
@@ -58,13 +58,21 @@ case class CreateDataSourceTableCommand(table: CatalogTable, ignoreIfExists: Boo
     // Create the relation to validate the arguments before writing the metadata to the metastore,
     // and infer the table schema and partition if users didn't specify schema in CREATE TABLE.
     val pathOption = table.storage.locationUri.map("path" -> _)
+    // Fill in some default table options from the session conf
+    val uncreatedTable = table.copy(
```
cloud-fan: how about `tableWithDefaultOptions`?

sql/core/src/main/scala/org/apache/spark/sql/execution/command/createDataSourceTables.scala
```diff
@@ -58,13 +58,21 @@ case class CreateDataSourceTableCommand(table: CatalogTable, ignoreIfExists: Boo
...
+      identifier = table.identifier.copy(
+        database = Some(
+          table.identifier.database.getOrElse(sessionState.catalog.getCurrentDatabase))),
+      tracksPartitionsInCatalog = sparkSession.sessionState.conf.manageFilesourcePartitions)
```
cloud-fan: Logically we don't know the value of `tracksPartitionsInCatalog` here, as the partition columns are not inferred yet.

sql/core/src/main/scala/org/apache/spark/sql/execution/command/createDataSourceTables.scala
```diff
@@ -58,13 +58,21 @@ case class CreateDataSourceTableCommand(table: CatalogTable, ignoreIfExists: Boo
...
     val dataSource: BaseRelation =
       DataSource(
         sparkSession = sparkSession,
         userSpecifiedSchema = if (table.schema.isEmpty) None else Some(table.schema),
+        partitionColumns = table.partitionColumnNames,
         className = table.provider.get,
         bucketSpec = table.bucketSpec,
-        options = table.storage.properties ++ pathOption).resolveRelation()
+        options = table.storage.properties ++ pathOption,
+        catalogTable = Some(uncreatedTable)).resolveRelation()
```
cloud-fan: why we need to pass in the `catalogTable` here?

sql/core/src/test/scala/org/apache/spark/sql/execution/command/DDLSuite.scala
```diff
@@ -312,7 +312,13 @@ class DDLSuite extends QueryTest with SharedSQLContext with BeforeAndAfterEach {
           pathToNonPartitionedTable,
           userSpecifiedSchema = Option("num int, str string"),
           userSpecifiedPartitionCols = partitionCols,
-          expectedSchema = new StructType().add("num", IntegerType).add("str", StringType),
+          expectedSchema = if (partitionCols.isDefined) {
```
cloud-fan: shall we just change the test to use `str` as partition column?

sql/core/src/main/scala/org/apache/spark/sql/execution/command/createDataSourceTables.scala
```diff
@@ -58,13 +58,21 @@ case class CreateDataSourceTableCommand(table: CatalogTable, ignoreIfExists: Boo
...
+    val uncreatedTable = table.copy(
+      identifier = table.identifier.copy(
+        database = Some(
+          table.identifier.database.getOrElse(sessionState.catalog.getCurrentDatabase))),
+      tracksPartitionsInCatalog = sparkSession.sessionState.conf.manageFilesourcePartitions)
     val dataSource: BaseRelation =
       DataSource(
         sparkSession = sparkSession,
         userSpecifiedSchema = if (table.schema.isEmpty) None else Some(table.schema),
+        partitionColumns = table.partitionColumnNames,
```
cloud-fan: it looks to me that this line and https://github.com/apache/spark/pull/16090/files#diff-7a6cb188d2ae31eb3347b5629a679cecR135 are the key change. Did I miss something?

sql/core/src/main/scala/org/apache/spark/sql/execution/command/createDataSourceTables.scala
ericl: Done

sql/core/src/main/scala/org/apache/spark/sql/execution/command/createDataSourceTables.scala
ericl: I think this is true for all new tables. If the table is unpartitioned the flag is harmless.

sql/core/src/main/scala/org/apache/spark/sql/execution/command/createDataSourceTables.scala
ericl: You also need to pass catalogTable in, so that on line 390 of DataSource we create a CatalogFileIndex instead of an InMemoryFileIndex.

sql/core/src/test/scala/org/apache/spark/sql/execution/command/DDLSuite.scala
ericl: I think that would be testing something slightly different.

sql/core/src/main/scala/org/apache/spark/sql/execution/command/createDataSourceTables.scala
ericl: Otherwise, we will construct an InMemoryFileIndex which scans the filesystem.

sql/core/src/main/scala/org/apache/spark/sql/execution/command/createDataSourceTables.scala
cloud-fan: I think it's fine to create an `InMemoryFileIndex` for this case, as we call `DataSource.resolveRelation` just to infer the schema and partition columns.

ericl: You don't want to do that though. Resolve relation also does not always scan the filesystem if you pass in a user defined schema.

ericl: You do need to pass it in though.

```
        val fileCatalog = if (sparkSession.sqlContext.conf.manageFilesourcePartitions &&
            catalogTable.isDefined && catalogTable.get.tracksPartitionsInCatalog) {
          new CatalogFileIndex(
            sparkSession,
            catalogTable.get,
            catalogTable.get.stats.map(_.sizeInBytes.toLong).getOrElse(0L))
        } else {
          new InMemoryFileIndex(sparkSession, globbedPaths, options, Some(partitionSchema))
        }
```

Otherwise, this code will perform a full filesystem scan, independent of the other change to prevent getOrInferFileFormatSchema from performing a scan as well.

cloud-fan:
My main concern is that, in `CreateDataSourceTableCommand`, we call `DataSource.resolveRelation` to infer the schema and partition columns. At that time, the table is not created yet, so logically we should not pass a `CatalogTable` to `DataSource` and create `CatalogFileIndex` inside it, which looks like a hack.

It seems to me that it's more logical to tweak the `InMemoryFileIndex` to scan the files lazily, to avoid unnecessary file scan for cases like this one.

sql/core/src/main/scala/org/apache/spark/sql/execution/command/createDataSourceTables.scala
ericl: That's a pretty big change, considering how many classes depend on the eager behavior of `InMemoryFileIndex`.

ericl:
I looked at avoiding the creation of a CatalogFileIndex, but the way table resolution works right now, the only way is to create some sort of dummy file index class that does not support scans. It's not clear to me that is any better than just creating a CatalogFileIndex, which seems better than incorrectly creating an InMemoryFileIndex.

ericl:
Seems like we also create InMemoryFileIndex twice for non-catalog tables. Let me try to fix that too.

ericl:
Fixed by adding a private cache to `Datasource`, which is used to avoid the duplicate file reads with InMemoryIndex.

ericl:
cc @rxin please merge unless wenchen gets to it first

ericl:
Seems like the caching broke a bunch of tests. I'll take a look at this
again tomorrow.

cloud-fan:
If we are going to hack it, how about this?
```
val dataSource = DataSource(...)
if (classOf[FileFormat].isAssignableFrom(dataSource.providingClass)) {
  dataSource.getOrInferFileFormatSchema()
} else {
  dataSoure.resolveRelation().schema -> new StructType
}
```
Then we don't need to create `FileIndex` and scan files.

ericl:
Not sure I follow - could you explain more on why that would resolve the issue?

Btw, I reverted this pr to b405635, which passes all tests.

cloud-fan:
After looking more at the code, now I agree with your approach. One question, seems we still scan the files when creating a unpartitioned external data source table?

ericl:
Yeah I was wondering if we should also try to fix that. It seems maybe not
as bad since unpartitioned tables usually aren't that big.

We can create separate tickets for investigating that, and also the
duplicate file scan issue (that might not be a regression).

cloud-fan:
LGTM, merging to master/2.1!
@ericl please create tickets for the other 2 issues

ericl:
Filed
https://issues.apache.org/jira/browse/SPARK-18725
https://issues.apache.org/jira/browse/SPARK-18726

PR #16092, opened 2016-12-01
[SPARK-18662] Move resource managers to separate directory

foxish:

* Moves yarn and mesos scheduler backends to resource-managers/ sub-directory (in preparation for https://issues.apache.org/jira/browse/SPARK-18278) 
* Corresponding change in top-level pom.xml.

Ref: https://github.com/apache/spark/pull/16061#issuecomment-263649340

## How was this patch tested?

* Manual tests

/cc @rxin


JoshRosen:
There's a few other places that will need updating, including the `source_file_regexes` in the test runner configuration (`modules.py`): https://github.com/apache/spark/blob/0a811210f809eb5b80eae14694d484d45b48b3f6/dev/sparktestsupport/modules.py#L472 and https://github.com/apache/spark/blob/0a811210f809eb5b80eae14694d484d45b48b3f6/dev/sparktestsupport/modules.py#L488

JoshRosen:
Jenkins, this is ok to test

srowen:
Hm, is this really necessary? I know it's just a move but for 2-3 modules, does it matter enough?

tnachen:
Changed suggested by @rxin

ash211:
With a k8s backend on the way I do think it adds a nice organization for these 3 clearly grouped modules

foxish:
@JoshRosen Thanks for reviewing.
Updated. PTAL.

foxish:
The test failure appears to be an unrelated flake.

vanzin:
@foxish this need to be rebased.

foxish:
@vanzin Done. Thanks!

dev/sparktestsupport/modules.py
```diff
@@ -469,7 +469,7 @@ def __hash__(self):
     name="yarn",
     dependencies=[],
     source_file_regexes=[
-        "yarn/",
+        "resource-managers/yarn/",
```
vanzin: You also need to update L477 (and probably a similar change in the mesos entry).

foxish: @vanzin If I understand correctly, the sbt test goals can stay the same? 
I can see them defined the same way for other modules such as `sketch` (as `sketch/test`), but it is nested under `common/`.

vanzin: Oh. Nevermind, I thought those were file paths like these. So fine the way they are.

foxish: @vanzin also rebased after recent changes to HEAD. Is this okay to merge?

vanzin:
LGTM, merging to master.

foxish:
Thanks @vanzin :)

mgummelt:
@foxish @vanzin I wish you would have cc'd me on this.  If you're going to do something significant to Mesos code, please cc me.

foxish:
Will cc you on future issues. Apologies if this disrupted your workflow. It was meant to be a quick change only in code structure.

PR #16093, opened 2016-12-01
[SPARK-18663][SQL] Simplify CountMinSketch aggregate implementation

rxin:
SPARK-18429 introduced count-min sketch aggregate function for SQL, but the implementation and testing is more complicated than needed. This simplifies the test cases and removes support for data types that don't have clear equality semantics:

1. Removed support for floating point and decimal types.

2. Removed the heavy randomized tests. The underlying CountMinSketch implementation already had pretty good test coverage through randomized tests, and the SPARK-18429 implementation is just to add an aggregate function wrapper around CountMinSketch. There is no need for randomized tests at three different levels of the implementations.

## How was this patch tested?
A lot of the change is to simplify test cases.


rxin:
cc @wzhfy @hvanhovell

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/CountMinSketchAgg.scala
```diff
@@ -75,13 +73,13 @@ case class CountMinSketchAgg(
     } else if (!epsExpression.foldable || !confidenceExpression.foldable ||
       !seedExpression.foldable) {
       TypeCheckFailure(
-        "The eps, confidence or seed provided must be a literal or constant foldable")
+        "The eps, confidence or seed provided must be a literal or foldable")
```
cloud-fan: literal is also foldable, I think we can just way `foldable`

cloud-fan:
LGTM

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/aggregate/CountMinSketchAggSuite.scala
```diff
@@ -17,199 +17,112 @@
...
-  }
-
-  def testItemType[T: ClassTag](dataType: DataType)(itemGenerator: Random => T): Unit = {
-    // Uses fixed seed to ensure reproducible test execution
-    val r = new Random(31)
+  testDataType[Long](LongType, Seq.fill(100) { rand.nextInt(10) })
 
-    val numAllItems = 1000000
-    val allItems = Array.fill(numAllItems)(itemGenerator(r))
+  testDataType[UTF8String](StringType, Seq.fill(100) { UTF8String.fromString(rand.nextString(1)) })
```
wzhfy: also test BinaryType?

wzhfy:
LGTM except one minor point about test case.

rxin:
Merging in master.

PR #16094, opened 2016-12-01
[SPARK-18541][Python]Add metadata parameter to pyspark.sql.Column.alias()

shea-parkes:

Add a `metadata` keyword parameter to `pyspark.sql.Column.alias()` to allow users to mix-in metadata while manipulating `DataFrame`s in `pyspark`.  Without this, I believe it was necessary to pass back through `SparkSession.createDataFrame` each time a user wanted to manipulate `StructField.metadata` in `pyspark`.  

This pull request also improves consistency between the Scala and Python APIs (i.e. I did not add any functionality that was not already in the Scala API).

Discussed ahead of time on JIRA with @marmbrus 

## How was this patch tested?

Added unit tests (and doc tests).  Ran the pertinent tests manually.

### DIFF
--- a/python/pyspark/sql/column.py
+++ b/python/pyspark/sql/column.py
@@ -299,19 +299,22 @@ class Column(object):
     isNotNull = _unary_op("isNotNull", "True if the current expression is not null.")
 
     @since(1.3)
-    def alias(self, *alias, metadata=None):
+    def alias(self, *alias, **kwargs):
         """
         Returns this column aliased with a new name or names (in the case of expressions that
         return more than one column, such as explode).
 
-        Optional ``metadata`` can be passed when aliasing a single column.  It will be stored
-        in the ``metadata`` attribute of the corresponding ``StructField``.
+        Optional ``metadata`` keyword argument can be passed when aliasing a single column.
+        Its contents will be stored in the ``metadata`` attribute of the matching ``StructField``.
 
         >>> df.select(df.age.alias("age2")).collect()
         [Row(age2=2), Row(age2=5)]
         >>> df.select(df.age.alias("age3", metadata={'max': 99})).schema['age3'].metadata['max']
         99
         """
+
+        metadata = kwargs.pop('metadata', None)
+        assert not kwargs, 'Unexpected kwargs where passed: %s' % kwargs
 
         sc = SparkContext._active_spark_context
         if len(alias) == 1:
--- a/python/pyspark/sql/column.py
+++ b/python/pyspark/sql/column.py
@@ -307,6 +307,9 @@ class Column(object):
         Optional ``metadata`` keyword argument can be passed when aliasing a single column.
         Its contents will be stored in the ``metadata`` attribute of the matching ``StructField``.
 
+        .. versionchanged:: 2.2
+           Added optional metadata argument.
+
         >>> df.select(df.age.alias("age2")).collect()
         [Row(age2=2), Row(age2=5)]
         >>> df.select(df.age.alias("age3", metadata={'max': 99})).schema['age3'].metadata['max']
--- a/python/pyspark/sql/column.py
+++ b/python/pyspark/sql/column.py
@@ -17,6 +17,7 @@
 
 import sys
 import warnings
+import json
 
 if sys.version >= '3':
     basestring = str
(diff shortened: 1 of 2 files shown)


marmbrus:
ok to test

shea-parkes:
Gah, I'm sorry.  I forgot python 2.6 compatibility.  I'll switch this over shortly to use the old-school style.  Likely do something like this:

```python
def alias(self, *alias, **kwargs):
    metadata = kwargs.pop('metadata', None)
    assert not kwargs
```

I'll also add some unit tests and raise a more proper exception (if I can find one).

Sorry again.

marmbrus:
No worries, thanks for working on this!  It's great to ensure our Python APIs aren't lagging behind the Scala ones.

shea-parkes:
Alright, I've pushed up a new commit that should make this compatible with python 2.6.  I don't have a convenient python 2.6 environment to test in, but it still is valid in my python 3 one.  I'll check back in on the CI results in a bit.

shea-parkes:
That last failure was a syntax screw-up in my intentionally failed unit tests; I'm too used to `py.test` style tests.  I've hopefully got the syntax correct now (and went ahead and switch to the cleaner context manager style while I was at it).  I'll check in again after the next round of CI results.

shea-parkes:
Alright @marmbrus , the tests are at least passing now (including the pesky python2.6).  Let me know if you'd like any further changes.

python/pyspark/sql/column.py
```diff
@@ -298,19 +299,34 @@ def isin(self, *cols):
     isNotNull = _unary_op("isNotNull", "True if the current expression is not null.")
 
     @since(1.3)
-    def alias(self, *alias):
+    def alias(self, *alias, **kwargs):
         """
         Returns this column aliased with a new name or names (in the case of expressions that
         return more than one column, such as explode).
 
+        Optional ``metadata`` keyword argument can be passed when aliasing a single column.
```
holdenk: This is very minor, but it might make sense to have a versionChanged directive in the docString so its clear when metaData support was added (since its addiative and shouldn't break existing code more on the optional but possibly useful side).

shea-parkes: I actually thought about this, but I'm struggling to find an example of how to properly document this.  I'll try grepping a little wider to see if I can find a "versionChanged" example somewhere...

shea-parkes:
There we go.  I found other examples of the `versionchanged` directive syntax and added a corresponding entry to `pyspark.sql.Column.alias`.  I think this should be good to look over again.  Let me know if there is anything else you would like added.

python/pyspark/sql/column.py
shea-parkes: Oh, and I inferred that this would be a v2.2 feature since v2.1 is already in pre-release.  If I got the version wrong, just let me know.

holdenk: 2.2 is probably right, although the current 2.1 RC is more a of a strawman so it is possible (but up to @davies / @marmbrus if this warrants going into 2.1).

holdenk:
Lets see if maybe @marmbrus or @davies has some time to look at this?

python/pyspark/sql/column.py
shea-parkes: Congrats on the release of v2.1.  I believe the `versionchanged` target of v2.2 is now accurate.

shea-parkes:
I'm sure everyone was busy over the holidays and with the release of v2.1 (congrats).

Anecdotally, we've been using this functionality a *lot* in our internal fork, so I think it could be quite valuable to get more broadly available.  It also exposes more of the Scala API to Python.

I'd be happy to make any requested changes.  Just let me know what you'd like done.

shea-parkes:
Looks like v2.2 hasn't been branched off yet.  We're still using the dickens out of this feature locally.  It's also a good step forward for Python API parity.  Is there anyone with some time to look at this?

I'd be happy to make any further requested changes, although I think it's in a good state now.  Thanks.

holdenk:
I'm pretty busy in the run up to Spark Summit - let's chat at the summit if you are around or I can take a look at this after once my schedule is back to normal. Sorry for the delay in getting to this @shea-parkes & thanks for sticking with it.

python/pyspark/sql/column.py
holdenk: Excellent, so I think maybe using the `:param:` docstring notation here would be make sense (as in the other functions).

holdenk (commented):
This look reasonable, one minor point with maybe improving the docstring a bit (which will also give us a fresh jenkins test since its been awhile). If you have chance to update the docstring and jenkins passes I'll try and merge this :)

holdenk:
Thank you for updating this, looks good to me! :)

holdenk:
Merged to master :)

PR #16098, opened 2016-12-01
[SPARK-18672][CORE] Close recordwriter in SparkHadoopMapReduceWriter before committing

HyukjinKwon:

It seems some APIs such as `PairRDDFunctions.saveAsHadoopDataset()` do not close the record writer before issuing the commit for the task.

On Windows, the output in the temp directory is being open and output committer tries to rename it from temp directory to the output directory after finishing writing. 

So, it fails to move the file. It seems we should close the writer actually before committing the task like the other writers such as `FileFormatWriter`.

Identified failure was as below:

```
FAILURE! - in org.apache.spark.JavaAPISuite
writeWithNewAPIHadoopFile(org.apache.spark.JavaAPISuite)  Time elapsed: 0.25 sec  <<< ERROR!
org.apache.spark.SparkException: Job aborted.
	at org.apache.spark.JavaAPISuite.writeWithNewAPIHadoopFile(JavaAPISuite.java:1231)
Caused by: org.apache.spark.SparkException: 
Job aborted due to stage failure: Task 0 in stage 0.0 failed 1 times, most recent failure: Lost task 0.0 in stage 0.0 (TID 0, localhost, executor driver): org.apache.spark.SparkException: Task failed while writing rows
	at org.apache.spark.internal.io.SparkHadoopMapReduceWriter$.org$apache$spark$internal$io$SparkHadoopMapReduceWriter$$executeTask(SparkHadoopMapReduceWriter.scala:182)
	at org.apache.spark.internal.io.SparkHadoopMapReduceWriter$$anonfun$3.apply(SparkHadoopMapReduceWriter.scala:100)
	at org.apache.spark.internal.io.SparkHadoopMapReduceWriter$$anonfun$3.apply(SparkHadoopMapReduceWriter.scala:99)
Caused by: java.io.IOException: Could not rename file:/C:/projects/spark/core/target/tmp/1480553515529-0/output/_temporary/0/_temporary/attempt_20161201005155_0000_r_000000_0 to file:/C:/projects/spark/core/target/tmp/1480553515529-0/output/_temporary/0/task_20161201005155_0000_r_000000
	at org.apache.hadoop.mapreduce.lib.output.FileOutputCommitter.commitTask(FileOutputCommitter.java:436)
	at org.apache.hadoop.mapreduce.lib.output.FileOutputCommitter.commitTask(FileOutputCommitter.java:415)
	at org.apache.spark.mapred.SparkHadoopMapRedUtil$.performCommit$1(SparkHadoopMapRedUtil.scala:50)
	at org.apache.spark.mapred.SparkHadoopMapRedUtil$.commitTask(SparkHadoopMapRedUtil.scala:76)
	at org.apache.spark.internal.io.HadoopMapReduceCommitProtocol.commitTask(HadoopMapReduceCommitProtocol.scala:153)
	at org.apache.spark.internal.io.SparkHadoopMapReduceWriter$$anonfun$4.apply(SparkHadoopMapReduceWriter.scala:167)
	at org.apache.spark.internal.io.SparkHadoopMapReduceWriter$$anonfun$4.apply(SparkHadoopMapReduceWriter.scala:156)
	at org.apache.spark.util.Utils$.tryWithSafeFinallyAndFailureCallbacks(Utils.scala:1341)
	at org.apache.spark.internal.io.SparkHadoopMapReduceWriter$.org$apache$spark$internal$io$SparkHadoopMapReduceWriter$$executeTask(SparkHadoopMapReduceWriter.scala:168)
	... 8 more
Driver stacktrace:
	at org.apache.spark.JavaAPISuite.writeWithNewAPIHadoopFile(JavaAPISuite.java:1231)
Caused by: org.apache.spark.SparkException: Task failed while writing rows
Caused by: java.io.IOException: Could not rename file:/C:/projects/spark/core/target/tmp/1480553515529-0/output/_temporary/0/_temporary/attempt_20161201005155_0000_r_000000_0 to file:/C:/projects/spark/core/target/tmp/1480553515529-0/output/_temporary/0/task_20161201005155_0000_r_000000
```

This PR proposes to close this before committing the task.

## How was this patch tested?

Manually tested via AppVeyor. 

**Before**

https://ci.appveyor.com/project/spark-test/spark/build/94-scala-tests

**After**

https://ci.appveyor.com/project/spark-test/spark/build/93-scala-tests

### DIFF
--- a/core/src/main/scala/org/apache/spark/internal/io/SparkHadoopMapReduceWriter.scala
+++ b/core/src/main/scala/org/apache/spark/internal/io/SparkHadoopMapReduceWriter.scala
@@ -146,7 +146,7 @@ object SparkHadoopMapReduceWriter extends Logging {
       case c: Configurable => c.setConf(hadoopConf)
       case _ => ()
     }
-    val writer = taskFormat.getRecordWriter(taskContext)
+    var writer = taskFormat.getRecordWriter(taskContext)
       .asInstanceOf[RecordWriter[K, V]]
     require(writer != null, "Unable to obtain RecordWriter")
     var recordsWritten = 0L
@@ -163,12 +163,20 @@ object SparkHadoopMapReduceWriter extends Logging {
             outputMetricsAndBytesWrittenCallback, recordsWritten)
           recordsWritten += 1
         }
-
+        if (writer != null) {
+          writer.close(taskContext)
+          writer = null
+        }
         committer.commitTask(taskContext)
       }(catchBlock = {
         committer.abortTask(taskContext)
         logError(s"Task ${taskContext.getTaskAttemptID} aborted.")
-      }, finallyBlock = writer.close(taskContext))
+      }, finallyBlock = {
+        if (writer != null) {
+          writer.close(taskContext)
+          writer = null
+        }
+      })
 
       outputMetricsAndBytesWrittenCallback.foreach {
         case (om, callback) =>
--- a/core/src/main/scala/org/apache/spark/internal/io/SparkHadoopMapReduceWriter.scala
+++ b/core/src/main/scala/org/apache/spark/internal/io/SparkHadoopMapReduceWriter.scala
@@ -164,7 +164,10 @@ object SparkHadoopMapReduceWriter extends Logging {
             outputMetricsAndBytesWrittenCallback, recordsWritten)
           recordsWritten += 1
         }
-        writer.close(taskContext)
+        if (writer != null) {
+          writer.close(taskContext)
+          writer = null
+        }
         committer.commitTask(taskContext)
       }(catchBlock = {
         // If there is an error, release resource and then abort the task.
--- a/core/src/main/scala/org/apache/spark/internal/io/SparkHadoopMapReduceWriter.scala
+++ b/core/src/main/scala/org/apache/spark/internal/io/SparkHadoopMapReduceWriter.scala
@@ -169,12 +169,14 @@ object SparkHadoopMapReduceWriter extends Logging {
         }
         committer.commitTask(taskContext)
       }(catchBlock = {
-        committer.abortTask(taskContext)
-        logError(s"Task ${taskContext.getTaskAttemptID} aborted.")
-      }, finallyBlock = {
-        if (writer != null) {
-          writer.close(taskContext)
-          writer = null
+        try {
+          if (writer != null) {
+            writer.close(taskContext)
+            writer = null
+          }
+        } finally {
+          committer.abortTask(taskContext)
+          logError(s"Task ${taskContext.getTaskAttemptID} aborted.")
         }
       })
 
--- a/core/src/main/scala/org/apache/spark/internal/io/SparkHadoopMapReduceWriter.scala
+++ b/core/src/main/scala/org/apache/spark/internal/io/SparkHadoopMapReduceWriter.scala
@@ -154,6 +154,7 @@ object SparkHadoopMapReduceWriter extends Logging {
     // Write all rows in RDD partition.
     try {
       val ret = Utils.tryWithSafeFinallyAndFailureCallbacks {
+        // Write rows out, release resource and commit the task.
         while (iterator.hasNext) {
           val pair = iterator.next()
           writer.write(pair._1, pair._2)
@@ -169,6 +170,7 @@ object SparkHadoopMapReduceWriter extends Logging {
         }
         committer.commitTask(taskContext)
       }(catchBlock = {
+        // If there is an error, release resource and then abort the task.
         try {
           if (writer != null) {
             writer.close(taskContext)
--- a/core/src/main/scala/org/apache/spark/internal/io/SparkHadoopMapReduceWriter.scala
+++ b/core/src/main/scala/org/apache/spark/internal/io/SparkHadoopMapReduceWriter.scala
@@ -164,10 +164,7 @@ object SparkHadoopMapReduceWriter extends Logging {
             outputMetricsAndBytesWrittenCallback, recordsWritten)
           recordsWritten += 1
         }
-        if (writer != null) {
-          writer.close(taskContext)
-          writer = null
-        }
+        writer.close(taskContext)
         committer.commitTask(taskContext)
       }(catchBlock = {
         // If there is an error, release resource and then abort the task.


HyukjinKwon:
Build started: [CORE] `org.apache.spark.JavaAPISuite` [PR-16098](https://ci.appveyor.com/project/spark-test/spark/branch/5DFF30F9-0FAB-4B27-8BE1-377D64B5B653)
Diff: https://github.com/apache/spark/compare/master...spark-test:5DFF30F9-0FAB-4B27-8BE1-377D64B5B653

srowen:
I see, so the writer should be closed before committing?

Can the writer be closed before aborting too? 
In that case, maybe it's cleaner to use nested try-finally? an inner try-finally manages closing the writer, and the outer try-catch commits or aborts.

But this is reasonable too.

HyukjinKwon:
Yes, it seems so because the output seems written in temp dir first and then it is being renamed as the task is committed. Yes, let me try to neat up as possible as I can.

HyukjinKwon:
(cc @mridulm as well who I guess is interested in this issue.)

HyukjinKwon:
Build started: [CORE] `org.apache.spark.JavaAPISuite` [PR-16098](https://ci.appveyor.com/project/spark-test/spark/branch/B67F4C09-43DD-44D7-90F3-058FB436CA44)
Diff: https://github.com/apache/spark/compare/master...spark-test:B67F4C09-43DD-44D7-90F3-058FB436CA44

mridulm:
close is idempotent - simply close it before 'committer.commitTask(taskContext)' and you should be done. The second close in finally will become no-op in case of successful commit, and will handle failure cases.

HyukjinKwon:
Ah, @marmbrus thanks for pointing that out. I see close should be idempotent by design but I am worried if it is guaranteed that they actually are. I mean.. for example, `ParquetRecordWriter`'s `close()` is not idempotent up to my understanding because at least it'd throw `NullPointerException` currently when we try to close multiple times.

For example, If we close this via [ParquetRecordWriter.close()](https://github.com/apache/parquet-mr/blob/apache-parquet-1.8.0/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetRecordWriter.java#L110-L116) which calls [InternalParquetRecordWriter. flushRowGroupToStore()](https://github.com/apache/parquet-mr/blob/apache-parquet-1.8.0/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/InternalParquetRecordWriter.java#L155-L175)  through [InternalParquetRecordWriter.close()](https://github.com/apache/parquet-mr/blob/apache-parquet-1.8.0/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/InternalParquetRecordWriter.java#L113), it would throws at least `NullPointerException` as the column store (containing column writers) becomes `null` for the first close().

HyukjinKwon:
I just manually tested this via `ParquetRecordWriter` and it seems throws an exception as below:

```
java.lang.NullPointerException
	at org.apache.parquet.hadoop.InternalParquetRecordWriter.flushRowGroupToStore(InternalParquetRecordWriter.java:157)
	at org.apache.parquet.hadoop.InternalParquetRecordWriter.close(InternalParquetRecordWriter.java:113)
	at org.apache.parquet.hadoop.ParquetRecordWriter.close(ParquetRecordWriter.java:112)
...
```

HyukjinKwon:
Would this make sense if we leave them just as a safe guard maybe?

mridulm:
That looks like a bug in ParquetRecordWriter (the contract of close() is unambiguous) ... but then, I guess there is no point in fighting against buggy code : we have to integrate with a lot of Closeable's.
If this is common enough bug, my suggestion is a bad idea - thanks for verifying.

Btw, if you are depending on writer being null to prevent 'double close', copy it to a local variable, null it and close the local variable. This prevents close itself throwing exception masking the initial exception.

core/src/main/scala/org/apache/spark/internal/io/SparkHadoopMapReduceWriter.scala
```diff
@@ -163,12 +163,22 @@ object SparkHadoopMapReduceWriter extends Logging {
...
+        try {
+          if (writer != null) {
+            writer.close(taskContext)
+            writer = null
+          }
+        } finally {
+          committer.abortTask(taskContext)
+          logError(s"Task ${taskContext.getTaskAttemptID} aborted.")
+        }
+      })
```
srowen: What about something like

```
val committer = ...
try {
  ...
  val writer = ...
  try {
    ...
  } finally {
    writer.close()
  }
  committer.commitTask(...)
} catch {
  ...
  commiter.abortTask(...)
}
```

It's simpler but am I missing why it wouldn't work?

mridulm:
@srowen This is where try-with-resources is really useful in java.
Util.tryWithSafeFinallyAndFailureCallbacks is an attempt at doing that in scala with mixed results.

I agree, the formulation you proposed is cleaner - the code has migrated a bit since it was initially written using tryWithSafeFinallyAndFailureCallbacks

HyukjinKwon:
Thank you both sincerely. Let me fix it.

core/src/main/scala/org/apache/spark/internal/io/SparkHadoopMapReduceWriter.scala
HyukjinKwon: I think I was worried of trivial and useless things. Let me try to follow this suggestion.

HyukjinKwon:
Build started: [CORE] `org.apache.spark.JavaAPISuite` [PR-16098](https://ci.appveyor.com/project/spark-test/spark/branch/10593DBD-863E-40A2-BC1E-86FDB1555790)
Diff: https://github.com/apache/spark/compare/master...spark-test:10593DBD-863E-40A2-BC1E-86FDB1555790

core/src/main/scala/org/apache/spark/internal/io/SparkHadoopMapReduceWriter.scala
```diff
@@ -146,29 +146,32 @@ object SparkHadoopMapReduceWriter extends Logging {
...
       val ret = Utils.tryWithSafeFinallyAndFailureCallbacks {
-        while (iterator.hasNext) {
-          val pair = iterator.next()
-          writer.write(pair._1, pair._2)
-
-          // Update bytes written metric every few records
-          SparkHadoopWriterUtils.maybeUpdateOutputMetrics(
-            outputMetricsAndBytesWrittenCallback, recordsWritten)
-          recordsWritten += 1
+        val writer = taskFormat.getRecordWriter(taskContext)
```
HyukjinKwon: I was a bit worried of moving this in here because we now abort the task when it fails to create a record reader. Just to convince myself, I at least checked `FileSystem.create(..)` is called within `TextOutputFormat.getRecordWriter` and it seems removed via `FileOutputCommitter.abortTask(..)`.

srowen (approved):

core/src/main/scala/org/apache/spark/internal/io/SparkHadoopMapReduceWriter.scala
```diff
@@ -146,29 +146,32 @@ object SparkHadoopMapReduceWriter extends Logging {
...
+        require(writer != null, "Unable to obtain RecordWriter")
```
HyukjinKwon: Ah, actually, I left a comment about this and then removed because I thought I might be too much worried.

I at least checked `TextOutputFormat. getRecordWriter(..)` is able to create the attempt folder and `FileOutputCommitter.abortTask(..)` removes this. If this is located in the original place, it'd not remove the directory (although this case would be really rare).

HyukjinKwon:
retest this please

HyukjinKwon:
Oh wait, the test seems related with this. Sorry.

HyukjinKwon:
I see. Let me please just revert the change back to https://github.com/apache/spark/pull/16098/commits/66ff918847a57b3ce4b5abbdf56498e3b871bbf0. It seems `TaskContextImpl.markTaskFailed(...)` should be called ahead when the task should be aborted (see https://issues.apache.org/jira/browse/SPARK-13601) and `Utils. tryWithSafeFinallyAndFailureCallbacks` calls this in `catchBlock`. I understand it seems not the cleanest way but it seems they can be refactored together if they have the same logics.

HyukjinKwon:
@srowen and @mridulm Would this be acceptable maybe?

core/src/main/scala/org/apache/spark/internal/io/SparkHadoopMapReduceWriter.scala
```diff
@@ -163,12 +164,23 @@ object SparkHadoopMapReduceWriter extends Logging {
             outputMetricsAndBytesWrittenCallback, recordsWritten)
           recordsWritten += 1
         }
-
+        if (writer != null) {
```
srowen: I don't think writer can be null here, but no big deal.

core/src/main/scala/org/apache/spark/internal/io/SparkHadoopMapReduceWriter.scala
```diff
@@ -163,12 +164,23 @@ object SparkHadoopMapReduceWriter extends Logging {
...
+          writer = null
+        }
         committer.commitTask(taskContext)
       }(catchBlock = {
-        committer.abortTask(taskContext)
-        logError(s"Task ${taskContext.getTaskAttemptID} aborted.")
-      }, finallyBlock = writer.close(taskContext))
+        // If there is an error, release resource and then abort the task.
+        try {
+          if (writer != null) {
```
srowen: So this is also adding a defensive close before aborting, not just before committing? if that's important at all, OK. It didn't do that before.

HyukjinKwon: Oh yes. Thank you. Will address the comment above.

HyukjinKwon:
Build started: [CORE] `org.apache.spark.JavaAPISuite` [PR-16098](https://ci.appveyor.com/project/spark-test/spark/branch/C6FA0845-A500-4CB8-A511-B7B5C418691F)
Diff: https://github.com/apache/spark/compare/master...spark-test:C6FA0845-A500-4CB8-A511-B7B5C418691F

core/src/main/scala/org/apache/spark/internal/io/SparkHadoopMapReduceWriter.scala
```diff
@@ -163,12 +164,20 @@ object SparkHadoopMapReduceWriter extends Logging {
             outputMetricsAndBytesWrittenCallback, recordsWritten)
           recordsWritten += 1
         }
-
+        writer.close(taskContext)
```
srowen: Ah, but do you still want to null it here so that it doesn't also cause a close() in the catch block? the commit could fail.

HyukjinKwon: Doh!

HyukjinKwon:
Build started: [CORE] `org.apache.spark.JavaAPISuite` [PR-16098](https://ci.appveyor.com/project/spark-test/spark/branch/EDC7F98E-6542-4A51-A725-D0BEF753E438)
Diff: https://github.com/apache/spark/compare/master...spark-test:EDC7F98E-6542-4A51-A725-D0BEF753E438

HyukjinKwon:
It seems the failed test is not related and passed in my local.

HyukjinKwon:
retest this please

HyukjinKwon:
retest this please

HyukjinKwon:
retest this please

srowen:
Merged to master

PR #16099, opened 2016-12-01
[SPARK-18665][SQL] set statement state to "ERROR" after user cancel job

cenyuhai:
set statement state to error after user canceled job

### DIFF
--- a/sql/hive-thriftserver/src/main/scala/org/apache/spark/sql/hive/thriftserver/SparkExecuteStatementOperation.scala
+++ b/sql/hive-thriftserver/src/main/scala/org/apache/spark/sql/hive/thriftserver/SparkExecuteStatementOperation.scala
@@ -241,6 +241,8 @@ private[hive] class SparkExecuteStatementOperation(
       dataTypes = result.queryExecution.analyzed.output.map(_.dataType).toArray
     } catch {
       case e: HiveSQLException =>
+        HiveThriftServer2.listener.onStatementError(
+          statementId, e.getMessage, SparkUtils.exceptionString(e))
         if (getStatus().getState() == OperationState.CANCELED) {
           return
         } else {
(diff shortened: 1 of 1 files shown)


rxin:
cc @liancheng

sql/hive-thriftserver/src/main/scala/org/apache/spark/sql/hive/thriftserver/SparkExecuteStatementOperation.scala
```diff
@@ -241,6 +241,8 @@ private[hive] class SparkExecuteStatementOperation(
       dataTypes = result.queryExecution.analyzed.output.map(_.dataType).toArray
     } catch {
       case e: HiveSQLException =>
+        HiveThriftServer2.listener.onStatementError(
+          statementId, e.getMessage, SparkUtils.exceptionString(e))
```
BruceXu1991: I tried your code. there seems have a bug.
e.getMessage may get a null value, which may cause NPE in ThriftServerPage.errorMessageCell(detail)

gatorsmile:
cc @cenyuhai

sql/hive-thriftserver/src/main/scala/org/apache/spark/sql/hive/thriftserver/SparkExecuteStatementOperation.scala
BruceXu1991: This PR still occurs in Spark 2.2.1. A workaround is to use "String.valueOf(e.getMessage)" instead of "e.getMessage" to avoid of NPE problem.

cenyuhai: ok, but My pr is closed by community...

cenyuhai: please catch the exception in ThriftServerPage.errorMessageCell. Without my pr, that function still throw exception.

cenyuhai:
@gatorsmile two years passed... I don't know what to say.

tooptoop4:
@cenyuhai want to open new pr?

PR #16102, opened 2016-12-01
[SPARK-18586][BUILD] netty-3.8.0.Final.jar has vulnerability CVE-2014-3488 and CVE-2014-0193

srowen:

Force update to latest Netty 3.9.x, for dependencies like Flume, to resolve two CVEs. 3.9.2 is the first version that resolves both, and, this is the latest in the 3.9.x line.

## How was this patch tested?

Existing tests

### DIFF
--- a/dev/deps/spark-deps-hadoop-2.2
+++ b/dev/deps/spark-deps-hadoop-2.2
@@ -122,7 +122,7 @@ metrics-graphite-3.1.2.jar
 metrics-json-3.1.2.jar
 metrics-jvm-3.1.2.jar
 minlog-1.3.0.jar
-netty-3.8.0.Final.jar
+netty-3.9.9.Final.jar
 netty-all-4.0.42.Final.jar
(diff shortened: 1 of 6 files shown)


rxin:
Do these libraries actually use Netty in Spark?

srowen:
@rxin it's a weird one. This is all I know:

https://issues.apache.org/jira/browse/SPARK-17875
https://github.com/apache/spark/pull/15436

It's unused but couldn't get the Python tests to pass without it. It's ultimately related to the flume assembly, but I could not see the problem.

Therefore I think this change will basically have no effect either way. I'm OK with merging this or punting, since I don't think the CVE actually has any effect on Spark as a result.

rxin:
OK up to you. I don't have a strong preference either way.

srowen:
Merged to master, though as I say I don't think the CVE actually impacted Spark to begin with.

PR #16104, opened 2016-12-01
[SPARK-18675][SQL] CTAS for hive serde table should work for all hive versions

cloud-fan:

Before hive 1.1, when inserting into a table, hive will create the staging directory under a common scratch directory. After the writing is finished, hive will simply empty the table directory and move the staging directory to it.

After hive 1.1, hive will create the staging directory under the table directory, and when moving staging directory to table directory, hive will still empty the table directory, but will exclude the staging directory there.

In `InsertIntoHiveTable`, we simply copy the code from hive 1.2, which means we will always create the staging directory under the table directory, no matter what the hive version is. This causes problems if the hive version is prior to 1.1, because the staging directory will be removed by hive when hive is trying to empty the table directory.

This PR copies the code from hive 0.13, so that we have 2 branches to create staging directory. If hive version is prior to 1.1, we'll go to the old style branch(i.e. create the staging directory under a common scratch directory), else, go to the new style branch(i.e. create the staging directory under the table directory)

## How was this patch tested?

new test

### DIFF
--- a/sql/hive/src/main/scala/org/apache/spark/sql/hive/execution/InsertIntoHiveTable.scala
+++ b/sql/hive/src/main/scala/org/apache/spark/sql/hive/execution/InsertIntoHiveTable.scala
@@ -126,16 +126,15 @@ case class InsertIntoHiveTable(
   }
 
   def getExternalTmpPath(path: Path): Path = {
-    val hiveVersion = externalCatalog.asInstanceOf[HiveExternalCatalog].client.version.fullVersion
-    if (hiveVersion.startsWith("0.12") ||
-      hiveVersion.startsWith("0.13") ||
-      hiveVersion.startsWith("0.14") ||
-      hiveVersion.startsWith("1.0")) {
+    import org.apache.spark.sql.hive.client.hive._
+
+    val hiveVersion = externalCatalog.asInstanceOf[HiveExternalCatalog].client.version
+    if (hiveVersion == v12 || hiveVersion == v13 || hiveVersion == v14 || hiveVersion == v1_0) {
       oldStyleExternalTempPath(path)
-    } else if (hiveVersion.startsWith("1.1") || hiveVersion.startsWith("1.2")) {
+    } else if (hiveVersion == v1_1 || hiveVersion == v1_2) {
       newStyleExternalTempPath(path)
     } else {
-      throw new IllegalStateException("Unsupported hive version: " + hiveVersion)
+      throw new IllegalStateException("Unsupported hive version: " + hiveVersion.fullVersion)
     }
   }
 
--- a/sql/hive/src/main/scala/org/apache/spark/sql/hive/execution/InsertIntoHiveTable.scala
+++ b/sql/hive/src/main/scala/org/apache/spark/sql/hive/execution/InsertIntoHiveTable.scala
@@ -129,17 +129,26 @@ case class InsertIntoHiveTable(
     import org.apache.spark.sql.hive.client.hive._
 
     val hiveVersion = externalCatalog.asInstanceOf[HiveExternalCatalog].client.version
+    // Before Hive 1.1, when inserting into a table, Hive will create the staging directory under
+    // a common scratch directory. After the writing is finished, Hive will simply empty the table
+    // directory and move the staging directory to it.
+    // After Hive 1.1, Hive will create the staging directory under the table directory, and when
+    // moving staging directory to table directory, Hive will still empty the table directory, but
+    // will exclude the staging directory there.
+    // We have to follow the Hive behavior here, to avoid troubles. For example, if we create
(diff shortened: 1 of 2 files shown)


cloud-fan:
cc @yhuai @gatorsmile

dongjoon-hyun:
Retest this please.

dongjoon-hyun:
Last time,
```
[info] - set hive.metastore.warehouse.dir *** FAILED *** (5 minutes, 0 seconds)
```
At this time,
```
[info] - SPARK-18360: default table path of tables in default database should depend on the location of default database *** FAILED *** (5 minutes, 0 seconds)
```
It's strange.

gatorsmile:
Will review this PR tonight. Thanks!

gatorsmile:
retest this please

sql/hive/src/main/scala/org/apache/spark/sql/hive/execution/InsertIntoHiveTable.scala
```diff
@@ -86,14 +85,15 @@ case class InsertIntoHiveTable(
 
   val hadoopConf = sessionState.newHadoopConf()
   val stagingDir = hadoopConf.get("hive.exec.stagingdir", ".hive-staging")
+  val scratchDir = hadoopConf.get("hive.exec.scratchdir", "/tmp/hive")
```
gatorsmile: The default value of the `hive.exec.scratchdir` also depends on the version. 
> Default Value: /tmp/${user.name} in Hive 0.2.0 through 0.8.0; /tmp/hive-${user.name} in Hive 0.8.1 through 0.14.0; or /tmp/hive in Hive 0.14.0 and later

sql/hive/src/main/scala/org/apache/spark/sql/hive/execution/InsertIntoHiveTable.scala
```diff
@@ -121,21 +121,61 @@ case class InsertIntoHiveTable(
...
+    var dirPath = new Path(
+      extURI.getScheme,
+      extURI.getAuthority,
+      scratchPath.toUri.getPath + "-" + TaskRunner.getTaskRunnerID())
+
+    try {
+      val fs: FileSystem = dirPath.getFileSystem(hadoopConf)
+      dirPath = new Path(fs.makeQualified(dirPath).toString())
+
+      if (!FileUtils.mkdir(fs, dirPath, true, hadoopConf)) {
```
gatorsmile: uh, we drop it after the normal termination of vm by calling `fs.deleteOnExit`

gatorsmile: Keeping these useless directories and files might not make sense, right? Many many files could accumulate?

cloud-fan: I followed the way we clean up staging dir in https://github.com/apache/spark/pull/16104/files#diff-d579db9a8f27e0bbef37720ab14ec3f6R114

sql/hive/src/main/scala/org/apache/spark/sql/hive/execution/InsertIntoHiveTable.scala
cloud-fan: yea that's true, but I don't think the default value here is a big deal and need to worry about.

cloud-fan:
retest this please

sql/hive/src/main/scala/org/apache/spark/sql/hive/execution/InsertIntoHiveTable.scala
```diff
@@ -121,21 +121,61 @@ case class InsertIntoHiveTable(
...
+    val hiveVersion = externalCatalog.asInstanceOf[HiveExternalCatalog].client.version.fullVersion
+    if (hiveVersion.startsWith("0.12") ||
+      hiveVersion.startsWith("0.13") ||
+      hiveVersion.startsWith("0.14") ||
+      hiveVersion.startsWith("1.0")) {
+      oldStyleExternalTempPath(path)
+    } else if (hiveVersion.startsWith("1.1") || hiveVersion.startsWith("1.2")) {
+      newStyleExternalTempPath(path)
+    } else {
+      throw new IllegalStateException("Unsupported hive version: " + hiveVersion)
```
gatorsmile: Do we have to issue an exception here? It sounds like this is the only place we block it. Do you think users might be able to use the higher version, although it might have a few issues?

gatorsmile: uh, I see. Thanks!

gatorsmile:
LGTM

yhuai:
LGTM. Thanks. Merging to master

PR #16106, opened 2016-12-01
[SPARK-17213][SQL] Disable Parquet filter push-down for string and binary columns due to PARQUET-686

liancheng:
This PR targets to both master and branch-2.1.

Due to PARQUET-686, Parquet doesn't do string comparison correctly while doing filter push-down for string columns. This PR disables filter push-down for both string and binary columns to work around this issue. Binary columns are also affected because some Parquet data models (like Hive) may store string columns as a plain Parquet `binary` instead of a `binary (UTF8)`.

## How was this patch tested?

New test case added in `ParquetFilterSuite`.


liancheng:
retest this please

liancheng:
The last build failure doesn't seem to be relevant.

rxin:

PR #16107, opened 2016-12-01
SPARK-18677: Fix parsing ['key'] in JSON path expressions.

rdblue:

This fixes the parser rule to match named expressions, which doesn't work for two reasons:
1. The name match is not coerced to a regular expression (missing .r)
2. The surrounding literals are incorrect and attempt to escape a single quote, which is unnecessary

## How was this patch tested?

This adds test cases for named expressions using the bracket syntax, including one with quoted spaces.


rxin:
cc @hvanhovell

srowen:
Seems correct to me, yes.

NathanHowell:
I wrote the buggy version, doh... but this LGTM. Thanks for fix.

hvanhovell:

rdblue:
Thanks for the quick review!

PR #16108, opened 2016-12-01
[SPARK-18670][SS]Limit the number of StreamingQueryListener.StreamProgressEvent when there is no data

zsxwing:

This PR adds a sql conf `spark.sql.streaming.noDataReportInterval` to control how long to wait before outputing the next StreamProgressEvent when there is no data.

## How was this patch tested?

The added unit test.

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamExecution.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamExecution.scala
@@ -63,6 +63,8 @@ class StreamExecution(
 
   private val pollingDelayMs = sparkSession.sessionState.conf.streamingPollingDelay
 
+  private val noDataReportInterval = sparkSession.sessionState.conf.streamingNoDataReportInterval
+
   /**
    * A lock used to wait/notify when batches complete. Use a fair lock to avoid thread starvation.
    */
@@ -196,6 +198,9 @@ class StreamExecution(
       // While active, repeatedly attempt to run batches.
       SparkSession.setActiveSession(sparkSession)
 
+      // The timestamp we report an event that has no input data
+      var noDataReportTimestamp = Long.MinValue
+
       triggerExecutor.execute(() => {
         startTrigger()
 
@@ -218,7 +223,15 @@ class StreamExecution(
 
             // Report trigger as finished and construct progress object.
             finishTrigger(dataAvailable)
-            postEvent(new QueryProgressEvent(lastProgress))
+            if (dataAvailable) {
+              postEvent(new QueryProgressEvent(lastProgress))
+            } else {
+              val now = triggerClock.getTimeMillis()
+              if (now - noDataReportInterval >= noDataReportTimestamp) {
+                noDataReportTimestamp = now
+                postEvent(new QueryProgressEvent(lastProgress))
+              }
+            }
 
             if (dataAvailable) {
               // We'll increase currentBatchId after we complete processing current batch's data
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamExecution.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamExecution.scala
@@ -63,7 +63,8 @@ class StreamExecution(
 
   private val pollingDelayMs = sparkSession.sessionState.conf.streamingPollingDelay
 
-  private val noDataEventInterval = sparkSession.sessionState.conf.streamingNoDataEventInterval
+  private val noDataProgressEventInterval =
+    sparkSession.sessionState.conf.streamingNoDataProgressEventInterval
 
   /**
    * A lock used to wait/notify when batches complete. Use a fair lock to avoid thread starvation.
@@ -199,7 +200,7 @@ class StreamExecution(
       SparkSession.setActiveSession(sparkSession)
 
       // The timestamp we report an event that has no input data
-      var noDataEventTimestamp = Long.MinValue
+      var lastNoDataProgressEventTime = Long.MinValue
 
       triggerExecutor.execute(() => {
         startTrigger()
@@ -225,12 +226,12 @@ class StreamExecution(
             finishTrigger(dataAvailable)
             if (dataAvailable) {
               // Reset noDataEventTimestamp if we processed any data
-              noDataEventTimestamp = Long.MinValue
+              lastNoDataProgressEventTime = Long.MinValue
               postEvent(new QueryProgressEvent(lastProgress))
             } else {
               val now = triggerClock.getTimeMillis()
-              if (now - noDataEventInterval >= noDataEventTimestamp) {
-                noDataEventTimestamp = now
+              if (now - noDataProgressEventInterval >= lastNoDataProgressEventTime) {
+                lastNoDataProgressEventTime = now
                 postEvent(new QueryProgressEvent(lastProgress))
               }
             }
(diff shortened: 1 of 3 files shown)


zsxwing:
cc @tdas

sql/core/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala
```diff
@@ -603,6 +603,13 @@ object SQLConf {
       .timeConf(TimeUnit.MILLISECONDS)
       .createWithDefault(10L)
 
+  val STREAMING_NO_DATA_REPORT_INTERVAL =
```
tdas: `Report` is not a term used anywhere. How about streamingNoDataEventInterval

sql/core/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala
```diff
@@ -603,6 +603,13 @@ object SQLConf {
...
+    SQLConfigBuilder("spark.sql.streaming.noDataReportInterval")
+      .internal()
+      .doc("How long to wait between two progress events when there is no data")
```
tdas: ... (in ms) ...

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamExecution.scala
```diff
@@ -218,7 +223,15 @@ class StreamExecution(
 
             // Report trigger as finished and construct progress object.
             finishTrigger(dataAvailable)
-            postEvent(new QueryProgressEvent(lastProgress))
+            if (dataAvailable) {
+              postEvent(new QueryProgressEvent(lastProgress))
+            } else {
+              val now = triggerClock.getTimeMillis()
```
tdas: Shouldnt there be an event every time the system switches from dataAvailable = true, to dataAvailable = false. This logic wont do that right?

sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamingQueryListenerSuite.scala
```diff
@@ -191,6 +192,48 @@ class StreamingQueryListenerSuite extends StreamTest with BeforeAndAfter {
     assert(queryQueryTerminated.exception === newQueryTerminated.exception)
   }
 
+  test("noDataReportInterval") {
+    withSQLConf(SQLConf.STREAMING_NO_DATA_REPORT_INTERVAL.key -> "100ms") {
+      @volatile var progressEventCount = 0
+
+      val listener = new StreamingQueryListener {
```
tdas: can you write a short explanation on what this test does. it hard to understand.

sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamingQueryListenerSuite.scala
```diff
@@ -191,6 +192,48 @@ class StreamingQueryListenerSuite extends StreamTest with BeforeAndAfter {
     assert(queryQueryTerminated.exception === newQueryTerminated.exception)
   }
 
+  test("noDataReportInterval") {
```
tdas: nit: make the name better - only one progressevent per interval when no data

sql/core/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala
zsxwing: TimeConf supports various units like `10s`, `100ms`. The unit here is telling which unit we want to use when getting this conf.

sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamingQueryListenerSuite.scala
```diff
@@ -191,6 +192,50 @@ class StreamingQueryListenerSuite extends StreamTest with BeforeAndAfter {
...
+            assert(progressEventCount > 0)
+          }
+          true
+        }
+        for (_ <- 1 to 100) {
+          actions += AdvanceManualClock(10)
+          actions += AssertOnQuery { _ =>
+            // Sleep so that if the config `noDataEventInterval` doesn't work, it has enough time
+            // to report too many events.
+            Thread.sleep(10)
```
tdas: this combination of manual clock and real time sleep is quite confusing

sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamingQueryListenerSuite.scala
```diff
@@ -191,6 +192,50 @@ class StreamingQueryListenerSuite extends StreamTest with BeforeAndAfter {
...
+            true
+          }
+        }
+        testStream(MemoryStream[Int].toDS)(actions: _*)
+        // 11 is the max value of the possible numbers of events.
```
tdas: I think this test looks a little complicated. Also you are assuming the fact the when there is no data, it will sleep for 10 ms real time. why not something like this. 
```
val input = new MemoryStream() {
  @volatile var numTriggers = 0
  override def getOffset(): Offset = { numTriggers +=1 ; super.getOffset() } 
}
...
// add listener with progressEventCount and flag isTerminationEventPosted
...
testStream() {   // on real clock
   AddData(...)
   AssertOnQuery {  q => 
      eventually (timeout(streamingTimeout)) {
          assert(input.numTriggers > 100)   // at least 100 triggers have occurred
      }
   }, 
   StopStream, 
   AssertOnQuery { q => 
     eventually(timeout(streamingTimeout) {
       assert(isTerminationEventPosted)   // event queue has drained
       assert(numProgressEvent >=1 && numProgressEvent <= 100)  // no more than 10 events. 
    } 
   }
}

```

tdas (commented):
I think the test can be simplified.

sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamingQueryListenerSuite.scala
tdas: Okay, that wont work as well, because there is not guarantee how long the 100 triggers would take. It may take 100 seconds in a slow machine, in which numProgress events could legitimately have 100 events.

sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamingQueryListenerSuite.scala
```diff
@@ -191,6 +193,48 @@ class StreamingQueryListenerSuite extends StreamTest with BeforeAndAfter {
...
+        actions += AssertOnQuery { _ =>
+          eventually(timeout(streamingTimeout)) {
+            assert(input.numTriggers > 100) // at least 100 triggers have occurred
+          }
+          true
+        }
+        testStream(input.toDS)(actions: _*)
+        spark.sparkContext.listenerBus.waitUntilEmpty(10000)
+        // 11 is the max value of the possible numbers of events.
+        assert(numProgressEvent >= 1 && numProgressEvent <= 11)
```
tdas: Shouldnt this be > 1? What if there is a bug that it only outputs once when isData goes from true -> false, and never outputs again, irrespective of the noDataInterval?

tdas (commented):
Test looks good now. But there are other comments that need to be addressed. The name is still confusing, in a different way now. I should have thought of that before suggesting.

tdas:
LGTM. Merging to master and 2.1

PR #16112, opened 2016-12-02
[SPARK-18679] [SQL] Fix regression in file listing performance for non-catalog tables

ericl:

In Spark 2.1 ListingFileCatalog was significantly refactored (and renamed to InMemoryFileIndex). This introduced a regression where parallelism could only be introduced at the very top of the tree. However, in many cases (e.g. `spark.read.parquet(topLevelDir)`), the top of the tree is only a single directory.

This PR simplifies and fixes the parallel recursive listing code to allow parallelism to be introduced at any level during recursive descent (though note that once we decide to list a sub-tree in parallel, the sub-tree is listed in serial on executors).

cc @mallman  @cloud-fan 

## How was this patch tested?

Checked metrics in unit tests.


cloud-fan:
LGTM, @ericl have you run some local benchmark to make sure the performance regression is fixed?

ericl:
Yep

cloud-fan:

PR #16113, opened 2016-12-02
[SPARK-18657][SPARK-18668] Make StreamingQuery.id persists across restart and not auto-generate StreamingQuery.name

tdas:
Here are the major changes in this PR. 
- Added the ability to recover `StreamingQuery.id` from checkpoint location, by writing the id to `checkpointLoc/metadata`.
- Added `StreamingQuery.runId` which is unique for every query started and does not persist across restarts. This is to identify each restart of a query separately (same as earlier behavior of `id`).
- Removed auto-generation of `StreamingQuery.name`. The purpose of name was to have the ability to define an identifier across restarts, but since id is precisely that, there is no need for a auto-generated name. This means name becomes purely cosmetic, and is null by default.
- Added `runId` to `StreamingQueryListener` events and `StreamingQueryProgress`.

Implementation details
- Renamed existing `StreamExecutionMetadata` to `OffsetSeqMetadata`, and moved it to the file `OffsetSeq.scala`, because that is what this metadata is tied to. Also did some refactoring to make the code cleaner (got rid of a lot of `.json` and `.getOrElse("{}")`).
- Added the `id` as the new `StreamMetadata`. 
- When a StreamingQuery is created it gets or writes the `StreamMetadata` from `checkpointLoc/metadata`.
- All internal logging in `StreamExecution` uses `(name, id, runId)` instead of just `name`

TODO
- [x] Test handling of name=null in json generation of StreamingQueryProgress
- [x] Test handling of name=null in json generation of StreamingQueryListener events
- [x] Test python API of runId

## How was this patch tested?
Updated unit tests and new unit tests


sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamExecution.scala
```diff
@@ -605,34 +629,64 @@ class StreamExecution(
...
+ * in the checkpoint location the first time a query is started and recovered every time the query
+ * is restarted.
  *
- * @param batchWatermarkMs: The current eventTime watermark, used to
- * bound the lateness of data that will processed. Time unit: milliseconds
- * @param batchTimestampMs: The current batch processing timestamp.
- * Time unit: milliseconds
+ * @param id  unique id of the [[StreamingQuery]] that needs to be persisted across restarts
  */
-case class StreamExecutionMetadata(
```
tdas: This has been renamed to OffsetSeqLog and moved to OffsetSeq.scala

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/OffsetSeq.scala
```diff
@@ -54,6 +61,26 @@ object OffsetSeq {
...
+/**
+ * Contains metadata associated with a [[OffsetSeq]]. This information is
+ * persisted to the offset log in the checkpoint location via the [[OffsetSeq]] metadata field.
+ *
+ * @param batchWatermarkMs: The current eventTime watermark, used to
+ * bound the lateness of data that will processed. Time unit: milliseconds
+ * @param batchTimestampMs: The current batch processing timestamp.
+ * Time unit: milliseconds
+ */
+case class OffsetSeqMetadata(var batchWatermarkMs: Long = 0, var batchTimestampMs: Long = 0) {
```
marmbrus: Can we put this in its own file?

sql/core/src/main/scala/org/apache/spark/sql/streaming/StreamingQuery.scala
```diff
@@ -32,21 +32,33 @@ import org.apache.spark.sql.SparkSession
...
    * @since 2.0.0
    */
   def name: String
 
   /**
-   * Returns the unique id of this query.
+   * Returns the unique id of this query that persists across restarts from checkpoint data.
+   * That is, this id is generated when a query is started for the first time, and
+   * will be the same every time it is restarted from checkpoint data.
+   * There can only be one query with the same id active in a Spark cluster.
```
marmbrus: This makes it sound like its okay to have more than one running as long as they aren't on the same spark cluster.

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamExecution.scala
```diff
@@ -605,34 +629,64 @@ class StreamExecution(
...
-    var batchWatermarkMs: Long = 0,
-    var batchTimestampMs: Long = 0) {
-  private implicit val formats = StreamExecutionMetadata.formats
-
-  /**
-   * JSON string representation of this object.
-   */
-  def json: String = Serialization.write(this)
+case class StreamMetadata(id: String) {
```
marmbrus: Can we move these to their own file?

marmbrus (approved):
Only minor comments.  LGTM

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/OffsetSeq.scala
tdas: Sure. But its a small class and closely tied with OffsetSeq, so I thought its not worth having a separate file for this.

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamExecution.scala
tdas: Sure.

sql/core/src/main/scala/org/apache/spark/sql/streaming/StreamingQuery.scala
tdas: I will just remove that line then.

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/OffsetSeq.scala
tdas: Not worth moving these 6 lines of code in a new file.

tdas:
Merging to master and 2.1

PR #16114, opened 2016-12-02
[SPARK-18620][Streaming][Kinesis] Flatten input rates in timeline for streaming + kinesis

maropu:
This pr is to make input rates in timeline more flat for spark streaming + kinesis.
Since kinesis workers fetch records and push them into block generators in bulk, timeline in web UI has many spikes when `maxRates` applied (See a Figure.1 below). This fix splits fetched input records into multiple `adRecords` calls.

Figure.1 Apply `maxRates=500` in vanilla Spark
apply_limit in_vanilla_spark

Figure.2 Apply `maxRates=500` in Spark with my patch
apply_limit in_spark_with_my_patch

## How was this patch tested?
Add tests to check to split input records into multiple `addRecords` calls.

### DIFF
--- a/external/kinesis-asl/src/main/scala/org/apache/spark/streaming/kinesis/KinesisReceiver.scala
+++ b/external/kinesis-asl/src/main/scala/org/apache/spark/streaming/kinesis/KinesisReceiver.scala
@@ -221,6 +221,12 @@ private[kinesis] class KinesisReceiver[T](
     }
   }
 
+  /** Return the current rate limit defined in [[BlockGenerator]]. */
+  private[kinesis] def getCurrentLimit: Int = {
+    assert(blockGenerator != null)
+    math.min(blockGenerator.getCurrentLimit, Int.MaxValue).toInt
+  }
+
   /** Get the latest sequence number for the given shard that can be checkpointed through KCL */
   private[kinesis] def getLatestSeqNumToCheckpoint(shardId: String): Option[String] = {
     Option(shardIdToLatestStoredSeqNum.get(shardId))
--- a/external/kinesis-asl/src/main/scala/org/apache/spark/streaming/kinesis/KinesisRecordProcessor.scala
+++ b/external/kinesis-asl/src/main/scala/org/apache/spark/streaming/kinesis/KinesisRecordProcessor.scala
@@ -18,7 +18,6 @@ package org.apache.spark.streaming.kinesis
 
 import java.util.List
 
-import scala.collection.JavaConverters._
 import scala.util.Random
 import scala.util.control.NonFatal
 
@@ -74,11 +73,13 @@ private[kinesis] class KinesisRecordProcessor[T](receiver: KinesisReceiver[T], w
         // in `KinesisClientLibConfiguration`. For example, if we set 10 to the number of max
         // records in a worker and a producer aggregates two records into one message, the worker
         // possibly 20 records every callback function called.
-        batch.asScala.grouped(receiver.getCurrentLimit).foreach { batch =>
-          receiver.addRecords(shardId, batch.asJava)
-          logDebug(s"Stored: Worker $workerId stored ${batch.size} records for shardId $shardId")
-          receiver.setCheckpointer(shardId, checkpointer)
+        val maxRecords = receiver.getCurrentLimit
+        for (start <- 0 until batch.size by maxRecords) {
+          val miniBatch = batch.subList(start, math.min(start + maxRecords, batch.size))
+          receiver.addRecords(shardId, miniBatch)
         }
+        logDebug(s"Stored: Worker $workerId stored ${batch.size} records for shardId $shardId")
+        receiver.setCheckpointer(shardId, checkpointer)
       } catch {
         case NonFatal(e) =>
           /*
--- a/external/kinesis-asl/src/main/scala/org/apache/spark/streaming/kinesis/KinesisRecordProcessor.scala
+++ b/external/kinesis-asl/src/main/scala/org/apache/spark/streaming/kinesis/KinesisRecordProcessor.scala
@@ -77,8 +77,9 @@ private[kinesis] class KinesisRecordProcessor[T](receiver: KinesisReceiver[T], w
         for (start <- 0 until batch.size by maxRecords) {
           val miniBatch = batch.subList(start, math.min(start + maxRecords, batch.size))
           receiver.addRecords(shardId, miniBatch)
+          logDebug(s"Stored: Worker $workerId stored ${miniBatch.size} records " +
+            s"for shardId $shardId")
         }
-        logDebug(s"Stored: Worker $workerId stored ${batch.size} records for shardId $shardId")
         receiver.setCheckpointer(shardId, checkpointer)
       } catch {
         case NonFatal(e) =>
--- a/external/kinesis-asl/src/main/scala/org/apache/spark/streaming/kinesis/KinesisRecordProcessor.scala
+++ b/external/kinesis-asl/src/main/scala/org/apache/spark/streaming/kinesis/KinesisRecordProcessor.scala
@@ -79,11 +79,11 @@ private[kinesis] class KinesisRecordProcessor[T](receiver: KinesisReceiver[T], w
       val rem = batch.size % maxRecords
       for (idx <- 0 until numIter) {
         val fromIdx = idx * maxRecords
-        val toIdx = fromIdx + maxRecords - 1
-        addRecords(batch.subList(fromIdx, toIdx), checkpointer)
+        val untilIdx = fromIdx + maxRecords
+        addRecords(batch.subList(fromIdx, untilIdx), checkpointer)
       }
       if (rem > 0) {
-        addRecords(batch.subList(numIter * maxRecords, batch.size - 1), checkpointer)
+        addRecords(batch.subList(numIter * maxRecords, batch.size), checkpointer)
       }
     }
   }
--- a/external/kinesis-asl/src/main/scala/org/apache/spark/streaming/kinesis/KinesisRecordProcessor.scala
+++ b/external/kinesis-asl/src/main/scala/org/apache/spark/streaming/kinesis/KinesisRecordProcessor.scala
@@ -18,7 +18,7 @@ package org.apache.spark.streaming.kinesis
 
 import java.util.List
 
-import scala.collection.JavaConversions._
+import scala.collection.JavaConverters._
 import scala.util.Random
 import scala.util.control.NonFatal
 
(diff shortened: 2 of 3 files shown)


dav009:
:+1: just had a play with it, it solves my original issue.

external/kinesis-asl/src/main/scala/org/apache/spark/streaming/kinesis/KinesisRecordProcessor.scala
```diff
@@ -56,6 +56,38 @@ private[kinesis] class KinesisRecordProcessor[T](receiver: KinesisReceiver[T], w
...
+   * in a worker and a producer aggregates two records into one message, the worker possibly
+   * 20 records every callback function called.
+   */
+  private def processRecordsWithLimit(
+      batch: List[Record], checkpointer: IRecordProcessorCheckpointer): Unit = {
+    val maxRecords = receiver.getCurrentLimit
+    if (batch.size() <= maxRecords) {
+      addRecords(batch, checkpointer)
+    } else {
+      val numIter = batch.size / maxRecords
```
srowen: Is this clause a bit simpler as ...

```
for (start <- 0 until batch.size by maxRecords) {
  addRecords(batch.sublist(start, math.min(start + maxRecords, batch.size)), checkpointer)
}
```

external/kinesis-asl/src/main/scala/org/apache/spark/streaming/kinesis/KinesisReceiver.scala
```diff
@@ -221,6 +221,12 @@ private[kinesis] class KinesisReceiver[T](
     }
   }
 
+  /** Return the current rate limit defined in [[BlockGenerator]]. */
+  private[kinesis] def getCurrentLimit: Int = {
+    assert(blockGenerator != null)
```
srowen: This is pretty trivial but do we use runtime assertions in general in the project? the next line fails already when it's null whether assertions are on or not.

external/kinesis-asl/src/main/scala/org/apache/spark/streaming/kinesis/KinesisRecordProcessor.scala
maropu: Thanks! I'll fix

external/kinesis-asl/src/main/scala/org/apache/spark/streaming/kinesis/KinesisRecordProcessor.scala
```diff
@@ -56,6 +56,31 @@ private[kinesis] class KinesisRecordProcessor[T](receiver: KinesisReceiver[T], w
...
+   * control the number of aggregated records to be fetched even if we set `MaxRecords`
+   * in `KinesisClientLibConfiguration`. For example, if we set 10 to the number of max records
+   * in a worker and a producer aggregates two records into one message, the worker possibly
+   * 20 records every callback function called.
+   */
+  private def processRecordsWithLimit(
+      batch: List[Record], checkpointer: IRecordProcessorCheckpointer): Unit = {
+    val maxRecords = receiver.getCurrentLimit
+    if (batch.size() <= maxRecords) {
+      addRecords(batch, checkpointer)
```
srowen: I think the for loop even takes care of this case, but no big deal either way. It seems like a reasonable change.

maropu: Aha, I see. I'll fix, thanks!

maropu:
@srowen Do u know qualified maintainers on this component?

external/kinesis-asl/src/main/scala/org/apache/spark/streaming/kinesis/KinesisRecordProcessor.scala
```diff
@@ -56,6 +56,27 @@ private[kinesis] class KinesisRecordProcessor[T](receiver: KinesisReceiver[T], w
...
+   * Limit the number of processed records from Kinesis stream. This is because the KCL cannot
+   * control the number of aggregated records to be fetched even if we set `MaxRecords`
+   * in `KinesisClientLibConfiguration`. For example, if we set 10 to the number of max records
+   * in a worker and a producer aggregates two records into one message, the worker possibly
+   * 20 records every callback function called.
+   */
+  private def processRecordsWithLimit(
+      batch: List[Record], checkpointer: IRecordProcessorCheckpointer): Unit = {
+    val maxRecords = receiver.getCurrentLimit
+    for (start <- 0 until batch.size by maxRecords) {
```
srowen: Hm, it just occurred to me that you would have a problem here if batch.size and maxRecords were both over Int.MaxValue / 2, and maxRecords were a bit smaller than batch.size. The addition below overflows.

It seems like a corner case but I note above you already defensively capped the maxRecords at Int.MaxValue so maybe it's less unlikely than it sounds.

You can fix it by letting the addition and min comparison take place over longs and then convert back to int.

Alternatively I think this is even simpler in Scala, though I imagine there's some extra overhead here:

```
batch.grouped(maxRecords).foreach(batch => addRecords(batch, checkpointer))
```

I don't know of a good reviewer for this component but I think I'm comfortable merging a straightforward change like this.

external/kinesis-asl/src/main/scala/org/apache/spark/streaming/kinesis/KinesisRecordProcessor.scala
```diff
@@ -56,6 +56,27 @@ private[kinesis] class KinesisRecordProcessor[T](receiver: KinesisReceiver[T], w
     logInfo(s"Initialized workerId $workerId with shardId $shardId")
   }
 
+  private def addRecords(batch: List[Record], checkpointer: IRecordProcessorCheckpointer): Unit = {
+    receiver.addRecords(shardId, batch)
+    logDebug(s"Stored: Worker $workerId stored ${batch.size} records for shardId $shardId")
+    receiver.setCheckpointer(shardId, checkpointer)
```
srowen: BTW is this supposed to be called on every batch or once at the end? I don't know how it works.

external/kinesis-asl/src/main/scala/org/apache/spark/streaming/kinesis/KinesisRecordProcessor.scala
maropu: Actually, since each kinesis shard has strict read limits of throughput (http://docs.aws.amazon.com/streams/latest/dev/service-sizes-and-limits.html), `batch.size` hardly exceeds `Int.MaxValue / 2`. But, since I like your idea in terms of code clearness, I fixed.

maropu:
Jenkins, retest this please.

srowen:
@brkyvz maybe you can give this a look to make sure it makes sense? especially the bit about the checkpointer.

external/kinesis-asl/src/main/scala/org/apache/spark/streaming/kinesis/KinesisRecordProcessor.scala
```diff
@@ -68,9 +69,16 @@ private[kinesis] class KinesisRecordProcessor[T](receiver: KinesisReceiver[T], w
...
       try {
-        receiver.addRecords(shardId, batch)
-        logDebug(s"Stored: Worker $workerId stored ${batch.size} records for shardId $shardId")
-        receiver.setCheckpointer(shardId, checkpointer)
+        // Limit the number of processed records from Kinesis stream. This is because the KCL cannot
+        // control the number of aggregated records to be fetched even if we set `MaxRecords`
+        // in `KinesisClientLibConfiguration`. For example, if we set 10 to the number of max
+        // records in a worker and a producer aggregates two records into one message, the worker
+        // possibly 20 records every callback function called.
+        batch.asScala.grouped(receiver.getCurrentLimit).foreach { batch =>
```
srowen: Sorry, one last comment -- `batch` is used for the overall data set and each subset. They should be named differently for clarity.

It's also my fault for not realizing the collections here were Java not Scala, and you have to convert to use the nice Scala idiom. I think it's OK as it's just going to wrap and not copy the class, but it does bear being careful about performance here.

external/kinesis-asl/src/main/scala/org/apache/spark/streaming/kinesis/KinesisRecordProcessor.scala
```diff
@@ -68,9 +69,16 @@ private[kinesis] class KinesisRecordProcessor[T](receiver: KinesisReceiver[T], w
...
+          receiver.addRecords(shardId, batch.asJava)
+          logDebug(s"Stored: Worker $workerId stored ${batch.size} records for shardId $shardId")
+          receiver.setCheckpointer(shardId, checkpointer)
```
brkyvz: this should be outside, after the `foreach`

srowen: Yeah, that's what I suspected at https://github.com/apache/spark/pull/16114#discussion_r90756702 -- thanks for confirming

maropu: thanks, I'll fix

external/kinesis-asl/src/main/scala/org/apache/spark/streaming/kinesis/KinesisRecordProcessor.scala
maropu: yea, I also think, when `maxRecords` is small and `batch` is large, many iterations cause a little overheads. So, I restored the code to the previous java-style one.

external/kinesis-asl/src/main/scala/org/apache/spark/streaming/kinesis/KinesisRecordProcessor.scala
```diff
@@ -68,7 +68,16 @@ private[kinesis] class KinesisRecordProcessor[T](receiver: KinesisReceiver[T], w
...
+        // control the number of aggregated records to be fetched even if we set `MaxRecords`
+        // in `KinesisClientLibConfiguration`. For example, if we set 10 to the number of max
+        // records in a worker and a producer aggregates two records into one message, the worker
+        // possibly 20 records every callback function called.
+        val maxRecords = receiver.getCurrentLimit
+        for (start <- 0 until batch.size by maxRecords) {
+          val miniBatch = batch.subList(start, math.min(start + maxRecords, batch.size))
+          receiver.addRecords(shardId, miniBatch)
+        }
         logDebug(s"Stored: Worker $workerId stored ${batch.size} records for shardId $shardId")
```
brkyvz: I would leave this comment inside the `for` loop, because IIRC `addRecords` will be a blocking call where it needs to be written to the WAL

brkyvz:
I will need to take a deeper look at this to remember the code. I'm not sure but there may be some issues with the checkpointing happening to the WriteAheadLog and DynamoDB. Going to come back to this in a couple hours.

external/kinesis-asl/src/main/scala/org/apache/spark/streaming/kinesis/KinesisRecordProcessor.scala
maropu: okay

maropu:
@brkyvz Could you also check this pr #16213? Thanks!

external/kinesis-asl/src/main/scala/org/apache/spark/streaming/kinesis/KinesisReceiver.scala
brkyvz: I would be okay to keep it if we add a useful error in case this assertion doesn't hold, e.g.
`assert(blockGenerator != null, "Expected blockGenerator to be set for the receiver before the processor received records")`

or something like that

brkyvz:
I've taken a look at the code. This change seems safe to me. Even if we process extra data, but fail to checkpoint to Kinesis, Spark streaming will re-process the exact same batch on a restart providing at least once semantics but with a stronger guarantee (that the data will be processed with exactly the same batching).

LGTM! Thanks @maropu

srowen:
Merged to master

PR #16119, opened 2016-12-02
[SPARK-18687][Pyspark][SQL]Backward compatibility - creating a Dataframe on a new SQLContext object fails with a Derby error

vijoshi:

Change is for SQLContext to reuse the active SparkSession during construction if the sparkContext supplied is the same as the currently active SparkContext. Without this change, a new SparkSession is instantiated that results in a Derby error when attempting to create a dataframe using a new SQLContext object even though the SparkContext supplied to the new SQLContext is same as the currently active one. Refer https://issues.apache.org/jira/browse/SPARK-18687 for details on the error and a repro.

## How was this patch tested?

Existing unit tests and a new unit test added to pyspark-sql:

/python/run-tests --python-executables=python --modules=pyspark-sql


vijoshi:
ping @holdenk @srowen

holdenk:
Since the current tests pass without this change I'd say that we should add a test for the behaviour we are planning to support that isn't currently supported (would also make the purpose of the change a bit clearer).

vijoshi:
@holdenk test case added

vijoshi:
@holdenk reminder 
also pinging @davies

python/pyspark/sql/context.py
```diff
@@ -72,8 +72,13 @@ def __init__(self, sparkContext, sparkSession=None, jsqlContext=None):
         self._sc = sparkContext
         self._jsc = self._sc._jsc
         self._jvm = self._sc._jvm
+
         if sparkSession is None:
-            sparkSession = SparkSession(sparkContext)
+            if sparkContext is SparkContext._active_spark_context:
+                sparkSession = SparkSession.builder.getOrCreate()
```
srowen: I don't know enough to usefully review, but is the better fix to call the getOrCreate that lets you pass a context and use that for both cases? I don't know if it is the same behavior but seems more natural to use the logic in SparkSession for this.

holdenk: Thats a good question, lets see what @andrewor14 thinks of changing the builder interface for `SparkSession`

srowen: @viirya since you opened a similar change, what do you think of this?

viirya: So you can see in the pr, I simply do `sparkSession = SparkSession.builder.getOrCreate()` here.

python/pyspark/sql/tests.py
```diff
@@ -1851,6 +1851,71 @@ def test_hivecontext(self):
         self.assertIn("default", out.decode('utf-8'))
         self.assertTrue(os.path.exists(metastore_path))
 
+    def test_hivecontext_create_dataframe(self):
```
cloud-fan: can we write a more low-level unit test to check if we create 2 `SparkSession`s?

python/pyspark/sql/context.py
```diff
@@ -72,8 +72,13 @@ def __init__(self, sparkContext, sparkSession=None, jsqlContext=None):
...
+            else:
```
cloud-fan: when will hit the else branch? The if-else seems the only difference between this PR and https://github.com/apache/spark/pull/16369

python/pyspark/sql/context.py
vijoshi: 1. the `SQLContext __init__` suggested that any sparkSession object could be injected. And the `SparkSession __init__` suggested any sparkContext could be injected. If not so - I thought they would all be invoking `getOrCreate()` to get instances of each other.  

2.  I found a few hits around this property - `spark.driver.allowMultipleContexts` which though not a python property - showed that there may be cases where someone (for testing etc) wanted the ability to bypass the one sparkContext limitation. 

These two combined I was not 100% sure someone couldn't send in different sparkContext objects / sparkSession objects tied to different sparkContexts when constructing an SQLContext. 

I will admit I am not familiar with Python side of things, so I hoped this PR review would be the best way to get this clarified :). 

If everyone agrees that we needn't worry about what sparkSession is passed in when constructing SQLContext, then I can get rid of the `if sparkContext is SparkContext._active_spark_context:` check.

python/pyspark/sql/context.py
vijoshi: @cloud-fan refer my comments above.

python/pyspark/sql/context.py
viirya: I found a relate discussion about multiple `SparkContext` at http://stackoverflow.com/questions/28259756/how-to-create-multiple-sparkcontexts-in-a-console

I think the usage of multiple contexts is discouraged, even you could do it in Scala/Java. Specially, you can't have multiple `SparkContext` in Python. So I am not worrying that another `SparkContext` other than the active one is passed in when constructing `SQLContext`.

python/pyspark/sql/tests.py
vijoshi: did you mean have the same scenario where multiple SQLContext objects are constructed but the test success criteria is to verify that no more than a single `SparkSession` object ever gets created?

viirya: I think we can just create two `SQLContext` and compare if they share the same `SparkSession` instance.

cloud-fan:
@vijoshi do you mind updating your PR according to the dicussion? i.e. simplify the fix and test

vijoshi:
@cloud-fan sure, i'll push an update soon

vijoshi:
@cloud-fan @viirya - rebased and updated the code change per review comments.

vijoshi:
retest this please

viirya:
The test failure is caused by another issue. I've submitted another PR to fix it: #16454. Once that is fixed, this test can be passed.

vijoshi:
thanks @viirya. will wait for your PR to go through.

cloud-fan:
https://github.com/apache/spark/pull/16454 has been merged

viirya:
@vijoshi Can you solve the conflict?

cloud-fan:
thanks, merging to master/2.1/2.0!

PR #14248, opened 2016-07-18
[SPARK-16589][PYTHON] Chained cartesian produces incorrect number of records

zero323:
## What changes were proposed in this pull request?

If either self or other is serialized using `CartesianSerializer` it is reserialized with default serializer before executing `_jrdd.cartesian`.
## How was this patch tested?

Using existing unit tests as well as additional test case to address  SPARK-16589.

PR #16121, opened 2016-12-02
[SPARK-16589][PYTHON] Chained cartesian produces incorrect number of records

aray:

Fixes a bug in the python implementation of rdd cartesian product related to batching that showed up in repeated cartesian products with seemingly random results. The root cause being multiple iterators pulling from the same stream in the wrong order because of logic that ignored batching.

`CartesianDeserializer` and `PairDeserializer` were changed to implement `_load_stream_without_unbatching` and borrow the one line implementation of `load_stream` from `BatchedSerializer`. The default implementation of `_load_stream_without_unbatching` was changed to give consistent results (always an iterable) so that it could be used without additional checks.

`PairDeserializer` no longer extends `CartesianDeserializer` as it was not really proper. If wanted a new common super class could be added.

Both `CartesianDeserializer` and `PairDeserializer` now only extend `Serializer` (which has no `dump_stream` implementation) since they are only meant for *de*serialization.

## How was this patch tested?

Additional unit tests (sourced from #14248) plus one for testing a cartesian with zip.


davies:
It's pretty tricky to make the chained CartesianDeserializer work, maybe it's easier to have a workaround in the RDD.cartesian() to add an _reserialize() between chained cartesian (or zipped), it will be less performant, but will be easy to reason about.

The current patch may still be wrong in case of chained DartesianDeserializer and PairSerializer, for example, a.cartesian(b.zip(c)) (have not verified yet)

aray:
@davies I was trying to make minimal changes to `PairDeserializer`, but you are right it needs changed also. I'll update the PR shortly.

zero323:
@davies I suggested workaround before but I remember that @holdenk had some reservations. 

Moreover it would have to be done proactively for all (?) calls. For example [SPARK-17756](https://issues.apache.org/jira/browse/SPARK-17756) seems to hit a similar problem.

holdenk:
I was hesistant with the previous PR since it seemed like we didn't fully understand why we were changing what we were at the time, I can try and take a closer look at this over the next few days if it is in a good place for that to happen.

aray:
@davies, @zero323, and @holdenk this is in a good place for review if you want to take a look.

python/pyspark/serializers.py
```diff
@@ -96,7 +96,7 @@ def load_stream(self, stream):
         raise NotImplementedError
 
     def _load_stream_without_unbatching(self, stream):
```
holdenk: Even though this is internal it might make sense to have a docstring for this since were changing its behaviour.

python/pyspark/serializers.py
```diff
@@ -278,50 +278,51 @@ def __repr__(self):
         return "AutoBatchedSerializer(%s)" % self.serializer
 
 
-class CartesianDeserializer(FramedSerializer):
+class CartesianDeserializer(Serializer):
 
     """
     Deserializes the JavaRDD cartesian() of two PythonRDDs.
```
holdenk: Maybe we should document this a bit given that we had problems with the implementation. (e.g. expand on the "Due to batching, we can't use the Java cartesian method." comment from `rdd.py` to explain how this is intended to function).

python/pyspark/serializers.py
```diff
@@ -278,50 +278,51 @@ def __repr__(self):
...
-        val_is_batched = isinstance(self.val_ser, BatchedSerializer)
-        for (keys, vals) in zip(key_stream, val_stream):
-            keys = keys if key_is_batched else [keys]
-            vals = vals if val_is_batched else [vals]
-            yield (keys, vals)
+    def _load_stream_without_unbatching(self, stream):
+        key_batch_stream = self.key_ser._load_stream_without_unbatching(stream)
+        val_batch_stream = self.val_ser._load_stream_without_unbatching(stream)
+        for (key_batch, val_batch) in zip(key_batch_stream, val_batch_stream):
+            yield product(key_batch, val_batch)
```
holdenk: Maybe consider adding a comment here explaining why the interaction of batching & product

holdenk (changes requested):
Thanks for working on this, just doing a quick first pass it looks like really good work - but I'd encourage you to add a few more comments in some places (since we had this bug before it seems like just the code wasn't sufficiently self explanatory). I'll do a deeper look later on this week.

davies:
LGTM, merging into master and 2.1 branch, thanks!

stuarteberg:
This PR seems to have introduced a bug, which I have reported here:
https://issues.apache.org/jira/browse/SPARK-21985

Any thoughts, @aray?  Can the check in question simply be removed, or is there a better solution to consider?

aray:
I'll take a look, sorry about that.

PR #16122, opened 2016-12-02
[SPARK-18681][SQL] Fix filtering to compatible with partition keys of type int

wangyum:

Cloudera put `/var/run/cloudera-scm-agent/process/15000-hive-HIVEMETASTORE/hive-site.xml` as the configuration file for the Hive Metastore Server, where `hive.metastore.try.direct.sql=false`. But Spark isn't reading this configuration file and get default value `hive.metastore.try.direct.sql=true`. As @mallman said, we should use `getMetaConf` method to obtain the original configuration from Hive Metastore Server. I have tested this method few times and the return value is always consistent with Hive Metastore Server.

## How was this patch tested?

The existing tests.

### DIFF
--- a/sql/hive/src/main/scala/org/apache/spark/sql/hive/client/HiveShim.scala
+++ b/sql/hive/src/main/scala/org/apache/spark/sql/hive/client/HiveShim.scala
@@ -591,7 +591,7 @@ private[client] class Shim_v0_13 extends Shim_v0_12 {
         logDebug(s"Hive metastore filter is '$filter'.")
         val tryDirectSqlConfVar = HiveConf.ConfVars.METASTORE_TRY_DIRECT_SQL
         // We should get this config value from the metaStore. otherwise hit SPARK-18681.
-        // To be compatible with hive-0.12 and hive-0.13,In the future we can achieve this by:
+        // To be compatible with hive-0.12 and hive-0.13, In the future we can achieve this by:
         // val tryDirectSql = hive.getMetaConf(tryDirectSqlConfVar.varname).toBoolean
         val tryDirectSql = hive.getMSC.getConfigValue(tryDirectSqlConfVar.varname,
           tryDirectSqlConfVar.defaultBoolVal.toString).toBoolean
--- a/sql/hive/src/main/scala/org/apache/spark/sql/hive/client/HiveShim.scala
+++ b/sql/hive/src/main/scala/org/apache/spark/sql/hive/client/HiveShim.scala
@@ -590,6 +590,7 @@ private[client] class Shim_v0_13 extends Shim_v0_12 {
       } else {
         logDebug(s"Hive metastore filter is '$filter'.")
         val tryDirectSqlConfVar = HiveConf.ConfVars.METASTORE_TRY_DIRECT_SQL
+        // We should get this config value from the metaStore. otherwise hit SPARK-18681.
         // To be compatible with hive-0.12 and hive-0.13,In the future we can achieve this by:
         // val tryDirectSql = hive.getMetaConf(tryDirectSqlConfVar.varname).toBoolean
         val tryDirectSql = hive.getMSC.getConfigValue(tryDirectSqlConfVar.varname,
--- a/sql/hive/src/main/scala/org/apache/spark/sql/hive/client/HiveShim.scala
+++ b/sql/hive/src/main/scala/org/apache/spark/sql/hive/client/HiveShim.scala
@@ -590,7 +590,7 @@ private[client] class Shim_v0_13 extends Shim_v0_12 {
       } else {
         logDebug(s"Hive metastore filter is '$filter'.")
         val tryDirectSqlConfVar = HiveConf.ConfVars.METASTORE_TRY_DIRECT_SQL
-        val tryDirectSql = hive.getMetaConf(tryDirectSqlConfVar.varname).toBoolean
+        val tryDirectSql = hive.getMSC.getMetaConf(tryDirectSqlConfVar.varname).toBoolean
         try {
           // Hive may throw an exception when calling this method in some circumstances, such as
           // when filtering on a non-string partition column when the hive config key
--- a/sql/hive/src/main/scala/org/apache/spark/sql/hive/client/HiveShim.scala
+++ b/sql/hive/src/main/scala/org/apache/spark/sql/hive/client/HiveShim.scala
@@ -590,8 +590,10 @@ private[client] class Shim_v0_13 extends Shim_v0_12 {
       } else {
         logDebug(s"Hive metastore filter is '$filter'.")
         val tryDirectSqlConfVar = HiveConf.ConfVars.METASTORE_TRY_DIRECT_SQL
+        // To be compatible with hive-0.12 and hive-0.13,In the future we can achieve this by:
+        // val tryDirectSql = hive.getMetaConf(tryDirectSqlConfVar.varname).toBoolean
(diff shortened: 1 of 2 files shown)


rxin:
cc @mallman @ericl

ericl:
Can you also update the warning message to match the new condition?

sql/hive/src/main/scala/org/apache/spark/sql/hive/client/HiveShim.scala
```diff
@@ -600,11 +600,14 @@ private[client] class Shim_v0_13 extends Shim_v0_12 {
             .asInstanceOf[JArrayList[Partition]]
         } catch {
           case ex: InvocationTargetException if ex.getCause.isInstanceOf[MetaException] &&
-              !tryDirectSql =>
-            logWarning("Caught Hive MetaException attempting to get partition metadata by " +
-              "filter from Hive. Falling back to fetching all partition metadata, which will " +
-              "degrade performance. Modifying your Hive metastore configuration to set " +
-              s"${tryDirectSqlConfVar.varname} to true may resolve this problem.", ex)
+              (!tryDirectSql || ex.getCause.getMessage.contains(
```
ericl: How about checking for the other flag instead of string matching.

mallman:
@wangyum Can you tell me what your underlying metastore database provider is? Postgres? MySQL?

sql/hive/src/main/scala/org/apache/spark/sql/hive/client/HiveShim.scala
mallman: In hindsight, I believe this approach of testing the hive configuration may be fundamentally flawed. When testing the value of `HiveConf.ConfVars.METASTORE_TRY_DIRECT_SQL`, we need to know the value set in the Hive Metastore server. What we're testing here is the setting in this Hive client. They won't necessarily be the same, so we can get false positives and false negatives of this test here.

mallman: The `Hive` class has a `getMetaConf` method. That sounds like it might return config values from the metastore server. I can take a little time to investigate.

wangyum:
@mallman  MySQL and version is 5.6.29

sql/hive/src/main/scala/org/apache/spark/sql/hive/client/HiveShim.scala
wangyum: Yes, you are right.
I found the root cause, I will update the pr description later.

wangyum:
This patch fails because hive-0.12 and hive-0.13 doesn't has `getMetaConf` method.
see [HIVE-7532](https://issues.apache.org/jira/browse/HIVE-7532),

sql/hive/src/main/scala/org/apache/spark/sql/hive/client/HiveShim.scala
```diff
@@ -590,8 +590,10 @@ private[client] class Shim_v0_13 extends Shim_v0_12 {
       } else {
         logDebug(s"Hive metastore filter is '$filter'.")
         val tryDirectSqlConfVar = HiveConf.ConfVars.METASTORE_TRY_DIRECT_SQL
-        val tryDirectSql =
-          hive.getConf.getBoolean(tryDirectSqlConfVar.varname, tryDirectSqlConfVar.defaultBoolVal)
+        // To be compatible with hive-0.12 and hive-0.13,In the future we can achieve this by:
```
mallman: Can you add a note briefly explaining why we need to get this config value from the metastore instead of from the local hive config?

mallman:
@wangyum Thanks for fixing this. The fact that our tests did not catch this bug means we have a gap in our test coverage. It looks like the test in `HiveClientSuite` is incorrect. Can you fix it? I'd really like to ensure we have a valid unit test for this behavior.

mallman:
@wangyum I'm going to see if I can help with the unit testing on this PR.

wangyum:
@mallman OK, thanks

mallman:
I haven't been able to get a proper unit test environment running where the embedded metastore conf is different from the client conf. I did validate that Spark without this patch failed to execute a query on a table with an integer type partition column filtering on that column where the metastore has direct sql access disabled, whereas Spark with this patch works and behaves as expected.

@wangyum I don't believe your test uses Hive in a way that's compatible with Spark. Can you please remove it?

@ericl Any ideas on how to unit test the case where the client and metastore have different configurations?

ericl:
@mallman could a test using getMSC.setMetaConf work?

mallman:
I tried that but got a `NoSuchMethodException` in the call to `getMSC`.

mallman:
There may be a way to do it, but the classloader tricks being used in the hive client implementation are beyond my comprehension.

ericl:
Hm, doesn't the setMetaConf call work in the removed test? https://github.com/apache/spark/pull/16122/commits/c0668f08dddd6d75d170d7a0ace30831ae81b0ba

mallman:
It does, yes. My concern around that test is that its behavior doesn't seem to be independent of other tests. For example, the value of

```hive.getConf.getBoolean(tryDirectSqlConfVar.varname, tryDirectSqlConfVar.defaultBoolVal)```

seemed to vary depending on what it was set to in the other test in that suite.

ericl:
I see. If we set both the client and server conf, and then reverted them in a finally block, would that be sufficient? Then you can test the client falls back correctly under various config scenarios.

mallman:
I'm not sure configuration-level rollback will guarantee an absence of interactions with other tests. For one thing, I think we need to create and clean up an independent metastore directory for Derby. Otherwise we may end up with https://amplab.cs.berkeley.edu/jenkins/job/SparkPullRequestBuilder/69669/testReport/junit/org.apache.spark.sql.hive/HiveExternalCatalogSuite/_It_is_not_a_test_/.

I guess I'd like a stronger assurance that we're doing this the right way. The only call to `Hive.get` in the Hive codebase I can find is in a private method in `HiveClientImpl`. Spark seems to be going to great lengths to discourage calling that method directly.

I've also thought this might be better suited to testing in the thriftserver codebase, where there seems to be a stronger sense of client/server separation in testing. I haven't gone too far in investigating that direction.

ericl:
I did some digging into HiveClientImpl and Hive.java, and I think it would be pretty safe to call `Hive.get()` within the scope of a `client.withHiveState {  }`. That call sets the thread-local hive to that used by the client before executing the passed function block.

mallman:
I think that's exactly what I tried and got the `NoSuchMethodException`.

On Dec 7, 2016, at 3:35 PM, Eric Liang <notifications@github.com> wrote:

I did some digging into HiveClientImpl and Hive.java, and I think it would
be pretty safe to call Hive.get() within the scope of a client.withHiveState
{ }. That call sets the thread-local hive to that used by the client before
executing the passed function block.

—
<https://github.com/apache/spark/pull/16122#issuecomment-265608215>, or mute
the thread

ericl:
I see. In that case I think manual testing may be sufficient.

mallman:
I believe that may be the case, unfortunately. At least, I have no immediate ideas otherwise.

hvanhovell:
I am going to merge this after a successful build. Ping me if anyone objects.

hvanhovell:
retest this please

hvanhovell:
LGTM, merging to master/2.1. Thanks!

mallman:
@wangyum Thank you for this important bug fix! :)

PR #16128, opened 2016-12-03
[SPARK-18671][SS][TEST] Added tests to ensure stability of that all Structured Streaming log formats

tdas:

To be able to restart StreamingQueries across Spark version, we have already made the logs (offset log, file source log, file sink log) use json. We should added tests with actual json files in the Spark such that any incompatible changes in reading the logs is immediately caught. This PR add tests for FileStreamSourceLog, FileStreamSinkLog, and OffsetSeqLog.

## How was this patch tested?
new unit tests


sql/core/src/test/scala/org/apache/spark/sql/execution/streaming/OffsetSeqLogSuite.scala
```diff
@@ -60,4 +60,39 @@ class OffsetSeqLogSuite extends SparkFunSuite with SharedSQLContext {
...
   }
+
+  test("read Spark 2.1.0 log format") {
+    val (batchId, offsetSeq) = readFromResource("offset-log-version-2.1.0")
+    assert(batchId === 16)
+    assert(offsetSeq ===
+      OffsetSeq(
+        offsets = Seq(Some(SerializedOffset(
+          """
+            |{"kafka-topic":{"23":0,"8":1,"17":1,"11":1,"20":0,"2":6,"5":2,"14":0,"4":4,"13":1,
```
tdas: This needs to be revised.

zsxwing: We should add a test for Kafka source to check if we can convert json to KafkaSourceOffset. This doesn't check that.

external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/JsonUtils.scala
```diff
@@ -81,7 +81,14 @@ private object JsonUtils {
...
   def partitionOffsets(partitionOffsets: Map[TopicPartition, Long]): String = {
     val result = new HashMap[String, HashMap[Int, Long]]()
-    partitionOffsets.foreach { case (tp, off) =>
+    implicit val ordering = new Ordering[TopicPartition] {
+      override def compare(x: TopicPartition, y: TopicPartition): Int = {
+        Ordering.Tuple2[String, Int].compare((x.topic, x.partition), (y.topic, y.partition))
+      }
+    }
+    val partitions = partitionOffsets.keySet.toSeq.sorted  // sort for more determinism
+    partitions.foreach { tp =>
```
zsxwing: nit: You can use `partitionOffsets.toSeq.sortBy(_._1).foreach { case (tp, off) =>` to simplify the codes.

external/kafka-0-10-sql/src/test/scala/org/apache/spark/sql/kafka010/KafkaSourceOffsetSuite.scala
```diff
@@ -89,4 +89,16 @@ class KafkaSourceOffsetSuite extends OffsetSuite with SharedSQLContext {
         Array(0 -> batch0Serialized, 1 -> batch1Serialized))
     }
   }
+
+  test("read Spark 2.1.0 log format") {
+    val offset = readFromResource("kafka-source-offset-version-2.1.0.txt")
```
zsxwing: nit: maybe not need to read json from a file since we never write them into a single file.

sql/core/src/test/scala/org/apache/spark/sql/streaming/FileStreamSourceSuite.scala
```diff
@@ -1022,6 +1021,33 @@ class FileStreamSourceSuite extends FileStreamSourceTest {
     val options = new FileStreamOptions(Map("maxfilespertrigger" -> "1"))
     assert(options.maxFilesPerTrigger == Some(1))
   }
+
+  test("FileStreamSource offset - read Spark 2.1.0 log format") {
+    val offset = readOffsetFromResource("file-source-offset-version-2.1.0.txt")
```
zsxwing: nit: maybe not need to read json from a file since we never write them into a single file.

zsxwing (commented):
LGTM overall. Just some nits.

external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/JsonUtils.scala
tdas: I want to sort by topic and partitions together. so that partitions are ordered when json is generated (currently is not) and hard to read.

external/kafka-0-10-sql/src/test/scala/org/apache/spark/sql/kafka010/KafkaSourceOffsetSuite.scala
tdas: yeah. but its good to have it in a separate file in the same place as other formats. will be easier to track all the things that need compatibility guarantees.

sql/core/src/test/scala/org/apache/spark/sql/streaming/FileStreamSourceSuite.scala
tdas: same comment as above.

zsxwing:
LGTM pending tests.

zsxwing:
retest this please

zsxwing:
LGTM. Thanks. Merging to master and 2.1.

PR #16131, opened 2016-12-04
[SPARK-18701][ML] Fix Poisson GLM failure due to wrong initialization

actuaryzhang:
Poisson GLM fails for many standard data sets (see example in test or JIRA). The issue is incorrect initialization leading to almost zero probability and weights. Specifically, the mean is initialized as the response, which could be zero. Applying the log link results in very negative numbers (protected against -Inf), which again leads to close to zero probability and weights in the weighted least squares. Fix and test are included in the commits.

Update initialization in Poisson GLM

## How was this patch tested?
Add test in GeneralizedLinearRegressionSuite

@srowen @sethah @yanboliang @HyukjinKwon @mengxr

### DIFF
--- a/mllib/src/main/scala/org/apache/spark/ml/regression/GeneralizedLinearRegression.scala
+++ b/mllib/src/main/scala/org/apache/spark/ml/regression/GeneralizedLinearRegression.scala
@@ -505,7 +505,8 @@ object GeneralizedLinearRegression extends DefaultParamsReadable[GeneralizedLine
     override def initialize(y: Double, weight: Double): Double = {
       require(y >= 0.0, "The response variable of Poisson family " +
         s"should be non-negative, but got $y")
-      y + 0.1
+      // Set lower bound for mean in the FIRST step in IWLS
+      math.max(y, 0.1)
     }
 
     override def variance(mu: Double): Double = mu
--- a/mllib/src/main/scala/org/apache/spark/ml/regression/GeneralizedLinearRegression.scala
+++ b/mllib/src/main/scala/org/apache/spark/ml/regression/GeneralizedLinearRegression.scala
@@ -505,7 +505,7 @@ object GeneralizedLinearRegression extends DefaultParamsReadable[GeneralizedLine
     override def initialize(y: Double, weight: Double): Double = {
       require(y >= 0.0, "The response variable of Poisson family " +
         s"should be non-negative, but got $y")
-      // Set lower bound for mean in the FIRST step in IWLS
+      // force Poisson mean > 0 to avoid numerical instability in IRLS
       math.max(y, 0.1)
     }
 
--- a/mllib/src/main/scala/org/apache/spark/ml/regression/GeneralizedLinearRegression.scala
+++ b/mllib/src/main/scala/org/apache/spark/ml/regression/GeneralizedLinearRegression.scala
@@ -505,7 +505,7 @@ object GeneralizedLinearRegression extends DefaultParamsReadable[GeneralizedLine
     override def initialize(y: Double, weight: Double): Double = {
       require(y >= 0.0, "The response variable of Poisson family " +
         s"should be non-negative, but got $y")
-      y
+      y + 0.1
     }
 
     override def variance(mu: Double): Double = mu
--- a/mllib/src/main/scala/org/apache/spark/ml/regression/GeneralizedLinearRegression.scala
+++ b/mllib/src/main/scala/org/apache/spark/ml/regression/GeneralizedLinearRegression.scala
@@ -505,7 +505,10 @@ object GeneralizedLinearRegression extends DefaultParamsReadable[GeneralizedLine
     override def initialize(y: Double, weight: Double): Double = {
       require(y >= 0.0, "The response variable of Poisson family " +
         s"should be non-negative, but got $y")
-      // force Poisson mean > 0 to avoid numerical instability in IRLS
+      /*
+        Force Poisson mean > 0 to avoid numerical instability in IRLS.
+        R uses y + 0.1 for initialization. See poisson()$initialize.
+       */
       math.max(y, 0.1)
     }
 
--- a/mllib/src/test/scala/org/apache/spark/ml/regression/GeneralizedLinearRegressionSuite.scala
+++ b/mllib/src/test/scala/org/apache/spark/ml/regression/GeneralizedLinearRegressionSuite.scala
@@ -89,11 +89,23 @@ class GeneralizedLinearRegressionSuite
       xVariance = Array(0.7, 1.2), nPoints = 10000, seed, noiseLevel = 0.01,
       family = "poisson", link = "log").toDF()
 
-    datasetPoissonLogWithZero = generateGeneralizedLinearRegressionInput(
-      intercept = -1.5, coefficients = Array(0.22, 0.06), xMean = Array(2.9, 10.5),
-      xVariance = Array(0.7, 1.2), nPoints = 100, seed, noiseLevel = 0.01,
-      family = "poisson", link = "log")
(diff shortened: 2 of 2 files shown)


actuaryzhang:
Jenkins, add to whitelist

mllib/src/main/scala/org/apache/spark/ml/regression/GeneralizedLinearRegression.scala
```diff
@@ -505,7 +505,7 @@ object GeneralizedLinearRegression extends DefaultParamsReadable[GeneralizedLine
     override def initialize(y: Double, weight: Double): Double = {
       require(y >= 0.0, "The response variable of Poisson family " +
         s"should be non-negative, but got $y")
-      y
+      y + 0.1
```
srowen: You're saying that the initial value of the response variable could be 0 because it could be a mean over all 0 values? yes, maybe. Why 0.1? shouldn't this just be `math.max(y, EPSILON)` for a small epsilon?

actuaryzhang:
@srowen 
Try this example below or the example @sethah had issue with in #15683. 

I have tried running the 2.1 version Poisson GLM on our data and it fails for most of them (it does work when there are not lots of zero sometimes). I traced down the cause and the fix proposed here seems to be the correct. At least the Poisson GLM is working on the data where it failed before.  

```
val datasetPoissonLogWithZero = Seq(
      LabeledPoint(0.0, Vectors.dense(18, 1.0)),
      LabeledPoint(1.0, Vectors.dense(12, 0.0)),
      LabeledPoint(0.0, Vectors.dense(15, 0.0)),
      LabeledPoint(0.0, Vectors.dense(13, 2.0)),
      LabeledPoint(0.0, Vectors.dense(15, 1.0)),
      LabeledPoint(1.0, Vectors.dense(16, 1.0)),
      LabeledPoint(0.0, Vectors.dense(10, 0.0)),
      LabeledPoint(0.0, Vectors.dense(15, 0.0)),
      LabeledPoint(0.0, Vectors.dense(12, 2.0)),
      LabeledPoint(0.0, Vectors.dense(13, 0.0)),
      LabeledPoint(1.0, Vectors.dense(15, 0.0)),
      LabeledPoint(1.0, Vectors.dense(15, 0.0)),
      LabeledPoint(0.0, Vectors.dense(15, 0.0)),
      LabeledPoint(0.0, Vectors.dense(12, 2.0)),
      LabeledPoint(1.0, Vectors.dense(12, 2.0))
    ).toDF()
    
val glr = new GeneralizedLinearRegression()
  .setFamily("poisson")
  .setLink("log")
  .setMaxIter(20)
  .setRegParam(0)

val model = glr.fit(datasetPoissonLogWithZero)
```

mllib/src/main/scala/org/apache/spark/ml/regression/GeneralizedLinearRegression.scala
yanboliang: @srowen Actually we have protected value against ```0.0``` by:
```
override def project(mu: Double): Double = {
      if (mu < epsilon) {
        epsilon
      } else if (mu.isInfinity) {
        Double.MaxValue
      } else {
        mu
      }
    }
```
but it seems that ```epsilon``` is not enough, since the function curve is very steep near zero. I'm ok for this change, and matching R make sense.

yanboliang (approved):

mllib/src/main/scala/org/apache/spark/ml/regression/GeneralizedLinearRegression.scala
actuaryzhang: @yanboliang Thanks much for clarifying and approving this. 

@sethah That's exactly the issue. Using avg(y) for Binomial would be more costly than the current approach since it takes one more pass through the data, right?

srowen: This is still an argument for making tiny values at least some epsilon (here 0.1) right? why does 30 need to become 30.1? isn't that just introducing some small inaccuracy needlessly?

actuaryzhang: @srowen This is not adjusting **the data** (y). Rather, the adjustment is for **the mean** E(y) in the very **first** step of the iteration. When you observe y = 30, it probably does not  matter whether E(y) = 30 or E(y) = 30.1 because the posterior will be very close? But when you observe y = 0, it is certainly not E(y) = 0, because Poisson must have positive mean. The mean E(y) in this case is probably a small number, say 0.1. One may set it to be 1e-16, but this will cause numerical issues in the subsequent iterations. Do you prefer just adding 0.1 to y = 0?

srowen: c) seems most consistent with the intent. Unless I'm totally missing something that seems preferable (with a comment to explain what it's for)

actuaryzhang:
@srowen @yanboliang 
I have updated the code and further cleaned up the test. Please review and let me know if there is any question. Thanks.

mllib/src/test/scala/org/apache/spark/ml/regression/GeneralizedLinearRegressionSuite.scala
```diff
@@ -497,6 +500,7 @@ class GeneralizedLinearRegressionSuite
         .setFitIntercept(fitIntercept).setLinkPredictionCol("linkPrediction")
       val model = trainer.fit(dataset)
       val actual = Vectors.dense(model.intercept, model.coefficients(0), model.coefficients(1))
+      println("coeff is " + actual)
```
sethah: remove :)

mllib/src/main/scala/org/apache/spark/ml/regression/GeneralizedLinearRegression.scala
```diff
@@ -505,7 +505,8 @@ object GeneralizedLinearRegression extends DefaultParamsReadable[GeneralizedLine
     override def initialize(y: Double, weight: Double): Double = {
       require(y >= 0.0, "The response variable of Poisson family " +
         s"should be non-negative, but got $y")
-      y
+      // Set lower bound for mean in the FIRST step in IWLS
```
sethah: let's change this to `// force Poisson mean > 0 to avoid numerical instability in IRLS`

sethah (commented):
I don't have a problem with adding 0.1 in all cases since this is just an initialization step, but it'd still be nice to have a reference to point to. Is there anything that discusses initializations for different families in IRLS?

actuaryzhang:
@sethah Thanks for the review. I have updated according to your suggestion. 

@yanboliang @srowen  Please take another look. Thanks.

mllib/src/main/scala/org/apache/spark/ml/regression/GeneralizedLinearRegression.scala
```diff
@@ -505,7 +505,8 @@ object GeneralizedLinearRegression extends DefaultParamsReadable[GeneralizedLine
     override def initialize(y: Double, weight: Double): Double = {
       require(y >= 0.0, "The response variable of Poisson family " +
         s"should be non-negative, but got $y")
-      y
+      // force Poisson mean > 0 to avoid numerical instability in IRLS
```
srowen: It wouldn't hurt to give a reference to the R source code here if you can, even just pointing out that this is (sort of) how R deals with it, as justification. It is a surprisingly big epsilon after all.

actuaryzhang:
@srowen Done. Thanks for the suggestion.

actuaryzhang:
@srowen Is this ready to be merged?

sethah:
LGTM

srowen:
Merged to master, and to 2.1 to match SPARK-18166

PR #16135, opened 2016-12-04
[SPARK-18700][SQL] Add StripedLock for each table's relation in cache

xuanyuanking:

As the scenario describe in [SPARK-18700](https://issues.apache.org/jira/browse/SPARK-18700), when cachedDataSourceTables invalided, the coming few queries will fetch all FileStatus in listLeafFiles function. In the condition of table has many partitions, these jobs will occupy much memory of driver finally may cause driver OOM.

In this patch, add StripedLock for each table's relation in cache not for the whole cachedDataSourceTables, each table's load cache operation protected by it.

## How was this patch tested?

Add a multi-thread access table test in `PartitionedTablePerfStatsSuite` and check it only loading once using metrics in `HiveCatalogMetrics`

### DIFF
--- a/core/src/main/scala/org/apache/spark/metrics/source/StaticSources.scala
+++ b/core/src/main/scala/org/apache/spark/metrics/source/StaticSources.scala
@@ -97,6 +97,12 @@ object HiveCatalogMetrics extends Source {
     MetricRegistry.name("parallelListingJobCount"))
 
   /**
+   * Tracks the total number of cachedDataSourceTables hits.
+   */
+  val METRIC_DATASOUCE_TABLE_CACHE_HITS = metricRegistry.counter(
+    MetricRegistry.name("dataSourceTableCacheHits"))
+
+  /**
    * Resets the values of all metrics to zero. This is useful in tests.
    */
   def reset(): Unit = {
@@ -105,6 +111,7 @@ object HiveCatalogMetrics extends Source {
     METRIC_FILE_CACHE_HITS.dec(METRIC_FILE_CACHE_HITS.getCount())
     METRIC_HIVE_CLIENT_CALLS.dec(METRIC_HIVE_CLIENT_CALLS.getCount())
     METRIC_PARALLEL_LISTING_JOB_COUNT.dec(METRIC_PARALLEL_LISTING_JOB_COUNT.getCount())
+    METRIC_DATASOUCE_TABLE_CACHE_HITS.dec(METRIC_DATASOUCE_TABLE_CACHE_HITS.getCount())
   }
 
   // clients can use these to avoid classloader issues with the codahale classes
@@ -113,4 +120,5 @@ object HiveCatalogMetrics extends Source {
   def incrementFileCacheHits(n: Int): Unit = METRIC_FILE_CACHE_HITS.inc(n)
   def incrementHiveClientCalls(n: Int): Unit = METRIC_HIVE_CLIENT_CALLS.inc(n)
   def incrementParallelListingJobCount(n: Int): Unit = METRIC_PARALLEL_LISTING_JOB_COUNT.inc(n)
+  def incrementDataSourceTableCacheHits(n: Int): Unit = METRIC_DATASOUCE_TABLE_CACHE_HITS.inc(n)
 }
--- a/core/src/main/scala/org/apache/spark/metrics/source/StaticSources.scala
+++ b/core/src/main/scala/org/apache/spark/metrics/source/StaticSources.scala
@@ -97,12 +97,6 @@ object HiveCatalogMetrics extends Source {
     MetricRegistry.name("parallelListingJobCount"))
 
   /**
-   * Tracks the total number of cachedDataSourceTables hits.
-   */
-  val METRIC_DATASOUCE_TABLE_CACHE_HITS = metricRegistry.counter(
-    MetricRegistry.name("dataSourceTableCacheHits"))
-
-  /**
    * Resets the values of all metrics to zero. This is useful in tests.
    */
   def reset(): Unit = {
@@ -111,7 +105,6 @@ object HiveCatalogMetrics extends Source {
     METRIC_FILE_CACHE_HITS.dec(METRIC_FILE_CACHE_HITS.getCount())
     METRIC_HIVE_CLIENT_CALLS.dec(METRIC_HIVE_CLIENT_CALLS.getCount())
     METRIC_PARALLEL_LISTING_JOB_COUNT.dec(METRIC_PARALLEL_LISTING_JOB_COUNT.getCount())
-    METRIC_DATASOUCE_TABLE_CACHE_HITS.dec(METRIC_DATASOUCE_TABLE_CACHE_HITS.getCount())
   }
 
   // clients can use these to avoid classloader issues with the codahale classes
@@ -120,5 +113,4 @@ object HiveCatalogMetrics extends Source {
   def incrementFileCacheHits(n: Int): Unit = METRIC_FILE_CACHE_HITS.inc(n)
   def incrementHiveClientCalls(n: Int): Unit = METRIC_HIVE_CLIENT_CALLS.inc(n)
   def incrementParallelListingJobCount(n: Int): Unit = METRIC_PARALLEL_LISTING_JOB_COUNT.inc(n)
-  def incrementDataSourceTableCacheHits(n: Int): Unit = METRIC_DATASOUCE_TABLE_CACHE_HITS.inc(n)
 }
--- a/sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveMetastoreCatalog.scala
+++ b/sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveMetastoreCatalog.scala
@@ -53,7 +53,7 @@ private[hive] class HiveMetastoreCatalog(sparkSession: SparkSession) extends Log
       tableIdent.table.toLowerCase)
   }
 
-  /** Locks for preventing driver mem waste when concurrent table instantiation */
+  /** These locks guard against multiple attempts to instantiate a table, which wastes memory. */
   private val tableCreationLocks = Striped.lazyWeakLock(100)
 
   /** Acquires a lock on the table cache for the duration of `f`. */
@@ -221,9 +221,6 @@ private[hive] class HiveMetastoreCatalog(sparkSession: SparkSession) extends Log
         }
       }
 
-      // Here we should protect all relation get and create operation with lock while big
-      // table's CatalogFileIndex will take some time, only lock cachedDataSourceTables.put
-      // will still cause driver memory waste. More detail see SPARK-18700.
       withTableCreationLock(tableIdentifier, {
         val cached = getCached(
           tableIdentifier,
--- a/sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveMetastoreCatalog.scala
+++ b/sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveMetastoreCatalog.scala
@@ -34,8 +34,6 @@ import org.apache.spark.sql.execution.datasources.parquet.{ParquetFileFormat, Pa
 import org.apache.spark.sql.hive.orc.OrcFileFormat
 import org.apache.spark.sql.types._
 
-
-
 /**
  * Legacy catalog for interacting with the Hive metastore.
  *
@@ -56,12 +54,12 @@ private[hive] class HiveMetastoreCatalog(sparkSession: SparkSession) extends Log
       tableIdent.table.toLowerCase)
   }
 
-  /** ReadWriteLock for each tables, protect the read and write cached */
-  private val tableLockStripes = Striped.lazyWeakLock(10)
+  /** Locks for preventing driver mem waste when concurrent table instantiation */
+  private val tableCreationLocks = Striped.lazyWeakLock(100)
 
   /** Acquires a lock on the table cache for the duration of `f`. */
-  private def cacheLock[A](tableName: QualifiedTableName, f: => A): A = {
-    val lock = tableLockStripes.get(tableName)
+  private def withTableCreationLock[A](tableName: QualifiedTableName, f: => A): A = {
+    val lock = tableCreationLocks.get(tableName)
     lock.lock()
     try f finally {
       lock.unlock()
@@ -225,10 +223,10 @@ private[hive] class HiveMetastoreCatalog(sparkSession: SparkSession) extends Log
         }
       }
 
-      // Here we should protect all relation get and create operation with writeLock while
-      // big table's CatalogFileIndex will take some time, only lock cachedDataSourceTables.put
+      // Here we should protect all relation get and create operation with lock while big
+      // table's CatalogFileIndex will take some time, only lock cachedDataSourceTables.put
       // will still cause driver memory waste. More detail see SPARK-18700.
-      cacheLock(tableIdentifier, {
+      withTableCreationLock(tableIdentifier, {
         val cached = getCached(
           tableIdentifier,
           rootPaths,
@@ -272,7 +270,7 @@ private[hive] class HiveMetastoreCatalog(sparkSession: SparkSession) extends Log
       })
     } else {
       val rootPath = metastoreRelation.hiveQlTable.getDataLocation
-      cacheLock(tableIdentifier, {
+      withTableCreationLock(tableIdentifier, {
         val cached = getCached(tableIdentifier,
           Seq(rootPath),
           metastoreRelation,
--- a/sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveMetastoreCatalog.scala
+++ b/sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveMetastoreCatalog.scala
@@ -18,12 +18,11 @@
 package org.apache.spark.sql.hive
 
 import com.google.common.cache.{CacheBuilder, CacheLoader, LoadingCache}
-import java.util.concurrent.ConcurrentHashMap
-import java.util.concurrent.locks.ReentrantReadWriteLock
+import com.google.common.util.concurrent.Striped
 import org.apache.hadoop.fs.Path
-import scala.collection.JavaConverters._
 
 import org.apache.spark.internal.Logging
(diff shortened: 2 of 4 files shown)


xuanyuanking:
@rxin @liancheng

rxin:
cc @ericl  can you take a look at this?

ericl:
Isn't it sufficient to lock around the `catalog.filterPartitions(Nil)`? Why do we need reader locks?

sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveMetastoreCatalog.scala
```diff
@@ -53,6 +56,10 @@ private[hive] class HiveMetastoreCatalog(sparkSession: SparkSession) extends Log
       tableIdent.table.toLowerCase)
   }
 
+  /** ReadWriteLock for each tables, protect the read and write cached */
+  protected[hive] val tableLockMap =
+    new ConcurrentHashMap[QualifiedTableName, ReentrantReadWriteLock]
```
ericl: https://google.github.io/guava/releases/19.0/api/docs/com/google/common/util/concurrent/Striped.html is usually good enough

xuanyuanking: Yes, I considered the Striped.lazyWeakReadWriteLock here but I need deal with invalidateAllCache(), it need all <K, V>, so I finally new a HashMap self.

xuanyuanking:
@ericl Thanks for your review.

> Is it sufficient to lock around the catalog.filterPartitions(Nil)?
Yes, this patch port from 1.6.2 and I missed the diff here. Fixed in next patch.

>  Why do we need reader locks?
Write or Invalid the table cache operation fewer than read it. Reader waiting when there is same table writing cache.

ericl:
I guess the large number of lock sites is confusing me. We only want to prevent concurrent instantiation of a single table, so shouldn't you only need 1 lock for that site?

Also, we should have a unit test that tries to concurrently read from a table from many threads, and verifies via the catalog metrics that it is only loaded once (see `TablePerfStatsSuite` for how to access the metrics).

sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveSessionCatalog.scala
```diff
@@ -105,7 +105,8 @@ private[sql] class HiveSessionCatalog(
   // For testing only
   private[hive] def getCachedDataSourceTable(table: TableIdentifier): LogicalPlan = {
     val key = metastoreCatalog.getQualifiedTableName(table)
-    metastoreCatalog.cachedDataSourceTables.getIfPresent(key)
+    metastoreCatalog.readLock(key,
+      metastoreCatalog.cachedDataSourceTables.getIfPresent(key))
```
ericl: Why a read lock here?

sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveSessionCatalog.scala
```diff
@@ -95,7 +95,7 @@ private[sql] class HiveSessionCatalog(
   }
 
   def invalidateCache(): Unit = {
-    metastoreCatalog.cachedDataSourceTables.invalidateAll()
+    metastoreCatalog.invalidateAllCache()
```
ericl: Why this change?

xuanyuanking:
hi @ericl 
This commit do the 3 things below, thanks for your check:
1. Delete the unnecessary lock use and simplify the lock operation
2. Add UT test in `PartitionedTablePerfStatsSuite`
3. Add cache hit metrics in `HiveCatalogMetrics`
Also change the description of this PR.

core/src/main/scala/org/apache/spark/metrics/source/StaticSources.scala
```diff
@@ -97,6 +97,12 @@ object HiveCatalogMetrics extends Source {
     MetricRegistry.name("parallelListingJobCount"))
 
   /**
+   * Tracks the total number of cachedDataSourceTables hits.
+   */
+  val METRIC_DATASOUCE_TABLE_CACHE_HITS = metricRegistry.counter(
+    MetricRegistry.name("dataSourceTableCacheHits"))
```
ericl: Could we use one of the other metrics, rather than add a new one?

sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveMetastoreCatalog.scala
```diff
@@ -33,6 +35,7 @@ import org.apache.spark.sql.hive.orc.OrcFileFormat
 import org.apache.spark.sql.types._
 
 
+
```
ericl: nit: extra newline

sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveMetastoreCatalog.scala
```diff
@@ -53,6 +56,18 @@ private[hive] class HiveMetastoreCatalog(sparkSession: SparkSession) extends Log
       tableIdent.table.toLowerCase)
   }
 
+  /** ReadWriteLock for each tables, protect the read and write cached */
```
ericl: Could you update this comment to say that the reason we lock is to prevent concurrent table instantiation?

sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveMetastoreCatalog.scala
```diff
@@ -53,6 +56,18 @@ private[hive] class HiveMetastoreCatalog(sparkSession: SparkSession) extends Log
...
+  private val tableLockStripes = Striped.lazyWeakLock(10)
+
+  /** Acquires a lock on the table cache for the duration of `f`. */
+  private def cacheLock[A](tableName: QualifiedTableName, f: => A): A = {
```
ericl: withTableCreationLock

sql/hive/src/test/scala/org/apache/spark/sql/hive/PartitionedTablePerfStatsSuite.scala
```diff
@@ -352,4 +353,28 @@ class PartitionedTablePerfStatsSuite
...
+              spark.sql("select * from test where partCol1 = 999").count()
+            }
+          }
+          executorPool.execute(runnable)
+          None
+        })
+        executorPool.shutdown()
+        executorPool.awaitTermination(30, TimeUnit.SECONDS)
+        // check the cache hit, the cache only load once
+        assert(HiveCatalogMetrics.METRIC_DATASOUCE_TABLE_CACHE_HITS.getCount() == 9)
```
ericl: Does this test fail without the lock?

sql/hive/src/test/scala/org/apache/spark/sql/hive/PartitionedTablePerfStatsSuite.scala
```diff
@@ -352,4 +353,28 @@ class PartitionedTablePerfStatsSuite
       }
     }
   }
+
+  test("SPARK-18700: add lock for each table's realation in cache") {
```
ericl: "table loaded only once even when resolved concurrently"

ericl (changes requested):

sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveMetastoreCatalog.scala
```diff
@@ -53,6 +56,18 @@ private[hive] class HiveMetastoreCatalog(sparkSession: SparkSession) extends Log
       tableIdent.table.toLowerCase)
   }
 
+  /** ReadWriteLock for each tables, protect the read and write cached */
+  private val tableLockStripes = Striped.lazyWeakLock(10)
```
ericl: Hm, may as well make it 100 if it's a lazy weak locks.

ericl: nit: `tableCreationLocks`.

sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveMetastoreCatalog.scala
xuanyuanking: fix done

sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveMetastoreCatalog.scala
xuanyuanking: update done, and more comments at HiveMetastoreCatalog.scala+226

sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveMetastoreCatalog.scala
xuanyuanking: fix done

sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveMetastoreCatalog.scala
xuanyuanking: update done

sql/hive/src/test/scala/org/apache/spark/sql/hive/PartitionedTablePerfStatsSuite.scala
xuanyuanking: update done

sql/hive/src/test/scala/org/apache/spark/sql/hive/PartitionedTablePerfStatsSuite.scala
xuanyuanking: It may failed sometimes without lock, but when I add `Thread.sleep(1000)` before `cachedDataSourceTables.put(tableIdentifier, created)` in HiveMetastoreCatalog.scala +265 to make the build table relation slow.It will failed every time. How can I do this hook in UT? Or how to make the cache build operation slow without really make a big table?  : )

ericl: Yeah that's fine, as long as it fails some fraction of the time it will eventually show up as a flaky test.

core/src/main/scala/org/apache/spark/metrics/source/StaticSources.scala
```diff
@@ -105,6 +111,7 @@ object HiveCatalogMetrics extends Source {
     METRIC_FILE_CACHE_HITS.dec(METRIC_FILE_CACHE_HITS.getCount())
     METRIC_HIVE_CLIENT_CALLS.dec(METRIC_HIVE_CLIENT_CALLS.getCount())
     METRIC_PARALLEL_LISTING_JOB_COUNT.dec(METRIC_PARALLEL_LISTING_JOB_COUNT.getCount())
+    METRIC_DATASOUCE_TABLE_CACHE_HITS.dec(METRIC_DATASOUCE_TABLE_CACHE_HITS.getCount())
```
ericl: s/souce/source

core/src/main/scala/org/apache/spark/metrics/source/StaticSources.scala
ericl: That's kind of odd, I'd expect the duplicate table building to cause more file accesses or at least cache hits since we are scanning the filesystem multiple times. Is that not the case?

xuanyuanking: I may find the reason of the odd scenario, please check my conclusion:
In 2.0 add a new config lazyPruningEnabled, it's default value is true and while multi-thread do the building same time, it will not do listLeafFile.
So I set the `HIVE_MANAGE_FILESOURCE_PARTITIONS`=false and set the partition larger than `PARALLEL_PARTITION_DISCOVERY_THRESHOLD`(this will cause METRIC_PARALLEL_LISTING_JOB_COUNT +1), test results list below:
without lock:
```
0
0
550   (50 file * 11, 1 is cache.load() and the other 10 is 10 threads)
90
11      (also 1 * 11)
1000
```
and with lock:
```
9
0
100  (2 * 50, 1 is cache.load() and the other 1 is the first threads)
54
2     (also 1 * 2)
550
```
so I can delete the added `dataSourceTableCacheHits` metrics and use `parallelListingJobCount` and `filesDiscovered` instead. Do this in next patch.

core/src/main/scala/org/apache/spark/metrics/source/StaticSources.scala
xuanyuanking: e...sorry, this new added metric will delete next patch like before comment

sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveMetastoreCatalog.scala
```diff
@@ -209,72 +221,79 @@ private[hive] class HiveMetastoreCatalog(sparkSession: SparkSession) extends Log
...
-        val fileCatalog = {
-          val catalog = new CatalogFileIndex(
-            sparkSession, metastoreRelation.catalogTable, sizeInBytes)
-          if (lazyPruningEnabled) {
-            catalog
-          } else {
-            catalog.filterPartitions(Nil)  // materialize all the partitions in memory
+      // Here we should protect all relation get and create operation with lock while big
+      // table's CatalogFileIndex will take some time, only lock cachedDataSourceTables.put
+      // will still cause driver memory waste. More detail see SPARK-18700.
```
ericl: nit: probably don't need this comment

sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveMetastoreCatalog.scala
```diff
@@ -53,6 +53,18 @@ private[hive] class HiveMetastoreCatalog(sparkSession: SparkSession) extends Log
       tableIdent.table.toLowerCase)
   }
 
+  /** Locks for preventing driver mem waste when concurrent table instantiation */
+  private val tableCreationLocks = Striped.lazyWeakLock(100)
```
ericl: nit: "These locks guard against multiple attempts to instantiate a table, which wastes memory."

sql/hive/src/test/scala/org/apache/spark/sql/hive/PartitionedTablePerfStatsSuite.scala
```diff
@@ -352,4 +353,34 @@ class PartitionedTablePerfStatsSuite
...
+            }
+            executorPool.execute(runnable)
+            None
+          })
+          executorPool.shutdown()
+          executorPool.awaitTermination(30, TimeUnit.SECONDS)
+          // check the cache hit, we use the metric of METRIC_FILES_DISCOVERED and
+          // METRIC_PARALLEL_LISTING_JOB_COUNT to check this, while the lock take effect,
+          // only one thread can really do the build, so the listing job count is 2, the other
+          // one is cahce.load func. Also METRIC_FILES_DISCOVERED is $partition_num * 2
```
ericl: s/cahce/cache

ericl (approved):
Thanks for making these changes, this lgtm with some nits.

sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveMetastoreCatalog.scala
xuanyuanking: fix done

sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveMetastoreCatalog.scala
xuanyuanking: done, delete it

sql/hive/src/test/scala/org/apache/spark/sql/hive/PartitionedTablePerfStatsSuite.scala
xuanyuanking: ……sorry, fix done

xuanyuanking:
Thanks for ericl's review!

ericl:
cc @rxin

xuanyuanking:
cc @rxin thanks for check.

hvanhovell:
retest this please

hvanhovell:
I am merging this one after a successful test run. Ping me if you object.

rxin:
Go for it.

hvanhovell:
Merging to master/2.1. @xuanyuanking can you open a backport for 2.0, if we also need to merge this to that branche?

xuanyuanking:
@hvanhovell Sure, I open a new BACKPORT-2.0.
There's a little diff in branch-2.0, the ut test of this patch based on the `HiveCatalogMetrics` which not added in 2.0, so I added the patch need metric. Thanks for check.

sql/hive/src/test/scala/org/apache/spark/sql/hive/PartitionedTablePerfStatsSuite.scala
```diff
@@ -352,4 +353,34 @@ class PartitionedTablePerfStatsSuite
...
+            }
+            executorPool.execute(runnable)
+            None
+          })
+          executorPool.shutdown()
+          executorPool.awaitTermination(30, TimeUnit.SECONDS)
+          // check the cache hit, we use the metric of METRIC_FILES_DISCOVERED and
+          // METRIC_PARALLEL_LISTING_JOB_COUNT to check this, while the lock take effect,
+          // only one thread can really do the build, so the listing job count is 2, the other
+          // one is cache.load func. Also METRIC_FILES_DISCOVERED is $partition_num * 2
```
gatorsmile: This comment is wrong. The extra counts are from the `DataFrameWriter`'s save() API.

gatorsmile: Working on a fix to avoid the useless filesystem scan caused by the save() API.

xuanyuanking: @gatorsmile Xiao fixed this in https://github.com/apache/spark/pull/16481

PR #16137, opened 2016-12-04
[SPARK-18708][CORE] Improvement/improve docs in spark context file

Mironor:

SparkContext.scala was created a long time ago and contains several types of Scaladocs/Javadocs mixed together. Public methods/fields should have a Scaladoc that is formatted in the same way everywhere. This pull request also adds scaladoc to methods/fields that did not have it before.

## How was this patch tested?

No actual code was modified, only comments.

### DIFF
--- a/core/src/main/scala/org/apache/spark/SparkContext.scala
+++ b/core/src/main/scala/org/apache/spark/SparkContext.scala
@@ -224,7 +224,7 @@ class SparkContext(config: SparkConf) extends Logging {
   private[spark] def conf: SparkConf = _conf
 
   /**
-   * @return copy of this `SparkContext`'s configuration. The configuration ''cannot'' be
+   * Return a copy of this SparkContext's configuration. The configuration ''cannot'' be
    * changed at runtime.
    */
   def getConf: SparkConf = conf.clone()
@@ -358,9 +358,7 @@ class SparkContext(config: SparkConf) extends Logging {
     }
   }
 
-  /**
-   * Control our logLevel. This overrides any user-defined log settings.
-   *
+  /** Control our logLevel. This overrides any user-defined log settings.
    * @param logLevel The desired log level as a string.
    * Valid log levels include: ALL, DEBUG, ERROR, FATAL, INFO, OFF, TRACE, WARN
    */
@@ -707,8 +705,7 @@ class SparkContext(config: SparkConf) extends Logging {
 
   // Methods for creating RDDs
 
-  /**
-   * Distribute a local Scala collection to form an RDD.
+  /** Distribute a local Scala collection to form an RDD.
    *
    * @note Parallelize acts lazily. If `seq` is a mutable collection and is altered after the call
    * to parallelize and before the first action on the RDD, the resultant RDD will reflect the
@@ -727,10 +724,11 @@ class SparkContext(config: SparkConf) extends Logging {
   }
 
   /**
-   * Creates a new `RDD[Long]` containing elements from `start` to `end`(exclusive), increased by
+   * Creates a new RDD[Long] containing elements from `start` to `end`(exclusive), increased by
    * `step` every element.
    *
    * @note if we need to cache this RDD, we should make sure each partition does not exceed limit.
+   *
    * @param start the start value.
    * @param end the end value.
    * @param step the incremental step
@@ -797,10 +795,9 @@ class SparkContext(config: SparkConf) extends Logging {
     }
   }
 
-  /**
-   * Distribute a local Scala collection to form an RDD.
-   * This method is identical to `parallelize`.
+  /** Distribute a local Scala collection to form an RDD.
    *
+   * This method is identical to `parallelize`.
    * @param seq Scala collection to distribute
    * @param numSlices number of partitions to divide the collection into
    * @return RDD representing distributed collection
@@ -815,9 +812,8 @@ class SparkContext(config: SparkConf) extends Logging {
    * Distribute a local Scala collection to form an RDD, with one or more
    * location preferences (hostnames of Spark nodes) for each object.
    * Create a new partition for each collection item.
-   *
-   * @param seq a list of tuples containing data and location preferences
-   * @return RDD with data sliced according to location preferences
+   * @param seq list of tuples of data and location preferences (hostnames of Spark nodes)
+   * @return RDD representing data partitioned according to location preferences
    */
   def makeRDD[T: ClassTag](seq: Seq[(T, Seq[String])]): RDD[T] = withScope {
     assertNotStopped()
@@ -828,10 +824,9 @@ class SparkContext(config: SparkConf) extends Logging {
   /**
    * Read a text file from HDFS, a local file system (available on all nodes), or any
    * Hadoop-supported file system URI, and return it as an RDD of Strings.
-   *
    * @param path path to the text file on a supported file system
    * @param minPartitions suggested minimum number of partitions for the resulting RDD
-   * @return RDD representing the content of the file
+   * @return RDD of lines of the text file
    */
   def textFile(
       path: String,
@@ -846,7 +841,7 @@ class SparkContext(config: SparkConf) extends Logging {
    * Hadoop-supported file system URI. Each file is read as a single record and returned in a
    * key-value pair, where the key is the path of each file, the value is the content of each file.
    *
-   * For example, if you have the following files:
+   * <p> For example, if you have the following files:
    * {{{
    *   hdfs://a-hdfs-path/part-00000
    *   hdfs://a-hdfs-path/part-00001
@@ -856,7 +851,7 @@ class SparkContext(config: SparkConf) extends Logging {
    *
    * Do `val rdd = sparkContext.wholeTextFile("hdfs://a-hdfs-path")`,
    *
-   * then `rdd` contains
+   * <p> then `rdd` contains
    * {{{
    *   (a-hdfs-path/part-00000, its content)
    *   (a-hdfs-path/part-00001, its content)
@@ -867,6 +862,7 @@ class SparkContext(config: SparkConf) extends Logging {
    * @note Small files are preferred, large file is also allowable, but may cause bad performance.
    * @note On some filesystems, `.../path/&#42;` can be a more efficient way to read all files
    *       in a directory rather than `.../path/` or `.../path`
+   *
    * @param path Directory to the input data files, the path can be comma separated paths as the
    *             list of inputs.
    * @param minPartitions A suggestion value of the minimal splitting number for input data.
@@ -945,10 +941,12 @@ class SparkContext(config: SparkConf) extends Logging {
    *
    * @note We ensure that the byte array for each record in the resulting RDD
    * has the provided record length.
+   *
    * @param path Directory to the input data files, the path can be comma separated paths as the
    *             list of inputs.
    * @param recordLength The length at which to split the records
    * @param conf Configuration for setting up the dataset.
+   *
    * @return An RDD of data with values, represented as byte arrays
    */
   def binaryRecords(
@@ -971,25 +969,25 @@ class SparkContext(config: SparkConf) extends Logging {
   }
 
   /**
-   * Get an RDD for a Hadoop-readable dataset from a Hadoop `JobConf` given its `InputFormat`
-   * and other necessary info (e.g. file name for a filesystem-based dataset, table name
-   * for HyperTable), using the older MapReduce API (`org.apache.hadoop.mapred`).
+   * Get an RDD for a Hadoop-readable dataset from a Hadoop JobConf given its InputFormat and other
+   * necessary info (e.g. file name for a filesystem-based dataset, table name for HyperTable),
+   * using the older MapReduce API (`org.apache.hadoop.mapred`).
    *
-   * @param conf `JobConf` for setting up the dataset. Note: This will be put into a Broadcast.
+   * @param conf JobConf for setting up the dataset. Note: This will be put into a Broadcast.
    *             Therefore if you plan to reuse this conf to create multiple RDDs, you need to make
    *             sure you won't modify the conf. A safe approach is always creating a new conf for
    *             a new RDD.
-   * @param inputFormatClass Class of the InputFormat
-   * @param keyClass Class of the keys
-   * @param valueClass Class of the values
-   * @param minPartitions minimum number of Hadoop partitions to generate.
+   * @param inputFormatClass storage format of the data to be read
+   * @param keyClass `Class` of the key associated with the `inputFormatClass` parameter
+   * @param valueClass `Class` of the value associated with the `inputFormatClass` parameter
+   * @param minPartitions Minimum number of Hadoop Splits to generate.
    * @return RDD of tuples of key and corresponding value
    *
-   * @note Because Hadoop's `RecordReader` class re-uses the same Writable object for each
-   * record, directly caching the returned RDD or directly passing it to an aggregation
-   * or shuffle operation will create many references to the same object.
-   * If you plan to directly cache, sort, or aggregate Hadoop writable objects, you
-   * should first copy them using a `map` function.
+   * @note Because Hadoop's RecordReader class re-uses the same Writable object for each
+   * record, directly caching the returned RDD or directly passing it to an aggregation or shuffle
+   * operation will create many references to the same object.
+   * If you plan to directly cache, sort, or aggregate Hadoop writable objects, you should first
+   * copy them using a `map` function.
    */
   def hadoopRDD[K, V](
       conf: JobConf,
@@ -1008,20 +1006,19 @@ class SparkContext(config: SparkConf) extends Logging {
     new HadoopRDD(this, conf, inputFormatClass, keyClass, valueClass, minPartitions)
   }
 
-  /**
-   * Get an RDD for a Hadoop file with an arbitrary `InputFormat`
+  /** Get an RDD for a Hadoop file with an arbitrary InputFormat
    *
-   * @note Because Hadoop's `RecordReader` class re-uses the same `Writable` object for each
-   * record, directly caching the returned RDD or directly passing it to an aggregation
-   * or shuffle operation will create many references to the same object.
-   * If you plan to directly cache, sort, or aggregate Hadoop writable objects, you
-   * should first copy them using a `map` function.
+   * @note Because Hadoop's RecordReader class re-uses the same Writable object for each
+   * record, directly caching the returned RDD or directly passing it to an aggregation or shuffle
+   * operation will create many references to the same object.
+   * If you plan to directly cache, sort, or aggregate Hadoop writable objects, you should first
+   * copy them using a `map` function.
    * @param path directory to the input data files, the path can be comma separated paths
    * as a list of inputs
-   * @param inputFormatClass Class of the `InputFormat`
-   * @param keyClass Class of the keys
-   * @param valueClass Class of the values
-   * @param minPartitions minimum number of Hadoop partitions to generate
+   * @param inputFormatClass storage format of the data to be read
+   * @param keyClass `Class` of the key associated with the `inputFormatClass` parameter
+   * @param valueClass `Class` of the value associated with the `inputFormatClass` parameter
+   * @param minPartitions suggested minimum number of partitions for the resulting RDD
    * @return RDD of tuples of key and corresponding value
    */
   def hadoopFile[K, V](
@@ -1050,24 +1047,21 @@ class SparkContext(config: SparkConf) extends Logging {
   }
 
   /**
-   * Smarter version of `hadoopFile()` that uses class tags to figure out the classes of keys,
+   * Smarter version of hadoopFile() that uses class tags to figure out the classes of keys,
    * values and the InputFormat so that users don't need to pass them directly. Instead, callers
    * can just write, for example,
    * {{{
    * val file = sparkContext.hadoopFile[LongWritable, Text, TextInputFormat](path, minPartitions)
    * }}}
    *
-   * @note Because Hadoop's `RecordReader` class re-uses the same `Writable` object for each
-   * record, directly caching the returned RDD or directly passing it to an aggregation
-   * or shuffle operation will create many references to the same object.
-   * If you plan to directly cache, sort, or aggregate Hadoop writable objects, you
-   * should first copy them using a `map` function.
+   * @note Because Hadoop's RecordReader class re-uses the same Writable object for each
+   * record, directly caching the returned RDD or directly passing it to an aggregation or shuffle
+   * operation will create many references to the same object.
+   * If you plan to directly cache, sort, or aggregate Hadoop writable objects, you should first
+   * copy them using a `map` function.
    * @param path directory to the input data files, the path can be comma separated paths
    * as a list of inputs
-   * @param minPartitions minimum number of Hadoop partitions to generate
-   * @param km implicitly found class of the keys
-   * @param vm implicitly found class of the values
-   * @param fm implicitly found class of InputFormat
+   * @param minPartitions suggested minimum number of partitions for the resulting RDD
    * @return RDD of tuples of key and corresponding value
    */
   def hadoopFile[K, V, F <: InputFormat[K, V]]
@@ -1081,23 +1075,20 @@ class SparkContext(config: SparkConf) extends Logging {
   }
 
   /**
-   * Smarter version of `hadoopFile()` that uses class tags to figure out the classes of keys,
-   * values and the `InputFormat` so that users don't need to pass them directly. Instead, callers
+   * Smarter version of hadoopFile() that uses class tags to figure out the classes of keys,
+   * values and the InputFormat so that users don't need to pass them directly. Instead, callers
    * can just write, for example,
    * {{{
    * val file = sparkContext.hadoopFile[LongWritable, Text, TextInputFormat](path)
    * }}}
    *
-   * @note Because Hadoop's `RecordReader` class re-uses the same `Writable` object for each
-   * record, directly caching the returned RDD or directly passing it to an aggregation
-   * or shuffle operation will create many references to the same object.
-   * If you plan to directly cache, sort, or aggregate Hadoop writable objects, you
-   * should first copy them using a `map` function.
+   * @note Because Hadoop's RecordReader class re-uses the same Writable object for each
+   * record, directly caching the returned RDD or directly passing it to an aggregation or shuffle
+   * operation will create many references to the same object.
+   * If you plan to directly cache, sort, or aggregate Hadoop writable objects, you should first
+   * copy them using a `map` function.
    * @param path directory to the input data files, the path can be comma separated paths as
    * a list of inputs
-   * @param km implicitly found class of the keys
-   * @param vm implicitly found class of the values
-   * @param fm implicitly found class of `InputFormat`
    * @return RDD of tuples of key and corresponding value
    */
   def hadoopFile[K, V, F <: InputFormat[K, V]](path: String)
@@ -1105,7 +1096,7 @@ class SparkContext(config: SparkConf) extends Logging {
     hadoopFile[K, V, F](path, defaultMinPartitions)
   }
 
-  /** Get an RDD for a Hadoop file with an arbitrary new API `InputFormat`. */
+  /** Get an RDD for a Hadoop file with an arbitrary new API InputFormat. */
   def newAPIHadoopFile[K, V, F <: NewInputFormat[K, V]]
       (path: String)
       (implicit km: ClassTag[K], vm: ClassTag[V], fm: ClassTag[F]): RDD[(K, V)] = withScope {
@@ -1117,19 +1108,19 @@ class SparkContext(config: SparkConf) extends Logging {
   }
 
   /**
-   * Get an RDD for a given Hadoop file with an arbitrary new API `InputFormat`
+   * Get an RDD for a given Hadoop file with an arbitrary new API InputFormat
    * and extra configuration options to pass to the input format.
    *
-   * @note Because Hadoop's `RecordReader` class re-uses the same Writable object for each
-   * record, directly caching the returned RDD or directly passing it to an aggregation
-   * or shuffles operation will create many references to the same object.
-   * If you plan to directly cache, sort, or aggregate Hadoop writable objects, you
-   * should first copy them using a `map` function.
+   * @note Because Hadoop's RecordReader class re-uses the same Writable object for each
+   * record, directly caching the returned RDD or directly passing it to an aggregation or shuffle
+   * operation will create many references to the same object.
+   * If you plan to directly cache, sort, or aggregate Hadoop writable objects, you should first
+   * copy them using a `map` function.
    * @param path directory to the input data files, the path can be comma separated paths
    * as a list of inputs
-   * @param fClass class of the `InputFormat`
-   * @param kClass class of the keys
-   * @param vClass class of the values
+   * @param fClass storage format of the data to be read
+   * @param kClass `Class` of the key associated with the `fClass` parameter
+   * @param vClass `Class` of the value associated with the `fClass` parameter
    * @param conf Hadoop configuration
    * @return RDD of tuples of key and corresponding value
    */
@@ -1156,22 +1147,22 @@ class SparkContext(config: SparkConf) extends Logging {
   }
 
   /**
-   * Get an RDD for a given Hadoop file with an arbitrary new API `InputFormat`
+   * Get an RDD for a given Hadoop file with an arbitrary new API InputFormat
    * and extra configuration options to pass to the input format.
    *
-   * @note because Hadoop's `RecordReader` class re-uses the same `Writable` object for each
-   * record, directly caching the returned RDD or directly passing it to an aggregation
-   * or shuffle operation will create many references to the same object.
-   * If you plan to directly cache, sort, or aggregate Hadoop writable objects, you
-   * should first copy them using a `map` function.
    * @param conf Configuration for setting up the dataset. Note: This will be put into a Broadcast.
    *             Therefore if you plan to reuse this conf to create multiple RDDs, you need to make
    *             sure you won't modify the conf. A safe approach is always creating a new conf for
    *             a new RDD.
-   * @param fClass Class of the `InputFormat`
-   * @param kClass Class of the keys
-   * @param vClass Class of the values
-   * @return RDD of tuples of key and corresponding value
+   * @param fClass storage format of the data to be read
+   * @param kClass `Class` of the key associated with the `fClass` parameter
+   * @param vClass `Class` of the value associated with the `fClass` parameter
+   *
+   * @note Because Hadoop's RecordReader class re-uses the same Writable object for each
+   * record, directly caching the returned RDD or directly passing it to an aggregation or shuffle
+   * operation will create many references to the same object.
+   * If you plan to directly cache, sort, or aggregate Hadoop writable objects, you should first
+   * copy them using a `map` function.
    */
   def newAPIHadoopRDD[K, V, F <: NewInputFormat[K, V]](
       conf: Configuration = hadoopConfiguration,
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/apache/spark/SparkContext.scala
```diff
@@ -119,22 +119,22 @@ class SparkContext(config: SparkConf) extends Logging {
   /**
    * Alternative constructor that allows setting common Spark properties directly
    *
-   * @param master Cluster URL to connect to (e.g. mesos://host:port, spark://host:port, local[4]).
-   * @param appName A name for your application, to display on the cluster web UI
-   * @param conf a [[org.apache.spark.SparkConf]] object specifying other Spark parameters
+   * @param master cluster URL to connect to (e.g. mesos://host:port, spark://host:port, local[4]).
```
srowen: It's not worth un-capitalizing these words

core/src/main/scala/org/apache/spark/SparkContext.scala
```diff
@@ -119,22 +119,22 @@ class SparkContext(config: SparkConf) extends Logging {
...
+   * @param appName a name for your application, to display on the cluster web UI
+   * @param conf a `org.apache.spark.SparkConf` object specifying other Spark parameters
```
srowen: Does it not render properly? then just `SparkConf` is fine

core/src/main/scala/org/apache/spark/SparkContext.scala
```diff
@@ -224,8 +224,8 @@ class SparkContext(config: SparkConf) extends Logging {
   private[spark] def conf: SparkConf = _conf
 
   /**
-   * Return a copy of this SparkContext's configuration. The configuration ''cannot'' be
-   * changed at runtime.
+   * @return copy of this `SparkContext`'s configuration. The configuration ''cannot'' be
+   *         changed at runtime.
```
srowen: This is fine but you don't need a continuation indent.

core/src/main/scala/org/apache/spark/SparkContext.scala
```diff
@@ -276,13 +279,17 @@ class SparkContext(config: SparkConf) extends Logging {
 
   private[spark] def ui: Option[SparkUI] = _ui
 
+  /**
+   * @return `Option` containing url to web ui.
```
srowen: url -> URL and ui -> UI, but, I don't this comment adds much to the reader beyond what the signature says already. I'd improve or remove it.

core/src/main/scala/org/apache/spark/SparkContext.scala
```diff
@@ -276,13 +279,17 @@ class SparkContext(config: SparkConf) extends Logging {
...
   def uiWebUrl: Option[String] = _ui.map(_.webUrl)
 
   /**
    * A default Hadoop Configuration for the Hadoop code (e.g. file systems) that we reuse.
    *
-   * @note As it will be reused in all Hadoop RDDs, it's better not to modify it unless you
-   * plan to set some global configurations for all Hadoop RDDs.
+   * @note as it will be reused in all Hadoop RDDs, it's better not to modify it unless you
+   *       plan to set some global configurations for all Hadoop RDDs.
+   * @return Hadoop configuration
```
srowen: This isn't a useful comment; I'd revert changes like this

core/src/main/scala/org/apache/spark/SparkContext.scala
```diff
@@ -461,11 +478,12 @@ class SparkContext(config: SparkConf) extends Logging {
 
     _hadoopConfiguration = SparkHadoopUtil.get.newConfiguration(_conf)
 
-    // Add each JAR given through the constructor
+    // Adds each JAR given through the constructor
```
srowen: Ditto, lots of comments like this just aren't adding anything

srowen (changes requested):
Rather than review the rest I'd say that most of these changes aren't worth making. I would start over and only make changes where you're fixing a typo or markdown, or adding a non-trivial public API comment. Formatting and trivia just won't be worth it.

Mironor:
Hi @srowen , it's quite hard to directly jump on something like "refactoring core scheduling logic", I thought fixing small things would be a good start. To make it simple I limited all the changes to only one file, I also took DAGSceduler.scala as an example for what documenting things should look like (as currently SparkContext.scala is a mix of different document styles, but maybe it's all fine). That's where the continuation indent comes from.

In any case thanks for taking your time to review this PR, I would really appreciate if you could tell me if it's worth removing obvious or insignificant (like de-capitalizing sentences) comments to make this PR less overwhelming and worthwhile of taking a review.

srowen:
It is good to start small, but not all small changes are worthwhile. It would be better to focus on fixing actual problems like a typo or missing doc. You can then expand to improve / fix docs across many files, not just one.

core/src/main/scala/org/apache/spark/SparkContext.scala
HyukjinKwon: This should render fine if I remember correctly although I am fine with either way but I would like to know the reason to make this wrapped with backticks if there is.

core/src/main/scala/org/apache/spark/SparkContext.scala
```diff
@@ -1401,8 +1532,11 @@ class SparkContext(config: SparkConf) extends Logging {
 
   /**
    * Broadcast a read-only variable to the cluster, returning a
-   * [[org.apache.spark.broadcast.Broadcast]] object for reading it in distributed functions.
+   * `org.apache.spark.broadcast.Broadcast` object for reading it in distributed functions.
```
HyukjinKwon: I believe this should  also render fine already.

core/src/main/scala/org/apache/spark/SparkContext.scala
```diff
@@ -1417,27 +1551,31 @@ class SparkContext(config: SparkConf) extends Logging {
 
   /**
    * Add a file to be downloaded with this Spark job on every node.
-   * The `path` passed can be either a local file, a file in HDFS (or other Hadoop-supported
-   * filesystems), or an HTTP, HTTPS or FTP URI.  To access the file in Spark jobs,
-   * use `SparkFiles.get(fileName)` to find its download location.
+   *
+   * @param path can be either a local file, a file in HDFS (or other Hadoop-supported
+   *             filesystems), or an HTTP, HTTPS or FTP URI.  To access the file in Spark jobs,
```
HyukjinKwon: nit: we could make this single spaced here` URI.  To` and also same instances, at least for the lines this PR changes.

core/src/main/scala/org/apache/spark/SparkContext.scala
```diff
@@ -923,15 +971,13 @@ class SparkContext(config: SparkConf) extends Logging {
   /**
    * Load data from a flat binary file, assuming the length of each record is constant.
    *
-   * @note We ensure that the byte array for each record in the resulting RDD
+   * @note we ensure that the byte array for each record in the resulting RDD
```
HyukjinKwon: I am not sure of making this lower-cased because we will see this as the start of sentence 
- scaladoc
  

- javadoc

core/src/main/scala/org/apache/spark/SparkContext.scala
```diff
@@ -119,22 +119,22 @@ class SparkContext(config: SparkConf) extends Logging {
...
-   * @param appName A name for your application, to display on the cluster web UI.
-   * @param sparkHome Location where Spark is installed on cluster nodes.
-   * @param jars Collection of JARs to send to the cluster. These can be paths on the local file
+   * @param master cluster URL to connect to (e.g. mesos://host:port, spark://host:port, local[4]).
+   * @param appName a name for your application, to display on the cluster web UI.
+   * @param sparkHome location where Spark is installed on cluster nodes.
+   * @param jars collection of JARs to send to the cluster. These can be paths on the local file
    *             system or HDFS, HTTP, HTTPS, or FTP URLs.
-   * @param environment Environment variables to set on worker nodes.
+   * @param environment environment variables to set on worker nodes.
```
HyukjinKwon: Do we have a rule (or other references) to make them lower-cased? I am worried of similar changes in the future and it might be great if we have a concrete reason here to change so.

HyukjinKwon: I am fine if there is not too if there can be coherent and this can be decided here too.

HyukjinKwon: Oh, I am sorry it was mentioned in https://github.com/apache/spark/pull/16137/files#r90948641.

core/src/main/scala/org/apache/spark/SparkContext.scala
```diff
@@ -1144,13 +1218,19 @@ class SparkContext(config: SparkConf) extends Logging {
...
   /**
-   * Get an RDD for a Hadoop SequenceFile with given key and value types.
+   * Get an RDD for a Hadoop `SequenceFile` with given key and value types.
    *
-   * @note Because Hadoop's RecordReader class re-uses the same Writable object for each
-   * record, directly caching the returned RDD or directly passing it to an aggregation or shuffle
-   * operation will create many references to the same object.
-   * If you plan to directly cache, sort, or aggregate Hadoop writable objects, you should first
-   * copy them using a `map` function.
+   * @note because Hadoop's `RecordReader` class re-uses the same `Writable` object for each
```
HyukjinKwon: I am worried of this change as well. I at least intendedly pick up a one style of those `@note` and followed it when I changed this in the last PR.  I guess we would not want to change those indentation without a concrete reason to prevent changing forward and backward.

core/src/main/scala/org/apache/spark/SparkContext.scala
Mironor: Yes, the initial intent was to make it the same everywhere (either backquotes or brackets)

core/src/main/scala/org/apache/spark/SparkContext.scala
```diff
@@ -1417,27 +1551,31 @@ class SparkContext(config: SparkConf) extends Logging {
...
+   *             filesystems), or an HTTP, HTTPS or FTP URI.  To access the file in Spark jobs,
+   *             use `SparkFiles.get(fileName)` to find its download location.
    */
   def addFile(path: String): Unit = {
     addFile(path, false)
   }
 
   /**
-   * Returns a list of file paths that are added to resources.
+   * A list of file paths that are added to resources.
```
HyukjinKwon: Would this be better if we leave `Returns` for a description for a function even if `@return` covers it?

Mironor: The second line is redundant here, my question here is whether it's worth to replace `Return` with `@return` or just to leave it as it is.

core/src/main/scala/org/apache/spark/SparkContext.scala
```diff
@@ -1620,8 +1766,9 @@ class SparkContext(config: SparkConf) extends Logging {
 
   /**
    * :: DeveloperApi ::
-   * Return information about what RDDs are cached, if they are in mem or on disk, how much space
-   * they take, etc.
```
HyukjinKwon: If you want to remove `Return` or make the description into `@returns`, I guess it should be at least consistent. It seem https://github.com/apache/spark/pull/16137/files#r91009309 is a bit different.

core/src/main/scala/org/apache/spark/SparkContext.scala
```diff
@@ -2584,10 +2807,11 @@ private object SparkMasterRegex {
 }
 
 /**
- * A class encapsulating how to convert some type T to Writable. It stores both the Writable class
- * corresponding to T (e.g. IntWritable for Int) and a function for doing the conversion.
- * The getter for the writable class takes a ClassTag[T] in case this is a generic object
- * that doesn't know the type of T when it is created. This sounds strange but is necessary to
+ * A class encapsulating how to convert some type `T` to `Writable`. It stores both the
+ * `Writable` class corresponding to `T`   (e.g. `IntWritable` for `Int`) and a function for doing
```
HyukjinKwon: nit: spaces :) ` `T`   (e.g`

core/src/main/scala/org/apache/spark/SparkContext.scala
HyukjinKwon: Actually brackets are better because they make links. Some were backquoted because javadoc8 complains about this.

core/src/main/scala/org/apache/spark/SparkContext.scala
srowen: We shouldn't duplicate the documentation. converting this to a `@return` is fine.

core/src/main/scala/org/apache/spark/SparkContext.scala
Mironor: [Spark style guide](http://spark.apache.org/contributing.html) doesn't contain anything about continuation indentation and refers to the [Scala's own style guide](http://docs.scala-lang.org/style/scaladoc.html) which shows that indentation should be used.

HyukjinKwon: I believe strictly we follow javadoc style according to http://spark.apache.org/contributing.html

> For Scala doc / Java doc comment before classes, objects and methods, use Java docs style instead of Scala docs style.

HyukjinKwon: I understand they are used in a mixed way and I see the example of multiple lines with `@return` in the scaladoc. I am fine with this but I just wanted to note my worry here.

Mironor: Correct, but [they](http://www.oracle.com/technetwork/articles/java/index-137868.html) also contain continuation indentation (they even align parameter description)

Mironor:
@srowen I reversed obvious comments as well as some minor changes (such as capitalizing). I only left javadoc for some of the non-trivial public api. I can reverse changes for comments where the only diff is the wrapping of references in backquotes.
I'd like to know if you have some idea on whether to use indentation (as in [javadoc](http://www.oracle.com/technetwork/articles/java/index-137868.html)/[scaladoc](http://docs.scala-lang.org/style/scaladoc.html)) as well as what type of character to use when linking a reference (backquote or brackets?), I could modify Spark code style documentation if if's different from javadoc.

core/src/main/scala/org/apache/spark/SparkContext.scala
```diff
@@ -64,11 +64,11 @@ import org.apache.spark.util._
  * Main entry point for Spark functionality. A SparkContext represents the connection to a Spark
  * cluster, and can be used to create RDDs, accumulators and broadcast variables on that cluster.
  *
- * Only one SparkContext may be active per JVM.  You must `stop()` the active SparkContext before
- * creating a new one.  This limitation may eventually be removed; see SPARK-2243 for more details.
+ * Only one SparkContext may be active per JVM. You must `stop()` the active SparkContext before
```
srowen: I still just don't think it's worth removing spaces. They don't even matter in rendered HTML

core/src/main/scala/org/apache/spark/SparkContext.scala
```diff
@@ -358,9 +361,11 @@ class SparkContext(config: SparkConf) extends Logging {
     }
   }
 
-  /** Control our logLevel. This overrides any user-defined log settings.
+  /**
+   * Control our logLevel. This overrides any user-defined log settings.
+   *
    * @param logLevel The desired log level as a string.
-   * Valid log levels include: ALL, DEBUG, ERROR, FATAL, INFO, OFF, TRACE, WARN
+   *                 Valid log levels include: ALL, DEBUG, ERROR, FATAL, INFO, OFF, TRACE, WARN
```
srowen: Yes we don't need continuation indents. The code doesn't use it in general and wastes some space

core/src/main/scala/org/apache/spark/SparkContext.scala
```diff
@@ -670,10 +677,14 @@ class SparkContext(config: SparkConf) extends Logging {
    * sc.cancelJobGroup("some_job_to_cancel")
    * }}}
    *
-   * If interruptOnCancel is set to true for the job group, then job cancellation will result
-   * in Thread.interrupt() being called on the job's executor threads. This is useful to help ensure
-   * that the tasks are actually stopped in a timely manner, but is off by default due to HDFS-1208,
-   * where HDFS may respond to Thread.interrupt() by marking nodes as dead.
+   * @param groupId job group's identification
+   * @param description job group's decription
```
srowen: The `interruptOnCancel` doc makes sense but these don't say much beyond the name. I'd expand them or remove them.

core/src/main/scala/org/apache/spark/SparkContext.scala
```diff
@@ -901,6 +926,7 @@ class SparkContext(config: SparkConf) extends Logging {
    * @param path Directory to the input data files, the path can be comma separated paths as the
    *             list of inputs.
    * @param minPartitions A suggestion value of the minimal splitting number for input data.
+   * @return RDD containing `Tuple2` with path to the file and the corresponding file content
```
srowen: Although these don't add a lot it's kind of nice to doc `@return` for completeness here. I might change the language in all similar docs a bit to something more like "return an RDD of tuples of file path and corresponding file content". That is there are many tuples.

core/src/main/scala/org/apache/spark/SparkContext.scala
```diff
@@ -953,24 +977,24 @@ class SparkContext(config: SparkConf) extends Logging {
...
   /**
-   * Get an RDD for a Hadoop-readable dataset from a Hadoop JobConf given its InputFormat and other
-   * necessary info (e.g. file name for a filesystem-based dataset, table name for HyperTable),
-   * using the older MapReduce API (`org.apache.hadoop.mapred`).
+   * Get an RDD for a Hadoop-readable dataset from a Hadoop `JobConf` given its `InputFormat`
+   * and other necessary info (e.g. file name for a filesystem-based dataset, table name
+   * for HyperTable), using the older MapReduce API (`org.apache.hadoop.mapred`).
    *
-   * @param conf JobConf for setting up the dataset. Note: This will be put into a Broadcast.
+   * @note because Hadoop's `RecordReader` class re-uses the same Writable object for each
```
srowen: Do you need to move this? you don't need to indent it

srowen (changes requested):
Again I didn't fully review yet; see another round of comments here.

srowen:
Ping @Mironor are you able to pare this down further? I want to figure out whether this will proceed or not. It needs a rebase anyhow.

Mironor:
@srowen I see that my rebase didn't go well, sorry for that. Is there a way to revert the last push or it would be better to create another pull request and close this one?

srowen:
Up to you but it's fine to just start from scratch in a new PR. Sometimes that's simpler, esp. because it will probably be fairly different anyway.

Mironor:
@srowen I fixed everything. All comments are squashed, please review the current one.

core/src/main/scala/org/apache/spark/SparkContext.scala
```diff
@@ -276,6 +276,9 @@ class SparkContext(config: SparkConf) extends Logging {
 
   private[spark] def ui: Option[SparkUI] = _ui
 
+  /**
+   * @return `Option` containing URL to web UI.
```
srowen: This kind of doc doesn't add value, because it doesn't add anything the method signature doesn't already say. I'd remove it or expand it. It's picky, but I think we did assert this before.

core/src/main/scala/org/apache/spark/SparkContext.scala
```diff
@@ -486,9 +491,11 @@ class SparkContext(config: SparkConf) extends Logging {
       value <- Option(System.getenv(envKey)).orElse(Option(System.getProperty(propKey)))} {
       executorEnvs(envKey) = value
     }
+
```
srowen: Likewise whitespace changes just aren't important and have a tiny cost (merge conflicts)

core/src/main/scala/org/apache/spark/SparkContext.scala
```diff
@@ -705,13 +712,17 @@ class SparkContext(config: SparkConf) extends Logging {
...
+  /**
+   * Distribute a local Scala collection to form an RDD.
    *
    * @note Parallelize acts lazily. If `seq` is a mutable collection and is altered after the call
    * to parallelize and before the first action on the RDD, the resultant RDD will reflect the
    * modified collection. Pass a copy of the argument to avoid this.
    * @note avoid using `parallelize(Seq())` to create an empty `RDD`. Consider `emptyRDD` for an
    * RDD with no partitions, or `parallelize(Seq[T]())` for an RDD of `T` with empty partitions.
+   * @param seq Scala collection to distribute
+   * @param numSlices how many sub-collections should it be sliced into
```
srowen: numSlices is a legacy name. Although the name must be kept its doc should refer to "partitions". How about "number of partitions to divide collection into"

core/src/main/scala/org/apache/spark/SparkContext.scala
```diff
@@ -721,16 +732,15 @@ class SparkContext(config: SparkConf) extends Logging {
...
    * `step` every element.
    *
    * @note if we need to cache this RDD, we should make sure each partition does not exceed limit.
-   *
    * @param start the start value.
    * @param end the end value.
    * @param step the incremental step
    * @param numSlices the partition number of the new RDD.
-   * @return
+   * @return RDD containing distributed range
```
srowen: Nit: RDDs are really pointers and bookeeping. The "represent" rather than "contain"

core/src/main/scala/org/apache/spark/SparkContext.scala
```diff
@@ -816,6 +833,10 @@ class SparkContext(config: SparkConf) extends Logging {
   /**
    * Read a text file from HDFS, a local file system (available on all nodes), or any
    * Hadoop-supported file system URI, and return it as an RDD of Strings.
+   *
+   * @param path URI to Hadoop-supported file system
```
srowen: Well, it's a URI representing a path on a supported file system. But the latter fact was already in the docs. The meaningful doc here is that it's the path of the text file!

core/src/main/scala/org/apache/spark/SparkContext.scala
```diff
@@ -816,6 +833,10 @@ class SparkContext(config: SparkConf) extends Logging {
...
+   * @param minPartitions suggested value of the minimal splitting number for input data
```
srowen: I think the doc string is hard to understand. "Suggested minimum number of partitions for the resulting RDD"?

core/src/main/scala/org/apache/spark/SparkContext.scala
```diff
@@ -956,24 +976,24 @@ class SparkContext(config: SparkConf) extends Logging {
...
    * @param keyClass Class of the keys
    * @param valueClass Class of the values
-   * @param minPartitions Minimum number of Hadoop Splits to generate.
-   *
-   * @note Because Hadoop's RecordReader class re-uses the same Writable object for each
-   * record, directly caching the returned RDD or directly passing it to an aggregation or shuffle
-   * operation will create many references to the same object.
-   * If you plan to directly cache, sort, or aggregate Hadoop writable objects, you should first
-   * copy them using a `map` function.
+   * @param minPartitions minimum number of Hadoop Splits to generate.
```
srowen: Say "partitions" not "splits" despite what the existing string says

core/src/main/scala/org/apache/spark/SparkContext.scala
```diff
@@ -956,24 +976,24 @@ class SparkContext(config: SparkConf) extends Logging {
...
+   * should first copy them using a `map` function.
+   * @param conf `JobConf` for setting up the dataset. Note: This will be put into a Broadcast.
    *             Therefore if you plan to reuse this conf to create multiple RDDs, you need to make
    *             sure you won't modify the conf. A safe approach is always creating a new conf for
    *             a new RDD.
    * @param inputFormatClass Class of the InputFormat
    * @param keyClass Class of the keys
    * @param valueClass Class of the values
-   * @param minPartitions Minimum number of Hadoop Splits to generate.
-   *
```
srowen: As before, don't move this please

srowen (changes requested):
I stopped about halfway through. I don't think most of my comment were addressed. This will be easier if you try to respond to each of them and apply them across your change at each round. Otherwise, for a relatively minor change this takes a lot of bandwidth.

core/src/main/scala/org/apache/spark/SparkContext.scala
Mironor: You said:
> url -> URL and ui -> UI, but, I don't this comment adds much to the reader beyond what the signature says already. I'd improve or remove it.

I thought you were talking about capitalizing as an improvement. I will remove these changes.

core/src/main/scala/org/apache/spark/SparkContext.scala
Mironor: This was my answer to your original comment about moving this:
> Following 2 functions have @note above the parameter description, that was the reason why I moved it. Its content is directly related to the returned value, so it may be justified to move it after the @return statement, but in this case the doc of the following 2 functions could be modified as well.

As I understand, you mean that this is not worth moving because of the possible merge conflicts, right?

srowen: I think it won't affect where it actually renders in the docs, and so yes I think it should be left where it is

Mironor:
@srowen 
> I don't think most of my comment were addressed. This will be easier if you try to respond to each of them and apply them across your change at each round. 

I genuinely would like to know which ones. I removed obvious javadocs (with the exception of `uiWebUrl` because I misunderstood you), indentation, upper-casing and applied suggested modifications to all the comments in this PR. 

I also don't know if I should reply to everyone fo your comments or if it's just ok to apply the modification (so that the comment would be "collapsed" after the push).

> Otherwise, for a relatively minor change this takes a lot of bandwidth.

I could remove the difference below a certain line to make this PR more lightweight.

srowen:
Yes I'm referring to 
https://github.com/apache/spark/pull/16137#discussion_r90948858

and to some extent these issues remain in some form, though it's a judgment call:
https://github.com/apache/spark/pull/16137#discussion_r90948903
https://github.com/apache/spark/pull/16137#discussion_r91170210

Given the sense of the feedback I think you can probably extrapolate what of this is clearly useful and what probably isn't worth it. If you can take your best shot at applying that to the whole change then I'll have another look.

No, don't split up the PR. I just meant that it wouldn't be useful to make 20 variations of the same comments if you can just run with them and apply to the rest first.

Mironor:
Right, I see what you are referring to. I will apply those ones to a greater degree than before.

Mironor:
I removed all trivial (unhelpful) additions and expended javadoc where it didn't add any value to the method signature. I payed special attention to all your previous comments and reflected them in the whole file.

The actual diff was reduced from the initial 400+ lines to 150. This is my best shot at this PR.

core/src/main/scala/org/apache/spark/SparkContext.scala
```diff
@@ -1949,20 +2038,35 @@ class SparkContext(config: SparkConf) extends Logging {
...
+   * a result from one partition)
    */
   def runJob[T, U: ClassTag](rdd: RDD[T], func: (TaskContext, Iterator[T]) => U): Array[U] = {
     runJob(rdd, func, 0 until rdd.partitions.length)
   }
 
   /**
-   * Run a job on all partitions in an RDD and return the results in an array.
+   * Run a job on all partitions in an RDD and return the results in an array, but take a function
+   * of type `Iterator[T] => U` instead of `(TaskContext, Iterator[T]) => U`.
```
Mironor: Two `runJob` methods had exactly the same javadoc, I added this line (similar to the already existing line 2020) to show the difference between them.

core/src/main/scala/org/apache/spark/SparkContext.scala
```diff
@@ -1038,6 +1059,13 @@ class SparkContext(config: SparkConf) extends Logging {
    * operation will create many references to the same object.
    * If you plan to directly cache, sort, or aggregate Hadoop writable objects, you should first
    * copy them using a `map` function.
+   * @param path directory to the input data files, the path can be comma separated paths
+   * as a list of inputs
+   * @param minPartitions suggested minimum number of partitions for the resulting RDD
+   * @param km implicitly found Class of the key associated with the `fm` implicit parameter
```
srowen: Interesting, are these javadoc-able? a caller wouldn't really specify them explicitly. Does javadoc succeed when you build with this change? I don't know if these docs help much; they just say it's some implicit thing associated to another value. I wouldn't add docs just for completeness.

core/src/main/scala/org/apache/spark/SparkContext.scala
```diff
@@ -816,6 +824,9 @@ class SparkContext(config: SparkConf) extends Logging {
   /**
    * Read a text file from HDFS, a local file system (available on all nodes), or any
    * Hadoop-supported file system URI, and return it as an RDD of Strings.
+   * @param path path to the text file on a supported file system
+   * @param minPartitions suggested minimum number of partitions for the resulting RDD
+   * @return RDD representing the content of the file
```
srowen: I still might lean towards saying more, if saying anything at all here: it's an RDD of lines of the text file. Many of these comments are OK but minimal.

core/src/main/scala/org/apache/spark/SparkContext.scala
srowen: That's pretty OK, though the type info is already documented. The prose version of the difference is more like: this one's arg doesn't take a `TaskContext` and the other one does.

core/src/main/scala/org/apache/spark/SparkContext.scala
```diff
@@ -2092,15 +2215,17 @@ class SparkContext(config: SparkConf) extends Logging {
    * @param checkSerializable whether or not to immediately check <tt>f</tt> for serializability
    * @throws SparkException if <tt>checkSerializable</tt> is set but <tt>f</tt> is not
    *   serializable
+   * @return reference cleaned from closures
```
srowen: Might just say `@return the cleaned closure`; "cleaned from closures" doesn't really scan

core/src/main/scala/org/apache/spark/SparkContext.scala
```diff
@@ -2307,6 +2432,8 @@ object SparkContext extends Logging {
    *
    * @note This function cannot be used to create multiple SparkContext instances
    * even if multiple contexts are allowed.
+   * @param config Spark's config that will be used for initialisation of the `SparkContext`
```
srowen: `SparkConf` instead of `Spark's config` ? just because elsewhere you're referring to concepts by class name in this line.

core/src/main/scala/org/apache/spark/SparkContext.scala
```diff
@@ -2412,6 +2540,9 @@ object SparkContext extends Logging {
   /**
    * Find the JAR from which a given class was loaded, to make it easy for users to pass
    * their JARs to SparkContext.
+   *
+   * @param cls Class that should be inside of the jar
```
srowen: Nit, I don't even feel strongly about it: "class" isn't capitalized, except where used as the class name `Class`. Most uses in your new docs are just the noun, not class name. So you might just write "class", or else back-tick it where it can be referred to as a class name.

srowen (changes requested):
Yes, looking pretty good.

core/src/main/scala/org/apache/spark/SparkContext.scala
Mironor: I ran `build/sbt unidoc` as described [here](https://github.com/apache/spark/blob/master/docs/README.md)
I got this output:
```
[info] 100 warnings
[info] Main Java API documentation successful.
```
Searching `/SparkContext.scala` through warnings didn't yield any results. So I guess it's fine.

This is how it looks in Scaladoc: [https://i.imgsafe.org/ea9b9e32bb.png](https://i.imgsafe.org/ea9b9e32bb.png)

Scala's Scaladocs [do contain](http://www.scala-lang.org/api/current/scala/collection/immutable/List.html#sorted[B>:A](implicitord:scala.math.Ordering[B]):Repr) implicit parameters descriptions, otherwise it would be hard to understand method's signature.

I hope this answers your questions, it's up to you to tell me if you want them removed or not.

srowen: The thing is that callers won't be specifying these. They're just an implementation detail related to various generic types. If anything, it's the generic types that should be documented, though they're kind of obvious.

These things aren't Classes, but ClassTags. fm is the parameter; it's not something related to fm. Really, fm is related to F.

If this is all we can say about these types I'd remove them, because I'm not sure anyone who doesn't already understand implicit ClassTags is going to be helped by this.

Mironor:
I addressed the feedback about the things that are left to modify. Thank you for taking your time to review this PR, I learnt a lot in terms of your expectations of what is useful and what isn't in a pull request.

core/src/main/scala/org/apache/spark/SparkContext.scala
```diff
@@ -1973,7 +2067,12 @@ class SparkContext(config: SparkConf) extends Logging {
   }
 
   /**
-   * Run a job on all partitions in an RDD and pass the results to a handler function.
+   * Run a job on all partitions in an RDD and pass the results to a handler function, but take
+   * a function of type `Iterator[T] => U` instead of `(TaskContext, Iterator[T]) => U`.
```
srowen: I think this one should change too -- instead of talking about the types (which are already in the signature) refer to the salient difference here, that this one doesn't take a function that takes a TaskContext.

srowen (commented):
One last comment then I think this is ready

core/src/main/scala/org/apache/spark/SparkContext.scala
```diff
@@ -2067,8 +2067,8 @@ class SparkContext(config: SparkConf) extends Logging {
   }
 
   /**
-   * Run a job on all partitions in an RDD and pass the results to a handler function, but take
-   * a function of type `Iterator[T] => U` instead of `(TaskContext, Iterator[T]) => U`.
+   * Run a job on all partitions in an RDD and pass the results to a handler function, this time the
```
srowen: I wasn't specific enough here ... I don't think this is the best way to write it. Both functions do X. One lets you do X with an additional argument. Both should say they do X. The one with the additional argument should mention its additional argument. Saying "the other one doesn't have this arg" seems reversed. I'm not that against merging it but because the point is to improve docs we should probably take extra care to make the wording clear.

Mironor: If I understood you correctly, both of the functions should have `X` as the description, but the one with `TaskContext` has this added to its description: `The function that is run against each partition should additionally take TaskContext argument.`, right? 
At the same time the actual argument description stays the same in both cases because the `TaskContext` thing is already described in the method signature.

srowen: Sure, sounds more natural to me. Instead of "should", say "... each partition additionally takes ..." because it's not optional.

core/src/main/scala/org/apache/spark/SparkContext.scala
```diff
@@ -1062,13 +1087,32 @@ class SparkContext(config: SparkConf) extends Logging {
...
+   * a list of inputs
+   * @return RDD of tuples of key and corresponding value
    */
   def hadoopFile[K, V, F <: InputFormat[K, V]](path: String)
       (implicit km: ClassTag[K], vm: ClassTag[V], fm: ClassTag[F]): RDD[(K, V)] = withScope {
     hadoopFile[K, V, F](path, defaultMinPartitions)
   }
 
-  /** Get an RDD for a Hadoop file with an arbitrary new API InputFormat. */
+  /**
```
Mironor: This one completely lacked documentation (I spotted it while going through the generated javadoc) in-between similar non-trivial public API that had documentation already set up.

srowen (approved):
OK, looks like a good set of elaborations of the docs

srowen:
Merged to master

PR #16138, opened 2016-12-04
[SPARK-16609] Add to_date/to_timestamp with format functions

bllchmbrs:

This pull request adds two new user facing functions:
- `to_date` which accepts an expression and a format and returns a date.
- `to_timestamp` which accepts an expression and a format and returns a timestamp.

For example, Given a date in format: `2016-21-05`. (YYYY-dd-MM)

### Date Function
*Previously*
```
to_date(unix_timestamp(lit("2016-21-05"), "yyyy-dd-MM").cast("timestamp"))
```
*Current*
```
to_date(lit("2016-21-05"), "yyyy-dd-MM")
```

### Timestamp Function
*Previously*
```
unix_timestamp(lit("2016-21-05"), "yyyy-dd-MM").cast("timestamp")
```
*Current*
```
to_timestamp(lit("2016-21-05"), "yyyy-dd-MM")
```
### Tasks

- [X] Add `to_date` to Scala Functions
- [x] Add `to_date` to Python Functions
- [x] Add `to_date` to SQL Functions
- [X] Add `to_timestamp` to Scala Functions
- [x] Add `to_timestamp` to Python Functions
- [x] Add `to_timestamp` to SQL Functions
- [x] Add function to R

## How was this patch tested?

- [x] Add Functions to `DateFunctionsSuite`
- Test new `ParseToTimestamp` Expression (*not necessary*)
- Test new `ParseToDate` Expression (*not necessary*)
- [x] Add test for R
- [x] Add test for Python in test.py


sql/core/src/main/scala/org/apache/spark/sql/functions.scala
```diff
@@ -2666,7 +2666,18 @@ object functions {
    * @group datetime_funcs
    * @since 1.5.0
    */
-  def to_date(e: Column): Column = withExpr { ToDate(e.expr) }
+  def to_date(e: Column): Column = withExpr { ToDate(e.expr, Literal("yyyy-MM-dd")) }
```
anabranch: this maintains previous capabilities.

sql/core/src/main/scala/org/apache/spark/sql/functions.scala
```diff
@@ -2661,12 +2661,31 @@ object functions {
...
+
+  /**
+   * Converts the column into a DateType with a specified format.
+   * (see [http://docs.oracle.com/javase/tutorial/i18n/format/simpleDateFormat.html])
+   *
+   * @group datetime_funcs
+   * @since 2.2.0
```
anabranch: Will need to make sure this version is correct.

sql/core/src/main/scala/org/apache/spark/sql/functions.scala
```diff
@@ -2661,12 +2661,30 @@ object functions {
   def unix_timestamp(s: Column, p: String): Column = withExpr {UnixTimestamp(s.expr, Literal(p)) }
 
   /**
+    * Convert time string with given pattern
+    * (see [http://docs.oracle.com/javase/tutorial/i18n/format/simpleDateFormat.html])
+    * to Unix time stamp (in seconds), return null if fail.
+    * @group datetime_funcs
+    * @since 1.5.0
```
anabranch: will need to change this version to the correct branch.

rxin:
Can you use "SPARK-16609" instead of "Spark-16609"?

sql/core/src/main/scala/org/apache/spark/sql/functions.scala
```diff
@@ -2661,12 +2661,30 @@ object functions {
...
+    */
+  def to_timestamp(s: Column, p: String): Column = withExpr {UnixTimestamp(s.expr, Literal(p)) }
```
anabranch: still missing a cast here...

sql/core/src/test/scala/org/apache/spark/sql/DateFunctionsSuite.scala
```diff
@@ -351,34 +351,81 @@ class DateFunctionsSuite extends QueryTest with SharedSQLContext {
...
+      df.select(to_date(col("t"), "yyyy-MM-dd")),
+      Seq(Row(Date.valueOf("2015-07-22")), Row(Date.valueOf("2014-12-31")),
+        Row(Date.valueOf("2014-12-31"))))
+    checkAnswer(
+      df.select(to_date(col("d"), "yyyy-MM-dd")),
+      Seq(Row(Date.valueOf("2015-07-22")), Row(Date.valueOf("2015-07-01")),
+        Row(Date.valueOf("2014-12-31"))))
+    checkAnswer(
+      df.select(to_date(col("s"), "yyyy-MM-dd")),
+      Seq(Row(Date.valueOf("2015-07-22")), Row(Date.valueOf("2014-12-31")), Row(null)))
```
anabranch:
```
scala> val format = new java.text.SimpleDateFormat("yyyy-MM-dd")
format: java.text.SimpleDateFormat = java.text.SimpleDateFormat@f67a0200

scala> format.parse("2014-31-12")
res1: java.util.Date = Tue Jul 12 00:00:00 PDT 2016
```

This is strange to me but is expected behavior.

anabranch: nevermind, this does fail on here too. Seems like I was just looking at the wrong result.

marmbrus:
This will be very convenient!  Looking forward to the whole patch.  For SQL I think you should look at [`RuntimeReplaceable`](https://github.com/apache/spark/blob/fd90541c35af2bccf0155467bec8cea7c8865046/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/Expression.scala#L238).  Also, we might consider having a default version that doen't take a pattern (and parses something standard such as `2016-12-01T02:03:01Z`)

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/datetimeExpressions.scala
```diff
@@ -1047,6 +1047,53 @@ case class ToDate(child: Expression) extends UnaryExpression with ImplicitCastIn
...
+  extended = """
+    Examples:
+      > SELECT _FUNC_('2016-12-31', 'yyyy-MM-dd');
+       2016-12-31 00:00:00.0
+  """)
+case class ParseToTimestamp(left: Expression, right: Expression, child: Expression)
+  extends RuntimeReplaceable with ImplicitCastInputTypes {
+
+  def this(left: Expression, format: Expression) = {
+  this(left, format, Cast(UnixTimestamp(left, format), TimestampType))
```
anabranch: For reasons unbeknownst to me, Cast(...., TimestampType) is not equivalent to ....cast("timestamp")

anabranch: This is returning the date in millis (1970) as opposed to seconds. But if you use the `UnixTimestamp` expression via the function, then cast, it will work just fine.

anabranch:
```
scala> spark.range(1).select(to_date(to_timestamp(lit("2016-12-31"), "yyyy-MM-dd"))).explain(true)
== Parsed Logical Plan ==
'Project [to_date(parsetotimestamp(2016-12-31, yyyy-MM-dd)) AS to_date(parsetotimestamp('2016-12-31', 'yyyy-MM-dd'))#48]
+- Range (0, 1, step=1, splits=Some(8))

== Analyzed Logical Plan ==
to_date(parsetotimestamp('2016-12-31', 'yyyy-MM-dd')): date
Project [to_date(cast(parsetotimestamp(2016-12-31, yyyy-MM-dd) as date)) AS to_date(parsetotimestamp('2016-12-31', 'yyyy-MM-dd'))#48]
+- Range (0, 1, step=1, splits=Some(8))

== Optimized Logical Plan ==
Project [17166 AS to_date(parsetotimestamp('2016-12-31', 'yyyy-MM-dd'))#48]
+- Range (0, 1, step=1, splits=Some(8))

== Physical Plan ==
*Project [17166 AS to_date(parsetotimestamp('2016-12-31', 'yyyy-MM-dd'))#48]
+- *Range (0, 1, step=1, splits=Some(8))
```

Calling `to_date` on the result of my function works (that's before I do that cast).

However
```
scala> spark.range(1).select(to_timestamp(lit("2016-12-31"), "yyyy-MM-dd")).explain(true)
== Parsed Logical Plan ==
'Project [parsetotimestamp(2016-12-31, yyyy-MM-dd) AS parsetotimestamp('2016-12-31', 'yyyy-MM-dd')#34]
+- Range (0, 1, step=1, splits=Some(8))

== Analyzed Logical Plan ==
parsetotimestamp('2016-12-31', 'yyyy-MM-dd'): timestamp
Project [parsetotimestamp(2016-12-31, yyyy-MM-dd) AS parsetotimestamp('2016-12-31', 'yyyy-MM-dd')#34]
+- Range (0, 1, step=1, splits=Some(8))

== Optimized Logical Plan ==
Project [2016-12-31 00:00:00 AS parsetotimestamp('2016-12-31', 'yyyy-MM-dd')#34]
+- Range (0, 1, step=1, splits=Some(8))

== Physical Plan ==
*Project [2016-12-31 00:00:00 AS parsetotimestamp('2016-12-31', 'yyyy-MM-dd')#34]
+- *Range (0, 1, step=1, splits=Some(8))
```

Doesn't return the correct result. It appears to be in microseconds.

anabranch: Now if I remove the cast call and turn this function into:

```scala
  def this(left: Expression, format: Expression) = {
  this(left, format, UnixTimestamp(left, format))
}
```

I will end up with
```scala
scala> import org.apache.spark.sql.functions._
import org.apache.spark.sql.functions._

scala>

scala> //////////// TIME INFO

scala> spark.range(1).select(to_timestamp(lit("2016-12-31"), "yyyy-MM-dd")).collect()
res0: Array[org.apache.spark.sql.Row] = Array([1970-01-01 11:05:19.476746])

scala> // expected is what the output of the below command

scala> spark.range(1).select(to_timestamp(lit("2016-12-31"), "yyyy-MM-dd").cast("timestamp")).collect()
res1: Array[org.apache.spark.sql.Row] = Array([null])

scala>

scala>

scala> //////////// DATE INFO

scala> spark.range(1).select(to_date(lit("2016-12-31"), "yyyy-MM-dd")).collect()
res2: Array[org.apache.spark.sql.Row] = Array([null])

scala> // expected is the output of the below command

scala> spark.range(1).select(to_date(to_timestamp(lit("2016-12-31"), "yyyy-MM-dd"))).collect()
res3: Array[org.apache.spark.sql.Row] = Array([null])
```

anabranch: Where i am stuck is, there seems to be some implicit converting between types somewhere that I don't understand. The two casts are not completely equivalent for reasons beyond my understanding.

anabranch:
and I found the error, I shouldn't be overriding the `DataType`.

anabranch:
Now that my outputs are correct (in format), there's a new problem. The types are *still* wrong.

```
scala> /////////// DETAILS

scala> // Schema

scala> spark.range(1).select(to_timestamp(lit("2016-12-31"), "yyyy-MM-dd")).schema
// res4: org.apache.spark.sql.types.StructType = StructType(StructField(to_timestamp('2016-12-31', 'yyyy-MM-dd'),StringType,true))

scala> spark.range(1).select(to_date(lit("2016-12-31"), "yyyy-MM-dd")).schema
// res5: org.apache.spark.sql.types.StructType = StructType(StructField(to_date('2016-12-31', 'yyyy-MM-dd'),StringType,true))
```

Those should be `DateType` and `TimestampType` respectively.

However if we inspect the expression directly, it seems that we do have the correct type.

```
scala> import org.apache.spark.sql.catalyst.expressions.ParseToDate
scala> val x = new ParseToDate(lit("hello").expr, lit("world").expr)
x: org.apache.spark.sql.catalyst.expressions.ParseToDate = to_date(hello, world)

scala> x.dataType
res9: org.apache.spark.sql.types.DataType = DateType
```

It seems to be getting lost somewhere.

anabranch:
More details are here: https://gist.github.com/anabranch/7a42292593976878eb14e2d86a9966d4

This is completely perplexing to me.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/datetimeExpressions.scala
```diff
@@ -1047,6 +1047,57 @@ case class ToDate(child: Expression) extends UnaryExpression with ImplicitCastIn
...
+}
+
+  override def inputTypes: Seq[AbstractDataType] = Seq(StringType, StringType)
+  override def flatArguments: Iterator[Any] = Iterator(left, right)
+  override def sql: String = s"$prettyName(${left.sql}, ${right.sql})"
+
+  override def prettyName: String = "to_timestamp"
+  override def dataType: DataType = TimestampType
```
anabranch: If I remove this `DataType` override I will end up with the right answer but the wrong type (string).

If I have this `DataType` override I will end up with the wrong answer but the right type.

If I cast to the expected type by either having the `dataType` override or removing it, I will get the expected output. But I have to cast after performing `to_timestamp` or `to_date`. The cast within the expression does not seem to be respected.

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/DateExpressionsSuite.scala
```diff
@@ -389,6 +389,20 @@ class DateExpressionsSuite extends SparkFunSuite with ExpressionEvalHelper {
     checkConsistencyBetweenInterpretedAndCodegen(ToDate, DateType)
   }
 
+  test("function parse_to_date") {
```
anabranch: I'm confused as to why this is failing. I need some guidance on setting these tests up.

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/DateExpressionsSuite.scala
```diff
@@ -389,6 +389,20 @@ class DateExpressionsSuite extends SparkFunSuite with ExpressionEvalHelper {
...
+    def testDateParse(input: String, format: String, output: String): Unit = {
+      val parsed = new ParseToDate(Literal(input), Literal(format))
+      checkEvaluation(parsed, DateTimeUtils.fromJavaDate(Date.valueOf((output))), null)
+    }
+    // This fails but I am not sure why, it doesn't really make sense that it would
+    // simply because it's the right answer. I don't understand CheckEvaluation.
+    testDateParse("2015-07-15", "yyyy-MM-dd", "2015-07-15")
+  }
+
+  test("function parse_to_timestamp") {
```
anabranch: same here obviously

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/FunctionRegistry.scala
```diff
@@ -342,7 +342,8 @@ object FunctionRegistry {
     expression[CurrentTimestamp]("now"),
     expression[Quarter]("quarter"),
     expression[Second]("second"),
-    expression[ToDate]("to_date"),
+    expression[ParseToTimestamp]("to_timestamp"),
+    expression[ParseToDate]("to_date"),
```
anabranch: I think this is ok to do because I just remapped it, I'm not sure how to register multiple functions with the same signature for SQL.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/datetimeExpressions.scala
```diff
@@ -1047,6 +1048,60 @@ case class ToDate(child: Expression) extends UnaryExpression with ImplicitCastIn
...
+  """)
+case class ParseToDate(left: Expression, format: Expression, child: Expression)
+  extends RuntimeReplaceable {
+
+  def this(left: Expression, format: Expression) = {
+    this(left, format, Cast(new ParseToTimestamp(left, format), DateType))
+  }
+
+  def this(left: Expression) = {
+    this(left, Literal(""), ToDate(left))
```
anabranch: so this second value gets ignored completely. Not sure exactly how to handle this.

python/pyspark/sql/functions.py
```diff
@@ -143,6 +143,12 @@ def _():
                'measured in radians.',
 }
 
+_functions_2_2 = {
```
anabranch: I want to double check all these docs, but I tried the functions and they work!

bllchmbrs (commented):
Think I am ready for a review here @hvanhovell

anabranch:
I believe now why my previous implementation did not work.

My implementation originally looked like this:

```scala
case class ParseToTimestamp(left: Expression, format: Expression, child: Expression)
  extends RuntimeReplaceable {

  def this(left: Expression, format: Expression) = {
  this(left, format, Cast(UnixTimestamp(left, format), TimestampType))
}

  override def checkInputDataTypes(): TypeCheckResult = {
    if (left.dataType != StringType) {
      TypeCheckResult.TypeCheckFailure(s"TO_TIMESTAMP requires both inputs to be strings")
    }
    TypeCheckResult.TypeCheckSuccess
  }

  override def flatArguments: Iterator[Any] = Iterator(left, format)
  override def sql: String = s"$prettyName(${left.sql}, ${format.sql})"

  override def prettyName: String = "to_timestamp"
  override def dataType: DataType = TimestampType
}
```

This implementation with a simple example would fail.

```scala
import org.apache.spark.sql.functions._

val ss1 = "2015-07-24 10:00:00"
val ss2 = "2015-07-25 02:02:02"
val df2 = Seq((ss1), (ss2)).toDF("ss")

df2.select(to_timestamp(col("ss"))).show
```
This throws a 

```
org.apache.spark.sql.catalyst.analysis.UnresolvedException: Invalid call to dataType on unresolved object, tree: 'ss
```

A `Trace` log level shows that the columns are resolved however the error originates after analysis and during `CheckInputDataTypes`. This function seeks to analyze the `left` input, however because this column is input into a `RuntimeReplaceable` function the relevant, and resolved, tree is actually the `child` argument - `left` remains unresolved (and therefore throws the above error).

I believe this to be the root cause and that has in turn showed me that I do not need to perform input validation for this function in the first place. Since I only wrap functions, they are performing the exact same input validation that I would be. Since no new logic is implemented, there's no point in redundantly validating something that will be validated again anyways, especially when the system won't let me.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/datetimeExpressions.scala
```diff
@@ -1047,6 +1048,60 @@ case class ToDate(child: Expression) extends UnaryExpression with ImplicitCastIn
...
+  extended = """
+    Examples:
+      > SELECT _FUNC_('2016-12-31', 'yyyy-MM-dd');
+       2016-12-31
+  """)
+case class ParseToDate(left: Expression, format: Expression, child: Expression)
+  extends RuntimeReplaceable {
+
+  def this(left: Expression, format: Expression) = {
+    this(left, format, Cast(new ParseToTimestamp(left, format), DateType))
```
anabranch: Change this to `UnixTimestamp` instead of ParseToTimestamp

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/DateExpressionsSuite.scala
hvanhovell: This is not needed since the functions are `RuntimeReplaceable`.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/datetimeExpressions.scala
```diff
@@ -1047,6 +1048,60 @@ case class ToDate(child: Expression) extends UnaryExpression with ImplicitCastIn
...
+  override def prettyName: String = "to_date"
+  override def dataType: DataType = DateType
+}
+
+/**
+ * Parses a column to a timestamp based on the supplied format.
+ */
+// scalastyle:off line.size.limit
+@ExpressionDescription(
+  usage = "_FUNC_(timestamp, fmt) - Parses the `left` expression with the `format` expression to a timestamp.",
```
hvanhovell: Could you mention that this function will return null with the input is invalid.

hvanhovell (changes requested):
This looks pretty good. Could you add a function to R, and a simple test to python.

anabranch:
@felixcheung could you give me some pointers on these R functions? I don't quite know if I am registering them correctly and they're failing my builds.

felixcheung:
Could you make sure your code doc is starting with `#'` instead of `#`? It's not obvious what the error is but it might be related to parsing / alignment.

anabranch:
@felixcheung Just tried that it doesn't seem to work.

Here's the strange thing, it should follow the *exact* same structure as [unix_timestamp](https://github.com/apache/spark/blob/master/R/pkg/R/functions.R#L3015) except that I don't have a signature that takes 0 parameters (i.e., [unix_timestamp L3031](https://github.com/apache/spark/blob/master/R/pkg/R/functions.R#L3031). 

Now if I copy and paste the `unix_timestamp` section and change the names, it **will compile**. However as soon as I remove the first function (like L3031, that takes no parameters). It will fail again with the same error.

I also tried removing everything and just creating one function at a time.

just this code will work.
```
#' to_date
#'
#' Converts the column into DateType, optionally with a format. Returns null if input is invalid.
#' Specify formats according to: http://docs.oracle.com/javase/tutorial/i18n/format/simpleDateFormat.html
#'
#' @param x Column to parse.
#' @param format string to use to parse x Column to DateType.
#'
#' @rdname to_date
#' @name to_date
#' @family datetime_funcs
#' @aliases to_date,Column-method
#' @export
#' @examples
#' \dontrun{
#' to_date(df$c, 'yyyy-MM-dd')
#' }
#' @note to_date(Column, format) since 2.2.0
setMethod("to_date",
          signature(x = "Column"),
          function(x) {
            jc <- callJStatic("org.apache.spark.sql.functions", "to_date", x@jc)
            column(jc)
          })
```

However if I add *any* other parameters to that method (or signature). It will fail.

felixcheung:
you would need to update the signature here https://github.com/apache/spark/blob/master/R/pkg/R/generics.R#L1255
this affects the signature matching

anabranch:
@felixcheung yup! Thanks!

R/pkg/R/functions.R
```diff
@@ -1730,24 +1730,82 @@ setMethod("toRadians",
...
+          signature(x = "Column", format = "missing"),
+          function(x, format) {
             jc <- callJStatic("org.apache.spark.sql.functions", "to_date", x@jc)
             column(jc)
           })
 
+#' @rdname to_date
+#' @name to_date
+#' @family datetime_funcs
+#' @aliases to_date,Column-method,character-method
```
felixcheung: this should be `#' @aliases to_date,Column,character-method`

R/pkg/R/functions.R
```diff
@@ -1730,24 +1730,82 @@ setMethod("toRadians",
...
+signature(x = "Column", format = "missing"),
+function(x, format) {
+    jc <- callJStatic("org.apache.spark.sql.functions", "to_timestamp", x@jc)
+    column(jc)
+})
+
+#' @rdname to_timestamp
+#' @name to_timestamp
+#' @family datetime_funcs
+#' @aliases to_timestamp,Column-method,character-method
```
felixcheung: this should be `#' @aliases to_timestamp,Column,character-method`

R/pkg/R/generics.R
```diff
@@ -1274,6 +1270,14 @@ setGeneric("unbase64", function(x) { standardGeneric("unbase64") })
 #' @export
 setGeneric("unhex", function(x) { standardGeneric("unhex") })
 
+#' @rdname to_date
+#' @export
+setGeneric("to_date", function(x, format) { standardGeneric("to_date") })
+
+#' @rdname to_timestamp
+#' @export
+setGeneric("to_timestamp", function(x, format) { standardGeneric("to_timestamp") })
```
felixcheung: could you keep this sorted by function name please

R/pkg/R/functions.R
```diff
@@ -1730,24 +1730,82 @@ setMethod("toRadians",
 
 #' to_date
 #'
-#' Converts the column into DateType.
+#' Converts the column into DateType, optionally with a format. Returns null if input is invalid.
+#' Specify formats according to: http://docs.oracle.com/javase/tutorial/i18n/format/simpleDateFormat.html
```
felixcheung: put url inside `\url{https://spark.apache.org/docs/latest/sparkr.html#sparkr-dataframes}` or as a separate line `#' @seealso \url{https://spark.apache.org/docs/latest/sparkr.html#sparkr-dataframes}`

R/pkg/R/functions.R
```diff
@@ -1730,24 +1730,82 @@ setMethod("toRadians",
 
 #' to_date
 #'
-#' Converts the column into DateType.
+#' Converts the column into DateType, optionally with a format. Returns null if input is invalid.
```
felixcheung: `Returns null if input is invalid.` - seems like this could confuse things since the sql function we don't actually NULL - would the column contains NULL value as a result? If so we should explain that here. Also, I'm pretty sure null in JVM gets turn into `NA` in R, might want to check for this case.

R/pkg/R/functions.R
```diff
@@ -1730,24 +1730,82 @@ setMethod("toRadians",
...
 #' @aliases to_date,Column-method
 #' @export
-#' @examples \dontrun{to_date(df$c)}
-#' @note to_date since 1.5.0
+#' @examples
+#' \dontrun{
+#' to_date(df$c)
+#' to_date(df$c, 'yyyy-MM-dd')
+#' }
+#' @note to_date(Column, format) since 2.2.0
```
felixcheung: this should be the type, not parameter name; so it should be
`#' @note to_date(Column, character) since 2.2.0`
ditto below

R/pkg/R/functions.R
```diff
@@ -1730,24 +1730,82 @@ setMethod("toRadians",
...
           })
 
+#' @rdname to_date
+#' @name to_date
+#' @family datetime_funcs
+#' @aliases to_date,Column-method,character-method
+#' @export
+#' @note to_date(Column, format) since 2.2.0
+setMethod("to_date",
+signature(x = "Column", format = "character"),
```
felixcheung: could you please align this like L1751, L1753?

R/pkg/R/functions.R
```diff
@@ -1730,24 +1730,82 @@ setMethod("toRadians",
...
+function(x, format) {
+    jc <- callJStatic("org.apache.spark.sql.functions", "to_date", x@jc, format)
+    column(jc)
+})
+
+#' to_timestamp
+#'
+#' Converts the column into DateType, optionally with a format. Returns null if input is invalid.
+#' Specify formats according to: http://docs.oracle.com/javase/tutorial/i18n/format/simpleDateFormat.html
```
felixcheung: ditto on the description

R/pkg/R/functions.R
```diff
@@ -1730,24 +1730,82 @@ setMethod("toRadians",
...
+#' @name to_timestamp
+#' @family datetime_funcs
+#' @aliases to_timestamp,Column-method
+#' @export
+#' @examples
+#' \dontrun{
+#' to_timestamp(df$c)
+#' to_timestamp(df$c, 'yyyy-MM-dd')
+#' }
+#' @note to_timestamp(Column, format) since 2.2.0
```
felixcheung: ditto

R/pkg/R/functions.R
felixcheung: same for to_timestamp

R/pkg/R/functions.R
anabranch: done in next push.

R/pkg/R/functions.R
anabranch: done in next push.

R/pkg/R/generics.R
anabranch: absolutely! done in next push.

R/pkg/R/functions.R
anabranch: Take a look at the next push, I think that this will resolve it. Because this is not an action I don't think it's relevant to define what will happen when the data is collected. That is already documented according to how the types line up with one another.

R/pkg/R/functions.R
anabranch: changed for both relevant function definitions.

R/pkg/R/functions.R
anabranch: done!

R/pkg/R/functions.R
anabranch: cleaned this up.

R/pkg/R/functions.R
anabranch: this too.

anabranch:
@felixcheung Thank you for your feedback! Small request, can you tell me if my R test case is sufficient for this? It doesn't seem like there is extensive R testing right now for virtually any function.

Obviously tests will pass soon, running into strange edge cases.

anabranch:
@cloud-fan - Reynold referred me to your for this test failure. 

My two tests are failing because Hive tests *allegedly* cover something like this.

```
SELECT to_date('2001-10-30 10:30:00', '')
```
However, Hive doesn't support passing in multiple parameters to `to_date` as specified in the [language manual](https://cwiki.apache.org/confluence/display/Hive/LanguageManual+UDF#LanguageManualUDF-DateFunctions). The only instance I see for `to_date` with multiple parameters is [when it talks to ORACLE DB as the metastore](https://github.com/apache/hive/blob/2d813f4d4a0bb42345d153c362f7416f05ab2749/metastore/src/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java#L1122) although I don't know the code base (I grepped for `to_date` and tried to find instances of this occurring).

It seems like this test case should not be running in the first place. Can you advise on any suggestions you might have?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/datetimeExpressions.scala
```diff
@@ -1047,6 +1048,64 @@ case class ToDate(child: Expression) extends UnaryExpression with ImplicitCastIn
...
+// scalastyle:off line.size.limit
+@ExpressionDescription(
+  usage = "_FUNC_(date_str, fmt) - Parses the `left` expression with the `fmt` expression. Returns null with invalid input.",
+  extended = """
+    Examples:
+      > SELECT _FUNC_('2016-12-31', 'yyyy-MM-dd');
+       2016-12-31
+  """)
+// scalastyle:on line.size.limit
+case class ParseToDate(left: Expression, format: Expression, child: Expression)
```
cloud-fan: we don't need to put `child` in constructor, but simply add a `def child` in the class.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/datetimeExpressions.scala
```diff
@@ -1047,6 +1048,64 @@ case class ToDate(child: Expression) extends UnaryExpression with ImplicitCastIn
...
+    // RuntimeReplaceable forces the signature, the second value
+    // is ignored completely
+    this(left, Literal(""), ToDate(left))
+  }
+
+  override def flatArguments: Iterator[Any] = Iterator(left, format)
+  override def sql: String = s"$prettyName(${left.sql}, ${format.sql})"
+
+  override def prettyName: String = "to_date"
+  override def dataType: DataType = DateType
```
cloud-fan: this is already defined in `RuntimeReplaceable`

felixcheung:
@anabranch right, since there is little work done on the R side and this is mostly a pass through, just checking if the functions are callable without error like the test you have added should be fine for R.

The gate for functional correctness would be more with the Scala tests in this case.

cloud-fan:
The failed test is `udf5.q` in `HiveComparisonTest`, but I can't find `SELECT to_date('2001-10-30 10:30:00', '')` in it, can you point out where we write such a test?

R/pkg/R/functions.R
```diff
@@ -1746,7 +1750,7 @@ setMethod("toRadians",
 #' to_date(df$c)
 #' to_date(df$c, 'yyyy-MM-dd')
 #' }
-#' @note to_date(Column, format) since 2.2.0
+#' @note to_date(Column) since 1.5
```
felixcheung: this should revert back to `1.5.0` I believe

R/pkg/inst/tests/testthat/test_sparkSQL.R
```diff
@@ -1177,6 +1177,9 @@ test_that("column functions", {
   c17 <- cov(c, c1) + cov("c", "c1") + covar_samp(c, c1) + covar_samp("c", "c1")
   c18 <- covar_pop(c, c1) + covar_pop("c", "c1")
   c19 <- spark_partition_id()
+  c20 <- to_timestamp(c) + trim(c) + unbase64(c) + unhex(c) + upper(c)
+  c21 <- to_timestamp(c, "yyyy") + trim(c) + unbase64(c) + unhex(c) + upper(c)
+  c22 <- to_date(c, "yyyy") + trim(c) + unbase64(c) + unhex(c) + upper(c)
```
felixcheung: you probably don't need to duplicate `+ trim(c) + unbase64(c) + unhex(c) + upper(c)` - as explained on the other comment, the goal (from what I can see) is to call each of these once.

So you could condense these 3 lines into 
```
c20 <- to_timestamp(c) + to_timestamp(c, "yyyy") + to_date(c, "yyyy")
```

anabranch:
@cloud-fan The error i see is the one in this [test case](https://amplab.cs.berkeley.edu/jenkins/job/SparkPullRequestBuilder/71913/testReport/org.apache.spark.sql.catalyst/ExpressionToSQLSuite/datetime_functions/)

What it states at the top

```
# Converted SQL query string:
SELECT `gen_attr_0` AS `to_date('2001-10-30 10:30:00', '')` FROM (SELECT to_date('2001-10-30 10:30:00', '') AS `gen_attr_0`) AS gen_subquery_0

# Original HiveQL query string:
SELECT to_date('2001-10-30 10:30:00')

# Resolved query plan:
Project [to_date(2001-10-30 10:30:00, ) AS to_date('2001-10-30 10:30:00', '')#13365]
```

I don't really understand the hive failure either.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/datetimeExpressions.scala
```diff
@@ -1047,6 +1048,64 @@ case class ToDate(child: Expression) extends UnaryExpression with ImplicitCastIn
...
+  }
+
+  def this(left: Expression) = {
+    // RuntimeReplaceable forces the signature, the second value
+    // is ignored completely
+    this(left, Literal(""), ToDate(left))
+  }
+
+  override def flatArguments: Iterator[Any] = Iterator(left, format)
+  override def sql: String = s"$prettyName(${left.sql}, ${format.sql})"
```
cloud-fan: This is the problem: you set `format` to empty string at https://github.com/apache/spark/pull/16138/files#diff-b83497f7bc11578a0b63a814a2a30f48R1072 , but this is not safe. Although the `format` will be ignored, we use it to build the sql string of this expression, which maybe something like `to_date('2012-12-12 12:12:12', '')`.

we should make `format` an `Option[String]`, and generate corrected sql string here.

R/pkg/R/functions.R
anabranch: fixed.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/datetimeExpressions.scala
anabranch: I don't really understand this feedback. This is how I saw other `RuntimeReplaceable` expressions created.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/datetimeExpressions.scala
anabranch: Removed.

R/pkg/inst/tests/testthat/test_sparkSQL.R
anabranch: fixed

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/datetimeExpressions.scala
anabranch: Fixed!

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/datetimeExpressions.scala
```diff
@@ -1047,6 +1048,69 @@ case class ToDate(child: Expression) extends UnaryExpression with ImplicitCastIn
...
+// scalastyle:on line.size.limit
+case class ParseToTimestamp(left: Expression, format: Expression, child: Expression)
+  extends RuntimeReplaceable {
+
+  def this(left: Expression, format: Expression) = {
+  this(left, format, Cast(UnixTimestamp(left, format), TimestampType))
+}
+
+  override def flatArguments: Iterator[Any] = Iterator(left, format)
+  override def sql: String = s"$prettyName(${left.sql}, ${format.sql})"
```
cloud-fan: does it have the same issue as `ParseToDate`?

anabranch: No, `to_timestamp` doesn't exist at all in hive, it's a brand new Spark SQL only function. I also can't actually find a failure in the test results. Do you know what actually failed?

hvanhovell:
retest this please

felixcheung:
this is the CRAN check error:
```
* checking for missing documentation entries ... WARNING
Undocumented S4 methods:
  generic 'to_date' and siglist 'Column,missing'
  generic 'to_timestamp' and siglist 'Column,missing'
All user-level objects in a package (including S4 classes and methods)
should have documentation entries.
See the chapter 'Writing R documentation files' in the 'Writing R
Extensions' manual.
* checking for code/documentation mismatches ... OK
* checking Rd \usage sections ... WARNING
Objects in \usage without \alias in documentation object 'to_date':
  '\S4method{to_date}{Column,missing}'

Objects in \usage without \alias in documentation object 'to_timestamp':
  '\S4method{to_timestamp}{Column,missing}'

Functions with \usage entries need to have the appropriate \alias
entries, and all their arguments documented.
The \usage entries must correspond to syntactically valid R code.
See the chapter 'Writing R documentation files' in the 'Writing R
```

you probably change it to `@alias to_date,Column,missing-method`

anabranch:
@felixcheung thanks, made those changes :). Hopefully this will start passing sometime :P

sql/core/src/test/scala/org/apache/spark/sql/DateFunctionsSuite.scala
```diff
@@ -354,31 +354,58 @@ class DateFunctionsSuite extends QueryTest with SharedSQLContext {
...
+        Row(Date.valueOf("2014-12-31"))))
+    checkAnswer(
+      df.select(to_date(col("s"), "yyyy-MM-dd")),
+      Seq(Row(Date.valueOf("2015-07-22")), Row(Date.valueOf("2014-12-31")),
+        Row(Date.valueOf("2016-07-12"))))
+
+    //  now switch format
+    checkAnswer(
+      df.select(to_date(col("s"), "yyyy-dd-MM")),
+      Seq(Row(Date.valueOf("2016-10-07")), Row(Date.valueOf("2016-07-12")),
```
cloud-fan: I'm surprised, converting `2015-07-22 10:00:00` to date with format `yyyy-dd-MM` will result to `2016-10-07`? When we convert `2014-31-12` to date with default format, it returns null.

anabranch:
@cloud-fan great questions.

I thought that was strange too. However this is the **current** behavior as well as Java `SimpleDateFormat`'s behavior. I did not implement that logic.

For example, today:

```scala
spark.sql("""
   SELECT 
   cast(
	   cast(
	   	unix_timestamp('2016-30-12', 'yyyy-MM-dd') 
	   as timestamp) 
   as date)
   """).take(5)
// res4: Array[org.apache.spark.sql.Row] = Array([2018-06-12])
```

Which seems wrong to me, but that's the logic that exists. The issue seems to be that unix_timestamp returns an incorrect value when it should just return null. The source is the behavior of `SimpleDateFormat`.

```scala
scala> import java.text.SimpleDateFormat
// import java.text.SimpleDateFormat

scala> val x = new SimpleDateFormat("yyyy-MM-dd")
// x: java.text.SimpleDateFormat = java.text.SimpleDateFormat@f67a0200

scala> x.parse("2016-30-12")
// res5: java.util.Date = Tue Jun 12 00:00:00 PDT 2018
```

anabranch:
Manual Correctness tests:

Python
```
>>> from pyspark.sql.functions import to_date, to_timestamp, lit
>>> spark.range(1).select(to_date(lit('2016-01-02'),'yyyy-MM-dd').alias("dt")).collect()
# [Row(dt=datetime.date(2016, 1, 2))]
>>> spark.range(1).select(to_timestamp(lit('2016-01-02'),'yyyy-MM-dd').alias("dt")).collect()
# [Row(dt=datetime.datetime(2016, 1, 2, 0, 0))]
```

R
```
> collect(select(people, to_date(lit('2016-01-02'), 'yyyy-MM-dd')))
  to_date('2016-01-02', 'yyyy-MM-dd'
#1                         2016-01-02
#2                         2016-01-02
#3                         2016-01-02
> collect(select(people, to_timestamp(lit('2016-01-02'), 'yyyy-MM-dd')))
  to_timestamp('2016-01-02', 'yyyy-MM-dd')
#1                               2016-01-02
#2                               2016-01-02
#3                               2016-01-02
> select(people, to_timestamp(lit('2016-01-02'), 'yyyy-MM-dd'))
```

cloud-fan:
Since this is an existing behavior, this PR LGTM

hvanhovell:
LGTM - merging master.

YannByron:
@anabranch 

I met a failure case when using `to_date`, like this:
```
 spark-sql> select func(to_date('2017-08-25 12:00:00'));
```
`func` is a custom udf which extends `org.apache.hadoop.hive.ql.udf.generic.GenericUDF`.

And this following exception happened:
```
Error in query: No handler for Hive UDF 'xxx.xxx.func': java.lang.UnsupportedOperationException: Cannot evaluate expression: to_date(2017-08-25 12:00:00, None);
```

I locate this problem. When resolving function `func(to_date('2017-08-25 12:00:00'))`(at this time Spark has already finished the resolving to `func`'s child `to_date('2017-08-25 12:00:00')`),  it will call `func`'s FunctionBuilder that is defined in [HiveSessionCatalog.makeFunctionBuilder](https://github.com/apache/spark/blob/v2.2.0/sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveSessionCatalog.scala)). 
In it, the GenericUDF instance will be forced to check input data type. During this process, All of the child catalyst expression which are the parameters of the subclass of GenericUDF, need to be mapped to `ObjectInspector`(see details: [HiveInspector.toInspector](https://github.com/apache/spark/blob/v2.2.0/sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveInspectors.scala)).  Inside, the child expression need to be evaluated though `eval()` firstly, if expression doesn't match with `Literal`.

```
def toInspector(expr: Expression): ObjectInspector = expr match {
    case Literal(value, XXXType) =>
        //...
    // ideally, we don't test the foldable here(but in optimizer), however, some of the
    // Hive UDF / UDAF requires its argument to be constant objectinspector, we do it eagerly.
    case _ if expr.foldable => toInspector(Literal.create(expr.eval(), expr.dataType))
}
```

In my case, the child expression is `to_date('2017-08-25 12:00:00')` that is the instance of [ParseToDate](https://github.com/apache/spark/blob/v2.2.0/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/datetimeExpressions.scala) (extends [RuntimeReplaceable](https://github.com/apache/spark/blob/v2.2.0/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/Expression.scala)). When calling its `eval()`,  we call the [Unevaluable](https://github.com/apache/spark/blob/v2.2.0/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/Expression.scala)'s `eval()` that will throw this exception mentioned above.
```
trait Unevaluable extends Expression {

  final override def eval(input: InternalRow = null): Any =
    throw new UnsupportedOperationException(s"Cannot evaluate expression: $this")

  final override protected def doGenCode(ctx: CodegenContext, ev: ExprCode): ExprCode =
    throw new UnsupportedOperationException(s"Cannot evaluate expression: $this")
}
```

Considering that we barely use the second parameter `pattern`, I remap `to_date` to `ToDate` again to fix the problem.

So, does my case relate to this PR?

anabranch:
I don't think this is necessarily my call but you're effectively writing a
hive udf, not a Spark one that depends on this. Writing a Spark UDF would
allow this to work just fine and the expression to be evaluated. Your work
around seems fine in your local version though.

PR #16140, opened 2016-12-05
[SPARK-18714][SQL] Add a simple time function to SparkSession

rxin:
Many Spark developers often want to test the runtime of some function in interactive debugging and testing. This patch adds a simple time function to SparkSession:

```
scala> spark.time { spark.range(1000).count() }
Time taken: 77 ms                                                               
res1: Long = 1000
```

## How was this patch tested?
I tested this interactively in spark-shell.

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/SparkSession.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/SparkSession.scala
@@ -618,6 +618,22 @@ class SparkSession private(
   @InterfaceStability.Evolving
   def readStream: DataStreamReader = new DataStreamReader(self)
 
+  /**
+   * Executes some code block and prints to stdout the time taken to execute the block. This is
+   * available in Scala only and is used primarily for interactive testing and debugging.
+   *
+   * @since 2.1.0
+   */
+  @InterfaceStability.Stable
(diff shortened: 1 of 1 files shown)


rxin:
cc @marmbrus

dongjoon-hyun:
Retest this please

rxin:
It's ok no need to retest. The change is fine.

dongjoon-hyun:
I see. BTW, @rxin . Do we need `spark` to measure time? Is it possible to add that into `SparkSession` companion object?

rxin:
That's much more difficult to type though.

dongjoon-hyun:
True. Never mind. I just thought we were able to use like the following.
```
    SparkSession.time { SparkSession.builder.appName("Spark Pi").getOrCreate() }
```

srowen:
_shrug_ It seems pretty unrelated to SparkSession. I get that it's a convenience method but something Spark really needs to expose as an API method?

rxin:
The cost to maintain this seems very small though, and I'd definitely use it all the time in the repl. In Databricks this is not an issue since the environment always appends the time, but I really miss this in the repl.

hvanhovell:

PR #16142, opened 2016-12-05
[SPARK-18716][CORE] Restrict the disk usage of spark event log.

uncleGen:

We've had reports of overfull disk usage of spark event log file. Current implementation has following drawbacks:
1. If we did not start Spark HistoryServer or Spark HistoryServer just failed, there is no chance to do clean work.
2. Spark HistoryServer is cleaning event log file based on file age only. If there are abundant applications constantly, the disk usage in every **spark.history.fs.cleaner.maxAge** can still be very large.

In this PR:

1. Do once clean work before save new application event log.
2. Provide a new clean mode, i.e. **space** mode with **age** mode default as existing implementation. We can set the max disk space to preserve the spark job event log. The excess part of job history files will be deleted, oldest file first. 

## How was this patch tested?

add new unit test


uncleGen:
retest it please

uncleGen:
@SparkQA Test it agagin

uncleGen:
cc @srowen

srowen:
Hm, does Spark generally  manage log rotation? I confess ignorance. It's something you often leave to a cron job or something to archive and clean up.

uncleGen:
@srowen Spark History Server may do the clean-up work. The precondition is we start it and it keeps running. Besides, if there are abundant applications constantly, the event log may still take up abundant storage space. This PR gives system another chance to clean-up work before each application begins saving event log. What's more, if you are more concerned with storage cost, this PR provides 'space' mode to restrict the disk usage of spark event log.

> It's something you often leave to a cron job or something to archive and clean up.

IMHO, I do not think it is a stable way to do this work. We may make sure it works and keeps working, just like Spark history server.

srowen:
Yes, but the alternative is reimplementing an ad-hoc log rotation system here, which isn't great either. Are you saying the history server already manages logs? pardon, I don't know it at all.

uncleGen:
@srowen If I have understand what you mean correctly,  the **"log rotation"** is different with **"job event log clean up"**.  The "job event log" is used to reply to build spark history ui. Right？：）

uncleGen:
@vanzin I found you provided this base work, could you please take a look? Thank you.

core/src/main/scala/org/apache/spark/scheduler/EventLoggingListener.scala
```diff
@@ -90,6 +91,10 @@ private[spark] class EventLoggingListener(
    * Creates the log file in the configured log directory.
    */
   def start() {
+    val statusList = Option(fileSystem.listStatus(new Path(logBaseDir))).map(_.toSeq)
+      .getOrElse(Seq[FileStatus]())
+    EventLoggingListener.cleanRedundantLogFiles(sparkConf, fileSystem, statusList)
```
vanzin: No, don't touch this file. First because applying history server settings in Spark apps is weird.

Second because users don't even have permission to do this in a properly deployed Spark. Users can only delete their own log files.

Spark apps should only be concerned with their own state, not figuring out how to manage things left over by other Spark apps.

uncleGen:
@vanzin I have removed related changes in `EventLoggingListener`, and provide a new clean mode, i.e. `space` based mode. Please take a review.

uncleGen:
retest please

uncleGen:
retest it please

uncleGen:
@AmplabJenkins  retest please

uncleGen:
cc @vanzin

vanzin:
I'm not such a big fan of this feature, but mostly I'm not a big fan of the current implementation.

For the feature, it feels like it's trying to make the SHS more like a "log management system" than a history server. There were already people who were concerned about this when the existing cleaner functionality was added. But I'll entertain the thought, even though you can probably get pretty close to this by using time-based deletion with a shorter max age, coupled with log compression (both features that already exist).

For the implementation, you cannot delete things just based on size. You need to account for time too; you have to delete older logs first, otherwise you risk deleting the logs for just finished applications instead of a large log that's been sitting there for months. It's also a reactive change; you're already using more space then you want too, the history server will just bring that down, eventually.

You change will also bombard the NameNode with requests on every scan, to get the size of each log.

At this point I'm not so convinced of the usefulness of the feature, and implementing it correctly will be a larger change than you have here.

uncleGen:
@vanzin 

> For the feature, it feels like it's trying to make the SHS more like a "log management system" than a history server. 

Sorry, I do not get it. I just add a new clean-up mode, but not add the cleaner itself. One is based on age, and the another one is base on space usage. That's all.

> But I'll entertain the thought, even though you can probably get pretty close to this by using time-based deletion with a shorter max age, coupled with log compression (both features that already exist).

I think you may not get what I mean. You can set the `max age` a very very small value, and use compression or whatever. Yes, you can use few space finally. But, this way will delete many latest job event logs, and then we can not review job history. Because, we use a wrong way to achieve restricting space usage.

> For the implementation, you cannot delete things just based on size. You need to account for time too; you have to delete older logs first, otherwise you risk deleting the logs for just finished applications instead of a large log that's been sitting there for months. 

I think current `space` mode is deleting logs base on space usage and oldest file first. If you do not think so, there may be something wrong in my implementation. I will check it.

> You change will also bombard the NameNode with requests on every scan, to get the size of each log.

The `cleanLog` will be called in every `CLEAN_INTERNAL_S` (1 day default), https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/deploy/history/FsHistoryProvider.scala#L218
If `CLEAN_INTERNAL_S` is set as a very very small value, current implementation has hurt NameNode already. So, I do not think `get the size of each log` will hurt NameNode greatly.

uncleGen:
Besides, the unit test has proved that the older file will be cleaned first.

vanzin:
> I just add a new clean-up mode, but not add the cleaner itself. 

But that's kinda the point. How many different ways of cleaning need to be added? Will this one be enough? Will people ask for archiving next? I'm wary of going down that path.

> I think you may not get what I mean.

I get what you mean. I just disagree with you that it's an important feature to have.

> So, I do not think get the size of each log will hurt NameNode greatly.

The current scan code does not make one request to the NameNode per log file in the directory. Your code does. That should be avoided.

> Besides, the unit test has proved that the older file will be cleaned first.

Your code doesn't do that, so if the unit test shows that it's not by design. Your code is scanning the list of apps in the order they're kept in memory (descending end time). I don't remember whether in progress apps come first or last. But if they come first, an old attempt of an in progress app will have precedence over newer attempts of apps that have already finished. If they come last, then you're first accounting for log sizes of apps that have already finished and might end up trying to delete logs from apps that are still running (!!!).

The way the current cleaner code works for time does not work if you're doing the `shouldClean` check solely based on space used. So this feature is not as trivial as your code make it seem.

srowen:
I personally don't see much value in deleting by 'size' because there's no particular reason to believe that files that happen to be bigger contain less important info.  If you want to delete more stuff, delete more from the oldest onwards.

uncleGen:
> The current scan code does not make one request to the NameNode per log file in the directory. Your code does. That should be avoided.

Make sense, current implementation can be optimized absolutely, it's my fault.

> If they come last, then you're first accounting for log sizes of apps that have already finished and might end up trying to delete logs from apps that are still running (!!!).

I get what you mean. Currently, order of apps log depends on the last attempts log time:

```
private def compareAppInfo(
      i1: FsApplicationHistoryInfo,
      i2: FsApplicationHistoryInfo): Boolean = {
    val a1 = i1.attempts.head
    val a2 = i2.attempts.head
    if (a1.endTime != a2.endTime) a1.endTime >= a2.endTime else a1.startTime >= a2.startTime
  }
```

So, if need clean up, completed job logs will be cleaned firstly, and the older in-progress (if exists).
#16165 has supported deleting too old in-progress job logs. So I think it is OK in this case.

uncleGen:
@srowen It has real demand. I have received some reports from our users: they care about space cost from job event logs, but not care about logs age, no matter how old it is.

If you guys do not like this feature, I will close it.

vanzin:
> #16165 has supported deleting too old in-progress job logs. So I think it is OK in this case.

It's *not* OK. That change uses an heuristic to decide when to delete "inprogress" logs: if they haven't been written to for a long time, it assumes the app is dead. That's an acceptable heuristic, since if an app hasn't written a single byte to the log in several days, it's probably not running anymore.

But your change may delete logs from *actively running* apps, just because you're exceeding the threshold you defined. That's not acceptable.

As for your users, I'd first try compression + shorter max age and see how that works out. If it doesn't, try enabling logs selectively just for applications they care about. Otherwise, there's not much sense in saving the logs in the first place, since they may disappear at any time.

If they have a custom need, you can always write a small daemon to do the cleanup, without having to modify Spark.

uncleGen:
@vanzin I will close this PR, but can't take your point completely.

PR #16143, opened 2016-12-05
[SPARK-18711][SQL] should disable subexpression elimination for LambdaVariable

cloud-fan:

This is kind of a long-standing bug, it's hidden until https://github.com/apache/spark/pull/15780 , which may add `AssertNotNull` on top of `LambdaVariable` and thus enables subexpression elimination.

However, subexpression elimination will evaluate the common expressions at the beginning, which is invalid for `LambdaVariable`. `LambdaVariable` usually represents loop variable, which can't be evaluated ahead of the loop.

This PR skips expressions containing `LambdaVariable` when doing subexpression elimination.

## How was this patch tested?

updated test in `DatasetAggregatorSuite`


cloud-fan:
cc @rxin @yhuai @hvanhovell

hvanhovell:

koertkuipers:
thanks for getting this fixed so fast

PR #16145, opened 2016-12-05
[MINOR][CORE][SQL] Remove explicit RDD and Partition overrides

jaceklaskowski:

I **believe** that I _only_ removed duplicated code (that adds nothing but noise). I'm gonna remove the comment after Jenkins has built the changes with no issues and Spark devs has agreed to include the changes.

Remove explicit `RDD` and `Partition` overrides (that turn out code duplication)

## How was this patch tested?

Local build. Awaiting Jenkins.

…cation)

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/ShuffledRowRDD.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/ShuffledRowRDD.scala
@@ -128,9 +128,7 @@ class ShuffledRowRDD(
   private[this] val part: Partitioner =
     new CoalescedPartitioner(dependency.partitioner, partitionStartIndices)
 
-  // Don't even think of removing the override
-  // and making the dependency explicit in `extends RDD` above. It has once been tried.
-  override def getDependencies: Seq[Dependency[_]] = Seq(dependency)
+  override def getDependencies: Seq[Dependency[_]] = List(dependency)
 
   override val partitioner: Option[Partitioner] = Some(part)
 
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/ShuffledRowRDD.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/ShuffledRowRDD.scala
@@ -128,6 +128,10 @@ class ShuffledRowRDD(
   private[this] val part: Partitioner =
     new CoalescedPartitioner(dependency.partitioner, partitionStartIndices)
 
+  // Don't even think of removing the override
+  // and making the dependency explicit in `extends RDD` above. It has once been tried.
+  override def getDependencies: Seq[Dependency[_]] = Seq(dependency)
+
   override val partitioner: Option[Partitioner] = Some(part)
 
   override def getPartitions: Array[Partition] = {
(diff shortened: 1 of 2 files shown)


sql/core/src/main/scala/org/apache/spark/sql/execution/ShuffledRowRDD.scala
```diff
@@ -117,7 +113,7 @@ class CoalescedPartitioner(val parent: Partitioner, val partitionStartIndices: A
 class ShuffledRowRDD(
     var dependency: ShuffleDependency[Int, InternalRow, InternalRow],
     specifiedPartitionStartIndices: Option[Array[Int]] = None)
-  extends RDD[InternalRow](dependency.rdd.context, Nil) {
+  extends RDD[InternalRow](dependency.rdd.context, Seq(dependency)) {
```
rxin: this might impact checkpointing - i rather not spend time digging into this to verify that it is correct ...

srowen: Agree, though it only has one other use at all:

```
  private[spark] def markCheckpointed(): Unit = {
    clearDependencies()
    partitions_ = null
    deps = null    // Forget the constructor argument for dependencies too
  }
```

Right now `ShuffledRowRDD` would always 'remember' its constructor parameter separately, when this comment suggests that's not desirable. This change would get rid of that. No idea if it makes any difference -- probably doesn't either way.

srowen: @rxin what do you think of that argument? if it's still not of interest, then @jaceklaskowski could you back out this part?

rxin: Yea I'd prefer not to change it unless somebody goes in and take a very careful look.

srowen: OK. I think it's actually an error as it stands given the comment, but we have no evidence it actually manifests as a problem.

jaceklaskowski:
@rxin @srowen Please re-review and act accordingly. Thanks a lot for your help so far!

sql/core/src/main/scala/org/apache/spark/sql/execution/ShuffledRowRDD.scala
```diff
@@ -132,8 +128,6 @@ class ShuffledRowRDD(
   private[this] val part: Partitioner =
     new CoalescedPartitioner(dependency.partitioner, partitionStartIndices)
 
-  override def getDependencies: Seq[Dependency[_]] = List(dependency)
```
srowen: Without this it will return nothing, because the superclass is passed Nil for deps.

sql/core/src/main/scala/org/apache/spark/sql/execution/ShuffledRowRDD.scala
rxin: If it is a bug fix then it's a very different story (and the title probably shouldn't say "minor" ...)

srowen:
Merged to master

PR #16146, opened 2016-12-05
[SPARK-18091] [SQL] [BACKPORT-1.6] Deep if expressions cause Generated SpecificUnsafeProjection code to exceed JVM code size limit

kapilsingh5050:

Fix for SPARK-18091 which is a bug related to large if expressions causing generated SpecificUnsafeProjection code to exceed JVM code size limit.

This PR changes if expression's code generation to place its predicate, true value and false value expressions' generated code in separate methods in context so as to never generate too long combined code.

## How was this patch tested?

Added a unit test and also tested manually with the application (having transformations similar to the unit test) which caused the issue to be identified in the first place.

### DIFF
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/conditionalExpressions.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/conditionalExpressions.scala
@@ -57,20 +57,70 @@ case class If(predicate: Expression, trueValue: Expression, falseValue: Expressi
     val trueEval = trueValue.gen(ctx)
     val falseEval = falseValue.gen(ctx)
 
-    s"""
-      ${condEval.code}
-      boolean ${ev.isNull} = false;
-      ${ctx.javaType(dataType)} ${ev.value} = ${ctx.defaultValue(dataType)};
-      if (!${condEval.isNull} && ${condEval.value}) {
-        ${trueEval.code}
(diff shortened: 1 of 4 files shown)


cloud-fan:
ok to test

cloud-fan:
hi @kapilsingh5050  can you also include https://github.com/apache/spark/pull/16244? It does fix the maven tests.

kapilsingh5050:
Yes, I'll do that but the test failures here are different. I'm still to figure out the root cause.

kapilsingh5050:
ExpressionEncoderSuite and RowEncoderSuite tests were failing because of following fix missing in branch-1.6:
https://github.com/apache/spark/commit/372baf0479840695388515170e6eae0b3fc4125e

kapilsingh5050:
@cloud-fan The last test failure is because of following error:
"impossible to get artifacts when data has not been loaded. IvyNode = org.scala-lang#scala-library;2.10.3"
for which [this](https://github.com/sbt/sbt/issues/2015) post suggests changing sbt version to 0.13.9 and I also found an open SPARK bug corresponding to this error. 
Are you aware of any recent changes in the build configuration for PR builder?
Can we retry the build?

cloud-fan:
retest this please

kapilsingh5050:
@cloud-fan please review and merge

rezasafi:
Hi, will this pull request merge to branch-1.6?

zzcclp:
ping @cloud-fan , will this pr be merged into branch-1.6?

PR #16149, opened 2016-12-05
[SPARK-18715][ML]Fix AIC calculations in Binomial GLM

actuaryzhang:
The AIC calculation in Binomial GLM seems to be off when the response or weight is non-integer: the result is different from that in R. This issue arises when one models rates, i.e, num of successes normalized over num of trials, and uses num of trials as weights. In this case, the effective likelihood is  weight * label ~ binomial(weight, mu), where weight = number of trials, and weight * label = number of successes and mu = is the success rate.  

@srowen @sethah @yanboliang @HyukjinKwon @zhengruifeng 

I suggest changing the current aic calculation for the Binomial family from 
```
-2.0 * predictions.map { case (y: Double, mu: Double, weight: Double) =>
        weight * dist.Binomial(1, mu).logProbabilityOf(math.round(y).toInt)
      }.sum()
```
to the following which generalizes to the case of real-valued response and weights. 
```
      -2.0 * predictions.map { case (y: Double, mu: Double, weight: Double) =>
        val wt = math.round(weight).toInt
        if (wt == 0){
          0.0
        } else {
          dist.Binomial(wt, mu).logProbabilityOf(math.round(y * weight).toInt)
        }
      }.sum()
``` 
## How was this patch tested?
I will write the unit test once the community wants to include the proposed change. For now, the following modifies existing tests in weighted Binomial GLM to illustrate the issue. The second label is changed from 0 to 0.5. 

```
val datasetWithWeight = Seq(
    (1.0, 1.0, 0.0, 5.0),
    (0.5, 2.0, 1.0, 2.0),
    (1.0, 3.0, 2.0, 1.0),
    (0.0, 4.0, 3.0, 3.0)
  ).toDF("y", "w", "x1", "x2")

val formula = (new RFormula()
  .setFormula("y ~ x1 + x2")
  .setFeaturesCol("features")
  .setLabelCol("label"))
val output = formula.fit(datasetWithWeight).transform(datasetWithWeight).select("features", "label", "w")

val glr = new GeneralizedLinearRegression()
    .setFamily("binomial")
    .setWeightCol("w")
    .setFitIntercept(false)
    .setRegParam(0)

val model = glr.fit(output)
model.summary.aic
```
The AIC from Spark is 17.3227, and the AIC from R is 15.66454.

### DIFF
--- a/mllib/src/main/scala/org/apache/spark/ml/regression/GeneralizedLinearRegression.scala
+++ b/mllib/src/main/scala/org/apache/spark/ml/regression/GeneralizedLinearRegression.scala
@@ -215,6 +215,7 @@ class GeneralizedLinearRegression @Since("2.0.0") (@Since("2.0.0") override val 
    * Sets the value of param [[weightCol]].
    * If this is not set or empty, we treat all instance weights as 1.0.
    * Default is not set, so all instances have weight one.
+   * In the Binomial model, weights correspond to number of trials.
    *
    * @group setParam
    */
@@ -468,11 +469,7 @@ object GeneralizedLinearRegression extends DefaultParamsReadable[GeneralizedLine
     override def variance(mu: Double): Double = mu * (1.0 - mu)
 
     private def ylogy(y: Double, mu: Double): Double = {
-      if (y == 0) {
-        0.0
-      } else {
-        y * math.log(y / mu)
-      }
+      if (y == 0) 0.0 else y * math.log(y / mu)
     }
 
     override def deviance(y: Double, mu: Double, weight: Double): Double = {
@@ -485,6 +482,7 @@ object GeneralizedLinearRegression extends DefaultParamsReadable[GeneralizedLine
         numInstances: Double,
         weightSum: Double): Double = {
       -2.0 * predictions.map { case (y: Double, mu: Double, weight: Double) =>
+        // weights for Binomial distribution correspond to number of trials
         val wt = math.round(weight).toInt
         if (wt == 0) {
           0.0
--- a/mllib/src/main/scala/org/apache/spark/ml/regression/GeneralizedLinearRegression.scala
+++ b/mllib/src/main/scala/org/apache/spark/ml/regression/GeneralizedLinearRegression.scala
@@ -468,9 +468,14 @@ object GeneralizedLinearRegression extends DefaultParamsReadable[GeneralizedLine
     override def variance(mu: Double): Double = mu * (1.0 - mu)
 
     override def deviance(y: Double, mu: Double, weight: Double): Double = {
-      val my = 1.0 - y
-      2.0 * weight * (y * math.log(math.max(y, 1.0) / mu) +
-        my * math.log(math.max(my, 1.0) / (1.0 - mu)))
+      val y_logy = (y: Double, mu: Double) => {
+        if (y == 0) {
+          0.0
+        } else {
+          y * math.log(y / mu)
+        }
+      }
+      2.0 * weight * (y_logy(y, mu) + y_logy(1.0 - y, 1.0 - mu))
     }
 
     override def aic(
--- a/mllib/src/main/scala/org/apache/spark/ml/regression/GeneralizedLinearRegression.scala
+++ b/mllib/src/main/scala/org/apache/spark/ml/regression/GeneralizedLinearRegression.scala
@@ -215,7 +215,8 @@ class GeneralizedLinearRegression @Since("2.0.0") (@Since("2.0.0") override val 
    * Sets the value of param [[weightCol]].
    * If this is not set or empty, we treat all instance weights as 1.0.
    * Default is not set, so all instances have weight one.
-   * In the Binomial model, weights correspond to number of trials.
+   * In the Binomial family, weights correspond to number of trials and should be integer.
+   * Non-integer weights are rounded to integer in AIC calculation.
    *
    * @group setParam
    */
--- a/mllib/src/main/scala/org/apache/spark/ml/regression/GeneralizedLinearRegression.scala
+++ b/mllib/src/main/scala/org/apache/spark/ml/regression/GeneralizedLinearRegression.scala
@@ -467,15 +467,16 @@ object GeneralizedLinearRegression extends DefaultParamsReadable[GeneralizedLine
 
     override def variance(mu: Double): Double = mu * (1.0 - mu)
 
-    override def deviance(y: Double, mu: Double, weight: Double): Double = {
-      val y_logy = (y: Double, mu: Double) => {
-        if (y == 0) {
-          0.0
-        } else {
-          y * math.log(y / mu)
-        }
+    private def ylogy(y: Double, mu: Double): Double = {
+      if (y == 0) {
+        0.0
+      } else {
+        y * math.log(y / mu)
       }
-      2.0 * weight * (y_logy(y, mu) + y_logy(1.0 - y, 1.0 - mu))
+    }
+
+    override def deviance(y: Double, mu: Double, weight: Double): Double = {
+      2.0 * weight * (ylogy(y, mu) + ylogy(1.0 - y, 1.0 - mu))
     }
 
     override def aic(
--- a/mllib/src/main/scala/org/apache/spark/ml/regression/GeneralizedLinearRegression.scala
+++ b/mllib/src/main/scala/org/apache/spark/ml/regression/GeneralizedLinearRegression.scala
@@ -479,7 +479,12 @@ object GeneralizedLinearRegression extends DefaultParamsReadable[GeneralizedLine
         numInstances: Double,
         weightSum: Double): Double = {
       -2.0 * predictions.map { case (y: Double, mu: Double, weight: Double) =>
-        weight * dist.Binomial(1, mu).logProbabilityOf(math.round(y).toInt)
+        val wt = math.round(weight).toInt
+        if (wt == 0) {
+          0.0
+        } else {
+          dist.Binomial(wt, mu).logProbabilityOf(math.round(y * weight).toInt)
+        }
       }.sum()
     }
 
--- a/mllib/src/test/scala/org/apache/spark/ml/regression/GeneralizedLinearRegressionSuite.scala
+++ b/mllib/src/test/scala/org/apache/spark/ml/regression/GeneralizedLinearRegressionSuite.scala
@@ -715,7 +715,7 @@ class GeneralizedLinearRegressionSuite
     val datasetWithWeight = Seq(
       Instance(1.0, 1.0, Vectors.dense(0.0, 5.0).toSparse),
       Instance(0.5, 2.0, Vectors.dense(1.0, 2.0)),
-      Instance(1.0, 3.0, Vectors.dense(2.0, 1.0)),
+      Instance(1.0, 0.3, Vectors.dense(2.0, 1.0)),
       Instance(0.0, 4.0, Vectors.dense(3.0, 3.0))
     ).toDF()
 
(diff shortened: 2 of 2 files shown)


actuaryzhang:
Jenkins, add to whitelist

mllib/src/main/scala/org/apache/spark/ml/regression/GeneralizedLinearRegression.scala
```diff
@@ -479,7 +479,12 @@ object GeneralizedLinearRegression extends DefaultParamsReadable[GeneralizedLine
         numInstances: Double,
         weightSum: Double): Double = {
       -2.0 * predictions.map { case (y: Double, mu: Double, weight: Double) =>
-        weight * dist.Binomial(1, mu).logProbabilityOf(math.round(y).toInt)
+        val wt = math.round(weight).toInt
+        if (wt == 0) {
+          0.0
+        } else {
+          dist.Binomial(wt, mu).logProbabilityOf(math.round(y * weight).toInt)
```
srowen: I think I understand the problem, because a weighted instance is treated like _n_ copies of one instance, whose likelihood is treated like a Bernoulli trial (this could have been much simpler in the existing code right -- log(mu) or log(1-mu) depending on y?). 

This loses some info because y is rounded, so a weight=100 instance with y=0.5 is treated like 100 instance of y=1, when it kind of should be treated like 50 instances each of y=0 and y=1.

This matters not must for non-integer weights right? you would also get a different answer in the case above.

I am out of my depth here, but is this the best generalization? noticeably, you have to round the weight, making this inaccurate for non-integer weights.

You have to handle wt=0 separately because dist.Binomial will reject it? Binomial with n=0 ought to be well defined. The problem is this says the log probability of an instance with weight<0.5 is always 0, but that's not 'true' -- I guess it depends on how one defines these weights.

Does this match what R does or something? I'm trying to figure out if this is the right thing to do.

sethah: So I think the real issue here is that we don't currently allow users to specify a binomial GLM using success/outcome pairs. One way to mash that kind of grouped data into the format Spark requires is using the process described above by @actuaryzhang, but then we need to adjust the log-likelihood computation as was also noted. 

So @srowen is correct in saying that this is inaccurate for non-integer weights. I checked with R's glmnet, and it seems that they obey the semantics of data weights for a binomial GLM corresponding to the number of successes. So they log a warning when you input data weights of non-integer values, then proceed with the method proposed in this patch. 

So, this actually _does_ match R's behavior and I am in favor of the change. But we need to log appropriate warnings and write good unit tests. What are others' thoughts?

srowen: SGTM assuming tests pass. I'm still curious if we need to special case wt=0 but it's fine.

actuaryzhang: @srowen The current uses [breeze](https://github.com/scalanlp/breeze/blob/master/math/src/main/scala/breeze/stats/distributions/Binomial.scala) for computing the Binomial density, which does not allow 0 trials. `require(n > 0, "n must be positive!")`. That is why I have to handle the case of `wt = 0`

actuaryzhang:
@srowen @sethah 
Thanks for the comments. Yes, the major use case is to be able to handle multiple trials (integer weight, real-valued response). Indeed, a better way to do this is through `offset`, which I have proposed to do in this JIRA [SPARK-18710](https://issues.apache.org/jira/browse/SPARK-18710). Please let me know if this is worth pursuing. 

I have submitted another two commits.

1. One commit makes minimal modification to the exiting Binomial GLM test so that one response record is now non-integer. This test still failed because the deviance residual calculation seems to work only for `y in (0, 1)`
2. The second commit fixes the issue in calculating the deviance. But I think the code can be improved, especially regarding the function `y_logy`. What's the best way to create a utility function like this? Please advise. 

With the two commits, all tests now passed, including the ones on AIC and the deviance.

srowen:
What you have is fine (though I might name it `ylogy` or something). I don't see other places in the code that compute x ln x or something similar, so it's OK to make this a private function. You might just make a `private def` of a local helper method rather than instantiate a lambda, but the difference is trivial.

actuaryzhang:
@srowen @sethah 
I have cleaned up the change as suggested. Please review and let me know if there is any question.

srowen (approved):
Looking OK to me. It looks like the tests update to cover the new behavior. Logging a warning may still be a good idea, though ideally in some way that it doesn't spew warnings continuously. That could be a little tricky.

mllib/src/test/scala/org/apache/spark/ml/regression/GeneralizedLinearRegressionSuite.scala
```diff
@@ -708,73 +708,72 @@ class GeneralizedLinearRegressionSuite
...
        A <- matrix(c(0, 1, 2, 3, 5, 2, 1, 3), 4, 2)
-       b <- c(1, 0, 1, 0)
+       b <- c(1, 0.5, 1, 0)
        w <- c(1, 2, 3, 4)
        df <- as.data.frame(cbind(A, b))
      */
     val datasetWithWeight = Seq(
       Instance(1.0, 1.0, Vectors.dense(0.0, 5.0).toSparse),
-      Instance(0.0, 2.0, Vectors.dense(1.0, 2.0)),
+      Instance(0.5, 2.0, Vectors.dense(1.0, 2.0)),
```
sethah: Let's add a fractional weight as well.

mllib/src/main/scala/org/apache/spark/ml/regression/GeneralizedLinearRegression.scala
```diff
@@ -467,10 +467,16 @@ object GeneralizedLinearRegression extends DefaultParamsReadable[GeneralizedLine
 
     override def variance(mu: Double): Double = mu * (1.0 - mu)
 
+    private def ylogy(y: Double, mu: Double): Double = {
```
sethah: nit: `private def ylogy(y: Double, mu: Double): Double = if (y == 0) 0.0 else y * math.log(y / mu)`

sethah:
@srowen We can add a note to the doc for `setWeightCol`. We could also use `logInfo` about weights needing to be integer values for `Binomial` family, but that may not be very effective.

mllib/src/main/scala/org/apache/spark/ml/regression/GeneralizedLinearRegression.scala
```diff
@@ -479,7 +485,12 @@ object GeneralizedLinearRegression extends DefaultParamsReadable[GeneralizedLine
         numInstances: Double,
         weightSum: Double): Double = {
       -2.0 * predictions.map { case (y: Double, mu: Double, weight: Double) =>
-        weight * dist.Binomial(1, mu).logProbabilityOf(math.round(y).toInt)
+        val wt = math.round(weight).toInt
+        if (wt == 0) {
+          0.0
+        } else {
+          dist.Binomial(wt, mu).logProbabilityOf(math.round(y * weight).toInt)
```
sethah: Maybe a comment here: `// weights for Binomial distribution correspond to number of trials`

actuaryzhang: @sethah Code updated as suggested

actuaryzhang:
@sethah @srowen I have added a comment to the weigthCol doc for the Binomial case. 
I also updated to test the case `weight < 0.5`, i.e., `round(weight) = 0`. All tests passed.

mllib/src/test/scala/org/apache/spark/ml/regression/GeneralizedLinearRegressionSuite.scala
```diff
@@ -715,7 +715,7 @@ class GeneralizedLinearRegressionSuite
     val datasetWithWeight = Seq(
       Instance(1.0, 1.0, Vectors.dense(0.0, 5.0).toSparse),
       Instance(0.5, 2.0, Vectors.dense(1.0, 2.0)),
-      Instance(1.0, 3.0, Vectors.dense(2.0, 1.0)),
+      Instance(1.0, 0.3, Vectors.dense(2.0, 1.0)),
```
sethah: Hm, so I know this is a pain, but we have special handling implemented for the weight = 0 case, but we never test it. I think we should add a test.

mllib/src/main/scala/org/apache/spark/ml/regression/GeneralizedLinearRegression.scala
```diff
@@ -468,11 +469,7 @@ object GeneralizedLinearRegression extends DefaultParamsReadable[GeneralizedLine
     override def variance(mu: Double): Double = mu * (1.0 - mu)
 
     private def ylogy(y: Double, mu: Double): Double = {
-      if (y == 0) {
-        0.0
-      } else {
-        y * math.log(y / mu)
-      }
+      if (y == 0) 0.0 else y * math.log(y / mu)
```
sethah: Well the entire thing can go on one line, but only change it if you make another commit since it's trivial.

mllib/src/main/scala/org/apache/spark/ml/regression/GeneralizedLinearRegression.scala
```diff
@@ -215,6 +215,7 @@ class GeneralizedLinearRegression @Since("2.0.0") (@Since("2.0.0") override val
    * Sets the value of param [[weightCol]].
    * If this is not set or empty, we treat all instance weights as 1.0.
    * Default is not set, so all instances have weight one.
+   * In the Binomial model, weights correspond to number of trials.
```
sethah: We should note that the weights should therefore be integers and that they'll be rounded if they are not. Also say "Binomial family" instead of "Binomial model."

mllib/src/test/scala/org/apache/spark/ml/regression/GeneralizedLinearRegressionSuite.scala
actuaryzhang: Isn't `round(0.3) = 0`? 
Or do you mean testing the raw (input) weight  = 0? How is that handled? I would assume that is just leaving out that observation, since that's how R does it.

actuaryzhang: If you mean input weight = 0, then it is better to be tested separately along with all the other families from the current PR. I think the current test is good for the current PR.

mllib/src/main/scala/org/apache/spark/ml/regression/GeneralizedLinearRegression.scala
actuaryzhang: I actually prefer not putting the whole thing in one line. This is more readable.

actuaryzhang:
@sethah @srowen 
I updated the documentation. I think we have everything needed for this fix. Please merge and close this PR if there is no other issue. Thanks much for all the comments.

srowen (approved):
OK by me, let's see if @sethah approves

actuaryzhang:
@sethah Would you please review this? Thanks.

mllib/src/main/scala/org/apache/spark/ml/regression/GeneralizedLinearRegression.scala
```diff
@@ -467,10 +469,12 @@ object GeneralizedLinearRegression extends DefaultParamsReadable[GeneralizedLine
 
     override def variance(mu: Double): Double = mu * (1.0 - mu)
 
+    private def ylogy(y: Double, mu: Double): Double = {
+      if (y == 0) 0.0 else y * math.log(y / mu)
+    }
+
     override def deviance(y: Double, mu: Double, weight: Double): Double = {
```
sethah: I would prefer to remove the added method `ylogy`, which is only used in one place, and replace it with the following logic:

````scala
override def deviance(y: Double, mu: Double, weight: Double): Double = {
      val clamped = y.min(1 - epsilon).max(epsilon)
      2.0 * weight * (clamped * math.log(clamped / mu) + (1 - clamped) * math.log((1 - clamped) / (1 - mu)))
    }
````

I think it's better to avoid adding a static method that is only used in one other method, and additionally I think it's clearer what the intent is here. I'm open to other opinions though.

srowen: Why do you need to avoid the y=0 case? y ln y approaches 0 as y approaches 0, and the utility method correctly captures that.

sethah: The utility method is not incorrect, I just think the above is more clear. That is how I've seen it done in e.g. statsmodels, and I think adding a static method for this simple logic is a bit superfluous. Isn't the entire point of the static method to avoid the y=0 case, which would calculate log(0), which is undefined?

Anyway, if we alternately elect to keep the `ylogy` method, I think a short doc would be helpful.

srowen: Yes, but isn't handling y=0 better than introducing a tiny error in computation? there's no need to avoid computing y ln y at y=0 because it's value is well defined for this purpose. It seems quite less clear to me; I've only ever seen computing y ln y by special-casing y=0 and not restricting its domain.

sethah: Fair enough. It's a minor point and not worth delaying for, thanks for your feedback!

actuaryzhang:
@srowen @sethah 
One more commit that adds a test case with `weight = 4.7` which will round up to 5 to test the case @sethah described. All tests passed.  I'm pretty sure R's rounding is the same as what I'm doing here. Please merge if there is no other issue. Thanks.

sethah:
LGTM

srowen:
Merged to master

actuaryzhang:
@srowen @sethah 
Thanks for all the helpful discussions!

PR #16150, opened 2016-12-05
[SPARK-18349][SparkR]:Update R API documentation on ml model summary

wangmiao1981:
In this PR, the document of `summary` method is improved in the format: 

returns summary information of the fitted model, which is a list. The list includes .......

Since `summary` in R is mainly about the model, which is not the same as `summary` object on scala side, if there is one, the scala API doc is not pointed here.

In current document, some `@return` have `.` and some don't have. `.` is added to missed ones.

Since spark.logit `summary` has a big refactoring, this PR doesn't include this one. It will be changed when the `spark.logit` PR is merged.

## How was this patch tested?

Manual build.

### DIFF
--- a/R/pkg/R/mllib.R
+++ b/R/pkg/R/mllib.R
@@ -666,7 +666,8 @@ setMethod("fitted", signature(object = "KMeansModel"),
 
 #' @param object a fitted k-means model.
 #' @return \code{summary} returns summary information of the fitted model, which is a list.
-#'         The list includes the model's \code{coefficients} (model cluster centers),
+#'         The list includes the model's \code{k} (number of cluster centers),
+#'         \code{coefficients} (model cluster centers),
 #'         \code{size} (number of data points in each cluster), and \code{cluster}
 #'         (cluster centers of the transformed data).
 #' @rdname spark.kmeans
@@ -688,7 +689,7 @@ setMethod("summary", signature(object = "KMeansModel"),
             } else {
               dataFrame(callJMethod(jobj, "cluster"))
             }
-            list(coefficients = coefficients, size = size,
+            list(k = k, coefficients = coefficients, size = size,
                  cluster = cluster, is.loaded = is.loaded)
           })
 
--- a/R/pkg/R/mllib.R
+++ b/R/pkg/R/mllib.R
@@ -278,10 +278,10 @@ setMethod("glm", signature(formula = "formula", family = "ANY", data = "SparkDat
 
 #' @param object a fitted generalized linear model.
 #' @return \code{summary} returns summary information of the fitted model, which is a list.
-#'         The list of components includes at least the coefficients matrix (which includes
+#'         The list of components includes at least the \code{coefficients} (coefficients matrix, which includes
 #'         coefficients, standard error of coefficients, t value and p value),
-#'         null/residual deviance, null/residual degrees of freedom, AIC and number of
-#'         iterations IRLS takes. If there are collinear columns in the data,
+#'         \code{null.deviance} (null/residual degrees of freedom), \code{aic} (AIC)
+#'         and \code{iter} (number of iterations IRLS takes). If there are collinear columns in the data,
 #'         the coefficients matrix only provides coefficients.
 #' @rdname spark.glm
 #' @export
@@ -567,7 +567,8 @@ setMethod("predict", signature(object = "IsotonicRegressionModel"),
 #  Get the summary of an IsotonicRegressionModel model
 
 #' @return \code{summary} returns summary information of the fitted model, which is a list.
-#'         The list includes model's boundaries and prediction.
+#'         The list includes model's \code{boundaries} (boundaries in increasing order)
+#'         and \code{predictions} (predictions associated with the boundaries at the same index).
 #' @rdname spark.isoreg
 #' @aliases summary,IsotonicRegressionModel-method
 #' @export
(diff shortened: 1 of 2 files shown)


wangmiao1981:
cc @felixcheung

felixcheung:
thanks, there is also the issue with `\code{numOfInputs}` vs `number of iterations IRLS takes`
- should it be a "variable" (and thus wrapped with `\code{something}`
- or should it be a description?

felixcheung:
there is also this form
`\code{apriori} (the label distribution)`

and this form
`\item{\code{docConcentration}}{concentration parameter commonly named \code{alpha} `

wangmiao1981:
The ideal case might be "variable"+description. Let me give a try and push it again.

R/pkg/R/mllib.R
```diff
@@ -661,7 +665,10 @@ setMethod("fitted", signature(object = "KMeansModel"),
 #  Get the summary of a k-means model
 
 #' @param object a fitted k-means model.
-#' @return \code{summary} returns the model's features, coefficients, k, size and cluster.
+#' @return \code{summary} returns summary information of the fitted model, which is a list.
+#'         The list includes the model's \code{coefficients} (model cluster centers),
```
felixcheung: are we missing features and k?

R/pkg/R/mllib.R
```diff
@@ -1389,7 +1399,9 @@ setMethod("spark.gaussianMixture", signature(data = "SparkDataFrame", formula =
 #  Get the summary of a multivariate gaussian mixture model
 
 #' @param object a fitted gaussian mixture model.
-#' @return \code{summary} returns the model's lambda, mu, sigma, k, dim and posterior.
+#' @return \code{summary} returns summary of the fitted model, which is a list.
+#'         The list includes the model's \code{lambda} (lambda), \code{mu} (mu),
+#'         \code{sigma} (sigma), and \code{posterior} (posterior).
```
felixcheung: missing k, dim?

R/pkg/R/mllib.R
wangmiao1981: For the return list, I didn't see features and k.  Does R function not only return the last line?

R/pkg/R/mllib.R
```diff
@@ -1852,9 +1867,9 @@ summary.treeEnsemble <- function(model) {
 
 #  Get the summary of a Random Forest Regression Model
 
-#' @return \code{summary} returns a summary object of the fitted model, a list of components
-#'         including formula, number of features, list of features, feature importances, number of
-#'         trees, and tree weights
+#' @return \code{summary} returns summary information of the fitted model, which is a list.
+#'         The list of components includes \code{ans} (formula, number of features, list of features,
+#'         feature importances, number of trees, and tree weights).
```
felixcheung: this should be something like
```
\code{formula} (something), \code{numFeatures} (something), \code{features} ....
```

`ans` is not the actual item or list of items. See https://github.com/apache/spark/pull/16150/files#diff-7ede1519b4a56647801b51af33c2dd18R1859

R/pkg/R/mllib.R
```diff
@@ -2076,9 +2091,9 @@ setMethod("write.ml", signature(object = "GBTClassificationModel", path = "chara
 
 #  Get the summary of a Gradient Boosted Tree Regression Model
 
-#' @return \code{summary} returns a summary object of the fitted model, a list of components
-#'         including formula, number of features, list of features, feature importances, number of
-#'         trees, and tree weights
+#' @return \code{summary} returns summary information of the fitted model, which is a list.
+#'         The list of components includes \code{ans} (formula, number of features, list of features,
+#'         feature importances, number of trees, and tree weights).
```
felixcheung: ditto previous comment. see https://github.com/apache/spark/pull/16150/files#diff-7ede1519b4a56647801b51af33c2dd18R1859

R/pkg/R/mllib.R
wangmiao1981: the two places returns `summary.treeEnsemble(object)`. What shall I put in the `\code{}`?

wangmiao1981 (commented):
I have questions in the two places.

R/pkg/R/mllib.R
wangmiao1981: Same reason as the above one.

R/pkg/R/mllib.R
felixcheung: ah, sorry you are right, it's not in the returned list

R/pkg/R/mllib.R
felixcheung: that's a function call, see https://github.com/apache/spark/pull/16150/files/6039c1f742f77b4ceff624b3979afbaa9e80d77a#diff-7ede1519b4a56647801b51af33c2dd18R1851

R/pkg/R/mllib.R
wangmiao1981: Do you think we should return these values? If so, I can file a followup PR.

felixcheung: Possibly `k` - as a description like the R `kmeans` summary

felixcheung:
LGTM.
@rxin I know rc2 has been cut but can this still go to branch-2.1? There're only 2 lines of code change and API doc improvements could really help usability in 2.1.
+ @shivaram

shivaram:
Yeah. This is mostly docs fixes - I think this is fine to go into branch-2.1. We can see which RC gets promoted etc.

felixcheung:
merged to master and branch-2.1

PR #16154, opened 2016-12-05
[SPARK-17822] [R] Make JVMObjectTracker a member variable of RBackend

mengxr:

* This PR changes `JVMObjectTracker` from `object` to `class` and let its instance associated with each RBackend. So we can manage the lifecycle of JVM objects when there are multiple `RBackend` sessions. `RBackend.close` will clear the object tracker explicitly.
* I assume that `SQLUtils` and `RRunner` do not need to track JVM instances, which could be wrong.
* Small refactor of `SerDe.sqlSerDe` to increase readability.

## How was this patch tested?

* Added unit tests for `JVMObjectTracker`.
* Wait for Jenkins to run full tests.


shivaram:
Thanks @mengxr for the change. Taking a look now

cc @felixcheung @falaki

mengxr:
@shivaram Let's wait Jenkins first:) I'm not sure it will pass.

mengxr:
@shivaram PR seems ready for you to make a pass:)

core/src/main/scala/org/apache/spark/api/r/RBackendHandler.scala
```diff
@@ -143,12 +142,8 @@ private[r] class RBackendHandler(server: RBackend)
       val cls = if (isStatic) {
         Utils.classForName(objId)
       } else {
-        JVMObjectTracker.get(objId) match {
-          case None => throw new IllegalArgumentException("Object not found " + objId)
-          case Some(o) =>
-            obj = o
-            o.getClass
-        }
+        obj = server.jvmObjectTracker.get(JVMObjectId(objId))
```
shivaram: This is minor but the exception now thrown will be a `NullPointerException` in place of what it was before  ?

core/src/main/scala/org/apache/spark/api/r/RRunner.scala
```diff
@@ -152,7 +152,7 @@ private[spark] class RRunner[U](
           dataOut.writeInt(mode)
 
           if (isDataFrame) {
-            SerDe.writeObject(dataOut, colNames)
+            SerDe.writeObject(dataOut, colNames, jvmObjectTracker = null)
```
shivaram: I think this change is fine as we are only writing out an array of string -- the other thing we can do is replace the code here and not call SerDe -- but instead just write out the code for writing an array of strings.

sql/core/src/main/scala/org/apache/spark/sql/api/r/SQLUtils.scala
```diff
@@ -158,7 +158,7 @@ private[sql] object SQLUtils extends Logging {
     val dis = new DataInputStream(bis)
     val num = SerDe.readInt(dis)
     Row.fromSeq((0 until num).map { i =>
-      doConversion(SerDe.readObject(dis), schema.fields(i).dataType)
+      doConversion(SerDe.readObject(dis, jvmObjectTracker = null), schema.fields(i).dataType)
```
shivaram: I think this one is safe because this is called on the output from the R UDF and the R UDF is assumed to return native types (i.e. no references to other JVM objects)

sql/core/src/main/scala/org/apache/spark/sql/api/r/SQLUtils.scala
```diff
@@ -167,7 +167,7 @@ private[sql] object SQLUtils extends Logging {
     val dos = new DataOutputStream(bos)
 
     val cols = (0 until row.length).map(row(_).asInstanceOf[Object]).toArray
-    SerDe.writeObject(dos, cols)
+    SerDe.writeObject(dos, cols, jvmObjectTracker = null)
```
shivaram: similarly this is called on the input to the R UDFs in `dapply` and again I dont think we allow passing in JVM objects there, so I think this is fine

sql/core/src/main/scala/org/apache/spark/sql/api/r/SQLUtils.scala
```diff
@@ -247,7 +247,7 @@ private[sql] object SQLUtils extends Logging {
     dataType match {
       case 's' =>
         // Read StructType for DataFrame
-        val fields = SerDe.readList(dis).asInstanceOf[Array[Object]]
+        val fields = SerDe.readList(dis, jvmObjectTracker = null).asInstanceOf[Array[Object]]
```
shivaram: The two changes below are functions used in serializing and deserializing StructType from R to JVM (when calling collect or parallelize)

Again I think since StructTypes themselves can't contain other JVM references I think its fine, but I'm less sure about this. (i.e what happens when we say embed an array into the structType). 

Another option here is to figure out which RBackend is calling this function and use its JVMObjectTracker ? Do we want to support cases where there are multiple active RBackends ? If not we can denote an `activeRBackend` and then use its jvm object tracker in these places ?

shivaram (commented):
Thanks @mengxr - I left a few comments inline. Some of them are just to reassure that this change is safe. We can add them as comments in the code. Also it will be good to document our assumptions more clearly somewhere -- i.e. having one RBackend (while handling re-connects) or handling multiple ones etc.

core/src/main/scala/org/apache/spark/api/r/JVMObjectTracker.scala
```diff
@@ -0,0 +1,65 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.spark.api.r
+
+import java.util.concurrent.atomic.AtomicInteger
+import java.util.concurrent.ConcurrentHashMap
+
+/** JVM object ID wrapper */
+private[r] case class JVMObjectId(id: String)
```
mengxr: I had the same question. But this is orthogonal to the changes in this PR. We can discussed it as a follow-up task.

core/src/main/scala/org/apache/spark/api/r/RBackendHandler.scala
mengxr: It is not an NullPointerException, it returns `null` instead. I thought about wrapping it with an `Option` but I don't see corner cases for internal use.

sql/core/src/main/scala/org/apache/spark/sql/api/r/SQLUtils.scala
mengxr: That would be ideal but requires a major refactoring, because we use `SerDe` as a singleton in many places. And we haven't fully discussed how to handle multiple active `RBackend`s yet.

core/src/main/scala/org/apache/spark/api/r/RBackendHandler.scala
shivaram: The `get` returns `null` but the next line will throw the exception. The old code was instead throwing a `IllegalArgumentException("Object not found " + objId)`

sql/core/src/main/scala/org/apache/spark/sql/api/r/SQLUtils.scala
shivaram: Yeah my take would be to not allow for multiple RBackends right now until we can figure this out. Or mark that as unsupported.

shivaram (commented):
One more thing - could @yhuai or @jkbradley test this out to see if it fixes the original memory pressure issue they were seeing ?

core/src/main/scala/org/apache/spark/api/r/RBackendHandler.scala
falaki: +1 for throwing an exception with old error message. Or even better add more description. The R users generally complain about cryptic error messages in SparkR.

core/src/main/scala/org/apache/spark/api/r/SerDe.scala
```diff
@@ -272,18 +282,22 @@ private[spark] object SerDe {
...
       throw new IllegalArgumentException(s"Invalid map key type: ${key.getClass.getName}")
     }
 
     writeString(dos, key.asInstanceOf[String])
-    writeObject(dos, value)
+    writeObject(dos, value, jvmObjectTracker)
   }
 
-  def writeObject(dos: DataOutputStream, obj: Object): Unit = {
+  def writeObject(dos: DataOutputStream, obj: Object, jvmObjectTracker: JVMObjectTracker): Unit = {
```
falaki: Do we have a policy against implicit parameter in Spark? If you make `JVMObjectTracker` an implicit parameter your code will become a lot cleaner. Basically you won't need to change every place `writeObject()` is called.

core/src/main/scala/org/apache/spark/api/r/JVMObjectTracker.scala
```diff
@@ -0,0 +1,65 @@
...
+
+  private[this] val objMap = new ConcurrentHashMap[JVMObjectId, Object]()
+  private[this] val objCounter = new AtomicInteger()
+
+  /**
+   * Returns the JVM object associated with the input key.
+   * Returns null if the key does not exist or has been removed.
+   */
+  @throws[NullPointerException]("if key is null")
+  final def get(id: JVMObjectId): Object = objMap.get(id)
```
falaki: Would be good to catch the exception here and throw another one with more R-friendly message.

falaki (commented):
Thanks @mengxr. I suggest adding another unit test to make sure, when `RBackend.close()` is called the member `JVMOBjectId` is cleared. That is the primary goal of this PR if I am correct and would be good to add a regression test for it.

core/src/main/scala/org/apache/spark/api/r/SerDe.scala
shivaram: I dont know if we have a policy, but I for one like the explicit parameter in this case. It at least makes it clear where we need to reason about the JVM object tracker being used.  Also if somebody adds a new `SerDe.writeObject` call in the future it ensures they think about it.

falaki: I agree. Using implicits has its readability downsides. I don't have a strong opinion in this case.

core/src/main/scala/org/apache/spark/api/r/JVMObjectTracker.scala
felixcheung: I'd agree completely, but generally the R-friendly handling would need to be done on the R side.
Related JIRA:
https://issues.apache.org/jira/browse/SPARK-17838
https://issues.apache.org/jira/browse/SPARK-18040

core/src/main/scala/org/apache/spark/api/r/RBackendHandler.scala
mengxr: Switched back to `Option`.

mengxr:
@falaki I'm happy to add the first Scala unit test for `RBackend`:)

core/src/main/scala/org/apache/spark/api/r/SerDe.scala
mengxr: I didn't make any changes. I think we should discuss more about supporting multiple R sessions before we do more refactoring work here.

mengxr:
@yhuai @jkbradley I'm testing the patch in our environment to see whether it actually solves the problem.

mengxr:
Confirmed this patch fixed the issue in our environment:)

core/src/main/scala/org/apache/spark/api/r/JVMObjectTracker.scala
```diff
@@ -0,0 +1,87 @@
...
+    get(id).getOrElse(
+      throw new NoSuchElementException(s"$id does not exist.")
+    )
+  }
+
+  /**
+   * Adds a JVM object to track and returns assigned ID, which is unique within this tracker.
+   */
+  final def addAndGetId(obj: Object): JVMObjectId = {
+    val id = JVMObjectId(objCounter.getAndIncrement().toString)
```
shivaram: I just noticed you made the `get` synchronized. Can we also make this synchronized ? In that case we will also have solved the thread safety issue

shivaram:
One minor comment about synchronized - LGTM otherwise. And thanks for the RBackend unit test !

core/src/main/scala/org/apache/spark/api/r/JVMObjectTracker.scala
mengxr: Is it necessary? `objCounter` is an `AtomicLong`.

mengxr:
test this please

mengxr:
All test failures seem unrelated.

core/src/main/scala/org/apache/spark/api/r/JVMObjectTracker.scala
shivaram: The counter is atomic - but the map insert is not ?

mengxr: But I don't see any safety issue. Two concurrent calls to `addAndGetId` will never share the same `id` because `getAndIncrement` is atomic and hence `objMap.put` won't overwrite each other. Let me know if I missed something.

shivaram: I was going to say that the keys wont clash, but that it might be unsafe to have two insert functions running at the same time. But I then realized you are using a ConcurrentHashMap, so that should be safe as well. I think the code is fine !

mengxr:
test this please (unrelated test failures)

mengxr:
Merged into master and branch-2.1.

PR #9940, opened 2015-11-24
[SPARK-11955][SQL] Mark optional fields in merging schema for safely pushdowning filters in Parquet

viirya:
JIRA: https://issues.apache.org/jira/browse/SPARK-11955

Currently we simply skip pushdowning filters in parquet if we enable schema merging.

However, we can actually mark particular fields in merging schema for safely pushdowning filters in parquet.

PR #16156, opened 2016-12-05
[SPARK-18539][SQL]: tolerate pushed-down filter on non-existing parquet columns

xwu0226:

When `spark.sql.parquet.filterPushdown` is on, SparkSQL's Parquet reader pushes down filters to Parquet file when creating reader, in order to start with filtered blocks. However, when the parquet file does not have the predicate column(s), Parquet-mr throw exceptions complaining the filter column does not existing. This issue will be fixed in parquet-mr 1.9. But Spark 2.1 is still on parquet 1.8. 

This PR is to tolerate such exception thrown by Parquet-mr and just return all the blocks from the current parquet file to the created SparkSQL parquet reader. Filters will be applied again anyway in later physical plan operation. According to following example physical plan:

```
== Physical Plan ==
*Project [a#2805, b#2806, c#2807]
+- *Filter ((isnotnull(a#2805) && isnull(c#2807)) && (a#2805 < 2))
  +- *FileScan parquet [a#2805,b#2806,c#2807] Batched: true, Format: ParquetFormat, Location: InMemoryFileIndex[file:/Users/xinwu/spark/target/tmp/spark-ed6f0c12-6494-4ac5-b485-5b986ef475cc], PartitionFilters: [], PushedFilters: [IsNotNull(a), IsNull(c), LessThan(a,2)], ReadSchema: struct<a:int,b:string,c:int>
```

## How was this patch tested?
A unit test case is added.

### DIFF
--- a/sql/core/src/main/java/org/apache/spark/sql/execution/datasources/parquet/SpecificParquetRecordReaderBase.java
+++ b/sql/core/src/main/java/org/apache/spark/sql/execution/datasources/parquet/SpecificParquetRecordReaderBase.java
@@ -107,7 +107,13 @@ public abstract class SpecificParquetRecordReaderBase<T> extends RecordReader<Vo
       footer = readFooter(configuration, file, range(split.getStart(), split.getEnd()));
       MessageType fileSchema = footer.getFileMetaData().getSchema();
       FilterCompat.Filter filter = getFilter(configuration);
-      blocks = filterRowGroups(filter, footer.getBlocks(), fileSchema);
+      try {
+        blocks = filterRowGroups(filter, footer.getBlocks(), fileSchema);
+      } catch (IllegalArgumentException e) {
+        // In the case where a particular parquet files does not contain
+        // the column(s) in the filter, we don't do filtering at this level
+        blocks = footer.getBlocks();
+      }
     } else {
       // otherwise we find the row groups that were selected on the client
       footer = readFooter(configuration, file, NO_FILTER);
--- a/sql/core/src/main/java/org/apache/spark/sql/execution/datasources/parquet/SpecificParquetRecordReaderBase.java
+++ b/sql/core/src/main/java/org/apache/spark/sql/execution/datasources/parquet/SpecificParquetRecordReaderBase.java
@@ -112,6 +112,9 @@ public abstract class SpecificParquetRecordReaderBase<T> extends RecordReader<Vo
       } catch (IllegalArgumentException e) {
         // In the case where a particular parquet files does not contain
         // the column(s) in the filter, we don't do filtering at this level
+        // PARQUET-389 will resolve this issue in Parquet 1.9, which may be used
+        // by future Spark versions. This is a workaround for current Spark version.
+        // Also the assumption here is that the predicates will be applied later
         blocks = footer.getBlocks();
       }
     } else {
--- a/sql/core/src/test/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetFilterSuite.scala
+++ b/sql/core/src/test/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetFilterSuite.scala
@@ -634,6 +634,8 @@ class ParquetFilterSuite extends QueryTest with ParquetTest with SharedSQLContex
 
           checkAnswer(readDf2.filter("c is null and a < 2"),
             Row(1, "abc", null) :: Nil)
+
+          readDf2.filter("c is null and a < 2").explain(true)
         }
       }
     }
--- a/sql/core/src/test/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetFilterSuite.scala
+++ b/sql/core/src/test/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetFilterSuite.scala
@@ -578,4 +578,64 @@ class ParquetFilterSuite extends QueryTest with ParquetTest with SharedSQLContex
       // scalastyle:on nonascii
     }
   }
+
+  test("SPARK-18539 - filtered by non-existing parquet column") {
+    Seq("parquet").foreach { format =>
+      withSQLConf(SQLConf.PARQUET_FILTER_PUSHDOWN_ENABLED.key -> "true") {
+        withTempPath { path =>
+          import testImplicits._
+          Seq((1, "abc"), (2, "hello")).toDF("a", "b").write.format(format).save(path.toString)
+
+          // user-specified schema contains nonexistent columns
+          val schema = StructType(
+            Seq(StructField("a", IntegerType),
+              StructField("b", StringType),
+              StructField("c", IntegerType)))
+          val readDf1 = spark.read.schema(schema).format(format).load(path.toString)
+
+          // Read the table without any filter
+          checkAnswer(readDf1, Row(1, "abc", null) :: Row(2, "hello", null) :: Nil)
+          // Read the table with a filter on existing columns
+          checkAnswer(readDf1.filter("a < 2"), Row(1, "abc", null) :: Nil)
+
+          // Read the table with a filter on nonexistent columns
+          checkAnswer(readDf1.filter("c < 2"), Nil)
+          checkAnswer(readDf1.filter("c is not null"), Nil)
+
+          checkAnswer(readDf1.filter("c is null"),
+            Row(1, "abc", null) :: Row(2, "hello", null) :: Nil)
+
+          checkAnswer(readDf1.filter("c is null and a < 2"),
+            Row(1, "abc", null) :: Nil)
+
+          // With another parquet file that contains the column "c"
+          Seq((2, "abc", 3), (3, "hello", 4)).toDF("a", "b", "c")
+            .write.format(format).mode(SaveMode.Append).save(path.toString)
+
+          // Right now, there are 2 parquet files. one with the column "c",
+          // the other does not have it.
+          val readDf2 = spark.read.schema(schema).format(format).load(path.toString)
+          // Read the table without any filter
+          checkAnswer(readDf2,
+            Row(1, "abc", null) :: Row(2, "hello", null) ::
+            Row(2, "abc", 3) :: Row(3, "hello", 4) :: Nil)
+
+          // Read the table with a filter on existing columns
+          checkAnswer(readDf2.filter("a < 2"), Row(1, "abc", null) :: Nil)
+
+          // column "c" is in one parquet file, not in the other parquet file
+          checkAnswer(readDf2.filter("c < 4"),
+            Row(2, "abc", 3) :: Nil)
+          checkAnswer(readDf2.filter("c is not null"),
+            Row(2, "abc", 3) :: Row(3, "hello", 4) :: Nil)
+
+          checkAnswer(readDf2.filter("c is null"),
+            Row(1, "abc", null) :: Row(2, "hello", null) :: Nil)
+
+          checkAnswer(readDf2.filter("c is null and a < 2"),
+            Row(1, "abc", null) :: Nil)
+        }
+      }
+    }
+  }
 }


xwu0226:
@gatorsmile @liancheng Thanks!

sql/core/src/main/java/org/apache/spark/sql/execution/datasources/parquet/SpecificParquetRecordReaderBase.java
```diff
@@ -107,7 +107,16 @@ public void initialize(InputSplit inputSplit, TaskAttemptContext taskAttemptCont
...
-      blocks = filterRowGroups(filter, footer.getBlocks(), fileSchema);
+      try {
+        blocks = filterRowGroups(filter, footer.getBlocks(), fileSchema);
+      } catch (IllegalArgumentException e) {
+        // In the case where a particular parquet files does not contain
+        // the column(s) in the filter, we don't do filtering at this level
+        // PARQUET-389 will resolve this issue in Parquet 1.9, which may be used
+        // by future Spark versions. This is a workaround for current Spark version.
+        // Also the assumption here is that the predicates will be applied later
+        blocks = footer.getBlocks();
```
xwu0226: Yes. Will do. Thanks!

sql/core/src/test/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetFilterSuite.scala
```diff
@@ -578,4 +578,66 @@ class ParquetFilterSuite extends QueryTest with ParquetTest with SharedSQLContex
       // scalastyle:on nonascii
     }
   }
+
+  test("SPARK-18539 - filtered by non-existing parquet column") {
+    Seq("parquet").foreach { format =>
```
gatorsmile: Since this is the parquet-only test case, we can just set format to `parquet` directly.

sql/core/src/test/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetFilterSuite.scala
```diff
@@ -578,4 +578,66 @@ class ParquetFilterSuite extends QueryTest with ParquetTest with SharedSQLContex
...
+          checkAnswer(readDf2.filter("c is not null"),
+            Row(2, "abc", 3) :: Row(3, "hello", 4) :: Nil)
+
+          checkAnswer(readDf2.filter("c is null"),
+            Row(1, "abc", null) :: Row(2, "hello", null) :: Nil)
+
+          checkAnswer(readDf2.filter("c is null and a < 2"),
+            Row(1, "abc", null) :: Nil)
+
+          readDf2.filter("c is null and a < 2").explain(true)
```
gatorsmile: Nit: remove it. Thanks!

sql/core/src/test/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetFilterSuite.scala
xwu0226: Ok. Thanks!

sql/core/src/test/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetFilterSuite.scala
```diff
@@ -578,4 +578,66 @@ class ParquetFilterSuite extends QueryTest with ParquetTest with SharedSQLContex
...
+          checkAnswer(readDf1.filter("c is null and a < 2"),
+            Row(1, "abc", null) :: Nil)
+
+          // With another parquet file that contains the column "c"
+          Seq((2, "abc", 3), (3, "hello", 4)).toDF("a", "b", "c")
+            .write.format(format).mode(SaveMode.Append).save(path.toString)
+
+          // Right now, there are 2 parquet files. one with the column "c",
+          // the other does not have it.
+          val readDf2 = spark.read.schema(schema).format(format).load(path.toString)
```
liancheng: By default, we don't enable Parquet schema merging. Could you please reorganize this test case into something like this to have both code paths tested?

```scala
Seq(true, false).foreach { merge =>
  test("SPARK-18539 - filtered by non-existing parquet column - schema merging enabled: $merge") {
    // ...
    val readDf2 = spark.read
      .option("mergeSchema", merge.toString)
      .schema(schema)
      .format(format)
      .load(path.toString)
    // ...
  }
}
```

xwu0226: I will add it. Thanks!

liancheng: In fact, we probably also want to test both the normal Parquet reader and the vectorized Parquet reader using the similar trick by setting  `SQLConf.PARQUET_VECTORIZED_READER_ENABLED.key`.

liancheng:
Actually, PR #9940 should have already fixed this issue. I'm checking why it doesn't work under 2.0.1 for 2.0.2.

liancheng:
BTW, I think this PR is a cleaner fix than #9940, which introduces a temporary metadata while merging two `StructType`s and erased it in a later phase. We may want to remove the hack done in #9940 later if possible. But for now, let's make the fix as surgical as possible to lower the risk for 2.1 release.

sql/core/src/test/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetFilterSuite.scala
xwu0226: Since the code change is in `SpecificParquetRecordReaderBase.initialize`, which is a common code path for both parquet readers, I did not include it originally. But it is safe to test both. I will add it . Thanks!

sql/core/src/main/java/org/apache/spark/sql/execution/datasources/parquet/SpecificParquetRecordReaderBase.java
```diff
@@ -107,7 +107,16 @@ public void initialize(InputSplit inputSplit, TaskAttemptContext taskAttemptCont
...
       FilterCompat.Filter filter = getFilter(configuration);
-      blocks = filterRowGroups(filter, footer.getBlocks(), fileSchema);
+      try {
+        blocks = filterRowGroups(filter, footer.getBlocks(), fileSchema);
+      } catch (IllegalArgumentException e) {
+        // In the case where a particular parquet files does not contain
+        // the column(s) in the filter, we don't do filtering at this level
+        // PARQUET-389 will resolve this issue in Parquet 1.9, which may be used
+        // by future Spark versions. This is a workaround for current Spark version.
+        // Also the assumption here is that the predicates will be applied later
```
dongjoon-hyun: What does this mean? PushedFilter seems to be eaten silently here.
```
// Also the assumption here is that the predicates will be applied later
```

sql/core/src/main/java/org/apache/spark/sql/execution/datasources/parquet/SpecificParquetRecordReaderBase.java
dongjoon-hyun: What about creating the JIRA issue first and writing the JIRA number here together in comment?

sql/core/src/main/java/org/apache/spark/sql/execution/datasources/parquet/SpecificParquetRecordReaderBase.java
liancheng: No matter filters are pushed down to Parquet reader or not, Spark will always apply all the filters again at a higher level to ensure that all filters are applied as expected. Spark treats data source filter push-down in a "best effort" manner.

dongjoon-hyun: I see! Thank you, @liancheng .

xwu0226: The filters pushed down to parquet reader and fail Parquet's schema validation will be ignored at this level. But after SparkSQL's parquet reader is created with all the blocks for this particular parquet file, the parent physical plan operation `Filter` will re-apply the filters anyway. 

Example:
```
== Physical Plan ==
*Project [a#2805, b#2806, c#2807]
+- *Filter ((isnotnull(a#2805) && isnull(c#2807)) && (a#2805 < 2))
  +- *FileScan parquet [a#2805,b#2806,c#2807] Batched: true, Format: ParquetFormat, Location: InMemoryFileIndex[file:/Users/xinwu/spark/target/tmp/spark-ed6f0c12-6494-4ac5-b485-5b986ef475cc], PartitionFilters: [], PushedFilters: [IsNotNull(a), IsNull(c), LessThan(a,2)], ReadSchema: struct<a:int,b:string,c:int>
```
This is why my test case still works.

sql/core/src/test/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetFilterSuite.scala
liancheng: Yea, what I concern is that we are using the vectorized Parquet reader by default now and may not have enough test coverage of the normal Parquet reader. Would be good to have test cases to prevent regression for the normal Parquet reader code path.

sql/core/src/main/java/org/apache/spark/sql/execution/datasources/parquet/SpecificParquetRecordReaderBase.java
dongjoon-hyun: Thank you, @xwu0226 !

sql/core/src/main/java/org/apache/spark/sql/execution/datasources/parquet/SpecificParquetRecordReaderBase.java
liancheng: Thanks. We may just mention [SPARK-18140][1] here. It's for upgrading parquet-mr to 1.9.

[1]: https://issues.apache.org/jira/browse/SPARK-18140

dongjoon-hyun: +1

sql/core/src/test/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetFilterSuite.scala
gatorsmile: Please add a test case. Thanks!

liancheng:
@xwu0226 Just tested that this issue also affects the normal Parquet reader (by setting `spark.sql.parquet.enableVectorizedReader` to `false`). That's also why #9940 couldn't take a similar approach as this one. Because `ParquetRecordReader` is a 3rd party class provided by parquet-mr.

xwu0226:
@liancheng I see. In normal parquet reader, ParquetFileFormat is using hadoop's `ParquetRecordReader`, which we can not add such toleration code.

xwu0226:
For normal parquet reader case, we have the following code
```Scala
} else {
        logDebug(s"Falling back to parquet-mr")
        // ParquetRecordReader returns UnsafeRow
        val reader = pushed match {
          case Some(filter) =>
            new ParquetRecordReader[UnsafeRow](
              new ParquetReadSupport,
              FilterCompat.get(filter, null))
          case _ =>
            new ParquetRecordReader[UnsafeRow](new ParquetReadSupport)
        }
        reader.initialize(split, hadoopAttemptContext)
        reader
      }
```
I am wondering we could try-catch the` reader.initialize` and recreate create the ParquetRecordReader without the filter and initialize again. 
What do you think?

sql/core/src/main/java/org/apache/spark/sql/execution/datasources/parquet/SpecificParquetRecordReaderBase.java
HyukjinKwon: @liancheng I just wonder if it is safe to rely on exception handling where it might be a common case for reading multiple Parquet files with a merged schema. Up to my understanding, we here know the schema of the target parquet file from its footer and the referenced column from the filter. Would we maybe try to check this and simply use if-else condition?

liancheng:
Hey @xwu0226 @gatorsmile, did some investigation, and I don't think this is a bug now. Please refer to [my JIRA comment][1] for more details.

[1]: https://issues.apache.org/jira/browse/SPARK-18539?focusedCommentId=15723747&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-15723747

HyukjinKwon:
Would there be another way to avoid try-catch? I think it is a normal reading path logic and it seems it might not be safe to rely on exception handling.

HyukjinKwon:
@liancheng Ah, thank you. I should have tested this first.

xwu0226:
https://issues.apache.org/jira/browse/SPARK-19409 is resolved to upgrade to parquet-1.8.2 that fixes this issue.

PR #16157, opened 2016-12-05
[SPARK-18723][DOC] Expanded programming guide information on wholeTex…

michalsenkyr:

Add additional information to wholeTextFiles in the Programming Guide. Also explain partitioning policy difference in relation to textFile and its impact on performance.

Also added reference to the underlying CombineFileInputFormat

## How was this patch tested?

Manual build of documentation and inspection in browser

```
cd docs
jekyll serve --watch
```


docs/programming-guide.md
```diff
@@ -347,7 +347,7 @@ Some notes on reading files with Spark:
 
 Apart from text files, Spark's Scala API also supports several other data formats:
 
-* `SparkContext.wholeTextFiles` lets you read a directory containing multiple small text files, and returns each of them as (filename, content) pairs. This is in contrast with `textFile`, which would return one record per line in each file.
+* `SparkContext.wholeTextFiles` lets you read a directory containing multiple small text files, and returns each of them as (filename, content) pairs. This is in contrast with `textFile`, which would return one record per line in each file. It takes an optional second argument for controlling the minimal number of partitions (by default this is 2). It uses [CombineFileInputFormat](https://hadoop.apache.org/docs/current/api/org/apache/hadoop/mapreduce/lib/input/CombineFileInputFormat.html) internally in order to process large numbers of small files effectively by grouping files on the same node into a single split. (This can lead to non-optimal partitioning. It is therefore advisable to set the minimal number of partitions explicitly.)
```
srowen: (What is a 'node' here in the context of Spark -- executor? I'm also not sure this behavior is guaranteed, from reading the code and docs)

I don't know that the implementation detail matters here as much as what problem the end user might solve by setting this. You might instead say that this can lead to many small files in relatively few partitions, and this is why you might wish to set a minimum. That's kind of what the docs suggest already; maybe this is better as a tiny improvement to the scaladoc?

michalsenkyr: Yes, you are right. A 'node' should probably be an 'executor'. Also 'split' should probably be replaced by 'partition' to prevent confusion. I used Hadoop terminology because it is used in the linked Hadoop documentation, but Spark terminology may be more appropriate here.

As I understand it, this behavior is guaranteed as setting minPartitions directly sets maxSplitSize by dividing the length of the file ([code](https://github.com/apache/spark/blob/39e2bad6a866d27c3ca594d15e574a1da3ee84cc/core/src/main/scala/org/apache/spark/input/WholeTextFileInputFormat.scala#L57)). Please note that [minSplitSize](https://hadoop.apache.org/docs/current/api/org/apache/hadoop/mapreduce/lib/input/CombineFileInputFormat.html#setMinSplitSizeNode-long-) is only used for leftover data.

As the programming guide already contains quite a detailed description of partitioning on `textFile`, I thought it would make sense to mention it in `wholeTextFiles` as well. Especially when the behavior differs that much.

srowen: A few more sentences in the docs here, and possibly the scaladoc, can't hurt. It would call attention to the fact that you may wish to set a minimum, and why you would do that.

docs/programming-guide.md
```diff
@@ -347,7 +347,7 @@ Some notes on reading files with Spark:
 
 Apart from text files, Spark's Scala API also supports several other data formats:
 
-* `SparkContext.wholeTextFiles` lets you read a directory containing multiple small text files, and returns each of them as (filename, content) pairs. This is in contrast with `textFile`, which would return one record per line in each file.
+* `SparkContext.wholeTextFiles` lets you read a directory containing multiple small text files, and returns each of them as (filename, content) pairs. This is in contrast with `textFile`, which would return one record per line in each file. It takes an optional second argument for controlling the minimal number of partitions (by default this is 2). It uses [CombineFileInputFormat](https://hadoop.apache.org/docs/current/api/org/apache/hadoop/mapreduce/lib/input/CombineFileInputFormat.html) internally in order to process large numbers of small files effectively by grouping files on the same executor into a single partition. (This can lead to non-optimal partitioning. It is therefore advisable to set the minimal number of partitions explicitly.)
```
srowen: I'm still not sure this accurate, that you "should" set this. It would be important to set in some contexts but not all, like lots of parameters. I still think promising a particular implementation here serves no purpose. This should just call attention to the parameter and when one would use it.

michalsenkyr:
I added a few more sentences describing the cases in which the user might want to use the argument. However, I am afraid this might be a little too descriptive.

docs/programming-guide.md
```diff
@@ -347,7 +347,7 @@ Some notes on reading files with Spark:
...
+* `SparkContext.wholeTextFiles` lets you read a directory containing multiple small text files, and returns each of them as (filename, content) pairs. This is in contrast with `textFile`, which would return one record per line in each file. It takes an optional second argument for controlling the minimal number of partitions (by default this is 2). It uses [CombineFileInputFormat](https://hadoop.apache.org/docs/current/api/org/apache/hadoop/mapreduce/lib/input/CombineFileInputFormat.html) internally in order to process large numbers of small files effectively by grouping files on the same executor into a single partition. This can lead to sub-optimal partitioning when the file sets would benefit from residing in multiple partitions (e.g., larger partitions would not fit in memory, files are replicated but a large subset is locally reachable from a single executor, subsequent transformations would benefit from multi-core processing). In those cases, set the `minPartitions` argument to enforce splitting.
```
srowen: I don't see value in the `@note` about the implementation - really it's an implementation detail. Adding additional info to doc about the parameter makes some sense, but it can just echo whatever note goes here. 

I think this is a bit complicated. How about: "In some cases, the result may have relatively few partitions containing relatively many files. Consider increasing minPartitions in this case to prefer more, smaller partitions, which may allow for more efficient processing of the result."

michalsenkyr: Also, the number of files in this case is irrelevant. Even with, say, 5 large files we got only 2 partitions by default.

srowen: Every element of the result is a file; it's fundamentally different from `textFile`, and can't be that each file therefore ends up in a partition. I don't think this merges files, right?

srowen:
OK, net-net I am OK merging this doc change even if it's not exactly what I might write. It does contain useful additional detail that's accurate at the moment.

michalsenkyr:
Sorry for the delay. You are probably right that the partitioning is primarily determined by data locality and that it is therefore appropriate in some cases and shouldn't be worded such as to imply it isn't usually appropriate. I will think about how to word it more appropriately and remove the implementation specifics.

srowen (approved):

srowen:
Merged to master

PR #16158, opened 2016-12-05
[SPARK-18724][ML] Add TuningSummary for TrainValidationSplit and CrossValidator

hhbyyh:
jira: https://issues.apache.org/jira/browse/SPARK-18724

Currently TrainValidationSplitModel only provides tuning metrics in the format of Array[Double], which makes it harder for matching the metrics back to the paramMap generating them and affects the user experience for the tuning framework.
Add a Tuning Summary to provide better presentation for the tuning metrics, for now the idea is to use a DataFrame listing all the params and corresponding metrics.

image

The Tuning Summary Class can be further extended for CrossValidator.
We can also add training time statistics and metrics rank to the data frame if that sounds good.

Update:

To support pipeline estimator, change the tuning summary column name to include full param reference:
image

## How was this patch tested?

existing and new unit tests


hhbyyh:
@MLnick Does this match your thoughts?

mllib/src/main/scala/org/apache/spark/ml/tuning/TrainValidationSplit.scala
```diff
@@ -123,7 +124,10 @@ class TrainValidationSplit @Since("1.5.0") (@Since("1.5.0") override val uid: St
     logInfo(s"Best set of parameters:\n${epm(bestIndex)}")
     logInfo(s"Best train validation split metric: $bestMetric.")
     val bestModel = est.fit(dataset, epm(bestIndex)).asInstanceOf[Model[_]]
-    copyValues(new TrainValidationSplitModel(uid, bestModel, metrics).setParent(this))
+    val model = copyValues(new TrainValidationSplitModel(uid, bestModel, metrics).setParent(this))
+    val summary = new TuningSummary(bestModel.transform(dataset), epm, metrics, bestIndex)
```
MLnick: It seems wasteful to do `bestModel.transform(dataset)` just to get access to the `sqlContext`. Is it really necessary?

MLnick (commented):
Will go through in more detail but just a quick comment about needing the `DataFrame` ref in the summary ctor.

mllib/src/main/scala/org/apache/spark/ml/tuning/TrainValidationSplit.scala
hhbyyh: Indeed that's not necessary. I just replaced it with  SparkSession.builder().getOrCreate(). Is there a better way to get the default contexts?

MLnick:
Sorry this slipped! I'd like to revisit soon after 2.2. settles down.

I think we may need to consider how this integrates with training / evaluation summaries to create a holistic solution (see [SPARK-19053](https://issues.apache.org/jira/browse/SPARK-19053))

hhbyyh:
@MLnick Thanks for your attention. I'm not sure if SPARK-19053 is still active and maybe it's not a blocking issue for this change. If you don't mind, I'll extend the jira/PR scope to involve CrossValidator to have an integrated improvement.

MLnick:
Yeah maybe do the CV one in this PR too.

mllib/src/main/scala/org/apache/spark/ml/tuning/TrainValidationSplit.scala
```diff
@@ -226,6 +230,29 @@ class TrainValidationSplitModel private[ml] (
...
+
+  /**
+   * Gets summary of model on training set. An exception is
+   * thrown if `trainingSummary == None`.
+   */
+  @Since("2.0.0")
+  def summary: TuningSummary = trainingSummary.getOrElse {
+    throw new SparkException(
+      s"No training summary available for the ${this.getClass.getSimpleName}")
+  }
```
hhbyyh: addressed in https://github.com/apache/spark/pull/17654

hhbyyh:
add tuning summary for crossValidator.

mllib/src/main/scala/org/apache/spark/ml/tuning/TrainValidationSplit.scala
```diff
@@ -231,6 +235,29 @@ class TrainValidationSplitModel private[ml] (
...
+
+  /**
+   * Return true if there exists summary of model.
+   */
+  @Since("2.3.0")
+  def hasSummary: Boolean = trainingSummary.nonEmpty
+
+  /**
+   * Gets summary of model on training set. An exception is
```
MLnick: Should probably rather be "summary of model performance on the validation set"?

mllib/src/main/scala/org/apache/spark/ml/tuning/CrossValidator.scala
```diff
@@ -229,6 +233,29 @@ class CrossValidatorModel private[ml] (
...
+  }
+
+  /**
+   * Return true if there exists summary of model.
+   */
+  @Since("2.3.0")
+  def hasSummary: Boolean = trainingSummary.nonEmpty
+
+  /**
+   * Gets summary of model on training set. An exception is
```
MLnick: Likewise, "cross-validation performance of each model" or similar?

mllib/src/main/scala/org/apache/spark/ml/tuning/CrossValidator.scala
```diff
@@ -133,7 +134,10 @@ class CrossValidator @Since("1.2.0") (@Since("1.4.0") override val uid: String)
     logInfo(s"Best cross-validation metric: $bestMetric.")
     val bestModel = est.fit(dataset, epm(bestIndex)).asInstanceOf[Model[_]]
     instr.logSuccess(bestModel)
-    copyValues(new CrossValidatorModel(uid, bestModel, metrics).setParent(this))
+    val model = new CrossValidatorModel(uid, bestModel, metrics).setParent(this)
+    val summary = new TuningSummary(epm, metrics, bestIndex)
+    model.setSummary(Some(summary))
```
MLnick: Just to confirm, the tuning summary will not be saved? Since it's a small dataframe, perhaps we should consider saving it with the model? (Can do that in a later PR however)

mllib/src/main/scala/org/apache/spark/ml/tuning/TuningSummary.scala
```diff
@@ -0,0 +1,58 @@
...
+   * Summary of grid search tuning in the format of DataFrame. Each row contains one candidate
+   * paramMap and its corresponding metric.
+   */
+  def trainingMetrics: DataFrame = {
+    require(params.nonEmpty, "estimator param maps should not be empty")
+    require(params.length == metrics.length, "estimator param maps number should match metrics")
+    val spark = SparkSession.builder().getOrCreate()
+    val sqlContext = spark.sqlContext
+    val sc = spark.sparkContext
+    val fields = params(0).toSeq.sortBy(_.param.name).map(_.param.name) ++ Seq("metrics")
```
MLnick: "metrics" is a bit generic. Perhaps it's better (and more user-friendly) to make this be something like `metric_name metric` so that it's obvious what metric was being optimized for? such as `ROC metric` or `AUC metric` or `MSE metric`? etc

mllib/src/main/scala/org/apache/spark/ml/tuning/TuningSummary.scala
```diff
@@ -0,0 +1,58 @@
...
+ * @param params  ParamMaps for the Estimator
+ * @param metrics  corresponding evaluation metrics for the params
+ * @param bestIndex  index in params for the ParamMap of the best model.
+ */
+@Since("2.3.0")
+@Experimental
+private[tuning] class TuningSummary private[tuning](
+    private[tuning] val params: Array[ParamMap],
+    private[tuning] val metrics: Array[Double],
+    private[tuning] val bestIndex: Int) {
```
MLnick: It appears `bestIndex` is never used?

mllib/src/test/scala/org/apache/spark/ml/tuning/CrossValidatorSuite.scala
```diff
@@ -66,6 +66,29 @@ class CrossValidatorSuite
...
+    assert(cvModel.hasSummary)
+    assert(cvModel.summary.params === lrParamMaps)
+    assert(cvModel.summary.trainingMetrics.count() === lrParamMaps.length)
+
+    val expected = lrParamMaps.zip(cvModel.avgMetrics).map { case (map, metric) =>
+      Row.fromSeq(map.toSeq.sortBy(_.param.name).map(_.value.toString) ++ Seq(metric.toString))
+    }
+    assert(cvModel.summary.trainingMetrics.collect().toSet === expected.toSet)
+  }
+
```
MLnick: Shall we add a test for the exception being thrown if no summary?

MLnick:
@hhbyyh sorry for the delay. Left a few review comments. 

Tested the examples and it looks cool! Very useful

mllib/src/main/scala/org/apache/spark/ml/tuning/CrossValidator.scala
hhbyyh: If we want to just save the tuning summary in the model, Perhaps we can just discard the TuningSummary, and add a tuningSummary field in the models. Sounds good?

MLnick: Are there other obvious things that might go into the summary in future, that would make a `TuningSummary` class a better fit?

Future support for say, multiple metrics, could simply extend the dataframe columns so that is ok. But is there anything else you can think of?

hhbyyh: There might be something like detailed training log and training time for each model. But I'm thinking the current Summary pattern does have some room for improvement (e.g., save/load and API), it makes me feel bad when I have to duplicate the code like 
`def hasSummary: Boolean = trainingSummary.nonEmpty`. Thus saving it to the models sounds like a good idea to me.

hhbyyh:
Move the tuningSummary to Models, and updated the name of the metrics column. 
image

mllib/src/main/scala/org/apache/spark/ml/tuning/ValidatorParams.scala
```diff
@@ -85,6 +86,32 @@ private[ml] trait ValidatorParams extends HasSeed with Params {
...
+      case _ => "metrics"
+    }
+    val spark = SparkSession.builder().getOrCreate()
+    val sc = spark.sparkContext
+    val fields = params(0).toSeq.sortBy(_.param.name).map(_.param.name) ++ Seq(metricName)
+    val schema = new StructType(fields.map(name => StructField(name, StringType)).toArray)
+    val rows = sc.parallelize(params.zip(metrics)).map { case (param, metric) =>
+      val values = param.toSeq.sortBy(_.param.name).map(_.value.toString) ++ Seq(metric.toString)
+      Row.fromSeq(values)
+    }
```
WeichenXu123: Here the var names is a little confusing,
`params` ==> `paramMaps`
`case (param, metric)` ==> `case (paramMap, metric)`
will be more clear.

mllib/src/main/scala/org/apache/spark/ml/tuning/ValidatorParams.scala
```diff
@@ -85,6 +86,32 @@ private[ml] trait ValidatorParams extends HasSeed with Params {
...
+      case m: MulticlassClassificationEvaluator => m.getMetricName
+      case r: RegressionEvaluator => r.getMetricName
+      case _ => "metrics"
+    }
+    val spark = SparkSession.builder().getOrCreate()
+    val sc = spark.sparkContext
+    val fields = params(0).toSeq.sortBy(_.param.name).map(_.param.name) ++ Seq(metricName)
+    val schema = new StructType(fields.map(name => StructField(name, StringType)).toArray)
+    val rows = sc.parallelize(params.zip(metrics)).map { case (param, metric) =>
+      val values = param.toSeq.sortBy(_.param.name).map(_.value.toString) ++ Seq(metric.toString)
```
WeichenXu123: Here seems exists a problem:
Suppose `params(0)` (which is a `ParamMap`) contains ParamA and ParamB,
and `params(1)` (which is a `ParamMap`) contains ParamA and ParamC,
The code here will run into problems. Because you compose the row values sorted by param name but do not check whether every row exactly match the first row.
I think better way is, go though the whole `ParamMap` list and collect all params used, and sort them by name, as the dataframe schema.

WeichenXu123: And here use `param_value.toString`, some array type param will convert to unreadable string.
For example, `DoubleArrayParam`, doubleArray.toString will became "[DXXXXX"
use `Param.jsonEncode` is better.

WeichenXu123 (commented):
Thanks for the PR ! I leave some comments.

mllib/src/main/scala/org/apache/spark/ml/tuning/ValidatorParams.scala
hhbyyh: Thanks, we should support the case for custom paramMap.

mllib/src/main/scala/org/apache/spark/ml/tuning/ValidatorParams.scala
hhbyyh: OK

hhbyyh:
Update:

To support pipeline estimator, change the tuning summary column name to include full param reference:
image

mllib/src/main/scala/org/apache/spark/ml/tuning/CrossValidator.scala
hhbyyh: The latest implementation does not need to save the extra dataframe. Since basically the dataframe can be generated from $(estimatorParamMaps) and avgMetrics.

hhbyyh:
Please advice if this is a good feature to add. If not I'll close it. Thanks.

WeichenXu123:
@hhbyyh This PR is stale. If there's nobody interested in this and no further updates, would you mind to close it ? Thanks!

PR #16160, opened 2016-12-05
[SPARK-18721][SS]Fix ForeachSink with watermark + append

zsxwing:

Right now ForeachSink creates a new physical plan, so StreamExecution cannot retrieval metrics and watermark.

This PR changes ForeachSink to manually convert InternalRows to objects without creating a new plan.

## How was this patch tested?

`test("foreach with watermark: append")`.

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/ForeachSink.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/ForeachSink.scala
@@ -18,9 +18,8 @@
 package org.apache.spark.sql.execution.streaming
 
 import org.apache.spark.TaskContext
-import org.apache.spark.rdd.RDD
-import org.apache.spark.sql.{DataFrame, Dataset, Encoder, ForeachWriter}
-import org.apache.spark.sql.catalyst.plans.logical.CatalystSerde
+import org.apache.spark.sql.{DataFrame, Encoder, ForeachWriter}
+import org.apache.spark.sql.catalyst.encoders.encoderFor
 
 /**
  * A [[Sink]] that forwards all data into [[ForeachWriter]] according to the contract defined by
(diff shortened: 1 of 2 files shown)


sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/ForeachSink.scala
```diff
@@ -32,46 +31,26 @@ import org.apache.spark.sql.catalyst.plans.logical.CatalystSerde
 class ForeachSink[T : Encoder](writer: ForeachWriter[T]) extends Sink with Serializable {
 
   override def addBatch(batchId: Long, data: DataFrame): Unit = {
-    // TODO: Refine this method when SPARK-16264 is resolved; see comments below.
```
zsxwing: `SPARK-16264` was resolved as `Won't Fix`. So I removed it from the comment.

sql/core/src/test/scala/org/apache/spark/sql/execution/streaming/ForeachSinkSuite.scala
```diff
@@ -204,6 +204,55 @@ class ForeachSinkSuite extends StreamTest with SharedSQLContext with BeforeAndAf
       query.stop()
     }
   }
+
+  test("foreach with watermark: append") {
```
tdas: what does this test that is not covered in the previous test "watermark + complete"?

sql/core/src/test/scala/org/apache/spark/sql/execution/streaming/ForeachSinkSuite.scala
```diff
@@ -204,6 +204,55 @@ class ForeachSinkSuite extends StreamTest with SharedSQLContext with BeforeAndAf
...
+          ForeachSinkSuite.Open(partition = 0, version = 2),
+          ForeachSinkSuite.Process(value = 3),
+          ForeachSinkSuite.Close(None)
+        )
+      )
+      assert(allEvents === expectedEvents)
+    } finally {
+      query.stop()
+    }
+  }
```
tdas: I dont see a test that verifies whether the metrics are correct

tdas (changes requested):
Needs a bit more work with the tests

sql/core/src/test/scala/org/apache/spark/sql/execution/streaming/ForeachSinkSuite.scala
zsxwing: @tdas As no eviction with `complete` mode, it will always output all data. So basically, the test "watermark + complete" is not super helpful.

tdas: NVM. watermark is a noop in complete mode. false alarm.

tdas:
LGTM.

zsxwing:
Forgot that `lastProgress` may be the batch without data. Updated the test to drop no data progress.

PR #16161, opened 2016-12-05
[SPARK-18717][SQL] Make code generation for Scala Map work with immutable.Map also

aray:

Fixes compile errors in generated code when user has case class with a `scala.collections.immutable.Map` instead of a `scala.collections.Map`. Since ArrayBasedMapData.toScalaMap returns the immutable version we can make it work with both.

## How was this patch tested?

Additional unit tests.


sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/encoders/RowEncoder.scala
```diff
@@ -288,7 +287,7 @@ object RowEncoder {
 
       StaticInvoke(
         ArrayBasedMapData.getClass,
-        ObjectType(classOf[Map[_, _]]),
+        ObjectType(classOf[scala.collection.immutable.Map[_, _]]),
```
srowen: I don't know this bit well, but the comments on the JIRA indicate that immutable maps aren't supported. However, this change should be a no-op because Predef.Map is already immutable.Map right -- how does this address it?

rxin:
cc @cloud-fan

cloud-fan:
can you highlight your approach to fix this bug? It looks to me that this is not a general fix, e.g. `mutable.Map` will still fail.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/encoders/RowEncoder.scala
viirya: @srowen `scala.collection.Map` is imported at the beginning of `RowEncoder` and `ScalaReflection`.

aray:
The approach is to change the deserializer (via `ScalaReflection#deserializerFor`) to return the more specific type `scala.collections.immutable.Map` instead of `scala.collections.Map` as it does now (no change to `ArrayBasedMapData#toScalaMap` is needed as it already returns an immutable map). This allows the generated code for case classes using either of those Map's to work as they are both assignable from `scala.collections.immutable.Map`. This PR does not address `scala.collections.mutable.Map`, it still wont work.

srowen:
OK, it sounds like what you're doing is unsupported at the moment and this doesn't change that, and you have a workaround. Is this worth merging?

aray:
Right now it's not supported to have the following:
```
case class Foo(a: Map[Int, Int])
```
(using the scala Predef version of Map)

The [documented](http://spark.apache.org/docs/latest/sql-programming-guide.html#data-types) way to do this is:
```
case class Foo(a: scala.collections.Map[Int, Int])
```
and it works fine.

However if someone did not read the documentation carefully and used the former, instead of a reasonable error they get a compile error on the code spark generates. Therefore IMHO there are two options:

1. Make the generated code work with either version of Map (this PR)
2. Intercept definitions that use the wrong version of Map and issue a proper error.

If the consensus is that this PR is not worth it then I'll be happy to work on option 2. But in my opinion as a Spark user option 1 is better.

cloud-fan:
option 1 is better, but this PR only adds the support for `scala.collection.immutable.Map`, with a very specific way: we can't extend it to support `scala.collection.mutable.Map` or `List`.

I think we need a general solution to support custom collection type.

aray:
I would be happy to create a seperate PR for adding support for `mutable.Map` (and `List`) if that is wanted. But there is no _generic_ solution as there is no type that is assignable to both `mutable.Map` and `immutable.Map`.

cloud-fan:
> there is no type that is assignable to both mutable.Map and immutable.Map

Yea, and a general solution would be, remember the original type and create that type when converting catalyst value to external value.

But this may be hard to implement, it may worth to merge this simple PR, as the `scala.collection.immutable.Map` is so common in scala. cc @rxin @yhuai @liancheng

cloud-fan:
thanks, merging to master!

liancheng:
Shall we backport this to branch-2.1? I'd consider this as a bug because of the following snippet fail in Spark 2.1:

```scala
case class Wrapper1(value: Option[Map[String, String]])
case class Wrapper2(value: Map[String, String])

val ds1 = Seq.empty[Wrapper1].toDS()  // Fail
val ds2 = Seq.empty[Wrapper2].toDS()  // Succeed
```

cloud-fan:
I'm fine to backport this

liancheng:
Thanks. Backported to branch-2.1.

PR #16163, opened 2016-12-06
[SPARK-18730] Post Jenkins test report page instead of the full console output page to GitHub

liancheng:

Currently, the full console output page of a Spark Jenkins PR build can be as large as several megabytes. It takes a relatively long time to load and may even freeze the browser for quite a while.

This PR makes the build script to post the test report page link to GitHub instead. The test report page is way more concise and is usually the first page I'd like to check when investigating a Jenkins build failure.

Note that for builds that a test report is not available (ongoing builds and builds that fail before test execution), the test report link automatically redirects to the build page.

## How was this patch tested?

N/A.

### DIFF
--- a/dev/run-tests-jenkins.py
+++ b/dev/run-tests-jenkins.py
@@ -80,7 +80,7 @@ def pr_message(build_display_name,
                 short_commit_hash,
                 commit_url,
                 str(' ' + post_msg + '.') if post_msg else '.')
-    return '**[Test build %s %s](%sconsoleFull)** for PR %s at commit [`%s`](%s)%s' % str_args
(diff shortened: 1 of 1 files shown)


srowen:
It might not be exactly the same thing, but I remember we changed something like this in the opposite way -- to link straight to the full output since typically you need that to debug a failure, and loading the brief page is just an extra hop.

I don't feel strongly about it, just dumping an anecdote from the back of my brain.

liancheng:
@srowen Thanks. I sent this one because the `consoleFull` page frequently freezes my browser these days, not mentioning viewing Jenkins build results via mobile phone...

rxin:
Alright let's try this ... Merging in master.

PR #16164, opened 2016-12-06
[SPARK-18732][WEB-UI] The Y axis ranges of "schedulingDelay", "processingTime", and "totalDelay" should not keep the same.

uncleGen:

Currently, the Y axis ranges of "schedulingDelay", "processingTime", and "totalDelay" keeps the same. It is not convenient for users to see, like following：
image
 
This PR separates the Y axis ranges of "schedulingDelay", "processingTime", and "totalDelay".

## How was this patch tested?

existing ut

### DIFF
--- a/streaming/src/main/scala/org/apache/spark/streaming/ui/StreamingPage.scala
+++ b/streaming/src/main/scala/org/apache/spark/streaming/ui/StreamingPage.scala
@@ -229,22 +229,10 @@ private[ui] class StreamingPage(parent: StreamingTab)
       batchInfo.totalDelay.map(batchInfo.batchTime.milliseconds -> _)
     })
 
-    // Use the max value of "schedulingDelay", "processingTime", and "totalDelay" to make the
-    // Y axis ranges same.
-    val _maxTime =
-      (for (m1 <- schedulingDelay.max; m2 <- processingTime.max; m3 <- totalDelay.max) yield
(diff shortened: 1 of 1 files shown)


srowen:
CC @zsxwing because it works this way on purpose, so that one can compare the graphs visually. Usually these values aren't too different in scale; it's a problem here because scheduling delay is unusually large.

uncleGen:
@srowen Indeed, it is not a normal case. And I found this problem when the streaming job went wrong. As you said

> one can compare the graphs visually.

It still may mislead users in some cases, like 'scheduling delay' in 'ms' and 'processing time' in 's' or 'min'

srowen:
Aren't you arguing against letting these vary independently then? I'm not sure we should change this.

uncleGen:
cc @zsxwing

zsxwing:
It's on purpose.

> It still may misleads users in some cases, like 'scheduling delay' in 'ms' and 'processing time' in 's' or 'min'

I don't get it. I saw the time units in your screenshot are same: `min`. And from your current screenshot, it's easy to know that there must be something wrong.

uncleGen:
Wait for a moment, I will provide a better case :)

srowen:
@uncleGen do you have a follow-up on this?

uncleGen:
@srowen Indeed, there is no need to change current code, because this happens when something  goes wrong. I will close this PR.

PR #16165, opened 2016-12-06
[SPARK-8617] [WEBUI] HistoryServer: Include in-progress files during cleanup

seyfe:
- Removed the`attempt.completed ` filter so cleaner would include the orphan inprogress files.
- Use loading time for inprogress files as lastUpdated. Keep using the modTime for completed files. First one will prevent deletion of inprogress job files. Second one will ensure that lastUpdated time won't change for completed jobs in an event of HistoryServer reboot. 

## How was this patch tested?
Added new unittests and via existing tests.

### DIFF
--- a/core/src/main/scala/org/apache/spark/deploy/history/FsHistoryProvider.scala
+++ b/core/src/main/scala/org/apache/spark/deploy/history/FsHistoryProvider.scala
@@ -461,7 +461,7 @@ private[history] class FsHistoryProvider(conf: SparkConf, clock: Clock)
           appListener.appAttemptId,
           appListener.startTime.getOrElse(-1L),
           appListener.endTime.getOrElse(-1L),
-          fileStatus.getModificationTime(),
+          clock.getTimeMillis(),
           appListener.sparkUser.getOrElse(NOT_STARTED),
           appCompleted,
           fileStatus.getLen()
--- a/core/src/main/scala/org/apache/spark/deploy/history/FsHistoryProvider.scala
+++ b/core/src/main/scala/org/apache/spark/deploy/history/FsHistoryProvider.scala
@@ -446,8 +446,12 @@ private[history] class FsHistoryProvider(conf: SparkConf, clock: Clock)
       }
 
       val logPath = fileStatus.getPath()
-
       val appCompleted = isApplicationCompleted(fileStatus)
+
+      // Use loading time as lastUpdated since some filesystems don't update modifiedTime
+      // each time file is updated. However use modifiedTime for completed jobs so lastUpdated
+      // won't change whenever HistoryServer restarts and reloads the file.
+      val lastUpdated = if (appCompleted) fileStatus.getModificationTime else clock.getTimeMillis()
 
       val appListener = replay(fileStatus, appCompleted, new ReplayListenerBus(), eventsFilter)
 
@@ -461,7 +465,7 @@ private[history] class FsHistoryProvider(conf: SparkConf, clock: Clock)
           appListener.appAttemptId,
           appListener.startTime.getOrElse(-1L),
           appListener.endTime.getOrElse(-1L),
-          clock.getTimeMillis(),
+          lastUpdated,
           appListener.sparkUser.getOrElse(NOT_STARTED),
           appCompleted,
           fileStatus.getLen()
--- a/core/src/main/scala/org/apache/spark/deploy/history/FsHistoryProvider.scala
+++ b/core/src/main/scala/org/apache/spark/deploy/history/FsHistoryProvider.scala
@@ -540,23 +540,13 @@ private[history] class FsHistoryProvider(conf: SparkConf, clock: Clock)
    */
   private[history] def cleanLogs(): Unit = {
     try {
-      val maxAge = conf.getTimeAsMs("spark.history.fs.cleaner.maxAge", "7d")
-
-      val cleanInProgressFiles =
-        conf.getBoolean("spark.history.fs.cleaner.inProgress.files", false)
-      val inProgressMaxAge =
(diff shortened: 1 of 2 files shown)


srowen:
I don't think it makes sense to expose yet another set of settings for this.
I think the risk of course is that this accidentally cleans up another instance's work in progress.
However if it's quite old, I'd think it's as safe to clean up an in-progress file as any other?

seyfe:
Hi @srowen,

Thanks for the quick feedback. Let me get rid of the on/off knob for inprogress files. Would you like me to remove the maxAge setting for inprogress files as well?

I initially worried about long running jobs (streaming?) but I think even in that case the files will get updated.

ajbozarth:
After the last changes this looks like a good small update, LGTM

vanzin:
ok to test

core/src/main/scala/org/apache/spark/deploy/history/FsHistoryProvider.scala
```diff
@@ -546,7 +546,7 @@ private[history] class FsHistoryProvider(conf: SparkConf, clock: Clock)
       val appsToRetain = new mutable.LinkedHashMap[String, FsApplicationHistoryInfo]()
 
       def shouldClean(attempt: FsApplicationAttemptInfo): Boolean = {
-        now - attempt.lastUpdated > maxAge && attempt.completed
+        now - attempt.lastUpdated > maxAge
```
vanzin: Hmm, I went and took a look and it doesn't seem like this is a safe way of doing this. `lastUpdated` is set to the file's modification time;  if you look at the class comment, there's this bit there:

```
 * The use of log size, rather than simply relying on modification times, is needed to
 * address the following issues
 * - some filesystems do not appear to update the `modtime` value whenever data is flushed to
 * an open file output stream. Changes to the history may not be picked up.
 * - the granularity of the `modtime` field may be 2+ seconds. Rapid changes to the FS can be
 * missed.
```

So using this value is probably not totally safe for long-running applications.

ajbozarth: Nice catch @vanzin, I missed that, but (devil's advocate) should that level of granularity matter for something cleaning, which is measured in days?

vanzin: Yes, because some applications have run times also measured in days.

seyfe: Hi @vanzin , as far as I understand you are not worried about the 2+ seconds granularity but instead worried that for inprogress jobs `lastUpdated` field might never get updated based on below comment:

> some filesystems do not appear to update the `modtime` value whenever data is flushed to an open file output stream.

Do we know that above statement is true? If that's the case, your suggestion (using last loaded time) seems like proper way of solving the issue.

vanzin: Yes, I'm only worried about the first bullet. It's most probably true, since the initial implementation of the SHS only considered file mod time and that was obviously not enough.

vanzin:
retest this please

core/src/main/scala/org/apache/spark/deploy/history/FsHistoryProvider.scala
```diff
@@ -461,7 +461,7 @@ private[history] class FsHistoryProvider(conf: SparkConf, clock: Clock)
           appListener.appAttemptId,
           appListener.startTime.getOrElse(-1L),
           appListener.endTime.getOrElse(-1L),
-          fileStatus.getModificationTime(),
+          clock.getTimeMillis(),
```
vanzin: As an attempt at keeping this value a little more a reflection of reality, I think it would be better to keep `fileStatus.getModificationTime()` if the application is complete. Otherwise, for complete applications, this will reflect the time the history server started in a lot of cases.

seyfe: Hi @vanzin, we can achieve this by simply doing `if (appCompleted) fileStatus.getModificationTime else clock.getTimeMillis()`.

But I am worried about this scenario: If we assume that modtime will not change when job is completed, the last updated time will move forward while job is in progress and it will go back when job is completed.

But then I found [this code](https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/scheduler/EventLoggingListener.scala#L230) which is executed when we complete the job. We call `fileSystem.setTimes` after `fileSystem.rename`. 

So seems like my assumption is not a valid case. Let me think about it more/add more tests etc..

vanzin: After the job is finished, the file will be closed and I hope that at that time the FS will update the mod time. The issue the comment in the class is talking about is the mod time not being update while the file is still open.

vanzin:
Code LGTM but can you update the description to reflect the recent changes to the PR?

seyfe:
Thank @vanzin. Updated the description based on final state.

vanzin:
Merging to master.

merlintang:
should we backport this into 2.1? @vanzin

vanzin:
2.1 has already shipped and this is not a regression.

merlintang:
@vanzin sorry, I mean the 2.1.1

markhamstra:
@merlintang Marcelo's point remains the same for 2.1.1. We don't typically backport changes to maintenance branches unless they are fixes for regression errors or severe bugs.

merlintang:
@markhamstra  Thanks all. 

btw: what if there are many redundant inprogress files in the disk and impact the system performance?

PR #16168, opened 2016-12-06
[SPARK-18209][SQL] More robust view canonicalization without full SQL expansion

jiangxb1987:

Currently we canonicalize the view definition to provide the context for the database as well as star expansion, this is fragile for the following reasons:

1. It is non-trivial to guarantee that the generated SQL is correct without being extremely verbose, given the current set of operators.
2. We need extensive testing for all combination of operators.
3. Whenever we introduce a new logical plan operator, we need to be super careful because it might break SQL generation. 

We use a late binding approach to do view canonicalization, which resolve the dependent views every time the view is used. This PR is expected to achieve the following goals:

1. Views should reflect changes to the underlying tables and intermediate views. For example:
    ```
        CREATE TABLE T1(a int, b string)
        CREATE TABLE T2(a int, b string, c int)
        CREATE VIEW A AS SELECT * FROM T1
        CREATE VIEW B AS SELECT * FROM A
        ALTER VIEW A AS SELECT * FROM T2
        SELECT * FROM B
    ```
will return data from table T2 (Note that we still need to validate the schema).
2. Make views more robust by not using SQL generation to store the view’s logical plan.
3. The views generated by older versions of Spark or HIVE should still work or throw a meaningful error if the view definition is invalid.
4. A view should throw a meaningful error when the view in an invalid state due to an underlying change. Such errors should be thrown as early as possible.

The follow up works will be:

1. Remove the SQL generation (in particular the SQLBuilder class) for Spark SQL operators and the test suites(and also golden files) for them.
2. Disallow cyclic view reference(See [SPARK-18389](https://issues.apache.org/jira/browse/SPARK-18389)).
3. View dependency management similar to Postgres.
4. Underlying table schema change when star is used.
5. Support create permanent views on non-SQL DataFrames.

## How was this patch tested?
Add new test cases in `SQLViewSuite`.

### DIFF
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
@@ -22,7 +22,7 @@ import scala.collection.mutable.ArrayBuffer
 
 import org.apache.spark.sql.AnalysisException
 import org.apache.spark.sql.catalyst.{CatalystConf, ScalaReflection, SimpleCatalystConf}
-import org.apache.spark.sql.catalyst.catalog.{CatalogDatabase, InMemoryCatalog, SessionCatalog}
+import org.apache.spark.sql.catalyst.catalog._
 import org.apache.spark.sql.catalyst.encoders.OuterScopes
 import org.apache.spark.sql.catalyst.expressions._
 import org.apache.spark.sql.catalyst.expressions.aggregate._
@@ -509,32 +509,42 @@ class Analyzer(
    * Replaces [[UnresolvedRelation]]s with concrete relations from the catalog.
    */
   object ResolveRelations extends Rule[LogicalPlan] {
-    private def lookupTableFromCatalog(u: UnresolvedRelation): LogicalPlan = {
+    private def lookupTableFromCatalog(
+        u: UnresolvedRelation, db: Option[String] = None): LogicalPlan = {
       try {
-        catalog.lookupRelation(u.tableIdentifier, u.alias)
+        catalog.lookupRelation(u.tableIdentifier, u.alias, db)
       } catch {
         case _: NoSuchTableException =>
           u.failAnalysis(s"Table or view not found: ${u.tableName}")
       }
     }
 
-    def apply(plan: LogicalPlan): LogicalPlan = plan resolveOperators {
-      case i @ InsertIntoTable(u: UnresolvedRelation, parts, child, _, _) if child.resolved =>
-        i.copy(table = EliminateSubqueryAliases(lookupTableFromCatalog(u)))
-      case u: UnresolvedRelation =>
-        val table = u.tableIdentifier
-        if (table.database.isDefined && conf.runSQLonFile && !catalog.isTemporaryTable(table) &&
+    def apply(plan: LogicalPlan): LogicalPlan = {
+      var currentDatabase = catalog.getCurrentDatabase
+      plan resolveOperators {
+        case i@InsertIntoTable(u: UnresolvedRelation, parts, child, _, _) if child.resolved =>
+          i.copy(table = EliminateSubqueryAliases(lookupTableFromCatalog(u)))
+        case u: UnresolvedRelation =>
+          val table = u.tableIdentifier
+          if (table.database.isDefined && conf.runSQLonFile && !catalog.isTemporaryTable(table) &&
             (!catalog.databaseExists(table.database.get) || !catalog.tableExists(table))) {
-          // If the database part is specified, and we support running SQL directly on files, and
-          // it's not a temporary view, and the table does not exist, then let's just return the
-          // original UnresolvedRelation. It is possible we are matching a query like "select *
-          // from parquet.`/path/to/query`". The plan will get resolved later.
-          // Note that we are testing (!db_exists || !table_exists) because the catalog throws
-          // an exception from tableExists if the database does not exist.
-          u
-        } else {
-          lookupTableFromCatalog(u)
-        }
+            // If the database part is specified, and we support running SQL directly on files, and
+            // it's not a temporary view, and the table does not exist, then let's just return the
+            // original UnresolvedRelation. It is possible we are matching a query like "select *
+            // from parquet.`/path/to/query`". The plan will get resolved later.
+            // Note that we are testing (!db_exists || !table_exists) because the catalog throws
+            // an exception from tableExists if the database does not exist.
+            u
+          } else {
+            val logicalPlan = lookupTableFromCatalog(u, Some(currentDatabase))
+            currentDatabase = logicalPlan.collectFirst {
+              case relation: CatalogRelation if relation.catalogTable.currentDatabase.isDefined =>
+                relation.catalogTable.currentDatabase.get
+            }.getOrElse(currentDatabase)
+
+            logicalPlan
+          }
+      }
     }
   }
 
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/SessionCatalog.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/SessionCatalog.scala
@@ -557,9 +557,12 @@ class SessionCatalog(
    * If the relation is a view, the relation will be wrapped in a [[SubqueryAlias]] which will
    * track the name of the view.
    */
-  def lookupRelation(name: TableIdentifier, alias: Option[String] = None): LogicalPlan = {
+  def lookupRelation(
+      name: TableIdentifier,
+      alias: Option[String] = None,
+      databaseHint: Option[String] = None): LogicalPlan = {
     synchronized {
-      val db = formatDatabaseName(name.database.getOrElse(currentDb))
+      val db = formatDatabaseName(name.database.getOrElse(databaseHint.getOrElse(currentDb)))
       val table = formatTableName(name.table)
       val relationAlias = alias.getOrElse(table)
       if (db == globalTempViewManager.database) {
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/interface.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/interface.scala
@@ -154,6 +154,7 @@ case class CatalogTable(
     tableType: CatalogTableType,
     storage: CatalogStorageFormat,
     schema: StructType,
+    originalSchema: Option[StructType] = None,
     provider: Option[String] = None,
     partitionColumnNames: Seq[String] = Seq.empty,
     bucketSpec: Option[BucketSpec] = None,
@@ -164,6 +165,7 @@ case class CatalogTable(
     stats: Option[Statistics] = None,
     viewOriginalText: Option[String] = None,
     viewText: Option[String] = None,
+    currentDatabase: Option[String] = None,
     comment: Option[String] = None,
     unsupportedFeatures: Seq[String] = Seq.empty,
     tracksPartitionsInCatalog: Boolean = false) {
@@ -221,11 +223,14 @@ case class CatalogTable(
         s"Last Access: ${new Date(lastAccessTime).toString}",
         s"Type: ${tableType.name}",
         if (schema.nonEmpty) s"Schema: ${schema.mkString("[", ", ", "]")}" else "",
+        if (originalSchema.isDefined) s"Original Schema: ${originalSchema.mkString("[", ", ", "]")}"
+        else "",
         if (provider.isDefined) s"Provider: ${provider.get}" else "",
         if (partitionColumnNames.nonEmpty) s"Partition Columns: $partitionColumns" else ""
       ) ++ bucketStrings ++ Seq(
         viewOriginalText.map("Original View: " + _).getOrElse(""),
         viewText.map("View: " + _).getOrElse(""),
+        currentDatabase.map("Database Hint: " + _).getOrElse(""),
         comment.map("Comment: " + _).getOrElse(""),
         if (properties.nonEmpty) s"Properties: $tableProperties" else "",
         if (stats.isDefined) s"Statistics: ${stats.get.simpleString}" else "",
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/command/views.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/command/views.scala
@@ -20,13 +20,13 @@ package org.apache.spark.sql.execution.command
 import scala.util.control.NonFatal
 
 import org.apache.spark.sql.{AnalysisException, Row, SparkSession}
-import org.apache.spark.sql.catalyst.{SQLBuilder, TableIdentifier}
+import org.apache.spark.sql.catalyst.TableIdentifier
 import org.apache.spark.sql.catalyst.analysis.{UnresolvedFunction, UnresolvedRelation}
 import org.apache.spark.sql.catalyst.catalog.{CatalogStorageFormat, CatalogTable, CatalogTableType}
 import org.apache.spark.sql.catalyst.expressions.Alias
 import org.apache.spark.sql.catalyst.plans.QueryPlan
 import org.apache.spark.sql.catalyst.plans.logical.{LogicalPlan, Project}
-import org.apache.spark.sql.types.MetadataBuilder
+import org.apache.spark.sql.types.{MetadataBuilder, StructType}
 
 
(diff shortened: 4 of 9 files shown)


hvanhovell:
cc @rxin @srinathshankar @nsyca

sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveMetastoreCatalog.scala
```diff
@@ -114,9 +117,26 @@ private[hive] class HiveMetastoreCatalog(sparkSession: SparkSession) extends Log
       alias.map(a => SubqueryAlias(a, qualifiedTable, None)).getOrElse(qualifiedTable)
     } else if (table.tableType == CatalogTableType.VIEW) {
       val viewText = table.viewText.getOrElse(sys.error("Invalid view without text."))
+      val unresolvedPlan = sparkSession.sessionState.sqlParser.parsePlan(viewText).transform {
```
hvanhovell: This will probably break with common table expressions. We could just set the current database, have the analyzer solve it, and set the current database back to its old value.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/interface.scala
```diff
@@ -151,6 +151,7 @@ case class CatalogTable(
     tableType: CatalogTableType,
     storage: CatalogStorageFormat,
     schema: StructType,
+    originalSchema: Option[StructType] = None,
```
hvanhovell: The only difference between originalSchema and schema are the column names, right? If that is the case why not store the `source` column name in the schema field metadata?

hvanhovell (commented):
I took a first pass (I haven't looked at the tests yet). This is a good start. I still think that we still need to address a few issues.

It might be nice if you explain in the PR description that we are doing late binding, and how the binding of views work.

hvanhovell:
also cc @gatorsmile

sql/core/src/main/scala/org/apache/spark/sql/execution/command/views.scala
```diff
@@ -207,31 +205,56 @@ case class CreateViewCommand(
...
+        storage = CatalogStorageFormat.empty,
+        schema = aliasedPlan.schema,
+        originalSchema = Some(originalSchema),
+        properties = properties,
+        viewOriginalText = originalText,
+        viewText = Some(viewSQL),
+        currentDatabase = Some(currentDatabase),
+        comment = comment
+      )
+    } else {
```
nsyca: I guess the code in the `else` happens when the object is a table, not a view, correct? Please add this to the prologue of the function for clarity.

sql/hive/src/test/scala/org/apache/spark/sql/hive/execution/SQLViewSuite.scala
```diff
@@ -448,19 +476,105 @@ class SQLViewSuite extends QueryTest with SQLTestUtils with TestHiveSingleton {
     }
   }
 
+  test("Using view after change the origin view") {
+    withView("v1", "v2") {
+      sql("CREATE VIEW v1 AS SELECT id FROM jt")
+      sql("CREATE VIEW v2 AS SELECT * FROM v1")
+      withTable("jt2", "jt3") {
+        // Don't change the view schema
+        val df2 = (1 until 10).map(i => i + i).toDF("id")
```
nsyca: I suggest to twist this test case a bit.
The first part of the test case has the table jt defined as columns ("j", "i") and then the second part jt2 defined as, what you currently have, columns("i", "j") to test that the code checks the order of the columns correctly.

sql/hive/src/test/scala/org/apache/spark/sql/hive/execution/SQLViewSuite.scala
```diff
@@ -448,19 +476,105 @@ class SQLViewSuite extends QueryTest with SQLTestUtils with TestHiveSingleton {
...
+      sql("CREATE VIEW v2 AS SELECT * FROM v1")
+      // Drop the referenced view
+      sql("DROP VIEW v1")
+      val e = intercept[RuntimeException] {
+        sql("SELECT * FROM v2")
+      }
+      assert(e.getMessage.contains(
+        "Failed to analyze the canonicalized SQL"))
+    }
+  }
```
nsyca: I suggest adding a variation of this test case, dropping the underlining table "jt" and v2 should be invalid. This will test the multiple-level expansion of a view. It is also no harm to check that v1 is also invalid in this new test case.

sql/hive/src/test/scala/org/apache/spark/sql/hive/execution/SQLViewSuite.scala
```diff
@@ -448,19 +476,105 @@ class SQLViewSuite extends QueryTest with SQLTestUtils with TestHiveSingleton {
...
+      df.write.format("json").saveAsTable("tab1")
+      withView("v1", "v2") {
+        sql("CREATE VIEW v1 AS SELECT * FROM tab1")
+        sql("CREATE VIEW v2 AS SELECT * FROM v1")
+        // Don't change the table schema
+        val df2 = (1 until 10).map(i => i * i).toDF("i")
+        df2.write.format("json").mode("overwrite").saveAsTable("tab1")
+        // the view v2 should have the same output with the view v1
+        checkAnswer(sql("SELECT * FROM v2"), sql("SELECT * FROM v1"))
+        // the view v2 should have the same output with the table testTable
```
nsyca: `testTable` or `tab1`?

sql/hive/src/test/scala/org/apache/spark/sql/hive/execution/SQLViewSuite.scala
```diff
@@ -448,19 +476,105 @@ class SQLViewSuite extends QueryTest with SQLTestUtils with TestHiveSingleton {
...
+        sql("DROP TABLE tab1")
+        val e = intercept[RuntimeException] {
+          sql("SELECT * FROM v1")
+        }
+        assert(e.getMessage.contains(
+          "Failed to analyze the canonicalized SQL"))
+      }
+    }
+  }
+
```
nsyca: If you add a new test case to test the multiple-level expansion as I suggested before, this test case is just a duplicate. It may be removed.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/SessionCatalog.scala
```diff
@@ -557,9 +557,12 @@ class SessionCatalog(
    * If the relation is a view, the relation will be wrapped in a [[SubqueryAlias]] which will
    * track the name of the view.
    */
-  def lookupRelation(name: TableIdentifier, alias: Option[String] = None): LogicalPlan = {
+  def lookupRelation(
```
gatorsmile: General suggestion: when you made a change like this, please update the above function descriptions.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
```diff
@@ -509,32 +509,42 @@ class Analyzer(
...
     }
 
-    def apply(plan: LogicalPlan): LogicalPlan = plan resolveOperators {
-      case i @ InsertIntoTable(u: UnresolvedRelation, parts, child, _, _) if child.resolved =>
-        i.copy(table = EliminateSubqueryAliases(lookupTableFromCatalog(u)))
-      case u: UnresolvedRelation =>
-        val table = u.tableIdentifier
-        if (table.database.isDefined && conf.runSQLonFile && !catalog.isTemporaryTable(table) &&
+    def apply(plan: LogicalPlan): LogicalPlan = {
+      var currentDatabase = catalog.getCurrentDatabase
```
gatorsmile: Why we need to do it here? This should be moved to https://github.com/jiangxb1987/spark/blob/434009ef7fcfc552314dca35dc9395bb98f5e9c6/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala#L539

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
```diff
@@ -509,32 +509,42 @@ class Analyzer(
...
-          lookupTableFromCatalog(u)
-        }
+            // If the database part is specified, and we support running SQL directly on files, and
+            // it's not a temporary view, and the table does not exist, then let's just return the
+            // original UnresolvedRelation. It is possible we are matching a query like "select *
+            // from parquet.`/path/to/query`". The plan will get resolved later.
+            // Note that we are testing (!db_exists || !table_exists) because the catalog throws
+            // an exception from tableExists if the database does not exist.
+            u
+          } else {
```
gatorsmile: Please roll back all the unnecessary changes between line 522 and line 538

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
```diff
@@ -509,32 +509,42 @@ class Analyzer(
...
+            val logicalPlan = lookupTableFromCatalog(u, Some(currentDatabase))
+            currentDatabase = logicalPlan.collectFirst {
```
gatorsmile: Please avoid using `var` here.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/SessionCatalog.scala
```diff
@@ -557,9 +557,12 @@ class SessionCatalog(
...
+      name: TableIdentifier,
+      alias: Option[String] = None,
+      databaseHint: Option[String] = None): LogicalPlan = {
```
gatorsmile: Need a document for these parameters.

sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveSessionCatalog.scala
```diff
@@ -55,16 +55,19 @@ private[sql] class HiveSessionCatalog(
...
-    val db = formatDatabaseName(name.database.getOrElse(currentDb))
+    val db = formatDatabaseName(name.database.getOrElse(databaseHint.getOrElse(currentDb)))
     if (db == globalTempViewManager.database) {
       val relationAlias = alias.getOrElse(table)
       globalTempViewManager.get(table).map { viewDef =>
         SubqueryAlias(relationAlias, viewDef, Some(name))
       }.getOrElse(throw new NoSuchTableException(db, table))
     } else if (name.database.isDefined || !tempTables.contains(table)) {
-      val database = name.database.map(formatDatabaseName)
+      val database = Some(db).map(formatDatabaseName)
```
gatorsmile: -> `val database = formatDatabaseName(db)`

Then, you can change the following line to Some(database)

gatorsmile:
@jiangxb1987 We need a low level design about the changes you made here, especially document the issue you hit and the solution you choose.

gatorsmile:
also cc @cloud-fan

rxin:
@jiangxb1987 - please take a look at my comment on the jira ticket: https://issues.apache.org/jira/browse/SPARK-18209

sql/core/src/main/scala/org/apache/spark/sql/execution/command/views.scala
```diff
@@ -207,31 +205,56 @@ case class CreateViewCommand(
...
 
-    // Validate the view SQL - make sure we can parse it and analyze it.
-    // If we cannot analyze the generated query, there is probably a bug in SQL generation.
-    try {
-      sparkSession.sql(viewSQL).queryExecution.assertAnalyzed()
-    } catch {
-      case NonFatal(e) =>
-        throw new RuntimeException(s"Failed to analyze the canonicalized SQL: $viewSQL", e)
-    }
+    if (originalText.isDefined) {
```
jiangxb1987: I should add comments over this code. The `originalText` is non-empty only if the command is generated from SQL.

sql/core/src/main/scala/org/apache/spark/sql/execution/command/views.scala
```diff
@@ -207,31 +205,56 @@ case class CreateViewCommand(
...
+      // Validate the view SQL - make sure we can resolve it with currentDatabase.
+      val originalSchema = try {
+        val unresolvedPlan = sparkSession.sessionState.sqlParser.parsePlan(viewSQL)
+        val resolvedPlan = sparkSession.sessionState.analyzer.execute(unresolvedPlan)
+        sparkSession.sessionState.analyzer.checkAnalysis(resolvedPlan)
+
+        resolvedPlan.schema
+      } catch {
+        case NonFatal(e) =>
+          throw new RuntimeException(s"Failed to analyze the canonicalized SQL: $viewSQL", e)
```
jiangxb1987: Yeah I agree we should throw a AnalysisException here.

sql/core/src/main/scala/org/apache/spark/sql/execution/command/views.scala
jiangxb1987: I should add comments over this code. The originalText is non-empty only if the command is generated from SQL.

sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveMetastoreCatalog.scala
```diff
@@ -114,9 +117,26 @@ private[hive] class HiveMetastoreCatalog(sparkSession: SparkSession) extends Log
...
+      val resolvedPlan = try {
+        val resolvedPlan = sparkSession.sessionState.analyzer.execute(unresolvedPlan)
+        sparkSession.sessionState.analyzer.checkAnalysis(resolvedPlan)
+
+        resolvedPlan
+      } catch {
+        case NonFatal(e) =>
+          throw new RuntimeException(s"Failed to analyze the canonicalized SQL: $viewText", e)
+      }
+      val planWithProjection = table.originalSchema.map(withProjection(resolvedPlan, _))
```
jiangxb1987: I think so.

sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveMetastoreCatalog.scala
```diff
@@ -114,9 +117,26 @@ private[hive] class HiveMetastoreCatalog(sparkSession: SparkSession) extends Log
...
+        .getOrElse(resolvedPlan)
+
       SubqueryAlias(
         alias.getOrElse(table.identifier.table),
-        sparkSession.sessionState.sqlParser.parsePlan(viewText),
+        aliasColumns(planWithProjection, table.schema.fields),
```
jiangxb1987: It will might be a bit more complex, but I think we certainly can do that.

sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveMetastoreCatalog.scala
```diff
@@ -126,6 +146,55 @@ private[hive] class HiveMetastoreCatalog(sparkSession: SparkSession) extends Log
...
+  private def compareStructField(field: StructField, other: StructField): Boolean = {
+    field.name == other.name && field.dataType == other.dataType
+  }
+
+  /**
+   * Aliases the schema of the LogicalPlan to the view attribute names
+   */
+  private def aliasColumns(plan: LogicalPlan, fields: Seq[StructField]): LogicalPlan = {
+    val output = fields.map(field => (field.name, field.getComment))
+    if (plan.output.size != output.size) {
```
jiangxb1987: This should not happen. I just want to ensure we are safe here in case the `withProjection` has been modified.

sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveSessionCatalog.scala
jiangxb1987: +1

sql/hive/src/test/scala/org/apache/spark/sql/hive/execution/SQLViewSuite.scala
jiangxb1987: Good point! I'll add the case.

jiangxb1987:
@hvanhovell @nsyca @gatorsmile @rxin Thank you for your suggestions! I will try to make a better approach ASAP! Thank you！

jiangxb1987:
This is closed because we should achieve the final goal by submitting a sequence of smaller PRs.

PR #16169, opened 2016-12-06
[SPARK-18326][SPARKR][ML] Review SparkR ML wrappers API for 2.1

yanboliang:
Reviewing SparkR ML wrappers API for 2.1 release, mainly two issues:
* Remove ```probabilityCol``` from the argument list of ```spark.logit``` and ```spark.randomForest```. Since it was used when making prediction and should be an argument of ```predict```, and we will work on this at [SPARK-18618](https://issues.apache.org/jira/browse/SPARK-18618) in the next release cycle. 
* Fix ```spark.als``` params to make it consistent with MLlib.

## How was this patch tested?
Existing tests.

### DIFF
--- a/R/pkg/R/mllib.R
+++ b/R/pkg/R/mllib.R
@@ -733,7 +733,6 @@ setMethod("predict", signature(object = "KMeansModel"),
 #'                  excepting that at most one value may be 0. The class with largest value p/t is predicted, where p
 #'                  is the original probability of that class and t is the class's threshold.
 #' @param weightCol The weight column name.
-#' @param probabilityCol column name for predicted class conditional probabilities.
 #' @param ... additional arguments passed to the method.
 #' @return \code{spark.logit} returns a fitted logistic regression model
 #' @rdname spark.logit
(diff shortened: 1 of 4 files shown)


yanboliang:
cc @felixcheung @jkbradley

felixcheung:
How would we set `probabilityCol` on the classifier which is where the method `.setProbabilityCol(probabilityCol)` is (suppose, from the pipeline model?)?
Is it ok to call this after calling `fit` on the pipeline? What happens when that is set multiple times to different values?

felixcheung:
Generally looks good, have a question above.

yanboliang:
@felixcheung This is a good question. It's reasonable to call ```.setProbabilityCol(probabilityCol)``` on a model or a pipeline model, and this should be handled at model transformation step(after fitting on the pipeline or a single estimator). Users can set multiple times to different values to a generated model, rather than re-training model several times. Thanks.

R/pkg/R/mllib.R
```diff
@@ -1464,20 +1463,20 @@ setMethod("predict", signature(object = "GaussianMixtureModel"),
...
+                   rank = 10, regParam = 0.1, maxIter = 10, nonnegative = FALSE,
                    implicitPrefs = FALSE, alpha = 1.0, numUserBlocks = 10, numItemBlocks = 10,
                    checkpointInterval = 10, seed = 0) {
 
             if (!is.numeric(rank) || rank <= 0) {
               stop("rank should be a positive number.")
             }
-            if (!is.numeric(reg) || reg < 0) {
+            if (!is.numeric(regParam) || regParam < 0) {
               stop("reg should be a nonnegative number.")
```
jkbradley: reg -> regParam

jkbradley:
I don't really see the harm in letting users specify probabilityCol beforehand, except that they may not have a good way to map the indices to String labels.  I'm OK with removing it for now though.

yanboliang:
Merged into master and branch-2.1. Thanks for all your reviewing.

PR #16171, opened 2016-12-06
[SPARK-18739][ML][PYSPARK] Classification and regression Models in pyspark support column setters

zhengruifeng:
0, move `JavaPredictionModel` from `util.py` to `wrapper.py`, for `HasFeaturesCol` and `HasPredictionCol` can not be imported in `util.py`.
1,  `JavaPredictionModel` inherit `HasFeaturesCol` and `HasPredictionCol` 
2,  `JavaClassificationModel` inherit `HasRawPredictionCol`
3, create class `JavaProbabilisticClassificationModel` inherit `HasProbabilityCol`
4, `LogisticRegressionModel`, `DecisionTreeClassificationModel`, `RandomForestClassificationModel`, `GBTClassificationModel`  and `NaiveBayesModel` inherit `JavaProbabilisticClassificationModel`
5, `MultilayerPerceptronClassificationModel` inherit `JavaClassificationModel`
6, `OneVsRestModel` inherit `HasFeaturesCol, HasPredictionCol`
7, Same modifications in `regression.py`
## How was this patch tested?
existing tests and added tests


srowen:
Hm, isn't this the third or fourth PR adding these setters to ML classes? As I said on the last one, we do not want to keep opening different issues to address quite logically related concerns. I would not merge this until you've assessed whether there are any other setter related issues you think need attention. In fact you can close this or mark it as WIP

zhengruifeng:
@srowen No, I think this maybe the third of five PR. Distinct from the first two PRs which only deal with the overlook, the PR modify the hierarchy of classification `Model`s in the python side. With regard to clustering and features, I suggest that we first change the hierarchy in the scala side: add some abstract classes like `ClusteringModel`, `ProbabilisticClusteringModel`, `FeatureModel`. And then modify pyspark to keep it in line with it. Otherwise we need to manually add setXXXCol methods one by one, like what we do now. I think the modifications in clustering and features are big enough to derserve a single JIRA.

zhengruifeng:
cc @yanboliang @sethah @jkbradley

python/pyspark/ml/wrapper.py
```diff
@@ -20,12 +20,14 @@
...
     xrange = range
 
+from pyspark import since, keyword_only
 from pyspark import SparkContext
 from pyspark.sql import DataFrame
 from pyspark.ml import Estimator, Transformer, Model
 from pyspark.ml.param import Params
 from pyspark.ml.util import _jvm
 from pyspark.ml.common import inherit_doc, _java2py, _py2java
+from pyspark.ml.param.shared import *
```
HyukjinKwon: Could we avoid an wild card import if it does not require many if more commits should be pushed?

zhengruifeng: @HyukjinKwon Thanks for pointing this out. I will modify this place.

zhengruifeng:
@jkbradley Could you please to make a pass?

zhengruifeng:
@holdenk @MLnick Can you help reveiwing this?

zhengruifeng:
It is out of date, and I will close it

PR #16105, opened 2016-12-01
[SPARK-18617][Tests]Fix flaky test: StreamingContextSuite.SPARK-18560 Receiver data should be deserialized properly

zsxwing:
## What changes were proposed in this pull request?

Avoid to create multiple threads to stop StreamingContext. Otherwise, the latch added in #16091 can be passed too early.

## How was this patch tested?

Jenkins

PR #16174, opened 2016-12-06
[SPARK-18741][STREAMING] Reuse or clean-up SparkContext in streaming tests

hvanhovell:
Tests in Spark Streaming currently create a `SparkContext` for each test, and sometimes do not clean-up afterwards. This is resource intensive and it can lead to unneeded test failures (flakyness) when park.driver.allowMultipleContexts is disabled (this happens when the order of tests changes).

This PR makes most test re-use a `SparkContext`. For tests that have to create a new context (for instance `CheckpointSuite`) we make sure that no active `SparkContext` exists before the test, and that the created `SparkContext` is cleaned up afterwards. I have refactored the `TestSuiteBase` into two classes `TestSuiteBase` and a parent class `ReusableSparkContext`; this to make `SparkContext` management relatively straightforward for most tests.

I have done a simple very unscientific benchmark (n=1), and streaming tests with this patch took 212 seconds and streaming tests without this patch took 252 seconds.

## How was this patch tested?
The patch only covers test code.

### DIFF
--- a/core/src/main/scala/org/apache/spark/SparkContext.scala
+++ b/core/src/main/scala/org/apache/spark/SparkContext.scala
@@ -2350,6 +2350,16 @@ object SparkContext extends Logging {
     }
   }
 
+  private[spark] def getActiveContext(): Option[SparkContext] = {
+    SPARK_CONTEXT_CONSTRUCTOR_LOCK.synchronized {
+      Option(activeContext.get())
+    }
+  }
+
+  private[spark] def stopActiveContext(): Unit = {
+    getActiveContext().foreach(_.stop())
+  }
+
   /**
    * Called at the end of the SparkContext constructor to ensure that no other SparkContext has
    * raced with this constructor and started.
--- a/streaming/src/test/java/org/apache/spark/streaming/LocalJavaStreamingContext.java
+++ b/streaming/src/test/java/org/apache/spark/streaming/LocalJavaStreamingContext.java
@@ -18,6 +18,7 @@
 package org.apache.spark.streaming;
 
 import org.apache.spark.SparkConf;
+import org.apache.spark.SparkContext$;
 import org.apache.spark.streaming.api.java.JavaStreamingContext;
 import org.junit.After;
 import org.junit.Before;
@@ -28,6 +29,7 @@ public abstract class LocalJavaStreamingContext {
 
     @Before
     public void setUp() {
+        SparkContext$.MODULE$.stopActiveContext();
         SparkConf conf = new SparkConf()
             .setMaster("local[2]")
             .setAppName("test")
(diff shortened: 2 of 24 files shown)


hvanhovell:
cc @tdas @zsxwing

streaming/src/test/java/org/apache/spark/streaming/LocalJavaStreamingContext.java
```diff
@@ -28,6 +29,7 @@
 
     @Before
     public void setUp() {
+        SparkContext$.MODULE$.stopActiveContext();
```
vanzin: Can't you just call `SparkContext.stopActiveContext()` from Java? I see static methods in the bytecode for things like `jarOfObject`.

The line is also indented incorrectly.

zsxwing:
> This is resource intensive and it can lead to unneeded test failures (flakyness) when park.driver.allowMultipleContexts is disabled (this happens when the order of tests changes).

Do you mean the recent failures? It was fixed by #16105.

core/src/main/scala/org/apache/spark/SparkContext.scala
```diff
@@ -2350,6 +2350,16 @@ object SparkContext extends Logging {
     }
   }
 
+  private[spark] def getActiveContext(): Option[SparkContext] = {
+    SPARK_CONTEXT_CONSTRUCTOR_LOCK.synchronized {
+      Option(activeContext.get())
+    }
+  }
+
+  private[spark] def stopActiveContext(): Unit = {
```
vanzin: I don't know. I'm not a big fan of the approach you're taking here: calling this method before running tests. That feels like a sledgehammer to fix flaky tests. I think it would be better for test code to be more careful about cleaning after itself. Kinda like most tests in spark-core use `LocalSparkContext` to more or less automatically do that without the need for these methods.

The `ReuseableSparkContext` trait you have is a step in that direction. If you make sure all needed streaming tests are using it, and keep this state within that class, I think it would be a better change.

zsxwing: +1. I don't like stopping SparkContext before running tests, either. It will hide the mistakes in other tests.

streaming/src/test/java/org/apache/spark/streaming/LocalJavaStreamingContext.java
hvanhovell: Intellij and SBT were both complaining. So I did this.

I'll try to rebuild and see what happens.

core/src/main/scala/org/apache/spark/SparkContext.scala
srowen: Shouldn't this be unnecessary with more carefully written tests? that always close the context etc when done?

PR #16175, opened 2016-12-06
[SPARK-17460][SQL]Make sure sizeInBytes in Statistics will not overflow

huaxingao:

1. In SparkStrategies.canBroadcast, I will add the check   plan.statistics.sizeInBytes >= 0
2. In LocalRelations.statistics, when calculate the statistics, I will change the size to BigInt so it won't overflow. 

  
## How was this patch tested?

I will add a test case to make sure the statistics.sizeInBytes won't overflow.

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/SparkStrategies.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/SparkStrategies.scala
@@ -115,7 +115,8 @@ abstract class SparkStrategies extends QueryPlanner[SparkPlan] {
      */
     private def canBroadcast(plan: LogicalPlan): Boolean = {
       plan.statistics.isBroadcastable ||
-        plan.statistics.sizeInBytes <= conf.autoBroadcastJoinThreshold
+        (plan.statistics.sizeInBytes >= 0 &&
+          plan.statistics.sizeInBytes <= conf.autoBroadcastJoinThreshold)
     }
 
     /**
--- a/sql/core/src/test/scala/org/apache/spark/sql/DatasetSuite.scala
+++ b/sql/core/src/test/scala/org/apache/spark/sql/DatasetSuite.scala
@@ -1112,10 +1112,12 @@ class DatasetSuite extends QueryTest with SharedSQLContext {
   }
 
 
-  test ("SPARK-17460: the sizeInBytes in Statistics shouldn't overflow") {
+  test ("SPARK-17460: If the sizeInBytes in Statistics exceeds the limit of an Int, " +
+    "it should convert to a BigInt instead of overflowing to a negative number") {
     val ds = (0 to 10000).map( i =>
       (i, Seq((i, Seq((i, "This is really not that long of a string")))))).toDS()
     val sizeInBytes = ds.logicalPlan.statistics.sizeInBytes
+    // sizeInBytes is 2404280404, before the fix, it will overflow to a negative number
     assert(sizeInBytes > 0)
   }
 }
--- a/sql/core/src/test/scala/org/apache/spark/sql/DatasetSuite.scala
+++ b/sql/core/src/test/scala/org/apache/spark/sql/DatasetSuite.scala
@@ -1111,10 +1111,9 @@ class DatasetSuite extends QueryTest with SharedSQLContext {
     assert(e.getMessage.contains("Cannot create encoder for Option of Product type"))
   }
 
-
   test ("SPARK-17460: the sizeInBytes in Statistics shouldn't overflow to a negative number") {
-    // If the sizeInBytes in Statistics exceeds the limit of an Int,
-    // it should convert to a BigInt instead of overflowing to a negative number
+    // Since the sizeInBytes in Statistics could exceed the limit of an Int, we should use BigInt
+    // instead of Int for avoiding possible overflow.
     val ds = (0 to 10000).map( i =>
       (i, Seq((i, Seq((i, "This is really not that long of a string")))))).toDS()
     val sizeInBytes = ds.logicalPlan.statistics.sizeInBytes
--- a/sql/core/src/test/scala/org/apache/spark/sql/DatasetSuite.scala
+++ b/sql/core/src/test/scala/org/apache/spark/sql/DatasetSuite.scala
@@ -1112,12 +1112,13 @@ class DatasetSuite extends QueryTest with SharedSQLContext {
   }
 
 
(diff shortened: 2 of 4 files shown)


huaxingao:
@gatorsmile Could you please take a look when you have time?  Thanks a lot!!

sql/core/src/test/scala/org/apache/spark/sql/DatasetSuite.scala
```diff
@@ -1110,6 +1110,14 @@ class DatasetSuite extends QueryTest with SharedSQLContext {
     }
     assert(e.getMessage.contains("Cannot create encoder for Option of Product type"))
   }
+
+
+  test ("SPARK-17460: the sizeInBytes in Statistics shouldn't overflow") {
+    val ds = (0 to 10000).map( i =>
+      (i, Seq((i, Seq((i, "This is really not that long of a string")))))).toDS()
+    val sizeInBytes = ds.logicalPlan.statistics.sizeInBytes
```
gatorsmile: Without the change, we got the following error because of overflow:

```
-1890686892 was not greater than 0
ScalaTestFailureLocation: org.apache.spark.sql.DatasetSuite$$anonfun$65 at (DatasetSuite.scala:554)
org.scalatest.exceptions.TestFailedException: -1890686892 was not greater than 0
```

gatorsmile:
cc @cloud-fan

sql/core/src/test/scala/org/apache/spark/sql/DatasetSuite.scala
```diff
@@ -1110,6 +1110,14 @@ class DatasetSuite extends QueryTest with SharedSQLContext {
     }
     assert(e.getMessage.contains("Cannot create encoder for Option of Product type"))
   }
+
+
+  test ("SPARK-17460: the sizeInBytes in Statistics shouldn't overflow") {
```
rxin: I think you'd need to explain what this test is actually doing better.

sql/core/src/test/scala/org/apache/spark/sql/DatasetSuite.scala
```diff
@@ -1110,6 +1110,16 @@ class DatasetSuite extends QueryTest with SharedSQLContext {
     }
     assert(e.getMessage.contains("Cannot create encoder for Option of Product type"))
   }
+
+
+  test ("SPARK-17460: If the sizeInBytes in Statistics exceeds the limit of an Int, " +
```
gatorsmile: You do not need to use such a long test case name. Just put the description inside the test case.

huaxingao: Thanks for the comment.  Changed.

sql/core/src/test/scala/org/apache/spark/sql/DatasetSuite.scala
```diff
@@ -1110,6 +1110,17 @@ class DatasetSuite extends QueryTest with SharedSQLContext {
     }
     assert(e.getMessage.contains("Cannot create encoder for Option of Product type"))
   }
+
+
```
gatorsmile: remove the extra line.

sql/core/src/test/scala/org/apache/spark/sql/DatasetSuite.scala
```diff
@@ -1110,6 +1110,17 @@ class DatasetSuite extends QueryTest with SharedSQLContext {
...
+  test ("SPARK-17460: the sizeInBytes in Statistics shouldn't overflow to a negative number") {
+    // If the sizeInBytes in Statistics exceeds the limit of an Int,
+    // it should convert to a BigInt instead of overflowing to a negative number
```
gatorsmile: How about? 
`Since the sizeInBytes in Statistics could exceed the limit of an Int, we should use BigInt instead of Int for avoiding possible overflow`

gatorsmile:
I checked the code base. We are having multiple potential overflow issues in `sizeInBytes` of `Statistics`. We might need a fix for all the other places.

huaxingao:
@gatorsmile Thanks a lot for reviewing this.  Sorry I just saw  your last comment after I pushed the change.  Will make more changes for other potential overflow issues.

sql/core/src/main/scala/org/apache/spark/sql/execution/SparkStrategies.scala
```diff
@@ -115,7 +115,8 @@ abstract class SparkStrategies extends QueryPlanner[SparkPlan] {
      */
     private def canBroadcast(plan: LogicalPlan): Boolean = {
       plan.statistics.isBroadcastable ||
-        plan.statistics.sizeInBytes <= conf.autoBroadcastJoinThreshold
+        (plan.statistics.sizeInBytes >= 0 &&
```
gatorsmile: We need to update the value of `-1` to the actual default value `Long.MaxValue` https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala#L139 

Update the code in https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala#L767 to
```
def defaultSizeInBytes: Long = getConf(DEFAULT_SIZE_IN_BYTES)
```

gatorsmile: So far, the incorrect statistics should be fine. Even if we add an `assert`, we still can disable it at runtime when it blocking the execution. Thus, either is fine to me.

huaxingao: @cloud-fan @gatorsmile 
Thank you all for your comments.  I will change defaultSizeInBytes based on the above comments. 
Seems all the other places of Statistics.sizeInBytes are safe and will not cause overflow.

gatorsmile:
retest this please

cloud-fan:
retest this please

cloud-fan:
retest this please

viirya:
In `UnaryNode`, although it is less problematic compared with `LocalRelations`, but it still possible to have similar issue:

https://github.com/apache/spark/blob/fa7d9d70825a6816495d239da925d0087f7cb94f/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/LogicalPlan.scala#L316-L317

It is better to fix it together in this.

cloud-fan:

PR #16177, opened 2016-12-06
[SPARK-17760][SQL] AnalysisException with dataframe pivot when groupBy column is not attribute

aray:

Fixes AnalysisException for pivot queries that have group by columns that are expressions and not attributes by substituting the expressions output attribute in the second aggregation and final projection.

## How was this patch tested?

existing and additional unit tests


rxin:
cc @cloud-fan / @hvanhovell

cloud-fan:
good catch! LGTM, cc @hvanhovell to take another look.

hvanhovell:

hvanhovell:
@aray could you open a backport for branch 2.0  if you feel that we should also fix it there?

PR #16178, opened 2016-12-06
[SPARK-18751][Core]Fix deadlock when SparkContext.stop is called in Utils.tryOrStopSparkContext

zsxwing:

When `SparkContext.stop` is called in `Utils.tryOrStopSparkContext` (the following three places), it will cause deadlock because the `stop` method needs to wait for the thread running `stop` to exit.

- ContextCleaner.keepCleaning
- LiveListenerBus.listenerThread.run
- TaskSchedulerImpl.start

This PR adds `SparkContext.stopInNewThread` and uses it to eliminate the potential deadlock. I also removed my changes in #15775 since they are not necessary now.

## How was this patch tested?

Jenkins

### DIFF
--- a/core/src/main/scala/org/apache/spark/SparkContext.scala
+++ b/core/src/main/scala/org/apache/spark/SparkContext.scala
@@ -1760,25 +1760,24 @@ class SparkContext(config: SparkConf) extends Logging {
   def listJars(): Seq[String] = addedJars.keySet.toSeq
 
   /**
-   * Shut down the SparkContext.
+   * When stopping SparkContext inside Spark components, it's easy to cause dead-lock since Spark
+   * may wait for some internal threads to finish. It's better to use this method to stop
+   * SparkContext instead.
    */
-  def stop(): Unit = {
-    if (env.rpcEnv.isInRPCThread) {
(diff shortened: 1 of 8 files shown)


zsxwing:
cc @rxin

core/src/main/scala/org/apache/spark/SparkContext.scala
```diff
@@ -1760,25 +1760,24 @@ class SparkContext(config: SparkConf) extends Logging {
...
 
-        override def run(): Unit = {
-          _stop()
-        }
-      }.start()
-    } else {
-      _stop()
-    }
+      override def run(): Unit = {
+        SparkContext.this.stop()
```
marmbrus: Will this ever throw an exception?  Should we register an `UncaughtExceptionHandler` or try catch with logging?

zsxwing: > Will this ever throw an exception? Should we register an UncaughtExceptionHandler or try catch with logging?

This happens in the driver, so we cannot use `SparkUncaughtExceptionHandler` to catch the error. The error will be sent to the user's UncaughtExceptionHandler if specified or just print to stderr.

vanzin:
LGTM

zsxwing:

PR #16185, opened 2016-12-07
Update Spark documentation to provide information on how to create External Table

c-sahuja:
Although, currently, the saveAsTable does not provide an API to save the table as an external table from a DataFrame, we can achieve this functionality by using options on DataFrameWriter where the key for the map is the String: "path" and the value is another String which is the location of the external table itself. This can be provided before the call to saveAsTable is performed.

## How was this patch tested?
Documentation was reviewed for formatting and content after the push was performed on the branch.
updated documentation


rxin:
Thanks - merging in master.

PR #16186, opened 2016-12-07
[SPARK-18758][SS] StreamingQueryListener events from a StreamingQuery should be sent only to the listeners in the same session as the query

tdas:

Listeners added with `sparkSession.streams.addListener(l)` are added to a SparkSession. So events only from queries in the same session as a listener should be posted to the listener. Currently, all the events gets rerouted through the Spark's main listener bus, that is, 
- StreamingQuery posts event to StreamingQueryListenerBus. Only the queries associated with the same session as the bus posts events to it.
- StreamingQueryListenerBus posts event to Spark's main LiveListenerBus as a SparkEvent.
- StreamingQueryListenerBus also subscribes to LiveListenerBus events thus getting back the posted event in a different thread. 
- The received is posted to the registered listeners.

The problem is that *all StreamingQueryListenerBuses in all sessions* gets the events and posts them to their listeners. This is wrong.

In this PR, I solve it by making StreamingQueryListenerBus track active queries (by their runIds) when a query posts the QueryStarted event to the bus. This allows the rerouted events to be filtered using the tracked queries.

Note that this list needs to be maintained separately
from the `StreamingQueryManager.activeQueries` because a terminated query is cleared from
`StreamingQueryManager.activeQueries` as soon as it is stopped, but the this ListenerBus must
clear a query only after the termination event of that query has been posted lazily, much after the query has been terminated.

Credit goes to @zsxwing for coming up with the initial idea.

## How was this patch tested?
Updated test harness code to use the correct session, and added new unit test.


sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/memory.scala
```diff
@@ -70,11 +70,11 @@ case class MemoryStream[A : Encoder](id: Int, sqlContext: SQLContext)
 
   def schema: StructType = encoder.schema
 
-  def toDS()(implicit sqlContext: SQLContext): Dataset[A] = {
```
tdas: removed this because this is not needed. the sqlContext is in the constructor.
rather this causes confusion in multi-session environment.

tdas:
@marmbrus @zsxwing @brkyvz Please review.

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamingQueryListenerBus.scala
```diff
@@ -63,18 +83,29 @@ class StreamingQueryListenerBus(sparkListenerBus: LiveListenerBus)
...
   }
 
+  /**
+   * Dispatch events to registered StreamingQueryListeners. Only the events associated queries
+   * started in the same SparkSession as this ListenerBus will be dispatched to the listeners.
+   */
   override protected def doPostEvent(
       listener: StreamingQueryListener,
       event: StreamingQueryListener.Event): Unit = {
+    val runIdsToReportTo = activeQueryRunIds.synchronized { activeQueryRunIds.toSet }
```
zsxwing: Why need to clone the set? You can just use `activeQueryRunIds.synchronized { activeQueryRunIds.contains(...) }`. Right?

zsxwing (changes requested):
Looks good overall. Just a minor style issue.

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamingQueryListenerBus.scala
```diff
@@ -35,12 +43,24 @@ class StreamingQueryListenerBus(sparkListenerBus: LiveListenerBus)
...
   /**
-   * Post a StreamingQueryListener event to the Spark listener bus asynchronously. This event will
-   * be dispatched to all StreamingQueryListener in the thread of the Spark listener bus.
+   * RunIds of active queries whose events are supposed to be forwarded by this ListenerBus
+   * to registered `StreamingQueryListeners`. Note that this list needs to be maintained separately
+   * from the `StreamingQueryManager.activeQueries` because a terminated query is cleared from
+   * `StreamingQueryManager.activeQueries` as soon as it is stopped, but the this ListenerBus must
+   * clear a query only after the termination event of that query has been posted.
+   */
+  private val activeQueryRunIds = new mutable.HashSet[UUID]
```
brkyvz: ~~just a qq: why do we use runId's instead of the id's of the streams? We already don't want concurrent runs for streams since the offset log directories will be messed up.~~

Wait, ok, got it. onStart's are called synchronously where as onTerminations are asynchronous. Basically we can get a second stream start report before the firs run completes. Do you think that's worth adding to the docs? You don't have to if you don't need a second pass.

zsxwing: +1

It's worth to document it. This is different from other Spark's listener buses because of the SPARK-18764 `synchronous` event.

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamingQueryListenerBus.scala
tdas: otherwise the code would look like 
```
case queryStarted: QueryStartedEvent =>
        if (activeQueryRunIds.synchronized { activeQueryRunIds.contains(queryStarted.runId) }) {
          listener.onQueryStarted(queryStarted)
        }
      case queryProgress: QueryProgressEvent =>
        if (activeQueryRunIds.synchronized { activeQueryRunIds.contains(queryProgress.progress.runId) }) {
          listener.onQueryProgress(queryProgress)
        }
      case queryTerminated: QueryTerminatedEvent =>
        if (activeQueryRunIds.synchronized { activeQueryRunIds.contains(queryTerminated.runId) }) {
          listener.onQueryTerminated(queryTerminated)
          activeQueryRunIds.synchronized { activeQueryRunIds -= queryTerminated.runId }
        }

```

zsxwing: Looks good? It also reduces the number of lines :)

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamingQueryListenerBus.scala
tdas: Even if this behavior was not different (that is, all async), this component should not be responsible preventing concurrent runs. This component should be simple and not deal with such issues. I have added more docs regarding why runIds instead of ids

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamingQueryListenerBus.scala
tdas: i think this looks uglier. repeated code. longer lines? :)

zsxwing:
LGTM pending tests

PR #14680, opened 2016-08-17
[SPARK-17101][SQL] Provide consistent format identifiers for TextFileFormat and ParquetFileFormat

jaceklaskowski:
## What changes were proposed in this pull request?

Define the format identifier that is used in Optimized Logical Plan in `explain` for text and parquet file formats (following CSV and JSON formats).
### Before
- Text

```
== Physical Plan ==
InMemoryTableScan [value#24]
   +- InMemoryRelation [value#24], true, 10000, StorageLevel(disk, memory, deserialized, 1 replicas)
         +- *FileScan text [value#24] Batched: false, Format: org.apache.spark.sql.execution.datasources.text.TextFileFormat@262e2c8c, InputPaths: file:/Users/jacek/dev/oss/spark/people.csv, PartitionFilters: [], PushedFilters: [], ReadSchema: struct<value:string>
```
- Parquet

```
== Physical Plan ==
*FileScan parquet [id#7L] Batched: true, Format: ParquetFormat, InputPaths: file:/tmp/test, PartitionFilters: [], PushedFilters: [], ReadSchema: struct<id:bigint>
```
### After
- Text

```
== Physical Plan ==
InMemoryTableScan [value#0]
   +- InMemoryRelation [value#0], true, 10000, StorageLevel(disk, memory, deserialized, 1 replicas)
         +- *FileScan text [value#0] Batched: false, Format: Text, InputPaths: file:/Users/jacek/dev/oss/spark/people.csv, PartitionFilters: [], PushedFilters: [], ReadSchema: struct<value:string>
```
- Parquet

```
== Physical Plan ==
*FileScan parquet [id#7L] Batched: true, Format: Parquet, InputPaths: file:/tmp/test, PartitionFilters: [], PushedFilters: [], ReadSchema: struct<id:bigint>
```
## How was this patch tested?

Local build.

PR #16187, opened 2016-12-07
[SPARK-18760][SQL] Consistent format specification for FileFormats

rxin:
This patch fixes the format specification in explain for file sources (Parquet and Text formats are the only two that are different from the rest):

Before:
```
scala> spark.read.text("test.text").explain()
== Physical Plan ==
*FileScan text [value#15] Batched: false, Format: org.apache.spark.sql.execution.datasources.text.TextFileFormat@xyz, Location: InMemoryFileIndex[file:/scratch/rxin/spark/test.text], PartitionFilters: [], PushedFilters: [], ReadSchema: struct<value:string>
```

After:
```
scala> spark.read.text("test.text").explain()
== Physical Plan ==
*FileScan text [value#15] Batched: false, Format: Text, Location: InMemoryFileIndex[file:/scratch/rxin/spark/test.text], PartitionFilters: [], PushedFilters: [], ReadSchema: struct<value:string>
```

Also closes #14680.

## How was this patch tested?
Verified in spark-shell.

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetFileFormat.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetFileFormat.scala
@@ -61,7 +61,7 @@ class ParquetFileFormat
 
   override def shortName(): String = "parquet"
 
-  override def toString: String = "ParquetFormat"
+  override def toString: String = "Parquet"
 
   override def hashCode(): Int = getClass.hashCode()
 
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/text/TextFileFormat.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/text/TextFileFormat.scala
@@ -39,6 +39,8 @@ class TextFileFormat extends TextBasedFileFormat with DataSourceRegister {
 
   override def shortName(): String = "text"
 
+  override def toString: String = "Text"
+
   private def verifySchema(schema: StructType): Unit = {
     if (schema.size != 1) {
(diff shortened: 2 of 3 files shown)


rxin:
cc @cloud-fan

rxin:
Actually if possible please merge this in branch-2.1.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetFileFormat.scala
```diff
@@ -61,7 +61,7 @@ class ParquetFileFormat
 
   override def shortName(): String = "parquet"
 
-  override def toString: String = "ParquetFormat"
+  override def toString: String = "Parquet"
```
gatorsmile: Just did a quick check. The other formats are using upper cases. Do you think we need to make them consistent? `JSON`, `CSV` and `ORC`

gatorsmile: Yeah. Got it. The offical documents of these three formats are using upper cases.

gatorsmile:
LGTM pending test.

gatorsmile:
retest this please

gatorsmile:
retest this please

gatorsmile:
retest this please

cloud-fan:
> We currently rely on FileFormat implementations to override toString in order to get a proper explain output. It'd be better to just depend on shortName for those.

Looks the PR description is outdated? We still depend on `toString` for explain output right?

rxin:
Fixed-  let me merge this in master/branch-2.1.

PR #16189, opened 2016-12-07
[SPARK-18761][CORE] Introduce "task reaper" to oversee task killing in executors

JoshRosen:

Spark's current task cancellation / task killing mechanism is "best effort" because some tasks may not be interruptible or may not respond to their "killed" flags being set. If a significant fraction of a cluster's task slots are occupied by tasks that have been marked as killed but remain running then this can lead to a situation where new jobs and tasks are starved of resources that are being used by these zombie tasks.

This patch aims to address this problem by adding a "task reaper" mechanism to executors. At a high-level, task killing now launches a new thread which attempts to kill the task and then watches the task and periodically checks whether it has been killed. The TaskReaper will periodically re-attempt to call `TaskRunner.kill()` and will log warnings if the task keeps running. I modified TaskRunner to rename its thread at the start of the task, allowing TaskReaper to take a thread dump and filter it in order to log stacktraces from the exact task thread that we are waiting to finish. If the task has not stopped after a configurable timeout then the TaskReaper will throw an exception to trigger executor JVM death, thereby forcibly freeing any resources consumed by the zombie tasks.

This feature is flagged off by default and is controlled by four new configurations under the `spark.task.reaper.*` namespace. See the updated `configuration.md` doc for details.

## How was this patch tested?

Tested via a new test case in `JobCancellationSuite`, plus manual testing.

### DIFF
--- a/core/src/main/scala/org/apache/spark/executor/Executor.scala
+++ b/core/src/main/scala/org/apache/spark/executor/Executor.scala
@@ -86,6 +86,14 @@ private[spark] class Executor(
   private val executorSource = new ExecutorSource(threadPool, executorId)
   // Pool used for threads that supervise task killing / cancellation
   private val taskReaperPool = ThreadUtils.newDaemonCachedThreadPool("Task reaper")
+  // For tasks which are in the process of being killed, this map the most recently created
+  // TaskReaper. All accesses to this map should be synchronized on the map itself (this isn't
+  // a ConcurrentHashMap because we use the synchronization for purposes other than simply guarding
+  // the integrity of the map's internal state). The purpose of this map is to prevent the creation
+  // of a separate TaskReaper for every killTask() of a given task. Instead, this map allows us to
+  // track whether an existing TaskReaper fulfills the role of a TaskReaper that we would otherwise
+  // create. The map key is a task id.
+  private val taskReaperForTask: HashMap[Long, TaskReaper] = HashMap[Long, TaskReaper]()
 
   if (!isLocal) {
     env.metricsSystem.registerSource(executorSource)
@@ -150,9 +158,19 @@ private[spark] class Executor(
   }
 
   def killTask(taskId: Long, interruptThread: Boolean): Unit = {
-    val tr = runningTasks.get(taskId)
-    if (tr != null) {
-      taskReaperPool.execute(new TaskReaper(tr, interruptThread = interruptThread))
+    val taskRunner = runningTasks.get(taskId)
+    if (taskRunner != null) {
+      taskReaperForTask.synchronized {
+        val shouldCreateReaper = taskReaperForTask.get(taskId) match {
+          case None => true
+          case Some(existingReaper) => interruptThread && !existingReaper.interruptThread
+        }
+        if (shouldCreateReaper) {
+          val taskReaper = new TaskReaper(taskRunner, interruptThread = interruptThread)
+          taskReaperForTask(taskId) = taskReaper
+          taskReaperPool.execute(taskReaper)
+        }
+      }
     }
   }
 
@@ -443,7 +461,12 @@ private[spark] class Executor(
    * Supervises the killing / cancellation of a task by sending the interrupted flag, optionally
    * sending a Thread.interrupt(), and monitoring the task until it finishes.
    */
-  private class TaskReaper(taskRunner: TaskRunner, interruptThread: Boolean) extends Runnable {
+  private class TaskReaper(
+      taskRunner: TaskRunner,
+      val interruptThread: Boolean)
+    extends Runnable {
+
+    private[this] val taskId: Long = taskRunner.taskId
 
     private[this] val killPollingFrequencyMs: Long =
       conf.getTimeAsMs("spark.task.killPollingFrequency", "10s")
@@ -456,36 +479,60 @@ private[spark] class Executor(
     override def run(): Unit = {
       val startTimeMs = System.currentTimeMillis()
       def elapsedTimeMs = System.currentTimeMillis() - startTimeMs
-
-      while (!taskRunner.isFinished && (elapsedTimeMs < killTimeoutMs || killTimeoutMs <= 0)) {
-        taskRunner.kill(interruptThread = interruptThread)
-        taskRunner.synchronized {
-          taskRunner.wait(killPollingFrequencyMs)
-        }
-        if (!taskRunner.isFinished) {
-          logWarning(s"Killed task ${taskRunner.taskId} is still running after $elapsedTimeMs ms")
-          if (takeThreadDump) {
-            try {
-              Utils.getThreadDumpForThread(taskRunner.getThreadId).foreach { thread =>
-                if (thread.threadName == taskRunner.threadName) {
-                  logWarning(s"Thread dump from task ${taskRunner.taskId}:\n${thread.stackTrace}")
+      var exceptionThrown: Boolean = true
+      try {
+        while (!taskRunner.isFinished && (elapsedTimeMs < killTimeoutMs || killTimeoutMs <= 0)) {
+          taskRunner.kill(interruptThread = interruptThread)
+          taskRunner.synchronized {
+            taskRunner.wait(killPollingFrequencyMs)
+          }
+          if (!taskRunner.isFinished) {
+            logWarning(s"Killed task $taskId is still running after $elapsedTimeMs ms")
+            if (takeThreadDump) {
+              try {
+                Utils.getThreadDumpForThread(taskRunner.getThreadId).foreach { thread =>
+                  if (thread.threadName == taskRunner.threadName) {
+                    logWarning(s"Thread dump from task $taskId:\n${thread.stackTrace}")
+                  }
                 }
+              } catch {
+                case NonFatal(e) =>
+                  logWarning("Exception thrown while obtaining thread dump: ", e)
               }
-            } catch {
-              case NonFatal(e) =>
-                logWarning("Exception thrown while obtaining thread dump: ", e)
             }
           }
         }
-      }
-      if (!taskRunner.isFinished && killTimeoutMs > 0 && elapsedTimeMs > killTimeoutMs) {
-        if (isLocal) {
-          logError(s"Killed task ${taskRunner.taskId} could not be stopped within " +
-            s"$killPollingFrequencyMs; not killing JVM because we are running in local mode")
-        } else {
-          throw new SparkException(
-            s"Killing executor JVM because killed task ${taskRunner.taskId} could not be" +
-              s" stopped within $killTimeoutMs ms")
+
+        if (!taskRunner.isFinished && killTimeoutMs > 0 && elapsedTimeMs > killTimeoutMs) {
+          if (isLocal) {
+            logError(s"Killed task $taskId could not be stopped within $killPollingFrequencyMs; " +
+              "not killing JVM because we are running in local mode.")
+          } else if (!exceptionThrown) {
+            // Only throw an exception here in case the finally block was entered via normal
+            // execution and not an exception in order to not mask exceptions thrown by TaskReaper
+            // itself
+            throw new SparkException(
+              s"Killing executor JVM because killed task $taskId could not be stopped within " +
+                s"$killTimeoutMs ms.")
+          } else {
+            logError(s"An unexpected exception was thrown in TaskReaper. The task $taskId could " +
+              s"not be stopped within $killTimeoutMs ms. The TaskReaper exception will now " +
+              s"propagate to the uncaught exception handler and trigger JVM exit.")
+          }
+        }
+        exceptionThrown = false
+      } finally {
+        // Clean up entries in the taskReaperForTask map.
+        taskReaperForTask.synchronized {
+          taskReaperForTask.get(taskId).foreach { taskReaperInMap =>
+            if (taskReaperInMap eq this) {
+              taskReaperForTask.remove(taskId)
+            } else {
+              // This must have been a TaskReaper where interruptThread == false where a subsequent
+              // killTask() call for the same task had interruptThread == true and overwrote the
+              // map entry.
+            }
+          }
         }
       }
     }
--- a/core/src/main/scala/org/apache/spark/executor/Executor.scala
+++ b/core/src/main/scala/org/apache/spark/executor/Executor.scala
@@ -259,6 +259,9 @@ private[spark] class Executor(
       // ClosedByInterruptException during execBackend.statusUpdate which causes
       // Executor to crash
       Thread.interrupted()
+      // Notify any waiting TaskReapers. Generally there will only be one reaper per task but there
+      // is a rare corner-case where one task can have two reapers in case cancel(interrupt=False)
+      // is followed by cancel(interrupt=True). Thus we use notifyAll() to avoid a lost wakeup:
       notifyAll()
     }
 
--- a/core/src/main/scala/org/apache/spark/executor/Executor.scala
+++ b/core/src/main/scala/org/apache/spark/executor/Executor.scala
@@ -84,6 +84,8 @@ private[spark] class Executor(
   // Start worker thread pool
   private val threadPool = ThreadUtils.newDaemonCachedThreadPool("Executor task launch worker")
   private val executorSource = new ExecutorSource(threadPool, executorId)
+  // Pool used for threads that supervise task killing / cancellation
+  private val taskReaperPool = ThreadUtils.newDaemonCachedThreadPool("Task reaper")
 
   if (!isLocal) {
     env.metricsSystem.registerSource(executorSource)
@@ -150,7 +152,7 @@ private[spark] class Executor(
   def killTask(taskId: Long, interruptThread: Boolean): Unit = {
     val tr = runningTasks.get(taskId)
     if (tr != null) {
-      tr.kill(interruptThread)
+      taskReaperPool.execute(new TaskReaper(tr, interruptThread = interruptThread))
     }
   }
 
@@ -161,12 +163,7 @@ private[spark] class Executor(
    * @param interruptThread whether to interrupt the task thread
    */
   def killAllTasks(interruptThread: Boolean) : Unit = {
-    // kill all the running tasks
-    for (taskRunner <- runningTasks.values().asScala) {
-      if (taskRunner != null) {
-        taskRunner.kill(interruptThread)
-      }
-    }
+    runningTasks.keys().asScala.foreach(t => killTask(t, interruptThread = interruptThread))
   }
 
   def stop(): Unit = {
@@ -192,12 +189,16 @@ private[spark] class Executor(
       serializedTask: ByteBuffer)
     extends Runnable {
 
+    val threadName = s"Executor task launch worker for task $taskId"
+
     /** Whether this task has been killed. */
     @volatile private var killed = false
 
     /** Whether this task has been finished. */
     @GuardedBy("TaskRunner.this")
     private var finished = false
+
+    def isFinished: Boolean = synchronized { finished }
 
     /** How much the JVM process has spent in GC when the task starts to run. */
     @volatile var startGCTime: Long = _
@@ -229,9 +230,11 @@ private[spark] class Executor(
       // ClosedByInterruptException during execBackend.statusUpdate which causes
       // Executor to crash
       Thread.interrupted()
+      notifyAll()
     }
 
     override def run(): Unit = {
+      Thread.currentThread().setName(threadName)
       val threadMXBean = ManagementFactory.getThreadMXBean
       val taskMemoryManager = new TaskMemoryManager(env.memoryManager, taskId)
       val deserializeStartTime = System.currentTimeMillis()
@@ -427,6 +430,57 @@ private[spark] class Executor(
 
       } finally {
         runningTasks.remove(taskId)
+      }
+    }
+  }
+
+  /**
+   * Supervises the killing / cancellation of a task by sending the interrupted flag, optionally
+   * sending a Thread.interrupt(), and monitoring the task until it finishes.
+   */
+  private class TaskReaper(taskRunner: TaskRunner, interruptThread: Boolean) extends Runnable {
+
+    private[this] val killPollingFrequencyMs: Long =
+      conf.getTimeAsMs("spark.task.killPollingFrequency", "10s")
+
+    private[this] val killTimeoutMs: Long = conf.getTimeAsMs("spark.task.killTimeout", "2m")
+
+    private[this] val takeThreadDump: Boolean =
+      conf.getBoolean("spark.task.threadDumpKilledTasks", true)
+
+    override def run(): Unit = {
+      val startTimeMs = System.currentTimeMillis()
+      def elapsedTimeMs = System.currentTimeMillis() - startTimeMs
+
+      while (!taskRunner.isFinished && elapsedTimeMs < killTimeoutMs) {
+        taskRunner.kill(interruptThread = interruptThread)
+        taskRunner.synchronized {
+          Thread.sleep(killPollingFrequencyMs)
+        }
+        if (!taskRunner.isFinished) {
+          logWarning(s"Killed task ${taskRunner.taskId} is still running after $elapsedTimeMs ms")
+          if (takeThreadDump) {
+            try {
+              val threads = Utils.getThreadDump()
+              threads.find(_.threadName == taskRunner.threadName).foreach { thread =>
+                logWarning(s"Thread dump from task ${taskRunner.taskId}:\n${thread.stackTrace}")
+              }
+            } catch {
+              case NonFatal(e) =>
+                logWarning("Exception thrown while obtaining thread dump: ", e)
+            }
+          }
+        }
+      }
+      if (!taskRunner.isFinished && killTimeoutMs > 0 && elapsedTimeMs > killTimeoutMs) {
+        if (isLocal) {
+          logError(s"Killed task ${taskRunner.taskId} could not be stopped within " +
+            s"$killPollingFrequencyMs; not killing JVM because we are running in local mode")
+        } else {
+          throw new SparkException(
+            s"Killing executor JVM because killed task ${taskRunner.taskId} could not be" +
+              s" stopped within $killTimeoutMs ms")
+        }
       }
     }
   }
--- a/core/src/main/scala/org/apache/spark/executor/Executor.scala
+++ b/core/src/main/scala/org/apache/spark/executor/Executor.scala
@@ -479,13 +479,14 @@ private[spark] class Executor(
     override def run(): Unit = {
       val startTimeMs = System.currentTimeMillis()
       def elapsedTimeMs = System.currentTimeMillis() - startTimeMs
+      def timeoutExceeded(): Boolean = killTimeoutMs > 0 && elapsedTimeMs > killTimeoutMs
       try {
         // Only attempt to kill the task once. If interruptThread = false then a second kill
         // attempt would be a no-op and if interruptThread = true then it may not be safe or
         // effective to interrupt multiple times:
         taskRunner.kill(interruptThread = interruptThread)
         // Monitor the killed task until it exits:
-        while (!taskRunner.isFinished && (elapsedTimeMs < killTimeoutMs || killTimeoutMs <= 0)) {
+        while (!taskRunner.isFinished && !timeoutExceeded()) {
           taskRunner.synchronized {
             taskRunner.wait(killPollingFrequencyMs)
           }
@@ -506,7 +507,7 @@ private[spark] class Executor(
           }
         }
 
-        if (!taskRunner.isFinished && killTimeoutMs > 0 && elapsedTimeMs > killTimeoutMs) {
+        if (!taskRunner.isFinished && timeoutExceeded()) {
           if (isLocal) {
             logError(s"Killed task $taskId could not be stopped within $killPollingFrequencyMs; " +
               "not killing JVM because we are running in local mode.")
--- a/core/src/main/scala/org/apache/spark/executor/Executor.scala
+++ b/core/src/main/scala/org/apache/spark/executor/Executor.scala
@@ -194,6 +194,10 @@ private[spark] class Executor(
     /** Whether this task has been killed. */
     @volatile private var killed = false
 
+    @volatile private var threadId: Long = -1
+
+    def getThreadId: Long = threadId
+
     /** Whether this task has been finished. */
     @GuardedBy("TaskRunner.this")
     private var finished = false
@@ -234,7 +238,8 @@ private[spark] class Executor(
     }
 
     override def run(): Unit = {
-      Thread.currentThread().setName(threadName)
+      threadId = Thread.currentThread.getId
+      Thread.currentThread.setName(threadName)
       val threadMXBean = ManagementFactory.getThreadMXBean
       val taskMemoryManager = new TaskMemoryManager(env.memoryManager, taskId)
       val deserializeStartTime = System.currentTimeMillis()
@@ -461,9 +466,10 @@ private[spark] class Executor(
           logWarning(s"Killed task ${taskRunner.taskId} is still running after $elapsedTimeMs ms")
           if (takeThreadDump) {
             try {
-              val threads = Utils.getThreadDump()
-              threads.find(_.threadName == taskRunner.threadName).foreach { thread =>
-                logWarning(s"Thread dump from task ${taskRunner.taskId}:\n${thread.stackTrace}")
+              Utils.getThreadDumpForThread(taskRunner.getThreadId).foreach { thread =>
+                if (thread.threadName == taskRunner.threadName) {
+                  logWarning(s"Thread dump from task ${taskRunner.taskId}:\n${thread.stackTrace}")
+                }
               }
             } catch {
               case NonFatal(e) =>
--- a/core/src/main/scala/org/apache/spark/executor/Executor.scala
+++ b/core/src/main/scala/org/apache/spark/executor/Executor.scala
@@ -485,7 +485,7 @@ private[spark] class Executor(
     private[this] val killPollingIntervalMs: Long =
       conf.getTimeAsMs("spark.task.reaper.pollingInterval", "10s")
 
-    private[this] val killTimeoutMs: Long = conf.getTimeAsMs("spark.task.reaper.killTimeout", "2m")
+    private[this] val killTimeoutMs: Long = conf.getTimeAsMs("spark.task.reaper.killTimeout", "-1")
 
     private[this] val takeThreadDump: Boolean =
       conf.getBoolean("spark.task.reaper.threadDump", true)
--- a/core/src/main/scala/org/apache/spark/executor/Executor.scala
+++ b/core/src/main/scala/org/apache/spark/executor/Executor.scala
@@ -474,6 +474,25 @@ private[spark] class Executor(
   /**
    * Supervises the killing / cancellation of a task by sending the interrupted flag, optionally
    * sending a Thread.interrupt(), and monitoring the task until it finishes.
+   *
+   * Spark's current task cancellation / task killing mechanism is "best effort" because some tasks
+   * may not be interruptable or may not respond to their "killed" flags being set. If a significant
+   * fraction of a cluster's task slots are occupied by tasks that have been marked as killed but
+   * remain running then this can lead to a situation where new jobs and tasks are starved of
+   * resources that are being used by these zombie tasks.
+   *
+   * The TaskReaper was introduced in SPARK-18761 as a mechanism to monitor and clean up zombie
+   * tasks. For backwards-compatibility / backportability this component is disabled by default
+   * and must be explicitly enabled by setting `spark.task.reaper.enabled=true`.
+   *
+   * A TaskReaper is created for a particular task when that task is killed / cancelled. Typically
+   * a task will have only one TaskReaper, but it's possible for a task to have up to two reapers
+   * in case kill is called twice with different values for the `interrupt` parameter.
+   *
+   * Once created, a TaskReaper will run until its supervised task has finished running. If the
+   * TaskReaper has not been configured to kill the JVM after a timeout (i.e. if
+   * `spark.task.reaper.killTimeout < 0`) then this implies that the TaskReaper may run indefinitely
+   * if the supervised task never exits.
    */
   private class TaskReaper(
       taskRunner: TaskRunner,
--- a/core/src/main/scala/org/apache/spark/executor/Executor.scala
+++ b/core/src/main/scala/org/apache/spark/executor/Executor.scala
@@ -457,7 +457,7 @@ private[spark] class Executor(
       val startTimeMs = System.currentTimeMillis()
       def elapsedTimeMs = System.currentTimeMillis() - startTimeMs
 
-      while (!taskRunner.isFinished && elapsedTimeMs < killTimeoutMs) {
+      while (!taskRunner.isFinished && (elapsedTimeMs < killTimeoutMs || killTimeoutMs <= 0)) {
         taskRunner.kill(interruptThread = interruptThread)
         taskRunner.synchronized {
           wait(killPollingFrequencyMs)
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/apache/spark/executor/Executor.scala
```diff
@@ -432,6 +435,57 @@ private[spark] class Executor(
...
+
+      while (!taskRunner.isFinished && elapsedTimeMs < killTimeoutMs) {
+        taskRunner.kill(interruptThread = interruptThread)
+        taskRunner.synchronized {
+          Thread.sleep(killPollingFrequencyMs)
+        }
+        if (!taskRunner.isFinished) {
+          logWarning(s"Killed task ${taskRunner.taskId} is still running after $elapsedTimeMs ms")
+          if (takeThreadDump) {
+            try {
```
JoshRosen: This `try` is a precautionary measure to avoid potential issues where the thread mx bean can't give us a thread dump.

core/src/main/scala/org/apache/spark/executor/Executor.scala
```diff
@@ -229,9 +230,11 @@ private[spark] class Executor(
       // ClosedByInterruptException during execBackend.statusUpdate which causes
       // Executor to crash
       Thread.interrupted()
+      notifyAll()
     }
 
     override def run(): Unit = {
+      Thread.currentThread().setName(threadName)
```
JoshRosen: Task ids should be unique so therefore this thread name should be unique. Hence, I don't think it's super important to reset the thread's name when returning it to this task thread pool because the thread will just be renamed as soon as it's recycled for a new task and if the task has already exited then it'll be clear from the thread state / context that this is just a completed task's thread that's been returned to the pool.

core/src/main/scala/org/apache/spark/executor/Executor.scala
```diff
@@ -192,13 +189,17 @@ private[spark] class Executor(
       serializedTask: ByteBuffer)
     extends Runnable {
 
+    val threadName = s"Executor task launch worker for task $taskId"
```
JoshRosen: This naming scheme was intentionally chosen to match the pattern that we use for sorting threads in the executor thread dump page. I'll manually verify that this worked as expected there.

core/src/main/scala/org/apache/spark/executor/Executor.scala
```diff
@@ -432,6 +435,57 @@ private[spark] class Executor(
...
+            } catch {
+              case NonFatal(e) =>
+                logWarning("Exception thrown while obtaining thread dump: ", e)
+            }
+          }
+        }
+      }
+      if (!taskRunner.isFinished && killTimeoutMs > 0 && elapsedTimeMs > killTimeoutMs) {
+        if (isLocal) {
+          logError(s"Killed task ${taskRunner.taskId} could not be stopped within " +
```
JoshRosen: Even if we did throw an exception here, it wouldn't exit the JVM in local mode because we don't set an uncaught exception handler in local mode (see code further up in this file).

core/src/test/scala/org/apache/spark/JobCancellationSuite.scala
```diff
@@ -209,6 +209,41 @@ class JobCancellationSuite extends SparkFunSuite with Matchers with BeforeAndAft
...
+      sc.setJobGroup("jobA", "this is a job to be cancelled", interruptOnCancel = true)
+      sc.parallelize(1 to 10000, 2).map { i =>
+        while (true) { }
+      }.count()
+    }
+
+    // Block until both tasks of job A have started and cancel job A.
+    sem.acquire(2)
+    // Small delay to ensure tasks actually start executing the task body
+    Thread.sleep(1000)
```
JoshRosen: This is slightly ugly but it's needed to avoid a race where this regression test can spuriously pass (and thereby fail to test anything) in case we cancel a task right after it has launched on the executor but before the UDF in the task has actually run.

core/src/main/scala/org/apache/spark/executor/Executor.scala
```diff
@@ -161,12 +163,7 @@ private[spark] class Executor(
    * @param interruptThread whether to interrupt the task thread
    */
   def killAllTasks(interruptThread: Boolean) : Unit = {
-    // kill all the running tasks
-    for (taskRunner <- runningTasks.values().asScala) {
-      if (taskRunner != null) {
-        taskRunner.kill(interruptThread)
-      }
-    }
+    runningTasks.keys().asScala.foreach(t => killTask(t, interruptThread = interruptThread))
```
JoshRosen: A careful reviewer will notice that it's possible for `killTask` to be called twice for the same task, either via multiple calls to `killTask` here or via a call to `killTask` followed by a later `killAllTasks` call. I think that this should technically be okay as of the code in this first draft of this patch since having multiple TaskReapers for the same task should be fine, but I can also appreciate how this could cause resource exhaustion issues in the pathological case where killTask is spammed continuously. If we think it's important to avoid multiple reapers in this case then a simple solution would be to add a `synchronized` method on `TaskRunner` which submits a `TaskReaper` on the first kill request and is a no-op on subsequent requests.

core/src/main/scala/org/apache/spark/executor/Executor.scala
```diff
@@ -432,6 +435,57 @@ private[spark] class Executor(
...
+      conf.getBoolean("spark.task.threadDumpKilledTasks", true)
+
+    override def run(): Unit = {
+      val startTimeMs = System.currentTimeMillis()
+      def elapsedTimeMs = System.currentTimeMillis() - startTimeMs
+
+      while (!taskRunner.isFinished && elapsedTimeMs < killTimeoutMs) {
+        taskRunner.kill(interruptThread = interruptThread)
+        taskRunner.synchronized {
+          Thread.sleep(killPollingFrequencyMs)
```
mridulm: Why is the sleep inside synchronized ?
Also, would be good to use wait/notify instead of sleep - so that the TaskReaper's are released proactively as soon as task finishes, so that number of threads in the cached pool does not go very high.

core/src/main/scala/org/apache/spark/executor/Executor.scala
```diff
@@ -432,6 +435,57 @@ private[spark] class Executor(
...
+        }
+        if (!taskRunner.isFinished) {
+          logWarning(s"Killed task ${taskRunner.taskId} is still running after $elapsedTimeMs ms")
+          if (takeThreadDump) {
+            try {
+              val threads = Utils.getThreadDump()
+              threads.find(_.threadName == taskRunner.threadName).foreach { thread =>
+                logWarning(s"Thread dump from task ${taskRunner.taskId}:\n${thread.stackTrace}")
```
mridulm: This is expensive only to get to a single thread dump.
Why not keep track of thread id, dump only for the single thread (instead of all) and validate threadName is same as expected ?

core/src/main/scala/org/apache/spark/executor/Executor.scala
```diff
@@ -432,6 +435,57 @@ private[spark] class Executor(
...
+      conf.getTimeAsMs("spark.task.killPollingFrequency", "10s")
+
+    private[this] val killTimeoutMs: Long = conf.getTimeAsMs("spark.task.killTimeout", "2m")
+
+    private[this] val takeThreadDump: Boolean =
+      conf.getBoolean("spark.task.threadDumpKilledTasks", true)
+
+    override def run(): Unit = {
+      val startTimeMs = System.currentTimeMillis()
+      def elapsedTimeMs = System.currentTimeMillis() - startTimeMs
```
mridulm: Use Clock instead ?

core/src/main/scala/org/apache/spark/executor/Executor.scala
```diff
@@ -432,6 +435,57 @@ private[spark] class Executor(
...
+                logWarning(s"Thread dump from task ${taskRunner.taskId}:\n${thread.stackTrace}")
+              }
+            } catch {
+              case NonFatal(e) =>
+                logWarning("Exception thrown while obtaining thread dump: ", e)
+            }
+          }
+        }
+      }
+      if (!taskRunner.isFinished && killTimeoutMs > 0 && elapsedTimeMs > killTimeoutMs) {
```
mridulm: Why not move the logging into finally block so that even if there are exceptions in the method we have the status ?
(Makes sense to keep the throw here ofcourse).

core/src/main/scala/org/apache/spark/executor/Executor.scala
JoshRosen: Yeah, I meant to use `wait()` (was prototyping this quickly late at night, hence this oversight). Will change now.

core/src/main/scala/org/apache/spark/executor/Executor.scala
markhamstra: Or access TaskReapers via something like a guava LoadingCache.

JoshRosen: One corner-case that I just thought of: what should happen if the first call to `killTask` sets `interruptThread = false` and the second call sets it to true? If we used a loading cache naively then we would miss the second call. Instead, it might make sense to have the loading cache key be a tuple of `(tid, interrupt)`.

markhamstra: And the more extreme case is if `interruptThread` gets repeatedly reset, sometimes `true` and sometimes `false`.  Probably the only tractable logic will be that if `interruptThread` ever gets set to `true`, then the Task can be interrupted regardless of whether or not `interruptThread` is subsequently reset to `false`.

core/src/main/scala/org/apache/spark/executor/Executor.scala
```diff
@@ -432,6 +435,57 @@ private[spark] class Executor(
...
   /**
+   * Supervises the killing / cancellation of a task by sending the interrupted flag, optionally
+   * sending a Thread.interrupt(), and monitoring the task until it finishes.
+   */
+  private class TaskReaper(taskRunner: TaskRunner, interruptThread: Boolean) extends Runnable {
+
+    private[this] val killPollingFrequencyMs: Long =
+      conf.getTimeAsMs("spark.task.killPollingFrequency", "10s")
+
+    private[this] val killTimeoutMs: Long = conf.getTimeAsMs("spark.task.killTimeout", "2m")
```
JoshRosen: Added a test.

core/src/main/scala/org/apache/spark/executor/Executor.scala
mridulm: Exactly - for testing.
If not unit tests, atleast functional tests should be done.

core/src/main/scala/org/apache/spark/executor/Executor.scala
mridulm: I meant have a try/finally for the entire run method - and log if unable to kill.

JoshRosen: I thought about this and it seems like there are only two possibilities here:

1. We're running in local mode, in which case we don't actually want to throw an exception to kill the JVM and even if we did throw then it would keep on running because there's not an uncaught exception handler here.
2. We're running in a separate JVM, in which case any exception thrown in this thread and not caught will cause the JVM to exit. The only place in the body of this code that might actually throw unexpected exceptions is the taskThreadDump, which is already in a `try-catch` block to prevent exceptions from bubbling up.

Thus the only purpose of a finally block would be to detect whether it was reached via an exception patch and to log a warning to state that task kill progress will no longer be monitored. Basically, I'm not sure what the finally block is buying us in terms of actionable / useful logs and it's only going to add complexity because then we need to be careful to not throw from the finally block in case it was entered via an exception, etc.

JoshRosen: That said, however, it may be useful to have a finally to perform cleanup in a map that I'm about to add / to guarantee that entries cannot possibly be leaked in local mode.

core/src/main/scala/org/apache/spark/executor/Executor.scala
```diff
@@ -432,6 +458,78 @@ private[spark] class Executor(
...
+
+    private[this] val takeThreadDump: Boolean =
+      conf.getBoolean("spark.task.threadDumpKilledTasks", true)
+
+    override def run(): Unit = {
+      val startTimeMs = System.currentTimeMillis()
+      def elapsedTimeMs = System.currentTimeMillis() - startTimeMs
+      try {
+        while (!taskRunner.isFinished && (elapsedTimeMs < killTimeoutMs || killTimeoutMs <= 0)) {
+          taskRunner.kill(interruptThread = interruptThread)
```
JoshRosen: In the case where `interruptThread = false` `taskRunner.kill()` will be idempotent and subsequent calls won't have any effect. In the case where we _do_ interrupt, however, the introduction of this polling loop means that we'll interrupt the same task multiple times. Note that this could, in principle, have happened before, but in practice I think it never would.

Are there cases where back-to-back interrupts could cause user code to break in really bad ways? I'm wondering whether you might have a case where you, say, are interrupted when issuing a SQL query and then use a finally block to do some kind of rollback and then have that rollback / cleanup step itself itself be interrupted a little while later. If this is a scenario that's concerning, then maybe the subsequent polls should only periodically thread dump and not interrupt. In that case, however, we'd need to make sure that back-to-back `killTask(interrupt=true)` calls both interrupt once on the first time, so the logic of avoiding the creation of a second TaskReaper would need to change a bit so that we issue a one-time interrupt in the case where we decide that the current TaskReaper subsumes the interrupt=true one that we'd otherwise create.

core/src/main/scala/org/apache/spark/executor/Executor.scala
JoshRosen: By the way, I looked at the current code and I'm actually not sure that the back-to-back `killTask` for the same task can actually happen very easily in the current architecture.

AFICT, the `interruptThread` flag comes from the job group properties at the time of job submission or is hardcoded to `true` in the case where we're killing redundant task attempts in `handleSuccessfulTask`. So, unless users are using speculation then I don't think this case can arise, and if they are using it then it should only crop up in a weird context where the speculative task refuses to die and then we cancel the job.

So, in a nutshell, I think the case that we're guarding against with this logic isn't actually that common in practice, but it's a good idea to be resilient in case anyone ever refactors this code.

core/src/main/scala/org/apache/spark/executor/Executor.scala
```diff
@@ -432,6 +458,78 @@ private[spark] class Executor(
...
+                }
+              } catch {
+                case NonFatal(e) =>
+                  logWarning("Exception thrown while obtaining thread dump: ", e)
+              }
+            }
+          }
+        }
+
+        if (!taskRunner.isFinished && killTimeoutMs > 0 && elapsedTimeMs > killTimeoutMs) {
```
lins05: I see we also have `(elapsedTimeMs < killTimeoutMs || killTimeoutMs <= 0)` above, what about extracting `killTimeoutMs > 0 && elapsedTimeMs > killTimeoutMs` to a method, e.g `def timeoutExceeded(): Boolean`?

lins05 (commented):
When the executor is killed by the reaper, the driver side would see these messags;

``` 
16/12/11 05:48:44 ERROR TaskSchedulerImpl: Lost executor 1 on 172.16.90.3: Remote
RPC client disassociated. Likely due to containers exceeding thresholds, or
network issues. Check driver logs for WARN messages.

16/12/11 05:48:44 WARN TaskSetManager: Lost task 1.0 in stage 0.0 (TID 1,
172.16.90.3, executor 1): ExecutorLostFailure (executor 1 exited caused by one of
the running tasks) Reason: Remote RPC client disassociated. Likely due to
containers exceeding thresholds, or network issues. Check driver logs for WARN
messages.

16/12/11 05:48:44 ERROR TaskSchedulerImpl: Lost executor 0 on 172.16.90.3: Remote
RPC client disassociated. Likely due to containers exceeding thresholds, or
network issues. Check driver logs for WARN messages.

16/12/11 05:48:44 WARN TaskSetManager: Lost task 0.0 in stage 0.0 (TID 0,
172.16.90.3, executor 0): ExecutorLostFailure (executor 0 exited caused by one of
the running tasks) Reason: Remote RPC client disassociated. Likely due to
containers exceeding thresholds, or network issues. Check driver logs for WARN
messages.
```

The above messages are not helpful for showing the real reason of the executor loss. Can we improve this?

core/src/main/scala/org/apache/spark/executor/Executor.scala
```diff
@@ -432,6 +458,78 @@ private[spark] class Executor(
...
+   */
+  private class TaskReaper(
+      taskRunner: TaskRunner,
+      val interruptThread: Boolean)
+    extends Runnable {
+
+    private[this] val taskId: Long = taskRunner.taskId
+
+    private[this] val killPollingFrequencyMs: Long =
+      conf.getTimeAsMs("spark.task.killPollingFrequency", "10s")
```
lins05: Maybe the name `killPollingInterval` would be slightly better? Also I think the poll interval should not be larger than `spark.task.killTimeout`.

core/src/main/scala/org/apache/spark/executor/Executor.scala
lins05: > In the case where we do interrupt, however, the introduction of this polling loop means that we'll interrupt the same task multiple times

My 2c: if the application code doesn't respond to the first interrupt immediately, chances are very low that it would respond to the following interrupts (it may got stuck in some dead loop, or some blocking JNI call), so sending multiple interrupt may not be necessary.

JoshRosen: That's a good point. I'll update this tomorrow to only interrupt once.

JoshRosen:
@lins05, I'll see if there's a way to get a nicer executor exit status to be reported back to the driver.

core/src/main/scala/org/apache/spark/executor/Executor.scala
JoshRosen: +1 on the naming suggestion; I'll do this tomorrow.

core/src/main/scala/org/apache/spark/executor/Executor.scala
markhamstra: Yes, it is not at all a likely case.  What is the most likely case does concern me, though.

The default value for `spark.job.interruptOnCancel` is `false`; and from the little bit of discussion on SPARK-17064, it doesn't seem that it will be very easy to change that with confidence.  That means that `interruptThread` will also be `false` except for the hard-coded `true` in `handleSuccessfulTask` (which is itself suspect).  The way that I am understanding the default behavior of this PR, then, is that if the reaper functionality is enabled, any kill request of a task (other than from within `handleSuccessfulTask`) that doesn't manage to complete within `killTimeoutMs` of that kill request will result in the Executor's JVM being killed -- i.e. the most likely default behavior looks to boil down to `if(killTask) killJvmAfterKillTimeoutMs`.  That is potentially expensive in terms of lost state from that Executor.  If the external shuffle service is being used, then we shouldn't need to lose the state of the shuffle files; and if one of the external block storage options is being used, then we shouldn't need to lose state on cached RDDs/tables, broadcast variables and/or accumulators.  But none of that externalization and persistence of state across Executor restarts is currently the default -- and that concerns me some.

JoshRosen: Given the current defaults, it's probably best to flag this feature off by default and then I can turn it on in my production environment. For a bit of context, my motivation for this patch is to add a last-ditch safety mechanism to prevent zombie'd executors from causing resource starvation issues on multi-tenant production clusters. This watchdog timer mechanism would have prevented an outage due to uncancellable tasks.

zsxwing:
LGTM

JoshRosen:
Jenkins retest this please

zsxwing:
retest this please

zsxwing:
@JoshRosen Looks like this test is flaky:

```
[info] - task reaper kills JVM if killed tasks keep running for too long *** FAILED *** (23 seconds, 412 milliseconds)
[info]   java.util.concurrent.TimeoutException: Futures timed out after [15 seconds]
[info]   at scala.concurrent.impl.Promise$DefaultPromise.ready(Promise.scala:219)
[info]   at scala.concurrent.impl.Promise$DefaultPromise.ready(Promise.scala:153)
[info]   at org.apache.spark.SimpleFutureAction.result(FutureAction.scala:131)
[info]   at org.apache.spark.util.ThreadUtils$.awaitResult(ThreadUtils.scala:201)
[info]   at org.apache.spark.JobCancellationSuite$$anonfun$15.apply$mcV$sp(JobCancellationSuite.scala:246)
[info]   at org.apache.spark.JobCancellationSuite$$anonfun$15.apply(JobCancellationSuite.scala:212)
[info]   at org.apache.spark.JobCancellationSuite$$anonfun$15.apply(JobCancellationSuite.scala:212)
[info]   at org.scalatest.Transformer$$anonfun$apply$1.apply$mcV$sp(Transformer.scala:22)
[info]   at org.scalatest.OutcomeOf$class.outcomeOf(OutcomeOf.scala:85)
[info]   at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
[info]   at org.scalatest.Transformer.apply(Transformer.scala:22)
[info]   at org.scalatest.Transformer.apply(Transformer.scala:20)
[info]   at org.scalatest.FunSuiteLike$$anon$1.apply(FunSuiteLike.scala:166)
[info]   at org.apache.spark.SparkFunSuite.withFixture(SparkFunSuite.scala:68)
[info]   at org.scalatest.FunSuiteLike$class.invokeWithFixture$1(FunSuiteLike.scala:163)
[info]   at org.scalatest.FunSuiteLike$$anonfun$runTest$1.apply(FunSuiteLike.scala:175)
[info]   at org.scalatest.FunSuiteLike$$anonfun$runTest$1.apply(FunSuiteLike.scala:175)
[info]   at org.scalatest.SuperEngine.runTestImpl(Engine.scala:306)
[info]   at org.scalatest.FunSuiteLike$class.runTest(FunSuiteLike.scala:175)
[info]   at org.apache.spark.JobCancellationSuite.org$scalatest$BeforeAndAfter$$super$runTest(JobCancellationSuite.scala:37)
[info]   at org.scalatest.BeforeAndAfter$class.runTest(BeforeAndAfter.scala:200)
[info]   at org.apache.spark.JobCancellationSuite.org$scalatest$BeforeAndAfterEach$$super$runTest(JobCancellationSuite.scala:37)
[info]   at org.scalatest.BeforeAndAfterEach$class.runTest(BeforeAndAfterEach.scala:255)
[info]   at org.apache.spark.JobCancellationSuite.runTest(JobCancellationSuite.scala:37)
[info]   at org.scalatest.FunSuiteLike$$anonfun$runTests$1.apply(FunSuiteLike.scala:208)
[info]   at org.scalatest.FunSuiteLike$$anonfun$runTests$1.apply(FunSuiteLike.scala:208)
[info]   at org.scalatest.SuperEngine$$anonfun$traverseSubNodes$1$1.apply(Engine.scala:413)
[info]   at org.scalatest.SuperEngine$$anonfun$traverseSubNodes$1$1.apply(Engine.scala:401)
[info]   at scala.collection.immutable.List.foreach(List.scala:381)
[info]   at org.scalatest.SuperEngine.traverseSubNodes$1(Engine.scala:401)
[info]   at org.scalatest.SuperEngine.org$scalatest$SuperEngine$$runTestsInBranch(Engine.scala:396)
[info]   at org.scalatest.SuperEngine.runTestsImpl(Engine.scala:483)
[info]   at org.scalatest.FunSuiteLike$class.runTests(FunSuiteLike.scala:208)
[info]   at org.scalatest.FunSuite.runTests(FunSuite.scala:1555)
[info]   at org.scalatest.Suite$class.run(Suite.scala:1424)
[info]   at org.scalatest.FunSuite.org$scalatest$FunSuiteLike$$super$run(FunSuite.scala:1555)
[info]   at org.scalatest.FunSuiteLike$$anonfun$run$1.apply(FunSuiteLike.scala:212)
[info]   at org.scalatest.FunSuiteLike$$anonfun$run$1.apply(FunSuiteLike.scala:212)
[info]   at org.scalatest.SuperEngine.runImpl(Engine.scala:545)
[info]   at org.scalatest.FunSuiteLike$class.run(FunSuiteLike.scala:212)
[info]   at org.apache.spark.SparkFunSuite.org$scalatest$BeforeAndAfterAll$$super$run(SparkFunSuite.scala:31)
[info]   at org.scalatest.BeforeAndAfterAll$class.liftedTree1$1(BeforeAndAfterAll.scala:257)
[info]   at org.scalatest.BeforeAndAfterAll$class.run(BeforeAndAfterAll.scala:256)
[info]   at org.apache.spark.JobCancellationSuite.org$scalatest$BeforeAndAfter$$super$run(JobCancellationSuite.scala:37)
[info]   at org.scalatest.BeforeAndAfter$class.run(BeforeAndAfter.scala:241)
[info]   at org.apache.spark.JobCancellationSuite.run(JobCancellationSuite.scala:37)
[info]   at org.scalatest.tools.Framework.org$scalatest$tools$Framework$$runSuite(Framework.scala:357)
[info]   at org.scalatest.tools.Framework$ScalaTestTask.execute(Framework.scala:502)
[info]   at sbt.ForkMain$Run$2.call(ForkMain.java:296)
[info]   at sbt.ForkMain$Run$2.call(ForkMain.java:286)
[info]   at java.util.concurrent.FutureTask.run(FutureTask.java:266)
[info]   at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
[info]   at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
[info]   at java.lang.Thread.run(Thread.java:745)
```
I'm not sure if new executors will be ready in 15 seconds.

yhuai:
test this please

yhuai:
let's trigger more tests and see if the test is flaky.

yhuai:
I am reviewing it now

JoshRosen:
I should just bump up the timeout here from 15s, since in practice it'll frequently be faster.

core/src/main/scala/org/apache/spark/executor/Executor.scala
```diff
@@ -229,9 +259,12 @@ private[spark] class Executor(
       // ClosedByInterruptException during execBackend.statusUpdate which causes
       // Executor to crash
       Thread.interrupted()
+      notifyAll()
```
yhuai: How about we also comment the purpose of this `notifyAll`?

core/src/main/scala/org/apache/spark/executor/Executor.scala
```diff
@@ -432,6 +465,93 @@ private[spark] class Executor(
...
+        var finished: Boolean = false
+        while (!finished && !timeoutExceeded()) {
+          taskRunner.synchronized {
+            // We need to synchronize on the TaskRunner while checking whether the task has
+            // finished in order to avoid a race where the task is marked as finished right after
+            // we check and before we call wait().
+            if (taskRunner.isFinished) {
+              finished = true
+            } else {
+              taskRunner.wait(killPollingIntervalMs)
```
yhuai: oh, the notifyAll is used for this, right?

JoshRosen: Yes.

core/src/main/scala/org/apache/spark/executor/Executor.scala
```diff
@@ -432,6 +465,93 @@ private[spark] class Executor(
...
+        }
+
+        if (!taskRunner.isFinished && timeoutExceeded()) {
+          if (isLocal) {
+            logError(s"Killed task $taskId could not be stopped within $killTimeoutMs ms; " +
+              "not killing JVM because we are running in local mode.")
+          } else {
+            throw new SparkException(
+              s"Killing executor JVM because killed task $taskId could not be stopped within " +
+                s"$killTimeoutMs ms.")
```
JoshRosen: Yep, the uncaught exception handler will catch this and force the JVM to exit. This is covered by tests added in this patch.

core/src/main/scala/org/apache/spark/executor/Executor.scala
```diff
@@ -432,6 +465,93 @@ private[spark] class Executor(
   }
 
   /**
+   * Supervises the killing / cancellation of a task by sending the interrupted flag, optionally
+   * sending a Thread.interrupt(), and monitoring the task until it finishes.
+   */
+  private class TaskReaper(
```
yhuai: It will be good to explain more about how this class works. It is great that we have comments at specific parts, but a section explains the workflow will be very helpful to future readers.

JoshRosen:
@lins05, I looked into this and I don't know that there's an easy way to get nicer executor death explanations reported back to the driver. This is somewhat unfortunate but I'm hoping that it won't be a huge issue in practice because this feature is flagged off by default and this path should only be hit in the case of bad errors that would have generally otherwise caused a cluster to become bricked / unusable. That said, I'd love to see if we can fix the error message reporting in a followup patch in case there's an easy way to do it that I'm overlooking.

core/src/main/scala/org/apache/spark/executor/Executor.scala
```diff
@@ -148,9 +161,23 @@ private[spark] class Executor(
...
+      if (taskReaperEnabled) {
+        taskReaperForTask.synchronized {
+          val shouldCreateReaper = taskReaperForTask.get(taskId) match {
+            case None => true
+            case Some(existingReaper) => interruptThread && !existingReaper.interruptThread
+          }
+          if (shouldCreateReaper) {
+            val taskReaper = new TaskReaper(taskRunner, interruptThread = interruptThread)
+            taskReaperForTask(taskId) = taskReaper
+            taskReaperPool.execute(taskReaper)
```
mridulm: Nit: Would be great if the execute is outside the synchronized block. Should not be an issue here though, but just paranoia ...

core/src/main/scala/org/apache/spark/executor/Executor.scala
mridulm: Another note: Compared to earlier, where we were simply running down the values (TaskRunner), here we run down the keys (taskId) and do point lookups for each taskId - it does increase the complexity and to be kept in mind (it is iteration/lookup on a ConcurrentHashMap, not regular HashMap).
Having said that, this should be uncommon enough not to be an issue.

core/src/main/scala/org/apache/spark/executor/Executor.scala
mridulm: Do we have multiple threads waiting on this monitor ? (nofify vs notifyAll)

core/src/main/scala/org/apache/spark/executor/Executor.scala
```diff
@@ -432,6 +465,93 @@ private[spark] class Executor(
...
+      val startTimeMs = System.currentTimeMillis()
+      def elapsedTimeMs = System.currentTimeMillis() - startTimeMs
+      def timeoutExceeded(): Boolean = killTimeoutMs > 0 && elapsedTimeMs > killTimeoutMs
+      try {
+        // Only attempt to kill the task once. If interruptThread = false then a second kill
+        // attempt would be a no-op and if interruptThread = true then it may not be safe or
+        // effective to interrupt multiple times:
+        taskRunner.kill(interruptThread = interruptThread)
+        // Monitor the killed task until it exits:
+        var finished: Boolean = false
```
mridulm: Can we move this variable ?
finished and taskRunner.isFinished seem to be used interchangeably in the code.

core/src/main/scala/org/apache/spark/executor/Executor.scala
```diff
@@ -432,6 +465,93 @@ private[spark] class Executor(
...
+              taskRunner.wait(killPollingIntervalMs)
+            }
+          }
+          if (taskRunner.isFinished) {
+            finished = true
+          } else {
+            logWarning(s"Killed task $taskId is still running after $elapsedTimeMs ms")
+            if (takeThreadDump) {
+              try {
+                Utils.getThreadDumpForThread(taskRunner.getThreadId).foreach { thread =>
```
mridulm: Neat !

core/src/test/scala/org/apache/spark/JobCancellationSuite.scala
```diff
@@ -209,6 +209,83 @@ class JobCancellationSuite extends SparkFunSuite with Matchers with BeforeAndAft
...
+
+    sc.clearJobGroup()
+    val jobB = sc.parallelize(1 to 100, 2).countAsync()
+    sc.cancelJobGroup("jobA")
+    val e = intercept[SparkException] { ThreadUtils.awaitResult(jobA, 15.seconds) }.getCause
+    assert(e.getMessage contains "cancel")
+
+    // Once A is cancelled, job B should finish fairly quickly.
+    assert(ThreadUtils.awaitResult(jobB, 15.seconds) === 100)
+  }
```
mridulm: What is the average and worst case build time hit for this test in your expts ?

core/src/test/scala/org/apache/spark/JobCancellationSuite.scala
mridulm: coutdownlatch for this ?

core/src/main/scala/org/apache/spark/executor/Executor.scala
```diff
@@ -432,6 +465,93 @@ private[spark] class Executor(
...
+      def elapsedTimeMs = System.currentTimeMillis() - startTimeMs
+      def timeoutExceeded(): Boolean = killTimeoutMs > 0 && elapsedTimeMs > killTimeoutMs
+      try {
+        // Only attempt to kill the task once. If interruptThread = false then a second kill
+        // attempt would be a no-op and if interruptThread = true then it may not be safe or
+        // effective to interrupt multiple times:
+        taskRunner.kill(interruptThread = interruptThread)
+        // Monitor the killed task until it exits:
+        var finished: Boolean = false
+        while (!finished && !timeoutExceeded()) {
```
mridulm: killTimeoutMs <= 0 will mean this loop will never finish : do we want to defend against that config ?

core/src/main/scala/org/apache/spark/executor/Executor.scala
```diff
@@ -148,9 +161,23 @@ private[spark] class Executor(
...
+            case Some(existingReaper) => interruptThread && !existingReaper.interruptThread
+          }
+          if (shouldCreateReaper) {
+            val taskReaper = new TaskReaper(taskRunner, interruptThread = interruptThread)
+            taskReaperForTask(taskId) = taskReaper
+            taskReaperPool.execute(taskReaper)
+          }
+        }
+      } else {
+        taskRunner.kill(interruptThread = interruptThread)
```
mridulm: Do we want to introduce dumping stack for this path ? I really like the idea of doing that !

mridulm (commented):
Some nits, thanks for the fix @JoshRosen !

core/src/test/scala/org/apache/spark/JobCancellationSuite.scala
zsxwing: @mridulm that doesn't work since driver and executors are not in the same process.

core/src/main/scala/org/apache/spark/executor/Executor.scala
JoshRosen: Basically, my concern here was that if I did optimize to just use `notify()` then we'd risk bugs if someone else came along later and added more `wait()`ing threads.

And actually I think we might be able to wind up with two waiting threads in the case where you first cancel with interrupt=False and then cancel again with interrupt=True... an example of why it's hard to reason about whether only `notify()` is correct.

core/src/main/scala/org/apache/spark/executor/Executor.scala
JoshRosen: We could but I'd rather not: my goal here is to not affect the `task.reaper.enabled=false` case at all, which is going to be necessary in order for me to backport this back to 1.6.x / 2.0.x.

core/src/main/scala/org/apache/spark/executor/Executor.scala
JoshRosen: This is intentional. The goal is that we'll keep polling and logging until the task exits.

core/src/main/scala/org/apache/spark/executor/Executor.scala
JoshRosen: I think that my original concern was about holding `taskRunner.synchronized` while taking the thread dump. We need to synchronize from the time that we check `taskRunner.isFinished` to when we call `taskRunner.wait()` in order to avoid a race where the task finishes right before we'd begin waiting and if we don't want to synchronize the entire loop then we need to redundantly check again in the `while` condition. My goal here was to avoid doing the `isFinished` call twice.

I suppose that lock contention shouldn't really matter here, though, since this only comes into play once a task has already been marked for killing / cancellation. Let me just lift the synchronized out of the body of the loop and I think I'll be able to simplify this slightly.

core/src/test/scala/org/apache/spark/JobCancellationSuite.scala
JoshRosen: Yeah, in practice this particular usage isn't flaky.

core/src/test/scala/org/apache/spark/JobCancellationSuite.scala
JoshRosen: The test itself usually runs in less than 15 seconds. The worst case could be longer, but it's not significantly worse than similar tests in this suite. This wallclock-based timeout isn't ideal, but it's a lot of work to refactor away from it given the difficulties of dependency injection in this code and we can't do those refactorings in a backportable patch.

core/src/test/scala/org/apache/spark/JobCancellationSuite.scala
mridulm: Ah, missed the obvious - thanks for clarifying.

core/src/main/scala/org/apache/spark/executor/Executor.scala
mridulm: I was not looking at it from performance point of view, but from semantics.
notify implies we expect only single thread to be waiting, notifyAll implies we can expect multiple to be waiting.

If we are unsure if multiple threads could be waiting, then it is fine.

core/src/main/scala/org/apache/spark/executor/Executor.scala
mridulm: sounds good

core/src/main/scala/org/apache/spark/executor/Executor.scala
mridulm: Interesting, is that documented someplace ? I might have missed it ...

core/src/main/scala/org/apache/spark/executor/Executor.scala
JoshRosen: I realized that one argument for _not_ changing this code is the concern that the thread-dump could hang / lock in some weird way. The current code eliminates this possibility by dropping the synchronized before taking the dump.

core/src/main/scala/org/apache/spark/executor/Executor.scala
JoshRosen: It's mentioned on line 495:

> `// Monitor the killed task until it exits.`

core/src/test/scala/org/apache/spark/JobCancellationSuite.scala
```diff
@@ -209,6 +209,83 @@ class JobCancellationSuite extends SparkFunSuite with Matchers with BeforeAndAft
...
+    // Once A is cancelled, job B should finish fairly quickly.
+    assert(ThreadUtils.awaitResult(jobB, 60.seconds) === 100)
+  }
+
+  test("task reaper will not kill JVM if spark.task.killTimeout == -1") {
+    val conf = new SparkConf()
+      .set("spark.task.reaper.enabled", "true")
+      .set("spark.task.reaper.killTimeout", "-1")
+      .set("spark.task.reaper.PollingInterval", "1s")
+      .set("spark.deploy.maxExecutorRetries", "1")
```
yhuai: We set it to 1 to make sure that we will not kill JVM, right (if we kill JVM, we will remove the application because spark.deploy.maxExecutorRetries is 1.)?

yhuai:
LGTM!

yhuai:
Thank you for those comments. I am merging this to master.

mridulm:
Sounds good @JoshRosen.
In general @yhuai it would have been better to give some more time for reviewers to get to ongoing conversations before commiting a patch under active review unless hotfix, thanks.

yhuai:
@mridulm Sure. Also, please feel free to leave more comments :)

JoshRosen:
This cherry-picked cleanly into branch-2.1 but there were some merge conflicts for branch-2.0, so I've opened #16358 for the branch-2.0 backport.

PR #12127, opened 2016-04-02
[SPARK-12981] [SQL] extract Pyhton UDF in physical plan

davies:
## What changes were proposed in this pull request?

Currently we extract Python UDFs into a special logical plan EvaluatePython in analyzer, But EvaluatePython is not part of catalyst, many rules have no knowledge of it , which will break many things (for example, filter push down or column pruning).

We should treat Python UDFs as normal expressions, until we want to evaluate in physical plan, we could extract them in end of optimizer, or physical plan.

This PR extract Python UDFs in physical plan.

Closes #10935
## How was this patch tested?

Added regression tests.

PR #16193, opened 2016-12-07
[SPARK-18766] [SQL] Push Down Filter Through BatchEvalPython (Python UDF)

gatorsmile:
Currently, when users use Python UDF in Filter, BatchEvalPython is always generated below FilterExec. However, not all the predicates need to be evaluated after Python UDF execution. Thus, this PR is to push down the determinisitc predicates through `BatchEvalPython`.
```Python
>>> df = spark.createDataFrame([(1, "1"), (2, "2"), (1, "2"), (1, "2")], ["key", "value"])
>>> from pyspark.sql.functions import udf, col
>>> from pyspark.sql.types import BooleanType
>>> my_filter = udf(lambda a: a < 2, BooleanType())
>>> sel = df.select(col("key"), col("value")).filter((my_filter(col("key"))) & (df.value < "2"))
>>> sel.explain(True)
```
Before the fix, the plan looks like
```
== Optimized Logical Plan ==
Filter ((isnotnull(value#1) && <lambda>(key#0L)) && (value#1 < 2))
+- LogicalRDD [key#0L, value#1]

== Physical Plan ==
*Project [key#0L, value#1]
+- *Filter ((isnotnull(value#1) && pythonUDF0#9) && (value#1 < 2))
   +- BatchEvalPython [<lambda>(key#0L)], [key#0L, value#1, pythonUDF0#9]
      +- Scan ExistingRDD[key#0L,value#1]
```

After the fix, the plan looks like
```
== Optimized Logical Plan ==
Filter ((isnotnull(value#1) && <lambda>(key#0L)) && (value#1 < 2))
+- LogicalRDD [key#0L, value#1]

== Physical Plan ==
*Project [key#0L, value#1]
+- *Filter pythonUDF0#9: boolean
   +- BatchEvalPython [<lambda>(key#0L)], [key#0L, value#1, pythonUDF0#9]
      +- *Filter (isnotnull(value#1) && (value#1 < 2))
         +- Scan ExistingRDD[key#0L,value#1]
```

Added both unit test cases for `BatchEvalPythonExec` and also add an end-to-end test case in Python test suite.

### DIFF
--- a/python/pyspark/sql/tests.py
+++ b/python/pyspark/sql/tests.py
@@ -360,6 +360,15 @@ class SQLTests(ReusedPySparkTestCase):
         [res] = self.spark.sql("SELECT MYUDF('')").collect()
         self.assertEqual("", res[0])
 
+    def test_udf_with_filter_function(self):
+        df = self.spark.createDataFrame([(1, "1"), (2, "2"), (1, "2"), (1, "2")], ["key", "value"])
+        from pyspark.sql.functions import udf, col
+        from pyspark.sql.types import BooleanType
+
+        my_filter = udf(lambda a: a < 2, BooleanType())
+        sel = df.select(col("key"), col("value")).filter((my_filter(col("key"))) & (df.value < "2"))
+        self.assertEqual(sel.collect(), [Row(key=1, value='1')])
+
     def test_udf_with_aggregate_function(self):
         df = self.spark.createDataFrame([(1, "1"), (2, "2"), (1, "2"), (1, "2")], ["key", "value"])
         from pyspark.sql.functions import udf, col, sum
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/python/ExtractPythonUDFs.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/python/ExtractPythonUDFs.scala
@@ -25,7 +25,7 @@ import org.apache.spark.sql.catalyst.expressions.aggregate.AggregateExpression
 import org.apache.spark.sql.catalyst.plans.logical.{Aggregate, LogicalPlan, Project}
 import org.apache.spark.sql.catalyst.rules.Rule
 import org.apache.spark.sql.execution
-import org.apache.spark.sql.execution.SparkPlan
+import org.apache.spark.sql.execution.{FilterExec, SparkPlan}
 
 
 /**
@@ -111,7 +111,15 @@ object ExtractPythonUDFs extends Rule[SparkPlan] {
   }
 
   def apply(plan: SparkPlan): SparkPlan = plan transformUp {
-    case plan: SparkPlan => extract(plan)
+    case plan: SparkPlan =>
+      val newPlan = extract(plan)
+      if (newPlan != plan) {
+        // Found and build BatchEvalPythonExec, and then push FilterExec
+        // through BatchEvalPythonExec
+        PushPredicateThroughBatchEvalPython.apply(newPlan)
+      } else {
+        plan
+      }
   }
 
   /**
@@ -163,6 +171,43 @@ object ExtractPythonUDFs extends Rule[SparkPlan] {
       } else {
         newPlan
       }
+    }
+  }
+}
+
+// This rule is to push deterministic predicates through BatchEvalPythonExec
+object PushPredicateThroughBatchEvalPython extends Rule[SparkPlan] with PredicateHelper {
+  def apply(plan: SparkPlan): SparkPlan = plan transform {
+    case filter @ FilterExec(_, child: BatchEvalPythonExec)
+        if child.expressions.forall(_.deterministic) =>
+      pushDownPredicate(filter, child.child) { predicate =>
+        child.withNewChildren(Seq(FilterExec(predicate, child.child)))
+      }
+  }
+
+  private def pushDownPredicate(
+      filter: FilterExec,
+      grandchild: SparkPlan)(insertFilter: Expression => SparkPlan): SparkPlan = {
+    // Only push down the predicates that is deterministic and all the referenced attributes
+    // come from grandchild.
+    val (candidates, containingNonDeterministic) =
+    splitConjunctivePredicates(filter.condition).span(_.deterministic)
+
+    val (pushDown, rest) = candidates.partition { cond =>
+      cond.references.subsetOf(grandchild.outputSet)
+    }
+
+    val stayUp = rest ++ containingNonDeterministic
+
+    if (pushDown.nonEmpty) {
+      val newChild = insertFilter(pushDown.reduceLeft(And))
+      if (stayUp.nonEmpty) {
+        FilterExec(stayUp.reduceLeft(And), newChild)
+      } else {
+        newChild
+      }
+    } else {
+      filter
     }
   }
 }
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/python/ExtractPythonUDFs.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/python/ExtractPythonUDFs.scala
@@ -177,20 +177,14 @@ object ExtractPythonUDFs extends Rule[SparkPlan] with PredicateHelper {
         // come from child.
         val (candidates, containingNonDeterministic) =
           splitConjunctivePredicates(filter.condition).span(_.deterministic)
+        // Python UDF is always deterministic
         val (pushDown, rest) = candidates.partition(!hasPythonUDF(_))
-        val stayUp = rest ++ containingNonDeterministic
-
         if (pushDown.nonEmpty) {
           val newChild = FilterExec(pushDown.reduceLeft(And), filter.child)
-          if (stayUp.nonEmpty) {
-            FilterExec(stayUp.reduceLeft(And), newChild)
-          } else {
-            newChild
-          }
+          FilterExec((rest ++ containingNonDeterministic).reduceLeft(And), newChild)
         } else {
           filter
         }
-
       case o => o
     }
   }
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/python/ExtractPythonUDFs.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/python/ExtractPythonUDFs.scala
@@ -168,12 +168,11 @@ object ExtractPythonUDFs extends Rule[SparkPlan] with PredicateHelper {
     }
   }
 
-  // Split the original FilterExec to two FilterExecs. The upper FilterExec only contains
-  // Python UDF and non-deterministic predicates.
+  // Split the original FilterExec to two FilterExecs. Only push down the first few predicates
+  // that are all deterministic.
   private def trySplitFilter(plan: SparkPlan): SparkPlan = {
     plan match {
       case filter: FilterExec =>
-        // Only push down the first few predicates that are all deterministic
         val (candidates, containingNonDeterministic) =
           splitConjunctivePredicates(filter.condition).span(_.deterministic)
         val (pushDown, rest) = candidates.partition(!hasPythonUDF(_))
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/python/ExtractPythonUDFs.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/python/ExtractPythonUDFs.scala
@@ -126,12 +126,11 @@ object ExtractPythonUDFs extends Rule[SparkPlan] with PredicateHelper {
       plan
     } else {
       val attributeMap = mutable.HashMap[PythonUDF, Expression]()
-      val splittedFilter = trySplitFilter(plan)
+      val splitFilter = trySplitFilter(plan)
       // Rewrite the child that has the input required for the UDF
-      val newChildren =
-        splittedFilter.children.map { child =>
+      val newChildren = splitFilter.children.map { child =>
         // Pick the UDF we are going to evaluate
-        val validUdfs = udfs.filter { case udf =>
+        val validUdfs = udfs.filter { udf =>
           // Check to make sure that the UDF can be evaluated with only the input of this child.
           udf.references.subsetOf(child.outputSet)
         }.toArray  // Turn it into an array since iterators cannot be serialized in Scala 2.10
@@ -152,7 +151,7 @@ object ExtractPythonUDFs extends Rule[SparkPlan] with PredicateHelper {
         sys.error(s"Invalid PythonUDF $udf, requires attributes from more than one child.")
       }
 
-      val rewritten = splittedFilter.withNewChildren(newChildren).transformExpressions {
+      val rewritten = splitFilter.withNewChildren(newChildren).transformExpressions {
         case p: PythonUDF if attributeMap.contains(p) =>
           attributeMap(p)
       }
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/python/ExtractPythonUDFs.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/python/ExtractPythonUDFs.scala
@@ -90,7 +90,7 @@ object ExtractPythonUDFFromAggregate extends Rule[LogicalPlan] {
  * This has the limitation that the input to the Python UDF is not allowed include attributes from
  * multiple child operators.
  */
-object ExtractPythonUDFs extends Rule[SparkPlan] {
+object ExtractPythonUDFs extends Rule[SparkPlan] with PredicateHelper {
 
   private def hasPythonUDF(e: Expression): Boolean = {
     e.find(_.isInstanceOf[PythonUDF]).isDefined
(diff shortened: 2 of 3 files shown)


sql/core/src/test/scala/org/apache/spark/sql/execution/python/BatchEvalPythonExecSuite.scala
```diff
@@ -0,0 +1,99 @@
...
+    super.afterAll()
+  }
+
+  test("Python UDF: push down deterministic FilterExec predicates") {
+    val df = Seq(("Hello", 4)).toDF("a", "b")
+      .where("dummyPythonUDF(b) and dummyPythonUDF(a) and a in (3, 4)")
+    val qualifiedPlanNodes = df.queryExecution.executedPlan.collect {
+      case f @ FilterExec(And(_: AttributeReference, _: AttributeReference), _) => f
+      case b: BatchEvalPythonExec => b
+      case f @ FilterExec(_: In, _) => f
```
gatorsmile: The physical plan has a few hidden nodes that are not shown in Explain output. Thus, I did not compare the result with the expected tree structure.

sql/core/src/test/scala/org/apache/spark/sql/execution/python/BatchEvalPythonExecSuite.scala
```diff
@@ -0,0 +1,99 @@
...
+    val df = Seq(("Hello", 4)).toDF("a", "b")
+      .where("dummyPythonUDF(a) and rand() > 3")
+    val qualifiedPlanNodes = df.queryExecution.executedPlan.collect {
+      case f: FilterExec => f
+      case b: BatchEvalPythonExec => b
+    }
+    assert(qualifiedPlanNodes.size == 2)
+  }
+
+  test("Python UDF refers to the attributes from more than one child") {
```
gatorsmile: This test case is not directly related to this PR. In the future, we need to add more unit test cases in Scala side for verifying `BatchEvalPythonExec` for improving the test case coverage.

sql/core/src/main/scala/org/apache/spark/sql/execution/python/ExtractPythonUDFs.scala
```diff
@@ -166,3 +174,40 @@ object ExtractPythonUDFs extends Rule[SparkPlan] {
     }
   }
 }
+
+// This rule is to push deterministic predicates through BatchEvalPythonExec
+object PushPredicateThroughBatchEvalPython extends Rule[SparkPlan] with PredicateHelper {
```
gatorsmile: Most of codes are from the optimizer rule `PushDownPredicate`. Not sure whether we should combine them. You know, this rule is for `SparkPlan`.

gatorsmile:
cc @cloud-fan @davies @liancheng

cloud-fan:
Would it be easier if we create a logical node for python evaluator? We do have one in Spark 1.6 but get removed in 2.0, not sure why

sql/core/src/test/scala/org/apache/spark/sql/execution/python/BatchEvalPythonExecSuite.scala
```diff
@@ -0,0 +1,99 @@
...
+
+import org.apache.spark.api.python.PythonFunction
+import org.apache.spark.sql.catalyst.expressions.{And, AttributeReference, In}
+import org.apache.spark.sql.execution.{FilterExec, SparkPlanTest}
+import org.apache.spark.sql.test.SharedSQLContext
+import org.apache.spark.sql.types.BooleanType
+
+class BatchEvalPythonExecSuite extends SparkPlanTest with SharedSQLContext {
+    import testImplicits.newProductEncoder
+    import testImplicits.localSeqToDatasetHolder
```
dongjoon-hyun: nit: indentation?

sql/core/src/main/scala/org/apache/spark/sql/execution/python/ExtractPythonUDFs.scala
```diff
@@ -166,3 +174,40 @@ object ExtractPythonUDFs extends Rule[SparkPlan] {
...
+      }
+  }
+
+  private def pushDownPredicate(
+      filter: FilterExec,
+      grandchild: SparkPlan)(insertFilter: Expression => SparkPlan): SparkPlan = {
+    // Only push down the predicates that is deterministic and all the referenced attributes
+    // come from grandchild.
+    val (candidates, containingNonDeterministic) =
+    splitConjunctivePredicates(filter.condition).span(_.deterministic)
```
dongjoon-hyun: nit. Indentation?

python/pyspark/sql/tests.py
```diff
@@ -360,6 +360,15 @@ def test_broadcast_in_udf(self):
         [res] = self.spark.sql("SELECT MYUDF('')").collect()
         self.assertEqual("", res[0])
 
+    def test_udf_with_filter_function(self):
+        df = self.spark.createDataFrame([(1, "1"), (2, "2"), (1, "2"), (1, "2")], ["key", "value"])
+        from pyspark.sql.functions import udf, col
+        from pyspark.sql.types import BooleanType
+
+        my_filter = udf(lambda a: a < 2, BooleanType())
+        sel = df.select(col("key"), col("value")).filter((my_filter(col("key"))) & (df.value < "2"))
```
gatorsmile: Nope. This case works well.

gatorsmile:
@cloud-fan Let me do a history search and see why we dropped the logical plan node `EvaluatePython`

gatorsmile:
https://github.com/apache/spark/pull/12127 dropped the node `EvaluatePython `.  Based on the PR description, we removed the node for the following reasons:

>Currently we extract Python UDFs into a special logical plan EvaluatePython in analyzer, But EvaluatePython is not part of catalyst, many rules have no knowledge of it , which will break many things (for example, filter push down or column pruning).
> We should treat Python UDFs as normal expressions, until we want to evaluate in physical plan, we could extract them in end of optimizer, or physical plan.

gatorsmile:
I also checked the plan of our 1.6.3 branch. The filter is not appropriately pushed down, even if we have the logical node `EvaluatePython`.
```
== Parsed Logical Plan ==
'Filter (PythonUDF#<lambda>('key) && (value#1 < 2))
+- Project [key#0L,value#1]
   +- LogicalRDD [key#0L,value#1], MapPartitionsRDD[4] at applySchemaToPythonRDD at NativeMethodAccessorImpl.java:-2

== Analyzed Logical Plan ==
key: bigint, value: string
Project [key#0L,value#1]
+- Filter (pythonUDF#2 && (value#1 < 2))
   +- EvaluatePython PythonUDF#<lambda>(key#0L), pythonUDF#2: boolean
      +- Project [key#0L,value#1]
         +- LogicalRDD [key#0L,value#1], MapPartitionsRDD[4] at applySchemaToPythonRDD at NativeMethodAccessorImpl.java:-2

== Optimized Logical Plan ==
Project [key#0L,value#1]
+- Filter (pythonUDF#2 && (value#1 < 2))
   +- EvaluatePython PythonUDF#<lambda>(key#0L), pythonUDF#2: boolean
      +- LogicalRDD [key#0L,value#1], MapPartitionsRDD[4] at applySchemaToPythonRDD at NativeMethodAccessorImpl.java:-2

== Physical Plan ==
Project [key#0L,value#1]
+- Filter (pythonUDF#2 && (value#1 < 2))
   +- !BatchPythonEvaluation PythonUDF#<lambda>(key#0L), [key#0L,value#1,pythonUDF#2]
      +- Scan ExistingRDD[key#0L,value#1]
```

dongjoon-hyun:
Retest this please

viirya:
As we can push down predicates through data source scan, those predicates should be already pushed down if they are in the query plan above data source scan node. This seems only work on ExistingRDD scan which the predicates cannot be pushed down.

So the question is, should we push down predicates to ExistingRDD scan? I think there is not much benefit.

viirya:
If we really want to do this, I'd suggest to push down predicates to rdd scan node during query planning stage. So we don't need to push down predicates to SparkPlan like this.

gatorsmile:
@viirya I did not get your points. Why pushing down predicates through Python UDF does not have significant benefit? Based on my understanding, it could greatly reduce the number of rows consumed/processed by UDF. Normally, UDF is much more expensive than the built-in expressions.

gatorsmile:
ExistingRDD might not be always the child of `Filter`. For example, 
```Python
>>> sel = df.select('key', 'value', rand()).filter((my_filter(col("key"))) & (df.value < "2"))
```
```
== Optimized Logical Plan ==
Filter ((isnotnull(value#1) && <lambda>(key#0L)) && (value#1 < 2))
+- Project [key#0L, value#1, rand(9089678730530723370) AS rand(9089678730530723370)#20]
   +- LogicalRDD [key#0L, value#1]

== Physical Plan ==
*Project [key#0L, value#1, rand(9089678730530723370)#20]
+- *Filter pythonUDF0#26: boolean
   +- BatchEvalPython [<lambda>(key#0L)], [key#0L, value#1, rand(9089678730530723370)#20, pythonUDF0#26]
      +- *Filter (isnotnull(value#1) && (value#1 < 2))
         +- *Project [key#0L, value#1, rand(9089678730530723370) AS rand(9089678730530723370)#20]
            +- Scan ExistingRDD[key#0L,value#1]
```

sql/core/src/main/scala/org/apache/spark/sql/execution/python/ExtractPythonUDFs.scala
gatorsmile: Good idea! The new commit does it.

sql/core/src/main/scala/org/apache/spark/sql/execution/python/ExtractPythonUDFs.scala
```diff
@@ -165,4 +167,31 @@ object ExtractPythonUDFs extends Rule[SparkPlan] {
...
+        // Only push down the predicates that is deterministic and all the referenced attributes
+        // come from child.
+        val (candidates, containingNonDeterministic) =
+          splitConjunctivePredicates(filter.condition).span(_.deterministic)
+        val (pushDown, rest) = candidates.partition(!hasPythonUDF(_))
+        val stayUp = rest ++ containingNonDeterministic
+
+        if (pushDown.nonEmpty) {
+          val newChild = FilterExec(pushDown.reduceLeft(And), filter.child)
+          if (stayUp.nonEmpty) {
```
davies: There are should be some UDFs, so this will not be empty

gatorsmile: True. : )

sql/core/src/main/scala/org/apache/spark/sql/execution/python/ExtractPythonUDFs.scala
```diff
@@ -165,4 +167,31 @@ object ExtractPythonUDFs extends Rule[SparkPlan] {
...
+  // Split the original FilterExec to two FilterExecs. The upper FilterExec only contains
+  // Python UDF and non-deterministic predicates.
+  private def trySplitFilter(plan: SparkPlan): SparkPlan = {
+    plan match {
+      case filter: FilterExec =>
+        // Only push down the predicates that is deterministic and all the referenced attributes
+        // come from child.
+        val (candidates, containingNonDeterministic) =
+          splitConjunctivePredicates(filter.condition).span(_.deterministic)
+        val (pushDown, rest) = candidates.partition(!hasPythonUDF(_))
```
gatorsmile: Sure.

gatorsmile: This will change the semantics. `span` and `partition` have different semantics. Thus, we still have to keep the existing behavior. 

Let me write a comment to explain `PythonUDF` is always assumed to deterministic.

sql/core/src/main/scala/org/apache/spark/sql/execution/python/ExtractPythonUDFs.scala
```diff
@@ -165,4 +167,25 @@ object ExtractPythonUDFs extends Rule[SparkPlan] {
       }
     }
   }
+
+  // Split the original FilterExec to two FilterExecs. The upper FilterExec only contains
+  // Python UDF and non-deterministic predicates.
+  private def trySplitFilter(plan: SparkPlan): SparkPlan = {
+    plan match {
+      case filter: FilterExec =>
+        // Only push down the predicates that is deterministic and all the referenced attributes
```
davies: Only push down the first few predicates that are all deterministic

sql/core/src/main/scala/org/apache/spark/sql/execution/python/ExtractPythonUDFs.scala
```diff
@@ -165,4 +167,25 @@ object ExtractPythonUDFs extends Rule[SparkPlan] {
...
+        // come from child.
+        val (candidates, containingNonDeterministic) =
+          splitConjunctivePredicates(filter.condition).span(_.deterministic)
+        // Python UDF is always deterministic
```
davies: Is this one useful? We just won't push down expressions that has Python UDFs here.

viirya:
@gatorsmile I do mean that we don't push down predicates into `ExistingRDD` scan as it doesn't help generally. But `BatchEvalPython` is a special case.

It doesn't matter `ExistingRDD` is the child of `Filter` or not. We already do the similar predicate pushdown in the rules like `FileSourceStrategy`.

If we push down predicates into `ExistingRDD` scan node during query planning. We can avoid to do predicate pushdown in SparkPlan. I think it is not a good idea.

cloud-fan:
is this a general problem? Does it apply to R UDF?

davies:
@cloud-fan There is no R UDF at this point.

davies:
If no objection in next two hours, I will merge this one into master.

gatorsmile:
@davies Just updated the code comments, as you suggested. It does not affect the code logics. Sorry for the late update.

gatorsmile:
@cloud-fan If the functions in `dapply` and `gapply` might be called as UDF in SparkR, we have very limited support. In the plan output, it is represented as `MapPartitionsInR` and `FlatMapGroupsInR`. More strictly, as @davies said, SparkR does not have actual SQL-level registered UDF.

gatorsmile:
@viirya I think your idea is trying to resolve a different issue. It does not apply to all the cases for PythonUDF pushdown.

gatorsmile:
retest this please

sql/core/src/main/scala/org/apache/spark/sql/execution/python/ExtractPythonUDFs.scala
```diff
@@ -126,8 +126,10 @@ object ExtractPythonUDFs extends Rule[SparkPlan] {
       plan
     } else {
       val attributeMap = mutable.HashMap[PythonUDF, Expression]()
+      val splittedFilter = trySplitFilter(plan)
       // Rewrite the child that has the input required for the UDF
-      val newChildren = plan.children.map { child =>
+      val newChildren =
+        splittedFilter.children.map { child =>
```
cloud-fan: nit: no need to start a new line here?

sql/core/src/test/scala/org/apache/spark/sql/execution/python/BatchEvalPythonExecSuite.scala
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  test("Python UDF: no push down on non-deterministic") {
+    val df = Seq(("Hello", 4)).toDF("a", "b")
+      .where("b > 4 and dummyPythonUDF(a) and rand() > 3")
+    val qualifiedPlanNodes = df.queryExecution.executedPlan.collect {
+      case f: FilterExec => f
+      case b: BatchEvalPythonExec => b
+    }
+    assert(qualifiedPlanNodes.size == 3)
```
cloud-fan: it's really hard to tell the correctness by checking the number of plan nodes...

cloud-fan:
It's a little hacky to me that we do optimization in a planner. How hard is it if we introduce a logical node for python evaluator? We can define an interface in catalyst, e.g. `ExternalUDFEvaluator`, so that R(or other languages in the future) UDF can also benefit from it.

davies:
@cloud-fan It's not trivial to do this in optimizer, for example, we should split one Filter into two, that will conflict with another optimizer rule, that combine two filter into one.

viirya:
If we add a logical node for python evaluator, we'd push down the Filter through it, so the optimizer rule won't combine two Filter into one again?

davies:
The reason we move the PythonUDFEvaluator from logical plan into physical plan, because this one-off break many things, many rules need to treat specially.

davies:
Pushing down predicates into data source is also during optimization in planner, I think this one is not the first that do optimization outside Optimizer.

sql/core/src/test/scala/org/apache/spark/sql/execution/python/BatchEvalPythonExecSuite.scala
gatorsmile: Let me improve them.

cloud-fan:
LGTM

viirya:
retest this please.

viirya:
LGTM

gatorsmile:
Thanks! Merging to master!

PR #16194, opened 2016-12-07
[SPARK-18767][ML] Unify Models' toString methods

zhengruifeng:
override models' toString methods to output some key info

## How was this patch tested?
existing tests

### DIFF
--- a/mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala
+++ b/mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala
@@ -932,7 +932,7 @@ class LogisticRegressionModel private[spark] (
   @Since("1.6.0")
   override def write: MLWriter = new LogisticRegressionModel.LogisticRegressionModelWriter(this)
 
-  @Since("2.1.0")
+  @Since("2.2.0")
   override def toString: String = {
     s"LogisticRegressionModel (uid=$uid) with $numClasses classes"
   }
--- a/mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala
+++ b/mllib/src/main/scala/org/apache/spark/ml/classification/LogisticRegression.scala
@@ -931,6 +931,11 @@ class LogisticRegressionModel private[spark] (
    */
   @Since("1.6.0")
   override def write: MLWriter = new LogisticRegressionModel.LogisticRegressionModelWriter(this)
+
+  @Since("2.1.0")
+  override def toString: String = {
+    s"LogisticRegressionModel (uid=$uid) with $numClasses classes"
+  }
 }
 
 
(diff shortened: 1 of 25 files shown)


hhbyyh:
I haven't really checked about it, but is it possible to just provide the default `toString` implementation in some upstream abstract class, like Model?
And we can do the similar thing for Estimator.

srowen:
Yes, it would be nicer to just have a default like `ClassName(uid=...)` at least. That however would not include the class-specific details like number of layers. I am neutral on whether that's so useful, but OK.

At least, this should not say `@Since("2.1.0")` since it can only go in for 2.2.0

zhengruifeng:
@srowen Ok, I will update the version. @hhbyyh as Owen said, if we add a `toString` in class `Model`, we still need to override it in most algs if we want to output some specific info: for example, current  `DecisionTreeClassifier` will output the depth and the number of nodes.

srowen (approved):
OK, toString can help a bit in the shell and in debugging. Although people might have different opinions about the format, this seems reasonable.

hhbyyh:
Sorry but I would still strongly prefer to override the toString() in Model or Params. The default implementation can provide more information about the params. And only in a few cases you need to override the default implementation.
More specifically,
```
  override def toString = s"${this.logName}(uid=$uid)" +
    this.params.map(p => p.name + ": " + $(p)).mkString(" {\n  ", "\n  ", "\n}")
    // or use this.extractParamMap().toString()
```

would give a result:
```
org.apache.spark.ml.clustering.NewModel(uid=1122) {
  featuresCol: features
  initMode: k-means||
  initSteps: 2
  k: 2
  maxIter: 20
  predictionCol: prediction
  seed: 1387313927
  tol: 1.0E-4
}
```

srowen:
Oh that's a good point, you can enumerate the params generically? that makes a lot more sense.

zhengruifeng:
Good point. This make me think of the usage of `instr.logParams(params: _*)` in training instrumentation [https://github.com/apache/spark/pull/15671].
IMIO, params are copied from its trainer, and some are not significant for the model.
For `KMeansModel`, I think only param `k` make sense, others are just the params to create the model.
I am not sure whether we should print all shared params of model.

srowen:
It still may be better than manually implementing so many toString methods. The worst case is that this carries some extra output.

sethah:
One problem I see is that some models have relevant information stored in class members instead of params. LogisticRegression for example would probably list the number of classes and the number of features, neither of which are params. Also, as @zhengruifeng mentioned, there are a lot of not-useful params that we would end up listing. However, I do not like the idea of implementing something different for every single estimator/model either. 

I guess I'm not entirely convinced the added benefit is worth the cost. What is the motivation for this, other than that it is convenient in the repl? Are there users asking for it?

zhengruifeng:
@sethah No. Just for the convenience in repl. It's somewhat confusing when the model's desc is just the uid of its trainer when I work in repl.

srowen:
To move towards a resolution, I'd either implement one generic `toString` that enumerates params, or not do anything at this stage. Up to you @zhengruifeng

sethah:
I lean towards doing nothing, unless we can find a solution that is both generic AND lists all/only relevant information.

zhengruifeng:
@sethah +1, I prefer not to do any update. I think this isuue will be resolved by the way after some refactor on SharedParam.

PR #16196, opened 2016-12-07
[SPARK-18231] Optimise SizeEstimator implementation

a-roberts:

Several improvements to the SizeEstimator for performance, most of the benefit comes from, when estimating, contending to not contending on multiple threads. There can be a small boost in uncontended scenarios from the removal of the synchronisation code but the cost of that synchronisation when not truly contended is low. On the PageRank workload for HiBench we see 10-15% performance improvements (measuring elapsed times on average) with both IBM's SDK for Java and OpenJDK 8. I don't see any changes other than noise for the other workloads on this benchmark.

## How was this patch tested?

Existing unit tests but there are problems to resolve.

I see SizeEstimatorSuite and SizeTrackerSuite failing with at least IBM Java now due to smaller sizes being reported than the test expects (let's see what happens with OpenJDK on the community runs). 

In SizeTrackerSuite I think the failures are caused by using ThreadLocalRandom and not Random - because with Random we see these tests passing again. Not sure how robust SizeTrackerSuite is though.

For performance testing I've used HiBench, large profile, with one executor ranging from 10g to 25g, experimenting with fixed and dynamic heaps. The Spark code I've based my results on is from December the 1st (master branch, so 2.1.0 snapshot).

More details on the optimisations (this being phase one and JDK agnostic) at www.spark.tc/improvements-to-the-sizeestimator-class

### DIFF
--- a/core/src/main/scala/org/apache/spark/util/SizeEstimator.scala
+++ b/core/src/main/scala/org/apache/spark/util/SizeEstimator.scala
@@ -19,10 +19,11 @@ package org.apache.spark.util
 
 import java.lang.management.ManagementFactory
 import java.lang.reflect.{Field, Modifier}
-import java.util.{IdentityHashMap, Random}
+import java.util.{IdentityHashMap, WeakHashMap}
+import java.util.concurrent.ThreadLocalRandom
 
 import scala.collection.mutable.ArrayBuffer
-import scala.runtime.ScalaRunTime
+import scala.concurrent.util.Unsafe
 
 import com.google.common.collect.MapMaker
 
@@ -89,7 +90,13 @@ object SizeEstimator extends Logging {
 
   // A cache of ClassInfo objects for each class
   // We use weakKeys to allow GC of dynamically created classes
-  private val classInfos = new MapMaker().weakKeys().makeMap[Class[_], ClassInfo]()
+  private val classInfos = new ThreadLocal[WeakHashMap[Class[_], ClassInfo]] {
+    override def initialValue(): java.util.WeakHashMap[Class[_], ClassInfo] = {
+      val toReturn = new WeakHashMap[Class[_], ClassInfo]()
+      toReturn.put(classOf[Object], new ClassInfo(objectSize, new Array[Int](0)))
+      return toReturn
+    }
+  }
 
   // Object and pointer sizes are arch dependent
   private var is64bit = false
@@ -119,8 +126,6 @@ object SizeEstimator extends Logging {
       }
     }
     pointerSize = if (is64bit && !isCompressedOops) 8 else 4
-    classInfos.clear()
-    classInfos.put(classOf[Object], new ClassInfo(objectSize, Nil))
   }
 
   private def getIsCompressedOops: Boolean = {
@@ -192,7 +197,7 @@ object SizeEstimator extends Logging {
    */
   private class ClassInfo(
     val shellSize: Long,
-    val pointerFields: List[Field]) {}
+    val fieldOffsets: Array[Int]) {}
 
   private def estimate(obj: AnyRef, visited: IdentityHashMap[AnyRef, AnyRef]): Long = {
     val state = new SearchState(visited)
@@ -221,8 +226,13 @@ object SizeEstimator extends Logging {
         case _ =>
           val classInfo = getClassInfo(cls)
           state.size += alignSize(classInfo.shellSize)
-          for (field <- classInfo.pointerFields) {
-            state.enqueue(field.get(obj))
+          // avoid an iterator based for loop for performance
+          var index = 0
+          val fieldCount = classInfo.fieldOffsets.length
+          val us = Unsafe.instance
+          while (index < fieldCount) {
+            state.enqueue(us.getObject(obj, classInfo.fieldOffsets(index).toLong))
+            index += 1
           }
       }
     }
@@ -233,7 +243,7 @@ object SizeEstimator extends Logging {
   private val ARRAY_SAMPLE_SIZE = 100 // should be lower than ARRAY_SIZE_FOR_SAMPLING
 
   private def visitArray(array: AnyRef, arrayClass: Class[_], state: SearchState) {
-    val length = ScalaRunTime.array_length(array)
+    val length = java.lang.reflect.Array.getLength(array)
     val elementClass = arrayClass.getComponentType()
 
     // Arrays have object header and length field which is an integer
@@ -243,24 +253,32 @@ object SizeEstimator extends Logging {
       arrSize += alignSize(length.toLong * primitiveSize(elementClass))
       state.size += arrSize
     } else {
+      // We know that the array we are dealing with is an array of references
+      // so explicitly expose this type so we can directly manipulate the array
+      // without help form the Scala runtime for efficency
       arrSize += alignSize(length.toLong * pointerSize)
       state.size += arrSize
+
+      val objArray = array.asInstanceOf[Array[AnyRef]]
 
       if (length <= ARRAY_SIZE_FOR_SAMPLING) {
         var arrayIndex = 0
         while (arrayIndex < length) {
-          state.enqueue(ScalaRunTime.array_apply(array, arrayIndex).asInstanceOf[AnyRef])
+          state.enqueue(objArray(arrayIndex))
           arrayIndex += 1
         }
       } else {
         // Estimate the size of a large array by sampling elements without replacement.
         // To exclude the shared objects that the array elements may link, sample twice
-        // and use the min one to calculate array size.
-        val rand = new Random(42)
+        //  and use the min one to calculate array size.
+        //  Use ThreadLocalRandom here since the random is only accessed from 1 thread
+        // and we can save the overhead of the full thread-safe Random
+        val rand = ThreadLocalRandom.current
         val drawn = new OpenHashSet[Int](2 * ARRAY_SAMPLE_SIZE)
-        val s1 = sampleArray(array, state, rand, drawn, length)
-        val s2 = sampleArray(array, state, rand, drawn, length)
+        val s1 = sampleArray(objArray, state, rand, drawn, length)
+        val s2 = sampleArray(objArray, state, rand, drawn, length)
         val size = math.min(s1, s2)
+
         state.size += math.max(s1, s2) +
           (size * ((length - ARRAY_SAMPLE_SIZE) / (ARRAY_SAMPLE_SIZE))).toLong
       }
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/apache/spark/util/SizeEstimator.scala
```diff
@@ -243,47 +253,59 @@ object SizeEstimator extends Logging {
...
+        //  Use ThreadLocalRandom here since the random is only accessed from 1 thread
+        // and we can save the overhead of the full thread-safe Random
+        val rand = ThreadLocalRandom.current
         val drawn = new OpenHashSet[Int](2 * ARRAY_SAMPLE_SIZE)
-        val s1 = sampleArray(array, state, rand, drawn, length)
-        val s2 = sampleArray(array, state, rand, drawn, length)
+        val s1 = sampleArray(objArray, state, rand, drawn, length)
+        val s2 = sampleArray(objArray, state, rand, drawn, length)
         val size = math.min(s1, s2)
+
```
mridulm: Changes to this method are excellent and should speed things up !

core/src/main/scala/org/apache/spark/util/SizeEstimator.scala
```diff
@@ -243,47 +253,59 @@ object SizeEstimator extends Logging {
...
-      rand: Random,
+      rand: ThreadLocalRandom,
       drawn: OpenHashSet[Int],
       length: Int): Long = {
     var size = 0L
-    for (i <- 0 until ARRAY_SAMPLE_SIZE) {
+    // avoid the use of an iterator derrived from the range syntax here for performance
+    var count = 0
+    val end = ARRAY_SAMPLE_SIZE
+    while (count <= end) {
```
mridulm: < end for until semantics

core/src/main/scala/org/apache/spark/util/SizeEstimator.scala
```diff
@@ -316,62 +338,40 @@ object SizeEstimator extends Logging {
...
-    }
-
-    // Should choose a larger size to be new shellSize and clearly alignedSize >= shellSize, and
-    // round up the instance filed blocks
-    shellSize = alignSizeUp(alignedSize, pointerSize)
-
     // Create and cache a new ClassInfo
-    val newInfo = new ClassInfo(shellSize, pointerFields)
-    classInfos.put(cls, newInfo)
+    val newInfo = new ClassInfo(shellSize, fieldOffsets.toArray)
```
mridulm: We are loosing out on padding due to allignment here which the earlier code was computing. No ?

core/src/main/scala/org/apache/spark/util/SizeEstimator.scala
a-roberts: Ah yes, should be just < not <=, will add into the next commit

core/src/main/scala/org/apache/spark/util/SizeEstimator.scala
```diff
@@ -119,8 +126,6 @@ object SizeEstimator extends Logging {
       }
     }
     pointerSize = if (is64bit && !isCompressedOops) 8 else 4
-    classInfos.clear()
-    classInfos.put(classOf[Object], new ClassInfo(objectSize, Nil))
```
a-roberts: No need to clear() because we're using ThreadLocal WeakHashMaps so the initialisation will occur once per thread, we put the new value in and we assume the size won't be changing which is going to be the case except for running in some debug modes*. As they're in a ThreadLocal WeakHashMap the classes can still be unloaded if no longer used.

* I'm referring to JDK debug modes such as fullspeeddebug where we can change class layouts at runtime. Normal execution modes prohibit changes to the class layout during execution without the class being unloaded and reloaded which will trigger the map entry to clear and be recreated due to its weakness (so in normal cases the map will always be correct)

core/src/main/scala/org/apache/spark/util/SizeEstimator.scala
a-roberts: Will look into this and determine if the padding is needed

core/src/main/scala/org/apache/spark/util/SizeEstimator.scala
```diff
@@ -89,7 +90,13 @@ object SizeEstimator extends Logging {
 
   // A cache of ClassInfo objects for each class
   // We use weakKeys to allow GC of dynamically created classes
-  private val classInfos = new MapMaker().weakKeys().makeMap[Class[_], ClassInfo]()
+  private val classInfos = new ThreadLocal[WeakHashMap[Class[_], ClassInfo]] {
+    override def initialValue(): java.util.WeakHashMap[Class[_], ClassInfo] = {
+      val toReturn = new WeakHashMap[Class[_], ClassInfo]()
+      toReturn.put(classOf[Object], new ClassInfo(objectSize, new Array[Int](0)))
+      return toReturn
```
mridulm: What I meant was, continue to use ThreadLocal, but maintain the MapMaker's result for thlocal.get()

And move the initilization to initialize() instead of in initialValue()

a-roberts: Got it, thanks will give this a try

core/src/main/scala/org/apache/spark/util/SizeEstimator.scala
mridulm: The map is expected to contain class info from the current execution; not from past runs (which might or might not be relevant - increasing the map size).

core/src/main/scala/org/apache/spark/util/SizeEstimator.scala
```diff
@@ -243,47 +253,59 @@ object SizeEstimator extends Logging {
...
       }
     }
   }
 
   private def sampleArray(
-      array: AnyRef,
+      array: Array[AnyRef],
       state: SearchState,
-      rand: Random,
+      rand: ThreadLocalRandom,
```
srowen: I don't think this has to change

core/src/main/scala/org/apache/spark/util/SizeEstimator.scala
```diff
@@ -89,7 +90,13 @@ object SizeEstimator extends Logging {
 
   // A cache of ClassInfo objects for each class
   // We use weakKeys to allow GC of dynamically created classes
-  private val classInfos = new MapMaker().weakKeys().makeMap[Class[_], ClassInfo]()
+  private val classInfos = new ThreadLocal[WeakHashMap[Class[_], ClassInfo]] {
+    override def initialValue(): java.util.WeakHashMap[Class[_], ClassInfo] = {
```
srowen: Nit: remove java.util, and 'return' below. "map" is better than "toReturn"

This is going to expand the memory footprint, because redundant copies of this info will be maintained per thread. Is the contention that significant?

core/src/main/scala/org/apache/spark/util/SizeEstimator.scala
```diff
@@ -221,8 +226,13 @@ object SizeEstimator extends Logging {
...
           val classInfo = getClassInfo(cls)
           state.size += alignSize(classInfo.shellSize)
-          for (field <- classInfo.pointerFields) {
-            state.enqueue(field.get(obj))
+          // avoid an iterator based for loop for performance
+          var index = 0
+          val fieldCount = classInfo.fieldOffsets.length
+          val us = Unsafe.instance
+          while (index < fieldCount) {
+            state.enqueue(us.getObject(obj, classInfo.fieldOffsets(index).toLong))
```
srowen: I understand avoiding reflection, but this is a dicier way to access fields of an object. I don't have a specific reason this would fail but the fact that it uses unsafe is riskier. Is this worth it?

core/src/main/scala/org/apache/spark/util/SizeEstimator.scala
```diff
@@ -243,47 +253,59 @@ object SizeEstimator extends Logging {
...
       state: SearchState,
-      rand: Random,
+      rand: ThreadLocalRandom,
       drawn: OpenHashSet[Int],
       length: Int): Long = {
     var size = 0L
-    for (i <- 0 until ARRAY_SAMPLE_SIZE) {
+    // avoid the use of an iterator derrived from the range syntax here for performance
+    var count = 0
+    val end = ARRAY_SAMPLE_SIZE
```
srowen: end is redundant here

core/src/main/scala/org/apache/spark/util/SizeEstimator.scala
```diff
@@ -316,62 +338,40 @@ object SizeEstimator extends Logging {
...
 
-    // iterate through the fields of this class and gather information.
-    for (field <- cls.getDeclaredFields) {
+    while (index < fieldCount) {
+      val field = fields(index)
       if (!Modifier.isStatic(field.getModifiers)) {
         val fieldClass = field.getType
         if (fieldClass.isPrimitive) {
-          sizeCount(primitiveSize(fieldClass)) += 1
+          if (cls == classOf[Double] || cls == classOf[Long]) {
```
srowen: This and the logic changes below aren't obviously OK. this seems to lose a lot of logic. I think this has to be explained or backed out

srowen (changes requested):
I'm kind of concerned that this is changing a lot, some of which isn't obviously without problems or risk, for marginal gains. I'd prefer to stick to obviously correct wins

rxin:
+1 on @srowen's suggestion. This change is not surgical at all. It is going to be difficult to guarantee no behavior change. If anything I'd trade off performance over correctness here.

srowen:
Ping @a-roberts -- I think some sections of this are clearly a win, like near https://github.com/apache/spark/pull/16196#discussion_r91270660  but maybe best to back out anything controversial. And touch up the style. Then I think this could be ready.

a-roberts:
Agreed, I'll be back working on this and answering the queries after the 2.1.0 release vote passes, that's my current priority as we're nearing the Christmas break period

srowen:
Ping to keep this on the radar; these couple PRs have been open a long time

a-roberts:
It's on my to-do list, working on this once I'm back from an end of year break, can I get a list of concerns here please? I'll run with them and I think one concern is that we'll underestimate and this'll lead to insufficient memory problems at runtime

Once we figure this out I can add the scenario(s) above as unit tests to ensure any changes here are entirely correct

srowen:
See the discussion above; I think the request is to back out much of this and leave only the clearly correct improvements like https://github.com/apache/spark/pull/16196#discussion_r91270660 There are still a number of small comments and questions outstanding, which you can go back and browse.

srowen:
Ping @a-roberts ; let's close this if not going to proceed, though at least part of it looks like a clear win.

a-roberts:
We can close it for now and I'll reopen it once the changes are more conservative and I've done plenty of testing/profiling - or anybody else could do so by keeping in only the changes we deem as safe, unfortunately I have too much other Spark work on at the moment to give this PR the attention it deserves

PR #9759, opened 2015-11-17
[SPARK-11753][SQL][test-hadoop2.2] Make allowNonNumericNumbers option work

viirya:
## What changes were proposed in this pull request?

Jackson suppprts `allowNonNumericNumbers` option to parse non-standard non-numeric numbers such as "NaN", "Infinity", "INF".  Currently used Jackson version (2.5.3) doesn't support it all. This patch upgrades the library and make the two ignored tests in `JsonParsingOptionsSuite` passed.
## How was this patch tested?

`JsonParsingOptionsSuite`.

PR #16199, opened 2016-12-07
[SPARK-18772][SQL] NaN/Infinite float parsing in JSON is inconsistent

NathanHowell:

This relaxes the parsing of `Float` and `Double` columns to properly support mixed case values of `NaN` and (+/-)`Infinity`, as well as properly supporting (+/-)`Inf`. Currently a string literal of `Nan` or `InfinitY` will cause a task to fail instead of placing the record in the corrupt record column, and `Inf` causes a failure instead of being a valid double.

## How was this patch tested?

Additional unit tests have been added

### DIFF
--- a/sql/core/src/test/scala/org/apache/spark/sql/execution/datasources/json/JsonSuite.scala
+++ b/sql/core/src/test/scala/org/apache/spark/sql/execution/datasources/json/JsonSuite.scala
@@ -1764,4 +1764,37 @@ class JsonSuite extends QueryTest with SharedSQLContext with TestJsonData {
     val df2 = spark.read.option("PREfersdecimaL", "true").json(records)
     assert(df2.schema == schema)
   }
+
+  test("SPARK-18772: Special floats") {
+    val records = sparkContext
+      .parallelize(
+        """{"a": "NaN"}""" ::
+          """{"a": "nAn"}""" ::
+          """{"a": "-iNf"}""" ::
+          """{"a": "inF"}""" ::
+          """{"a": "+Inf"}""" ::
+          """{"a": "-iNfInity"}""" ::
+          """{"a": "InFiNiTy"}""" ::
+          """{"a": "+InfiNitY"}""" ::
+          """{"a": "+Infi"}""" ::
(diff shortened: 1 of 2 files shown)


NathanHowell:
Hello @HyukjinKwon, can you take a look at this one? I am unsure if we should be accepting lowercased values like `nan` (versus strictly testing for `NaN`) but I think this PR matches the original intent of the code.

HyukjinKwon:
@NathanHowell Thank you for cc'ing me. I will try to take a look within tomorrow at my best.

HyukjinKwon:
@NathanHowell, while tracking down the history, I found similar PR including this in https://github.com/apache/spark/pull/9759/files#diff-8affe5ec7d691943a88e43eb30af656e (this seems reverted due to conflicts of `dev/deps/spark-deps-hadoop*` which is not related with this PR).

Would this make sense if we take out the valid changes from there? It seems safe to follow it as the changes there were checked by several reviewers.

sql/core/src/test/scala/org/apache/spark/sql/execution/datasources/json/JsonSuite.scala
```diff
@@ -1764,4 +1764,37 @@ class JsonSuite extends QueryTest with SharedSQLContext with TestJsonData {
     val df2 = spark.read.option("PREfersdecimaL", "true").json(records)
     assert(df2.schema == schema)
   }
+
+  test("SPARK-18772: Special floats") {
+    val records = sparkContext
```
HyukjinKwon: I think it would be nicer if it has some roundtrip tests in reading and writing.

HyukjinKwon:
What do you think about my suggestion @NathanHowell ?

NathanHowell:
@HyukjinKwon Good idea, I'll take another stab and try to revive the original pull request.

HyukjinKwon:
Gentle ping @NathanHowell, how is it going?

HyukjinKwon:
@NathanHowell, please let me know. I can pick up the commits and take over.

PR #16200, opened 2016-12-07
[SPARK-18773][core] Make commons-crypto config translation consistent.

vanzin:
This change moves the logic that translates Spark configuration to
commons-crypto configuration to the network-common module. It also
extends TransportConf and ConfigProvider to provide the necessary
interfaces for the translation to work.

As part of the change, I removed SystemPropertyConfigProvider, which
was mostly used as an "empty config" in unit tests, and adjusted the
very few tests that required a specific config.

I also changed the config keys for AES encryption to live under the
"spark.network." namespace, which is more correct than their previous
names under "spark.authenticate.".

Tested via existing unit test.


vanzin:
hmm. retest this please.

vanzin:
@zsxwing want to take a look?

zsxwing:
LGTM. Merging to master and 2.1. If this patch is not in 2.1.0, then I think we need to support old configs for several versions. Could you create a ticket in case we forget it?

zsxwing:
@vanzin there are conflicts with 2.1. Could you submit another PR for 2.1, please?

vanzin:
The original change is not in 2.1, we don't need this patch there.

zsxwing:
> The original change is not in 2.1, we don't need this patch there.

Nice :)

PR #16201, opened 2016-12-08
[SPARK-3359][DOCS] Fix greater-than symbols in Javadoc to allow building with Java 8

michalsenkyr:

The API documentation build was failing when using Java 8 due to incorrect character `>` in Javadoc.

Replace `>` with literals in Javadoc to allow the build to pass.

## How was this patch tested?

Documentation was built and inspected manually to ensure it still displays correctly in the browser

```
cd docs && jekyll serve
```

### DIFF
--- a/mllib/src/main/scala/org/apache/spark/ml/classification/DecisionTreeClassifier.scala
+++ b/mllib/src/main/scala/org/apache/spark/ml/classification/DecisionTreeClassifier.scala
@@ -81,7 +81,7 @@ class DecisionTreeClassifier @Since("1.4.0") (
    * E.g. 10 means that the cache will get checkpointed every 10 iterations.
    * This is only used if cacheNodeIds is true and if the checkpoint directory is set in
    * [[org.apache.spark.SparkContext]].
-   * Must be {@literal >= 1}.
+   * Must be at least 1.
    * (default = 10)
    * @group setParam
    */
--- a/mllib/src/main/scala/org/apache/spark/ml/classification/DecisionTreeClassifier.scala
+++ b/mllib/src/main/scala/org/apache/spark/ml/classification/DecisionTreeClassifier.scala
@@ -81,7 +81,7 @@ class DecisionTreeClassifier @Since("1.4.0") (
    * E.g. 10 means that the cache will get checkpointed every 10 iterations.
    * This is only used if cacheNodeIds is true and if the checkpoint directory is set in
    * [[org.apache.spark.SparkContext]].
-   * Must be &gt;= 1.
+   * Must be {@literal >= 1}.
    * (default = 10)
    * @group setParam
    */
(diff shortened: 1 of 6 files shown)


mllib/src/main/scala/org/apache/spark/ml/classification/GBTClassifier.scala
```diff
@@ -98,7 +98,7 @@ class GBTClassifier @Since("1.4.0") (
    * E.g. 10 means that the cache will get checkpointed every 10 iterations.
    * This is only used if cacheNodeIds is true and if the checkpoint directory is set in
    * [[org.apache.spark.SparkContext]].
-   * Must be >= 1.
+   * Must be &gt;= 1.
```
HyukjinKwon: I believe it should be {@literal >}. This would not be printed as `>` but `&gt` in javadoc IIRC.

michalsenkyr: Oh, I am sorry. I checked the scaladoc, which renders it correctly, and IDEA quick docs. Thanks. I'll fix it

michalsenkyr:
This time I inspected both the generated Javadoc and Scaladoc. It should be fine now.

rxin:
Jenkins, test this please.

mllib/src/main/scala/org/apache/spark/ml/classification/DecisionTreeClassifier.scala
```diff
@@ -81,7 +81,7 @@ class DecisionTreeClassifier @Since("1.4.0") (
    * E.g. 10 means that the cache will get checkpointed every 10 iterations.
    * This is only used if cacheNodeIds is true and if the checkpoint directory is set in
    * [[org.apache.spark.SparkContext]].
-   * Must be >= 1.
+   * Must be {@literal >= 1}.
```
HyukjinKwon: Actually, it might be already fine and too trivial to mention but in scaladoc, inlined tag becomes monospace text and this will be `<code>&gt;= 1</code>` in HTML whereas it becomes `&gt;= 1` in javadoc.

So, I just decided to use `{@literal >}` instead of `{@literal >= 1}` to minimise the appearance differences bettwen javadoc and scaladoc. This is trivial and maybe not meant to be forced but I just wanted note.

michalsenkyr: Yes, I read about that but figured that the literal text is probably meant to be a logical part of code. Which, in this case, is `>= 1`.

Making just `>` literal would make sense if you want to minimize the cosmetic differences between the characters. However, as I did not know whether the resulting code block would retain the same styling in the future, I opted for logical separation. If scaladoc later changed the styling of literals to, for example, the GitHub style for code, the result would look very bad. `>= 1` would still look relatively good.

But if the `{@literal >}` style is widely used in the Spark project, I can change it.

HyukjinKwon: Thanks for the detailed explanation. For me, I am fine with this.

srowen:
I see, so this is just another case where changes will keep breaking this. We do need a build that can run this at some point soon here to catch it. But yes, just keep fixing.

I would just write "Must be at least 1" instead of "Must be >= 1".

srowen:
Merged to master/2.1

PR #16202, opened 2016-12-08
[SPARK-18662][hotfix] Add new resource-managers directories to SparkLauncher.

vanzin:
These directories are added to the classpath of applications when testing or
using SPARK_PREPEND_CLASSES, otherwise updated classes are not seen. Also,
add the mesos directory which was missing.


vanzin:
I ran the yarn tests locally (without doing "sbt package" first) and they're happy.

vanzin:
@JoshRosen

rxin:
Was mesos never included?

vanzin:
Yeah, mesos was missing. But I guess that only affects a tiny number of people who'd like to use SPARK_PREPEND_CLASSES while changing that mesos code, so probably nobody noticed.

vanzin:
Merging to master to make the builds happy.

PR #16204, opened 2016-12-08
[SPARK-18775][SQL] Limit the max number of records written per file

rxin:
Currently, Spark writes a single file out per task, sometimes leading to very large files. It would be great to have an option to limit the max number of records written per file in a task, to avoid humongous files.

This patch introduces a new write config option `maxRecordsPerFile` (default to a session-wide setting `spark.sql.files.maxRecordsPerFile`) that limits the max number of records written to a single file. A non-positive value indicates there is no limit (same behavior as not having this flag).

## How was this patch tested?
Added test cases in PartitionedWriteSuite for both dynamic partition insert and non-dynamic partition insert.


sql/core/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala
```diff
@@ -821,8 +831,6 @@ private[sql] class SQLConf extends Serializable with CatalystConf with Logging {
 
   def warehousePath: String = new Path(getConf(StaticSQLConf.WAREHOUSE_PATH)).toString
 
-  def ignoreCorruptFiles: Boolean = getConf(IGNORE_CORRUPT_FILES)
```
rxin: i moved this closer to the controls for files.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/FileFormatWriter.scala
```diff
@@ -225,32 +228,50 @@ object FileFormatWriter extends Logging {
       taskAttemptContext: TaskAttemptContext,
       committer: FileCommitProtocol) extends ExecuteWriteTask {
 
-    private[this] var outputWriter: OutputWriter = {
+    private[this] var currentWriter: OutputWriter = _
```
ssimeonov: Looking through the code, three things stand out:

1. There is code duplication between `SingleDirectoryWriteTask` and `DynamicPartitionWriteTask` when it comes to current writer management and cleanup.

2. There is duplication within `releaseResources()` and `newOutputWriter()` of the write tasks when it comes to releasing resources. 

3. Write task state management is leaky because `releaseResources()` is called explicitly by `executeTask()`. Also, `releaseResources()` will be called twice when there are no exceptions and once if there is an exception in `execute()`, which is a bit confusing. 

What about asking the base trait to do a bit more work and present a stronger contract to its users, e.g.:

```scala
  private trait ExecuteWriteTask {

    protected[this] var currentWriter: OutputWriter = null

    def execute(iterator: Iterator[InternalRow]): Set[String] = {
      try {
        executeImp(iterator)
      } finally {
        releaseResources()
      }
    }

    /**
     * Writes data out to files, and then returns the list of partition strings written out.
     * The list of partitions is sent back to the driver and used to update the catalog.
     */
    protected def executeImp(iterator: Iterator[InternalRow]): Set[String]

    protected def resetCurrentWriter(): Unit = {
      if (currentWriter != null) {
        currentWriter.close()
        currentWriter = null
      }
    }

    protected def releaseResources(): Unit = {
      resetCurrentWriter()
    }
  }
```

ssimeonov (commented):
This will be a much-loved feature. :)

rxin:
cc @ericl

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/FileFormatWriter.scala
```diff
@@ -379,30 +405,40 @@ object FileFormatWriter extends Logging {
...
+          fileCounter = 0
+
+          releaseResources()
+          newOutputWriter(currentKey, getPartitionStringFunc, fileCounter)
+          val partitionPath = getPartitionStringFunc(currentKey).getString(0)
           if (partitionPath.nonEmpty) {
             updatedPartitions.add(partitionPath)
           }
+        } else if (description.maxRecordsPerFile > 0 &&
+            recordsInFile == description.maxRecordsPerFile) {
```
ericl: \>= is probably a little more clear intent, here and in the other write task

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/FileFormatWriter.scala
ericl: I'm wondering if we should just remove `SingleDirectoryWriteTask`. The few tests I tried still seem to pass with the dynamic implementation as well.

rxin: The dynamic one forces a sort, which is highly inefficient.

ericl: Ah ok. I guess you could still refactor out the record counting code to be shared, but I'm not sure it's worth it.

ericl:
LGTM

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/FileFormatWriter.scala
```diff
@@ -311,7 +335,10 @@ object FileFormatWriter extends Logging {
       } else {
         ""
       }
-      val ext = bucketId + description.outputWriterFactory.getFileExtension(taskAttemptContext)
+
+      // This must be in a form that matches our bucketing format. See BucketingUtils.
+      val ext = f"$bucketId.c$fileCounter%03d" +
+        description.outputWriterFactory.getFileExtension(taskAttemptContext)
```
mridulm: Is there an assumption here on number of files per output partition ?

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/FileFormatWriter.scala
```diff
@@ -379,30 +405,40 @@ object FileFormatWriter extends Logging {
       val sortedIterator = sorter.sortedIterator()
 
       // If anything below fails, we should abort the task.
+      var recordsInFile = 0
```
mridulm: long

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/FileFormatWriter.scala
rxin: No.

rxin:
cc @hvanhovell

rxin:
@hvanhovell don't forget this one!

sql/core/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala
```diff
@@ -466,6 +466,19 @@ object SQLConf {
...
     .createWithDefault(4 * 1024 * 1024)
 
+  val IGNORE_CORRUPT_FILES = SQLConfigBuilder("spark.sql.files.ignoreCorruptFiles")
+    .doc("Whether to ignore corrupt files. If true, the Spark jobs will continue to run when " +
+      "encountering corrupted or non-existing and contents that have been read will still be " +
+      "returned.")
+    .booleanConf
+    .createWithDefault(false)
+
+  val MAX_RECORDS_PER_FILE = SQLConfigBuilder("spark.sql.files.maxRecordsPerFile")
```
hvanhovell: Should we also mention that there is a limit to the number of files produced? This might not be the best location.

hvanhovell (approved):
One tiny comment regarding documentation. Otherwise LGTM.

sql/core/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala
rxin: the limit realistically speaking is so high that i doubt it'd matter unless this value is set to 1.

hvanhovell: Okay that is fair. Lets merge this.

hvanhovell:

PR #16205, opened 2016-12-08
[SPARK-18776][SS] Make Offset for FileStreamSource corrected formatted in json

tdas:

- Changed FileStreamSource to use new FileStreamSourceOffset rather than LongOffset. The field is named as `logOffset` to make it more clear that this is a offset in the file stream log.
- Fixed bug in FileStreamSourceLog, the field endId in the FileStreamSourceLog.get(startId, endId) was not being used at all. No test caught it earlier. Only my updated tests caught it.

Other minor changes
- Dont use batchId in the FileStreamSource, as calling it batch id is extremely miss leading. With multiple sources, it may happen that a new batch has no new data from a file source. So offset of FileStreamSource != batchId after that batch.

## How was this patch tested?

Updated unit test.


tdas:
@marmbrus @zsxwing

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/FileStreamSourceOffset.scala
```diff
@@ -0,0 +1,47 @@
...
+package org.apache.spark.sql.execution.streaming
+
+import org.json4s.NoTypeHints
+import org.json4s.jackson.Serialization
+
+/**
+ * Offset for the [[FileStreamSource]].
+ * @param logOffset  Position in the [[FileStreamSourceLog]]
+ */
+case class FileStreamSourceOffset(logOffset: Long) extends Offset {
```
tdas: I named it logoffset to make a more clear that this is a offset in the FileStreamSourceLog.

zsxwing:
LGTM. Merging to master and 2.1.

zsxwing:
Oh, please also support to read the old format in case RC2 is passed.

zsxwing:
LGTM pending tests

tdas:
Merging to master and 2.1

PR #16209, opened 2016-12-08
[SPARK-10849][SQL] Adds option to the JDBC data source write for user to specify database column type for the create table

sureshthalamati:
Currently JDBC data source creates tables in the target database using the default type mapping, and the JDBC dialect mechanism.  If users want to specify different database data type for only some of columns, there is no option available. In scenarios where default mapping does not work, users are forced to create tables on the target database before writing. This workaround is probably not acceptable from a usability point of view. This PR is to provide a user-defined type mapping for specific columns.

The solution is to allow users to specify database column data type for the create table  as JDBC datasource option(createTableColumnTypes) on write. Data type information can be specified in the same format as table schema DDL format (e.g: `name CHAR(64), comments VARCHAR(1024)`).

All supported target database types can not be specified ,  the data types has to be valid spark sql data types also.  For example user can not specify target database  CLOB data type. This will be supported in the follow-up PR.

Example:
```Scala
df.write
.option("createTableColumnTypes", "name CHAR(64), comments VARCHAR(1024)")
.jdbc(url, "TEST.DBCOLTYPETEST", properties)
```
## How was this patch tested?
Added new test cases to the JDBCWriteSuite

### DIFF
--- a/docs/sql-programming-guide.md
+++ b/docs/sql-programming-guide.md
@@ -1223,6 +1223,13 @@ the following case-insensitive options:
      This is a JDBC writer related option. If specified, this option allows setting of database-specific table and partition options when creating a table (e.g., <code>CREATE TABLE t (name string) ENGINE=InnoDB.</code>). This option applies only to writing.
    </td>
   </tr>
+  
+  <tr>
+    <td><code>createTableColumnTypes</code></td>
+    <td>
+     The database column data types to use instead of the defaults, when creating the table. Data type information should be specified as key(column name)-value(data type) pairs in JSON (e.g: <code>{"name":"varchar(128)", "comments":"clob(20k)"})</code>. You can use <code>org.apache.spark.sql.types.MetadataBuilder</code> to build the metadata and generate the JSON string required for this option. This option applies only to writing.
+    </td>
+  </tr>  
 </table>
 
 <div class="codetabs">
--- a/docs/sql-programming-guide.md
+++ b/docs/sql-programming-guide.md
@@ -1227,7 +1227,7 @@ the following case-insensitive options:
   <tr>
     <td><code>createTableColumnTypes</code></td>
     <td>
-     The database column data types to use instead of the defaults, when creating the table. Data type information should be specified as key(column name)-value(data type) pairs in JSON (e.g: <code>{"name":"varchar(128)", "comments":"clob(20k)"})</code>. You can use <code>org.apache.spark.sql.types.MetadataBuilder</code> to build the metadata and generate the JSON string required for this option. This option applies only to writing.
+     The database column data types to use instead of the defaults, when creating the table. Data type information should be specified in the same format as CREATE TABLE columns syntax (e.g: <code>"name CHAR(64), comments VARCHAR(1024)")</code>. The specified types should be valid spark sql data types. This option applies only to writing.
     </td>
   </tr>  
 </table>
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
@@ -31,7 +31,7 @@ import org.apache.spark.sql.catalyst.InternalRow
 import org.apache.spark.sql.catalyst.encoders.RowEncoder
 import org.apache.spark.sql.catalyst.expressions.SpecificInternalRow
 import org.apache.spark.sql.catalyst.parser.CatalystSqlParser
-import org.apache.spark.sql.catalyst.util.{DateTimeUtils, GenericArrayData}
+import org.apache.spark.sql.catalyst.util.{CaseInsensitiveMap, DateTimeUtils, GenericArrayData}
 import org.apache.spark.sql.jdbc.{JdbcDialect, JdbcDialects, JdbcType}
 import org.apache.spark.sql.types._
 import org.apache.spark.unsafe.types.UTF8String
@@ -682,18 +682,18 @@ object JdbcUtils extends Logging {
    * Compute the schema string for this RDD.
    */
   def schemaString(
-      schema: StructType,
+      df: DataFrame,
       url: String,
       createTableColumnTypes: Option[String] = None): String = {
     val sb = new StringBuilder()
     val dialect = JdbcDialects.get(url)
     val userSpecifiedColTypesMap = createTableColumnTypes
-      .map(parseUserSpecifiedCreateTableColumnTypes(schema, _))
+      .map(parseUserSpecifiedCreateTableColumnTypes(df, _))
       .getOrElse(Map.empty[String, String])
-    schema.fields foreach { field =>
+    df.schema.fields.foreach { field =>
       val name = dialect.quoteIdentifier(field.name)
-      val typ: String = userSpecifiedColTypesMap.get(field.name)
-        .getOrElse(getJdbcType(field.dataType, dialect).databaseTypeDefinition)
+      val typ = userSpecifiedColTypesMap
+        .getOrElse(field.name, getJdbcType(field.dataType, dialect).databaseTypeDefinition)
       val nullable = if (field.nullable) "" else "NOT NULL"
       sb.append(s", $name $typ $nullable")
     }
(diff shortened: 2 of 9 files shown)


gatorsmile:
@rxin @joshrosen @srowen Which options are preferred for supporting customized column types? Table-level JDBC option or column-level metadata property? Thanks!

gatorsmile:
retest this please

gatorsmile:
This PR is following the same way as what AVRO did in Hive. [This linked file](https://github.com/apache/spark/blob/master/sql/hive/src/test/scala/org/apache/spark/sql/hive/QueryPartitionSuite.scala#L81-L117) is an example shown in our test case:

gatorsmile:
@sureshthalamati https://github.com/apache/spark/pull/17171 has been resolved. Can you update your PR by allowing users to specify the schema in DDL format?

sureshthalamati:
@gatorsmile  sure.  I will update the PR with the DDL format approach.

sureshthalamati:
@gatorsmile I like the DDL schema format approach. But the method `CatalystSqlParser.parseTableSchema(sql)`  will work only if user wants to specify the target database datatype that also exists in Spark. For example if user wants to specify CLOB(200K) ; it will not work because that is not a valid data type in spark. 

How about simple comma separate list with restriction of ,(comma) can not be in the column name to use this option ?. I am guessing that would work in most of the  scenarios.

Any suggestions ?

maropu:
Sorry to cutting in though, IMHO we need to have general logic to inject user-defined types via `UDTRegistration` into the DDL parser (`CatalystSqlParser`). If we have the logic, we could use the types in a expression string.

gatorsmile:
Yes, we need to extend the DDL parser to support the general user-defined types.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
```diff
@@ -680,19 +681,63 @@ object JdbcUtils extends Logging {
...
+    // check duplicate columns in the user specified column types.
+    if (userColNames.distinct.length != userColNames.length) {
+      val duplicates = userColNames.groupBy(identity).collect {
+        case (x, ys) if ys.length > 1 => x
+      }.mkString(", ")
+      throw new AnalysisException(
+        s"Found duplicate column(s) in createTableColumnTypes option value: $duplicates")
+    }
+    // check user specified column names exists in the data frame schema.
+    val commonNames = userColNames.intersect(schema.fieldNames)
```
gatorsmile: Are we assuming the name comparison is always case sensitive?

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
```diff
@@ -680,19 +681,63 @@ object JdbcUtils extends Logging {
...
+
+    // char/varchar gets translated to string type. Real data type specified by the user
+    // is available in the field metadata as HIVE_TYPE_STRING
+    userSchema.fields.map(f =>
+      f.name -> {
+        (if (f.metadata.contains(HIVE_TYPE_STRING)) {
+          f.metadata.getString(HIVE_TYPE_STRING)
+        } else {
+          f.dataType.catalogString
+        }).toUpperCase
```
gatorsmile: We can create a partial function here.

sql/core/src/test/scala/org/apache/spark/sql/jdbc/JDBCWriteSuite.scala
```diff
@@ -362,4 +363,80 @@ class JDBCWriteSuite extends SharedSQLContext with BeforeAndAfter {
...
+      StructField("id", IntegerType) ::
+        StructField("first#name", StringType) ::
+        StructField("city", StringType) ::
+        StructField("descr", StringType) ::
+        Nil)
+    val df = spark.createDataFrame(sparkContext.parallelize(data), schema)
+    // Use database specific CHAR/VARCHAR types instead of String data type.
+    val createTableColTypes = "`first#name` VARCHAR(123), city CHAR(20)"
+    assert(JdbcUtils.schemaString(df.schema, url1, Option(createTableColTypes)) ==
+      s""""id" INTEGER , "first#name" VARCHAR(123) , "city" CHAR(20) , "descr" TEXT """)
```
gatorsmile: This is the case that users specify all the columns. In this case, we should mix the order of the columns. 

In addition, we also need a case that users only specify one/two columns.

maropu: nit: Drop interpolation `s` in the head.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
sureshthalamati: Thank you for the review. Good question., updated the PR with case-sensitive handling.  Now column names from user specified schema are matched with data frame schema based on the SQLConf.CASE_SENSITIVE flag.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
sureshthalamati: Done. Moved it to separate function. Thanks for the suggestion.

sql/core/src/test/scala/org/apache/spark/sql/jdbc/JDBCWriteSuite.scala
sureshthalamati: Thanks for review @maropu . Fixed it.

gatorsmile:
LGTM pending Jenkins

cc @rxin @joshrosen @srowen 

This is a nice option to have for JDBC users. If no further comment, I will merge it to master tomorrow. Thanks!

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
```diff
@@ -680,19 +681,71 @@ object JdbcUtils extends Logging {
...
     if (sb.length < 2) "" else sb.substring(2)
   }
 
   /**
+   * Parses the user specified createTableColumnTypes option value string specified in the same
+   * format as create table ddl column types, and returns Map of field name and the data type to
+   * use in-place of the default data type.
+   */
+  private def parseUserSpecifiedCreateTableColumnTypes(
+    df: DataFrame,
```
viirya: wrong indent.

docs/sql-programming-guide.md
```diff
@@ -1223,6 +1223,13 @@ the following case-insensitive options:
      This is a JDBC writer related option. If specified, this option allows setting of database-specific table and partition options when creating a table (e.g., <code>CREATE TABLE t (name string) ENGINE=InnoDB.</code>). This option applies only to writing.
    </td>
   </tr>
+  
+  <tr>
+    <td><code>createTableColumnTypes</code></td>
+    <td>
+     The database column data types to use instead of the defaults, when creating the table. Data type information should be specified in the same format as CREATE TABLE columns syntax (e.g: <code>"name CHAR(64), comments VARCHAR(1024)")</code>. The specified types should be valid spark sql data types. This option applies only to writing.
```
viirya: `The specified types should be valid spark sql data types`? What it means?

sql/core/src/test/scala/org/apache/spark/sql/jdbc/JDBCWriteSuite.scala
```diff
@@ -362,4 +363,147 @@ class JDBCWriteSuite extends SharedSQLContext with BeforeAndAfter {
...
+        if (typeName.contains("CHAR")) {
+          val charMaxLength = row.getInt(2)
+          assert(expectedTypes(row.getString(0)) == s"$typeName($charMaxLength)")
+        } else {
+          assert(expectedTypes(row.getString(0)) == typeName)
+        }
+      }
+    }
+
+    val data = Seq[Row](Row(1, "dave", "Boston", "electric cars"))
```
viirya: Do I miss something? Looks like you have 4 columns. But the schema has only 3 fields? Is it intentional?

viirya:
LGTM except for few minor comments.

docs/sql-programming-guide.md
sureshthalamati: VARCHAR(1024) is a valid data type in spark sql, it gets translated to String internally in Spark.  The data types specified in this property are meant for target database, using VARCHAR for example because many RDBMS does not have String data type.

Thank you for reviewing @viirya .

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala
sureshthalamati: Will fix it.

sql/core/src/test/scala/org/apache/spark/sql/jdbc/JDBCWriteSuite.scala
sureshthalamati: Forgot to delete the extra value. Will fix it.

gatorsmile:
Thanks! Merging to master.

docs/sql-programming-guide.md
viirya: Yeah, I see it is working internally. However, looks like this kind of types is not explicitly documented: http://spark.apache.org/docs/latest/sql-programming-guide.html#data-types

So I have a little concern regarding the description here.

cbyn:
@sureshthalamati is it possible to write strings as LONGTEXT? I'm having difficulty understanding which types are possible.

robbyki:
Is there a recommended workaround to achieve exactly this in spark 2.1? I'm going through several resources to try and understand how to maintain my schema created outside of spark and then just truncating my tables from spark followed by writing with a savemode of overwrite. My problem exactly this issue with respect to my db netezza failing when it sees spark trying to save a text data type so I then have to go specify in my new jdbc dialect to use varchar(n) which does work however that just replaces all of my varchar columns (different lengths for different columns) with whatever I specified in my dialect which is not what I want. How can I just have it save the TEXT as varchar without specifying a length in the custom dialect?

sureshthalamati:
@cbyn The specified types should be valid spark sql data types. LONGTEXT probably is not one of those types supported by spark sql syntax. 

@robbyki Problem with dialect as you noticed it will be same for all the columns as you noticed. Only workaround is to  create table explicitly in the Netezza , and the save it.  There is a truncate option also
if you need to empty the table before saving , that typically keeps the table as you created.

Please post questions  to spark user list , you will get answers quickly from other users and developers. People will not notice comments on the closed PRS.

cbyn:
Thanks @sureshthalamati. I thought the idea was to specify the destination database type. E.g. writing spark sql strings as VARCHAR works but VARCHAR is not a spark sql type. (I'm using the VARCHAR feature and I'm very grateful for the addition!)

clwang24:
@sureshthalamati  It's now can specify target database  CLOB or BLOB data type? I use "createTableColumnTypes" option when write BLOB data in ORACLE DataBase but return error "ora-00902 invalid datatype", the scala code just as shown below:
            df.write.mode(SaveMode.Overwrite)
                 .option("createTableColumnTypes", "id int, Name binary")
                 .jdbc(url1, "TEST.USERDBTYPETEST", properties)
     Then i try change column type as BLOB, but return error "DataType blob is not supported.(line 1, pos 437)",  the scala code just as shown below:
           df.write.mode(SaveMode.Overwrite)
                 .option("createTableColumnTypes", "id int, Name blob")
                 .jdbc(url1, "TEST.USERDBTYPETEST", properties)
     The last i try to no specify column type, but return another error "ora-12899 value too large for column (actual 581 maximum 255)" because StrignDataType default mappping varchar(255). Can you spare time to help me with this question? thx

PR #16210, opened 2016-12-08
[Core][SPARK-18778]Fix the scala classpath under some environment

libratiger:
under some environment, the Dscala.usejavacp=true option seems not work, pass the -usejavacp directly to the repl fix this.

## How was this patch tested?
we test in our cluster environment.

### DIFF
--- a/bin/spark-shell
+++ b/bin/spark-shell
@@ -50,11 +50,11 @@ function main() {
     # (see https://github.com/sbt/sbt/issues/562).
     stty -icanon min 1 -echo > /dev/null 2>&1
     export SPARK_SUBMIT_OPTS="$SPARK_SUBMIT_OPTS -Djline.terminal=unix"
-    "${SPARK_HOME}"/bin/spark-submit --class org.apache.spark.repl.Main --name "Spark shell" "$@"
+    "${SPARK_HOME}"/bin/spark-submit --class org.apache.spark.repl.Main -usejavacp --name "Spark shell" "$@"
     stty icanon echo > /dev/null 2>&1
   else
     export SPARK_SUBMIT_OPTS
-    "${SPARK_HOME}"/bin/spark-submit --class org.apache.spark.repl.Main --name "Spark shell" "$@"
+    "${SPARK_HOME}"/bin/spark-submit --class org.apache.spark.repl.Main -usejavacp --name "Spark shell" "$@"
   fi
 }
 


rxin:
What are the environments?

srowen:
Paging @jodersky 
I'm not sure about this ... isn't `-usejavacp` a legacy option? I don't know of any other reports of this so it is possibly specific to your env.

djvulee:
@rxin our jdk is jdk1.8.0_91, and we do not install the scala, the OS is Debian 4.6.4.

jodersky:
The command line parameter `-Dscala.usejavacp` is something directly passed to the JVM and should not be OS-dependent. I'm wondering if the issue may not arise from the way the parameter is defined in the spark-shell launcher scripts. Debian Etch is very old and probably has a older version of bash.

vanzin:
I wonder if it wouldn't be better to avoid this by moving the logic to add this parameter to `SparkSubmitCommandBuilder`. As in, if the class is the REPL's main class, add that option to the JVM.

(That also avoids the duplicate logic in Unix and Windows scripts.)

jodersky:
+1 to implementing it in the command builder. However I'm not sure if the command option `-usejavacp` or property `scala.usejavacp` is the suggested way to add the java class path to the REPL. I skimmed through the scala compiler's code and it seems both do slightly different things. I'll check this out in further detail

djvulee:
@jodersky Yes. I try different ways, here is the result:

```
SPARK_SUBMIT_OPTS="$SPARK_SUBMIT_OPTS -Dscala.usejavacp=true -usejavacp"
```

and

```
SPARK_SUBMIT_OPTS="$SPARK_SUBMIT_OPTS -Dscala.usejavacp=true -Dusejavacp"
```

will output
```
Exception in thread "main" java.lang.AssertionError: assertion failed: null
        at scala.Predef$.assert(Predef.scala:179)
        at org.apache.spark.repl.SparkIMain.initializeSynchronous(SparkIMain.scala:247)
        at org.apache.spark.repl.SparkILoop$$anonfun$org$apache$spark$repl$SparkILoop$$process$1.apply$mcZ$sp(SparkILoop.scala:990)
        at org.apache.spark.repl.SparkILoop$$anonfun$org$apache$spark$repl$SparkILoop$$process$1.apply(SparkILoop.scala:945)
        at org.apache.spark.repl.SparkILoop$$anonfun$org$apache$spark$repl$SparkILoop$$process$1.apply(SparkILoop.scala:945)
        at scala.tools.nsc.util.ScalaClassLoader$.savingContextLoader(ScalaClassLoader.scala:135)
        at org.apache.spark.repl.SparkILoop.org$apache$spark$repl$SparkILoop$$process(SparkILoop.scala:945)
        at org.apache.spark.repl.SparkILoop.process(SparkILoop.scala:1059)
        at org.apache.spark.repl.Main$.main(Main.scala:31)
        at org.apache.spark.repl.Main.main(Main.scala)
```

```
SPARK_SUBMIT_OPTS="$SPARK_SUBMIT_OPTS -usejavacp"
```
will output:
```
Unrecognized option: -usejavacp
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.
```

djvulee:
I find the reason, because we pass some  SPARK_SUBMIT_OPTS defined by ourself, so it seem that spark only parse the opts defined by ourself, ignore the ```-Dscala.usejavacp=true```. 

Since we want user to use the `SPARK_SUBMIT_OPTS`, the best way it to separate the ```-Dscala.usejavacp=true```  from the SPARK_SUBMIT_OPTS, maybe move to SparkSubmitCommandBuilder is good idea as suggestioned by @vanzin.

srowen:
I'm still not clear how to reproduce this. I can't repro with Java 8, Debian, on 2.x. Is this really necessary or somehow band-aiding some other deployment problem? that is, why does this work for anyone at all?

djvulee:
@srowen Sorry for late reply and thanks for your reproduce!

As I have mentioned in last reply, this is not a  environment problem, but a misunderstand of SPARK_SUBMIT_OPTS by ourself, or a deployment  problem. This works for anyone because few people use the ```SPARK_SUBMIT_OPTS``` option and do not put ```SPARK_SUBMIT_OPTS``` in the spark-env.sh file.

 It maybe better to separate the ```Dscala.usejavacp=true``` from the ```SPARK_SUBMIT_OPTS``` in the spark-shell file to avoid the misunderstanding.

vanzin:
The bug was filed as an improvement, and I think this is fine as an improvement. (By this I mean the suggestion of moving the handling of that option to Java code.)

That way we can get rid of `SPARK_SUBMIT_OPTS`, which is not documented and only the Spark shell scripts use. Users should be using "spark.driver.extraJavaOptions" and friends.

srowen:
OK. This PR doesn't move the handling / get rid of SPARK_SUBMIT_OPTS. Should this then be closed and the JIRA updated?

djvulee:
Yes, this PR do not consider well, I will close this and update the JIRA.

PR #16214, opened 2016-12-08
[SPARK-18325][SPARKR] Add example for using native R package.

yanboliang:
SparkR provides the ability to distribute the computations by ```spark.lapply```. In many cases, the distributed computation depends on third-party R packages, and we need to install corresponding packages to executor in advance which is tedious and unappeasable. Since SparkR is an interactive analytic tools and users may load lots of native R packages across the session, it's difficult to install all necessary packages before starting SparkR session, so I think it's helpful to provide an example to illustrate how to handle this scenario in an elegant way.

## How was this patch tested?
Offline test.

### DIFF
--- a/docs/sparkr.md
+++ b/docs/sparkr.md
@@ -489,6 +489,14 @@ print(model.summaries)
 {% endhighlight %}
 </div>
 
+##### spark.lapply with third-party packages
+Many of the SparkR jobs distributed by `spark.lapply` need supports from third-party packages. Rather than installing all necessary packages to all executors in advance,
+we could install them during the SparkR interactive session or script. Users can add a file or directory to be downloaded with this SparkR job on every node by `spark.addFile` firstly,
+download them to every executor node, and install them to be used.
+<div>
+{% include_example r/native-r-package.R %}
+</div>
+
 ## Running SQL Queries from SparkR
 A SparkDataFrame can also be registered as a temporary view in Spark SQL and that allows you to run SQL queries over its data.
 The `sql` function enables applications to run SQL queries programmatically and returns the result as a `SparkDataFrame`.
--- a/docs/sparkr.md
+++ b/docs/sparkr.md
@@ -478,8 +478,8 @@ should fit in a single machine. If that is not the case they can do something li
 ##### spark.lapply with third-party packages
 
 Many of the SparkR jobs distributed by `spark.lapply` need supports from third-party packages. Rather than installing all necessary packages to all executors in advance,
-we could install them during the SparkR interactive session or script. Users can add package files or directories by `spark.addFile` firstly,
-download them to every executor node, and install them.
+we could install them during the SparkR interactive session or script. Users can add package files or directories by `spark.addFile` first,
+which automatically download them to every executor node, and then install them.
 
 <div data-lang="r"  markdown="1">
 {% include_example r/native-r-package.R %}
--- a/docs/sparkr.md
+++ b/docs/sparkr.md
@@ -472,28 +472,16 @@ should fit in a single machine. If that is not the case they can do something li
 `dapply`
 
 <div data-lang="r"  markdown="1">
-{% highlight r %}
-# Perform distributed training of multiple models with spark.lapply. Here, we pass
-# a read-only list of arguments which specifies family the generalized linear model should be.
-families <- c("gaussian", "poisson")
-train <- function(family) {
-  model <- glm(Sepal.Length ~ Sepal.Width + Species, iris, family = family)
-  summary(model)
-}
-# Return a list of model's summaries
-model.summaries <- spark.lapply(families, train)
-
-# Print the summary of each model
-print(model.summaries)
-
-{% endhighlight %}
+{% include_example lapply r/ml/ml.R %}
 </div>
 
 ##### spark.lapply with third-party packages
+
 Many of the SparkR jobs distributed by `spark.lapply` need supports from third-party packages. Rather than installing all necessary packages to all executors in advance,
-we could install them during the SparkR interactive session or script. Users can add a file or directory to be downloaded with this SparkR job on every node by `spark.addFile` firstly,
-download them to every executor node, and install them to be used.
-<div>
+we could install them during the SparkR interactive session or script. Users can add package files or directories by `spark.addFile` firstly,
+download them to every executor node, and install them.
+
+<div data-lang="r"  markdown="1">
 {% include_example r/native-r-package.R %}
 </div>
 
--- a/examples/src/main/r/native-r-package.R
+++ b/examples/src/main/r/native-r-package.R
@@ -52,7 +52,7 @@ path <- spark.getSparkFiles(filename)
 costs <- exp(seq(from = log(1), to = log(1000), length.out = 5))
 train <- function(cost) {
     if("e1071" %in% rownames(installed.packages(lib = libDir)) == FALSE) {
-        install.packages(path, repos = NULL, type = "source")
+        install.packages(path, lib = libDir, repos = NULL, type = "source")
     }
     library(e1071)
     model <- svm(Species ~ ., data = iris, cost = cost)
--- a/examples/src/main/r/native-r-package.R
+++ b/examples/src/main/r/native-r-package.R
@@ -16,14 +16,14 @@
 #
 
 # This example illustrates how to install third-party R packages to executors
-# in your SparkR jobs distributed by `spark.lapply`.
+# in your SparkR jobs distributed by "spark.lapply".
 #
 # Note: This example will install packages to a temporary directory on your machine.
 #       The directory will be removed automatically when the example exit.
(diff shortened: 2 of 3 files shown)


yanboliang:
@felixcheung @shivaram

felixcheung:
Great - I think we should be careful with example code that install stuff automatically, especially if that could run on a cluster without any change.

Perhaps we should have a blocking prompt at the beginning with a warning message?

Also it would be better to have more description on what it is doing, where things would go (eg. where package would be installed to), and why?
And perhaps a section on programming guide? Or that is too public (depending on how much we want to promote such things)?

examples/src/main/r/native-r-package.R
```diff
@@ -0,0 +1,80 @@
...
+# ./bin/spark-submit examples/src/main/r/native-r-package.R
+
+# Load SparkR library into your R session
+library(SparkR)
+
+# Initialize SparkSession
+sparkR.session(appName = "SparkR-native-r-package-example")
+
+# Get the location of the default library
+libDir <- .libPaths()[1]
```
felixcheung: `install.packages()` already defaults to the default library.
as discussed before, perhaps this is more useful to be set to the application directory (eg. with YARN) because default libPath is usually secured?

examples/src/main/r/native-r-package.R
```diff
@@ -0,0 +1,80 @@
...
+
+# Install third-party R packages from CRAN to executors directly if it does not exist,
+# then the packages can be used by the corresponding task.
+
+# Perform distributed training of multiple models with spark.lapply
+costs <- exp(seq(from = log(1), to = log(1000), length.out = 5))
+train <- function(cost) {
+    if("e1071" %in% rownames(installed.packages(libDir)) == FALSE) {
```
felixcheung: I'd prefer `installed.packages(lib = libDir)` to be more clear

examples/src/main/r/native-r-package.R
```diff
@@ -0,0 +1,80 @@
...
+filename <- list.files(packagesDir, "^e1071")
+packagesPath <- file.path(packagesDir, filename)
+# Add the third-party R package to be downloaded with this Spark job on every node.
+spark.addFile(packagesPath)
+
+path <- spark.getSparkFiles(filename)
+costs <- exp(seq(from = log(1), to = log(1000), length.out = 5))
+train <- function(cost) {
+    if("e1071" %in% rownames(installed.packages(libDir)) == FALSE) {
+        install.packages(path, repos=NULL, type="source")
```
felixcheung: although if this is an example of how R package can be distributed, I wouldn't call `install.packages` here, because of secure location and duplications.

instead, this could do `library(e1071, lib.loc = path)` - ie. the package doesn't need to be "installed" to be loaded.

felixcheung: https://stat.ethz.ch/R-manual/R-devel/library/base/html/library.html

mengxr:
@yanboliang What happens if there are multiple executors on the same machine? Are there concurrency issues?

examples/src/main/r/native-r-package.R
yanboliang: Good suggestion!

examples/src/main/r/native-r-package.R
yanboliang: I removed the test which directly install R packages from CRAN consider that most of users don't allow the cluster to connect internet.

yanboliang:
@mengxr I think there is no (at least I did not found) concurrency issue when multiple executors on the same machine, since native R has lock to protect this. You can verify it by installing package to a shared directory in a single node in multi-thread way. However, we should recommend to install package to an executor associated directory which means different executor has its own R lib directory even if they are in the same machine. For YARN mode, it works well since each executor has its own R lib directory natively. 
@felixcheung I updated the examples and add corresponding session in SparkR user guide. I'm not sure whether such an example is appropriate, since it has many dependencies on environment. But I think for an interactive analytical tool, installing packages across the session is not very rare. Still, I'm open to hear your thought, thanks.

examples/src/main/r/native-r-package.R
```diff
@@ -0,0 +1,68 @@
...
+# Add the third-party R package to be downloaded with this Spark job on every node.
+spark.addFile(packagesPath)
+
+path <- spark.getSparkFiles(filename)
+costs <- exp(seq(from = log(1), to = log(1000), length.out = 5))
+train <- function(cost) {
+    if("e1071" %in% rownames(installed.packages(lib = libDir)) == FALSE) {
+        install.packages(path, repos = NULL, type = "source")
+    }
+    library(e1071)
```
mengxr: Did you try standalone mode? Even R installation has locks, this might have concurrency issues, e.g.:

1. Executor A executes L55 and sets a lock.
2. Executor B executes L55 and wait for the lock.
3. A finished installation and released the lock.
4. Executor B sets the lock and start executing L55, i.e., re-installing the same package.
5. A executes L57 but failed due to partially installed package.

Just my guess. But it may be worth testing.

yanboliang: Yeah, I run it in standalone mode several times, all work well. But I think I can not guarantee it works well always before more careful test, may be I'm lucky and not hit the concurrent issue. I'll figure out a religious test to verify it later. Thanks.

docs/sparkr.md
```diff
@@ -472,21 +472,17 @@ should fit in a single machine. If that is not the case they can do something li
 `dapply`
 
 <div data-lang="r"  markdown="1">
-{% highlight r %}
-# Perform distributed training of multiple models with spark.lapply. Here, we pass
-# a read-only list of arguments which specifies family the generalized linear model should be.
```
felixcheung: I think it'll be good to copy the comment into the example ml.R

docs/sparkr.md
```diff
@@ -472,21 +472,17 @@ should fit in a single machine. If that is not the case they can do something li
...
-families <- c("gaussian", "poisson")
-train <- function(family) {
-  model <- glm(Sepal.Length ~ Sepal.Width + Species, iris, family = family)
-  summary(model)
-}
-# Return a list of model's summaries
```
felixcheung: ditto, but minor here

docs/sparkr.md
```diff
@@ -472,21 +472,17 @@ should fit in a single machine. If that is not the case they can do something li
...
+</div>
 
-# Print the summary of each model
-print(model.summaries)
+##### spark.lapply with third-party packages
 
-{% endhighlight %}
+Many of the SparkR jobs distributed by `spark.lapply` need supports from third-party packages. Rather than installing all necessary packages to all executors in advance,
+we could install them during the SparkR interactive session or script. Users can add package files or directories by `spark.addFile` firstly,
+download them to every executor node, and install them.
```
felixcheung: this kind of sounds like the user will need to separately "download them to executor node" - perhaps instead say "by `spark.addFile` first, which automatically download them to every executor node, and then install them`?

examples/src/main/r/native-r-package.R
```diff
@@ -0,0 +1,68 @@
...
+filename <- list.files(packagesDir, "^e1071")
+packagesPath <- file.path(packagesDir, filename)
+# Add the third-party R package to be downloaded with this Spark job on every node.
+spark.addFile(packagesPath)
+
+path <- spark.getSparkFiles(filename)
+costs <- exp(seq(from = log(1), to = log(1000), length.out = 5))
+train <- function(cost) {
+    if("e1071" %in% rownames(installed.packages(lib = libDir)) == FALSE) {
+        install.packages(path, repos = NULL, type = "source")
```
felixcheung: re: comment [here](https://github.com/apache/spark/pull/16214#discussion_r91562425) if you already have the package content from sparkFiles you do not need to call `install.packages()`, which I think would be better without it.

examples/src/main/r/native-r-package.R
felixcheung: re: [here](https://github.com/apache/spark/pull/16214#discussion_r91562425) we shouldn't need to install it.

docs/sparkr.md
yanboliang: Sounds good, updated.

docs/sparkr.md
yanboliang: Done.

examples/src/main/r/native-r-package.R
yanboliang: Yeah, we have the package content, but it's source package rather than binary package, so we can not use ```library``` to load the package. This is the pain point for this example. If we illustrate this example with binary package, we should provide scripts for different os version, and it require all nodes in users' cluster should have the same architecture. So I use source package, I think it's a more universal example.

yanboliang:
After rethinking this issue, I found this is not very appropriate as an example delivered with Spark, since it depends on the environment that users running the script. I'm closing this PR, thanks for all your reviewing.

PR #16215, opened 2016-12-08
[Streaming] Update PairDStreamFunctions.scala

Jimmy-Newtron:

Added methods to retrieve keys or values of a pair DStream

## How was this patch tested?

NONE

Added two methods that are available into the PairRDDFunctions.scala

### DIFF
--- a/streaming/src/main/scala/org/apache/spark/streaming/dstream/PairDStreamFunctions.scala
+++ b/streaming/src/main/scala/org/apache/spark/streaming/dstream/PairDStreamFunctions.scala
@@ -46,6 +46,16 @@ class PairDStreamFunctions[K, V](self: DStream[(K, V)])
   }
 
   /**
+   * Return a DStream with the keys of each tuple.
+   */
+  def keys: DStream[K] = self.map(_._1)
+
+  /**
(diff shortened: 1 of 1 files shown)


srowen:
I don't think this worth new API methods since it's so easy to do this directly. Please see http://spark.apache.org/contributing.html too. I'd close this.

Jimmy-Newtron:
I think that this new API methods is going to bring consistency between PairRDDFunctions and PairDStreamFunctions by providing the same API to access the same data.

I see the use case of a streaming application that is consuming a Kafka topic to retrieve the values as the simplest example of usage of this new API.

I accept the PR rejection, even if the new API is matching only positive aspects of the Code Review Criteria.

Positives:
    Adds functionality needed by a large number of users
    Simple, targeted
Negatives, Risks: NONE

Best regards

rxin:
This is actually parity with PairRDDFunctions so I think it's OK to have them.

srowen:
I get it, it still seems not so worth it. The `DStream` API has always been mostly redundant with `foreachRDD` + the RDD API, and is itself maybe not something being actively updated anyway.

PR #16220, opened 2016-12-08
[SPARK-18796][SS]StreamingQueryManager should not block when starting a query

zsxwing:

Major change in this PR:
- Add `pendingQueryNames` and `pendingQueryIds` to track that are going to start but not yet put into `activeQueries` so that we don't need to hold a lock when starting a query.

Minor changes:
- Fix a potential NPE when the user sets `checkpointLocation` using SQLConf but doesn't specify a query name.
- Add missing docs in `StreamingQueryListener`

## How was this patch tested?

Jenkins


zsxwing:
cc @tdas @brkyvz

sql/core/src/main/scala/org/apache/spark/sql/streaming/StreamingQueryManager.scala
```diff
@@ -265,31 +278,53 @@ class StreamingQueryManager private[sql] (sparkSession: SparkSession) {
       }
```
tdas: Moving all the code to prepare and create the query object to a separate function would make this function smaller and easier to read.

sql/core/src/main/scala/org/apache/spark/sql/streaming/StreamingQueryManager.scala
```diff
@@ -265,31 +278,53 @@ class StreamingQueryManager private[sql] (sparkSession: SparkSession) {
...
+            pendingQueryIds -= query.id
+            userSpecifiedName.foreach(name => pendingQueryNames -= name)
+          }
+        }
+        query
+      } finally {
+        // In case `query.start()` fails
+        activeQueriesLock.synchronized { pendingQueryIds -= query.id }
+      }
+    } finally {
```
tdas: you can reduce the nested finally by keeping the id in a var.

sql/core/src/main/scala/org/apache/spark/sql/streaming/StreamingQueryManager.scala
```diff
@@ -206,21 +219,21 @@ class StreamingQueryManager private[sql] (sparkSession: SparkSession) {
...
+        pendingQueryNames += name
       }
+    }
+    try {
       val checkpointLocation = userSpecifiedCheckpointLocation.map { userSpecified =>
         new Path(userSpecified).toUri.toString
       }.orElse {
         df.sparkSession.sessionState.conf.checkpointLocation.map { location =>
-          new Path(location, name).toUri.toString
+          new Path(location, userSpecifiedName.getOrElse(UUID.randomUUID().toString)).toUri.toString
```
tdas: Can you add a unit test to verify this behavior. The test can something like 
- with userspecified location, see if it creates the right directory
- without userspecified location and without SQL conf, see if it creates a unique CP directory
- without userspecified location and with SQL conf, see if it creates a unique CP directory 
This can be in the StreamingQueryManager and would be easier if creating the query becomes a separate function.

tdas (changes requested):
The code needs simplification, otherwise the approach with pending variables is good.

sql/core/src/main/scala/org/apache/spark/sql/streaming/StreamingQueryManager.scala
```diff
@@ -181,6 +194,74 @@ class StreamingQueryManager private[sql] (sparkSession: SparkSession) {
     listenerBus.post(event)
   }
 
+  private def prepareAndCreateQuery(
```
tdas: just createQuery is fine

sql/core/src/main/scala/org/apache/spark/sql/streaming/StreamingQueryManager.scala
```diff
@@ -206,90 +287,70 @@ class StreamingQueryManager private[sql] (sparkSession: SparkSession) {
...
+          throw new IllegalStateException(
+            s"Cannot start query with id ${query.id} as another query with same id is " +
+              s"already active. Perhaps you are attempting to restart a query from checkpoint " +
+              s"that is already active.")
+        }
+        pendingQueryIds += query.id
+        queryId = query.id
       }
-
       query.start()
```
tdas: add comment explaining why this should not be inside a lock. so that this is less likely to regress in future.

sql/core/src/main/scala/org/apache/spark/sql/streaming/StreamingQueryManager.scala
```diff
@@ -206,90 +287,70 @@ class StreamingQueryManager private[sql] (sparkSession: SparkSession) {
...
-    activeQueriesLock.synchronized {
-      val name = userSpecifiedName match {
-        case Some(n) =>
-          if (activeQueries.values.exists(_.name == userSpecifiedName.get)) {
+    // `queryName` and `queryId` will be set when we add them into `pendingQueryNames` and
+    // `pendingQueryIds` so that we can remove them at the end of this method when an error happens.
+    var queryName: String = null
+    var queryId: UUID = null
+    try {
+      userSpecifiedName.foreach { name =>
```
tdas: Would be nice to add comment in each section, like "Make sure no other query with same name is active", "...query with same id...." , etc.

sql/core/src/test/scala/org/apache/spark/sql/streaming/test/DataStreamReaderWriterSuite.scala
```diff
@@ -575,4 +576,58 @@ class DataStreamReaderWriterSuite extends StreamTest with BeforeAndAfter {
...
+          val queryName = "test_query"
+          val ds = MemoryStream[Int].toDS
+          ds.writeStream
+            .format("memory")
+            .queryName(queryName)
+            .option("checkpointLocation", userCheckpointPath.getAbsolutePath)
+            .start()
+            .stop()
+          // Should not use SQLConf
+          assert(checkpointPath.listFiles().isEmpty, s"${checkpointPath.listFiles()} is not empty")
```
tdas: nit: its more useful to comment what this failure really means in the context of the test. For example, "SQLConf path is used even if user specified checkpointLoc"

tdas: If that feels too long, its okay to not have comments as well, its not that complex a test.

sql/core/src/test/scala/org/apache/spark/sql/streaming/test/DataStreamReaderWriterSuite.scala
```diff
@@ -575,4 +576,58 @@ class DataStreamReaderWriterSuite extends StreamTest with BeforeAndAfter {
...
+          // Should use `userCheckpointPath` directly
+          assert(userCheckpointPath.exists(), s"$userCheckpointPath doesn't exist")
```
tdas: same nit on failure message as above

tdas (commented):
Looks better, some nits.

zsxwing:
retest this please

zsxwing:
retest this please

zsxwing:
retest this please

tdas (commented):
LGTM. only nits.

zsxwing:
retest this please

zsxwing:
@tdas `org.apache.spark.sql.kafka010.KafkaSourceStressForDontFailOnDataLossSuite.stress test for failOnDataLoss=false` is still flaky. I think maybe we should also recreate the consumer in KafkaSource when an error happens. What do you think?

tdas:
I think thats a good idea.

PR #16225, opened 2016-12-09
[SPARK-14932][SQL] Allow DataFrame.replace() to replace values with None

jiayue-zhang:

Allow DataFrame.replace() to replace with None/null values.

## How was this patch tested?

Python doctest and unit test. Scala unit test.


HyukjinKwon:
Would we need a test for Scala too? I checked this by myself and it seems working fine with Scala though. I could argue that this affects the language-specific functions of both Python and Scala as a not strong opinion.

sql/core/src/main/scala/org/apache/spark/sql/DataFrameNaFunctions.scala
```diff
@@ -342,11 +342,14 @@ final class DataFrameNaFunctions private[sql](df: DataFrame) {
     }
 
     // replacementMap is either Map[String, String] or Map[Double, Double] or Map[Boolean,Boolean]
-    val replacementMap: Map[_, _] = replacement.head._2 match {
-      case v: String => replacement
-      case v: Boolean => replacement
-      case _ => replacement.map { case (k, v) => (convertToDouble(k), convertToDouble(v)) }
-    }
+    val replacementMap: Map[_, _] =
+      if (replacement.head._2 == null)
```
JoshRosen: I think you can just write `case null =>` here and thereby avoid the `if` `else`

bravo-zhang: I tried that but failed. Scala doesn't allow pattern matching with null, scala.Nothing and scala.Null

bravo-zhang: Oh, wait `case null =>` works. I tried `case v null` which doesn't. Let me modify this. Thanks!

gatorsmile:
ok to test

gatorsmile:
@bravo-zhang Could you resolve the conflicts? I will review it then. Thanks!

bravo-zhang:
Thanks for taking a look, @gatorsmile The conflicts have been resolved.
I appreciate if @zero323 can take a look as well since you made improvement on this function recently.

python/pyspark/sql/dataframe.py
```diff
@@ -1391,9 +1401,10 @@ def all_of_(xs):
                 "to_replace should be a float, int, long, string, list, tuple, or dict. "
                 "Got {0}".format(type(to_replace)))
 
-        if not isinstance(value, valid_types) and not isinstance(to_replace, dict):
+        if not isinstance(value, valid_types) and value is not None \
```
holdenk: So slightly jet lagged style question: would this be clearer if we just add type(None) to L1398? I know PEP8 says we should only use `is` `is not` for checking if something is none rather than depending on the implicit conversion to boolean -- but since really checking the type here we aren't really in danger of that. (This is just a suggestion to make it easier to read - if others think its easier to read this way thats fine :)). @davies ?

python/pyspark/sql/dataframe.py
```diff
@@ -1409,7 +1420,7 @@ def all_of_(xs):
         if isinstance(to_replace, (float, int, long, basestring)):
             to_replace = [to_replace]
 
-        if isinstance(value, (float, int, long, basestring)):
+        if isinstance(value, (float, int, long, basestring)) or value is None:
```
holdenk: same as above

python/pyspark/sql/dataframe.py
```diff
@@ -1423,7 +1434,9 @@ def all_of_(xs):
             subset = [subset]
 
         # Verify we were not passed in mixed type generics."
-        if not any(all_of_type(rep_dict.keys()) and all_of_type(rep_dict.values())
+        if not any(all_of_type(rep_dict.keys())
+                   and (all_of_type(rep_dict.values())
+                        or list(rep_dict.values()).count(None) == len(rep_dict))
```
holdenk: In the Scala code null is allowed in to be the replacement value for some of the elements but not in the Python code. Is this intentional? If so we should document it clearly and expand on the error message bellow (otherwise we should make it more flexible).

holdenk (commented):
Thanks for working on this, sorry its taken so long to review. I did a first read through the Python side and I've got two minor questions. Hopefully @zero323,  @davie , or @gatorsmile can also have some time once the current release is finished to take a look.

gatorsmile:
We are closing it due to inactivity. please do reopen if you want to push it forward. Thanks!

bravo-zhang:
@holdenk Thanks for review. I'll combine type(None) in the `isinstance`. I also made Scala and Python to accept null more generally and in the same way.
PR is reopened at: https://github.com/apache/spark/pull/18820
@gatorsmile please also take a look.

PR #16228, opened 2016-12-09
[SPARK-17076] [SQL] Cardinality estimation for join based on basic column statistics

wzhfy:

Support cardinality estimation and stats propagation for all join types.

Limitations:
- For inner/outer joins without any equal condition, we estimate it like cartesian product.
- For left semi/anti joins, since we can't apply the heuristics for inner join to it, for now we just propagate the statistics from left side. We should support them when other advanced stats (e.g. histograms) are available in spark.

## How was this patch tested?

Add a new test suite.


wzhfy:
cc @rxin @srinathshankar @cloud-fan

wzhfy:
I still leave two issues undecided:
1. Where can we turn on/off cbo estimation? I think we need to have such switch, because otherwise we will use statistics as long as they are in the metastore, but they can become stale.
2. Currently we use column name as the key for column statistics, which is problematic because if the output of join have columns from different tables with the same column name, they can't be distinguished. Can we use a combination string like table name + column name? Will it have problem in case of table alias?

Tagar:
2. That is great. Would it be easier to use FK, when it is available (see HMS has FKs since Hive 2.1: https://issues.apache.org/jira/browse/HIVE-13076), and if FK between columns is not defined, then use stats. 

Also, do I understand correctly, that the assumption is if two tables being joined by columns with the same name, join columns have the same stats / set of values?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/basicLogicalOperators.scala
```diff
@@ -325,15 +326,17 @@ case class Join(
...
-    case _ =>
-      // make sure we don't propagate isBroadcastable in other joins, because
-      // they could explode the size.
-      super.statistics.copy(isBroadcastable = false)
-  }
+  override lazy val statistics: Statistics = JoinEstimation.estimate(this).getOrElse(
+    joinType match {
+      case LeftAnti | LeftSemi =>
+        // LeftSemi and LeftAnti won't ever be bigger than left
+        left.statistics.copy()
```
rxin: why do we need copy here?

cloud-fan: `Statistics` is immutable, I think it's safe without copy.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/estimation/JoinEstimation.scala
```diff
@@ -0,0 +1,175 @@
...
+      newMax: Option[Any]): ColumnStat = {
+    origin.copy(distinctCount = newDistinctCount, min = newMin, max = newMax, nullCount = 0)
+  }
+
+  def extractJoinKeys(
+      leftKeys: Seq[Expression],
+      rightKeys: Seq[Expression]): Seq[(AttributeReference, AttributeReference)] = {
+    leftKeys.zip(rightKeys).flatMap {
+      case (ExtractAttr(left), ExtractAttr(right)) => Some((left, right))
+      // Currently we don't deal with equal joins like key1 = key2 + 5.
```
cloud-fan: If we are not ready for expressions, I think we should not handle `Cast` either, as it may be tricky to handle overflow correctly(e.g. cast long to int)

wzhfy:
@Tagar Thanks for sharing this information. Yes, it would be better to use PK/FK, but it won't be done in this pr, and we need to implement PK/FK constraints in Spark first. 

> the assumption is if two tables being joined by columns with the same name, join columns have the same stats / set of values?

It is true for inner join, but not true for outer joins, right?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/basicLogicalOperators.scala
wzhfy: ok

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/estimation/JoinEstimation.scala
wzhfy: Yes, but I'm a little worried about losing estimation opportunities because of this rare case (in my understanding this kind of downgrading cast is not common)?

wzhfy: Can we define a new parent class for such downgrading cast? Will this change be big?

cloud-fan: I don't think we should add a new expression just for the current implementation limitations. BTW, only handle `Cast` may also lose a lot of estimation opportunities, we should support expression before we release this feature.

Tagar:
@wzhfy, thanks for the feedback.
For outer joins, cardinaility estimates should be:
- left_outer_join_cardinality(table_A, table_B) = MAX(cardinality(A), inner_join_cardinality(table_A, table_B))
  So with left outer join, you'll get join cardinality at least number of the rows on the left (table A);
- right outer join is similar to left outer join (with exception s/A/B/g);
- full_outer_join_cardinality(table_A, table_B) = cardinality(A) + cardinality(B) - inner_join_cardinality(table_A, table_B)).
Does this sound about right?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/estimation/JoinEstimation.scala
```diff
@@ -0,0 +1,175 @@
...
+   * The number of rows of A inner join B on A.k1 = B.k1 is estimated by this basic formula:
+   * T(A IJ B) = T(A) * T(B) / max(V(A.k1), V(B.k1)), where V is the number of distinct values of
+   * that column. The underlying assumption for this formula is: each value of the smaller domain
+   * is included in the larger domain.
+   * Generally, inner join with multiple join keys can also be estimated based on the above
+   * formula:
+   * T(A IJ B) = T(A) * T(B) / (max(V(A.k1), V(B.k1)) * max(V(A.k2), V(B.k2)) * ... * max(V(A.kn), V(B.kn)))
+   * However, the denominator can become very large and excessively reduce the result, so we use a
+   * conservative strategy to take only the largest max(V(A.ki), V(B.ki)) as the denominator.
+   *
```
ioana-delaney: Can we also include a short description of how column stats are computed after the join?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/estimation/JoinEstimation.scala
```diff
@@ -0,0 +1,175 @@
...
+   * statsitics.
+   * The number of rows of A inner join B on A.k1 = B.k1 is estimated by this basic formula:
+   * T(A IJ B) = T(A) * T(B) / max(V(A.k1), V(B.k1)), where V is the number of distinct values of
+   * that column. The underlying assumption for this formula is: each value of the smaller domain
+   * is included in the larger domain.
+   * Generally, inner join with multiple join keys can also be estimated based on the above
+   * formula:
+   * T(A IJ B) = T(A) * T(B) / (max(V(A.k1), V(B.k1)) * max(V(A.k2), V(B.k2)) * ... * max(V(A.kn), V(B.kn)))
+   * However, the denominator can become very large and excessively reduce the result, so we use a
+   * conservative strategy to take only the largest max(V(A.ki), V(B.ki)) as the denominator.
```
ioana-delaney: They probably estimate for the number of distinct values in a vector of columns using uniform distribution.

wzhfy:
@Tagar For full outer join, how about cardinality = MAX(card(A) + card(B), innerCard(AB)) ?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/estimation/JoinEstimation.scala
wzhfy: Yes, of course

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/estimation/JoinEstimation.scala
```diff
@@ -0,0 +1,175 @@
...
+      }
+
+      Some(Statistics(
+        sizeInBytes = outputRows * getRowSize(join.output, outputColStats.toMap),
+        rowCount = Some(outputRows),
+        colStats = outputColStats.toMap,
+        isBroadcastable = false))
+
+    case _ => None
+  }
```
ioana-delaney: I have a design comment. Given a join, this function computes predicate selectivity, plan cardinality, and column statistics. I wonder if it would make sense to encapsulate predicate selectivity computation in its own function i.e. selectivity is a property of a predicate and cardinality (i.e. number of rows) is a property of the data stream. Also, there might be different ways to compute selectivity of a predicate (e.g. uniform vs non uniform distribution) and therefore it might make sense to separate the computation of the two properties. Then, maybe in the future, selectivity hint might be used to overwrite the default CBO selectivity computation.

Tagar:
@wzhfy, it's easier to check validity of these type of expressions when you look at extreme cases.
Your formula for full outer join cardinality,

> cardinality = MAX(card(A) + card(B), innerCard(AB)) 

in one of extreme cases when set(A) and set(B) are the same sets, then calculated cardinality would be 2 times more of the actual cardinality.

While 

> full_outer_join_cardinality(table_A, table_B) = cardinality(A) + cardinality(B) - inner_join_cardinality(table_A, table_B))

will produce correct result. 

ps. I find this visualization http://www.radacad.com/wp-content/uploads/2015/07/joins.jpg very helpful. 
https://en.wikipedia.org/wiki/Inclusion%E2%80%93exclusion_principle A U B = A + B - A \ B

Hope this helps. Thanks!

wzhfy:
@Tagar We can always find extreme cases to which these formula can't apply. In my opinion, it's better to over-estimate than under-estimate, which can lead to OOM problems, e.g. broadcast a very large result.

If A is a big table and B is a small one, every A.k has a match in B (a common case for PK and FK), then 

> cardinality(A) + cardinality(B) - inner_join_cardinality(table_A, table_B))

becomes card(B), which is dramatically smaller than the real outer join card. Even more, it can be negative if all A.k and B.k has the same value, the inner join part becomes a cartesian product.

This formula,
> cardinality = MAX(card(A) + card(B), innerCard(AB))

although over estimates sometimes, it's still obviously better than the original one in spark: card(A) * card(B).

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/estimation/JoinEstimation.scala
wzhfy: Thanks for the advice. We will consider to have such function when we check in the code for Filter estimation, then we can look at this more comprehensively. I think the pr will be submitted soon.

Tagar:
@wzhfy, I think overestimating cardinality could be as bad as underestimating.
For example, Optimizer could prematurely switch to SortMergeJoin when it could used broadcast hash join.
But I agree, this PR is a great improvement over current cardinality estimates.

wzhfy:
To solve [the two issues](https://github.com/apache/spark/pull/16228#issuecomment-265946299) I mentioned above, a separate PR is sent [here](https://github.com/apache/spark/pull/16323).
We need to rebase this one after that one is resolved.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/estimation/JoinEstimation.scala
```diff
@@ -0,0 +1,310 @@
...
+        case RightOuter =>
+          // All rows from right side should be in the result.
+          rightRows.max(innerRows)
+        case FullOuter =>
+          // Simulate full outer join as obtaining the number of elements in the union of two
+          // finite sets: A \cup B = A + B - A \cap B => A FOJ B = A + B - A IJ B.
+          // But the "inner join" part can be much larger than A \cap B, making the simulated
+          // result much smaller. To prevent this, we choose the larger one between the simulated
+          // part and the inner part.
+          (leftRows + rightRows - innerRows).max(innerRows)
```
wzhfy: @Tagar I take your advice about full outer join, but with a little change by lower bounding it using `innerRows`.

wzhfy:
retest this please

wzhfy:
This pr is rebased and ready for review. @rxin

sameeragarwal:
cc @cloud-fan can you please take a look?

cloud-fan:
retest this please

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/basicLogicalOperators.scala
```diff
@@ -340,14 +340,22 @@ case class Join(
...
-      super.computeStats(conf).copy(isBroadcastable = false)
+  override def computeStats(conf: CatalystConf): Statistics = {
+    def simpleEstimation: Statistics = joinType match {
+      case LeftAnti | LeftSemi =>
+        // LeftSemi and LeftAnti won't ever be bigger than left
+        left.stats(conf)
+      case _ =>
+        // Make sure we don't propagate isBroadcastable in other joins, because
+        // they could explode the size.
+        super.computeStats(conf).copy(isBroadcastable = false)
```
cloud-fan: not related to this PR, but `super.computeStats(conf)` won't produce a stats with `isBroadcastable = true`

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/statsEstimation/JoinEstimation.scala
```diff
@@ -0,0 +1,314 @@
...
+      val outputRows = joinType match {
+        case LeftOuter =>
+          // All rows from left side should be in the result.
+          leftRows.max(innerRows)
+        case RightOuter =>
+          // All rows from right side should be in the result.
+          rightRows.max(innerRows)
+        case FullOuter =>
+          // Simulate full outer join as obtaining the number of elements in the union of two
+          // finite sets: A \cup B = A + B - A \cap B => A FOJ B = A + B - A IJ B.
```
cloud-fan: what does `\cup` means?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/statsEstimation/JoinEstimation.scala
```diff
@@ -0,0 +1,314 @@
...
+      outputRows: BigInt,
+      attributes: Seq[Attribute],
+      oldAttrStats: AttributeMap[ColumnStat],
+      joinKeyStats: AttributeMap[ColumnStat]): AttributeMap[ColumnStat] = {
+    val outputAttrStats = new mutable.HashMap[Attribute, ColumnStat]()
+    val leftRows = leftStats.rowCount.get
+    val rightRows = rightStats.rowCount.get
+    if (outputRows == 0) {
+      // empty output
+      attributes.foreach(a => outputAttrStats.put(a, emptyColumnStat(a.dataType)))
```
cloud-fan: what's the difference between no column stats and empty column stats?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/statsEstimation/JoinEstimation.scala
```diff
@@ -0,0 +1,314 @@
...
+    } else if (outputRows == leftRows * rightRows) {
+      // We do estimation like cartesian product and propagate the original column stats
+      attributes.foreach(a => outputAttrStats.put(a, oldAttrStats(a)))
```
cloud-fan: what's the difference between no column stats and empty column stats? it `rowCount == 0`, we won't look at column stats at all, isn't it?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/statsEstimation/JoinEstimation.scala
cloud-fan: what's the difference between no column stats and empty column stats? it `rowCount == 0`, we won't look at column stats at all, isn't it?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/statsEstimation/JoinEstimation.scala
```diff
@@ -0,0 +1,314 @@
...
+          // Don't update column stats for attributes from left side.
+          fromLeft.map(a => (a, inputAttrStats(a))).toMap ++
+            updateAttrStats(outputRows, fromRight, inputAttrStats, intersectedStats)
+        case RightOuter =>
+          // Don't update column stats for attributes from right side.
+          updateAttrStats(outputRows, fromLeft, inputAttrStats, intersectedStats) ++
+            fromRight.map(a => (a, inputAttrStats(a))).toMap
+        case FullOuter =>
+          // Don't update column stats for attributes from both sides.
+          attributesWithStat.map(a => (a, inputAttrStats(a))).toMap
```
cloud-fan: the `nullCount` will be inaccurate after this, right?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/statsEstimation/JoinEstimation.scala
```diff
@@ -0,0 +1,314 @@
...
+      1
+    } else if (ndvDenom == 0) {
+      // One of the join key pairs is disjoint, thus the two sides of join is disjoint.
+      0
+    } else {
+      1 / BigDecimal(ndvDenom)
+    }
+  }
+
+  /** Update column stats for output attributes. */
```
cloud-fan: can you explain the heuristic used in this method in the doc?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/statsEstimation/Range.scala
```diff
@@ -0,0 +1,117 @@
...
+    case (_, _: DefaultRange) | (_: DefaultRange, _) =>
+      // Skip overlapping check for binary/string types
+      true
+    case (_, _: NullRange) | (_: NullRange, _) =>
+      false
+    case (n1: NumericRange, n2: NumericRange) =>
+      n1.min.compareTo(n2.max) <= 0 && n1.max.compareTo(n2.min) >= 0
+  }
+
+  /** This is only for two overlapped ranges. */
```
cloud-fan: please explain the 4 outputs in the doc

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/basicLogicalOperators.scala
wzhfy: We can't make sure we won't change the behavior of super.computeStats(conf), so I think it's safer to do this here.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/statsEstimation/JoinEstimation.scala
wzhfy: It's a set operation. Please ignore this, because I'll update computing equation for FullOuter.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/statsEstimation/JoinEstimation.scala
wzhfy: Yes. It's difficult to get a reasonable nullCount estimation after outer joins. So in filter, we only use nullCount to estimate if the child is a table.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/statsEstimation/JoinEstimation.scala
wzhfy: no column stats means we don't collect stats, while empty column stats means we have the stats and we know after estimation the column has no satisfying data.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/statsEstimation/Range.scala
wzhfy: ok I'll improve the doc

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/statsEstimation/JoinEstimation.scala
cloud-fan: in the case of `rowCount == 0`, I think column stats becomes useless, like we don't need column stats for an empty table. Do I miss something?

wzhfy: we may union this plan with the other plan, then empty column stats is useful. Now we don't support union estimation because it's difficult to estimate distinct count after union.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/statsEstimation/Range.scala
```diff
@@ -0,0 +1,120 @@
...
+  def apply(min: Option[Any], max: Option[Any], dataType: DataType): Range = dataType match {
+    case StringType | BinaryType => new DefaultRange()
+    case _ if min.isEmpty || max.isEmpty => new NullRange()
+    case _ => toNumericRange(min.get, max.get, dataType)
+  }
+
+  def isIntersected(r1: Range, r2: Range): Boolean = (r1, r2) match {
+    case (_, _: DefaultRange) | (_: DefaultRange, _) =>
+      // Skip overlapping check for binary/string types
+      true
```
cloud-fan: shall we return false to skip overlapping check?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/statsEstimation/JoinEstimation.scala
```diff
@@ -0,0 +1,316 @@
...
+      newDistinctCount: BigInt,
+      newMin: Option[Any],
+      newMax: Option[Any]): ColumnStat = {
+    origin.copy(distinctCount = newDistinctCount, min = newMin, max = newMax, nullCount = 0)
+  }
+
+  private def extractJoinKeys(
+      leftKeys: Seq[Expression],
+      rightKeys: Seq[Expression]): Seq[(AttributeReference, AttributeReference)] = {
+    leftKeys.zip(rightKeys).flatMap {
```
cloud-fan: use `collect` instead of `flatMap`

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/statsEstimation/JoinEstimation.scala
```diff
@@ -0,0 +1,316 @@
...
+      Some(Statistics(
+        sizeInBytes = getOutputSize(join.output, outputRows, outputAttrStats),
+        rowCount = Some(outputRows),
+        attributeStats = outputAttrStats,
+        isBroadcastable = false))
+  }
+
+  // scalastyle:off
+  /**
+   * The number of rows of A inner join B on A.k1 = B.k1 is estimated by this basic formula:
```
cloud-fan: outer join will also hit this branch, right?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/statsEstimation/Range.scala
cloud-fan: let's update the comment to say: `The DefaultRange represents string/binary types which do not have max/min stats, we assume they are intersected to be conservative on estimation`

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/statsEstimation/JoinEstimation.scala
```diff
@@ -0,0 +1,316 @@
...
+
+    case ExtractEquiJoinKeys(joinType, leftKeys, rightKeys, condition, left, right) =>
+      // 1. Compute join selectivity
+      val joinKeyPairs = extractJoinKeys(leftKeys, rightKeys)
+      val selectivity = joinSelectivity(joinKeyPairs, leftStats, rightStats)
+
+      // 2. Estimate the number of output rows
+      val leftRows = leftStats.rowCount.get
+      val rightRows = rightStats.rowCount.get
+      val innerRows = ceil(BigDecimal(leftRows * rightRows) * selectivity)
```
cloud-fan: nit: `joinedRows`

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/statsEstimation/JoinEstimation.scala
```diff
@@ -0,0 +1,316 @@
...
+        case FullOuter =>
+          // T(A FOJ B) = T(A LOJ B) + T(A ROJ B) - T(A IJ B)
+          leftRows.max(innerRows) + rightRows.max(innerRows) - innerRows
+        case _ =>
+          // Don't change for inner or cross join
+          innerRows
+      }
+
+      // 3. Update statistics based on the output of join
+      val intersectedStats = if (selectivity == 0) {
```
cloud-fan: what's the difference between `selectivity == 0` and `outputRows == 0`? does it only matter for outer joins?

cloud-fan: for outer joins, if selectivity is 0, then the number of output rows is same as the number of left/right side rows. And the column stats should also be same as the left/right side columns, while the other side columns are all null.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/statsEstimation/JoinEstimation.scala
```diff
@@ -0,0 +1,316 @@
...
+      newMin: Option[Any],
+      newMax: Option[Any]): ColumnStat = {
+    origin.copy(distinctCount = newDistinctCount, min = newMin, max = newMax, nullCount = 0)
+  }
+
+  private def extractJoinKeys(
+      leftKeys: Seq[Expression],
+      rightKeys: Seq[Expression]): Seq[(AttributeReference, AttributeReference)] = {
+    leftKeys.zip(rightKeys).flatMap {
+      case (lk: AttributeReference, rk: AttributeReference) => Some((lk, rk))
```
cloud-fan: we can check column stats existence here, so that we don't need to do `columnStatsExist((leftStats, leftKey), (rightStats, rightKey))` again and again later.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/statsEstimation/JoinEstimation.scala
cloud-fan: let's name it `joinKeyStats`

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/statsEstimation/StatsEstimationTestBase.scala
```diff
@@ -21,14 +21,19 @@ import org.apache.spark.SparkFunSuite
...
+import org.apache.spark.sql.types.{IntegerType, StringType}
 
 
 class StatsEstimationTestBase extends SparkFunSuite {
 
   /** Enable stats estimation based on CBO. */
   protected val conf = SimpleCatalystConf(caseSensitiveAnalysis = true, cboEnabled = true)
 
+  def getColSize(attribute: Attribute, colStat: ColumnStat): Long = attribute.dataType match {
+    case StringType => colStat.avgLen + 8 + 4
```
cloud-fan: explain the `+ 8 + 4`?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/statsEstimation/Range.scala
```diff
@@ -0,0 +1,120 @@
...
+class DefaultRange extends Range
+
+/** This is for columns with only null values. */
+class NullRange extends Range
+
+object Range {
+  def apply(min: Option[Any], max: Option[Any], dataType: DataType): Range = dataType match {
+    case StringType | BinaryType => new DefaultRange()
+    case _ if min.isEmpty || max.isEmpty => new NullRange()
+    case _ => toNumericRange(min.get, max.get, dataType)
```
cloud-fan: This doesn't work for empty column stats you defined in https://github.com/apache/spark/pull/16228/files#diff-6387e7aaeb7d8e0cb1457b9d0fe5cd00R270

that's why I was worried about the empty stats, it breaks some assumptions, like numeric type stats must have max/min.

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/statsEstimation/JoinEstimationSuite.scala
```diff
@@ -0,0 +1,314 @@
...
+import org.apache.spark.sql.catalyst.plans._
+import org.apache.spark.sql.catalyst.plans.logical.{ColumnStat, Join, Project, Statistics}
+import org.apache.spark.sql.types.{DateType, TimestampType, _}
+
+
+class JoinEstimationSuite extends StatsEstimationTestBase {
+
+  /** Set up tables and its columns for testing */
+  private val columnInfo: AttributeMap[ColumnStat] = AttributeMap(Seq(
+    attr("key11") -> ColumnStat(distinctCount = 5, min = Some(1), max = Some(5), nullCount = 0,
```
cloud-fan: how about `key-1-5`, `key-5-9`, etc.? then we can know the key value range directly from the name.

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/statsEstimation/JoinEstimationSuite.scala
```diff
@@ -0,0 +1,314 @@
...
+      attributeStats = AttributeMap(Seq("key11", "key12", "key21", "key22").map(nameToColInfo)))
+    assert(join.stats(conf) == expectedStats)
+  }
+
+  test("disjoint inner join") {
+    // table1 (key11 int, key12 int): (1, 9), (2, 8), (3, 7), (4, 6), (5, 5)
+    // table2 (key21 int, key22 int): (1, 2), (2, 3), (2, 4)
+    // key12 and key22 are disjoint
+    val join = Join(table1, table2, Inner, Some(
+      And(EqualTo(nameToAttr("key11"), nameToAttr("key21")),
```
cloud-fan: actually we can just use `key12 = key22`, so that it's more different from the test `inner join with multiple equi-join keys`

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/statsEstimation/JoinEstimation.scala
wzhfy: This is only for estimate the inner joined part.

cloud-fan: look at the code path: `JoinEstimation.estimate` -> `InnerOuterEstimation.doEstimate` -> `joinSelectivity`, outer join will hit this branch.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/statsEstimation/JoinEstimation.scala
wzhfy: yea good point, thanks

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/statsEstimation/Range.scala
wzhfy: ok, let's remove empty column stats. When we know rowCount=0, we can derive the column stats is empty, we don't need to keep it. what do you think?

cloud-fan: SGTM

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/statsEstimation/JoinEstimation.scala
wzhfy: yes, I mean it computes the inner join part in the outer join.

wzhfy:
This pr is updated, please review @cloud-fan

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/statsEstimation/JoinEstimation.scala
```diff
@@ -0,0 +1,316 @@
...
+
+      val outputStats: Seq[(Attribute, ColumnStat)] = if (innerJoinedRows == 0) {
+        joinType match {
+          // For outer joins, if the inner join part is empty, the number of output rows is the
+          // same as that of the outer side. And column stats of join keys from the outer side
+          // keep unchanged, while column stats of join keys from the other side should be updated
+          // based on added null values.
+          case LeftOuter =>
+            fromLeft.map(a => (a, inputAttrStats(a))) ++
+              fromRight.map(a => (a, nullColumnStat(a.dataType, leftRows)))
```
cloud-fan: `fromRight` only includes the attributes that having column stats, but logically we can use null column stats for all the attributes from the right side.

We can do it later.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/statsEstimation/JoinEstimation.scala
```diff
@@ -0,0 +1,316 @@
...
+          case RightOuter =>
+            fromRight.map(a => (a, inputAttrStats(a))) ++
+              fromLeft.map(a => (a, nullColumnStat(a.dataType, rightRows)))
```
cloud-fan: same here.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/statsEstimation/JoinEstimation.scala
```diff
@@ -0,0 +1,316 @@
...
+    : AttributeMap[ColumnStat] = {
+
+    val intersectedStats = new mutable.HashMap[Attribute, ColumnStat]()
+    joinKeyPairs.foreach { case (leftKey, rightKey) =>
+      // Check if the two sides are disjoint
+      val leftKeyStats = leftStats.attributeStats(leftKey)
+      val rightKeyStats = rightStats.attributeStats(rightKey)
+      val lRange = Range(leftKeyStats.min, leftKeyStats.max, leftKey.dataType)
+      val rRange = Range(rightKeyStats.min, rightKeyStats.max, rightKey.dataType)
+      if (Range.isIntersected(lRange, rRange)) {
```
cloud-fan: is it better to write `assert(Range.isIntersected(lRange, rRange))` here? When we reach here, the join selectivity is not 0 and each join keys should be intersected.

cloud-fan: oh string/binary type will hit this.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/statsEstimation/JoinEstimation.scala
```diff
@@ -0,0 +1,316 @@
...
+        // Update intersected column stats
+        val minNdv = leftKeyStats.distinctCount.min(rightKeyStats.distinctCount)
+        val (newMin1, newMax1, newMin2, newMax2) =
+          Range.intersect(lRange, rRange, leftKey.dataType, rightKey.dataType)
+        intersectedStats.put(leftKey, intersectedColumnStat(leftKeyStats, minNdv,
```
cloud-fan: logically the join keys should have same column stats, we can write it more explicitly
```
assert(leftKey.dataType.sameType(rightKey.dataType))
val stats = ColumnStats(minNdv, newMin, newMax, nullCount = 0) // and some more logic to update the avg/max length.
intersectedStats.put(leftKey, stats)
intersectedStats.put(rightKey, stats)
```

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/statsEstimation/Range.scala
```diff
@@ -0,0 +1,121 @@
...
+
+  /**
+   * Intersected results of two ranges. This is only for two overlapped ranges.
+   * The outputs are the intersected min/max values of the two columns based on their data types.
+   */
+  def intersect(
+      r1: Range,
+      r2: Range,
+      dt1: DataType,
+      dt2: DataType): (Option[Any], Option[Any], Option[Any], Option[Any]) = {
```
cloud-fan: we can simplify this, we will only calculate intersection for same-type ranges.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/statsEstimation/JoinEstimation.scala
```diff
@@ -0,0 +1,316 @@
...
+        join.joinType match {
+          // For outer joins, don't update column stats from the outer side.
+          case LeftOuter =>
+            fromLeft.map(a => (a, inputAttrStats(a))) ++
+              updateAttrStats(outputRows, fromRight, inputAttrStats, joinKeyStats)
+          case RightOuter =>
+            updateAttrStats(outputRows, fromLeft, inputAttrStats, joinKeyStats) ++
+              fromRight.map(a => (a, inputAttrStats(a)))
+          case FullOuter =>
+            attributesWithStat.map(a => (a, inputAttrStats(a)))
```
cloud-fan: this is just `inputAttrStats` right?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/statsEstimation/JoinEstimation.scala
```diff
@@ -0,0 +1,316 @@
...
+      // One of the join key pairs is disjoint, thus the two sides of join is disjoint.
+      0
+    } else {
+      1 / BigDecimal(ndvDenom)
+    }
+  }
+
+  /**
+   * Propagate or update column stats for output attributes.
+   * 1. For empty output, we don't need to keep any column stats.
```
cloud-fan: when we hit this method, the `outputRows` will never be 0 right?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/statsEstimation/JoinEstimation.scala
```diff
@@ -0,0 +1,316 @@
...
+      val leftRatio =
+        if (leftRows != 0) BigDecimal(outputRows) / BigDecimal(leftRows) else BigDecimal(0)
+      val rightRatio =
+        if (rightRows != 0) BigDecimal(outputRows) / BigDecimal(rightRows) else BigDecimal(0)
+      attributes.foreach { a =>
+        // check if this attribute is a join key
+        if (joinKeyStats.contains(a)) {
+          outputAttrStats += a -> joinKeyStats(a)
+        } else {
+          val oldCS = oldAttrStats(a)
```
cloud-fan: nit: `oldColumnStats`

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/statsEstimation/JoinEstimation.scala
```diff
@@ -0,0 +1,316 @@
...
+        rowCount = Some(outputRows),
+        attributeStats = outputAttrStats,
+        isBroadcastable = false))
+
+    case _ =>
+      // When there is no equi-join condition, we do estimation like cartesian product.
+      val inputAttrStats = AttributeMap(
+        leftStats.attributeStats.toSeq ++ rightStats.attributeStats.toSeq)
+      // Propagate the original column stats
+      val outputAttrStats = getOutputMap(inputAttrStats, join.output)
```
cloud-fan: is it just `inputsAttrStats`?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/statsEstimation/JoinEstimation.scala
```diff
@@ -0,0 +1,316 @@
...
+
+case class LeftSemiAntiEstimation(conf: CatalystConf, join: Join) {
+  def doEstimate(): Option[Statistics] = {
+    // TODO: It's error-prone to estimate cardinalities for LeftSemi and LeftAnti based on basic
+    // column stats. Now we just propagate the statistics from left side. We should do more
+    // accurate estimation when advanced stats (e.g. histograms) are available.
+    if (rowCountsExist(conf, join.left)) {
+      val leftStats = join.left.stats(conf)
+      // Propagate the original column stats for cartesian product
+      val outputAttrStats = getOutputMap(leftStats.attributeStats, join.output)
```
cloud-fan: is it just `leftStats.attributeStats`?

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/statsEstimation/JoinEstimationSuite.scala
```diff
@@ -0,0 +1,356 @@
...
+      AttributeReference("cbinary", BinaryType)() -> ColumnStat(distinctCount = 1,
+        min = None, max = None, nullCount = 0, avgLen = 3, maxLen = 3),
+      AttributeReference("cdate", DateType)() -> ColumnStat(distinctCount = 1,
+        min = Some(d1), max = Some(d1), nullCount = 0, avgLen = 4, maxLen = 4),
+      AttributeReference("ctimestamp", TimestampType)() -> ColumnStat(distinctCount = 1,
+        min = Some(t1), max = Some(t1), nullCount = 0, avgLen = 8, maxLen = 8)
+    )
+
+    /** Columns in a table with two rows */
+    val columnInfo2 = mutable.LinkedHashMap[Attribute, ColumnStat](
```
cloud-fan: are these totally same with the `columnInfo1`? may we can create a method to do this

cloud-fan:
LGTM except some minor comments, thanks for working on it!

cloud-fan:
thanks, merging to master!

PR #16232, opened 2016-12-09
[SPARK-18800][SQL] Correct the assert in UnsafeKVExternalSorter which ensures array size

viirya:

`UnsafeKVExternalSorter` uses `UnsafeInMemorySorter` to sort the records of `BytesToBytesMap` if it is given a map.

Currently we use the number of keys in `BytesToBytesMap` to determine if the array used for sort is enough or not. We has an assert that ensures the size of the array is enough: `map.numKeys() <= map.getArray().size() / 2`.

However, each record in the map takes two entries in the array, one is record pointer, another is key prefix. So the correct assert should be `map.numKeys() * 2 <= map.getArray().size() / 2`.

## How was this patch tested?

N/A

### DIFF
--- a/sql/core/src/main/java/org/apache/spark/sql/execution/UnsafeKVExternalSorter.java
+++ b/sql/core/src/main/java/org/apache/spark/sql/execution/UnsafeKVExternalSorter.java
@@ -24,7 +24,6 @@ import com.google.common.annotations.VisibleForTesting;
 
 import org.apache.spark.SparkEnv;
 import org.apache.spark.TaskContext;
-import org.apache.spark.memory.MemoryConsumer;
 import org.apache.spark.memory.TaskMemoryManager;
 import org.apache.spark.serializer.SerializerManager;
 import org.apache.spark.sql.catalyst.expressions.UnsafeRow;
@@ -34,7 +33,6 @@ import org.apache.spark.sql.types.StructType;
 import org.apache.spark.storage.BlockManager;
 import org.apache.spark.unsafe.KVIterator;
 import org.apache.spark.unsafe.Platform;
-import org.apache.spark.unsafe.array.LongArray;
 import org.apache.spark.unsafe.map.BytesToBytesMap;
 import org.apache.spark.unsafe.memory.MemoryBlock;
 import org.apache.spark.util.collection.unsafe.sort.*;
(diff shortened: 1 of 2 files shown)


viirya:
retest this please.

viirya:
The Failed tests are passed locally. Let's test it again.

viirya:
Interesting...those failed tests are all passed locally. I will dig into it.

hvanhovell:
retest this please

viirya:
OK. It is passed now.

sql/core/src/main/java/org/apache/spark/sql/execution/UnsafeKVExternalSorter.java
```diff
@@ -96,13 +98,35 @@ public UnsafeKVExternalSorter(
...
-      assert(map.numKeys() <= map.getArray().size() / 2);
+      // Becasue we insert the number of values in the map into `UnsafeInMemorySorter`, if
+      // the number of values is more than the number of keys, and the array in the map is
+      // not big enough to do in-place sort, we must acquire new array.
+      // To insert a record into `UnsafeInMemorySorter` will consume two spaces in the array.
+      // We must have half of the array as empty. There are totally `map.numValues()` records
+      // to be inserted.
+      LongArray sortArray = null;
+      boolean useAllocatedArray = false;
+      if (map.numValues() > map.numKeys() && map.numValues() * 2 > map.getArray().size() / 2) {
```
hvanhovell: I mean the multiplication and division by 2?

sql/core/src/main/java/org/apache/spark/sql/execution/UnsafeKVExternalSorter.java
```diff
@@ -96,13 +98,35 @@ public UnsafeKVExternalSorter(
...
+        }
+      }
+
       // During spilling, the array in map will not be used, so we can borrow that and use it
       // as the underlying array for in-memory sorter (it's always large enough).
-      // Since we will not grow the array, it's fine to pass `null` as consumer.
+      // Although we will not grow the array, it's fine to pass `null` as consumer, however,
+      // if we have allocated array instead of using the map's array, we still need
+      // to pass the map as consumer in order to release the allocated array.
+      MemoryConsumer consumer = useAllocatedArray ? map : null;
```
hvanhovell: Just assign the consumer in the previous block. That makes it clearer what you are doing.

sql/core/src/main/java/org/apache/spark/sql/execution/UnsafeKVExternalSorter.java
viirya: oh. I added the comment to explain the correct number. So I keep the multiplication and division to make it clear and match the explanation. Do you prefer to simplify it?

sql/core/src/main/java/org/apache/spark/sql/execution/UnsafeKVExternalSorter.java
viirya: Ok.

sql/core/src/main/java/org/apache/spark/sql/execution/UnsafeKVExternalSorter.java
viirya: If you ask the reason behind this calculation, it is because the a value needs two spaces (pointer, key's hashcode) in the array. We need at least half of the array for sorting. So if the occupied space of all values are more than half of the array, we need to allocate new bigger array for sorting.

viirya:
ping @hvanhovell Do you think this fixing good now? Thanks.

cc @davies @JoshRosen

davies:
@viirya A map with multiple values for same key, is only used for hashed relation, not aggregation, will also not passed into UnsafeKVExternalSorter. So I think this is not actually a bug.

viirya:
@davies Yes, that is an extreme case. I use it to explain how we can cause the exception.

Even we don't consider the extreme case if we don't pass the kind of map in, the original assert `map.numKeys() <= map.getArray().size() / 2` looks problematic. If each key has different value, as a key/value pair occupied two spaces in the array, I think the assert for the array space should be `map.numKeys() * 2 <= map.getArray().size() / 2`.

davies:
@viirya  That's not correct, the values does not have entry in the array.

viirya:
@davies What I just said is not accurate. I don't meant the values have entry in the array. I meant each key/value pair will occupy two entries in the array.

We iterator all key/value pairs in the `BytesToBytesMap` and call `UnsafeInMemorySorter.insertRecord` which inserts record pointer and key prefix for each key/value pair in https://github.com/apache/spark/blob/5857b9ac2d9808d9b89a5b29620b5052e2beebf5/core/src/main/java/org/apache/spark/util/collection/unsafe/sort/UnsafeInMemorySorter.java#L241

So you will have `map.numKeys() * 2` entries in the array because for each you will have one entry for record pointer and one entry for key prefix.

davies:
That make sense, we should update the assert. But this still is not a bug, the other changes are not needed.

viirya:
OK. I will update accordingly.

viirya:
@davies Actually this pr is motivated by a reported error on dev mailling list at http://apache-spark-developers-list.1001551.n3.nabble.com/java-lang-IllegalStateException-There-is-no-space-for-new-record-tc20108.html

So if the array size is not enough, don't we need to allocate big enough array for the sorter like the current change?

The reporter doesn't have the repro, but I think this place is the only one which will cause this error.

davies:
@viirya without a repro, I don't think this is the root cause. There could be a random corrupt that cause the error.

viirya:
@davies Ok. I got it. I will update the assert.

davies:
lgtm

viirya:
ping @davies

srowen:
Looks like you are cleared to merge @viirya if you're OK with it

viirya:
I am ok.

srowen:
Merged to master

viirya:
Thanks @davies  @srowen

PR #16235, opened 2016-12-09
[SPARK-18745][SQL] Fix signed integer overflow due to toInt cast

kiszk:

This PR avoids that a result of a cast `toInt` is negative due to signed integer overflow (e.g. 0x0000_0000_1???????L.toInt < 0 ). This PR performs casts after we can ensure the value is within range of signed integer (the result of `max(array.length, ???)` is always integer).

## How was this patch tested?

Manually executed query68 of TPC-DS with 100TB

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/joins/HashedRelation.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/joins/HashedRelation.scala
@@ -670,9 +670,9 @@ private[execution] final class LongToUnsafeRowMap(val mm: TaskMemoryManager, cap
     var offset: Long = Platform.LONG_ARRAY_OFFSET
     val end = len * 8L + Platform.LONG_ARRAY_OFFSET
     while (offset < end) {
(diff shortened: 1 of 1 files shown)


hvanhovell:
LGTM - pending jenkins.

@kiszk do you think there is a sane way of testing this?

hvanhovell:
also cc @davies

hvanhovell:
retest this please

davies:
LGTM

kiszk:
@hvanhovell if we want to test these methods, it seems to be complicated since we have to allocate large `array` by a naive approach that allocates `LongToUnsafeRowMap` and expands `array`.
Another approach is to directly call private methods `writeLongArray` and `readLongArray` by using reflection.

hvanhovell:
Ok, lets not do that for now.

hvanhovell:
It might at some point be an idea to add a bunch of asserts to `Platform`, that would make diagnoses a lot easier.

kiszk:
Jenkins, retest this please

srowen (approved):

hvanhovell:

PR #16237, opened 2016-12-09
[SPARK-18807][SPARKR] Should suppress output print for calls to JVM methods with void return values

felixcheung:

Several SparkR API calling into JVM methods that have void return values are getting printed out, especially when running in a REPL or IDE.
example:
```
> setLogLevel("WARN")
NULL
```
We should fix this to make the result more clear.

Also found a small change to return value of dropTempView in 2.1 - adding doc and test for it.

## How was this patch tested?

manually - I didn't find a expect_*() method in testthat for this

### DIFF
--- a/R/pkg/inst/tests/testthat/test_sparkSQL.R
+++ b/R/pkg/inst/tests/testthat/test_sparkSQL.R
@@ -589,7 +589,7 @@ test_that(
   newdf <- sql("SELECT * FROM table1 where name = 'Michael'")
   expect_is(newdf, "SparkDataFrame")
   expect_equal(count(newdf), 1)
-  dropTempView("table1")
+  expect_true(dropTempView("table1"))
 
   createOrReplaceTempView(df, "dfView")
   sqlCast <- collect(sql("select cast('2' as decimal) as x from dfView limit 1"))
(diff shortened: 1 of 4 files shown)


shivaram:
Does `expert_silent` do what we want here ? https://github.com/hadley/testthat/issues/261

shivaram:
Actually looks like that doesn't do what we want - that checks for outputs on stdout, warnings or errors etc.

R/pkg/inst/tests/testthat/test_sparkSQL.R
```diff
@@ -576,7 +576,7 @@ test_that("test tableNames and tables", {
   tables <- tables()
   expect_equal(count(tables), 2)
   suppressWarnings(dropTempTable("table1"))
-  dropTempView("table2")
+  expect_true(dropTempView("table2"))
```
shivaram: We have a bunch of other places where we call `dropTempView` - Might be good to add this to those places which we think are testing temp views ?

shivaram:
LGTM. Will wait for Jenkins, AppVeyor

shivaram:
Merging into master, branch-2.1

PR #16238, opened 2016-12-09
[SPARK-18811] StreamSource resolution should happen in stream execution thread

brkyvz:

When you start a stream, if we are trying to resolve the source of the stream, for example if we need to resolve partition columns, this could take a long time. This long execution time should not block the main thread where `query.start()` was called on. It should happen in the stream execution thread possibly before starting any triggers.

## How was this patch tested?

Unit test added. Made sure test fails with no code changes.

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamExecution.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamExecution.scala
@@ -47,7 +47,7 @@ class StreamExecution(
     override val sparkSession: SparkSession,
     override val name: String,
     checkpointRoot: String,
-    val logicalPlan: LogicalPlan,
+    analyzedPlan: LogicalPlan,
     val sink: Sink,
     val trigger: Trigger,
     val triggerClock: Clock,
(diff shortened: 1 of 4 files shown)


brkyvz:
cc @marmbrus @zsxwing @tdas

sql/core/src/test/scala/org/apache/spark/sql/streaming/util/DefaultSource.scala
```diff
@@ -0,0 +1,66 @@
...
+package org.apache.spark.sql.streaming.util
+
+import org.apache.spark.sql.{SQLContext, _}
+import org.apache.spark.sql.execution.streaming.{LongOffset, Offset, Sink, Source}
+import org.apache.spark.sql.sources.{StreamSinkProvider, StreamSourceProvider}
+import org.apache.spark.sql.streaming.{OutputMode, StreamingQueryManagerSuite}
+import org.apache.spark.sql.types.{IntegerType, StructField, StructType}
+
+/** Dummy provider: returns a SourceProvider with a blocking `createSource` call. */
+class DefaultSource extends StreamSourceProvider with StreamSinkProvider {
```
marmbrus: Nit, I'd consider calling this `BlockingSource` or something more descriptive of what makes it special.

brkyvz: I wanted to call it that, but doesn't the magic require it to be called `DefaultSource`?

marmbrus:
This LGTM, I was just talking with @tdas about how I think that all of this initialization stuff should be lazy and happen on the stream execution thread.  I think this can simplify what they are trying to fix in #16220 as well.  /cc @zsxwing

sql/core/src/test/scala/org/apache/spark/sql/streaming/util/DefaultSource.scala
marmbrus: No, you can give fully qualified classnames as well.

brkyvz:
retest this please

brkyvz:
failed on `StateStoreSuite - maintenance`, which is unrelated

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamExecution.scala
```diff
@@ -214,6 +228,10 @@ class StreamExecution(
       // While active, repeatedly attempt to run batches.
       SparkSession.setActiveSession(sparkSession)
 
+      updateStatusMessage("Initializing sources")
```
zsxwing: Could you move `SparkSession.setActiveSession(sparkSession)`, `updateStatusMessage("Initializing sources")` and `logicalPlan` above `postEvent(new QueryStartedEvent(id, runId, name))`? Then we can initialize the logical plan before returning `StreamingQuery` to the user and ensure `logicalPlan` is always initialized in the stream thread. And you can also add `assert(Thread.currentThread().isInstanceOf[StreamExecutionThread])` when creating `logicalPlan`, and remove the unnecessary test you added since this assertion will be tested in all current stream tests.

zsxwing: Okey. I'm going to merge this PR since it passed tests. I will fix this in my PR.

zsxwing:
Merging to master and 2.1.

PR #16240, opened 2016-12-10
[SPARK-16792][SQL] Dataset containing a Case Class with a List type causes a CompileException (converting sequence to list)

michalsenkyr:

Added a `to` call at the end of the code generated by `ScalaReflection.deserializerFor` if the requested type is not a supertype of `WrappedArray[_]` that uses `CanBuildFrom[_, _, _]` to convert result into an arbitrary subtype of `Seq[_]`.

Care was taken to preserve the original deserialization where it is possible to avoid the overhead of conversion in cases where it is not needed

`ScalaReflection.serializerFor` could already be used to serialize any `Seq[_]` so it was not altered

`SQLImplicits` had to be altered and new implicit encoders added to permit serialization of other sequence types

Also fixes [SPARK-16815] Dataset[List[T]] leads to ArrayStoreException

## How was this patch tested?
```bash
./build/mvn -DskipTests clean package && ./dev/run-tests
```

Also manual execution of the following sets of commands in the Spark shell:
```scala
case class TestCC(key: Int, letters: List[String])

val ds1 = sc.makeRDD(Seq(
(List("D")),
(List("S","H")),
(List("F","H")),
(List("D","L","L"))
)).map(x=>(x.length,x)).toDF("key","letters").as[TestCC]

val test1=ds1.map{_.key}
test1.show
```

```scala
case class X(l: List[String])
spark.createDataset(Seq(List("A"))).map(X).show
```

```scala
spark.sqlContext.createDataset(sc.parallelize(List(1) :: Nil)).collect
```

After adding arbitrary sequence support also tested with the following commands:

```scala
case class QueueClass(q: scala.collection.immutable.Queue[Int])

spark.createDataset(Seq(List(1,2,3))).map(x => QueueClass(scala.collection.immutable.Queue(x: _*))).map(_.q.dequeue).collect
```

### DIFF
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/ScalaReflection.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/ScalaReflection.scala
@@ -312,11 +312,20 @@ object ScalaReflection extends ScalaReflection {
           "array",
           ObjectType(classOf[Array[Any]]))
 
-        StaticInvoke(
+        val wrappedArray = StaticInvoke(
           scala.collection.mutable.WrappedArray.getClass,
           ObjectType(classOf[Seq[_]]),
           "make",
           array :: Nil)
+
+        if (t <:< localTypeOf[List[_]]) {
+          Invoke(
+            wrappedArray,
+            "toList",
+            ObjectType(classOf[List[_]]))
+        } else {
+          wrappedArray
+        }
 
       case t if t <:< localTypeOf[Map[_, _]] =>
         // TODO: add walked type path for map
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/ScalaReflection.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/ScalaReflection.scala
@@ -318,13 +318,38 @@ object ScalaReflection extends ScalaReflection {
           "make",
           array :: Nil)
 
-        if (t <:< localTypeOf[List[_]]) {
+        if (localTypeOf[scala.collection.mutable.WrappedArray[_]] <:< t.erasure) {
+          wrappedArray
+        } else {
+          // Convert to another type using `to`
+          val cls = mirror.runtimeClass(t.typeSymbol.asClass)
+          import scala.collection.generic.CanBuildFrom
+          import scala.reflect.ClassTag
+          import scala.util.{Try, Success}
           Invoke(
             wrappedArray,
-            "toList",
-            ObjectType(classOf[List[_]]))
-        } else {
-          wrappedArray
+            "to",
+            ObjectType(cls),
+            StaticInvoke(
+              cls,
+              ObjectType(classOf[CanBuildFrom[_, _, _]]),
+              "canBuildFrom",
+              Try(cls.getDeclaredMethod("canBuildFrom", classOf[ClassTag[_]])) match {
+                case Success(_) =>
+                  StaticInvoke(
+                    ClassTag.getClass,
+                    ObjectType(classOf[ClassTag[_]]),
+                    "apply",
+                    StaticInvoke(
+                      cls,
+                      ObjectType(classOf[Class[_]]),
+                      "getClass"
+                    ) :: Nil
+                  ) :: Nil
+                case _ => Nil
+              }
+            ) :: Nil
+          )
         }
 
       case t if t <:< localTypeOf[Map[_, _]] =>
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/ScalaReflection.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/ScalaReflection.scala
@@ -325,7 +325,24 @@ object ScalaReflection extends ScalaReflection {
           val cls = mirror.runtimeClass(t.typeSymbol.asClass)
           import scala.collection.generic.CanBuildFrom
           import scala.reflect.ClassTag
-          import scala.util.{Try, Success}
+
+          // Some canBuildFrom methods take an implicit ClassTag parameter
+          val cbfParams = try {
+            cls.getDeclaredMethod("canBuildFrom", classOf[ClassTag[_]])
+            StaticInvoke(
+              ClassTag.getClass,
+              ObjectType(classOf[ClassTag[_]]),
+              "apply",
+              StaticInvoke(
+                cls,
+                ObjectType(classOf[Class[_]]),
+                "getClass"
+              ) :: Nil
+            ) :: Nil
+          } catch {
+            case _: NoSuchMethodException => Nil
+          }
+
           Invoke(
             wrappedArray,
             "to",
@@ -334,20 +351,7 @@ object ScalaReflection extends ScalaReflection {
               cls,
               ObjectType(classOf[CanBuildFrom[_, _, _]]),
               "canBuildFrom",
-              Try(cls.getDeclaredMethod("canBuildFrom", classOf[ClassTag[_]])) match {
-                case Success(_) =>
-                  StaticInvoke(
-                    ClassTag.getClass,
-                    ObjectType(classOf[ClassTag[_]]),
-                    "apply",
-                    StaticInvoke(
-                      cls,
-                      ObjectType(classOf[Class[_]]),
-                      "getClass"
-                    ) :: Nil
-                  ) :: Nil
-                case _ => Nil
-              }
+              cbfParams
             ) :: Nil
           )
         }
--- a/sql/core/src/main/scala/org/apache/spark/sql/SQLImplicits.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/SQLImplicits.scala
@@ -100,31 +100,72 @@ abstract class SQLImplicits {
   // Seqs
 
   /** @since 1.6.1 */
-  implicit def newIntSeqEncoder: Encoder[Seq[Int]] = ExpressionEncoder()
+  implicit def newIntSeqEncoder[T <: Seq[Int] : TypeTag]: Encoder[T] = ExpressionEncoder()
 
   /** @since 1.6.1 */
-  implicit def newLongSeqEncoder: Encoder[Seq[Long]] = ExpressionEncoder()
+  implicit def newLongSeqEncoder[T <: Seq[Long] : TypeTag]: Encoder[T] = ExpressionEncoder()
 
   /** @since 1.6.1 */
-  implicit def newDoubleSeqEncoder: Encoder[Seq[Double]] = ExpressionEncoder()
+  implicit def newDoubleSeqEncoder[T <: Seq[Double] : TypeTag]: Encoder[T] = ExpressionEncoder()
 
   /** @since 1.6.1 */
-  implicit def newFloatSeqEncoder: Encoder[Seq[Float]] = ExpressionEncoder()
+  implicit def newFloatSeqEncoder[T <: Seq[Float] : TypeTag]: Encoder[T] = ExpressionEncoder()
 
   /** @since 1.6.1 */
-  implicit def newByteSeqEncoder: Encoder[Seq[Byte]] = ExpressionEncoder()
+  implicit def newByteSeqEncoder[T <: Seq[Byte] : TypeTag]: Encoder[T] = ExpressionEncoder()
 
   /** @since 1.6.1 */
-  implicit def newShortSeqEncoder: Encoder[Seq[Short]] = ExpressionEncoder()
+  implicit def newShortSeqEncoder[T <: Seq[Short] : TypeTag]: Encoder[T] = ExpressionEncoder()
 
   /** @since 1.6.1 */
-  implicit def newBooleanSeqEncoder: Encoder[Seq[Boolean]] = ExpressionEncoder()
+  implicit def newBooleanSeqEncoder[T <: Seq[Boolean] : TypeTag]: Encoder[T] = ExpressionEncoder()
 
   /** @since 1.6.1 */
-  implicit def newStringSeqEncoder: Encoder[Seq[String]] = ExpressionEncoder()
+  implicit def newStringSeqEncoder[T <: Seq[String] : TypeTag]: Encoder[T] = ExpressionEncoder()
 
   /** @since 1.6.1 */
-  implicit def newProductSeqEncoder[A <: Product : TypeTag]: Encoder[Seq[A]] = ExpressionEncoder()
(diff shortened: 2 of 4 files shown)


michalsenkyr:
I would like to add that the conversion is specific to `List[_]`. I can add support for arbitrary sequence types through the use of `CanBuildFrom` if it is desirable.

We can also support `Set`s in this way by serializing into arrays. See SPARK-17414

michalsenkyr:
Added support for arbitrary sequences.

Now also Queues, ArrayBuffers and such can be used in datasets (all are serialized into ArrayType).

I had to alter and add new implicit encoders into `SQLImplicits`. The new encoders are for `Seq` with `Product` combination (essentially only `List`) to disambiguate between `Seq` and `Product` encoders.

However, I encountered a problem with implicits. When constructing a complex Dataset using `Seq.toDS` that includes a `Product` (like a case class) and a sequence, the encoder doesn't seem to be created. When constructed with `spark.createDataset` or when transforming an existing dataset, there is no problem.

I added a workaround by defining a specific implicit just for `Seq`s. This makes the problem go away for existing usages, however other collections cannot be constructed by `Seq.toDS` unless `newProductSeqEncoder[A, T]` is created with the correct type parameters.

If anybody knows how to fix this, let me know.

michalsenkyr:
Possible optimization: Instead of conversions using `to`, we can use `Builder`s. This way we could get rid of the conversion overhead. This would require adding a new codegen method that would operate similarly to `MapObjects` but use a provided `Builder` to build the collection directly.

I will wait for a response to this PR before attempting any more modifications.

sql/core/src/main/scala/org/apache/spark/sql/SQLImplicits.scala
```diff
@@ -100,31 +100,76 @@ abstract class SQLImplicits {
...
+  /** @since 2.2.0 */
+  implicit def newIntSeqWithProductEncoder[T <: Seq[Int] with Product : TypeTag]: Encoder[T] =
+    newIntSeqEncoder
+
+  /** @since 2.2.0 */
+  implicit def newLongSeqWithProductEncoder[T <: Seq[Long] with Product : TypeTag]: Encoder[T] =
+    newLongSeqEncoder
+
+  /** @since 2.2.0 */
+  implicit def newDoubleListEncoder[T <: Seq[Double] with Product : TypeTag]: Encoder[T] =
```
aray: Should this be `newDoubleSeqWithProductEncoder`?

cloud-fan:
The overall strategy LGTM.

> I had to alter and add new implicit encoders into SQLImplicits. The new encoders are for Seq with Product combination (essentially only List) to disambiguate between Seq and Product encoders.

Does scala have a clear definition for this case? i.e. we have implicit for both type `A` and `B`, given type `A with B`, which implicit will be picked?

For the optimization, we can do it in follow-up.

michalsenkyr:
None of them. The compilation will fail. That is why I had to provide those additional implicits.

```
scala> class Test[T]
defined class Test

scala> implicit def test1[T <: Seq[String]]: Test[T] = null
test1: [T <: Seq[String]]=> Test[T]

scala> implicit def test2[T <: Product]: Test[T] = null
test2: [T <: Product]=> Test[T]

scala> def test[T : Test](t: T) = null
test: [T](t: T)(implicit evidence$1: Test[T])Null

scala> test(List("abc"))
<console>:31: error: ambiguous implicit values:
 both method test1 of type [T <: Seq[String]]=> Test[T]
 and method test2 of type [T <: Product]=> Test[T]
 match expected type Test[List[String]]
       test(List("abc"))
```

cloud-fan:
How about we assign priority to implicit rules like http://stackoverflow.com/questions/1886953/is-there-a-way-to-control-which-implicit-conversion-will-be-the-default-used ?

I think we should prefer `Seq` encoder over `Product` encoder, for `Seq with Product`

michalsenkyr:
I actually read that but IDEA complained when I tried to place the `Product` encoder into a separate trait. So I opted for specificity.
However, I tried it again right now and even though IDEA still complains, scalac compiles it, all the tests pass and it all works. So I will change it.
I will also test whether this fixes the `toDS`-specific problems before pushing the changes.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/ScalaReflection.scala
```diff
@@ -312,12 +312,46 @@ object ScalaReflection extends ScalaReflection {
...
           array :: Nil)
 
+        if (localTypeOf[scala.collection.mutable.WrappedArray[_]] <:< t.erasure) {
+          wrappedArray
+        } else {
+          // Convert to another type using `to`
+          val cls = mirror.runtimeClass(t.typeSymbol.asClass)
+          import scala.collection.generic.CanBuildFrom
+          import scala.reflect.ClassTag
+          import scala.util.{Try, Success}
```
cloud-fan: spark code style discourage the usage of `Try` and `Success`, can you refactor your code a little bit? i.e. move `cls.getDeclaredMethod("canBuildFrom", classOf[ClassTag[_]])` out of the `Invoke` code block.

sql/core/src/main/scala/org/apache/spark/sql/SQLImplicits.scala
```diff
@@ -100,31 +97,36 @@ abstract class SQLImplicits {
...
-  implicit def newBooleanSeqEncoder: Encoder[Seq[Boolean]] = ExpressionEncoder()
+  implicit def newBooleanSeqEncoder[T <: Seq[Boolean] : TypeTag]: Encoder[T] = ExpressionEncoder()
 
   /** @since 1.6.1 */
-  implicit def newStringSeqEncoder: Encoder[Seq[String]] = ExpressionEncoder()
+  implicit def newStringSeqEncoder[T <: Seq[String] : TypeTag]: Encoder[T] = ExpressionEncoder()
 
   /** @since 1.6.1 */
-  implicit def newProductSeqEncoder[A <: Product : TypeTag]: Encoder[Seq[A]] = ExpressionEncoder()
+  implicit def newProductSeqEncoder[A <: Product : TypeTag, T <: Seq[A] : TypeTag]: Encoder[T] =
```
cloud-fan: This is my only concern now. Can you provide more details about it?

michalsenkyr: This one is the same as all the other ones, just with Product subclasses. If you were concerned about the `TypeTag` on `A`, it was actually not needed as `T`'s tag already contains all the information. I just tested it to be sure and removed it.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/ScalaReflection.scala
michalsenkyr: Done. I tried looking up the code style you mentioned, but only found the [Databricks' Scala Code Style Guide](https://github.com/databricks/scala-style-guide#exception-handling-try-vs-try). And that is not mentioned in the Spark docs as far as I know.

sql/core/src/main/scala/org/apache/spark/sql/SQLImplicits.scala
```diff
@@ -100,31 +97,36 @@ abstract class SQLImplicits {
...
-  implicit def newBooleanSeqEncoder: Encoder[Seq[Boolean]] = ExpressionEncoder()
+  implicit def newBooleanSeqEncoder[T <: Seq[Boolean] : TypeTag]: Encoder[T] = ExpressionEncoder()
 
   /** @since 1.6.1 */
-  implicit def newStringSeqEncoder: Encoder[Seq[String]] = ExpressionEncoder()
+  implicit def newStringSeqEncoder[T <: Seq[String] : TypeTag]: Encoder[T] = ExpressionEncoder()
 
   /** @since 1.6.1 */
-  implicit def newProductSeqEncoder[A <: Product : TypeTag]: Encoder[Seq[A]] = ExpressionEncoder()
+  implicit def newProductSeqEncoder[A <: Product, T <: Seq[A] : TypeTag]: Encoder[T] =
```
cloud-fan: can we just use `newProductSeqEncoder[T <: Seq[Product] : TypeTag]: Encoder[T]` here? Then we don't need to workaround implicit

michalsenkyr: You are right. `Seq` is declared covariant on `T` so it works and solves all the problems I was having. Thanks

sql/core/src/test/scala/org/apache/spark/sql/DatasetPrimitiveSuite.scala
```diff
@@ -130,6 +130,30 @@ class DatasetPrimitiveSuite extends QueryTest with SharedSQLContext {
     checkDataset(Seq(Array(Tuple1(1))).toDS(), Array(Tuple1(1)))
   }
 
+  test("arbitrary sequences") {
```
cloud-fan: let's also test nested sequences, e.g. `List(Queue(1))`, and sequences inside product, e.g. `List(1) -> Queue(1)`

michalsenkyr: I added some sequence-product combination tests.
Nested sequences were never supported (tried on master and 2.0.2).

cloud-fan:
retest this please

sql/core/src/test/scala/org/apache/spark/sql/DatasetPrimitiveSuite.scala
```diff
@@ -17,10 +17,21 @@
...
 package org.apache.spark.sql
 
+import scala.collection.immutable.Queue
+import scala.collection.mutable.ArrayBuffer
+
 import org.apache.spark.sql.test.SharedSQLContext
 
 case class IntClass(value: Int)
 
+case class SeqCC(s: Seq[Int])
```
cloud-fan: what does `CC` short for? How about `SeqClass`?

cloud-fan:
LGTM, please create 2 more tickets for the optimization you metioned in https://github.com/apache/spark/pull/16240#issuecomment-266318016 and the nested custom collection.

cloud-fan:
you need to fix mima:
```
[error]  * method newDoubleSeqEncoder()org.apache.spark.sql.Encoder in class org.apache.spark.sql.SQLImplicits does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("org.apache.spark.sql.SQLImplicits.newDoubleSeqEncoder")
[error]  * method newFloatSeqEncoder()org.apache.spark.sql.Encoder in class org.apache.spark.sql.SQLImplicits does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("org.apache.spark.sql.SQLImplicits.newFloatSeqEncoder")
[error]  * method newByteSeqEncoder()org.apache.spark.sql.Encoder in class org.apache.spark.sql.SQLImplicits does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("org.apache.spark.sql.SQLImplicits.newByteSeqEncoder")
[error]  * method newLongSeqEncoder()org.apache.spark.sql.Encoder in class org.apache.spark.sql.SQLImplicits does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("org.apache.spark.sql.SQLImplicits.newLongSeqEncoder")
[error]  * method newStringSeqEncoder()org.apache.spark.sql.Encoder in class org.apache.spark.sql.SQLImplicits does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("org.apache.spark.sql.SQLImplicits.newStringSeqEncoder")
[error]  * method newIntSeqEncoder()org.apache.spark.sql.Encoder in class org.apache.spark.sql.SQLImplicits does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("org.apache.spark.sql.SQLImplicits.newIntSeqEncoder")
[error]  * method newBooleanSeqEncoder()org.apache.spark.sql.Encoder in class org.apache.spark.sql.SQLImplicits does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("org.apache.spark.sql.SQLImplicits.newBooleanSeqEncoder")
[error]  * method newShortSeqEncoder()org.apache.spark.sql.Encoder in class org.apache.spark.sql.SQLImplicits does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("org.apache.spark.sql.SQLImplicits.newShortSeqEncoder")
```

michalsenkyr:
Not sure how to run MiMa tests locally so I tried my best to figure out what was necessary. Hope this fixes it.
The downside of the fix is that I had to restore the original methods in `SQLImplicits`. I removed the `implicit` keyword and added deprecation annotations as only the new methods should be used from now on. Old code importing the methods explicitly should be fine now.

marmbrus:
For future reference: https://github.com/apache/spark/blob/master/dev/mima (script to run mima)

marmbrus:
ok to test

sql/core/src/main/scala/org/apache/spark/sql/SQLImplicits.scala
```diff
@@ -99,33 +96,96 @@ abstract class SQLImplicits {
 
   // Seqs
 
-  /** @since 1.6.1 */
-  implicit def newIntSeqEncoder: Encoder[Seq[Int]] = ExpressionEncoder()
+  /**
+   * @since 1.6.1
+   * @deprecated use [[newIntSequenceEncoder]]
+   */
+  def newIntSeqEncoder: Encoder[Seq[Int]] = ExpressionEncoder()
```
cloud-fan: oh you doesn't need to do this, just update the `project/MimaExcludes.scala` to add something like `ProblemFilters.exclude[DirectMissingMethodProblem]("org.apache.spark.sql.SQLImplicits.newDoubleSeqEncoder")`. Please see the history of the `MimaExcludes.scala` to see how others update this file.

marmbrus: Wait, I'm not sure I agree... Do we want to break binary compatibility for libraries that might be using this function?  That could have even been resolved implicitly, so it would be confusing when it breaks.

cloud-fan: ah i see, makes sense

cloud-fan:
thanks, merging to master!

cloud-fan:
@michalsenkyr please create 2 more tickets for the optimization you metioned in https://github.com/apache/spark/pull/16240#issuecomment-266318016 and the nested custom collection problem.

PR #16242, opened 2016-12-10
[SPARK-18628][ML] Update Scala param and Python param to have quotes 

krishnakalyan3:

Updated Scala param and Python param to have quotes around the options making it easier for users to read.

## How was this patch tested?

Manually checked the docstrings


krishnakalyan3:
cc @MLnick and @holdenk

srowen (approved):
Trivial but OK. I wonder if there are other similar instances?

PR #16243, opened 2016-12-10
[SPARK-18815] [SQL] Fix NPE when collecting column stats for string/binary column having only null values

wzhfy:

During column stats collection, average and max length will be null if a column of string/binary type has only null values. To fix this, I use default size when avg/max length is null.

## How was this patch tested?

Add a test for handling null columns

### DIFF
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/Statistics.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/Statistics.scala
@@ -194,11 +194,12 @@ object ColumnStat extends Logging {
     val numNonNulls = if (col.nullable) Count(col) else Count(one)
     val ndv = Least(Seq(HyperLogLogPlusPlus(col, relativeSD), numNonNulls))
     val numNulls = Subtract(Count(one), numNonNulls)
+    val defaultSize = Literal(col.dataType.defaultSize, LongType)
 
     def fixedLenTypeStruct(castType: DataType) = {
       // For fixed width types, avg size should be the same as max size.
-      val avgSize = Literal(col.dataType.defaultSize, LongType)
-      struct(ndv, Cast(Min(col), castType), Cast(Max(col), castType), numNulls, avgSize, avgSize)
+      struct(ndv, Cast(Min(col), castType), Cast(Max(col), castType), numNulls, defaultSize,
+        defaultSize)
     }
 
     col.dataType match {
@@ -213,7 +214,8 @@ object ColumnStat extends Logging {
         val nullLit = Literal(null, col.dataType)
         struct(
           ndv, nullLit, nullLit, numNulls,
-          Ceil(Average(Length(col))), Cast(Max(Length(col)), LongType))
+          Coalesce(Seq(Ceil(Average(Length(col))), defaultSize)),
+          Coalesce(Seq(Cast(Max(Length(col)), LongType), defaultSize)))
       case _ =>
         throw new AnalysisException("Analyzing column statistics is not supported for column " +
             s"${col.name} of data type: ${col.dataType}.")
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/Statistics.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/Statistics.scala
@@ -214,6 +214,7 @@ object ColumnStat extends Logging {
         val nullLit = Literal(null, col.dataType)
         struct(
           ndv, nullLit, nullLit, numNulls,
+          // Set avg/max size to default size if all the values are null or there is no value.
           Coalesce(Seq(Ceil(Average(Length(col))), defaultSize)),
           Coalesce(Seq(Cast(Max(Length(col)), LongType), defaultSize)))
       case _ =>
--- a/sql/core/src/test/scala/org/apache/spark/sql/StatisticsCollectionSuite.scala
+++ b/sql/core/src/test/scala/org/apache/spark/sql/StatisticsCollectionSuite.scala
@@ -133,6 +133,79 @@ class StatisticsCollectionSuite extends StatisticsCollectionTestBase with Shared
     }
   }
 
+  test("column stats round trip serialization") {
+    // Make sure we serialize and then deserialize and we will get the result data
+    val df = data.toDF(stats.keys.toSeq :+ "carray" : _*)
+    stats.zip(df.schema).foreach { case ((k, v), field) =>
+      withClue(s"column $k with type ${field.dataType}") {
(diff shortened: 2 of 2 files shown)


wzhfy:
cc @cloud-fan @rxin

sql/core/src/test/scala/org/apache/spark/sql/StatisticsCollectionSuite.scala
```diff
@@ -133,6 +133,79 @@ class StatisticsCollectionSuite extends StatisticsCollectionTestBase with Shared
     }
   }
 
+  test("column stats round trip serialization") {
```
rxin: why did you move the test cases?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/Statistics.scala
```diff
@@ -213,7 +214,8 @@ object ColumnStat extends Logging {
         val nullLit = Literal(null, col.dataType)
         struct(
           ndv, nullLit, nullLit, numNulls,
-          Ceil(Average(Length(col))), Cast(Max(Length(col)), LongType))
+          Coalesce(Seq(Ceil(Average(Length(col))), defaultSize)),
```
rxin: add a line of comment explaining we set the value to defaultSize if all the values are null (of if there is no value).

sql/core/src/test/scala/org/apache/spark/sql/StatisticsCollectionSuite.scala
```diff
@@ -133,6 +133,79 @@ class StatisticsCollectionSuite extends StatisticsCollectionTestBase with Shared
...
+
+  test("analyze column command - result verification") {
+    val tableName = "column_stats_test2"
+    // (data.head.productArity - 1) because the last column does not support stats collection.
+    assert(stats.size == data.head.productArity - 1)
+    val df = data.toDF(stats.keys.toSeq :+ "carray" : _*)
+    checkColStats(df, tableName, stats)
+  }
+
+  private def checkColStats(
```
rxin: Need to add a line of comment documenting this function and change the signature to 

```
/**
  * Compute stats for the given DataFrame and compare it with colStats.
  */
private def checkStats(df: DataFrame, colStats: LinkedHashMap[String, ColumnStat]): Unit = {
}
```

Just generate a random tableName.

rxin: Also don't put this in the middle of the function. Move it into the abstract class StatisticsCollectionTestBase

sql/core/src/test/scala/org/apache/spark/sql/StatisticsCollectionSuite.scala
```diff
@@ -133,6 +133,79 @@ class StatisticsCollectionSuite extends StatisticsCollectionTestBase with Shared
...
+      }
+    }
+  }
+
+  test("column stats collection for null columns") {
+    def nullColumnStat(dataType: DataType): ColumnStat = {
+      ColumnStat(0, None, None, 1, dataType.defaultSize.toLong, dataType.defaultSize.toLong)
+    }
+
+    val tableName = "column_stats_test3"
```
rxin: You can simplify this a lot by doing it programmatically, e.g.

```

val dataTypes = Seq(BooleanType, ByteType, ...)
val df = sql("select " + dataTypes.map(tpe => s"cast(null as ${tpe.sql})").mkString(", "))

val expectedStats = dataTypes.map { tpe =>
  ColumnStat(0, None, None, 1, dataType.defaultSize.toLong, dataType.defaultSize.toLong)
}

...
```

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/Statistics.scala
wzhfy: ok

sql/core/src/test/scala/org/apache/spark/sql/StatisticsCollectionSuite.scala
wzhfy: ok, I'll fix it

cloud-fan: +1

sql/core/src/test/scala/org/apache/spark/sql/StatisticsCollectionSuite.scala
wzhfy: That's much better, thanks!

sql/core/src/test/scala/org/apache/spark/sql/StatisticsCollectionSuite.scala
```diff
@@ -180,35 +214,28 @@ abstract class StatisticsCollectionTestBase extends QueryTest with SQLTestUtils
...
-    val df = data.toDF(stats.keys.toSeq :+ "carray" : _*)
+  private val randomName = new Random(31)
 
+  /**
+   * Compute column stats for the given DataFrame and compare it with colStats.
+   */
+  def checkColStats(
+      df: DataFrame,
+      colStats: mutable.LinkedHashMap[String, ColumnStat]): Unit = {
+    val tableName = "column_stats_test_" + randomName.nextInt(10)
```
rxin: the chance of collision here is not low. just use a broader key space nextInt

wzhfy: ok, done.

sql/core/src/test/scala/org/apache/spark/sql/StatisticsCollectionSuite.scala
```diff
@@ -180,35 +214,28 @@ abstract class StatisticsCollectionTestBase extends QueryTest with SQLTestUtils
...
-    val df = data.toDF(stats.keys.toSeq :+ "carray" : _*)
+  private val randomName = new Random(31)
 
+  /**
+   * Compute column stats for the given DataFrame and compare it with colStats.
+   */
+  def checkColStats(
+      df: DataFrame,
+      colStats: mutable.LinkedHashMap[String, ColumnStat]): Unit = {
+    val tableName = "column_stats_test_" + randomName.nextInt(1000)
```
cloud-fan: why not just use a fixed name like `stats_test_tbl`? It will be used in `withTable`, which means we will drop this table every time.

wzhfy: It's in the parent class, if another subclass suite runs in a different thread and also calls this function, there's a chance of collision. Currently it's only called in one suite, but maybe some other suites will use it in the future, so it's safer to use a random name.

wzhfy:
retest this please

rxin:

PR #16244, opened 2016-12-10
[SQL][minor] simplify a test to fix the maven tests

cloud-fan:

After https://github.com/apache/spark/pull/15620 , all of the Maven-based 2.0 Jenkins jobs time out consistently. As I pointed out in https://github.com/apache/spark/pull/15620#discussion_r91829129 , it seems that the regression test is an overkill and may hit constants pool size limitation, which is a known issue and hasn't been fixed yet.

Since #15620 only fix the code size limitation problem, we can simplify the test to avoid hitting constants pool size limitation.

## How was this patch tested?

test only change

### DIFF
--- a/sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/CodeGenerationSuite.scala
+++ b/sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/CodeGenerationSuite.scala
@@ -98,20 +98,15 @@ class CodeGenerationSuite extends SparkFunSuite with ExpressionEvalHelper {
   }
 
   test("SPARK-18091: split large if expressions into blocks due to JVM code size limit") {
-    val inStr = "StringForTesting"
-    val row = create_row(inStr)
-    val inputStrAttr = 'a.string.at(0)
-
-    var strExpr: Expression = inputStrAttr
-    for (_ <- 1 to 13) {
-      strExpr = If(EqualTo(Decode(Encode(strExpr, "utf-8"), "utf-8"), inputStrAttr),
-        strExpr, strExpr)
+    var strExpr: Expression = Literal("abc")
+    for (_ <- 1 to 150) {
+      strExpr = Decode(Encode(strExpr, "utf-8"), "utf-8")
     }
 
-    val expressions = Seq(strExpr)
-    val plan = GenerateUnsafeProjection.generate(expressions, true)
-    val actual = plan(row).toSeq(expressions.map(_.dataType))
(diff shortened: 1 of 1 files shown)


cloud-fan:
cc @srowen @kapilsingh5050 @ueshin @viirya

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/CodeGenerationSuite.scala
```diff
@@ -98,20 +98,15 @@ class CodeGenerationSuite extends SparkFunSuite with ExpressionEvalHelper {
   }
 
   test("SPARK-18091: split large if expressions into blocks due to JVM code size limit") {
```
kapilsingh5050: @cloud-fan Just confirming, have you made sure that this test fails without the fix for SPARK-18091?

cloud-fan: what do you mean? this test was added by SPARK-18091

kapilsingh5050: This test simulates the scenario where large code for components of If expression causes JVM's method code size limit to be hit. So the point of having this test is if it fails before the fix for SPARK-18091 was committed and it passes afterwards.

cloud-fan: yea of course, I reverted SPARK-18091 and ran this test locally, it failed.

viirya:
How do we make sure the fixed test passes Maven-based tests?

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/CodeGenerationSuite.scala
kapilsingh5050: great!

cloud-fan:
Actually I can't reproduce this issue locally, but by looking at the logs, I'm 90% percent sure this is the cause. The only way to verify it may be merging and checking the jenkins maven status again.

srowen (approved):

rxin:
sgtm

viirya:
@cloud-fan Ok. I see. LGTM.

srowen:
Merged to master and 2.1, in order to see if this resolves the test failures for 2.1

PR #16245, opened 2016-12-10
[SPARK-18824][SQL] Add optimizer rule to reorder Filter predicates

viirya:

During evaluation of predicates in Filter, we can reorder the expressions in order to evaluate the more expensive expressions like ScalaUDF later. So if other expressions are evaluated to false, we can avoid evaluation of these UDFs.

## How was this patch tested?

Jenkins tests.

### DIFF
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/expressions.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/expressions.scala
@@ -514,6 +514,34 @@ case class OptimizeCodegen(conf: CatalystConf) extends Rule[LogicalPlan] {
 
 
 /**
+ * Reorders the predicates in `Filter` so more expensive expressions like UDF can evaluate later.
+ */
+object ReorderPredicatesInFilter extends Rule[LogicalPlan] with PredicateHelper {
+  def apply(plan: LogicalPlan): LogicalPlan = plan transform {
+    case f @ Filter(pred, child) =>
+      // Extracts deterministic suffix expressions from Filter predicate.
+      val expressions = splitConjunctivePredicates(pred)
+      // The beginning index of the deterministic suffix expressions.
+      var splitIndex = -1
+      (expressions.length - 1 to 0 by -1).foreach { idx =>
+        if (splitIndex == -1 && !expressions(idx).deterministic) {
+          splitIndex = idx + 1
+        }
+      }
+      if (splitIndex == expressions.length) {
+        // All expressions are non-deterministic, no reordering.
+        f
+      } else {
+        val (nonDeterminstics, deterministicExprs) = expressions.splitAt(splitIndex)
+        val (udfs, condExprs) = deterministicExprs.partition { e =>
+          e.find(_.isInstanceOf[ScalaUDF]).isDefined
+        }
+        Filter((nonDeterminstics ++ condExprs ++ udfs).reduce(And), child)
+      }
+  }
+}
+
+/**
  * Removes [[Cast Casts]] that are unnecessary because the input is already the correct type.
  */
 object SimplifyCasts extends Rule[LogicalPlan] {
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/Optimizer.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/Optimizer.scala
@@ -103,6 +103,7 @@ abstract class Optimizer(sessionCatalog: SessionCatalog, conf: CatalystConf)
       SimplifyConditionals,
       RemoveDispensableExpressions,
       SimplifyBinaryComparison,
+      ReorderPredicatesInFilter,
       PruneFilters,
       EliminateSorts,
       SimplifyCasts,
--- a/sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/optimizer/ReorderPredicatesInFilterSuite.scala
+++ b/sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/optimizer/ReorderPredicatesInFilterSuite.scala
@@ -1,0 +1,104 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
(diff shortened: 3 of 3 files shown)


sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/expressions.scala
```diff
@@ -514,6 +514,19 @@ case class OptimizeCodegen(conf: CatalystConf) extends Rule[LogicalPlan] {
 
 
 /**
+ * Reorders the predicates in `Filter` so more expensive expressions like UDF can evaluate later.
+ */
+object ReorderPredicatesInFilter extends Rule[LogicalPlan] with PredicateHelper {
+  def apply(plan: LogicalPlan): LogicalPlan = plan transform {
+    case Filter(pred, child) =>
+      val (deterministicExprs, others) = splitConjunctivePredicates(pred).reverse
```
hvanhovell: So this is generally a good idea. However we need to be super careful with reordering non-deterministic expressions here (we should not).

viirya: Yea I see. I only reorder the deterministic expression in the suffix of predicates.

hvanhovell: +1

viirya:
I will create Jira and add test for this later.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/expressions.scala
```diff
@@ -514,6 +514,25 @@ case class OptimizeCodegen(conf: CatalystConf) extends Rule[LogicalPlan] {
...
+ * Reorders the predicates in `Filter` so more expensive expressions like UDF can evaluate later.
+ */
+object ReorderPredicatesInFilter extends Rule[LogicalPlan] with PredicateHelper {
+  def apply(plan: LogicalPlan): LogicalPlan = plan transform {
+    case Filter(pred, child) =>
+      // Reverses the expressions to get the suffix deterministic expressions in the predicate.
+      // E.g., the original expressions are 'a > 1, rand(0), 'b > 2, 'c > 3.
+      // The reversed expressions are 'c > 3, 'b > 2, rand(0), 'a > 1.
+      // The suffix deterministic expressions are 'c > 3, 'b > 2.
+      val (deterministicExprs, others) = splitConjunctivePredicates(pred).reverse
```
rxin: how is the performance of this split, reverse, and span?

viirya: the split is widely used in optimizer. i think i may rewrite this reverse and span to alleviate performance concern.

viirya:
retest this please.

viirya:
ping @hvanhovell @rxin Any thing I need to address more? Thanks.

viirya:
ping @hvanhovell @rxin again. Please let me know if you have more comments on this. Thanks.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/expressions.scala
```diff
@@ -514,6 +514,34 @@ case class OptimizeCodegen(conf: CatalystConf) extends Rule[LogicalPlan] {
...
+      (expressions.length - 1 to 0 by -1).foreach { idx =>
+        if (splitIndex == -1 && !expressions(idx).deterministic) {
+          splitIndex = idx + 1
+        }
+      }
+      if (splitIndex == expressions.length) {
+        // All expressions are non-deterministic, no reordering.
+        f
+      } else {
+        val (nonDeterminstics, deterministicExprs) = expressions.splitAt(splitIndex)
```
chenghao-intel: I am a little confused why we need to separate the `non-deterministic`? Should be the `stateful` or `foldable`?

viirya: Reordering non-deterministic expressions might change the evaluation results. I think `foldable` expressions are handled by other rule already. And I remember we don't have explicit `stateful` expression or it is classified as `non-deterministic` too.

chenghao-intel:
Actually I doubt this is really an optimization, as the assumption of Scala UDF is slower than the non-SCALA UDF probably not always true.

viirya:
I think most of time it should be as Scala UDF needs extra conversion between internal format and external format on input and out.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/expressions.scala
chenghao-intel: I mean `(rand() > 0) && b)` should equals to `b && (rand() >0)`, and even, the latter probably has better performance, due to the short cut evaluation of `AND`. isn't it?

viirya: yes. however, if the first expression in the `AND` is `non-deterministic`, skipping it might change its next evaluation. so we can only reorder the deterministic expressions after non-deterministic expressions.

chenghao-intel:
I think that's true in most of time for`Scala UDF needs extra conversion between internal format and external format on input and out`, not all of the time, for example, some built-in string based operations and its combinations are also quite heavy in evaluation, and most likely, this probably causes concern for an experienced SQL developers, to write an optimal(business related short-cutting logic) SQL expressions.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/expressions.scala
chenghao-intel: Hmm, actually that's what I mean, probably some confusing with `non-deterministic` with `non-foldable`? I think we can skip them both in a short cut evaluation. as those expressions are not `stateful`(unfortunately, Spark SQL expression doesn't have the concept of `stateful`), so skip the evaluation of them are harmless, and this is exactly the short cut logic of expression `AND`.

viirya:
It is true of course you can construct a combination of complex string operations and compare it with a simple Scala UDF. But as you said, the previous claim is true in most of time. I also think Scala UDF is usually used to write complex logic which can't be achieved by built-in expressions.

viirya:
Close this for now since no further feedback.

PR #16246, opened 2016-12-10
[SPARK-18814][SQL] CheckAnalysis rejects TPCDS query 32

nsyca:
Move the checking of GROUP BY column in correlated scalar subquery from CheckAnalysis
to Analysis to fix a regression caused by SPARK-18504.

This problem can be reproduced with a simple script now.

Seq((1,1)).toDF("pk","pv").createOrReplaceTempView("p")
Seq((1,1)).toDF("ck","cv").createOrReplaceTempView("c")
sql("select * from p,c where p.pk=c.ck and c.cv = (select avg(c1.cv) from c c1 where c1.ck = p.pk)").show

The requirements are:
1. We need to reference the same table twice in both the parent and the subquery. Here is the table c.
2. We need to have a correlated predicate but to a different table. Here is from c (as c1) in the subquery to p in the parent.
3. We will then "deduplicate" c1.ck in the subquery to `ck#<n1>#<n2>` at `Project` above `Aggregate` of `avg`. Then when we compare `ck#<n1>#<n2>` and the original group by column `ck#<n1>` by their canonicalized form, which is #<n2> != #<n1>. That's how we trigger the exception added in SPARK-18504.

## How was this patch tested?

SubquerySuite and a simplified version of TPCDS-Q32


sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
```diff
@@ -1230,10 +1230,24 @@ class Analyzer(
...
-        outer: Seq[LogicalPlan]): (LogicalPlan, Seq[Expression]) = {
+        outer: Seq[LogicalPlan],
+        scalarSubq: Boolean = false): (LogicalPlan, Seq[Expression]) = {
       // Pull out the tagged predicates and rewrite the subquery in the process.
       val (basePlan, baseConditions) = pullOutCorrelatedPredicates(sub)
 
+      // SPARK-18504: block cases where GROUP BY columns
+      // are not part of the correlated columns
+      if (scalarSubq && sub.isInstanceOf[Aggregate]) {
+        val groupByCols = ExpressionSet.apply(sub.asInstanceOf[Aggregate].
```
gatorsmile: Nomally, when we need to do `isInstanceOf` and `asInstanceOf`, we use the patten matching.

gatorsmile:
Do you what code changes caused the regression? 

In addition, could you post your investigation in the PR description?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
nsyca: Thank you, @gatorsmile. I will make the change with other comments that may come.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
```diff
@@ -1230,10 +1230,24 @@ class Analyzer(
...
+          groupingExpressions.flatMap(_.references))
+        val conditionsCols = ExpressionSet.apply(baseConditions.flatMap(_.references))
```
gatorsmile: `ExpressionSet(baseConditions.flatMap(_.references))`

nsyca:
This problem can be reproduced with a simple script now.

````scala
Seq((1,1)).toDF("pk","pv").createOrReplaceTempView("p")
Seq((1,1)).toDF("ck","cv").createOrReplaceTempView("c")
sql("select * from p,c where p.pk=c.ck and c.cv = (select avg(c1.cv) from c c1 where c1.ck = p.pk)").show
````

The requirements are:
1. We need to reference the same table twice in both the parent and the subquery. Here is the table c.
2. We need to have a correlated predicate but to a different table. Here is from c (as c1) in the subquery to p in the parent.
3. We will then "deduplicate" c1.ck in the subquery to `ck#<n1>#<n2>` at `Project` above `Aggregate` of `avg`. Then when we compare `ck#<n1>#<n2>` and the original group by column `ck#<n1>` by their canonicalized form, which is #<n2> != #<n1>. That's how we trigger the exception I added in SPARK-18504.

File: sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala (hunk 2 above)
Comment (nsyca) [inline sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala]: Yep. Thank you. I guess it applies to the one at line 1241 too.

File: sql/core/src/test/scala/org/apache/spark/sql/SubquerySuite.scala [hunk 3]
```diff
@@ -495,6 +495,17 @@ class SubquerySuite extends QueryTest with SharedSQLContext {
     }
   }
 
+  test("SPARK-18814 extra GROUP BY column in correlated scalar subquery is not permitted") {
+    withTempView("p", "c") {
+      Seq((1, 1)).toDF("pk", "pv").createOrReplaceTempView("p")
+      Seq((1, 1)).toDF("ck", "cv").createOrReplaceTempView("c")
+      checkAnswer(
+        sql("select pk, cv from p,c where p.pk=c.ck and " +
+          "c.cv = (select avg(c1.cv) from c c1 where c1.ck = p.pk)"),
```
gatorsmile: Move it to `SQLQueryTestSuite`?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
```diff
@@ -1230,10 +1230,24 @@ class Analyzer(
...
+        val invalidCols = groupByCols.diff(conditionsCols)
+        if (invalidCols.nonEmpty) {
+          failAnalysis("a GROUP BY clause in a scalar correlated subquery " +
```
gatorsmile: `a` -> `A`

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
```diff
@@ -1230,10 +1230,24 @@ class Analyzer(
...
     private def rewriteSubQuery(
         sub: LogicalPlan,
-        outer: Seq[LogicalPlan]): (LogicalPlan, Seq[Expression]) = {
+        outer: Seq[LogicalPlan],
+        scalarSubq: Boolean = false): (LogicalPlan, Seq[Expression]) = {
       // Pull out the tagged predicates and rewrite the subquery in the process.
       val (basePlan, baseConditions) = pullOutCorrelatedPredicates(sub)
 
+      // SPARK-18504: block cases where GROUP BY columns
+      // are not part of the correlated columns
```
gatorsmile: Please update the comment.

gatorsmile:
@nsyca Edit the PR description and move the investigation there. Then, when we merge the PR, it will be part of the commit description.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
```diff
@@ -1298,7 +1312,7 @@ class Analyzer(
             s"does not match the required number of columns ($requiredColumns)")
         }
```
gatorsmile: Why we need to issue the error message in `rewriteSubQuery`? Instead of here? If we do it here, we do not need to pass `e.isInstanceOf[ScalarSubquery]` to `rewriteSubQuery `

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
```diff
@@ -1230,10 +1230,24 @@ class Analyzer(
      */
```
gatorsmile: Please also update the above function comments. Thanks!

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
```diff
@@ -1225,15 +1225,37 @@ class Analyzer(
...
+     *
+     * If this is a scalar subquery, check that GROUP BY columns are a subset
+     * of the columns used in the correlated predicate(s). Otherwise, pulling up
+     * correlated predicates could cause incorrect results.
      */
     private def rewriteSubQuery(
         sub: LogicalPlan,
-        outer: Seq[LogicalPlan]): (LogicalPlan, Seq[Expression]) = {
+        outer: Seq[LogicalPlan],
+        scalarSubq: Boolean = false): (LogicalPlan, Seq[Expression]) = {
```
gatorsmile: No need to define the default value, right?

gatorsmile: How about renaming it to `isScalarSubq `

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
```diff
@@ -1225,15 +1225,37 @@ class Analyzer(
...
       // Pull out the tagged predicates and rewrite the subquery in the process.
       val (basePlan, baseConditions) = pullOutCorrelatedPredicates(sub)
 
+      // SPARK-18504/SPARK-18814:
+      // Block cases where GROUP BY columns are not part of the correlated columns
+      // of a scalar subquery.
+      if (scalarSubq) {
+        sub match {
+          case a @ Aggregate(grouping, _, _) =>
```
gatorsmile: No need to add `if` before patten matching.
-> 
```
      sub match {
        case Aggregate(grouping, _, _) if scalarSubq =>
```

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
```diff
@@ -1225,15 +1225,37 @@ class Analyzer(
...
+            val groupByCols = ExpressionSet(grouping.flatMap(_.references))
+            val conditionsCols = ExpressionSet(baseConditions.flatMap(_.references))
+            val invalidCols = groupByCols.diff(conditionsCols)
+            if (invalidCols.nonEmpty) {
+              failAnalysis("A GROUP BY clause in a scalar correlated subquery " +
+                "cannot contain non-correlated columns: " +
+                invalidCols.mkString(","))
+            }
+          case _ => None
```
gatorsmile: This is not needed, if we use `collect` to replace the above `match`

gatorsmile:
cc @hvanhovell @ericl

nsyca:
Thank you, @gatorsmile, for your code review comments.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
```diff
@@ -1225,15 +1225,34 @@ class Analyzer(
...
+        outer: Seq[LogicalPlan],
+        isScalarSubq: Boolean): (LogicalPlan, Seq[Expression]) = {
       // Pull out the tagged predicates and rewrite the subquery in the process.
       val (basePlan, baseConditions) = pullOutCorrelatedPredicates(sub)
 
+      // SPARK-18504/SPARK-18814:
+      // Block cases where GROUP BY columns are not part of the correlated columns
+      // of a scalar subquery.
+      sub collect {
+        case a @ Aggregate(grouping, _, _) if (isScalarSubq) =>
```
gatorsmile: `case a @ Aggregate(grouping, _, _) if (isScalarSubq)`
-> 
`case Aggregate(grouping, _, _) if isScalarSubq`

nsyca:
retest this please.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
```diff
@@ -1225,15 +1225,34 @@ class Analyzer(
...
+        outer: Seq[LogicalPlan],
+        isScalarSubq: Boolean): (LogicalPlan, Seq[Expression]) = {
       // Pull out the tagged predicates and rewrite the subquery in the process.
       val (basePlan, baseConditions) = pullOutCorrelatedPredicates(sub)
 
+      // SPARK-18504/SPARK-18814:
+      // Block cases where GROUP BY columns are not part of the correlated columns
+      // of a scalar subquery.
+      sub collect {
+        case Aggregate(grouping, _, _) if isScalarSubq =>
```
hvanhovell: Doesn't this break if you have nested aggregates in the subquery? I.e.:
```sql
create or replace temporary view t as
select id,
       id % 10 val,
       id % 100 fk
from   range(1000);

select *
from   t
where t.id = (select max(id)
              from   (select tt.val,
                             max(tt.id) as id
                      from   t as tt
                      where  t.fk = tt.fk
                      group by tt.val))
```

hvanhovell:
I think we need to revisit the rule in CheckAnalysis, and make it `Alias` aware. We only need to keep track of the inner references. We traverse/recurse down the tree and do a few things:
- When we encounter any other than a `Project`/`Filter`/`SubqueryAlias`/`Aggregate`, we should fail.
- When we encounter a Project. Assume that there are only Alias or AttributeReference expressions (this should be the case anyway), and update the references for columns that are aliased.
- When we encounter a `Filter`/`SubqueryAlias`, do Nothing.
- When we encounter a toplevel Aggregate, validate the grouping expressions using the updated references (the current logic should be fine here), and return.

gatorsmile:
@hvanhovell I like your ideas. : ) 

This JIRA is in the `Blocker` level of Spark 2.1. If we are doing a major refactoring in `CheckAnalysis `, is it too risky at the last minute? I am fine if you think the above proposal has a limited impact. Thanks!

nsyca:
I am working on the code based on @hvanhovell's proposal.

gatorsmile:
Sure, thanks! @nsyca

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/CheckAnalysis.scala
```diff
@@ -124,6 +124,12 @@ trait CheckAnalysis extends PredicateHelper {
                 s"Scalar subquery must return only one column, but got ${query.output.size}")
 
           case s @ ScalarSubquery(query, conditions, _) if conditions.nonEmpty =>
+
+            // Collect the columns from the subquery for further checking.
+            var subqueryColumns = conditions.flatMap(_.references).collect {
```
hvanhovell: NIT: `conditions.flatMap(_.references).filter(query.output.contains)`

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/CheckAnalysis.scala
```diff
@@ -136,24 +142,36 @@ trait CheckAnalysis extends PredicateHelper {
...
               }
 
-              // SPARK-18504: block cases where GROUP BY columns
-              // are not part of the correlated columns
-              val groupByCols = ExpressionSet.apply(agg.groupingExpressions.flatMap(_.references))
-              val predicateCols = ExpressionSet.apply(conditions.flatMap(_.references))
-              val invalidCols = groupByCols.diff(predicateCols)
+              // SPARK-18504/SPARK-18814: Block cases where GROUP BY columns
+              // are not part of the correlated columns.
+              val groupByCols = ExpressionSet(agg.groupingExpressions.flatMap(_.references))
```
hvanhovell: Nit: Using an `AttributeSet` is more natural (I should have seen that for the initial PR), and probably a little faster. You can perform a `diff` by subtracting the sets, i.e.: 'val invalidCols = groupByCols -- correlatedCols`

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/CheckAnalysis.scala
```diff
@@ -136,24 +142,36 @@ trait CheckAnalysis extends PredicateHelper {
...
             def cleanQuery(p: LogicalPlan): LogicalPlan = p match {
               case s: SubqueryAlias => cleanQuery(s.child)
-              case p: Project => cleanQuery(p.child)
+              case p: Project =>
+                // SPARK-18814: Map any aliases to their AttributeReference children
+                // for the checking in the Aggregate operators below this Project.
+                subqueryColumns = subqueryColumns.map {
+                  case xs =>
+                    p.projectList.collectFirst {
+                      case e @ Alias(child : AttributeReference, _) if e.toAttribute equals xs =>
```
hvanhovell: Nit: It is quicker to compare the `exprId`s of the alias and the attribute, e.g.:
`case e @ Alias(child : AttributeReference, _) if e.exprId == xs.exprId =>`

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/CheckAnalysis.scala
```diff
@@ -136,24 +142,36 @@ trait CheckAnalysis extends PredicateHelper {
...
+            // Skip subquery aliases added by the Analyzer and the SQLBuilder.
+            // For projects, do the necessary mapping and skip to its child.
             def cleanQuery(p: LogicalPlan): LogicalPlan = p match {
               case s: SubqueryAlias => cleanQuery(s.child)
-              case p: Project => cleanQuery(p.child)
+              case p: Project =>
+                // SPARK-18814: Map any aliases to their AttributeReference children
+                // for the checking in the Aggregate operators below this Project.
+                subqueryColumns = subqueryColumns.map {
+                  case xs =>
```
hvanhovell: Nit: you don't need the case statement here.

hvanhovell (approved):
Some minor things, but otherwise LGTM.

hvanhovell:
LGTM - pending jenkins.

nsyca:
@hvanhovell, I really appreciate your time reviewing my code. I have addressed all your four comments. Thanks!

sql/core/src/test/resources/sql-tests/inputs/scalar-subquery.sql
```diff
@@ -0,0 +1,20 @@
...
+
+-- SPARK-18814.2: Adding stack of aggregates
+SELECT pk, cv
+FROM   p, c
+WHERE  p.pk = c.ck
+AND    c.cv = (SELECT max(avg)
+	       FROM   (SELECT   c1.cv, avg(c1.cv) avg
+		       FROM     c c1
+		       WHERE    c1.ck = p.pk
+                       GROUP BY c1.cv));
```
gatorsmile: I have a question that is not related to this JIRA. In the above query, if we do not have the `GROUP BY c1.cv`, it still works. It sounds like the subquery progressing ignores group by clauses. What is the reason?

hvanhovell:
Ok, I am merging this to master/2.1/2.0. Thanks!

sql/core/src/test/resources/sql-tests/inputs/scalar-subquery.sql
nsyca: Nice catch! I think this is a bug. There could be multiple values of c1.cv. Without a GROUP BY clause, which value does it return? I will investigate along with my subquery work. Do you think this is a blocker?

nsyca: I opened [SPARK-18863](https://issues.apache.org/jira/browse/SPARK-18863) to track this problem.

gatorsmile: Thank you confirming this is a bug. I expect I can get the error message like
```
org.apache.spark.sql.AnalysisException: grouping expressions sequence is empty, and 'c1.`cv`' is not an aggregate function. 
```

gatorsmile: Or in some other cases, we should see the error message like 
```
org.apache.spark.sql.AnalysisException: expression 'c1.`cv`' is neither present in the group by, nor is it an aggregate function.
```

Both error handling are missing.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/CheckAnalysis.scala
```diff
@@ -136,24 +140,35 @@ trait CheckAnalysis extends PredicateHelper {
...
               }
             }
 
-            // Skip projects and subquery aliases added by the Analyzer and the SQLBuilder.
+            // Skip subquery aliases added by the Analyzer and the SQLBuilder.
+            // For projects, do the necessary mapping and skip to its child.
             def cleanQuery(p: LogicalPlan): LogicalPlan = p match {
               case s: SubqueryAlias => cleanQuery(s.child)
-              case p: Project => cleanQuery(p.child)
+              case p: Project =>
```
dilipbiswal: Now the de-duplication of sub plan happens in optimizer when we actually pull up the correlated predicates. Therefore the following code block is simplified.

PR #16247, opened 2016-12-10
[SPARK-18817][SparkR] set default spark-warehouse path to tempdir()

bdwyer2:

Set the default location of `spark.sql.warehouse.dir` to be compliant with the CRAN policy (https://cran.r-project.org/web/packages/policies.html) regarding writing files outside of the tmp directory. Previously a folder named `spark-warehouse` was created in the working directory when `sparkR.session()` was called.

See SPARK-15799 for discussion.
cc @shivaram 

## How was this patch tested?

Added new test and manually verified nothing was created in my working directory after running the following code:
```R
sparkR.session(master = "local[*]",
               sparkConfig = list(spark.driver.memory = "2g"),
               enableHiveSupport = FALSE)
```

### DIFF
--- a/R/pkg/R/sparkR.R
+++ b/R/pkg/R/sparkR.R
@@ -362,6 +362,10 @@ sparkR.session <- function(
   enableHiveSupport = TRUE,
   ...) {
 
+  if(length(sparkConfig[["spark.sql.warehouse.dir"]]) == 0) {
+    sparkConfig[["spark.sql.warehouse.dir"]] <- tempdir()
+  }
+  
   sparkConfigMap <- convertNamedListToEnv(sparkConfig)
   namedParams <- list(...)
   if (length(namedParams) > 0) {
--- a/R/pkg/R/sparkR.R
+++ b/R/pkg/R/sparkR.R
@@ -362,10 +362,10 @@ sparkR.session <- function(
   enableHiveSupport = TRUE,
   ...) {
 
-  if(length(sparkConfig[["spark.sql.warehouse.dir"]]) == 0) {
+  if (length(sparkConfig[["spark.sql.warehouse.dir"]]) == 0) {
     sparkConfig[["spark.sql.warehouse.dir"]] <- tempdir()
   }
-  
+
   sparkConfigMap <- convertNamedListToEnv(sparkConfig)
   namedParams <- list(...)
   if (length(namedParams) > 0) {
--- a/R/pkg/inst/tests/testthat/test_sparkR.R
+++ b/R/pkg/inst/tests/testthat/test_sparkR.R
@@ -48,8 +48,13 @@ test_that("sparkCheckInstall", {
 test_that("sparkR.session", {
   # nothing should be written outside tempdir() without explicit user permission
   inital_working_directory_files <- list.files()
-  sparkR.session()
-  result <- sql("SELECT 1")
+  sparkR.session(enableHiveSupport = FALSE)
+  df <- data.frame("col1" = c(1, 2, 3, 4, 5, 6),
+                   "col2" = c(1, 0, 0, 1, 1, 0),
+                   "col3" = c(1, 0, 0, 2, 6, 2))
+  df <- as.DataFrame(df)
+  createOrReplaceTempView(df, "table")
+  result <- sql("SELECT * FROM `table`")
   sparkR.session.stop()
   expect_equal(inital_working_directory_files, list.files())
 })
--- a/R/pkg/inst/tests/testthat/test_sparkR.R
+++ b/R/pkg/inst/tests/testthat/test_sparkR.R
@@ -46,15 +46,10 @@ test_that("sparkCheckInstall", {
 })
 
 test_that("sparkR.session", {
-  # nothing should be written outside the tempdir() without explicit user premission
+  # nothing should be written outside tempdir() without explicit user permission
   inital_working_directory_files <- list.files()
   sparkR.session()
-  df <- data.frame("col1" = c(1, 2, 3, 4, 5, 6),
-                   "col2" = c(1, 0, 0, 1, 1, 0),
-                   "col3" = c(1, 0, 0, 2, 6, 2))
-  df <- as.DataFrame(df)
-  createOrReplaceTempView(df, "table")
-  result <- sql("SELECT * FROM `table`")
+  result <- sql("SELECT 1")
   sparkR.session.stop()
   expect_equal(inital_working_directory_files, list.files())
 })
--- a/R/pkg/inst/tests/testthat/test_sparkR.R
+++ b/R/pkg/inst/tests/testthat/test_sparkR.R
@@ -44,3 +44,11 @@ test_that("sparkCheckInstall", {
   deployMode <- "client"
   expect_error(sparkCheckInstall(sparkHome, master, deployMode))
 })
+
+test_that("sparkR.session", {
+  # nothing should be written outside the tempdir() without explicit user premission
+  inital_working_directory_files <- list.files()
+  sparkR.session()
+  expect_equal(inital_working_directory_files, list.files())
+
+})
--- a/R/pkg/inst/tests/testthat/test_sparkR.R
+++ b/R/pkg/inst/tests/testthat/test_sparkR.R
@@ -49,6 +49,12 @@ test_that("sparkR.session", {
   # nothing should be written outside the tempdir() without explicit user premission
   inital_working_directory_files <- list.files()
   sparkR.session()
+  df <- data.frame("col1" = c(1, 2, 3, 4, 5, 6),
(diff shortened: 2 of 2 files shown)


HyukjinKwon:
I may missed something but isn't tmpdir per R's session? Maybe I should have tried it first but I guess it wouldn't be accessible after the R session is ended and another one is started. Is this behaviour expected?

shivaram:
Jenkins, ok to test

bdwyer2:
@HyukjinKwon yes but we are restricted by CRAN policies

> Packages should not write in the users’ home filespace, nor anywhere else on the file system apart from the R session’s temporary directory

shivaram:
@HyukjinKwon That is a good point. As @bdwyer2 says the question is one of defaults. If the user does specify a more permanent location we will use that.  I wonder if we should print a warning or info message as well ?

HyukjinKwon:
I see. Thank you both for your comments. (BTW, this should have a JIRA I think.)

bdwyer2:
Should I open a JIRA under SPARK-15799 myself or leave that to one of the admins?

shivaram:
@bdwyer2 Feel free to open a JIRA as a sub-task under SPARK-15799

R/pkg/R/sparkR.R
```diff
@@ -362,6 +362,10 @@ sparkR.session <- function(
   enableHiveSupport = TRUE,
   ...) {
 
+  if (length(sparkConfig[["spark.sql.warehouse.dir"]]) == 0) {
+    sparkConfig[["spark.sql.warehouse.dir"]] <- tempdir()
```
shivaram: As I mentioned in the JIRA, I think this change is not enough. If the user has a hive-site.xml and the warehouse dir set inside that, this change will override that [1]. We might need to change the Scala code and/or add a new option to specify that a new default for SparkR.

[1] https://github.com/apache/spark/blob/cf33a86285629abe72c1acf235b8bfa6057220a8/sql/core/src/main/scala/org/apache/spark/sql/internal/SharedState.scala#L49

shivaram:
@bdwyer2 One more thing: Is there a good way to test this ?

felixcheung:
I think we should only change `spark.sql.warehouse.dir` when we are loading SparkR as a package. This should minimize changes in the case where we are running in cluster mode and so on.

For that purpose, let's check for R `interactive()`.

Conceptually, we might want to pass the R `tempdir()` along as a property and then check again after config and hive config is applied, if it is still not set, take that property and set it to `spark.sql.warehouse.dir`.

shivaram:
Or we could introduce a new property say spark.sql.default.warehouse and
set that to tmpdir()

felixcheung:
... that's actually what I meant with "we might want to pass the R tempdir() along as a property" <-- this would be a new property and not `spark.sql.warehouse.dir`

R/pkg/R/sparkR.R
bdwyer2: How about an argument named `sparkWorkingDirectory` that defaulted to `tempdir()`?

bdwyer2:
@shivaram I can create a test to verify the output of `list.files()` is the same before and after running `sparkR.session()`.

shivaram:
@bdwyer2 The test case idea sounds good ! 

Regarding the conf naming for warehouse dir lets also check with contributors who are more familiar with SQL. 
cc @yhuai @cloud-fan

bdwyer2:
I don't see how my last commit could of caused this
```
functions in sparkR.R: .....
SparkSQL functions: Spark package found in SPARK_HOME: /home/jenkins/workspace/SparkPullRequestBuilder
Error in handleErrors(returnStatus, conn) : 
  java.lang.IllegalArgumentException: Error while instantiating 'org.apache.spark.sql.hive.HiveSessionState':
	at org.apache.spark.sql.SparkSession$.org$apache$spark$sql$SparkSession$$reflect(SparkSession.scala:981)
	at org.apache.spark.sql.SparkSession.sessionState$lzycompute(SparkSession.scala:110)
	at org.apache.spark.sql.SparkSession.sessionState(SparkSession.scala:109)
	at org.apache.spark.sql.api.r.SQLUtils$$anonfun$setSparkContextSessionConf$2.apply(SQLUtils.scala:67)
	at org.apache.spark.sql.api.r.SQLUtils$$anonfun$setSparkContextSessionConf$2.apply(SQLUtils.scala:66)
	at scala.collection.TraversableLike$WithFilter$$anonfun$foreach$1.apply(TraversableLike.scala:733)
	at scala.collection.TraversableLike$WithF
Calls: test_package ... sparkR.session -> callJStatic -> invokeJava -> handleErrors
```

jodersky:
jenkins, retest this please

cloud-fan:
a new property for default warehouse LGTM

R/pkg/R/sparkR.R
felixcheung: I think Shivaram is talking about a spark property, not a parameter (if perhaps your camel casing is an indication)
Basically, we don't want to change `spark.sql.warehouse.dir` because it could be set already in an earlier point (just not accessible here)

felixcheung:
re: test failure. it might be related to this change? the call stack is hidden, it should be trying to call into https://github.com/apache/spark/blob/master/sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveSessionState.scala and failed.

Is it possible that tmpdir() is not writeable on Jenkins?

HyukjinKwon:
This test failure seems related with this PR. It seems because the previous Hive-enabled spark session is not closed properly between `test_saprkSQL.R` and `test_sparkR.R` here, in particular, I suspect the lock in derby via Hive client. It seems apparently related with SPARK-16027?

There was only single instance of Hive-enabled spark session there but the test here introduces another one. I manually tested for each after removing each other and it seems working fine.

felixcheung:
Possibly,  SPARK-16027 was just a hack, the root issue remains I think

bdwyer2:
Would calling the test with this be an acceptable solution?
```R
sparkR.session(enableHiveSupport = FALSE)
```

shivaram:
Yeah disabling Hive for the test is fine. @bdwyer2 Can you add the new config flag as well ? We can do one final pass of review after that

R/pkg/R/sparkR.R
aloknsingh: There is the pb of circular dependency for getting the hive warehouse dir.

I think currently in sparkR, to create SparkSession, we need all those parameter and once spark session is there we can call 
```R
sparkContext <- SparkR:::callJMethod(sc, "sparkContext") 
haddopConf <- SparkR:::callJMethod(sparkContext, "hadoopConfiguration")
hivewarehouseDir <- SparkR:::callJMethod(hadoopConf, "get", "hive.metastore.warehouse.dir"") 
```

but the above logic requires that we have sparksession available but there is no spark session yet at this line of code. So there must a way to get the hive.conf directory from env.
we have to make decision that is it ok to get the env from system or not? something like the following

```R
require(XML)
data <- xmlParse(file.path(Sys.env("HADOOP_HOME", "conf", "hive-site.xml")
xml_data <- xmlToList(data)
```

it will make sparkR dependent on xml parser . just some thoughts :)

bdwyer2: @felixcheung I'm confused. By "spark property" do you mean something passed to `sparkR.session()` via the `sparkConfig` argument?

felixcheung:
I mean it as something we set to the SparkContext or SparkSession and not a parameter of sparkR.session().

felixcheung:
But yes other Spark config properties would be set by the user in sparkConfig parameter of sparkR.session method. We would just add to that without adding another parameter to sparkR.session

felixcheung:
But yes _other_ Spark config properties would be set by the user in sparkConfig parameter of sparkR.session method. We would just add to that like
```
sparkConfig[["spark.sql.warehouse.default.dir"]] <- tempdir()
```
without adding another parameter to sparkR.session()

bdwyer2:
How would we access that value on the scala side? Would `sparkContext.hadoopConfiguration.get("spark.sql.warehouse.default.dir")` work?

I'm currently unable to compile Spark which makes experimenting with scala difficult.

felixcheung:
yes, something like what's being used here https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/internal/SharedState.scala#L49

shivaram:
@bdwyer2 Let us know if you have problems setting up the environment - If so me or @felixcheung can open a new PR that includes your changes (we can still assign the JIRA as your contribution)

The reason I ask is that it will be good to get this in before the RC3 cut as this helps the SparkR CRAN release etc.

bdwyer2:
@shivaram @felixcheung I'll close this PR so that one of you can take over in order to have it done in time for the RC.

PR #16248, opened 2016-12-11
[SPARK-18810][SPARKR] SparkR install.spark does not work for RCs, snapshots

felixcheung:

Support overriding the download url (include version directory) in an environment variable, `SPARKR_RELEASE_DOWNLOAD_URL`

## How was this patch tested?

unit test, manually testing
- snapshot build url
  - download when spark jar not cached
  - when spark jar is cached
- RC build url
  - download when spark jar not cached
  - when spark jar is cached
- multiple cached spark versions
- starting with sparkR shell

To use this,
```
SPARKR_RELEASE_DOWNLOAD_URL=http://this_is_the_url_to_spark_release_tgz R
```
then in R,
```
library(SparkR) # or specify lib.loc
sparkR.session()
```

### DIFF
--- a/R/pkg/R/install.R
+++ b/R/pkg/R/install.R
@@ -79,19 +79,28 @@ install.spark <- function(hadoopVersion = "2.7", mirrorUrl = NULL,
     dir.create(localDir, recursive = TRUE)
   }
 
-  packageLocalDir <- file.path(localDir, packageName)
-
   if (overwrite) {
     message(paste0("Overwrite = TRUE: download and overwrite the tar file",
                    "and Spark package directory if they exist."))
   }
 
+  releaseUrl <- Sys.getenv("SPARKR_RELEASE_DOWNLOAD_URL")
+  if (releaseUrl != "") {
+    packageName <- basenameSansExtFromUrl(releaseUrl)
+  }
+
+  packageLocalDir <- file.path(localDir, packageName)
+
   # can use dir.exists(packageLocalDir) under R 3.2.0 or later
   if (!is.na(file.info(packageLocalDir)$isdir) && !overwrite) {
-    fmt <- "%s for Hadoop %s found, with SPARK_HOME set to %s"
-    msg <- sprintf(fmt, version, ifelse(hadoopVersion == "without", "Free build", hadoopVersion),
-                   packageLocalDir)
-    message(msg)
+    if (releaseUrl != "") {
+      message(paste(packageName, "found, setting SPARK_HOME to", packageLocalDir))
+    } else {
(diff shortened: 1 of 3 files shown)


felixcheung:
Jenkins, retest this please

felixcheung:
appveyer passed before.

felixcheung:
@shivaram all tests passed.

R/pkg/R/install.R
```diff
@@ -104,7 +113,12 @@ install.spark <- function(hadoopVersion = "2.7", mirrorUrl = NULL,
   if (tarExists && !overwrite) {
     message("tar file found.")
   } else {
-    robustDownloadTar(mirrorUrl, version, hadoopVersion, packageName, packageLocalPath)
+    if (releaseUrl != "") {
+      message("Downloading from alternate URL:\n- ", releaseUrl)
+      downloadUrl(releaseUrl, packageLocalPath, paste0("Fetch failed from ", mirrorUrl))
```
shivaram: this should be `releaseUrl` instead of `mirrorUrl` in the `paste0` ?

R/pkg/R/utils.R
```diff
@@ -851,3 +851,12 @@ rbindRaws <- function(inputData){
   out[!rawcolumns] <- lapply(out[!rawcolumns], unlist)
   out
 }
+
+# Get basename without extension from URL
+basenameSansExtFromUrl <- function(url) {
```
shivaram: can we use file_path_sans_ext [1] for removing the extension ? I worry we might publish it as `.tar.gz` someday and then removing just the last `.` will be insufficient

[1] https://stat.ethz.ch/R-manual/R-patched/library/tools/html/fileutils.html

felixcheung: My concern was to bring in another dependencies just for this (it's in the tools)

The regex was in fact copy-paste from file_path_sans_ext (hence the name) except for the compression part which is what you are referring to.
I could copy that over as well. Would you prefer `compression` be TRUE (default is FALSE) to remove `.gz`?

```
> library(tools)
> file_path_sans_ext
function (x, compression = FALSE)
{
    if (compression)
        x <- sub("[.](gz|bz2|xz)$", "", x)
    sub("([^.]+)\\.[[:alnum:]]+$", "\\1", x)
}
```

shivaram: Got it - yeah it sounds good to remove compression and just copy the two lines of code.

shivaram (commented):
LGTM. Thanks @felixcheung - I'll just test this locally once and then merge

shivaram:
Tested locally - Works fine. Merging into master, branch-2.1

PR #16249, opened 2016-12-11
[SPARK-18828][SPARKR] Refactor scripts for R

felixcheung:

Refactored script to remove duplications and clearer purpose for each script

## How was this patch tested?

manually

### DIFF
--- a/R/install-source-package.sh
+++ b/R/install-source-package.sh
@@ -29,7 +29,7 @@
 set -o pipefail
 set -e
 
-FWDIR="$(cd `dirname $0`; pwd)"
+FWDIR="$(cd `dirname "${BASH_SOURCE[0]}"`; pwd)"
 pushd $FWDIR > /dev/null
 . $FWDIR/find-r.sh
 
--- a/R/install-source-package.sh
+++ b/R/install-source-package.sh
@@ -30,20 +30,22 @@ set -o pipefail
 set -e
 
 FWDIR="$(cd `dirname $0`; pwd)"
-LIB_DIR="$FWDIR/lib"
-
-mkdir -p $LIB_DIR
-
 pushd $FWDIR > /dev/null
 . $FWDIR/find-r.sh
 
-. $FWDIR/create-rd.sh
+if [ -z "$VERSION" ]; then
+  VERSION=`grep Version $FWDIR/pkg/DESCRIPTION | awk '{print $NF}'`
+fi
 
-# Install SparkR to $LIB_DIR
-"$R_SCRIPT_PATH/"R CMD INSTALL --library=$LIB_DIR $FWDIR/pkg/
+echo "Removing lib path and installing from source package"
+LIB_DIR="$FWDIR/lib"
+rm -rf $LIB_DIR
+mkdir -p $LIB_DIR
+"$R_SCRIPT_PATH/"R CMD INSTALL SparkR_"$VERSION".tar.gz --library=$LIB_DIR
 
 # Zip the SparkR package so that it can be distributed to worker nodes on YARN
-cd $LIB_DIR
+pushd $LIB_DIR > /dev/null
 jar cfM "$LIB_DIR/sparkr.zip" SparkR
-
 popd > /dev/null
+
+popd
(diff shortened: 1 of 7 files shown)


felixcheung:
flaky test?

felixcheung:
Jenkins, retest this please

shivaram:
Can we open a JIRA for this ? Its good to track this change

shivaram:
Just FYI, I'm holding off on this till 2.1 - I think its better to make this change after the release just to be safe

felixcheung:
jenkins, retest this please

felixcheung:
that's weird. but the last Jenkins run on commit 66fc83c passed.
how about we merge this @shivaram ?

shivaram:
Sure I'll take a final look today. Meanwhile

Jenkins, retest this please

felixcheung:
Jenkins, retest this please

felixcheung:
Jenkins, retest this please

felixcheung:
@shivaram how about it? ;)

shivaram:
Sorry for the delay - I was just trying this today. Looks like the `$FWDIR` is not right in `install-dev.sh` ?
```
# cd spark
# ./R/install-dev.sh
Using R_SCRIPT_PATH = /Library/Frameworks/R.framework/Versions/3.2/Resources/bin
/Users/shivaram/spark/R/create-rd.sh: line 34: /Users/shivaram/spark/R/R/find-r.sh: No such file or directory
```

felixcheung:
Hmm, I think it's the problem with running the first script from a relative directory.

./R/install-dev.sh

Vs

cd R
./install-dev.sh

Let me investigate.

felixcheung:
it turns out `$0` is not reliable (http://stackoverflow.com/questions/59895/getting-the-source-directory-of-a-bash-script-from-within). Updated and tested on Ubuntu and Mac.

R/install-source-package.sh
```diff
@@ -0,0 +1,51 @@
...
+
+if [ -z "$VERSION" ]; then
+  VERSION=`grep Version $FWDIR/pkg/DESCRIPTION | awk '{print $NF}'`
+fi
+
+echo "Removing lib path and installing from source package"
+LIB_DIR="$FWDIR/lib"
+rm -rf $LIB_DIR
+mkdir -p $LIB_DIR
+"$R_SCRIPT_PATH/"R CMD INSTALL SparkR_"$VERSION".tar.gz --library=$LIB_DIR
```
shivaram: Just so I understand correctly this script assumes that `SparkR_$VERSION.tar.gz` exists ? If so can we check that in the preceding lines ?

shivaram (commented):
Thanks for the fixes ! local checks work fine now - I just had one last comment

R/install-source-package.sh
felixcheung: great point - added.

dev/make-distribution.sh
```diff
@@ -232,14 +232,17 @@ if [ "$MAKE_R" == "true" ]; then
...
+  NO_TESTS=1 "$SPARK_HOME/"R/check-cran.sh
+
   # Move R source package to match the Spark release version if the versions are not the same.
   # NOTE(shivaram): `mv` throws an error on Linux if source and destination are same file
   if [ "$R_PACKAGE_VERSION" != "$VERSION" ]; then
     mv $SPARK_HOME/R/SparkR_"$R_PACKAGE_VERSION".tar.gz $SPARK_HOME/R/SparkR_"$VERSION".tar.gz
   fi
+
+  # Install source package to get it to generate vignettes rds files, etc.
+  VERSION=$VERSION "$SPARK_HOME/"R/install-source-package.sh
```
shivaram: One more thing I ran into while testing this locally - the file `install-source-package.sh` didn't have execute permission. I wonder if this due to the way I patched things or if we should set the executable bit while checking the file in ?

shivaram: Actually ignore this - If I use `git apply` it gets the executable bit. I was using `patch` before this.

shivaram:
LGTM

dev/make-distribution.sh
felixcheung: ah good.

felixcheung:
merging this to master.
AppVeyor seems to be stuck or running behind - since I have PR pending I'll check it in a couple of days.

felixcheung:
merged to master. Will this run in nightly for make-distribution.sh? if not, could you kick off Jenkins as some point? Thanks @shivaram

PR #16250, opened 2016-12-11
[CORE][MINOR] Stylistic changes in DAGScheduler (to ease comprehensio…

jaceklaskowski:

Stylistic changes in `DAGScheduler` to ease comprehension thereof.

## How was this patch tested?

Local build. Awaiting Jenkins to run the entire test suite.

…n thereof)

### DIFF
--- a/core/src/main/scala/org/apache/spark/scheduler/DAGScheduler.scala
+++ b/core/src/main/scala/org/apache/spark/scheduler/DAGScheduler.scala
@@ -333,16 +333,16 @@ class DAGScheduler(
       // (so we don't unnecessarily re-compute that data).
       val serLocs = mapOutputTracker.getSerializedMapOutputStatuses(shuffleDep.shuffleId)
       val locs = MapOutputTracker.deserializeMapStatuses(serLocs)
-      (0 until locs.length).foreach { i =>
-        if (locs(i) ne null) {
-          // locs(i) will be null if missing
-          stage.addOutputLoc(i, locs(i))
+      locs.zipWithIndex
+        .filter { case (status, _) => status != null } // null if missing
+        .foreach { case (ms, idx) =>
+          stage.addOutputLoc(idx, ms)
         }
-      }
     } else {
       // Kind of ugly: need to register RDDs with the cache and map output tracker here
       // since we can't do it in the RDD constructor because # of partitions is unknown
-      logInfo("Registering RDD " + rdd.id + " (" + rdd.getCreationSite + ")")
+      logInfo(s"Registering RDD ${rdd.id} (partitions: ${rdd.partitions.length}, " +
+        s" ${rdd.getCreationSite})")
       mapOutputTracker.registerShuffle(shuffleDep.shuffleId, rdd.partitions.length)
     }
     stage
@@ -1137,12 +1137,12 @@ class DAGScheduler(
     listenerBus.post(SparkListenerTaskEnd(
        stageId, task.stageAttemptId, taskType, event.reason, event.taskInfo, taskMetrics))
 
-    if (!stageIdToStage.contains(task.stageId)) {
+    if (!stageIdToStage.contains(stageId)) {
       // Skip all the actions if the stage has been cancelled.
       return
     }
 
-    val stage = stageIdToStage(task.stageId)
+    val stage = stageIdToStage(stageId)
     event.reason match {
       case Success =>
         stage.pendingPartitions -= task.partitionId
@@ -1176,7 +1176,7 @@ class DAGScheduler(
                   }
                 }
               case None =>
-                logInfo("Ignoring result from " + rt + " because its job has finished")
+                logInfo(s"Ignoring result from $rt because its job has finished")
             }
 
           case smt: ShuffleMapTask =>
@@ -1184,7 +1184,7 @@ class DAGScheduler(
             updateAccumulators(event)
             val status = event.result.asInstanceOf[MapStatus]
             val execId = status.location.executorId
-            logDebug("ShuffleMapTask finished on " + execId)
+            logDebug(s"ShuffleMapTask finished on $execId")
             if (failedEpoch.contains(execId) && smt.epoch <= failedEpoch(execId)) {
               logInfo(s"Ignoring possibly bogus $smt completion from executor $execId")
             } else {
@@ -1211,14 +1211,7 @@ class DAGScheduler(
 
               clearCacheLocs()
 
-              if (!shuffleStage.isAvailable) {
-                // Some tasks had failed; let's resubmit this shuffleStage
-                // TODO: Lower-level scheduler should also deal with this
-                logInfo("Resubmitting " + shuffleStage + " (" + shuffleStage.name +
-                  ") because some of its tasks had failed: " +
-                  shuffleStage.findMissingPartitions().mkString(", "))
-                submitStage(shuffleStage)
-              } else {
+              if (shuffleStage.isAvailable) {
                 // Mark any map-stage jobs waiting on this stage as finished
                 if (shuffleStage.mapStageJobs.nonEmpty) {
                   val stats = mapOutputTracker.getStatistics(shuffleStage.shuffleDep)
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/apache/spark/scheduler/DAGScheduler.scala
```diff
@@ -333,16 +333,16 @@ class DAGScheduler(
       // (so we don't unnecessarily re-compute that data).
       val serLocs = mapOutputTracker.getSerializedMapOutputStatuses(shuffleDep.shuffleId)
       val locs = MapOutputTracker.deserializeMapStatuses(serLocs)
-      (0 until locs.length).foreach { i =>
-        if (locs(i) ne null) {
-          // locs(i) will be null if missing
-          stage.addOutputLoc(i, locs(i))
+      locs.zipWithIndex
+        .filter { case (status, _) => status != null } // null if missing
+        .foreach { case (ms, idx) =>
```
srowen: Use ms or status consistency?

core/src/main/scala/org/apache/spark/scheduler/DAGScheduler.scala
```diff
@@ -333,16 +333,16 @@ class DAGScheduler(
...
+          stage.addOutputLoc(idx, ms)
         }
-      }
     } else {
       // Kind of ugly: need to register RDDs with the cache and map output tracker here
       // since we can't do it in the RDD constructor because # of partitions is unknown
-      logInfo("Registering RDD " + rdd.id + " (" + rdd.getCreationSite + ")")
+      logInfo(s"Registering RDD ${rdd.id} (partitions: ${rdd.partitions.length}, " +
```
srowen: No big deal but this is changing the log message too

core/src/main/scala/org/apache/spark/scheduler/DAGScheduler.scala
```diff
@@ -1227,12 +1220,19 @@ class DAGScheduler(
                   }
                 }
                 submitWaitingChildStages(shuffleStage)
+              } else {
+                // Some tasks had failed; let's resubmit this shuffleStage
+                // TODO: Lower-level scheduler should also deal with this
+                logInfo(s"Resubmitting $shuffleStage (${shuffleStage.name}) " +
+                  "because some of its tasks had failed: " +
+                  shuffleStage.findMissingPartitions().mkString(", "))
```
srowen: If you bother with this, could go all the way and pull this out as a val

srowen (commented):
Certainly, I like the code you propose more. I am still not clear if we want to encourage touch-ups that don't really have any impact on performance or readability. I am neutral on whether this is worth it.

core/src/main/scala/org/apache/spark/scheduler/DAGScheduler.scala
jaceklaskowski: @srowen That's what I was going to do, but didn't know how much would get approved :) Thanks for warm welcome.

core/src/main/scala/org/apache/spark/scheduler/DAGScheduler.scala
jaceklaskowski: Yes, but it was because I think it's important at this "stage". Do you think I should remove the partition-related addition?

jaceklaskowski:
Thanks @srowen for the review! I do understand your point and remember you and @rxin have always been telling me that I should not touch code unless there's a need for a change.

But the more I'm with the code the more I think there'd be more contributions if the code were even slightly more readable. I'm still having troubles getting the gist of it (but am closer).

Spark in general is very tricky to get the hang of and when the code is unnecessarily complex (like doing traversing, filtering and branching in a convoluted way) the more functional bits could certainly help the code, me and the community. That's my hope.

That's also why I'm sending very small changes to get myself going with more ease with the code (and get more comfortable with what's acceptable). All in all, I don't think I'm ready for heavier contributions yet, but I do think I'm closer and will soon be. Your help and patience have helped a lot. Thanks (and don't worry if you have to reject my changes).

core/src/main/scala/org/apache/spark/scheduler/DAGScheduler.scala
```diff
@@ -333,16 +333,16 @@ class DAGScheduler(
       // (so we don't unnecessarily re-compute that data).
       val serLocs = mapOutputTracker.getSerializedMapOutputStatuses(shuffleDep.shuffleId)
       val locs = MapOutputTracker.deserializeMapStatuses(serLocs)
-      (0 until locs.length).foreach { i =>
-        if (locs(i) ne null) {
-          // locs(i) will be null if missing
-          stage.addOutputLoc(i, locs(i))
+      locs.zipWithIndex
```
rxin: just fyi this is going to be slower than the original code. I don't know if it matters, but it will be slower.

srowen:
I don't think these superficial changes meaningfully affect readability at all, and I don't perceive that this is any significant barrier to anyone modifying the code anyway. It's both an argument that to do it -- hey, why not polish things a bit? -- and not do it -- not going to really help so why bother at all.

There is a tiny risk of breaking something or regressing performance. There is the very small problem of merge conflicts.

Mostly, it's a meta-issue: reviewer time. Remember it's not just me that reads this, but a thousand people for at least a little bit of time. It probably costs >$1000 of person-hours to get this change merged. 

I think we're often happy to take trivial contributions from a new person to encourage new participants, but, do look to more experienced folks to spend time on reviews, fixing deeper issues, fixing broader issues.

We can keep saying this stuff and merging anyway but I have not seen evidence that the message is getting across. This is why you may see at this point we're asking you to not make changes like this and just close it.

None of these changes look essential even if I would have written it your way in general, so I'd close this.

rxin:
+1 to what @srowen said.

core/src/main/scala/org/apache/spark/scheduler/DAGScheduler.scala
JoshRosen: I'd be wary of making this type of change since there are cases where we've purposely removed Scala-style transformations in hot scheduler loops.

shivaram (commented):
+1 on what @srowen said - Also I don't think the improvement in readability is worth the risk in terms of performance regression -- esp. in parts of the DAGScheduler

JoshRosen:
I agree with above comments and would be +1 to not performing further review on this set of cosmetic changes. The reason why the DAGScheduler is complicated isn't due to surface-level code issues but due to design complexity and just putting a fresh coat of paint on it via small-scale changes eats up reviewer time, makes it harder to backport fixes without merge conflicts, makes it harder to hunt for regressions since there are more diffs to bisect, etc.

rxin:
@jaceklaskowski please close it.

jaceklaskowski:
Just to have the list of the reasons not to accept the changes:

1. @rxin "just fyi this is going to be slower than the original code."
2. @shivaram "Its hard to profile these things closely enough to figure out how much overhead it adds"
3. @srowen "There is a tiny risk of breaking something or regressing performance."
4. @shivaram "I don't think the improvement in readability is worth the risk in terms of performance regression"

I however think the points above are the exact reasons why you could accept this and the following changes to DAGScheduler. We should *not* be afraid of making changes to the core. We've got no evidence the points above hold and that's what worries me the most.

Thanks for your help anyway! It was worth to have sent the PR to learn from you!

rxin:
You missed the fact that people don't even think this is a big improvement in usability.

srowen:
The question is not whether there are downsides; there almost always some. It's whether the upsides are bigger. We typically can't evaluate either of them perfectly, but can take good guesses. The problem here isn't that downsides are so clear or large, but that there is almost no upside. We change core all the time, that's not the issue.

This is part of a meta-issue of overhead of processing the change. It's not just whether a change is a tiny net win but whether, in the time needed to process it, we could have done something else that's a bigger win.

(This extended aside is worthwhile because I think the overhead issue is rarely appreciated; it is a non-issue on medium-sized projects but bites at scale. Most projects would happily accept any change at all; this project is way on the other end of that spectrum. It can be surprising.)

PR #16251, opened 2016-12-12
[SPARK-18826][SS]Add 'latestFirst' option to FileStreamSource

zsxwing:

When starting a stream with a lot of backfill and maxFilesPerTrigger, the user could often want to start with most recent files first. This would let you keep low latency for recent data and slowly backfill historical data.

This PR adds a new option `latestFirst` to control this behavior. When it's true, `FileStreamSource` will sort the files by the modified time from latest to oldest, and take the first `maxFilesPerTrigger` files as a new batch.

## How was this patch tested?

The added test.

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/FileStreamOptions.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/FileStreamOptions.scala
@@ -53,4 +53,15 @@ class FileStreamOptions(parameters: CaseInsensitiveMap) extends Logging {
   /** Options as specified by the user, in a case-insensitive map, without "path" set. */
   val optionMapWithoutPath: Map[String, String] =
     parameters.filterKeys(_ != "path")
+
+  /** Whether to scan new files first. */
+  val newestFirst: Boolean = parameters.get("newestFirst").map { str =>
+    try {
+      str.toBoolean
+    } catch {
+      case _: IllegalArgumentException =>
+        throw new IllegalArgumentException(
+          s"Invalid value '$str' for option 'newestFirst', must be 'true' or 'false'")
+    }
+  }.getOrElse(false)
 }
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/FileStreamOptions.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/FileStreamOptions.scala
@@ -54,14 +54,17 @@ class FileStreamOptions(parameters: CaseInsensitiveMap) extends Logging {
   val optionMapWithoutPath: Map[String, String] =
     parameters.filterKeys(_ != "path")
 
-  /** Whether to scan new files first. */
-  val newestFirst: Boolean = parameters.get("newestFirst").map { str =>
+  /**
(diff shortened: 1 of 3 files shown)


sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/FileStreamOptions.scala
```diff
@@ -53,4 +53,15 @@ class FileStreamOptions(parameters: CaseInsensitiveMap) extends Logging {
   /** Options as specified by the user, in a case-insensitive map, without "path" set. */
   val optionMapWithoutPath: Map[String, String] =
     parameters.filterKeys(_ != "path")
+
+  /** Whether to scan new files first. */
```
tdas: Can you elaborate this comment further. In a trigger when it finds unprocessed files, it will first process the latest file.

Also, isnt latest more common than newest?

zsxwing:
retest this please

sql/core/src/test/scala/org/apache/spark/sql/streaming/FileStreamSourceSuite.scala
```diff
@@ -1059,6 +1060,72 @@ class FileStreamSourceSuite extends FileStreamSourceTest {
...
+
+      // Read oldest files first, so the first batch is "1", and the second batch is "2".
+      val fileStream = createFileStream(
+        "text",
+        src.getCanonicalPath,
+        options = Map("latestFirst" -> "false", "maxFilesPerTrigger" -> "1"))
+      val clock = new StreamManualClock()
+      testStream(fileStream)(
+        StartStream(trigger = ProcessingTime(10), triggerClock = clock),
+        AssertOnQuery { _ =>
```
tdas: why do you need to wait on the manual clock? CheckLastBatch will automatically wait for the batch to complete?

sql/core/src/test/scala/org/apache/spark/sql/streaming/FileStreamSourceSuite.scala
```diff
@@ -1059,6 +1060,72 @@ class FileStreamSourceSuite extends FileStreamSourceTest {
...
+          eventually(timeout(streamingTimeout)) {
+            assert(clock.isStreamWaitingAt(10))
+          }
+          true
+        },
+        CheckLastBatch("2")
+      )
+
+      // Read latest files first, so the first batch is "2", and the second batch is "1".
+      val fileStream2 = createFileStream(
```
tdas: I think this code can be deduped by writing a function that make the query run two batches and collect the results in order. And then the function is called with latestFirst true or false, and the result order checked.

sql/core/src/test/scala/org/apache/spark/sql/streaming/FileStreamSourceSuite.scala
zsxwing: > why do you need to wait on the manual clock? CheckLastBatch will automatically wait for the batch to complete?

`CheckLastBatch` waits only when `AddData` is used, but in this test, I need to add data before starting the query.

tdas:
LGTM, pending tests.

tdas:
Merging to master and 2.1

PR #16255, opened 2016-12-12
[SPARK-18609][SQL]Fix when CTE with Join between two table with same column name

windpiger:
   After optimize the SQL with CTE and Join between two tables with the same column name,  there will throw a exception when the Physical Plan executed.
SQL:
```
>spark.sql("CREATE TABLE p1 (col STRING)"
>spark.sql("CREATE TABLE p2 (col STRING) "
>spark.sql("set spark.sql.crossJoin.enabled = true")
>spark.sql("WITH CTE AS (SELECT s2.col as col FROM p1 CROSS JOIN (SELECT e.col as col FROM p2 E) s2) SELECT T1.col as c1, T2.col as c2 FROM CTE T1 CROSS JOIN CTE T2")
```

before fixed the SQL explained:
```
== Physical Plan ==
*Project [col#123 AS c1#104, col#126 AS c2#105]
+- CartesianProduct
   :- CartesianProduct
   :  :- HiveTableScan MetastoreRelation default, p1
   :  +- HiveTableScan [col#123], MetastoreRelation default, p2, E
   +- *!Project [col#123 AS col#126]
      +- CartesianProduct
         :- HiveTableScan MetastoreRelation default, p1
         +- HiveTableScan MetastoreRelation default, p2, E
```
when execute the Physical Plan above, it will throw exception because of there is no attribute for col#123 to bind( the rules ColumnPruning/RemoveAliasOnlyProject are the root cause):
`
org.apache.spark.sql.catalyst.errors.package$TreeNodeException: Binding attribute, tree: col#123
	at org.apache.spark.sql.catalyst.errors.package$.attachTree(package.scala:56)
	at org.apache.spark.sql.catalyst.expressions.BindReferences$$anonfun$bindReference$1.applyOrElse(BoundAttribute.scala:88)
	at org.apache.spark.sql.catalyst.expressions.BindReferences$$anonfun$bindReference$1.applyOrElse(BoundAttribute.scala:87)
	at org.apache.spark.sql.catalyst.trees.TreeNode$$anonfun$3.apply(TreeNode.scala:286)
	at org.apache.spark.sql.catalyst.trees.TreeNode$$anonfun$3.apply(TreeNode.scala:286)
	at org.apache.spark.sql.catalyst.trees.CurrentOrigin$.withOrigin(TreeNode.scala:69)
	at org.apache.spark.sql.catalyst.trees.TreeNode.transformDown(TreeNode.scala:285)
	at org.apache.spark.sql.catalyst.trees.TreeNode$$anonfun$transformDown$1.apply(TreeNode.scala:291)
	at org.apache.spark.sql.catalyst.trees.TreeNode$$anonfun$transformDown$1.apply(TreeNode.scala:291)
	at org.apache.spark.sql.catalyst.trees.TreeNode$$anonfun$5.apply(TreeNode.scala:328)
	at org.apache.spark.sql.catalyst.trees.TreeNode.mapProductIterator(TreeNode.scala:186)
	at org.apache.spark.sql.catalyst.trees.TreeNode.transformChildren(TreeNode.scala:326)
	at org.apache.spark.sql.catalyst.trees.TreeNode.transformDown(TreeNode.scala:291)
	at org.apache.spark.sql.catalyst.trees.TreeNode.transform(TreeNode.scala:275)
	at org.apache.spark.sql.catalyst.expressions.BindReferences$.bindReference(BoundAttribute.scala:87)
	at org.apache.spark.sql.execution.ProjectExec$$anonfun$4.apply(basicPhysicalOperators.scala:61)
	at org.apache.spark.sql.execution.ProjectExec$$anonfun$4.apply(basicPhysicalOperators.scala:60)
` 

after fixed the SQL explained:
```
== Physical Plan ==
*Project [col#12 AS c1#0, col#12 AS c2#1]
+- CartesianProduct
   :- CartesianProduct
   :  :- HiveTableScan MetastoreRelation default, p1
   :  +- HiveTableScan [col#12], MetastoreRelation default, p2
   +- CartesianProduct
      :- HiveTableScan MetastoreRelation default, p1
      +- HiveTableScan [col#12], MetastoreRelation default, p2
```
and the result is 
```
+---+---+
| c1| c2|
+---+---+
+---+---+
```

## How was this patch tested?
unit test added

### DIFF
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/Optimizer.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/Optimizer.scala
@@ -418,8 +418,8 @@ object ColumnPruning extends Rule[LogicalPlan] {
     case w: Window if w.windowExpressions.isEmpty => w.child
 
     // Eliminate no-op Projects
-    case p @ Project(_, child) if sameOutput(child.output, p.output) =>
-      if (child.isInstanceOf[CatalogRelation]) p else child
+    case p @ Project(_, child) if sameOutput(child.output, p.output) => child
+
     // Can't prune the columns on LeafNode
     case p @ Project(_, _: LeafNode) => p
 
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/Optimizer.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/Optimizer.scala
@@ -26,7 +26,7 @@ import org.apache.spark.api.java.function.FilterFunction
 import org.apache.spark.sql.AnalysisException
 import org.apache.spark.sql.catalyst.{CatalystConf, SimpleCatalystConf}
 import org.apache.spark.sql.catalyst.analysis._
-import org.apache.spark.sql.catalyst.catalog.{InMemoryCatalog, SessionCatalog}
+import org.apache.spark.sql.catalyst.catalog.{CatalogRelation, InMemoryCatalog, SessionCatalog}
 import org.apache.spark.sql.catalyst.expressions._
 import org.apache.spark.sql.catalyst.expressions.aggregate._
 import org.apache.spark.sql.catalyst.expressions.Literal.{FalseLiteral, TrueLiteral}
@@ -200,6 +200,7 @@ object RemoveAliasOnlyProject extends Rule[LogicalPlan] {
         case plan: Project if plan eq proj => plan.child
         case plan => plan transformExpressions {
           case a: Attribute if attrMap.contains(a) => attrMap(a)
+          case b: Alias if (attrMap.exists(_._1.exprId == b.exprId)) => b.child
         }
       }
     }.getOrElse(plan)
@@ -416,8 +417,8 @@ object ColumnPruning extends Rule[LogicalPlan] {
     case w: Window if w.windowExpressions.isEmpty => w.child
 
     // Eliminate no-op Projects
-    case p @ Project(_, child) if sameOutput(child.output, p.output) => child
-
+    case p @ Project(_, child) if sameOutput(child.output, p.output) =>
+      if (child.isInstanceOf[CatalogRelation]) p else child
     // Can't prune the columns on LeafNode
     case p @ Project(_, _: LeafNode) => p
 
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/Optimizer.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/Optimizer.scala
@@ -200,7 +200,8 @@ object RemoveAliasOnlyProject extends Rule[LogicalPlan] {
         case plan: Project if plan eq proj => plan.child
         case plan => plan transformExpressions {
           case a: Attribute if attrMap.contains(a) => attrMap(a)
-          case b: Alias if (attrMap.exists(_._1.exprId == b.exprId)) => b.child
+          case b: Alias if attrMap.exists(_._1.exprId == b.exprId)
+            && b.child.isInstanceOf[NamedExpression] => b.child
         }
       }
     }.getOrElse(plan)
--- a/sql/hive/src/test/scala/org/apache/spark/sql/hive/execution/SQLQuerySuite.scala
+++ b/sql/hive/src/test/scala/org/apache/spark/sql/hive/execution/SQLQuerySuite.scala
@@ -2011,6 +2011,22 @@ class SQLQuerySuite extends QueryTest with SQLTestUtils with TestHiveSingleton {
     }
   }
 
+  test("test CTE with join between two table with the same column name ") {
+    sql("DROP TABLE IF EXISTS p1")
+    sql("DROP TABLE IF EXISTS p2")
+    sql("CREATE TABLE p1 (col String)" )
+    sql("CREATE TABLE p2 (col String)")
+
+    assert(
+      sql(
+        """
+          | WITH CTE AS
+          |  (SELECT s2.col as col FROM p1
+          |     CROSS JOIN (SELECT e.col as col FROM p2 E) s2)
+          | SELECT T1.col as c1,T2.col as c2 FROM CTE T1 CROSS JOIN CTE T2
+        """.stripMargin).collect.isEmpty)
+  }
+
   def testCommandAvailable(command: String): Boolean = {
     val attempt = Try(Process(command).run(ProcessLogger(_ => ())).exitValue())
     attempt.isSuccess && attempt.get == 0


hvanhovell:
@windpiger I am not sure that this is an analysis problem. I suspect that optimizer is messing up the query. Could you pinpoint the rule that causes the issue? Tip: in the spark shell (`./build/sbt sparkShell`) use `sc.setLogLevel("TRACE")`, and that will print all rules being applied.

windpiger:
ok, the optimizer rule apply as follows:

1. before optimizer:

> Project [col#7 AS c1#4, col#10 AS c2#5]
+- Join Cross
   :- Project [col#6 AS col#7]
   :  +- Join Cross
   :     :- MetastoreRelation default, p1
   :     +- Project [col#9 AS col#6]
   :        +- MetastoreRelation default, p2
   +- Project [col#6 AS col#10]
      +- Join Cross
         :- MetastoreRelation default, p1
         +- Project [col#9 AS col#6]
            +- MetastoreRelation default, p2


2. after first ColumnPruning: 
>Project [col#7 AS c1#4, col#10 AS c2#5]
+- Join Cross
   :- Project [col#6 AS col#7]
   :  +- Join Cross
   :     :- Project
   :     :  +- MetastoreRelation default, p1
   :     +- Project [col#9 AS col#6]
   :        +- MetastoreRelation default, p2
   +- Project [col#6 AS col#10]
      +- Join Cross
         :- Project
         :  +- MetastoreRelation default, p1
         +- Project [col#9 AS col#6]
            +- MetastoreRelation default, p2

3. after first RemoveAliasOnlyProject:
>Project [col#6 AS c1#4, col#10 AS c2#5]
+- Join Cross
   :- Join Cross
   :  :- Project
   :  :  +- MetastoreRelation default, p1
   :  +- Project [col#9 AS col#6]
   :     +- MetastoreRelation default, p2
   +- Project [col#6 AS col#10]
      +- Join Cross
         :- Project
         :  +- MetastoreRelation default, p1
         +- Project [col#9 AS col#6]
            +- MetastoreRelation default, p2
`

4. after second ColumnPruning:
>Project [col#6 AS c1#4, col#10 AS c2#5]
+- Join Cross
   :- Join Cross
   :  :- Project
   :  :  +- MetastoreRelation default, p1
   :  +- Project [col#9 AS col#6]
   :     +- MetastoreRelation default, p2
   +- Project [col#6 AS col#10]
      +- Join Cross
         :- Project
         :  +- MetastoreRelation default, p1
         +- Project [col#9 AS col#6]
            +- MetastoreRelation default, p2

5. after second RemoveAliasOnlyProject:
>Project [col#9 AS c1#4, col#10 AS c2#5]
+- Join Cross
   :- Join Cross
   :  :- Project
   :  :  +- MetastoreRelation default, p1
   :  +- MetastoreRelation default, p2
   +- !Project [col#9 AS col#10]
      +- Join Cross
         :- Project
         :  +- MetastoreRelation default, p1
         +- Project [col#9 AS col#6]
            +- MetastoreRelation default, p2

6. after third ColumnPruning:
>Project [col#9 AS c1#4, col#10 AS c2#5]
+- Join Cross
   :- Join Cross
   :  :- Project
   :  :  +- MetastoreRelation default, p1
   :  +- MetastoreRelation default, p2
   +- !Project [col#9 AS col#10]
      +- Join Cross
         :- Project
         :  +- MetastoreRelation default, p1
         +- Project
            +- Project
               +- MetastoreRelation default, p2

**here**
>   +- !Project [col#9 AS col#10]
      +- Join Cross
         :- Project
         :  +- MetastoreRelation default, p1
         +- Project
            +- Project
               +- MetastoreRelation default, p2

**col#9 has no attribute to bind when do BindRefference**

7. after CollapseProject:
>Project [col#9 AS c1#4, col#10 AS c2#5]
+- Join Cross
   :- Join Cross
   :  :- Project
   :  :  +- MetastoreRelation default, p1
   :  +- MetastoreRelation default, p2
   +- !Project [col#9 AS col#10]
      +- Join Cross
         :- Project
         :  +- MetastoreRelation default, p1
         +- Project
            +- MetastoreRelation default, p2

8. when optimizer finished, the logic plan is 
>Project [col#9 AS c1#4, col#10 AS c2#5]
+- Join Cross
   :- Join Cross
   :  :- Project
   :  :  +- MetastoreRelation default, p1
   :  +- MetastoreRelation default, p2
   +- !Project [col#9 AS col#10]
      +- Join Cross
         :- Project
         :  +- MetastoreRelation default, p1
         +- Project
            +- MetastoreRelation default, p2
`

windpiger:
@hvanhovell please help to review this, thanks a lot!

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/Optimizer.scala
```diff
@@ -200,6 +200,8 @@ object RemoveAliasOnlyProject extends Rule[LogicalPlan] {
         case plan: Project if plan eq proj => plan.child
         case plan => plan transformExpressions {
           case a: Attribute if attrMap.contains(a) => attrMap(a)
+          case b: Alias if attrMap.exists(_._1.exprId == b.exprId)
```
hvanhovell: It looks like semanticEquals might be broken (this should be, but perhaps this is for a reason. @cloud-fan any idea?

Please use a proper pattern match here.

hvanhovell: Nevermind.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/Optimizer.scala
```diff
@@ -416,8 +418,8 @@ object ColumnPruning extends Rule[LogicalPlan] {
     case w: Window if w.windowExpressions.isEmpty => w.child
 
     // Eliminate no-op Projects
-    case p @ Project(_, child) if sameOutput(child.output, p.output) => child
-
+    case p @ Project(_, child) if sameOutput(child.output, p.output) =>
+      if (child.isInstanceOf[CatalogRelation]) p else child
```
hvanhovell: This works in your example case, but will fail for a more complex query.

The problem is that the transform on L199 also modifies parts of the tree, that are totally unrelated to the projection at hand. In this case right side of the join. This can currently only happen in case of a self-join on a CTE, such a tree can contain duplicate attribute ids because we expand the tree after we have analyzed the CTE.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/Optimizer.scala
```diff
@@ -200,6 +200,8 @@ object RemoveAliasOnlyProject extends Rule[LogicalPlan] {
         case plan: Project if plan eq proj => plan.child
```
cloud-fan: a naive way to do this is, make sure we only replace attributes in the ancestor nodes of the alias-only project:
```
plan transform {
  case plan: Project if plan eq proj => plan.child
  case plan if plan.collect { case p if p eq project }.nonEmpty => // do the replace
}
```

It's very inefficient, maybe we can improve `TreeNode` to maintain the parent-child relationship between nodes.

windpiger: it is safe to only replace attributes in the ancestor nodes.
Alias with the same exprId but not the same object, replace the alias with it's child. it is not safe ,right?

cloud-fan: what do you mean?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/Optimizer.scala
```diff
@@ -200,6 +200,8 @@ object RemoveAliasOnlyProject extends Rule[LogicalPlan] {
...
         case plan => plan transformExpressions {
           case a: Attribute if attrMap.contains(a) => attrMap(a)
+          case b: Alias if attrMap.exists(_._1.exprId == b.exprId)
+            && b.child.isInstanceOf[NamedExpression] => b.child
```
cloud-fan: How do you reason about this? why we treat `Alias` differently here?

cloud-fan: How do you reason about this? why we treat `Alias` differently here?

windpiger: As you said, if we find an alias-only project, e.g. alias a#1 to a#2, it's safe to `remove this project` and `replace all a#2 with a#1` in this plan. So another Alias which is also alias a#1 to a#2, but not the same object with the first one, it will not be processed. 

here, the logic shows that we process the situation.

cloud-fan: I don't get it, for `alias a#1 to a#2`, we wanna `replace all a#2 with a#1`, so we will do nothing for `alias a#1 to a#2`, because we can't find an attribute `a#2`

windpiger: RemoveAliasOnlyProject will remove  `alias a#1 to a#2`, and `replace all a#2 with a#1`, so there is no `a#2` exists, If we do nothing for `alias a#1 to a#2`(not the same object with the removed one), it will cause the exception situation from `step 5` to `step 6` showed on the above comment.

cloud-fan: I know how the failure happens and this can fix it, but I think it's too hacky and does not catch the root cause.  https://github.com/apache/spark/pull/16255/files#r92348878 easily explains why the failure happens and how to fix it, can you make other people understand your fix easily?

PR #14382, opened 2016-07-27
[SPARK-16673] [Web UI] New Executor Page removed conditional for Logs and Thread Dump columns

ajbozarth:
## What changes were proposed in this pull request?

When #13670 switched `ExecutorsPage` to use JQuery DataTables it incidentally removed the conditional for the Logs and Thread Dump columns. I reimplemented the conditional display of the Logs and Thread dump columns as it was before the switch.
## How was this patch tested?

Manually tested and dev/run-tests

both
dump
logs

PR #16256, opened 2016-12-12
[SPARK-18816] [Web UI] Executors Logs column only ran visibility check on initial table load

ajbozarth:

When I added a visibility check for the logs column on the executors page in #14382 the method I used only ran the check on the initial DataTable creation and not subsequent page loads. I moved the check out of the table definition and instead it runs on each page load. The jQuery DataTable functionality used is the same.

## How was this patch tested?

Tested Manually

No visible UI changes to screenshot.

### DIFF
--- a/core/src/main/resources/org/apache/spark/ui/static/executorspage.js
+++ b/core/src/main/resources/org/apache/spark/ui/static/executorspage.js
@@ -412,10 +412,6 @@ $(document).ready(function () {
                     ],
                     "columnDefs": [
                         {
-                            "targets": [ 15 ],
-                            "visible": logsExist(response)
-                        },
-                        {
                             "targets": [ 16 ],
                             "visible": getThreadDumpEnabled()
                         }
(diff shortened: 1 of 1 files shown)


ajbozarth:
@srowen and @tgravescs could you take a quick look at this fix for a bug @yhuai found

ajbozarth:
Also in case the others are on vacation, @vanzin want to take a look too?

core/src/main/resources/org/apache/spark/ui/static/executorspage.js
```diff
@@ -412,18 +412,15 @@ $(document).ready(function () {
...
-                        {
                             "targets": [ 16 ],
                             "visible": getThreadDumpEnabled()
                         }
                     ],
                     "order": [[0, "asc"]]
                 };
     
-                $(selector).DataTable(conf);
+                var dt = $(selector).DataTable(conf);
```
srowen: You know better than I how to implement this. Does the same logic apply to column 16's init above?

ajbozarth: No, column 16 logics is based on a conf flag, 15 is based on a function.

srowen:
Merged to master

PR #16258, opened 2016-12-12
[SPARK-18834][SS] Expose event time stats through StreamingQueryProgress

tdas:

- Changed `StreamingQueryProgress.watermark` to `StreamingQueryProgress.queryTimestamps` which is a `Map[String, String]` containing the following keys: "eventTime.max", "eventTime.min", "eventTime.avg", "processingTime", "watermark". All of them UTC formatted strings.

- Renamed `StreamingQuery.timestamp` to `StreamingQueryProgress.triggerTimestamp` to differentiate from `queryTimestamps`. It has the timestamp of when the trigger was started. 

## How was this patch tested?

Updated tests

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/EventTimeWatermarkExec.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/EventTimeWatermarkExec.scala
@@ -17,8 +17,6 @@
 
 package org.apache.spark.sql.execution.streaming
 
-import scala.math.max
-
 import org.apache.spark.rdd.RDD
 import org.apache.spark.sql.catalyst.InternalRow
 import org.apache.spark.sql.catalyst.expressions.{Attribute, UnsafeProjection}
@@ -28,24 +26,48 @@ import org.apache.spark.sql.types.MetadataBuilder
 import org.apache.spark.unsafe.types.CalendarInterval
 import org.apache.spark.util.AccumulatorV2
 
-/** Tracks the maximum positive long seen. */
-class MaxLong(protected var currentValue: Long = 0)
-  extends AccumulatorV2[Long, Long] {
+/** Class for collecting event time stats with an accumulator */
+case class EventTimeStats(var max: Long, var min: Long, var sum: Long, var count: Long) {
+  def add(eventTime: Long): Unit = {
+    this.max = math.max(this.max, eventTime)
+    this.min = math.min(this.min, eventTime)
+    this.sum += eventTime
+    this.count += 1
+  }
 
-  override def isZero: Boolean = value == 0
-  override def value: Long = currentValue
-  override def copy(): AccumulatorV2[Long, Long] = new MaxLong(currentValue)
+  def merge(that: EventTimeStats): Unit = {
+    this.max = math.max(this.max, that.max)
+    this.min = math.min(this.min, that.min)
+    this.sum += that.sum
+    this.count += that.count
+  }
+
+  def avg: Long = (sum.toDouble / count).toLong
+}
+
+object EventTimeStats {
+  def zero: EventTimeStats = EventTimeStats(
+    max = Long.MinValue, min = Long.MaxValue, sum = 0L, count = 0L)
+}
+
+/** Accumulator that collects stats on event time in a batch. */
+class EventTimeStatsAccum(protected var currentStats: EventTimeStats = EventTimeStats.zero)
+  extends AccumulatorV2[Long, EventTimeStats] {
+
+  override def isZero: Boolean = value == EventTimeStats.zero
+  override def value: EventTimeStats = currentStats
+  override def copy(): AccumulatorV2[Long, EventTimeStats] = new EventTimeStatsAccum(currentStats)
 
   override def reset(): Unit = {
-    currentValue = 0
+    currentStats = EventTimeStats.zero
   }
 
   override def add(v: Long): Unit = {
-    currentValue = max(v, value)
+    currentStats.add(v)
   }
 
-  override def merge(other: AccumulatorV2[Long, Long]): Unit = {
-    currentValue = max(value, other.value)
+  override def merge(other: AccumulatorV2[Long, EventTimeStats]): Unit = {
+    currentStats.merge(other.value)
   }
 }
 
@@ -61,15 +83,14 @@ case class EventTimeWatermarkExec(
     delay: CalendarInterval,
     child: SparkPlan) extends SparkPlan {
 
-  // TODO: Use Spark SQL Metrics?
-  val maxEventTime = new MaxLong
-  sparkContext.register(maxEventTime)
+  val eventTimeStats = new EventTimeStatsAccum()
+  sparkContext.register(eventTimeStats)
 
   override protected def doExecute(): RDD[InternalRow] = {
     child.execute().mapPartitions { iter =>
       val getEventTime = UnsafeProjection.create(eventTime :: Nil, child.output)
       iter.map { row =>
-        maxEventTime.add(getEventTime(row).getLong(0))
+        eventTimeStats.add((getEventTime(row).getLong(0).toDouble / 1000).toLong)
         row
       }
     }
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/EventTimeWatermarkExec.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/EventTimeWatermarkExec.scala
@@ -42,7 +42,7 @@ case class EventTimeStats(var max: Long, var min: Long, var sum: Long, var count
     this.count += that.count
   }
 
-  def avg: Long = (sum.toDouble / count).toLong
+  def avg: Long = sum / count
 }
 
 object EventTimeStats {
@@ -76,7 +76,7 @@ class EventTimeStatsAccum(protected var currentStats: EventTimeStats = EventTime
  * adding appropriate metadata to this column, this operator also tracks the maximum observed event
  * time. Based on the maximum observed time and a user specified delay, we can calculate the
  * `watermark` after which we assume we will no longer see late records for a particular time
- * period.
+ * period. Note that event time is measured in milliseconds.
  */
 case class EventTimeWatermarkExec(
     eventTime: Attribute,
@@ -90,7 +90,7 @@ case class EventTimeWatermarkExec(
     child.execute().mapPartitions { iter =>
       val getEventTime = UnsafeProjection.create(eventTime :: Nil, child.output)
       iter.map { row =>
-        eventTimeStats.add((getEventTime(row).getLong(0).toDouble / 1000).toLong)
+        eventTimeStats.add(getEventTime(row).getLong(0) / 1000)
         row
       }
     }
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamExecution.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamExecution.scala
@@ -360,6 +360,24 @@ class StreamExecution(
     if (hasNewData) {
       // Current batch timestamp in milliseconds
       offsetSeqMetadata.batchTimestampMs = triggerClock.getTimeMillis()
+      // Update the eventTime watermark if we find one in the plan.
+      if (lastExecution != null) {
+        lastExecution.executedPlan.collect {
+          case e: EventTimeWatermarkExec if e.eventTimeStats.value.count > 0 =>
+            logDebug(s"Observed event time stats: ${e.eventTimeStats.value}")
+            e.eventTimeStats.value.max - e.delay.milliseconds
+        }.headOption.foreach { newWatermarkMs =>
+          if (newWatermarkMs > offsetSeqMetadata.batchWatermarkMs) {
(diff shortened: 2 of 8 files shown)


sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamingQueryStatusAndProgressSuite.scala
```diff
@@ -38,13 +38,18 @@ class StreamingQueryStatusAndProgressSuite extends SparkFunSuite {
...
-        |  "timestamp" : "2016-12-05T20:54:20.827Z",
+        |  "triggerTimestamp" : "2016-12-05T20:54:20.827Z",
         |  "numInputRows" : 678,
         |  "inputRowsPerSecond" : 10.0,
         |  "durationMs" : {
         |    "total" : 0
         |  },
-        |  "currentWatermark" : 3,
+        |  "queryTimestamps" : {
+        |    "eventTime.avg" : "2016-12-05T20:54:20.827Z",
```
tdas: @marmbrus the dot is in the key. when parsing using our json DF, we may have to refer to the column as "progress.queryTimestamps.`eventTim.max`"

sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamingQueryStatusAndProgressSuite.scala
```diff
@@ -38,13 +38,18 @@ class StreamingQueryStatusAndProgressSuite extends SparkFunSuite {
         |  "id" : "${testProgress1.id.toString}",
         |  "runId" : "${testProgress1.runId.toString}",
         |  "name" : "myName",
-        |  "timestamp" : "2016-12-05T20:54:20.827Z",
+        |  "triggerTimestamp" : "2016-12-05T20:54:20.827Z",
```
tdas: @marmbrus One idea is to not have a top level timestamp, and merge this with the `queryTimestamps`. Then this would be a key `triggerStartTime` in the `queryTimestamps` map. In fact then we can rename `queryTimestamps` to `timestamps`.

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamExecution.scala
```diff
@@ -360,6 +360,24 @@ class StreamExecution(
     if (hasNewData) {
       // Current batch timestamp in milliseconds
       offsetSeqMetadata.batchTimestampMs = triggerClock.getTimeMillis()
+      // Update the eventTime watermark if we find one in the plan.
```
tdas: I moved this such that the watermark is updated before starting a batch, rather than after finishing a batch. This keeps batchWatermarkMs consistent with batchTimestampsMs, both are set before starting a batch, and reduces complexities in the ProgressReporter.

sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamingQueryStatusAndProgressSuite.scala
marmbrus: One option is we could use dashes instead, but honestly I'm getting a little confused by all these timestamps.  In particular, I'm not sure what is the difference between `triggerTimestamp` and `processingTime` (and, I think that having `processingTime` mean anything different than `System.currentTimeMillis` will be confusing to users coming from other systems).

The two things that I think you really want to be able to track from the metrics feed are:
 - the actual timestamp when this progress update was produced.  I think this should remain top level and be called `timestamp`.
 - Stats about the event time, so that I can know roughly what data is present in this batch.  This can be useful for several reasons, including finding other problems upstream.  The more I think about this, I think you just want to see this as `"eventTime": { "min": ..., "max": ..., "watermark": ... }`

Its actually not clear to me why you need to know the original batch start time.  In 99% of cases this is the same as the `timestamp`.  If you are executing, a batch due to a failure, it will be different.  But, why does an outside monitoring job care?  I can't come up with any interesting graphs that you would make with this extra field.

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamExecution.scala
marmbrus: +1, this makes sense.

sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamingQueryStatusAndProgressSuite.scala
marmbrus: It is available, it is stored in a human readable format in the offset log (BTW, is that a long or a timestamp?).  I think in the log run we'll want to open up another API that gives you access to this log, but I think that can come later.  For now, its still pretty easy to find.

tdas: But the log gets cleaned up continuously, so will not be available if you are trying to debug it a day or an hour later.

marmbrus: Not any more... we keep 1000 now, right? This just really feels like we are reaching for a use case.  We can always add it, but I think the way it is done currently in this PR is very confusing and having `timestamp` and `eventTime` are very clear.

tdas: 1000 *  5 seconds batch is < 2 hours. Anyways, I think it will be equally or more confusing adding it later as a top-level field in Progress (e.g. `timestamp`, `eventTime`, `processingTime` ?). It may be better to have the top-level `timestamp`, and all other execution level detailed timestamps inside a single map. 

Anyways, I am updating the PR, but I think this is a little short-sighted.

sql/core/src/main/scala/org/apache/spark/sql/streaming/progress.scala
```diff
@@ -33,27 +34,6 @@ import org.apache.spark.sql.catalyst.util.DateTimeUtils
 
 /**
  * :: Experimental ::
- * Information about updates made to stateful operators in a [[StreamingQuery]] during a trigger.
```
tdas: I moved this below to keep the StreamingQueryProgress class first in the file. More important code first.

marmbrus: Thats actually the opposite of how most code in SQL is laid out, so I think it would be better to avoid this change.  The logic here is declarations that are use later should come first (references before declaration make it harder to read), and stuff at the end of the file is kind of hidden.

tdas: aah, then for consistency, SourceProgress and SinkProgress should also be before StreamingQueryProgress. But thats a bigger change should be done in a different PR.

sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamingQueryStatusAndProgressSuite.scala
```diff
@@ -44,7 +44,12 @@ class StreamingQueryStatusAndProgressSuite extends SparkFunSuite {
         |  "durationMs" : {
         |    "total" : 0
         |  },
-        |  "currentWatermark" : 3,
+        |  "eventTime" : {
```
brkyvz: awesome!

sql/core/src/main/scala/org/apache/spark/sql/streaming/progress.scala
```diff
@@ -67,7 +67,7 @@ class StateOperatorProgress private[sql](
  *                Similarly, when there is no data to be processed, the batchId will not be
  *                incremented.
  * @param durationMs The amount of time taken to perform various operations in milliseconds.
- * @param currentWatermark The current event time watermark in milliseconds
+ * @param eventTime Statistics of event time seen in this batch
```
zsxwing: nit: since you are touching this file, could you also fix the comment of `@param timestamp`? It's better to document the format as well, such as, `The beginning time of the trigger in ISO8601 format. (e.g., 2016-12-05T20:54:20.827Z)`

zsxwing: In addition, could you also add an example in the comment of `eventTime`?

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/EventTimeWatermarkExec.scala
```diff
@@ -28,24 +26,48 @@ import org.apache.spark.sql.types.MetadataBuilder
...
+  }
+
+  def merge(that: EventTimeStats): Unit = {
+    this.max = math.max(this.max, that.max)
+    this.min = math.min(this.min, that.min)
+    this.sum += that.sum
+    this.count += that.count
+  }
+
+  def avg: Long = (sum.toDouble / count).toLong
```
zsxwing: nit: why not use `sum / count`?

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/EventTimeWatermarkExec.scala
```diff
@@ -61,15 +83,14 @@ case class EventTimeWatermarkExec(
...
-  sparkContext.register(maxEventTime)
+  val eventTimeStats = new EventTimeStatsAccum()
+  sparkContext.register(eventTimeStats)
 
   override protected def doExecute(): RDD[InternalRow] = {
     child.execute().mapPartitions { iter =>
       val getEventTime = UnsafeProjection.create(eventTime :: Nil, child.output)
       iter.map { row =>
-        maxEventTime.add(getEventTime(row).getLong(0))
+        eventTimeStats.add((getEventTime(row).getLong(0).toDouble / 1000).toLong)
```
zsxwing: nit: could be `getEventTime(row).getLong(0) / 1000`

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/EventTimeWatermarkExec.scala
```diff
@@ -28,24 +26,48 @@ import org.apache.spark.sql.types.MetadataBuilder
 import org.apache.spark.unsafe.types.CalendarInterval
 import org.apache.spark.util.AccumulatorV2
 
-/** Tracks the maximum positive long seen. */
-class MaxLong(protected var currentValue: Long = 0)
-  extends AccumulatorV2[Long, Long] {
+/** Class for collecting event time stats with an accumulator */
```
zsxwing: nit: please document the time unit.

zsxwing (commented):
Looks good overall. Just some nits.

zsxwing:
LGTM

marmbrus:
LGTM

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/EventTimeWatermarkExec.scala
tdas: Documented it in EventTimeWatermarkExec

tdas:
Merging to master and 2.1

sql/core/src/main/scala/org/apache/spark/sql/streaming/progress.scala
```diff
@@ -61,13 +61,20 @@ class StateOperatorProgress private[sql](
...
+ * @param timestamp Beginning time of the trigger in ISO8601 format, i.e. UTC timestamps.
  * @param batchId A unique id for the current batch of data being processed.  Note that in the
  *                case of retries after a failure a given batchId my be executed more than once.
  *                Similarly, when there is no data to be processed, the batchId will not be
  *                incremented.
  * @param durationMs The amount of time taken to perform various operations in milliseconds.
- * @param currentWatermark The current event time watermark in milliseconds
+ * @param eventTime Statistics of event time seen in this batch. It may contain the following keys:
+ *                 {
+ *                   "max" -> "2016-12-05T20:54:20.827Z"  // maximum event time seen in this trigger
```
HyukjinKwon: Hi all, I am just leaving a comment as a gentle reminder to note that we probably should replace `<` or `>` to other ones such as `{@literal <}` or `{@literal >}` in the future. Please refer https://github.com/apache/spark/pull/16013#discussion_r89666914. This causes javadoc8 break.

```
[error] .../java/org/apache/spark/sql/streaming/StreamingQueryProgress.java:19: error: bad use of '>'
[error]  *                   "max" -> "2016-12-05T20:54:20.827Z"  // maximum event time seen in this trigger
[error]                             ^
[error] .../java/org/apache/spark/sql/streaming/StreamingQueryProgress.java:20: error: bad use of '>'
[error]  *                   "min" -> "2016-12-05T20:54:20.827Z"  // minimum event time seen in this trigger
[error]                             ^
[error] .../java/org/apache/spark/sql/streaming/StreamingQueryProgress.java:21: error: bad use of '>'
[error]  *                   "avg" -> "2016-12-05T20:54:20.827Z"  // average event time seen in this trigger
[error]                             ^
[error] .../java/org/apache/spark/sql/streaming/StreamingQueryProgress.java:22: error: bad use of '>'
[error]  *                   "watermark" -> "2016-12-05T20:54:20.827Z"  // watermark used in this trigger
[error]                                   ^
```

PR #16261, opened 2016-12-13
[SPARK-18836] [CORE] Serialize one copy of task metrics in DAGScheduler

shivaram:

Right now we serialize the empty task metrics once per task – Since this is shared across all tasks we could use the same serialized task metrics across all tasks of a stage. 

## How was this patch tested?

- [x] Run tests on EC2 to measure performance improvement

### DIFF
--- a/core/src/main/scala/org/apache/spark/scheduler/DAGScheduler.scala
+++ b/core/src/main/scala/org/apache/spark/scheduler/DAGScheduler.scala
@@ -1009,13 +1009,14 @@ class DAGScheduler(
     }
 
     val tasks: Seq[Task[_]] = try {
+      val serializedTaskMetrics = closureSerializer.serialize(stage.latestInfo.taskMetrics).array()
       stage match {
         case stage: ShuffleMapStage =>
           partitionsToCompute.map { id =>
             val locs = taskIdToLocations(id)
             val part = stage.rdd.partitions(id)
             new ShuffleMapTask(stage.id, stage.latestInfo.attemptId,
-              taskBinary, part, locs, stage.latestInfo.taskMetrics, properties, Option(jobId),
+              taskBinary, part, locs, properties, serializedTaskMetrics, Option(jobId),
               Option(sc.applicationId), sc.applicationAttemptId)
           }
 
@@ -1025,7 +1026,7 @@ class DAGScheduler(
             val part = stage.rdd.partitions(p)
             val locs = taskIdToLocations(id)
             new ResultTask(stage.id, stage.latestInfo.attemptId,
-              taskBinary, part, locs, id, properties, stage.latestInfo.taskMetrics,
+              taskBinary, part, locs, id, properties, serializedTaskMetrics,
               Option(jobId), Option(sc.applicationId), sc.applicationAttemptId)
           }
       }
--- a/core/src/main/scala/org/apache/spark/scheduler/Task.scala
+++ b/core/src/main/scala/org/apache/spark/scheduler/Task.scala
@@ -48,6 +48,8 @@ import org.apache.spark.util._
  * @param partitionId index of the number in the RDD
  * @param metrics a `TaskMetrics` that is created at driver side and sent to executor side.
  * @param localProperties copy of thread-local properties set by the user on the driver side.
+ * @param serializedTaskMetrics a `TaskMetrics` that is created and serialized on the driver side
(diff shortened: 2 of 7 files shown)


shivaram:
This code was originally written by @JoshRosen. BTW while I put this up for review,  I am going to run some performance tests on EC2 today / tomm to see exactly how much benefit this gives

cc @kayousterhout

JoshRosen:
Woohoo! This change has been sitting on my laptop for a while but I haven't had the time to clean it up for submission. Thanks for picking this up.

core/src/main/scala/org/apache/spark/scheduler/Task.scala
```diff
@@ -58,13 +60,17 @@ private[spark] abstract class Task[T](
     val stageId: Int,
     val stageAttemptId: Int,
     val partitionId: Int,
-    // The default value is only used in tests.
-    val metrics: TaskMetrics = TaskMetrics.registered,
     @transient var localProperties: Properties = new Properties,
+    // The default value is only used in tests.
+    serializedTaskMetrics: Array[Byte] =
```
srowen: Nit: `val`? No big deal. seems fine to me.

shivaram:
With the help of @kayousterhout I ran a scheduling microbenchmark (Code [1]) with 10000 tasks per stage on 20 m2.4xlarge machines on EC2 (160 cores). From 10 trials, I measured the average time taken per stage.
Before this PR (baseline): 2526.81 ms 
With this PR: 1741.99 ms

So overall we get a 785ms improvement (~30%) in this case. To figure out more closely where the speedup was coming from I added a timer inside the function `Task.serializeWithDependencies`[2]. 

Avg. Time to serialize one task without this PR: 0.119954 ms
Avg. Time to serialize one task with this PR: 0.0556422 ms

Thus we save around 0.064 ms in serialization time per task and that explains most of the improvements. 

[1] https://gist.github.com/shivaram/c84d18512fe8ba9c047e3d2b170b9f68
[2] https://github.com/apache/spark/blob/172a52f5d31337d90155feb7072381e8d5712288/core/src/main/scala/org/apache/spark/scheduler/Task.scala#L224

shivaram:
Jenkins, retest this please

shivaram:
Jenkins, retest this please

core/src/test/scala/org/apache/spark/scheduler/FakeTask.scala
```diff
@@ -17,12 +17,20 @@
...
+import org.apache.spark.executor.TaskMetrics
 
 class FakeTask(
     stageId: Int,
     partitionId: Int,
-    prefLocs: Seq[TaskLocation] = Nil) extends Task[Int](stageId, 0, partitionId) {
+    prefLocs: Seq[TaskLocation] = Nil,
+    serializedTaskMetrics: Array[Byte] =
+      SparkEnv.get.closureSerializer.newInstance().serialize(TaskMetrics.registered).array())
+  extends Task[Int](stageId, 0, partitionId, new Properties, serializedTaskMetrics) {
```
shivaram: Same as below - or to put it another way we use FakeTask in other test suites which dont have the mock problem I described below. For those cases we use the default value.

core/src/main/scala/org/apache/spark/scheduler/DAGScheduler.scala
```diff
@@ -1009,13 +1009,14 @@ class DAGScheduler(
...
     val tasks: Seq[Task[_]] = try {
+      val serializedTaskMetrics = closureSerializer.serialize(stage.latestInfo.taskMetrics).array()
       stage match {
         case stage: ShuffleMapStage =>
           partitionsToCompute.map { id =>
             val locs = taskIdToLocations(id)
             val part = stage.rdd.partitions(id)
             new ShuffleMapTask(stage.id, stage.latestInfo.attemptId,
-              taskBinary, part, locs, stage.latestInfo.taskMetrics, properties, Option(jobId),
+              taskBinary, part, locs, properties, serializedTaskMetrics, Option(jobId),
```
mridulm: Thanks for switching the order of param - makes it consistent now !

mridulm (approved):
LGTM. Would be great to get more eyeballs on this though.

kayousterhout:
LGTM

kayousterhout:
Merged this to master. Thanks @shivaram!

PR #16262, opened 2016-12-13
[SPARK-17932][SQL][FOLLOWUP] Change statement `SHOW TABLES EXTENDED` to `SHOW TABLE EXTENDED`

jiangxb1987:

Change the statement `SHOW TABLES [EXTENDED] [(IN|FROM) database_name] [[LIKE] 'identifier_with_wildcards'] [PARTITION(partition_spec)]` to the following statements:

- SHOW TABLES [(IN|FROM) database_name] [[LIKE] 'identifier_with_wildcards']
- SHOW TABLE EXTENDED [(IN|FROM) database_name] LIKE 'identifier_with_wildcards' [PARTITION(partition_spec)]

After this change, the statements `SHOW TABLE/SHOW TABLES` have the same syntax with that HIVE has.

## How was this patch tested?
Modified the test sql file `show-tables.sql`;
Modified the test suite `DDLSuite`.

### DIFF
--- a/sql/catalyst/src/main/antlr4/org/apache/spark/sql/catalyst/parser/SqlBase.g4
+++ b/sql/catalyst/src/main/antlr4/org/apache/spark/sql/catalyst/parser/SqlBase.g4
@@ -120,8 +120,10 @@ statement
         (USING resource (',' resource)*)?                              #createFunction
     | DROP TEMPORARY? FUNCTION (IF EXISTS)? qualifiedName              #dropFunction
     | EXPLAIN (LOGICAL | FORMATTED | EXTENDED | CODEGEN)? statement    #explain
-    | SHOW TABLES EXTENDED? ((FROM | IN) db=identifier)?
-        (LIKE? pattern=STRING)? partitionSpec?                         #showTables
+    | SHOW TABLES ((FROM | IN) db=identifier)?
+        (LIKE? pattern=STRING)?                                        #showTables
+    | SHOW TABLE EXTENDED ((FROM | IN) db=identifier)?
+        LIKE pattern=STRING partitionSpec?                             #showTable
     | SHOW DATABASES (LIKE pattern=STRING)?                            #showDatabases
     | SHOW TBLPROPERTIES table=tableIdentifier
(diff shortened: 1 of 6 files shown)


jiangxb1987:
cc @gatorsmile @hvanhovell

sql/catalyst/src/main/antlr4/org/apache/spark/sql/catalyst/parser/SqlBase.g4
```diff
@@ -120,8 +120,10 @@ statement
         (USING resource (',' resource)*)?                              #createFunction
     | DROP TEMPORARY? FUNCTION (IF EXISTS)? qualifiedName              #dropFunction
     | EXPLAIN (LOGICAL | FORMATTED | EXTENDED | CODEGEN)? statement    #explain
-    | SHOW TABLES EXTENDED? ((FROM | IN) db=identifier)?
-        (LIKE? pattern=STRING)? partitionSpec?                         #showTables
+    | SHOW TABLES ((FROM | IN) db=identifier)?
+        (LIKE? pattern=STRING)?                                        #showTables
+    | SHOW TABLE EXTENDED ((FROM | IN) db=identifier)?
+        LIKE pattern=STRING partitionSpec?                             #showTable
```
gatorsmile: Thanks for the split to be consistent with Hive! 

If we do not split, the outputs will be confusing when supporting `partitionSpec` in the future. In Hive, users cannot use regular expression for table name if a partition specification is present. That means, when supporting `partitionSpec`, we only allow users to specify a single table in `SHOW TABLE`.

gatorsmile:
LGTM

hvanhovell:
@jiangxb1987 could you open a backport for branch 2.1?

gatorsmile:
@hvanhovell We did not merge the original PR https://github.com/apache/spark/pull/15958 to 2.1. No need to backport this to Spark 2.1, right? Thanks!

hvanhovell:
@gatorsmile yeah you are right about that. @jiangxb1987 ignore my last request :)

PR #16263, opened 2016-12-13
[SPARK-18281][SQL][PySpark] Remove timeout for reading data through socket for local iterator

viirya:

There is a timeout failure when using `rdd.toLocalIterator()` or `df.toLocalIterator()` for a PySpark RDD and DataFrame:

    df = spark.createDataFrame([[1],[2],[3]])
    it = df.toLocalIterator()
    row = next(it)

    df2 = df.repartition(1000)  # create many empty partitions which increase materialization time so causing timeout 
    it2 = df2.toLocalIterator() 
    row = next(it2)

The cause of this issue is, we open a socket to serve the data from JVM side. We set timeout for connection and reading through the socket in Python side. In Python we use a generator to read the data, so we only begin to connect the socket once we start to ask data from it. If we don't consume it immediately, there is connection timeout.

In the other side, the materialization time for RDD partitions is unpredictable. So we can't set a timeout for reading data through the socket. Otherwise, it is very possibly to fail.

## How was this patch tested?

Added tests into PySpark.

### DIFF
--- a/python/pyspark/rdd.py
+++ b/python/pyspark/rdd.py
@@ -138,12 +138,9 @@ def _load_from_socket(port, serializer):
     # The RDD materialization time is unpredicable, if we set a timeout for socket reading
     # operation, it will very possibly fail. See SPARK-18281.
     sock.settimeout(None)
-    try:
-        rf = sock.makefile("rb", 65536)
-        for item in serializer.load_stream(rf):
-            yield item
-    finally:
-        sock.close()
+    # The socket will be automatically closed when garbage-collected.
+    rf = sock.makefile("rb", 65536)
+    return serializer.load_stream(sock.makefile("rb", 65536))
 
 
 def ignore_unicode_prefix(f):
@@ -2352,12 +2349,7 @@ class RDD(object):
         """
         with SCCallSiteSync(self.context) as css:
             port = self.ctx._jvm.PythonRDD.toLocalIteratorAndServe(self._jrdd.rdd())
-        # We set a timeout for connecting socket. The connection only begins when we start
-        # to consume the first element. If we do not begin to consume the returned iterator
-        # immediately, there will be a failure.
-        iter = _load_from_socket(port, self._jrdd_deserializer)
-        peek = next(iter)
-        return chain([peek], iter)
+        return _load_from_socket(port, self._jrdd_deserializer)
 
 
 def _prepare_for_python_RDD(sc, command):
--- a/python/pyspark/rdd.py
+++ b/python/pyspark/rdd.py
@@ -139,7 +139,6 @@ def _load_from_socket(port, serializer):
     # operation, it will very possibly fail. See SPARK-18281.
     sock.settimeout(None)
     # The socket will be automatically closed when garbage-collected.
-    rf = sock.makefile("rb", 65536)
     return serializer.load_stream(sock.makefile("rb", 65536))
 
 
--- a/python/pyspark/rdd.py
+++ b/python/pyspark/rdd.py
@@ -135,6 +135,9 @@ def _load_from_socket(port, serializer):
         break
     if not sock:
         raise Exception("could not open socket")
+    # The RDD materialization time is unpredicable, if we set a timeout for socket reading
+    # operation, it will very possibly fail. See SPARK-18281.
+    sock.settimeout(None)
     try:
         rf = sock.makefile("rb", 65536)
         for item in serializer.load_stream(rf):
@@ -2349,7 +2352,12 @@ class RDD(object):
         """
         with SCCallSiteSync(self.context) as css:
             port = self.ctx._jvm.PythonRDD.toLocalIteratorAndServe(self._jrdd.rdd())
-        return _load_from_socket(port, self._jrdd_deserializer)
+        # We set a timeout for connecting socket. The connection only begins when we start
+        # to consume the first element. If we do not begin to consume the returned iterator
+        # immediately, there will be a failure.
+        iter = _load_from_socket(port, self._jrdd_deserializer)
+        peek = next(iter)
+        return chain([peek], iter)
 
 
 def _prepare_for_python_RDD(sc, command):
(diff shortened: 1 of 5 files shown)


viirya:
retest this please.

viirya:
retest this please.

viirya:
Dealing with the difference between python2 and python3...

python/pyspark/sql/dataframe.py
```diff
@@ -400,10 +402,19 @@ def toLocalIterator(self):
 
         >>> list(df.toLocalIterator())
         [Row(age=2, name=u'Alice'), Row(age=5, name=u'Bob')]
+        >>> it = df.toLocalIterator()
+        >>> import time
+        >>> time.sleep(5)
+        >>> next(it)
```
holdenk: It might make sense to test consuming the second element given that the fix eagerly serves the first element, (and this additional test probably belongs in tests.py rather than a doctest though).

viirya: Ok. Thanks. I will update like that.

viirya:
retest this please.

holdenk:
So the more I look at this, the less I'm convinced this is the right solution. Eagerly evaluating the head of one of the partitions is fine, but if one of the middle partitions takes a long time to evaluate if one of the later partitions takes a long time to evaluate we will still run into trouble.

I think it might make sense to consider what the purpose of the timeouts are and if maybe there is a better solution on the JIRA.

holdenk:
But also thanks for taking this on, is an important issue to fix :)

viirya:
@holdenk Thanks for review. Yeah, I've considering this probability that a partition taking longer time to evaluate might cause problem again. But in the end as in Python side we have a buffered reading, I first guess we can alleviate this case.

I have thought to remove the timeout in Python side, or only apply timeout for connection. But I have concern that it may break the purpose of this timeout.

As reported on the jira, this eagerly evaluating seems don't completely solve the issue, I would try to take the approach that only sets timeout for connection and see if it solves this.

Then we can ask others' opinion about the timeout removing.

viirya:
@holdenk @dusenberrymw There are timeouts for connection and reading data. If we want to keep the connection timeout, which I think is reasonable, eagerly evaluating of the first element is necessary in both JVM and Python sides.

Timeout for reading data needs to be removed if we want to completely solve this. As the newer test case shows, the materialization time of RDD partitions is unpredictable and can easily cause timeout problem. I removed the timeout for reading operation on the socket. This can solve the test cases locally.

viirya:
cc @davies

python/pyspark/rdd.py
```diff
@@ -2349,7 +2352,12 @@ def toLocalIterator(self):
         """
         with SCCallSiteSync(self.context) as css:
             port = self.ctx._jvm.PythonRDD.toLocalIteratorAndServe(self._jrdd.rdd())
-        return _load_from_socket(port, self._jrdd_deserializer)
+        # We set a timeout for connecting socket. The connection only begins when we start
+        # to consume the first element. If we do not begin to consume the returned iterator
+        # immediately, there will be a failure.
+        iter = _load_from_socket(port, self._jrdd_deserializer)
```
davies: Since the timeout is removed, is this still needed?

viirya: We still keep connection timeout. Eagerly evaluating of the first element is needed for it. Because it is going to connect the socket only if we begin to consume the iterator. Otherwise, the server socket in JVM side will be timeout.

davies: I think we could have a wrapper to load items, and return the iterator in _load_from_socket(): 
```
def _load_from_socket(port):
     ...
    return  serializer.load_stream(sock.makefile("rb", 65536))
```

The `finally` may be optional, GC cloud close the socket (should be confirmed).

viirya: It is.

>   socket.close()
>   Close the socket. All future operations on the socket object will fail. The remote end will receive no more data (after queued data is flushed). Sockets are automatically closed when they are garbage-collected.

viirya:
ping @davies Can you take a look again? Thanks.

python/pyspark/rdd.py
```diff
@@ -135,12 +135,12 @@ def _load_from_socket(port, serializer):
...
-        for item in serializer.load_stream(rf):
-            yield item
-    finally:
-        sock.close()
+    # The RDD materialization time is unpredicable, if we set a timeout for socket reading
+    # operation, it will very possibly fail. See SPARK-18281.
+    sock.settimeout(None)
+    # The socket will be automatically closed when garbage-collected.
+    rf = sock.makefile("rb", 65536)
+    return serializer.load_stream(sock.makefile("rb", 65536))
```
dusenberrymw: Can you remove the `rf = sock.makefile("rb", 65536)` line since it is redundant?

dusenberrymw (changes requested):
Overall, this looks good now, @viirya.  I tried it out for my use-cases, and it works as expected now, without error.

Also, I tracked down the exact reason for the most recent case in which an increased number of partitions led to the error.  Since we started with a small DataFrame, the repartition to a much larger number of partitions led to a large number of partitions being empty.  Thus, the Python socket read would wait until the Scala side finally found a partition that was not empty (processing a bunch of empty partitions first), and thus would timeout.

This PR now fixes these issues. 👍  I just left one minor comment.

python/pyspark/sql/tests.py
```diff
@@ -558,6 +558,18 @@ def test_create_dataframe_from_objects(self):
         self.assertEqual(df.dtypes, [("key", "bigint"), ("value", "string")])
         self.assertEqual(df.first(), Row(key=1, value="1"))
 
+    def test_to_localiterator_for_dataframe(self):
```
davies: This test is duplicated, I think.

python/pyspark/rdd.py
viirya: ah, i miss this. :)

python/pyspark/sql/tests.py
viirya: Oh, the jira reported two test cases for rdd and dataframe. But yes, I think we can remove this.

dusenberrymw (approved):
LGTM.  Thanks, @viirya.

davies:
Merging this into master and 2.1 branch, thanks!

PR #16222, opened 2016-12-08
[SPARK-18797][SparkR]:Update spark.logit in sparkr-vignettes

wangmiao1981:
## What changes were proposed in this pull request?
spark.logit is added in 2.1. We need to update spark-vignettes to reflect the changes. This is part of SparkR QA work.

## How was this patch tested?

Manual build html. Please see attached image for the result.
test

PR #15697, opened 2016-10-31
[SparkR][Test]:remove unnecessary suppressWarnings

wangmiao1981:
## What changes were proposed in this pull request?

In test_mllib.R, there are two unnecessary suppressWarnings. This PR just removes them.

## How was this patch tested?

Existing unit tests.

PR #15686, opened 2016-10-30
[MINOR][DOC] Remove spaces following slashs

dongjoon-hyun:
## What changes were proposed in this pull request?

This PR merges multiple lines enumerating items in order to remove the redundant spaces following slashes in [Structured Streaming Programming Guide in 2.0.2-rc1](http://people.apache.org/~pwendell/spark-releases/spark-2.0.2-rc1-docs/structured-streaming-programming-guide.html).
- Before: `Scala/ Java/ Python`
- After: `Scala/Java/Python`
## How was this patch tested?

Manual by the followings because this is documentation update.

```
cd docs
SKIP_API=1 jekyll build
```

PR #16264, opened 2016-12-13
[SPARK-18793] [SPARK-18794] [R] add spark.randomForest/spark.gbt to vignettes

mengxr:

Mention `spark.randomForest` and `spark.gbt` in vignettes. Keep the content minimal since users can type `?spark.randomForest` to see the full doc.

cc: @jkbradley

### DIFF
--- a/R/pkg/vignettes/sparkr-vignettes.Rmd
+++ b/R/pkg/vignettes/sparkr-vignettes.Rmd
@@ -451,6 +451,8 @@ SparkR supports the following machine learning models and algorithms.
 
 * Random Forest
 
+* Gradient-Boosted Trees (GBT)
+
 * Naive Bayes Model
 
 * $k$-means Clustering
@@ -531,7 +533,7 @@ head(select(gaussianFitted, "model", "prediction", "mpg", "wt", "hp"))
 #### Random Forest
 
 `spark.randomForest` fits a [random forest](https://en.wikipedia.org/wiki/Random_forest) classification or regression model on a `SparkDataFrame`.
-Users can use `summary` to get a summary of the fitted model, `predict` to make predictions on new data, and `write.ml`/`read.ml` to save/load fitted models.
+Users can call `summary` to get a summary of the fitted model, `predict` to make predictions, and `write.ml`/`read.ml` to save/load fitted models.
 
 In the following example, we use the `longley` dataset to train a random forest and make predictions:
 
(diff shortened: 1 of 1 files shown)


felixcheung:
LGTM. maybe good to have one example for classification but optional

felixcheung:
and if we are doing "(Added in Spark 2.1.0)" in https://github.com/apache/spark/pull/16222/files perhaps we should have the same for these 2?

actually, I'd vote for removing them - vignettes is for that specific version of package you install. If you install SparkR 2.1.0 what is described there is in 2.1.0, by default

R/pkg/vignettes/sparkr-vignettes.Rmd
```diff
@@ -539,7 +539,7 @@ In the following example, we use the `longley` dataset to train a random forest
 
 ```{r}
 df <- createDataFrame(longley)
-rfModel <- spark.randomForest(df, Employed ~ ., type = "regression", numTrees = 5)
+rfModel <- spark.randomForest(df, Employed ~ ., type = "regression", maxDepth = 2, numTrees = 2)
```
felixcheung: you could also do this to limit output

```{r, include=FALSE}
ops <- options()
options(max.print=40)
```

there is another example in the vignettes

R/pkg/vignettes/sparkr-vignettes.Rmd
```diff
@@ -526,6 +530,34 @@ gaussianFitted <- predict(gaussianGLM, carsDF)
...
 ```
 
+#### Random Forest
+
+`spark.randomForest` fits a [random forest](https://en.wikipedia.org/wiki/Random_forest) classification or regression model on a `SparkDataFrame`.
+Users can call `summary` to get a summary of the fitted model, `predict` to make predictions, and `write.ml`/`read.ml` to save/load fitted models.
+
+In the following example, we use the `longley` dataset to train a random forest and make predictions:
+
+```{r}
```
felixcheung: longley has columns with `.` like Armed.Forces
suggest putting `warning = FALSE` like [here](https://github.com/apache/spark/pull/16222/files#diff-56790892076daac0bc43a9b38c0b1470R784) otherwise we will have a warning in the output vignettes

R/pkg/vignettes/sparkr-vignettes.Rmd
mengxr: The current length of the output seems fine (<40 lines).

R/pkg/vignettes/sparkr-vignettes.Rmd
mengxr: updated

mengxr:
Agree that it is not necessary to mention the version added here. I will send a follow-up PR after this one to rearrange the ordering of the ML algorithms. There is no logical ordering now.

mengxr:
Merged into master and branch-2.1. The last commit passed Jenkins.

mengxr:
@HyukjinKwon Could you take a look at the AppVeyor error?

felixcheung:
That looks like a network error when accessing github

HyukjinKwon:
Yup, it seems failed time to time due to network problem. FYI, there was some discussions about this in https://github.com/apache/spark/pull/15686#issuecomment-257217562 and https://github.com/apache/spark/pull/15697#issuecomment-257721992

I could not find a workaround to re-trigger this so far and it seems (I manually checked and privately asked if it is true) some Apahce projects using Travis CI/AppVeyor are also re-triggering this via closing and opening.

PR #16230, opened 2016-12-09
[SPARK-13747][Core]Fix potential ThreadLocal leaks in RPC when using ForkJoinPool

zsxwing:
## What changes were proposed in this pull request?

Some places in SQL may call `RpcEndpointRef.askWithRetry` (e.g., ParquetFileFormat.buildReader -> SparkContext.broadcast -> ... -> BlockManagerMaster.updateBlockInfo -> RpcEndpointRef.askWithRetry), which will finally call `Await.result`. It may cause `java.lang.IllegalArgumentException: spark.sql.execution.id is already set` when running in Scala ForkJoinPool.

This PR includes the following changes to fix this issue:

- Remove `ThreadUtils.awaitResult`
- Rename `ThreadUtils. awaitResultInForkJoinSafely` to `ThreadUtils.awaitResult`
- Replace `Await.result` in RpcTimeout with `ThreadUtils.awaitResult`.

## How was this patch tested?

Jenkins

PR #16268, opened 2016-12-13
[SPARK-18843][Core]Fix timeout in awaitResultInForkJoinSafely (branch 2.1, 2.0)

zsxwing:

This PR fixes the timeout value in `awaitResultInForkJoinSafely` for 2.1 and 2.0. Master has been fixed by https://github.com/apache/spark/pull/16230.

## How was this patch tested?

Jenkins


srowen:
I see, so there is no point in applying this to master, because it's as if it were already superseded by a bigger fix? but we 'back port' this stop-gap to earlier branches.

zsxwing:
Yep. I want to backport this obvious fix without the entire patch for master.

yhuai:
LGTM

zsxwing:
Thanks! Merging to 2.1 and 2.0.

PR #14079, opened 2016-07-06
[SPARK-8425][CORE] Application Level Blacklisting

squito:
## What changes were proposed in this pull request?

This builds upon the blacklisting introduced in SPARK-17675 to add blacklisting of executors and nodes for an entire Spark application.  Resources are blacklisted based on tasks that fail, in tasksets that eventually complete successfully; they are automatically returned to the pool of active resources based on a timeout.  Full details are available in a design doc attached to the jira. 
## How was this patch tested?

Added unit tests, ran them via Jenkins, also ran a handful of them in a loop to check for flakiness.

The added tests include:
- verifying BlacklistTracker works correctly
- verifying TaskSchedulerImpl interacts with BlacklistTracker correctly (via a mock BlacklistTracker)
- an integration test for the entire scheduler with blacklisting in a few different scenarios

PR #16270, opened 2016-12-13
[SPARK-18846][Scheduler] Fix flakiness in SchedulerIntegrationSuite

squito:

There is a small race in SchedulerIntegrationSuite.
The test assumes that the taskscheduler thread
processing that last task will finish before the DAGScheduler processes
the task event and notifies the job waiter, but that is not 100%
guaranteed.

## How was this patch tested?

ran the test locally a bunch of times, never failed, though admittedly it never failed locally for me either.  However I am nearly 100% certain this is what caused the failure of one jenkins build https://amplab.cs.berkeley.edu/jenkins/job/SparkPullRequestBuilder/68694/consoleFull (which is long gone now, sorry -- I fixed it as part of https://github.com/apache/spark/pull/14079 initially)

### DIFF
--- a/core/src/test/scala/org/apache/spark/scheduler/SchedulerIntegrationSuite.scala
+++ b/core/src/test/scala/org/apache/spark/scheduler/SchedulerIntegrationSuite.scala
@@ -163,7 +163,7 @@ abstract class SchedulerIntegrationSuite[T <: MockBackend: ClassTag] extends Spa
       // when the job succeeds.
       // When the final task of a taskset completes, we post
       // the event to the DAGScheduler event loop before we finish processing in the taskscheduler
-      // thread.  Its possible the DAGScheduler thread processes the event, finishes the job,
+      // thread.  It's possible the DAGScheduler thread processes the event, finishes the job,
       // and notifies the job waiter before our original thread in the task scheduler finishes
       // handling the event and marks the taskset as complete.  So its ok if we need to wait a
       // *little* bit longer for the original taskscheduler thread to finish up to deal w/ the race.
--- a/core/src/test/scala/org/apache/spark/scheduler/SchedulerIntegrationSuite.scala
+++ b/core/src/test/scala/org/apache/spark/scheduler/SchedulerIntegrationSuite.scala
@@ -27,9 +27,9 @@ import scala.language.existentials
 import scala.reflect.ClassTag
 
 import org.scalactic.TripleEquals
+import org.scalatest.Assertions.AssertionsHelper
 import org.scalatest.concurrent.Eventually._
 import org.scalatest.time.SpanSugar._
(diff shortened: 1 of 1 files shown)


core/src/test/scala/org/apache/spark/scheduler/SchedulerIntegrationSuite.scala
```diff
@@ -27,6 +27,9 @@ import scala.language.existentials
 import scala.reflect.ClassTag
 
 import org.scalactic.TripleEquals
+import org.scalatest.concurrent.Eventually._
+import org.scalatest.concurrent.PatienceConfiguration
```
kayousterhout: do you need this? (I compiled successfully w/o it)

core/src/test/scala/org/apache/spark/scheduler/SchedulerIntegrationSuite.scala
```diff
@@ -157,8 +160,16 @@ abstract class SchedulerIntegrationSuite[T <: MockBackend: ClassTag] extends Spa
...
-      // when the job succeeds
-      assert(taskScheduler.runningTaskSets.isEmpty)
+      // when the job succeeds.
+      // When the final task of a taskset completes, we post
+      // the event to the DAGScheduler event loop before we finish processing in the taskscheduler
+      // thread.  It's possible the DAGScheduler thread processes the event, finishes the job,
+      // and notifies the job waiter before our original thread in the task scheduler finishes
+      // handling the event and marks the taskset as complete.  So its ok if we need to wait a
+      // *little* bit longer for the original taskscheduler thread to finish up to deal w/ the race.
+      eventually(timeout(1 second), interval(100 millis)) {
```
kayousterhout: I might do 10 millis here since it seems like there's not much cost here to checking more frequently, and 100 millis is somewhat long to wait

core/src/test/scala/org/apache/spark/scheduler/SchedulerIntegrationSuite.scala
```diff
@@ -27,6 +27,8 @@ import scala.language.existentials
 import scala.reflect.ClassTag
 
 import org.scalactic.TripleEquals
+import org.scalatest.concurrent.Eventually._
+import org.scalatest.time.SpanSugar._
 import org.scalatest.Assertions.AssertionsHelper
```
kayousterhout: derp sorry just noticed this one is out of order (should be above org.scalatest.concurrent)

squito: no problem, sorry I never get these right -- I actually thought package imports were supposed to be first and checked some other places to realize you are right.

kayousterhout:
LGTM

squito:
merged to master

PR #16272, opened 2016-12-13
[SPARK-18850][SS]Make StreamExecution and progress classes serializable

zsxwing:

This PR adds StreamingQueryWrapper to make StreamExecution and progress classes serializable because it is too easy for it to get captured with normal usage. If StreamingQueryWrapper gets captured in a closure but no place calls its methods, it should not fail the Spark tasks. However if its methods are called, then this PR will throw a better message.

## How was this patch tested?

`test("StreamingQuery should be Serializable but cannot be used in executors")`
`test("progress classes should be Serializable")`


sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/ProgressReporter.scala
```diff
@@ -71,7 +71,7 @@ trait ProgressReporter extends Logging {
   private var metricWarningLogged: Boolean = false
 
   /** Holds the most recent query progress updates.  Accesses must lock on the queue itself. */
-  private val progressBuffer = new mutable.Queue[StreamingQueryProgress]()
+  @transient private val progressBuffer = new mutable.Queue[StreamingQueryProgress]()
```
zsxwing: mark `progressBuffer` `@transient` instead of changing it to serializable because it may contain a lot of progresses and we should try to avoid sending them to executors.

sql/core/src/main/scala/org/apache/spark/sql/streaming/progress.scala
```diff
@@ -38,7 +38,7 @@ import org.apache.spark.sql.catalyst.util.DateTimeUtils
 @Experimental
 class StateOperatorProgress private[sql](
     val numRowsTotal: Long,
-    val numRowsUpdated: Long) {
+    val numRowsUpdated: Long) extends Serializable {
```
tdas: +1

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamProgress.scala
```diff
@@ -24,7 +24,7 @@ import scala.collection.{immutable, GenTraversableOnce}
  */
 class StreamProgress(
     val baseMap: immutable.Map[Source, Offset] = new immutable.HashMap[Source, Offset])
-  extends scala.collection.immutable.Map[Source, Offset] {
+  extends scala.collection.immutable.Map[Source, Offset] with Serializable {
```
tdas: I dont think this is needed any more. This is purely internal

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamingQueryWrapper.scala
```diff
@@ -0,0 +1,113 @@
...
+import org.apache.spark.sql.SparkSession
+import org.apache.spark.sql.streaming.{StreamingQuery, StreamingQueryException, StreamingQueryProgress, StreamingQueryStatus}
+
+/**
+ * Wrap non-serializable StreamExecution to make the query serializable as it's easy to for it to
+ * get captured with normal usage. It's safe to capture the query but not use it in executors.
+ * However, if the user tries to call its methods, it will throw `IllegalStateException`.
+ */
+class StreamingQueryWrapper(
+    @transient val streamingQuery: StreamExecution) extends StreamingQuery with Serializable {
```
tdas: streamingQuery -> @transient private val _streamExecution
def streamExecution: StreamExecution = {
    if (_streamExecution == null) {
       throw new IllegalStateException("StreamingQuery cannot be used in executors")
    }
    _streamExecution
}

sql/core/src/main/scala/org/apache/spark/sql/streaming/progress.scala
```diff
@@ -83,7 +83,7 @@ class StreamingQueryProgress private[sql](
   val currentWatermark: Long,
   val stateOperators: Array[StateOperatorProgress],
   val sources: Array[SourceProgress],
-  val sink: SinkProgress) {
+  val sink: SinkProgress) extends Serializable {
```
tdas: There should be new tests for verifying serializability

sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamSuite.scala
```diff
@@ -300,6 +302,48 @@ class StreamSuite extends StreamTest {
       q.stop()
     }
   }
+
+  test("StreamingQuery should be Serializable but cannot be used in executors") {
```
tdas: Why is this in StreamSuite. This is better to be in StreamingQuerySuite as its related to the StreamingQuery.

sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamingQuerySuite.scala
```diff
@@ -439,6 +440,37 @@ class StreamingQuerySuite extends StreamTest with BeforeAndAfter with Logging {
     }
   }
 
+  test("progress classes should be Serializable") {
```
tdas: This should be in the StreamingStatusAndPRogressSuite

sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamingQueryStatusAndProgressSuite.scala
```diff
@@ -137,12 +146,13 @@ object StreamingQueryStatusAndProgressSuite {
     name = "myName",
     timestamp = "2016-12-05T20:54:20.827Z",
     batchId = 2L,
-    durationMs = Map("total" -> 0L).mapValues(long2Long).asJava,
-    eventTime = Map(
-      "max" -> "2016-12-05T20:54:20.827Z",
-      "min" -> "2016-12-05T20:54:20.827Z",
-      "avg" -> "2016-12-05T20:54:20.827Z",
-      "watermark" -> "2016-12-05T20:54:20.827Z").asJava,
+    durationMs = Collections.singletonMap("total", 0L),
```
zsxwing: I changed these two maps to pure Java maps, as the map created by `asJava` is not serializable.

tdas: Wait ... does this mean, that the maps we are generating in the StreamingQuery is NOT serializable?? Because we are probably doing the same conversion in ProgressReporter

sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamingQueryStatusAndProgressSuite.scala
```diff
@@ -137,12 +174,12 @@ object StreamingQueryStatusAndProgressSuite {
     name = "myName",
     timestamp = "2016-12-05T20:54:20.827Z",
     batchId = 2L,
-    durationMs = Map("total" -> 0L).mapValues(long2Long).asJava,
-    eventTime = Map(
+    durationMs = new java.util.HashMap(Map("total" -> 0L).mapValues(long2Long).asJava),
```
zsxwing: Changed them to use the same approach as `ProgressReporter` to create Java Map.

zsxwing:
retest this please

sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamingQueryManagerSuite.scala
```diff
@@ -244,7 +244,7 @@ class StreamingQueryManagerSuite extends StreamTest with BeforeAndAfter {
     failAfter(streamingTimeout) {
       val queries = withClue("Error starting queries") {
         datasets.zipWithIndex.map { case (ds, i) =>
-          @volatile var query: StreamExecution = null
+          var query: StreamingQuery = null
```
tdas: why was this volatile removed?

tdas:
LGTM. Just one minor question.

sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamingQueryManagerSuite.scala
zsxwing: It's not necessary. This is just a local variable.

PR #16273, opened 2016-12-13
[SPARK-18852][SS]StreamingQuery.lastProgress should be null when recentProgress is empty

zsxwing:

Right now `StreamingQuery.lastProgress` throws NoSuchElementException and it's hard to be used in Python since Python user will just see Py4jError.

This PR just makes it return null instead.

## How was this patch tested?

`test("lastProgress should be null when recentProgress is empty")`

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/ProgressReporter.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/ProgressReporter.scala
@@ -100,7 +100,7 @@ trait ProgressReporter extends Logging {
     progressBuffer.toArray
   }
 
-  /** Returns the most recent query progress update. */
+  /** Returns the most recent query progress update or null if there were no progress updates. */
   def lastProgress: StreamingQueryProgress = progressBuffer.synchronized {
     progressBuffer.lastOption.orNull
   }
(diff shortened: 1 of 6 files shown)


sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamingQuerySuite.scala
```diff
@@ -312,6 +313,24 @@ class StreamingQuerySuite extends StreamTest with BeforeAndAfter with Logging {
     )
   }
 
+  test("lastProgress should be null when recentProgress is empty") {
+    BlockingSource.latch = new CountDownLatch(1)
```
brkyvz: Glad that you got to re-use `BlockingSource`!

sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamingQuerySuite.scala
```diff
@@ -312,6 +313,24 @@ class StreamingQuerySuite extends StreamTest with BeforeAndAfter with Logging {
     )
   }
 
+  test("lastProgress should be null when recentProgress is empty") {
```
brkyvz: should we add a similar test to python just to make sure it returns `None`? Do we need to update any docs?

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/ProgressReporter.scala
```diff
@@ -102,7 +102,7 @@ trait ProgressReporter extends Logging {
 
   /** Returns the most recent query progress update. */
```
brkyvz: Maybe add `returns null if there were no progress updates`

brkyvz:
Left one comment on updating docs, otherwise LGTM

zsxwing:
retest this please

zsxwing:
retest this please

brkyvz:
LGTM!

PR #16274, opened 2016-12-14
[SPARK-18853][SQL] Project (UnaryNode) is way too aggressive in estimating statistics

rxin:
This patch reduces the default number element estimation for arrays and maps from 100 to 1. The issue with the 100 number is that when nested (e.g. an array of map), 100 * 100 would be used as the default size. This sounds like just an overestimation which doesn't seem that bad (since it is usually better to overestimate than underestimate). However, due to the way we assume the size output for Project (new estimated column size / old estimated column size), this overestimation can become underestimation. It is actually in general in this case safer to assume 1 default element.

## How was this patch tested?
This should be covered by existing tests.

### DIFF
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/types/ArrayType.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/types/ArrayType.scala
@@ -78,8 +78,8 @@ case class ArrayType(elementType: DataType, containsNull: Boolean) extends DataT
       ("containsNull" -> containsNull)
 
   /**
-   * The default size of a value of the ArrayType is 1 * the default size of the element type.
-   * (We assume that there are 1 elements).
+   * The default size of a value of the ArrayType is the default size of the element type.
+   * We assume that there is only 1 element on average in an array. See SPARK-18853.
    */
   override def defaultSize: Int = 1 * elementType.defaultSize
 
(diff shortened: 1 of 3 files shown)


rxin:
cc @mallman @davies @srinathshankar

davies:
lgtm

sql/catalyst/src/main/scala/org/apache/spark/sql/types/ArrayType.scala
```diff
@@ -78,10 +78,10 @@ case class ArrayType(elementType: DataType, containsNull: Boolean) extends DataT
       ("containsNull" -> containsNull)
 
   /**
-   * The default size of a value of the ArrayType is 100 * the default size of the element type.
-   * (We assume that there are 100 elements).
+   * The default size of a value of the ArrayType is 1 * the default size of the element type.
+   * (We assume that there are 1 elements).
    */
-  override def defaultSize: Int = 100 * elementType.defaultSize
+  override def defaultSize: Int = 1 * elementType.defaultSize
```
hvanhovell: Why multiply by 1?

hvanhovell:
Two nits, otherwise LGTM.

sql/catalyst/src/main/scala/org/apache/spark/sql/types/ArrayType.scala
```diff
@@ -78,10 +78,10 @@ case class ArrayType(elementType: DataType, containsNull: Boolean) extends DataT
       ("containsNull" -> containsNull)
 
   /**
-   * The default size of a value of the ArrayType is 100 * the default size of the element type.
-   * (We assume that there are 100 elements).
+   * The default size of a value of the ArrayType is 1 * the default size of the element type.
```
mallman: Suggest:

```The default size of a value of the ArrayType is the default size of the element type.```

mallman:
Outside of some comment grooming, LGTM.

gatorsmile:
LGTM

hvanhovell:

PR #16276, opened 2016-12-14
[SPARK-18855][CORE] Add RDD flatten function

linbojin:

Added a new flatten function for RDD.

## How was this patch tested?

Unit tests inside RDDSuite and manually tests:
```
scala> val rdd = sc.makeRDD(List(List(1, 2, 3), List(4, 5), List(6)))
rdd: org.apache.spark.rdd.RDD[List[Int]] = ParallelCollectionRDD[0] at makeRDD at <console>:24

scala> rdd.flatten.collect
res0: Array[Int] = Array(1, 2, 3, 4, 5, 6)
```

### DIFF
--- a/core/src/main/scala/org/apache/spark/rdd/RDD.scala
+++ b/core/src/main/scala/org/apache/spark/rdd/RDD.scala
@@ -381,6 +381,14 @@ abstract class RDD[T: ClassTag](
   }
 
   /**
+    *  Return a new RDD by flattening all elements from RDD with traversable elements
+    */
+  def flatten[U: ClassTag](implicit asTraversable: T => TraversableOnce[U]): RDD[U] = withScope {
+    val f = (x: T) => asTraversable(x)
+    flatMap(f)
+  }
+
+  /**
    * Return a new RDD containing only the elements that satisfy a predicate.
    */
   def filter(f: T => Boolean): RDD[T] = withScope {
--- a/core/src/main/scala/org/apache/spark/rdd/RDD.scala
+++ b/core/src/main/scala/org/apache/spark/rdd/RDD.scala
@@ -381,11 +381,14 @@ abstract class RDD[T: ClassTag](
   }
 
   /**
-    *  Return a new RDD by flattening all elements from RDD with traversable elements
-    */
+   * Return a new RDD by flattening all elements from RDD with traversable elements
+   */
   def flatten[U: ClassTag](implicit asTraversable: T => TraversableOnce[U]): RDD[U] = withScope {
-    val f = (x: T) => asTraversable(x)
-    flatMap(f)
+    new MapPartitionsRDD[U, T](this, (context, pid, iter) => {
+      var newIter: Iterator[U] = Iterator.empty
+      for (x <- iter) newIter ++= asTraversable(x)
+      newIter
+    })
   }
 
   /**
--- a/core/src/test/scala/org/apache/spark/rdd/RDDSuite.scala
+++ b/core/src/test/scala/org/apache/spark/rdd/RDDSuite.scala
@@ -88,6 +88,13 @@ class RDDSuite extends SparkFunSuite with SharedSparkContext {
     }
   }
 
+  test("flatten") {
+    val nums = sc.makeRDD(Array(Array(1, 2, 3), Array(4, 5), Array(6)), 2)
+    assert(nums.flatten.collect().toList === List(1, 2, 3, 4, 5, 6))
+    val strs = sc.makeRDD(Array(Array("a", "b", "c"), Array("d", "e"), Array("f")), 2)
+    assert(strs.flatten.collect().toList === List("a", "b", "c", "d", "e", "f"))
+  }
+
   test("serialization") {
     val empty = new EmptyRDD[Int](sc)
     val serial = Utils.serialize(empty)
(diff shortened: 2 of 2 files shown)


linbojin:
cc @rxin

core/src/test/scala/org/apache/spark/rdd/RDDSuite.scala
```diff
@@ -88,6 +88,13 @@ class RDDSuite extends SparkFunSuite with SharedSparkContext {
     }
   }
 
+  test("flatten") {
+    val nums = sc.makeRDD(Array(Array(1, 2, 3), Array(4, 5), Array(6)), 2)
+    assert(nums.flatten.collect().toList === List(1, 2, 3, 4, 5, 6))
+    val strs = sc.makeRDD(Array(Array("a", "b", "c"), Array("d", "e"), Array("f")), 2)
+    assert(strs.flatten.collect().toList === List("a", "b", "c", "d", "e", "f"))
```
BryanCutler: I don't think it's necessary to test both nums and strings, one would be fine and you should just move it to the "basic operations" test above and reuse an RDD if possible.

linbojin: Thx, i will move test codes into "basic operations".

core/src/main/scala/org/apache/spark/rdd/RDD.scala
```diff
@@ -381,6 +381,14 @@ abstract class RDD[T: ClassTag](
   }
 
   /**
+    *  Return a new RDD by flattening all elements from RDD with traversable elements
+    */
+  def flatten[U: ClassTag](implicit asTraversable: T => TraversableOnce[U]): RDD[U] = withScope {
```
linbojin: Hi @srowen, thx for your suggestion. I have one way to use scala flatMap as follows:
```
  def flatten[U: ClassTag](implicit asTraversable: T => TraversableOnce[U]): RDD[U] = withScope {
    val f = (x: T) => asTraversable(x)
    val cleanF = sc.clean(f)
    new MapPartitionsRDD[U, T](this, (context, pid, iter) => iter.flatMap(f))
  }
```
Or i implement the logic by myself:
```
  def flatten[U: ClassTag](implicit asTraversable: T => TraversableOnce[U]): RDD[U] = withScope {
    new MapPartitionsRDD[U, T](this, (context, pid, iter) => new Iterator[U] {
      private val empty = Iterator.empty
      private var cur: Iterator[U] = empty
      private def nextCur() { cur = asTraversable(iter.next).toIterator }
      def hasNext: Boolean = {
        while (!cur.hasNext) {
          if (!iter.hasNext) return false
          nextCur()
        }
        true
      }
      def next(): U = (if (hasNext) cur else empty).next()
    })
  }
```
ref: https://github.com/scala/scala/blob/v2.11.8/src/library/scala/collection/Iterator.scala#L432

Which one do you think is better?

linbojin: @srowen I think i figured out a simpler way:
```
  def flatten[U: ClassTag](implicit asTraversable: T => TraversableOnce[U]): RDD[U] = withScope {
    new MapPartitionsRDD[U, T](this, (context, pid, iter) => {
      var newIter: Iterator[U] = Iterator.empty
      for (x <- iter) newIter ++= asTraversable(x)
      newIter
    })
  }
```

srowen:
Yes you're hitting a key point, that `flatten` only makes sense on an RDD of `TraversableOnce`. I think we'd implement this in the way that `DoubleRDDFunctions` adds operations for only RDDs of doubles instead of adding this implicit transformation.

In that case you can get an `iterator` directly from each element, and then concatenating them. I think it goes something like:

```
iter => iter.foldLeft(Iterator.empty)((a,b) => a ++ b.toIterator)
```

... but I'm sure that's not 100% correct.

Anyway the complexity makes me wonder if it's worth doing unless the lack of a direct method is a bottleneck. (We could also look at whether Spark internally could replace flatMap(x => x) with the new method)

linbojin:
@srowen So any problems with my current implementation, i mean use `implicit asTraversable: T => TraversableOnce[U]`. Because I refer to `flatten` implementation in scala source code: https://github.com/scala/scala/blob/05016d9035ab9b1c866bd9f12fdd0491f1ea0cbb/src/library/scala/collection/generic/GenericTraversableTemplate.scala#L169
and if user call flatten on invalid rdd, it will just error out:
```
scala> val rdd = sc.parallelize(Seq(1,2,3))
rdd: org.apache.spark.rdd.RDD[Int] = ParallelCollectionRDD[0] at parallelize at <console>:24

scala> rdd.flatten
<console>:27: error: No implicit view available from Int => scala.collection.TraversableOnce[U].
       rdd.flatten
           ^
```
same as scala
```
scala> val l = List(1,2,3)
l: List[Int] = List(1, 2, 3)

scala> l.flatten
<console>:13: error: No implicit view available from Int => scala.collection.GenTraversableOnce[B].
       l.flatten
         ^
```

rxin:
Do we really need to add this? For the time we spent we can work on more impactful things ...

linbojin:
@rxin Thx, I just deep dive into spark, I hope i can try to contribute more impactful things later 😄 . I will close this one. @srowen I tried out your idea that i created `TraversableRDDFunctions` file and `implicit def rddToTraversableRDDFunctions[U](rdd: RDD[TraversableRDDFunctions[U]])` inside RDD object. It's not very complex, but the problem is class RDD is **invariant** so that it's hard to make this method generic.

rxin:
Thanks @linbojin !

PR #16278, opened 2016-12-14
[SPARK-18779][STREAMING][KAFKA] Messages being received only from one partition when using Spark Streaming integration for Kafka 0.10 with kafka client library at 0.10.1

pnakhe:

This pull request is to fix for SPARK-18779. When using kafka 0.10.1.0 messages are being read only from one partition. The current kafka-spark 0.10 integration ships with kafka 0.10.0.1 where messages are read from all partitions but using kafka client 0.10.1.0 client, messages are read from only one partition.

In the ConsumerStrategy class there is a pause on the consumer. We never resume the consumer and that seems to causing the issue. The KafkaConsumer implementation has changed between 10.0.1 and 10.1.0 which has exposed this issue. The solution to this issue is to resume the consumer before we find the position in DirectKafkaInputDStream class in the latestOffsets method.The reason the issue is not seen in the current setup is because pause/resume logic is changed in the latest kafka version. We dont seem to have a resume for the pause and hence this fix is necessary.

This patch fixes the issue.

## How was this patch tested?
The spark-kafka test cases were run to check no regressions were caused. I have checked that messages are being read from all partitions for both 0.10.0.1 kafka client and 0.10.1.0 client.

### DIFF
--- a/external/kafka-0-10/src/main/scala/org/apache/spark/streaming/kafka010/DirectKafkaInputDStream.scala
+++ b/external/kafka-0-10/src/main/scala/org/apache/spark/streaming/kafka010/DirectKafkaInputDStream.scala
@@ -195,6 +195,8 @@ private[spark] class DirectKafkaInputDStream[K, V](
     c.pause(newPartitions.asJava)
     // find latest available offsets
     c.seekToEnd(currentOffsets.keySet.asJava)
+    c.resume(newPartitions.asJava)
+    c.resume(c.assignment())
     parts.map(tp => tp -> c.position(tp)).toMap
   }
 


zsxwing:
> pause/resume logic is changed in the latest kafka version

Could you post the Kafka JIRA for this change? Just want to understand the issue.

HyukjinKwon:
(@pnakhe gentle ping, I am curious too)

pnakhe:
@HyukjinKwon Well the issue was not with spark after all. It was a regression on kafka between 0.10.1.0 and 0.10.1.1. Its fixed as part of https://issues.apache.org/jira/browse/KAFKA-4547

I have updated the defect with the same

HyukjinKwon:
Aha, thanks for the details, then is this PR/JIRA closable maybe?

PR #16279, opened 2016-12-14
[SPARK-18471][MLLIB][BACKPORT-2.0] In LBFGS, avoid sending huge vectors of 0 

AnthonyTruchet:
_Backport #16037 to 2.0 branch_

# What changes were proposed in this pull request? #

treeAggregate call. To avoid that, we change the aggregation operations
to convert sparse vectors into dense vectors on the fly if needed and we 
pass a sparse 0 vector which is lightweight.

# How was this patch tested? #

Author: Anthony Truchet <a.truchet@criteo.com>
Author: sethah <seth.hendrickson16@gmail.com>

### DIFF
--- a/mllib/src/main/scala/org/apache/spark/mllib/optimization/LBFGS.scala
+++ b/mllib/src/main/scala/org/apache/spark/mllib/optimization/LBFGS.scala
@@ -246,16 +246,24 @@ object LBFGS extends Logging {
       val bcW = data.context.broadcast(w)
       val localGradient = gradient
 
-      val (gradientSum, lossSum) = data.treeAggregate((Vectors.zeros(n), 0.0))(
-          seqOp = (c, v) => (c, v) match { case ((grad, loss), (label, features)) =>
-            val l = localGradient.compute(
-              features, label, bcW.value, grad)
-            (grad, loss + l)
-          },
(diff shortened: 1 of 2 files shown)


srowen:
I don't think we would backport this to even 2.1. You can close this.

AnthonyTruchet:
@srowen sorry to bother you but this is just the backport (without any conflict) of #16037. Could you please validate it ?

AnthonyTruchet:
May I ask why ? There was no conflicts so no additional qualification work is required and this looks like a performance bug fix to me, not a new feature. In order to adjust our contribution policy  (fix to our internal version vs pushing upstream) we would need to understand better the backporting policy: would you have any more detailed pointer at it ?

srowen:
The guidance such as it is is here: http://spark.apache.org/versioning-policy.html
It is in truth mostly a judgment call. Improvements are rarely back-ported to a maintenance branch, but it depends on impact vs risk. This is low-risk, but also seems minor in impact too. I'd back-port if someone else would second that.

AnthonyTruchet:
Ok, thanks for the pointer. I do agree this is a non critical judgement call .

sethah:
Can we close it?

AnthonyTruchet:
We have backported it to our internal version of Spark anyhow. As told above feel free to close it if you consider this is not worth officially backporting :-)

PR #16280, opened 2016-12-14
[SPARK-18856][SQL] non-empty partitioned table should not report zero size

cloud-fan:

In `DataSource`, if the table is not analyzed, we will use 0 as the default value for table size. This is dangerous, we may broadcast a large table and cause OOM. We should use `defaultSizeInBytes` instead.

## How was this patch tested?

new regression test

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSourceStrategy.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSourceStrategy.scala
@@ -303,7 +303,8 @@ class FindDataSourceTable(sparkSession: SparkSession) extends Rule[LogicalPlan] 
         partitionColumns = table.partitionColumnNames,
         bucketSpec = table.bucketSpec,
         className = table.provider.get,
-        options = table.storage.properties ++ pathOption)
+        options = table.storage.properties ++ pathOption,
+        catalogTable = Some(simpleCatalogRelation.metadata))
 
     LogicalRelation(
       dataSource.resolveRelation(),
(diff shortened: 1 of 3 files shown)


cloud-fan:
cc @rxin @ericl

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSourceStrategy.scala
```diff
@@ -303,7 +303,8 @@ class FindDataSourceTable(sparkSession: SparkSession) extends Rule[LogicalPlan]
         partitionColumns = table.partitionColumnNames,
         bucketSpec = table.bucketSpec,
         className = table.provider.get,
-        options = table.storage.properties ++ pathOption)
+        options = table.storage.properties ++ pathOption,
+        catalogTable = Some(simpleCatalogRelation.metadata))
```
cloud-fan: Previously we will always use `InMemoryFileIndex` because we don't pass the catalog table to `DataSource`...

ericl: Any way to catch this in a test? Presumably this would have caused all the files to be scanned.

ericl (approved):

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSourceStrategy.scala
cloud-fan: oh i know why. The `InMemoryExternalCatalog` hasn't implemented all the interfaces(e.g. some partition related ones), so we did it intentionally.

rxin:
Merging in master/branch-2.1.

PR #16281, opened 2016-12-14
[SPARK-13127][SQL] Update Parquet to 1.9.0

dongjoon-hyun:

This PR aims to update Parquet to 1.9.0 and remove the hacks due to Parquet 1.8.1 limitation.

```scala
-  // !! HACK ALERT !!
-  //
-  // PARQUET-363 & PARQUET-278: parquet-mr 1.8.1 doesn't allow constructing empty GroupType,
-  // which prevents us to avoid selecting any columns for queries like `SELECT COUNT(*) FROM t`.
-  // This issue has been fixed in parquet-mr 1.8.2-SNAPSHOT.
-  //
-  // To workaround this problem, here we first construct a `MessageType` with a single dummy
-  // field, and then remove the field to obtain an empty `MessageType`.
-  //
-  // TODO Reverts this change after upgrading parquet-mr to 1.8.2+
   val EMPTY_MESSAGE = Types
       .buildMessage()
-      .required(PrimitiveType.PrimitiveTypeName.INT32).named("dummy")
       .named(ParquetSchemaConverter.SPARK_PARQUET_SCHEMA_NAME)
-  EMPTY_MESSAGE.getFields.clear()
```

## How was this patch tested?

Pass the existing tests.


sql/core/src/test/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetSchemaSuite.scala
```diff
@@ -1424,7 +1424,7 @@ class ParquetSchemaSuite extends ParquetSchemaTest {
 
     catalystSchema = new StructType(),
 
-    expectedSchema = ParquetSchemaConverter.EMPTY_MESSAGE)
+    expectedSchema = "message root {}")
```
viirya: Why this change? Don't we expect `EMPTY_MESSAGE`?

HyukjinKwon: +1

srowen:
The tricky part about upgrading Parquet is whether it will continue to work with other dependencies that work with Parquet. It's also worth considering that a different Parquet may be on the classpath at runtime, certainly for older versions of Hadoop. 

We should probably proceed with this if it's fixing important bugs. It's worth scanning through what changes went in from 1.8 to 1.9 to see if anything looks like a potential behavior change or compatibility issue.

dongjoon-hyun:
Thank you for review, @srowen . I see.

First of all, I'll check the case of the mixed parquet jar files in the class path.

Second, among [83 issues](https://issues.apache.org/jira/issues/?jql=project%20%3D%20PARQUET%20AND%20status%20%3D%20Resolved%20AND%20resolution%20%3D%20Fixed%20AND%20fixVersion%20%3D%201.9.0) fixed by Parquet 1.9.0, the following seems to be notable.
- PARQUET-99: Large rows cause unnecessary OOM exceptions
- PARQUET-225: INT64 support for Delta Encoding
- PARQUET-341: Improve write performance with wide schema sparse data
- PARQUET-353: Compressors not getting recycled while writing parquet files, causing memory leak
- PARQUET-363: Cannot construct empty MessageType for ReadContext.requestedSchema
- PARQUET-415: ByteBufferBackedBinary serialization is broken
- PARQUET-511: Integer overflow on counting values in column
- PARQUET-569: ParquetMetadataConverter offset filter is broken
- PARQUET-571: Fix potential leak in ParquetFileReader.close()
- PARQUET-623: DeltaByteArrayReader has incorrect skip behaviour
- PARQUET-645: DictionaryFilter incorrectly handles null

@rdblue , could you give us some advice for compatibility issue between Parquet 1.8 and 1.9?

rxin:
I'm actually wondering if we should just fork Parquet and maintain a version of it, and add fixes ourselves to the fork. In the past Parquet updates often bring more regressions ...

dongjoon-hyun:
Thank you for review, @rxin .
Forking may give more controllability, but it goes invisible (in terms of documents) soon.
According to the recent mail on Spark dev list, only committers seem to know the latest repository location of Spark Hive fork.
I also want to contribute some on that, but it's difficult for me to find how to do.

Every Apache projects (including Apache Spark) have some bugs at every release.
I don't think there is no bug on Parquet 1.9.0. But, Parquet community exists to improve that, doesn't it?

BTW, @rxin and @srowen . To reduce the risk,
- Do you want to add more Spark-side test case here?
- Or, do you prefer to skip 1.9.0 and go 1.10 directly?

gatorsmile:
Parquet is the default format of Spark. It is pretty significant to Spark. Now, Parquet is becoming stable and might be the right time to fork it. We are just fixing the bugs. @liancheng and @rdblue are Parquet committers. They might be the right person to judge the changes we made in the forked version.

srowen:
Forking is a very bad thing, and only a last resort. We haven't properly managed the fork of Hive yet, even. I don't hear specific bugs to fork around here either. As such I can't see why this would even be considered.

gatorsmile:
We keep hitting the issues in Parquet. Below is another example: https://issues.apache.org/jira/browse/SPARK-18539

gatorsmile:
One more example: https://github.com/apache/spark/pull/16106 This issue degrades the performance.

viirya:
Those issues with Parquet are specified for certain Parquet versions. If upgrading Parquet can solve them, it can't justify the decision to fork Parquet. To fork such project we need more consideration about the pros and cons.

srowen:
Yes, but we face bugs in third-party components all the time and work around them or get them fixed. There is an unmentioned downside here too: not getting bug fixes and improvements that don't (yet) affect Spark. Running a fork risks incompatibility at the data format level, where it matters most to interop. I think this really can't be taken lightly -- better to find ways to increase integration testing or influence on Parquet or whatever to get things fixed faster where they bite Spark.

nsyca:
My two cents:
- Do we have a Parquet specific test suite **with sufficient coverage** to run and back us up that this upgrade won't cause any regressions? I think simply moving up the version of the jar files is a risky act. This practice of doing (sort of) integration test will gain the confidence of our user community that they can count on Spark to exercise its due diligence when it changes the versions of any third party modules Spark runs on. Yes, the activity comes with a cost. We can always define how much we can test and we want to test.
- On the topic of forking, it is a judgment call. It's a balance of having a full control on the dependent third party modules but deviating from their origin versus doing little work on our end but risking any contamination. In the world of interdependence and interconnect, my opinion leans towards "good fences make good neighbours." That comes back to my first point, we need to have a good test coverage to gauge the impact of an upgrade to Spark.

gatorsmile:
@srowen Even if we fork our own version, it does not mean we will give up the upgrading to the newer version. We just added a few fixes. 

This is very normal in the mission-critical system. When the customers on mainframe hit a bug, they will not upgrade it to the newer major version. Normally, they can get the special build with the needed fixes. Parquet community will not do this for us, but we can do it by ourselves especially when we have the Parquet experts @liancheng @rdblue

srowen:
@gatorsmile @rdblue also works directly on Parquet. I am not seeing "unfixable" Parquet problems here. You're just pointing at problems that can and should be fixed, preferably in one place. Forking is not at all normal as a response to this.

This, at least, must block on us figuring out how to manage forks. The Hive fork is still not really technically OK.

gatorsmile:
The problem is the Parquet community will not create a branch 1.8.2+ for us. Upgrading to newer versions 1.9 or 2.0 are always risky. Based on the history, we hit the bugs and performance degradation, when we try to upgrade the Parquet versions. We need more time and efforts to decide whether we need to upgrade to the newer Parquet version.

srowen:
I'd much rather lobby to release 1.8.2 and help with the legwork than do all that legwork and more to maintain a fork. It's still not clear to me that upgrading to 1.9.0 is not a solution?

gatorsmile:
Basically, the idea is to make a special build for Parquet 1.8.1 with the needed fixes by ourselves.

Upgrading to newer version like Parquet 1.9.0 is risky. Parquet 1.9.0 was just released this Oct. The bugs might not be exposed until more users/clients start using it. This is also the same to our Spark community. I personally do not suggest any enterprise customer use Spark 2.0.0, even if it resolved many bugs in Spark 1.6+

To evaluate whether upgrading Parquet 1.9.0., the biggest effort is the performance evaluation. We need to have our own standard performance workload benchmarks (TPC-DS is not enough) to test whether the upgrade can introduce any major performance degradation.

dongjoon-hyun:
Actually, this PR is about Apache Spark 2.2 on Late March in terms of RC1.
We have a lot of time to discuss. Why don't we discuss that on dev mailing list?

gatorsmile:
@dongjoon-hyun What kind of questions/requests should we ask in dev mailing list? 

IMO, the risk and cost are small if we make a special build by ourselves. We can get the bug fixes very quickly. Maybe @rxin @rdblue @liancheng can give their inputs here?

dongjoon-hyun:
I agree, but, in a long term perspective, the risk and cost of forking could be the worst.

gatorsmile:
I think we are not adding new features into Parquet. The fixes must be small. To avoid the cost and risk, we need to reject all the major fixes in our special build. At the same time, we also need to request Parquet community to resolve the bugs in the newer releases.

dongjoon-hyun:
Yep. At the beginning, it starts like that. But, please look at Spark Hive or Spark Thrift Server. I don't think we are maintaining that well or visibly.

gatorsmile:
We are adding major code changes in Spark Thrift Server? 

What is the Spark Hive?

srowen:
Has anyone even asked for a new 1.8.x build from Parquet and been told it won't happen?

You don't stop consuming non fix changes by forking. You do that by staying on a maintenance branch. If that branch is maintained of course. I'd be shocked if there were important bugs affecting a major ASF project and no way to get a maintenance release of a recent branch. Esp when we have experts here with influence. 

What does this block - why the urgency?

dongjoon-hyun:
Yep. Spark Thrift Server are different, but it's not actively maintained. For example, the default database feature is recently added.

I mean this one by `Spark Hive`.
```
<hive.version>1.2.1.spark2</hive.version>
```

rxin:
We haven't really added much to Hive though, and as a matter of fact the dependency on Hive is decreasing. Parquet is a much more manageable piece of code to fork. In the past we have seen fairly critical bugs with almost every upgrade. and coupled with the fact that Parquet cannot always make releases fast enough (yes it happened in the past when we asked to have released but didn't get them), or have proper testing, it has always been very risky to just upgrade a major version of Parquet. In addition, we already have a forked Parquet reader in Spark that is vectorized (that is different from the one in parquet mr).

rxin:
Btw issues are not just performance, but often correctness as well. As the default format, a bug in Parquet is much worse than a bug in say ORC.

gatorsmile:
@dongjoon-hyun 

I just checked the code changes in `1.2.1.spark2` compared with the official Hive 1.2.1: https://github.com/JoshRosen/hive/commits/release-1.2.1-spark2 

Very small changes, right?

rxin:
@nsync you raised an excellent question on test coverage. The kind of bugs we have seen in the past weren't really integration bugs, but bugs in parquet-mr. Technically it should be the jobs of parquet-mr to verify correctness and performance regressions. If we were to introduce a much more broader set of regression tests in Spark, then to me it makes even more sense to just move the Parquet code into Spark and fixed issues found there. 

Also I have spent some time understanding the Parquet codec, and I have to say it is pretty powerful and complicated and as a result fairly difficult to implement correctly. The dremel format optimizes for sparse nested data, but is much more difficult to get right than a simpler dense format. 

FWIW, the ideal scenario I can think of is to have parquet-mr publish big fix versions that don't include new features. That would make update auditing easier and updates lower risk. 

E.g. Parquet-mr 2 adds new features, and 1.x are just bug fixes.

mridulm:
I agree with @srowen, forking should be the last resort.

ash211:
What are the specific patches to parquet that folks are proposing should be included in a parquet 1.8.1-spark1 ?  Or what would be desired in a parquet-released 1.8.2 ?

dongjoon-hyun:
For this issue, it was initially PARQUET-363, but all maintenance fix (with new features) are welcome.

For me, the followings?

- PARQUET-99: Large rows cause unnecessary OOM exceptions
- PARQUET-353: Compressors not getting recycled while writing parquet files, causing memory leak
- PARQUET-363: Cannot construct empty MessageType for ReadContext.requestedSchema
- PARQUET-511: Integer overflow on counting values in column
- PARQUET-569: ParquetMetadataConverter offset filter is broken
- PARQUET-571: Fix potential leak in ParquetFileReader.close()
- PARQUET-623: DeltaByteArrayReader has incorrect skip behaviour
- PARQUET-645: DictionaryFilter incorrectly handles null

@gatorsmile . Do you have more?

gatorsmile:
@dongjoon-hyun I am not sure what is your search conditions. In the recent JIRA, we hit this issue: 
- PARQUET-389: Filter predicates should work with missing columns

gatorsmile:
There is still an open JIRA: [PARQUET-686 Allow for Unsigned Statistics in Binary Type](https://issues.apache.org/jira/browse/PARQUET-686)

Because of that bug, we disable the filter push down for String and Binary. For details, please see https://github.com/apache/spark/pull/16106

dongjoon-hyun:
Is the open issue the main purpose of forking?
To resolve it faster than Parquet community?

gatorsmile:
That is just an example. We definitely need the fix ASAP, right? See[ the past release dates](http://search.maven.org/#search%7Cgav%7C1%7Cg%3A%22org.apache.parquet%22%20AND%20a%3A%22parquet%22) of Apache Parquet (that is just the one I got from the web, it might not be 100% accurate):

- 1.9.0: Oct. 2016
- 1.8.1: Jul. 2015
- 1.8.0: Jul. 2015
- 1.7.0: May 2015

dongjoon-hyun:
I see the point. Then, the forked repository is going for 2.1.1 or 2.1-rc4 ASAP?

rxin:
I don't think we are making a decision at this point to fork ... if we can really push parquet to make another maintenance bug fix release, that'd be great.

dongjoon-hyun:
+1

rdblue:
I don't think a fork is a good idea, nor do I think there is a reasonable need for one.

@gatorsmile brought up that the Parquet community refused to build a patch release: "The problem is the Parquet community will not create a branch 1.8.2+ for us." I don't remember this happening, and as far as I can tell from both google and my inbox, the Parquet community never rejected the idea of patch releases. If there was a conversation that I don't know about, then I apologize that you were given the impression that patch releases aren't possible. That isn't the case.

I'm happy to work with the community to put out patch releases, especially if that's needed for Spark. To demonstrate, look at PARQUET-389 and PARQUET-654. @rxin asked the Parquet dev list about predicate push-down features and within a week and a half, both of those issues were resolved. (PARQUET-389 is the fix for SPARK-18539, cited as motivation to fork.)

As for the other motivating issue, PARQUET-686, a fork can't help solve this problem. This is an issue that requires updating the Parquet format spec so you couldn't simply fix your own fork without abandoning compatibility. The Parquet community put out a release that gives the user a choice between correctness and performance, which is a good compromise until this can be fixed.

It is fair to point out that Parquet has not had a regular release cadence for minor releases (1.8.1 to 1.9.0), which is something that the Parquet community knows about and has discussed. We have recently committed to quarterly releases to fix this, with patch releases whenever they are needed. I'd encourage anyone interested to get involved.

julienledem:
Like @rdblue said, I don't recall people asking for a 1.8.2 on the parquet dev list.
We are happy to help and if there is a 1.8.x patch release branch it's better to maintain it in the parquet repo. 
We can create that branch and will be happy to accept your patches for small bug fixes there.

mallman:
I'd love to see frequent, conservative patch releases. From my experience, parquet bugs cause significant trouble for downstream consumers. For example, we encountered a data corruption bug writing parquet format V2 that made a good deal of that data unreadable. (I can't recall the issue number, but it was reported and fixed.)

As another example, we recently upgraded some of our Spark clusters to use parquet-mr 1.9.0, because PARQUET-363 introduced a bug into one of our Spark 2.x patches. When we switched to 1.9, we found https://issues.apache.org/jira/browse/PARQUET-783, which breaks things in a different way. We needed a fix, so we forked 1.9 internally.

FWIW, we haven't found any other issues using parquet-mr 1.9.0 with Spark 2.1.

gatorsmile:
@julienledem @rdblue wow, that is so great!!! It will be much easier for us!

I never expected the Parquet community is willing to do the regular patch releases. Sorry, that statement is just my personal assumption based on the release history, since I did not see the backport PRs. If Parquet community can do the quarterly fixpack release, it makes our maintenance work much easier. Thank you very much!!!

rdblue:
Great! I'm glad it was just confusion. I completely agree with @srowen that forking should be a last resort.

In the future, please reach out to the community, whether its Parquet or another, to address concerns before it gets this far. It's better for everyone if we can use the feedback to improve and continue to work together.

gatorsmile:
@rdblue Thanks again! : )

viirya:
As we understand the Parquet community is willing to put out patch releases, I don't see any major reason now to motivate the forking. That is a great news for us. So do we postpone the upgrading to Parquet 1.9.0 and wait for upgrading to a patch release in the near future?

dongjoon-hyun:
Sure. We are going to wait for Parquet 1.8.2.
Maybe, in a month for Spark 2.1.1 on January?
Anyway, thank you all! I'm closing this PR happily.

rdblue:
I think we should move to a 1.8.2 patch release. The reason is that 1.9.0 moved to ByteBuffer based reads and we've found at least one problem with it. ByteBuffer based reads also changes an internal API that Spark uses for its vectorized reader. The changes here (now that I've looked at the actual PR) wouldn't work because Parquet is going to call the ByteBuffer method rather than the byte array method. I'm cleaning up the ByteBuffer code quite a bit right now for better performance with G1GC (see https://github.com/apache/parquet-mr/pull/390), so I think Spark should move to ByteBuffer reads after that makes it in.

rdblue:
Thanks @dongjoon-hyun! Lets get a Parquet 1.8.2 out in January.

dongjoon-hyun:
Thank you for confirming, @rdblue .

kiszk:
@rdblue Interesting, do you have any estimated or actual data for performance improvement?
I am interested in how you can achieve performance improvement over `byte[]`.

- Usage of `ByteBuffer`
  Usage of direct `ByteBuffer`
  Usage of multiple buffers described at [the PR](https://github.com/apache/parquet-mr/pull/390)

rdblue:
The improvement is in how row groups are garbage collected. G1GC puts humongous allocations directly into the old generation, so you end up needing a full GC to reclaim the space. That just increases memory pressure, so you run out of memory and run full GCs and/or spill to disk. We don't have data yet because I haven't pushed the feature or metrics collection for it.

julienledem:
FYI: Parquet 1.8.2 vote thread passed: https://mail-archives.apache.org/mod_mbox/parquet-dev/201701.mbox/%3CCAO4re1mHLT%2BLYn8s1RTEDZK8-9WSVugY8-HQqAN%2BtU%3DBOi1L9w%40mail.gmail.com%3E

dongjoon-hyun:
Hi, all.
Now, I'm trying to upgrade Apache Spark to 1.8.2.

mallman:
FYI, we've been using 1.9.0 patched with a fix for https://issues.apache.org/jira/browse/PARQUET-783 without problem.

dongjoon-hyun:
Thank you for sharing that information, @mallman .

PR #16282, opened 2016-12-14
[SPARK-18588][SS][Kafka]Create a new KafkaConsumer when error happens to fix the flaky test

zsxwing:

When KafkaSource fails on Kafka errors, we should create a new consumer to retry rather than using the existing broken one because it's possible that the broken one will fail again.

This PR also assigns a new group id to the new created consumer for a possible race condition:  the broken consumer cannot talk with the Kafka cluster in `close` but the new consumer can talk to Kafka cluster. I'm not sure if this will happen or not. Just for safety to avoid that the Kafka cluster thinks there are two consumers with the same group id in a short time window. (Note: CachedKafkaConsumer doesn't need this fix since `assign` never uses the group id.)

## How was this patch tested?

In https://amplab.cs.berkeley.edu/jenkins/job/SparkPullRequestBuilder/70370/console , it ran this flaky test 120 times and all passed.

### DIFF
--- a/dev/sparktestsupport/modules.py
+++ b/dev/sparktestsupport/modules.py
@@ -245,7 +245,8 @@ streaming_kafka_0_10 = Module(
     name="streaming-kafka-0-10",
     dependencies=[streaming],
     source_file_regexes=[
-        "external/kafka-0-10",
+        # The ending "/" is necessary otherwise it will include "sql-kafka" codes
+        "external/kafka-0-10/",
         "external/kafka-0-10-assembly",
     ],
     sbt_test_goals=[
--- a/external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/KafkaSource.scala
+++ b/external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/KafkaSource.scala
@@ -538,14 +538,6 @@ private[kafka010] object KafkaSource {
 
   sealed trait ConsumerStrategy {
     def createConsumer(kafkaParams: ju.Map[String, Object]): Consumer[Array[Byte], Array[Byte]]
-
-    def newKafkaParams(
-        kafkaParams: ju.Map[String, Object],
-        groupId: String): ju.Map[String, Object] = {
-      val newKafkaParams = new ju.HashMap[String, Object](kafkaParams)
-      newKafkaParams.put(ConsumerConfig.GROUP_ID_CONFIG, groupId)
-      newKafkaParams
-    }
   }
 
   case class AssignStrategy(partitions: Array[TopicPartition]) extends ConsumerStrategy {
--- a/external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/KafkaSource.scala
+++ b/external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/KafkaSource.scala
@@ -24,7 +24,7 @@ import java.nio.charset.StandardCharsets
 import scala.collection.JavaConverters._
 import scala.util.control.NonFatal
 
-import org.apache.kafka.clients.consumer.{Consumer, KafkaConsumer, OffsetOutOfRangeException}
+import org.apache.kafka.clients.consumer.{Consumer, ConsumerConfig, KafkaConsumer, OffsetOutOfRangeException}
 import org.apache.kafka.clients.consumer.internals.NoOpConsumerRebalanceListener
 import org.apache.kafka.common.TopicPartition
 
@@ -81,14 +81,16 @@ import org.apache.spark.util.UninterruptibleThread
  * To avoid this issue, you should make sure stopping the query before stopping the Kafka brokers
  * and not use wrong broker addresses.
  */
-private[kafka010] case class KafkaSource(
+private[kafka010] class KafkaSource(
     sqlContext: SQLContext,
     consumerStrategy: ConsumerStrategy,
+    driverKafkaParams: ju.Map[String, Object],
     executorKafkaParams: ju.Map[String, Object],
     sourceOptions: Map[String, String],
     metadataPath: String,
     startingOffsets: StartingOffsets,
-    failOnDataLoss: Boolean)
+    failOnDataLoss: Boolean,
+    driverGroupIdPrefix: String)
   extends Source with Logging {
 
   private val sc = sqlContext.sparkContext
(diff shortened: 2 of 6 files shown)


zsxwing:
retest this please

zsxwing:
retest this please

zsxwing:
retest this please

external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/KafkaSource.scala
```diff
@@ -81,14 +81,16 @@ import org.apache.spark.util.UninterruptibleThread
  * To avoid this issue, you should make sure stopping the query before stopping the Kafka brokers
  * and not use wrong broker addresses.
  */
-private[kafka010] case class KafkaSource(
+private[kafka010] class KafkaSource(
```
zsxwing: Remove `case` to make `equals` faster since this class has implemented its `toString`.

external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/KafkaSource.scala
```diff
@@ -441,13 +468,12 @@ private[kafka010] case class KafkaSource(
               try {
                 result = Some(body)
               } catch {
-                case x: OffsetOutOfRangeException =>
```
zsxwing: `OffsetOutOfRangeException` only happens when we call `seek` than `poll`. So it won't happen here. It's pretty confusing as it doesn't update `attempt`, so I just removed it.

external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/KafkaSource.scala
```diff
@@ -81,14 +81,16 @@ import org.apache.spark.util.UninterruptibleThread
  * To avoid this issue, you should make sure stopping the query before stopping the Kafka brokers
  * and not use wrong broker addresses.
  */
-private[kafka010] case class KafkaSource(
```
zsxwing: Remove `case` to make `equals` faster since this class has implemented its `toString`.

external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/KafkaSource.scala
zsxwing: `OffsetOutOfRangeException` only happens when we call `seek` than `poll`. So it won't happen here. It's pretty confusing as it doesn't update attempt, so I just removed it.

zsxwing:
retest this please

zsxwing:
retest this please

zsxwing:
SPARK-18907

zsxwing:
retest this please

zsxwing:
retest this please

zsxwing:
cc @tdas

external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/KafkaSource.scala
```diff
@@ -536,11 +570,12 @@ private[kafka010] object KafkaSource {
...
-  case class SubscribePatternStrategy(
-    topicPattern: String, kafkaParams: ju.Map[String, Object])
+  case class SubscribePatternStrategy(topicPattern: String)
     extends ConsumerStrategy {
-    override def createConsumer(): Consumer[Array[Byte], Array[Byte]] = {
-      val consumer = new KafkaConsumer[Array[Byte], Array[Byte]](kafkaParams)
+    override def createConsumer(
+        kafkaParams: ju.Map[String, Object]): Consumer[Array[Byte], Array[Byte]] = {
+      val consumer =
+        new KafkaConsumer[Array[Byte], Array[Byte]](kafkaParams)
```
brkyvz: nit: move to line above

external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/KafkaSource.scala
```diff
@@ -511,12 +537,20 @@ private[kafka010] object KafkaSource {
   ))
 
   sealed trait ConsumerStrategy {
-    def createConsumer(): Consumer[Array[Byte], Array[Byte]]
+    def createConsumer(kafkaParams: ju.Map[String, Object]): Consumer[Array[Byte], Array[Byte]]
+
+    def newKafkaParams(
+        kafkaParams: ju.Map[String, Object],
+        groupId: String): ju.Map[String, Object] = {
+      val newKafkaParams = new ju.HashMap[String, Object](kafkaParams)
```
tdas: Nit: please dont reuse the name of the function as a variable. its confusing while reading code.

external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/KafkaSource.scala
```diff
@@ -511,12 +537,20 @@ private[kafka010] object KafkaSource {
   ))
 
   sealed trait ConsumerStrategy {
-    def createConsumer(): Consumer[Array[Byte], Array[Byte]]
+    def createConsumer(kafkaParams: ju.Map[String, Object]): Consumer[Array[Byte], Array[Byte]]
+
+    def newKafkaParams(
```
tdas: What is the need of this new method in ConsumerStrategy. Does not seem to be overridden or used any where?

external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/KafkaSource.scala
tdas: But does hashcode use the toString to generate unique hashcode?

tdas: This is needed in StreamExecution for distinctSources, etc.

tdas: What is the performance concern here?

tdas (changes requested):
Overall approach looks good. Few questions though.

external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/KafkaSource.scala
zsxwing: Shouldn't the default hashCode is enough?

dev/sparktestsupport/modules.py
```diff
@@ -245,7 +245,8 @@ def __hash__(self):
     name="streaming-kafka-0-10",
     dependencies=[streaming],
     source_file_regexes=[
-        "external/kafka-0-10",
+        # The ending "/" is necessary otherwise it will include "sql-kafka" codes
+        "external/kafka-0-10/",
```
zsxwing: Fixed this to avoid triggering dstream kafka tests

tdas: Great!

external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/KafkaSource.scala
tdas: the unique id of the object? maybe. I am not sure. if the same kafkaDF is used twice in a query (say, union two duplicate the data), will it actually create two distinct source objects, or reuse the same one. And in that case do we want them to match or not?

zsxwing:
> Test build #70439 has finished for PR 16282 at commit 7c789e8.

This is DStream failure: DirectKafkaStreamSuite

> Test build #70444 has finished for PR 16282 at commit 9080acd.

Looks like Python tests were hung. But that's not related to this PR.

zsxwing:
retest this please

tdas:
Merging to master and 2.1

PR #16285, opened 2016-12-15
[SPARK-18867] [SQL] Throw cause if IsolatedClientLoad can't create cl…

jojochuang:
If IsolatedClientLoader can't instantiate a class object, it throws InvocationTargetException. But the caller doesn't need to know this exception. Instead, it should throw the exception that causes the InvocationTargetException, so that the caller may be able to handle it.

## How was this patch tested?

(If this patch involves UI changes, please attach a screenshot; otherwise, remove this)

…ient


rxin:
Is it ever possible cause is null?

jojochuang:
Good point @rxin. That seems possible. 
https://docs.oracle.com/javase/7/docs/api/java/lang/reflect/InvocationTargetException.html#getCause()

rxin:
Thanks for looking into it. In that case, maybe we should just keep the thing as is and don't change it?

HyukjinKwon:
Hi @jojochuang, is it still active? If so, I think we should address this comment ^.

PR #16288, opened 2016-12-15
[SPARK-18869][SQL] Add TreeNode.p that returns BaseType

rxin:
After the bug fix in SPARK-18854, TreeNode.apply now returns TreeNode[_] rather than a more specific type. It would be easier for interactive debugging to introduce a function that returns the BaseType.

## How was this patch tested?
N/A - this is a developer only feature used for interactive debugging. As long as it compiles, it should be good to go. I tested this in spark-shell.

### DIFF
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/QueryPlan.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/QueryPlan.scala
@@ -24,15 +24,6 @@ import org.apache.spark.sql.types.{DataType, StructType}
 abstract class QueryPlan[PlanType <: QueryPlan[PlanType]] extends TreeNode[PlanType] {
   self: PlanType =>
 
-  /**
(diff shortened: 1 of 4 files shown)


yhuai:
lgtm

srowen:
@rxin I'm not entirely sure, but it sort of looks like this causes branch-2.0 to fail:
https://amplab.cs.berkeley.edu/jenkins/view/Spark%20QA%20Test%20(Dashboard)/

Reading backwards, it seems like some failures were caused by numberedTreeString but then fixed, but this is the first one I can see in common across the various jobs that have failed since.

Here's some output; I don't yet know if it rules this change in or out:
https://amplab.cs.berkeley.edu/jenkins/view/Spark%20QA%20Test%20(Dashboard)/job/spark-branch-2.0-test-sbt-hadoop-2.7/865/consoleFull

rxin:
@srowen it looks like 2.0 has a different plan from master. I pushed a fix.

PR #16290, opened 2016-12-15
[SPARK-18817] [SPARKR] [SQL] Set default warehouse dir to tempdir

shivaram:
This PR sets the default warehouse dir to a temporary directory in SparkR to avoid creating directories in the working directory (see JIRA for more details).

To do this we introduce a new SQL config that is used to configure the default. For all other frontends, existing behavior is maintained

## How was this patch tested?
Running unit tests locally. Manually with SparkR shell


shivaram:
cc @bdwyer2  - who did an initial version at #16247 
cc @felixcheung @cloud-fan for review

R/pkg/R/sparkR.R
```diff
@@ -363,6 +363,13 @@ sparkR.session <- function(
   ...) {
 
   sparkConfigMap <- convertNamedListToEnv(sparkConfig)
+
+  # NOTE(shivaram): Set default warehouse dir to tmpdir to meet CRAN requirements
+  # See SPARK-18817 for more details
+  if (!exists("spark.sql.default.warehouse.dir", envir = sparkConfigMap)) {
+    assign("spark.sql.default.warehouse.dir", tempdir(), envir = sparkConfigMap)
```
felixcheung: I think we could just `sparkConfigMap[["spark.sql.warehouse.default.dir"]] <- tempdir()`

felixcheung: I think we should move this after L383 "overrideEnvs(sparkConfigMap, paramMap)" in case it is passed in named param

R/pkg/inst/tests/testthat/test_sparkSQL.R
```diff
@@ -2165,6 +2165,14 @@ test_that("SQL error message is returned from JVM", {
   expect_equal(grepl("blah", retError), TRUE)
 })
 
+test_that("Default warehouse dir should be set to tempdir", {
+  # nothing should be written outside tempdir() without explicit user permission
+  inital_working_directory_files <- list.files()
```
bdwyer2: Does Jenkins start with a new workspace every time it runs a test?

sql/core/src/main/scala/org/apache/spark/sql/internal/SharedState.scala
```diff
@@ -55,14 +55,19 @@ private[sql] class SharedState(val sparkContext: SparkContext) extends Logging {
         s"is set. Setting ${WAREHOUSE_PATH.key} to the value of " +
         s"hive.metastore.warehouse.dir ('$hiveWarehouseDir').")
       hiveWarehouseDir
-    } else {
+    } else if (sparkContext.conf.contains(WAREHOUSE_PATH.key) &&
+               sparkContext.conf.get(WAREHOUSE_PATH).isDefined) {
```
gatorsmile: Nit: indent is not right.

gatorsmile:
If the default database has already been created in the metastore, any following changes of `spark.sql.default.warehouse.dir` can trigger an issue when we create a data source table in the default database (Here, we assume Hive support is enabled). Note, we will not hit any issue if we create a Hive serde table in the default database, or create a data source table in the non-default database. 

The directory of managed data source tables is created by Hive. When creating a new data source table, the created directory is based on the current value of `hive.metastore.warehouse.dir`. However, the value of table location in the metastore is pointing to the child directory of the location of the default database. Thus, you will not hit any issue when you creating such a table. However, the mismatch will cause a problem (because the expected directory does not exist), when we try to select from /insert into this table. This is a bug of Hive metastore. 

@dilipbiswal hit this issue very recently. Below shows the location of these two tables. 

`t11` is a Hive managed data source table we created in the default database. 
```
spark-sql> describe extended t11;
...
    Storage(Location: file:/user/hive/warehouse/t11, InputFormat: org.apache.hadoop.hive.ql.io.parquet.MapredParquetInputFormat, OutputFormat: org.apache.hadoop.hive.ql.io.parquet.MapredParquetOutputFormat, Serde: org.apache.hadoop.hive.ql.io.parquet.serde.ParquetHiveSerDe, Properties: [serialization.format=1]))    
Time taken: 0.105 seconds, Fetched 8 row(s)
```  

`t1` is a Hive managed data source table we created in the non-default database. 
```
spark-sql> use dilip;
Time taken: 0.028 seconds
spark-sql> describe extended t1;
...
    Storage(Location: file:/home/cloudera/mygit/apache/spark/bin/spark-warehouse/dilip.db/t1, InputFormat: org.apache.hadoop.hive.ql.io.parquet.MapredParquetInputFormat, OutputFormat: org.apache.hadoop.hive.ql.io.parquet.MapredParquetOutputFormat, Serde: org.apache.hadoop.hive.ql.io.parquet.serde.ParquetHiveSerDe, Properties: [serialization.format=1]))    
```

R/pkg/inst/tests/testthat/test_sparkSQL.R
felixcheung: I'm referring to other tests in this test file, test_sparkSQL, that is calling to API that might already initialize the warehouse dir.

`sparkR.session()` is called at the top. Does this `createOrReplaceTempView` cause the warehouse dir to be created?

https://github.com/shivaram/spark-1/blob/25834109588e8e545deafb1da162958766a057e2/R/pkg/inst/tests/testthat/test_sparkSQL.R#L570

felixcheung: From my test, the `spark-warehouse` directory is created when I run `a <- createDataFrame(iris)`

gatorsmile:
retest this please

R/pkg/inst/tests/testthat/test_sparkSQL.R
shivaram: Yeah. I think @felixcheung point is right - The dir should be created early on. Also I think in the tests we sometimes configure the `hive.metastore.dir` in our tests and so I dont see it come up when I run the test cases.  I'll try to see if we can design a different test case.

shivaram:
Ugh - The failure seems to be from `HiveClientSuite` and I dont think its related to this PR (as pasted below). However I'm refactoring the SparkR test case, so let me do that and then re-trigger a test


```
[info] - getPartitionsByFilter returns all partitions when hive.metastore.try.direct.sql=false *** FAILED *** (8 seconds, 479 milliseconds)
[info]   java.lang.RuntimeException: [unresolved dependency: org.apache.hive#hive-metastore;1.2.1: not found, unresolved dependency: org.apache.hive#hive-exec;1.2.1: not found, unresolved dependency: org.apache.hive#hive-common;1.2.1: not found, unresolved dependency: org.apache.hive#hive-serde;1.2.1: not found]
```

R/pkg/R/sparkR.R
shivaram: Moved this below L383 - We still need the `exists` check to make sure we don't overwrite any user provided value ?

R/pkg/inst/tests/testthat/test_sparkSQL.R
shivaram: I refactored this test to recursively list all the files and check if `spark-warehouse` is in it. Another thing is that we could check if the specific table is in it

sql/core/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala
```diff
@@ -964,11 +970,17 @@ object StaticSQLConf {
...
-  val WAREHOUSE_PATH = buildConf("spark.sql.warehouse.dir")
-    .doc("The default location for managed databases and tables.")
+  val DEFAULT_WAREHOUSE_PATH = buildConf("spark.sql.default.warehouse.dir")
+    .doc("The default location for managed databases and tables. " +
+         "Used if spark.sql.warehouse.dir is not set")
     .stringConf
     .createWithDefault(Utils.resolveURI("spark-warehouse").toString)
 
+  val WAREHOUSE_PATH = buildConf("spark.sql.warehouse.dir")
+    .doc("The location for managed databases and tables.")
```
gatorsmile: The description is not right. `spark.sql.warehouse.dir` is still the default location when we create a database/table without providing the location value.

shivaram: Thats a good point. I misunderstood the meaning of `default` there. Will fix this now

sql/core/src/main/scala/org/apache/spark/sql/internal/SharedState.scala
shivaram: Indented 4 spaces now

shivaram:
@gatorsmile Thanks for taking a look. Addressed your comments now. Lets see if Jenkins passes.

shivaram:
@felixcheung @bdwyer2 Could you take another look ?

R/pkg/inst/tests/testthat/test_context.R
```diff
@@ -72,6 +72,20 @@ test_that("repeatedly starting and stopping SparkSession", {
...
 })
 
+test_that("Default warehouse dir should be set to tempdir", {
+  sparkR.session.stop()
+  sparkR.session(enableHiveSupport = FALSE)
+
+  # Create a temporary table
+  sql("CREATE TABLE people_warehouse_test")
+  # spark-warehouse should be written only tempdir() and not current working directory
+  res <- list.files(path = ".", pattern = ".*spark-warehouse.*",
```
bdwyer2: should we test to make sure that no files are created during this process instead of only checking for `spark-warehouse`?

sql/core/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala
```diff
@@ -819,7 +819,13 @@ private[sql] class SQLConf extends Serializable with CatalystConf with Logging {
 
   def variableSubstituteDepth: Int = getConf(VARIABLE_SUBSTITUTE_DEPTH)
 
-  def warehousePath: String = new Path(getConf(StaticSQLConf.WAREHOUSE_PATH)).toString
+  def warehousePath: String = {
+    if (contains(StaticSQLConf.WAREHOUSE_PATH.key)) {
```
gatorsmile: What is the reason we are not doing the same check, as shown in [another place](https://github.com/shivaram/spark-1/blob/6eec97d463b027c93b48621b55e3bd4005dc0f7e/sql/core/src/main/scala/org/apache/spark/sql/internal/SharedState.scala#L59)?

sql/core/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala
```diff
@@ -964,10 +970,16 @@ object StaticSQLConf {
     }
   }
 
+  val DEFAULT_WAREHOUSE_PATH = buildConf("spark.sql.default.warehouse.dir")
```
gatorsmile: Should we make it internal?

sql/core/src/test/scala/org/apache/spark/sql/internal/SQLConfSuite.scala
```diff
@@ -221,6 +221,19 @@ class SQLConfSuite extends QueryTest with SharedSQLContext {
...
   }
 
+  test("changing default value of warehouse path") {
+    try {
+      val newWarehouseDefault = "spark-warehouse2"
+      val newWarehouseDefaultPath = new Path(Utils.resolveURI(newWarehouseDefault)).toString
+      sparkContext.conf.set("spark.sql.default.warehouse.dir", newWarehouseDefaultPath)
+      val spark = new SparkSession(sparkContext)
+      assert(newWarehouseDefaultPath.stripSuffix("/") === spark
+        .sessionState.conf.warehousePath.stripSuffix("/"))
```
gatorsmile: Also need a check for `spark.sharedState.warehousePath` because we did the logic changes there.

gatorsmile:
I finished my review. cc @cloud-fan

R/pkg/inst/tests/testthat/test_context.R
shivaram: Well - given that this PR is only changing the warehouse dir, I think its only fair to test for that.  Or in other words, adding such a test would fail now because of `derby.log` etc. (per our JIRA discussion) ?

sql/core/src/test/scala/org/apache/spark/sql/internal/SQLConfSuite.scala
```diff
@@ -221,6 +221,19 @@ class SQLConfSuite extends QueryTest with SharedSQLContext {
       .sessionState.conf.warehousePath.stripSuffix("/"))
   }
 
+  test("changing default value of warehouse path") {
```
shivaram: Good point. Added tests for all 4 cases now

sql/core/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala
shivaram: Nice catch - Added the same check here as well

sql/core/src/test/scala/org/apache/spark/sql/internal/SQLConfSuite.scala
shivaram: Done

shivaram:
Thanks @gatorsmile - Addressed your comments.

gatorsmile:
retest this please

R/pkg/inst/tests/testthat/test_context.R
shivaram: Ah I see - Will check this today.  I think if `SPARK_HOME` is accessible I can just call `list.files` with that as `path`

shivaram:
@gatorsmile The test error in HiveSparkSubmitSuite seems related. I am debugging locally

R/pkg/inst/tests/testthat/test_context.R
felixcheung: I think a couple of other test files would have called `sparkR.session` already, so I'd propose adding a new test explicitly named to make sure it is called first, ie. https://github.com/apache/spark/pull/16330/files#diff-5ff1ba5d1751f3b1cc96a567e9ab25ffR18

shivaram: I'm not sure why it needs to run first ? because the default warehouse dir is in `tempdir` even if spark.session is called before it shouldn't create any warehouse dir in `SPARK_HOME/` ?

shivaram:
@gatorsmile I think I figured out the problem with `HiveSparkSubmitSuite` but I'm not sure how to solve it. The problem is that in one of the test cases we check the DB location to be same as warehouse path [1]. 

Now the warehouse path correctly points to `spark-warehouse` - The catalog though seems to be configured to an existing `metastore_db`. Now in my machine (and I guess in Jenkins) if we have a metastore_db left over from running SparkR tests then the default location points to the R `tempdir` which is no longer valid.

It seems to me that the right solution here is to also clear the `metastore_db` once a set of tests finish ? Is there a way to do this from within the tests or should we be running `rm` from some of the test running scripts ?

[1] https://github.com/apache/spark/blob/1e5c51f336b90cd1eed43e9c6cf00faee696174c/sql/hive/src/test/scala/org/apache/spark/sql/hive/HiveSparkSubmitSuite.scala#L824

R/pkg/inst/tests/testthat/test_context.R
felixcheung: You are right.
In this case we are specifically looking for spark-warehouse; I guess I was referring to a general check to make sure going forward the list of files before running anything in the package should  == list of files after running anything

felixcheung:
@shivaram with my PR #16330, `metastore_db` is in `tempdir` and is removed when the R process exits.

R/pkg/inst/tests/testthat/test_context.R
shivaram: I think my bigger concern for that is that usually tests are run all at time - i.e. core, sql, hive and then python, R. And there are no guarantees that other module tests won't create files inside `SPARK_HOME` afaik. So while we can check some basic things with our test, I dont think verifying a global property is always possible.

gatorsmile:
retest this please

gatorsmile:
I checked the most two recent failed test cases in Jenkins. They are not related to the changes in the PR. 

In the local environment, I can reproduce the error you mentioned above.  
```
[info]   2016-12-18 20:47:23.328 - stdout> path1: file:/Users/xiaoli/IdeaProjects/sparkDelivery/bin/spark-warehouse
[info]   2016-12-18 20:47:23.328 - stdout> path2: file:/Users/xiaoli/IdeaProjects/sparkDelivery/spark-warehouse/
```

The location of default database is still pointing to the original value of `hive.metastore.warehouse.dir` or `spark.sql.warehouse.dir` that was set in the previous test case or our previous local spark job. Ideally, our test suite should directly connect to Derby and drop the default database. Let me do more search. 

Also cc @yhuai

R/pkg/inst/tests/testthat/test_context.R
felixcheung: That I agree completely

gatorsmile:
After a research, for avoiding this flaky testcase, the simplest way is to remove the contents in `metastore_db` (whose location is specified through `javax.jdo.option.ConnectionURL`).

gatorsmile:
Generally, it looks good to me. The only concern is the external behavior change. Since Spark 2.0, users are able to get the warehouse location by the following two ways:
```Scala
      spark.sql("set spark.sql.warehouse.dir").show()
```
```Scala
      println(spark.conf.get("spark.sql.warehouse.dir"))
```

After this PR, the output will be `<undefined>`. Do we need extra code changes to fix this? @cloud-fan

shivaram:
@cloud-fan Any thoughts on this ?

R/pkg/R/sparkR.R
```diff
@@ -376,6 +377,12 @@ sparkR.session <- function(
     overrideEnvs(sparkConfigMap, paramMap)
   }
 
+  # NOTE(shivaram): Set default warehouse dir to tmpdir to meet CRAN requirements
+  # See SPARK-18817 for more details
+  if (!exists("spark.sql.default.warehouse.dir", envir = sparkConfigMap)) {
```
gatorsmile: After rethinking it, we might not need to add an extra sql conf. We just need to know whether the value of `spark.sql.warehouse.dir` is from the users or the original default. If it is the default, R can simply change it. 

Maybe it is a good to-have feature for users to know whether the SQLConf value is from users or from the default. cc @cloud-fan

cloud-fan: actually we can, `SessionState.conf.settings` contains all the user-setted entries.

shivaram: Ah I see - I will make try to use `SessionState` and see if it can avoid having to create a new option

shivaram:
@gatorsmile @cloud-fan @felixcheung I looked at the SharedState code more closely and it looks like the only time the warehousePath can be set is when the initialization of shared state happens. So I modified the code to set the temp dir for SparkR during SparkSession initialization if it has not already been set. 

I verified that this works locally and with a unit test -- Is there anything I might be missing with this approach ?

felixcheung:
so base on this comment https://github.com/apache/spark/pull/16330#issuecomment-282101389
doesn't it mean we shouldn't set warehouse dir to under tempdir()?

shivaram:
@felixcheung Yeah I think that sounds good. I can close this PR for now and we can revisit this if we have an issue in the future. Also I guess you will update #16330 for the derby.log change. Does this sound okay ?

felixcheung:
Yes that's the plan.

PR #16291, opened 2016-12-15
[SPARK-18838][CORE] Use separate executor service for each event listener

sitalkedia:

Currently we are observing the issue of very high event processing delay in driver's `ListenerBus` for large jobs with many tasks. Many critical component of the scheduler like `ExecutorAllocationManager`, `HeartbeatReceiver` depend on the `ListenerBus` events and this delay might hurt the job performance significantly or even fail the job. For example, a significant delay in receiving the `SparkListenerTaskStart` might cause `ExecutorAllocationManager` manager to mistakenly remove an executor which is not idle.
The problem is that the event processor in `ListenerBus` is a single thread which loops through all the Listeners for each event and processes each event synchronously https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/scheduler/LiveListenerBus.scala#L94. This single threaded processor often becomes the bottleneck for large jobs. Also, if one of the Listener is very slow, all the listeners will pay the price of delay incurred by the slow listener. In addition to that a slow listener can cause events to be dropped from the event queue which might be fatal to the job.
To solve the above problems, we propose to get rid of the event queue and the single threaded event processor. Instead each listener will have its own dedicate single threaded executor service . When ever an event is posted, it will be submitted to executor service of all the listeners. The Single threaded executor service will guarantee in order processing of the events per listener. The queue used for the executor service will be bounded to guarantee we do not grow the memory indefinitely. The downside of this approach is separate event queue per listener will increase the driver memory footprint.

## How was this patch tested?

Tested by running the job on the cluster and the average event processing latency dropped from 2 minutes to 3 milliseconds.


sitalkedia:
cc -  @zsxwing  - Please note that the PR is incomplete and there are some test failures. I just wanted some initial feedback on the design before investing more time on it.

rxin:
This was initially introduced by @kayousterhout

vanzin:
Hmm... I took a quick look and I'm not sure I understand exactly what's going on. It seems you're wrapping each listener with a `ListenerEvenProcessor` (note the typo), and each processor has its own thread pool for processing events. 

If that's the case, that sounds wrong. Each listener should process events serially otherwise you risk getting into funny situations like a task end event being processed before the task start event for the same task.

I think this would benefit from a proper explanation of the changes being proposed, instead of a bunch of code. How will the listener bus work with the changes? Where will we have thread pools, where will we have message queues? Will each listener get its own dedicated thread, or will there be a limit? What kind of controls will there be to avoid memory pressure? Would it be worth it to allow certain listeners to have dedicated threads while others share the same one like the current approach? That kind of thing. Can you write something up and post it to the bug?

sitalkedia:
@vanzin - Thanks for taking a look and sorry about putting my unfinished sloppy code out there. I updated the bug and the PR with the overall design idea, which hopefully will answer your questions. 

>> Each listener should process events serially otherwise you risk getting into funny situations like a task end event being processed before the task start event for the same task.

You are right, that's why the executor service is single threaded which guarantees ordered processing of the events per listener.

sitalkedia:
Jenkins test this please.

sitalkedia:
Jenkins retest this please.

sitalkedia:
This change is ready for review @rxin, @zsxwing, @kayousterhout

markhamstra:
> You are right, that's why the executor service is single threaded which guarantees ordered processing of the events per listener.

But this is still substantially different from the ordering guarantees in the current implementation, and I'm not sure that this PR wouldn't invalidate some assumptions made by current listeners.   Events will end up in the same order in each `ListenerEventExecutor` (at least by my understanding, which someone else should double-check), but there is no synchronization between `ListenerEventExecutor`s of the processing of their `eventQueue`s; so it is entirely possible. for example, for one `ListenerEventExecutor` to process a task end event for a particular task before another `ListenerEventExecutor` has worked sufficiently through its `eventQueue` to have even seen the corresponding task start event.  That is quite a bit different than the prior ordering guarantee implied by the comment "`onPostEvent` is guaranteed to be called in the same thread for all listeners."

Whether this actually constitutes a problem or not, I'm not certain; but serializing event processing within each `ListenerEventExecutor` isn't enough if `Listener`s on different `ListenerEventExecutor` threads expect to share mutable state.

vanzin:
I'm not sure whether that change would be a problem, but I wonder if a system where listeners can "opt in" to this behavior wouldn't be better. e.g., `EventLoggingListener` can slow down the bus, having it run in a separate thread would benefit all the others. Similarly, the dynamic allocation listener could have its own dedicated thread since it needs lower latency, and doesn't really depend on any others.

sitalkedia:
@markhamstra, @vanzin thanks for your comments!

>> so it is entirely possible. for example, for one ListenerEventExecutor to process a task end event for a particular task before another ListenerEventExecutor has worked sufficiently through its eventQueue to have even seen the corresponding task start event. That is quite a bit different than the prior ordering guarantee implied by the comment "onPostEvent is guaranteed to be called in the same thread for all listeners."

That is true. If a listener is slow in processing the events than it might not see the task start event where as other listeners might already have processed the corresponding task end event. I am not sure why would that cause any problem though. Each listener should in theory could work independent of each other and we should only guarantee ordered processing of the events within a listener. This is the only way I can think of to scale the driver for large workloads.


>>  but I wonder if a system where listeners can "opt in" to this behavior wouldn't be better. e.g., EventLoggingListener can slow down the bus, having it run in a separate thread would benefit all the others.

I think that will be an over-kill. Having a separate thread per listener simplifies the design. Also, for user added listeners, it will be hard to decide if we want to have a shared or separate thread per listener. The only downside of having a separate thread per listener is increased memory footprint in worst case. But I would argue that since the average event processing latency is decreased by multi threaded event processor, the backlog on the event queues will clear faster which actually might decrease the driver memory footprint on average case.

markhamstra:
> Each listener should in theory could work independent of each other and we should only guarantee ordered processing of the events within a listener.

If we were starting from nothing, then yes, it would be valid and advisable to design the Listener infrastructure using only this weaker guarantee.  The issue, though, is that we are not starting from nothing, but rather from a system that currently offers a much stronger guarantee on the synchronized behavior of Listeners.  If it is the case that no Listeners currently rely on the stronger guarantee and thus could work completely correctly under the weaker guarantee of this PR, then we could make this change without much additional concern.  But reaching that level of confidence in current Listeners is a difficult prerequisite -- strictly speaking, it's an impossible task.

We could carefully work through all the internal behavior of Spark's Listeners to convince ourselves that they can work correctly under the new, weaker guarantee.  At a bare minimum, we need to do that much before we can consider merging this PR -- but that's probably not enough.  The problem is that Listeners aren't just internal to Spark.  Users have also developed their own custom Listeners that either implement `SparkListenerInterface` or extend `SparkListener` or `SparkFirehoseListener`, and we can't just assume that those custom Listeners don't rely upon the current guarantee to either synchronize behavior with other custom Listeners or even with Spark internal Listeners.  Since we can't know that user Listeners don't already rely upon the current, stronger guarantee, the question now becomes whether we even have the freedom to change that guarantee within the lifetime of Spark 2.x, or whether any such change would have to wait for Spark 3.x.

`SparkListener` is still annotated as `@DeveloperAPI`, so if that were the only piece in play, then we could change its guarantee fairly freely.  `SparkListenerInterface` is almost as good, since it includes the admonition in a comment to "[n]ote that this is an internal interface which might change in different Spark releases."  The stickier issue is with `SparkFirehoseListener`, which carries no such annotations or comments, but is just a plain public class and API.  So, after convincing ourselves that Spark's internal Listeners would be fine with this PR, we'd still have to convince the Spark PMC that changing the public `SparkFirehoseListener` (with prominent warnings in the release notes, of course) before Spark 3.x would be acceptable.

And all of the above is still really only arguing about whether we *could* adopt this PR in essentially its present form.  There are still questions of whether we *should* do this or maybe instead we should do something a little different or more.  I can see some merit in Marcelo's "opt in" suggestion.  If there is utility in having groups of Listeners that can rely upon synchronized behavior, then we should probably retain one or more threads running synchronized Listeners.  For example, if Listener A relies upon synchronization with Listeners B and C while D needs to synchronize with E, but F, G and H are all independent, then there are a couple of things we could do.  First, the independent Listeners (F, G and H) can each run in its own thread, providing the scalable performance that this PR is aiming for.  After that, we could either have one synchronized Listener thread for all the other Listeners, or we could have one thread for A, B and C and one thread for D and E.  Whether we support only one synchronized Listener group/thread or multiple, we'd still need some mechanism for Listeners to select into a synchronized group or to indicate that they can and should be run independently on their own thread.

@rxin

vanzin:
> I think that will be an over-kill. Having a separate thread per listener simplifies the design. 

I don't think it's overkill, and it wouldn't really be that much of a change from your existing design. It also serves as a way of controlling how much extra memory you'll need for these things, since right now each listener would incur in its own thread + message queue.

e.g. you could just have a tagging interface (if listener implements "X" then it gets its own dedicated dispatch thread, or if you want to get fancy, you could also define a "group" so that listener in the same group share a dispatch thread).

mridulm:
I agree with @markhamstra and @vanzin - having ability to tag listeners  into groups (default = spark listener group) and preserving current synchronized behavior within group would be ensure backward compatibility at fairly minimal additional complexity.

sitalkedia:
jenkins test this please.

sitalkedia:
Jenkins retest this please.

sitalkedia:
Thanks guys for your input. I have made changes to support grouping of listeners to preserve the existing synchronization behavior. Currently, the `ExecutionAllocationManager`, `HeartbeatReceiver` and `EventLogger` use their separate threads.  All User developed `Listerners` use a separate thread and all other belong to a default group.  Let me know what you think the approach.

@rxin, @vanzin, @markhamstra , @mridulm .

core/src/main/scala/org/apache/spark/util/ListenerBus.scala
```diff
@@ -17,53 +17,222 @@
...
 import scala.collection.JavaConverters._
 import scala.reflect.ClassTag
 import scala.util.control.NonFatal
 
+import com.google.common.util.concurrent.ThreadFactoryBuilder
+
 import org.apache.spark.internal.Logging
+import org.apache.spark.scheduler.LiveListenerBus
+
+private class ListenerEventExecutor[L <: AnyRef] (listenerName: String, queueCapacity: Int)
```
vanzin: Not a fan of the name, since you don't execute events. I prefer `ListenerEventDispatcher.`

Also, `listenerName` is more of a group name now, right? You mention "listener" in many places of this class, those should be replaced with "listener group".

Finally, not sure why do you need the type parameter?

nit: no space before `(`

core/src/main/scala/org/apache/spark/util/ListenerBus.scala
```diff
@@ -17,53 +17,222 @@
...
+  extends Logging {
+  private val threadFactory = new ThreadFactoryBuilder().setDaemon(true)
```
vanzin: `ThreadUtils.namedThreadFactory`.

core/src/main/scala/org/apache/spark/util/ListenerBus.scala
```diff
@@ -17,53 +17,222 @@
...
+  private var lastReportTimestamp = 0L
+  /** Indicates if we are in the middle of processing some event */
+  private val processingEvent = new AtomicBoolean(false)
+  /**
+   * Indicates if the event executor is started. The executor thread will be
+   * blocked on the condition variable until the event executor is started to
+   * guarantee that we do not process any event before starting the event executor.
+   */
+  private val isStarted = new AtomicBoolean(false)
+  private val lock = new ReentrantLock()
```
vanzin: Minor, but you could achieve the same thing with synchronized/wait/notifyAll, with less code.

core/src/main/scala/org/apache/spark/util/ListenerBus.scala
```diff
@@ -17,53 +17,222 @@
...
+
+  private[this] def isProcessingEvent: Boolean = processingEvent.get()
+
+  def isEmpty: Boolean = {
+    executorService.getQueue.size() == 0 && !isProcessingEvent
+  }
+
+  def stop(): Unit = {
+    executorService.shutdown()
+    executorService.awaitTermination(10, TimeUnit.SECONDS)
```
vanzin: This changes the current behavior slightly. Right now the bus will wait until the listener thread stops:

    listenerThread.join()

Here you might return before things are really shut down. It's unlikely to happen, but if it does happen, it can lead to weird issues. Might be better to try `shutdownNow()` if this call fails, which at least makes a best effort to interrupt existing tasks that might be slow.

core/src/main/scala/org/apache/spark/util/ListenerBus.scala
```diff
@@ -17,53 +17,222 @@
...
+          waitForStart()
+          processingEvent.set(true)
+          task.run()
+          processingEvent.set(false)
+      }
+      })
+    } catch {
+      case e: RejectedExecutionException =>
+        droppedEventsCounter.incrementAndGet()
+        if (System.currentTimeMillis() - lastReportTimestamp >= 60 * 1000) {
```
vanzin: When measuring elapsed time it's generally better to use `System.nanoTime` since it's monotonic.

core/src/main/scala/org/apache/spark/util/ListenerBus.scala
```diff
@@ -17,53 +17,222 @@
...
+          val droppedEvents = droppedEventsCounter.get
+          // There may be multiple threads trying to decrease droppedEventsCounter.
+          // Use "compareAndSet" to make sure only one thread can win.
+          // And if another thread is increasing droppedEventsCounter, "compareAndSet" will fail and
+          // then that thread will update it.
+          if (droppedEventsCounter.compareAndSet(droppedEvents, 0)) {
+            droppedEventsCounter.set(0)
```
vanzin: Redundant.

core/src/main/scala/org/apache/spark/util/ListenerBus.scala
```diff
@@ -17,53 +17,222 @@
...
+            val prevLastReportTimestamp = lastReportTimestamp
+            lastReportTimestamp = System.currentTimeMillis()
+            logError(s"Dropping $droppedEvents SparkListenerEvent since " +
+              new java.util.Date(prevLastReportTimestamp) +
+              " because no remaining room in event queue. This likely means" +
+              s" $listenerName event processor is too slow and cannot keep up " +
+              "with the rate at which tasks are being started by the scheduler.")
```
vanzin: "at which events are being posted". Task end events can also flood the bus (and others, but at a lesser scale).

core/src/main/scala/org/apache/spark/util/ListenerBus.scala
```diff
@@ -17,53 +17,222 @@
...
+          }
+        }
+    }
+  }
+
+  private[this] def isProcessingEvent: Boolean = processingEvent.get()
+
+  def isEmpty: Boolean = {
+    executorService.getQueue.size() == 0 && !isProcessingEvent
```
vanzin: This feels racy. I can see a window after the task is removed from the queue and before the task starts to run where this would return `true`.

core/src/main/scala/org/apache/spark/util/ListenerBus.scala
```diff
@@ -17,53 +17,222 @@
...
 
   /**
    * Add a listener to listen events. This method is thread-safe and can be called in any thread.
    */
-  final def addListener(listener: L): Unit = {
-    listeners.add(listener)
+  final def addListener(
+    listener: L, eventListenerGroup: String = ListenerEventExecutor.DefaultEventListenerGroup):
+  Unit = synchronized {
+    var listenerEventExecutor = eventGroupToEventExecutors.get(eventListenerGroup)
```
vanzin: I'm not sure this is what we want. This is forcing asynchronous behavior on all buses, when we really only want this for `LiveListenerBus`. In particular, we do *not* want this to be asynchronous in `ReplayListenerBus`. You could change the history server to use `postToAllSync`, but that feels like a hack. It would be better to have each bus handle their listeners the way they see fit.

I'll comment on the implementation below, but I really think all this logic should be in `LiveListenerBus` only.

core/src/main/scala/org/apache/spark/util/ListenerBus.scala
```diff
@@ -76,7 +245,55 @@ private[spark] trait ListenerBus[L <: AnyRef, E] extends Logging {
...
 
+  /**
+   * Start sending events to attached listeners.
+   *
+   * This first sends out all buffered events posted before this listener bus has started, then
+   * listens for any additional events asynchronously while the listener bus is still running.
+   * This should only be called once.
+   *
+   */
+  def start(): Unit = {
```
vanzin: I'd make this synchronized to avoid possible races with `addListener`.

core/src/main/scala/org/apache/spark/util/ListenerBus.scala
```diff
@@ -17,53 +17,222 @@
...
-  final def addListener(listener: L): Unit = {
-    listeners.add(listener)
+  final def addListener(
+    listener: L, eventListenerGroup: String = ListenerEventExecutor.DefaultEventListenerGroup):
+  Unit = synchronized {
+    var listenerEventExecutor = eventGroupToEventExecutors.get(eventListenerGroup)
+    if (listenerEventExecutor == null) {
+      listenerEventExecutor =
+        new ListenerEventExecutor[L](listener.getClass.getName, eventQueueSize)
+      eventGroupToEventExecutors.put(eventListenerGroup, listenerEventExecutor)
```
vanzin: I think that if you make `start` also synchronized, and move this after the `started.get()` check below, you can avoid all the `waitForStart()` stuff in the dispatcher class. I haven't fully checked if that will work, but if it does, it will greatly simplify the code in the dispatcher class (no atomic booleans, locks, signals, etc).

core/src/main/scala/org/apache/spark/util/ListenerBus.scala
```diff
@@ -17,53 +17,222 @@
...
-    val iter = listeners.iterator
-    while (iter.hasNext) {
-      val listener = iter.next()
-      try {
-        doPostEvent(listener, event)
-      } catch {
-        case NonFatal(e) =>
-          logError(s"Listener ${Utils.getFormattedClassName(listener)} threw an exception", e)
+    for (listenerEventProcessor <- eventGroupToEventExecutors.values().asScala) {
+      // JavaConverters can create a JIterableWrapper if we use asScala.
```
vanzin: Shouldn't you then also avoid the `asScala` convertion above and use Java APIs all around?

core/src/main/scala/org/apache/spark/util/ListenerBus.scala
```diff
@@ -17,53 +17,222 @@
...
+      eventExecutor.removeListener(listener)
+    }
+  }
+
+   /**
+    * For testing only. Returns whether there is any event pending to be processed by
+    * any of the existing listener
+    */
+  def isListenerBusEmpty: Boolean = {
+    for (eventExecutor <- eventGroupToEventExecutors.values().asScala) {
```
vanzin: I think you can do this with `eventGroupToEventExecutors.values().asScala.forall`.

core/src/main/scala/org/apache/spark/util/ListenerBus.scala
```diff
@@ -17,53 +17,222 @@
...
-      try {
-        doPostEvent(listener, event)
-      } catch {
-        case NonFatal(e) =>
-          logError(s"Listener ${Utils.getFormattedClassName(listener)} threw an exception", e)
+    for (listenerEventProcessor <- eventGroupToEventExecutors.values().asScala) {
+      // JavaConverters can create a JIterableWrapper if we use asScala.
+      // However, this method will be called frequently. To avoid the wrapper cost, here we use
+      // Java Iterator directly.
+      val iter = listenerEventProcessor.listeners.iterator()
```
vanzin: It feels like this logic should be in the `submit()` implementation. Right now you're peeking into its internals, and actually wrapping a runnable inside another runnable, which feels hacky.

Instead, if you just `submit()` the event itself, the task you submit to the executor can iterate over the listeners in the group, and you don't need to expose internal state of the dispatcher to this code.

core/src/main/scala/org/apache/spark/util/ListenerBus.scala
```diff
@@ -17,53 +17,222 @@
...
+      }
+    }
+  }
+
+  /**
+   * For testing only. Post the event to all registered listeners.
+   * This guarantees processing the event in the same thread for all
+   * events.
+   */
+  final def postToAllSync(event: E): Unit = {
```
vanzin: Similarly to the above, this feels hacky. If you follow my previous suggestion, you could return a `Future` from the `submit()` method, and this method would just become a call like `submit(...).get()`.

core/src/main/scala/org/apache/spark/util/ListenerBus.scala
```diff
@@ -76,7 +245,55 @@ private[spark] trait ListenerBus[L <: AnyRef, E] extends Logging {
...
+    for (eventExecutor <- eventGroupToEventExecutors.values().asScala) {
+      eventExecutor.start()
+    }
+  }
+
+  /**
+   * Stop the listener bus. It will wait until the queued events have been processed, but drop the
+   * new events after stopping.
+   */
+  def stop(): Unit = {
```
vanzin: To follow the other suggestions, make this synchronized.

core/src/main/scala/org/apache/spark/util/ListenerBus.scala
```diff
@@ -76,7 +245,55 @@ private[spark] trait ListenerBus[L <: AnyRef, E] extends Logging {
...
+    if (!started.get()) {
+      throw new IllegalStateException(s"Attempted to stop hat has not yet started!")
+    }
+    if (stopped.compareAndSet(false, true)) {
```
vanzin: This block looks weird, seems like you forgot to actually write some code.

core/src/main/scala/org/apache/spark/util/ListenerBus.scala
```diff
@@ -76,7 +245,55 @@ private[spark] trait ListenerBus[L <: AnyRef, E] extends Logging {
...
+    } else {
+      // Keep quiet
+    }
+    val iter = eventGroupToEventExecutors.values().iterator()
```
vanzin: You could use `.asScala.foreach` here.

core/src/main/scala/org/apache/spark/util/ListenerBus.scala
```diff
@@ -76,7 +245,55 @@ private[spark] trait ListenerBus[L <: AnyRef, E] extends Logging {
...
+  }
+}
+
+private[spark] object ListenerEventExecutor {
+  val DefaultEventListenerGroup = "default-event-listener"
+  val DefaultUserEventListenerGroup = "default-user-event-listener"
+  val ExecutorAllocationManagerGroup = "executor-allocation-manager-listener"
+  val HeartBeatReceiverGroup = "heart-beat-receiver-listener"
+  val EventLoggingGroup = "event-logging-listener"
+  // Allows for Context to check whether stop() call is made within listener thread
```
vanzin: ?

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamingQueryListenerBus.scala
```diff
@@ -35,7 +35,7 @@ import org.apache.spark.util.ListenerBus
  * and StreamingQueryManager. So this bus will dispatch events to registered listeners for only
  * those queries that were started in the associated SparkSession.
  */
-class StreamingQueryListenerBus(sparkListenerBus: LiveListenerBus)
+class StreamingQueryListenerBus(val sparkListenerBus: LiveListenerBus)
```
vanzin: Do you need to make this a `val`?

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamingQueryListenerBus.scala
```diff
@@ -69,7 +69,7 @@ class StreamingQueryListenerBus(sparkListenerBus: LiveListenerBus)
         activeQueryRunIds.synchronized { activeQueryRunIds += s.runId }
         sparkListenerBus.post(s)
         // post to local listeners to trigger callbacks
-        postToAll(s)
+        postToAllSync(s)
```
vanzin: Do you need this change? Seems like it wasn't a synchronous call before. If you need it, then `postToAllSync` isn't really for testing only, and you need to change the command in that method.

Same below.

streaming/src/main/scala/org/apache/spark/streaming/StreamingContext.scala
```diff
@@ -661,7 +661,7 @@ class StreamingContext private[streaming] (
     var shutdownHookRefToRemove: AnyRef = null
     if (LiveListenerBus.withinListenerThread.value) {
       throw new SparkException(
-        s"Cannot stop StreamingContext within listener thread of ${LiveListenerBus.name}")
+        "Cannot stop SparkContext within listener event executor thread")
```
vanzin: `StreamingContext` was actually correct.

streaming/src/main/scala/org/apache/spark/streaming/scheduler/StreamingListenerBus.scala
```diff
@@ -26,7 +26,7 @@ import org.apache.spark.util.ListenerBus
  * registers itself with Spark listener bus, so that it can receive WrappedStreamingListenerEvents,
  * unwrap them as StreamingListenerEvent and dispatch them to StreamingListeners.
  */
-private[streaming] class StreamingListenerBus(sparkListenerBus: LiveListenerBus)
+private[streaming] class StreamingListenerBus(val sparkListenerBus: LiveListenerBus)
```
vanzin: Similarly, is `val` necessary?

core/src/main/scala/org/apache/spark/util/ListenerBus.scala
```diff
@@ -76,7 +245,55 @@ private[spark] trait ListenerBus[L <: AnyRef, E] extends Logging {
...
+    }
+    val iter = eventGroupToEventExecutors.values().iterator()
+    while (iter.hasNext) {
+      iter.next().stop()
+    }
+  }
+}
+
+private[spark] object ListenerEventExecutor {
+  val DefaultEventListenerGroup = "default-event-listener"
```
vanzin: Spark's style is to use `ALL_CAPS` for constants.

vanzin (commented):
I did a first pass to check the logic, and as I explained in the comments, my main concern is that I think this change should only apply to `LiveListenerBus`. This asynchronous behavior is actually not desired in `ReplayListenerBus`.

tgravescs:
@sitalkedia are you still working on this?

sitalkedia:
@tgravescs - I have not gotten time to work on this. Hopefully will get to it sometime next week.

CodingCat:
what's the current status of this PR?

HyukjinKwon:
what's the current status of this PR?

PR #16296, opened 2016-12-15
[SPARK-18885][SQL] unify CREATE TABLE syntax for data source and hive serde tables

cloud-fan:

Today we have different syntax to create data source or hive serde tables, we should unify them to not confuse users and step forward to make hive a data source.

Please read https://issues.apache.org/jira/secure/attachment/12843835/CREATE-TABLE.pdf for  details.

TODO(for follow-up PRs):
1. TBLPROPERTIES is not added to the new syntax, we should decide if we wanna add it later.
2. `SHOW CREATE TABLE` should be updated to use the new syntax.
3. we should decide if we wanna change the behavior of `SET LOCATION`.

## How was this patch tested?

new tests

### DIFF
--- a/sql/catalyst/src/main/antlr4/org/apache/spark/sql/catalyst/parser/SqlBase.g4
+++ b/sql/catalyst/src/main/antlr4/org/apache/spark/sql/catalyst/parser/SqlBase.g4
@@ -69,16 +69,18 @@ statement
     | ALTER DATABASE identifier SET DBPROPERTIES tablePropertyList     #setDatabaseProperties
     | DROP DATABASE (IF EXISTS)? identifier (RESTRICT | CASCADE)?      #dropDatabase
     | createTableHeader ('(' colTypeList ')')? tableProvider
-        (OPTIONS tablePropertyList)?
+        (OPTIONS options=tablePropertyList)?
         (PARTITIONED BY partitionColumnNames=identifierList)?
-        bucketSpec? (AS? query)?                                       #createTableUsing
+        bucketSpec? locationSpec?
+        (COMMENT comment=STRING)?
+        (AS? query)?                                                   #createTable
     | createTableHeader ('(' columns=colTypeList ')')?
-        (COMMENT STRING)?
+        (COMMENT comment=STRING)?
         (PARTITIONED BY '(' partitionColumns=colTypeList ')')?
         bucketSpec? skewSpec?
         rowFormat?  createFileFormat? locationSpec?
         (TBLPROPERTIES tablePropertyList)?
-        (AS? query)?                                                   #createTable
+        (AS? query)?                                                   #createHiveTable
     | CREATE TABLE (IF NOT EXISTS)? target=tableIdentifier
         LIKE source=tableIdentifier                                    #createTableLike
     | ANALYZE TABLE tableIdentifier partitionSpec? COMPUTE STATISTICS
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/SparkSqlParser.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/SparkSqlParser.scala
@@ -377,7 +377,9 @@ class SparkSqlAstBuilder(conf: SQLConf) extends AstBuilder {
     val storage = DataSource.buildStorageFormatFromOptions(options)
 
     if (location.isDefined && storage.locationUri.isDefined) {
-      throw new ParseException("Cannot specify LOCATION when there is 'path' in OPTIONS.", ctx)
+      throw new ParseException(
+        "LOCATION and 'path' in OPTIONS are both used to indicate the custom table path, " +
+          "you can only specify one of them.", ctx)
     }
     val customLocation = storage.locationUri.orElse(location)
 
@@ -1060,19 +1062,7 @@ class SparkSqlAstBuilder(conf: SQLConf) extends AstBuilder {
     val schema = StructType(dataCols ++ partitionCols)
 
     // Storage format
-    val defaultStorage: CatalogStorageFormat = {
-      val defaultStorageType = conf.getConfString("hive.default.fileformat", "textfile")
-      val defaultHiveSerde = HiveSerDe.sourceToSerDe(defaultStorageType)
-      CatalogStorageFormat(
-        locationUri = None,
-        inputFormat = defaultHiveSerde.flatMap(_.inputFormat)
-          .orElse(Some("org.apache.hadoop.mapred.TextInputFormat")),
-        outputFormat = defaultHiveSerde.flatMap(_.outputFormat)
-          .orElse(Some("org.apache.hadoop.hive.ql.io.HiveIgnoreKeyTextOutputFormat")),
-        serde = defaultHiveSerde.flatMap(_.serde),
-        compressed = false,
-        properties = Map())
-    }
+    val defaultStorage = HiveSerDe.getDefaultStorage(conf)
     validateRowFormatFileFormat(ctx.rowFormat, ctx.createFileFormat, ctx)
     val fileStorage = Option(ctx.createFileFormat).map(visitCreateFileFormat)
       .getOrElse(CatalogStorageFormat.empty)
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/SparkSqlParser.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/SparkSqlParser.scala
@@ -354,7 +354,7 @@ class SparkSqlAstBuilder(conf: SQLConf) extends AstBuilder {
    *    [SORTED BY (col_name [ASC|DESC], ...)]
    *    INTO num_buckets BUCKETS
    *   ]
-   *   [TBLPROPERTIES (property_name=property_value, ...)]
+   *   [LOCATION path]
    *   [COMMENT table_comment]
    *   [AS select_statement];
    * }}}
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/SparkSqlParser.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/SparkSqlParser.scala
@@ -342,11 +342,11 @@ class SparkSqlAstBuilder(conf: SQLConf) extends AstBuilder {
   }
 
   /**
-   * Create a data source table, returning a [[CreateTable]] logical plan.
+   * Create a table, returning a [[CreateTable]] logical plan.
    *
    * Expected format:
    * {{{
-   *   CREATE [EXTERNAL] TABLE [IF NOT EXISTS] [db_name.]table_name
+   *   CREATE [TEMPORARY] TABLE [IF NOT EXISTS] [db_name.]table_name
    *   USING table_provider
    *   [OPTIONS table_property_list]
    *   [PARTITIONED BY (col_name, col_name, ...)]
@@ -354,19 +354,18 @@ class SparkSqlAstBuilder(conf: SQLConf) extends AstBuilder {
    *    [SORTED BY (col_name [ASC|DESC], ...)]
    *    INTO num_buckets BUCKETS
    *   ]
+   *   [TBLPROPERTIES (property_name=property_value, ...)]
+   *   [COMMENT table_comment]
    *   [AS select_statement];
    * }}}
    */
-  override def visitCreateTableUsing(ctx: CreateTableUsingContext): LogicalPlan = withOrigin(ctx) {
+  override def visitCreateTable(ctx: CreateTableContext): LogicalPlan = withOrigin(ctx) {
     val (table, temp, ifNotExists, external) = visitCreateTableHeader(ctx.createTableHeader)
     if (external) {
       operationNotAllowed("CREATE EXTERNAL TABLE ... USING", ctx)
     }
-    val options = Option(ctx.tablePropertyList).map(visitPropertyKeyValues).getOrElse(Map.empty)
+    val options = Option(ctx.options).map(visitPropertyKeyValues).getOrElse(Map.empty)
     val provider = ctx.tableProvider.qualifiedName.getText
-    if (provider.toLowerCase == DDLUtils.HIVE_PROVIDER) {
-      throw new AnalysisException("Cannot create hive serde table with CREATE TABLE USING")
-    }
     val schema = Option(ctx.colTypeList()).map(createSchema)
     val partitionColumnNames =
       Option(ctx.partitionColumnNames)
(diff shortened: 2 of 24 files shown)


cloud-fan:
cc @yhuai

sql/catalyst/src/main/antlr4/org/apache/spark/sql/catalyst/parser/SqlBase.g4
```diff
@@ -69,16 +69,19 @@ statement
     | ALTER DATABASE identifier SET DBPROPERTIES tablePropertyList     #setDatabaseProperties
     | DROP DATABASE (IF EXISTS)? identifier (RESTRICT | CASCADE)?      #dropDatabase
     | createTableHeader ('(' colTypeList ')')? tableProvider
-        (OPTIONS tablePropertyList)?
+        (OPTIONS options=tablePropertyList)?
         (PARTITIONED BY partitionColumnNames=identifierList)?
-        bucketSpec? (AS? query)?                                       #createTableUsing
+        bucketSpec?
+        (TBLPROPERTIES properties=tablePropertyList)?
+        (COMMENT comment=STRING)?
```
gatorsmile: Do we need to keep the same order? For example, moving `(COMMENT comment=STRING)?` before `(PARTITIONED BY partitionColumnNames=identifierList)?`?

cloud-fan: I think it's more natural to write the table definition first, then the comment, i.e. important things first.

sql/core/src/main/scala/org/apache/spark/sql/execution/SparkSqlParser.scala
```diff
@@ -342,42 +342,46 @@ class SparkSqlAstBuilder(conf: SQLConf) extends AstBuilder {
...
-   *   CREATE [EXTERNAL] TABLE [IF NOT EXISTS] [db_name.]table_name
+   *   CREATE [TEMPORARY] TABLE [IF NOT EXISTS] [db_name.]table_name
    *   USING table_provider
    *   [OPTIONS table_property_list]
    *   [PARTITIONED BY (col_name, col_name, ...)]
    *   [CLUSTERED BY (col_name, col_name, ...)
    *    [SORTED BY (col_name [ASC|DESC], ...)]
    *    INTO num_buckets BUCKETS
    *   ]
+   *   [TBLPROPERTIES (property_name=property_value, ...)]
```
gatorsmile: Here, we need an update. Removed `TBLPROPERTIES` but added new `locationSpec`

gatorsmile:
Just read the design doc. What is the decision about adding `(TBLPROPERTIES tablePropertyList)?` in the CREATE TABLE syntax? So far, .g4 file does not have it.

gatorsmile:
There is a syntax difference in partition column definition between Hive serde tables and data source tables.  In Hive serde tables, the partitioning columns cannot be part of the table schema. Do we need to document the difference, or we can assume users understand this when they convert it?

gatorsmile:
We might need another PR for updating the output of `SHOW CREATE TABLE`, since we recommend users use the new syntax.

gatorsmile:
`CREATE TEMPORARY TABLE` is not supported for all the types of hive serde tables. However, in `CREATE TEMPORARY TABLE` is allowed for creating data souce tables if `AS query` is not specified.

gatorsmile:
Hive does not allow to use a CTAS statement to create a partitioned table, but we allow it in the Create Data Source table syntax.

cloud-fan:
> CREATE TEMPORARY TABLE is not supported for all the types of hive serde tables. However, in CREATE TEMPORARY TABLE is allowed for creating data souce tables if AS query is not specified.

`CREATE TEMPORARY TABLE` is also not allowed for data source tables, we will convert it to temp view and give an error message. Can we remove it in spark 2.2? cc @yhuai @liancheng 

> Hive does not allow to use a CTAS statement to create a partitioned table, but we allow it in the Create Data Source table syntax.

We can add an extra check to make sure we don't use CTAS to create partitioned table.

cloud-fan:
> There is a syntax difference in partition column definition between Hive serde tables and data source tables. In Hive serde tables, the partitioning columns cannot be part of the table schema. Do we need to document the difference, or we can assume users understand this when they convert it?

the hive syntax has data schema and partition schema, while the new syntax only has a table schema(logically data schema + partition schema). This syntax difference already exists between data source table syntax and hive table syntax, and users don't need to convert their old SQL statements, the legacy hive syntax still exists.

gatorsmile:
In the original `Create Hive Serde Table` command, users are allowed to specify the serde properties for `ROW FORMAT SERDE`. It sounds like the unified Create Table command is missing such a capability.

cloud-fan:
@gatorsmile , `OPTIONS` is serde properties, I should document this.

sql/core/src/main/scala/org/apache/spark/sql/execution/SparkSqlParser.scala
```diff
@@ -342,42 +342,46 @@ class SparkSqlAstBuilder(conf: SQLConf) extends AstBuilder {
   }
 
   /**
-   * Create a data source table, returning a [[CreateTable]] logical plan.
+   * Create a table, returning a [[CreateTable]] logical plan.
    *
    * Expected format:
    * {{{
-   *   CREATE [EXTERNAL] TABLE [IF NOT EXISTS] [db_name.]table_name
+   *   CREATE [TEMPORARY] TABLE [IF NOT EXISTS] [db_name.]table_name
```
yhuai: Was `external` a typo at here?

yhuai: oh it is a typo

sql/core/src/main/scala/org/apache/spark/sql/execution/SparkSqlParser.scala
```diff
@@ -342,42 +342,46 @@ class SparkSqlAstBuilder(conf: SQLConf) extends AstBuilder {
...
     val bucketSpec = Option(ctx.bucketSpec()).map(visitBucketSpec)
 
-    // TODO: this may be wrong for non file-based data source like JDBC, which should be external
-    // even there is no `path` in options. We should consider allow the EXTERNAL keyword.
+    val location = Option(ctx.locationSpec).map(visitLocationSpec)
     val storage = DataSource.buildStorageFormatFromOptions(options)
-    val tableType = if (storage.locationUri.isDefined) {
+
+    if (location.isDefined && storage.locationUri.isDefined) {
+      throw new ParseException("Cannot specify LOCATION when there is 'path' in OPTIONS.", ctx)
```
yhuai: Let's be more specific at here. These two approaches are the same and we only want users to use one, right?

sql/core/src/main/scala/org/apache/spark/sql/execution/SparkSqlParser.scala
cloud-fan: yup

sql/hive/src/main/scala/org/apache/spark/sql/hive/execution/HiveOptions.scala
```diff
@@ -0,0 +1,90 @@
...
+ */
+
+package org.apache.spark.sql.hive.execution
+
+import org.apache.spark.sql.catalyst.util.CaseInsensitiveMap
+
+/**
+ * Options for the Hive data source.
+ */
+class HiveOptions(@transient private val parameters: CaseInsensitiveMap) extends Serializable {
```
yhuai: Let's also mention that DetermineHiveSerde will fill in default values based on the file format.

sql/hive/src/main/scala/org/apache/spark/sql/hive/execution/HiveOptions.scala
```diff
@@ -0,0 +1,90 @@
...
+  import HiveOptions._
+
+  def this(parameters: Map[String, String]) = this(new CaseInsensitiveMap(parameters))
+
+  val format = parameters.get(FORMAT).map(_.toLowerCase)
```
yhuai: file format?

sql/hive/src/main/scala/org/apache/spark/sql/hive/execution/HiveOptions.scala
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  if (format.isDefined && inputFormat.isDefined) {
+    throw new IllegalArgumentException("Cannot specify format and inputFormat/outputFormat " +
+      "together for Hive data source.")
+  }
+
+  val serde = parameters.get(SERDE)
+
+  for (f <- format if serde.isDefined) {
```
yhuai: maybe using if is easier to read?

sql/hive/src/main/scala/org/apache/spark/sql/hive/execution/HiveOptions.scala
```diff
@@ -0,0 +1,90 @@
...
+    if (!Set("sequencefile", "textfile", "rcfile").contains(f)) {
+      throw new IllegalArgumentException(s"format '$f' already specifies a serde.")
+    }
+  }
+
+  val containsDelimiters = delimiterOptions.keys.exists(parameters.contains)
```
gatorsmile: Add one more check? The `DELIMITED` clause cannot be used with a custom serde. Previously, it is blocked by our parser. 
```
row_format
  : DELIMITED [FIELDS TERMINATED BY char [ESCAPED BY char]] [COLLECTION ITEMS TERMINATED BY char]
        [MAP KEYS TERMINATED BY char] [LINES TERMINATED BY char]
        [NULL DEFINED AS char]   -- (Note: Available in Hive 0.13 and later)
  | SERDE serde_name [WITH SERDEPROPERTIES (property_name=property_value, property_name=property_value, ...)]
```

sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveStrategies.scala
```diff
@@ -18,14 +18,79 @@
...
+        HiveSerDe.sourceToSerDe(options.format.get) match {
+          case Some(s) =>
+            CatalogStorageFormat.empty.copy(
+              inputFormat = s.inputFormat,
+              outputFormat = s.outputFormat,
+              serde = s.serde)
+          case None =>
+            throw new IllegalArgumentException(s"invalid format: '${options.format.get}'")
+        }
+      } else if (options.inputFormat.isDefined) {
```
yhuai: Maybe we should use a helper function to know if inputFormat and outputFormat are set? The current version assumes that the reader know the internal of HiveOptions.

sql/hive/src/test/scala/org/apache/spark/sql/hive/HiveDDLCommandSuite.scala
```diff
@@ -592,4 +597,77 @@ class HiveDDLCommandSuite extends PlanTest with SQLTestUtils with TestHiveSingle
...
+    assert(table.partitionColumnNames == Seq("c2"))
+    // check the default formats
+    assert(table.storage.serde == Some("org.apache.hadoop.hive.serde2.lazy.LazySimpleSerDe"))
+    assert(table.storage.inputFormat == Some("org.apache.hadoop.mapred.TextInputFormat"))
+    assert(table.storage.outputFormat ==
+      Some("org.apache.hadoop.hive.ql.io.HiveIgnoreKeyTextOutputFormat"))
+
+    val v2 = "CREATE TABLE t (c1 int, c2 int) USING hive CLUSTERED BY (c2) INTO 4 BUCKETS"
+    val e = intercept[AnalysisException](analyzeCreateTable(v2))
+    assert(e.message.contains("Cannot create bucketed Hive serde table"))
```
yhuai: Let's also have a test using both partitioning and bucketing.

sql/hive/src/test/scala/org/apache/spark/sql/hive/execution/HiveDDLSuite.scala
```diff
@@ -1198,4 +1198,23 @@ class HiveDDLSuite
...
+      sql("INSERT INTO t SELECT 1")
+      checkAnswer(spark.table("t"), Row(1))
+
+      sql("CREATE TABLE t2 USING HIVE AS SELECT 1 AS c1, 'a' AS c2")
+      val table2 = spark.sessionState.catalog.getTableMetadata(TableIdentifier("t2"))
+      assert(DDLUtils.isHiveTable(table2))
+      assert(table2.storage.serde == Some("org.apache.hadoop.hive.serde2.lazy.LazySimpleSerDe"))
+      checkAnswer(spark.table("t2"), Row(1, "a"))
+    }
+  }
```
yhuai: Let's also exercise partitioning.

Let's also test a orc's option.

sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveStrategies.scala
```diff
@@ -18,14 +18,79 @@
...
+          locationUri = None,
+          inputFormat = defaultHiveSerde.flatMap(_.inputFormat)
+            .orElse(Some("org.apache.hadoop.mapred.TextInputFormat")),
+          outputFormat = defaultHiveSerde.flatMap(_.outputFormat)
+            .orElse(Some("org.apache.hadoop.hive.ql.io.HiveIgnoreKeyTextOutputFormat")),
+          serde = defaultHiveSerde.flatMap(_.serde)
+            .orElse(Some("org.apache.hadoop.hive.serde2.lazy.LazySimpleSerDe")),
+          compressed = false,
+          properties = Map())
+      }
```
gatorsmile: Can we create a function to generate `defaultStorage`? We are having the [duplicate codes](TextInputFormat) in the parser.

sql/core/src/main/scala/org/apache/spark/sql/internal/HiveSerDe.scala
```diff
@@ -77,4 +79,16 @@ object HiveSerDe {
...
+  def getDefaultStorage(conf: SQLConf): CatalogStorageFormat = {
+    val defaultStorageType = conf.getConfString("hive.default.fileformat", "textfile")
+    val defaultHiveSerde = sourceToSerDe(defaultStorageType)
+    CatalogStorageFormat.empty.copy(
+      inputFormat = defaultHiveSerde.flatMap(_.inputFormat)
+        .orElse(Some("org.apache.hadoop.mapred.TextInputFormat")),
+      outputFormat = defaultHiveSerde.flatMap(_.outputFormat)
+        .orElse(Some("org.apache.hadoop.hive.ql.io.HiveIgnoreKeyTextOutputFormat")),
+      serde = defaultHiveSerde.flatMap(_.serde)
+        .orElse(Some("org.apache.hadoop.hive.serde2.lazy.LazySimpleSerDe")))
```
cloud-fan: it's a little different from what it was. Previously we don't set default serde at [parser](https://github.com/apache/spark/pull/16296/files#diff-1bb4f7bd5a2656f48bcd3c857167a11bL1069), but set it during [analysis](https://github.com/apache/spark/pull/16296/files#diff-c4ed9859978dd6ac271b6a40ee945e4bL54). However, this doesn't make sense and I think we should just provide the default serde at the beginning.

yhuai: Does this version break any test?

cloud-fan: no, because they are same fundamentally.

yhuai:
LGTM.

PR #16299, opened 2016-12-15
[MINOR] Only rename SparkR tar.gz if names mismatch

shivaram:

For release builds the R_PACKAGE_VERSION and VERSION are the same (e.g., 2.1.0). Thus `cp` throws an error which causes the build to fail. 

## How was this patch tested?

Manually by executing the following script
```
set -o pipefail
set -e
set -x

touch a

R_PACKAGE_VERSION=2.1.0
VERSION=2.1.0

if [ "$R_PACKAGE_VERSION" != "$VERSION" ]; then
  cp a a
fi
```

### DIFF
--- a/dev/make-distribution.sh
+++ b/dev/make-distribution.sh
@@ -229,7 +229,9 @@ if [ "$MAKE_R" == "true" ]; then
   # Do not source the check-cran.sh - it should be run from where it is for it to set SPARK_HOME
   NO_TESTS=1 CLEAN_INSTALL=1 "$SPARK_HOME/"R/check-cran.sh
   # Make a copy of R source package matching the Spark release version.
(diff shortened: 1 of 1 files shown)


shivaram:
cc @rxin

felixcheung:
can we `mv` instead?
https://github.com/apache/spark/pull/16249/files#diff-ba2c046d92a1d2b5b417788bfb5cb5f8R232

felixcheung:
I found an issue with snapshot build just now - `R CMD check` doesn't handle the `-SNAPSHOT` part of the file name in SparkR_*tar.gz. Since it's only an issue with SNAPSHOT build I'll open a PR separately.

shivaram:
Ah I see - yes - using `mv` is a simpler option - BTW when do we run `R CMD check` ? Is this for say when users download the tar.gz and run it ? (We can move this discussion to the JIRA)

rxin:
Merging in master/branch-2.1.

PR #16300, opened 2016-12-16
[SPARK-18892][SQL] Alias percentile_approx approx_percentile

rxin:
percentile_approx is the name used in Hive, and approx_percentile is the name used in Presto. approx_percentile is actually more consistent with our approx_count_distinct. Given the cost to alias SQL functions is low (one-liner), it'd be better to just alias them so it is easier to use.

## How was this patch tested?
Technically I could add an end-to-end test to verify this one-line change, but it seemed too trivial to me.

### DIFF
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/FunctionRegistry.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/FunctionRegistry.scala
@@ -252,6 +252,7 @@ object FunctionRegistry {
     expression[Percentile]("percentile"),
     expression[Skewness]("skewness"),
     expression[ApproximatePercentile]("percentile_approx"),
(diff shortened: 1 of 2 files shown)


rxin:
cc @cloud-fan

gatorsmile:
Could we also update [another two more lines](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/ApproximatePercentile.scala#L64-L66)?

```
      > SELECT _FUNC_(10.0, array(0.5, 0.4, 0.1), 100);
       [10.0,10.0,10.0]
      > SELECT _FUNC_(10.0, 0.5, 100);
       10.0
```

rxin:
hm weird that we hard coded the name.

rxin:
updated.

gatorsmile:
LGTM

rxin:
Merging in master/branch-2.1.

PR #16303, opened 2016-12-16
[SPARK-17807][core] Demote scalatest to "provided" in spark-tags.

vanzin:
This avoids exposing scalatest as a transitive dependencies to projects
that depend on spark-core, but don't declare a direct dependency on
scalatest in test scope.

Tested by checking "mvn dependency:tree" with a small pom file that just
declared a dependency on spark-core; scalatest would show up as "compile"
before the change, now it doesn't show at all.

### DIFF
--- a/common/tags/pom.xml
+++ b/common/tags/pom.xml
@@ -38,7 +38,7 @@
     <dependency>
       <groupId>org.scalatest</groupId>
       <artifactId>scalatest_${scala.binary.version}</artifactId>
-      <scope>compile</scope>
+      <scope>provided</scope>
     </dependency>
   </dependencies>
 


ryan-williams:
I appreciate the quick turn-around on this, though it seems like a mis-use of the `provided` scope.

FWIW, I am advocating for:
- `mv common/tags/src/{main,test}/java/org/apache/spark/tags`;
- publish the `test-jar` for the spark-tags module.
  - I know this is trivial in SBT, and I've seen Maven modules do this
  - bdgenomics:utils-misc does it with [this config blob](https://github.com/bigdatagenomics/utils/blob/utils-parent_2.11-0.2.10/utils-misc/pom.xml#L46-L56) afaict?

srowen:
This may work in practice, but it doesn't sound like the right change. The scalatest dependency is correctly non-test because it is used by the tags that this module exposes from its non-test code. 

One solution is to make spark-tags itself provided, because it won't matter if the annotations are present at runtime, as far as I can tell. It's not really provided, but, doesn't matter.

Another is to move the test-oriented tags to src/test/scala, and have tests depend on its test-jar artifact.

ryan-williams:
I filed #16311 splitting out a spark-tags test-jar; turns out my second bullet above is a no-op since all Spark modules' test-jars are already automatically published, so that's neat.

As I noted over on #16311, I now feel that we should just move each of the three test-tags to the modules they are pretty explicitly scoped to (in name and usage), and do away with spark-tags' containing these test-tags altogether, but we can discuss that further over there.

vanzin:
Closing in favor of the other PR.

PR #16304, opened 2016-12-16
[SPARK-18894][SS] Fix event time watermark delay threshold specified in months or years

tdas:

Two changes 
- Fix how delays specified in months and years are translated to milliseconds
- Following up on #16258, not show watermark when there is no watermarking in the query

## How was this patch tested?
Updated and new unit tests


common/unsafe/src/main/java/org/apache/spark/unsafe/types/CalendarInterval.java
```diff
@@ -252,6 +252,9 @@ public static long parseSecondNano(String secondNano) throws IllegalArgumentExce
   public final int months;
   public final long microseconds;
 
+  /**
+   * Return the interval in miliseconds, not including the months in interval.
+   */
   public final long milliseconds() {
```
tdas: @rxin Documented as per your requets

rxin:
event time?

common/unsafe/src/main/java/org/apache/spark/unsafe/types/CalendarInterval.java
rxin: hm why don't you remove this and i will send a patch for docs.

sql/core/src/main/scala/org/apache/spark/sql/Dataset.scala
```diff
@@ -572,6 +572,10 @@ class Dataset[T] private[sql](
     val parsedDelay =
       Option(CalendarInterval.fromString("interval " + delayThreshold))
         .getOrElse(throw new AnalysisException(s"Unable to parse time delay '$delayThreshold'"))
+    // Threshold specified in months/years is non-deterministic
```
marmbrus: I think waiting 1 month for late data is a reasonable use case. Based on the definition of the watermark, its actually okay for us to over estimate this delay too.  Why not take take the max (31 days, leap year)?

rxin: i worry users think it will align with month boundary. It is safer to just say "31 days".

marmbrus: what does "safe" mean in this context?  users must not rely on watermarks for correctness as they can be arbitrarily delayed based on batch boundaries.  I think this error actually confuses the point as its is enforcing precision when this API cannot provide that.

rxin: when i have a record that's 29 days late in feb what should I expect? If we want to just change month to a fix number of days, I'd say just use 30, and then document it clearly in the API (e.g. "if month is specified, 1 month = 30 days").

rxin: sgtm

rxin:
fyi the title has a typo

sql/core/src/test/scala/org/apache/spark/sql/streaming/EventTimeWatermarkSuite.scala
```diff
@@ -124,6 +137,29 @@ class WatermarkSuite extends StreamTest with BeforeAndAfter with Logging {
     )
   }
 
+  test("delay in years handled correctly") {
+    val input = MemoryStream[Long]
+    val aggWithWatermark = input.toDF()
+      .withColumn("eventTime", $"value".cast("timestamp"))
+      .withWatermark("eventTime", "1 month")
```
marmbrus: This seems to be testing months, but the title says years?

tdas: yeah. sorry .. i pushed too early. i updated the test now.

marmbrus: Just wanted to make sure I wasn't missing something :)

marmbrus:
LGTM

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamExecution.scala
```diff
@@ -387,7 +387,7 @@ class StreamExecution(
         lastExecution.executedPlan.collect {
           case e: EventTimeWatermarkExec if e.eventTimeStats.value.count > 0 =>
             logDebug(s"Observed event time stats: ${e.eventTimeStats.value}")
-            e.eventTimeStats.value.max - e.delay.milliseconds
+            math.max(0, e.eventTimeStats.value.max - e.delayMs)
```
zsxwing: `newWatermarkMs` can be negative. Image the user is processing some very old data which is before 1970.

tdas: i think a lot of things are going to break for that usecase. I dont think our sql functions, Java time format, etc are even designed to handle negative millis. The way I found this issue is that when i tried for convert -ve watermark to formatted string, it gave a very weird date. So I dont think we should add complexity for that use case. 
@marmbrus any thoughts?

zsxwing: It sounds like a SQL bug. IMO, a negative timestamp is valid, it's not a negative date, just a date before 1970.

tdas: okay, I think I was doing something to goof up the conversion using SimpleDateFormat. I am fixing that.

zsxwing:
LGTM

zsxwing:

sql/core/src/test/scala/org/apache/spark/sql/streaming/EventTimeWatermarkSuite.scala
```diff
@@ -124,6 +137,33 @@ class WatermarkSuite extends StreamTest with BeforeAndAfter with Logging {
...
+
+    testStream(aggWithWatermark)(
+      AddData(input, currentTimeMs / 1000),
+      CheckAnswer(),
+      AddData(input, currentTimeMs / 1000),
+      CheckAnswer(),
+      assertEventStats { e =>
+        assert(timestampFormat.parse(e.get("max")).getTime === (currentTimeMs / 1000) * 1000)
+        val watermarkTime = timestampFormat.parse(e.get("watermark"))
+        assert(monthsSinceEpoch(currentTime) - monthsSinceEpoch(watermarkTime) === 29)
```
gatorsmile: All the recent builds failed in this check.
```
[info] - delay in months and years handled correctly *** FAILED *** (325 milliseconds)
[info]   30 did not equal 29 (EventTimeWatermarkSuite.scala:193)
```

zsxwing: Thanks for reporting it. Working on a PR.

zsxwing: Fixing in #16449

PR #16308, opened 2016-12-16
[SPARK-18936][SQL] Infrastructure for session local timezone support.

ueshin:

As of Spark 2.1, Spark SQL assumes the machine timezone for datetime manipulation, which is bad if users are not in the same timezones as the machines, or if different users have different timezones.

We should introduce a session local timezone setting that is used for execution.

An explicit non-goal is locale handling.

### Semantics

Setting the session local timezone means that the timezone-aware expressions listed below should use the timezone to evaluate values, and also it should be used to convert (cast) between string and timestamp or between timestamp and date.

- `CurrentDate`
- `CurrentBatchTimestamp`
- `Hour`
- `Minute`
- `Second`
- `DateFormatClass`
- `ToUnixTimestamp`
- `UnixTimestamp`
- `FromUnixTime`

and below are implicitly timezone-aware through cast from timestamp to date:

- `DayOfYear`
- `Year`
- `Quarter`
- `Month`
- `DayOfMonth`
- `WeekOfYear`
- `LastDay`
- `NextDay`
- `TruncDate`

For example, if you have timestamp `"2016-01-01 00:00:00"` in `GMT`, the values evaluated by some of timezone-aware expressions are:

```scala
scala> val df = Seq(new java.sql.Timestamp(1451606400000L)).toDF("ts")
df: org.apache.spark.sql.DataFrame = [ts: timestamp]

scala> df.selectExpr("cast(ts as string)", "year(ts)", "month(ts)", "dayofmonth(ts)", "hour(ts)", "minute(ts)", "second(ts)").show(truncate = false)
+-------------------+----------------------+-----------------------+----------------------------+--------+----------+----------+
|ts                 |year(CAST(ts AS DATE))|month(CAST(ts AS DATE))|dayofmonth(CAST(ts AS DATE))|hour(ts)|minute(ts)|second(ts)|
+-------------------+----------------------+-----------------------+----------------------------+--------+----------+----------+
|2016-01-01 00:00:00|2016                  |1                      |1                           |0       |0         |0         |
+-------------------+----------------------+-----------------------+----------------------------+--------+----------+----------+
```

whereas setting the session local timezone to `"PST"`, they are:

```scala
scala> spark.conf.set("spark.sql.session.timeZone", "PST")

scala> df.selectExpr("cast(ts as string)", "year(ts)", "month(ts)", "dayofmonth(ts)", "hour(ts)", "minute(ts)", "second(ts)").show(truncate = false)
+-------------------+----------------------+-----------------------+----------------------------+--------+----------+----------+
|ts                 |year(CAST(ts AS DATE))|month(CAST(ts AS DATE))|dayofmonth(CAST(ts AS DATE))|hour(ts)|minute(ts)|second(ts)|
+-------------------+----------------------+-----------------------+----------------------------+--------+----------+----------+
|2015-12-31 16:00:00|2015                  |12                     |31                          |16      |0         |0         |
+-------------------+----------------------+-----------------------+----------------------------+--------+----------+----------+
```

Notice that even if you set the session local timezone, it affects only in `DataFrame` operations, neither in `Dataset` operations, `RDD` operations nor in `ScalaUDF`s. You need to properly handle timezone by yourself.

### Design of the fix

I introduced an analyzer to pass session local timezone to timezone-aware expressions and modified DateTimeUtils to take the timezone argument.

## How was this patch tested?

Existing tests and added tests for timezone aware expressions.


ueshin:
I'd like to discuss the boundary of session local timezone.

I assumed it affects only in `DataFrame` operations, neither in `Dataset` operations, before import from `RDD`, after export to `RDD` nor in `ScalaUDF`s because we can't control their timezones, which are basically based on system timezone, i.e. `TimeZone.getDefault()` value.
(And also for backward-compatibility)

What do you think?

rxin:
Can you document the semantics of time zones and when they are used?

ueshin:
@rxin Did you mean I should add a document to `sql-programming-guide.md` file or somewhere else, or describe it here?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/Cast.scala
```diff
@@ -135,6 +146,14 @@ case class Cast(child: Expression, dataType: DataType) extends UnaryExpression w
 
   override def nullable: Boolean = Cast.forceNullable(child.dataType, dataType) || child.nullable
 
+  override def timeZoneResolved: Boolean =
+    (!(childrenResolved && Cast.needTimeZone(child.dataType, dataType))) || super.timeZoneResolved
+
+  override lazy val resolved: Boolean =
+    childrenResolved && checkInputDataTypes().isSuccess && timeZoneResolved
+
+  override def withTimeZone(zoneId: String): TimeZoneAwareExpression = copy(zoneId = zoneId)
```
rxin: this is just a copy ctor isn't it? Maybe no need to add this? Not a big deal though.

```
copy(zoneId = zoneId)
```

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/datetimeExpressions.scala
```diff
@@ -30,21 +30,44 @@ import org.apache.spark.sql.types._
 import org.apache.spark.unsafe.types.{CalendarInterval, UTF8String}
 
 /**
+ * Common base class for time zone aware expressions.
+ */
+trait TimeZoneAwareExpression extends Expression {
+
+  def zoneId: String
```
rxin: is the reason you are using null rather than option to avoid a bunch of gets?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/Cast.scala
```diff
@@ -110,6 +110,14 @@ object Cast {
     case (_: FractionalType, _: IntegralType) => true  // NaN, infinity
     case _ => false
   }
+
+  def needTimeZone(from: DataType, to: DataType): Boolean = (from, to) match {
```
rxin: i think it's important to document this ...

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/Cast.scala
```diff
@@ -120,7 +128,10 @@ object Cast {
       > SELECT _FUNC_('10' as int);
        10
   """)
-case class Cast(child: Expression, dataType: DataType) extends UnaryExpression with NullIntolerant {
+case class Cast(child: Expression, dataType: DataType, zoneId: String = null)
```
rxin: Also we should add classdoc to explain what zoneId is. I'd probably call it timeZoneId.

rxin:
I'd rename all the zoneId to timeZoneId to reduce confusion ..

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
```diff
@@ -180,7 +181,7 @@ class Analyzer(
               case ne: NamedExpression => ne
               case e if !e.resolved => u
               case g: Generator => MultiAlias(g, Nil)
-              case c @ Cast(ne: NamedExpression, _) => Alias(c, ne.name)()
+              case c @ Cast(ne: NamedExpression, _, _) => Alias(c, ne.name)()
```
rxin: if we add a Cast.unapply that returns only the first two arguments, we can reduce a lot of the cast match changes. Not sure if it is worth it though.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/interface.scala
```diff
@@ -111,7 +112,8 @@ case class CatalogTablePartition(
    */
   def toRow(partitionSchema: StructType): InternalRow = {
     InternalRow.fromSeq(partitionSchema.map { field =>
-      Cast(Literal(spec(field.name)), field.dataType).eval()
+      Cast(Literal(spec(field.name)), field.dataType,
+        DateTimeUtils.defaultTimeZone().getID).eval()
```
rxin: could this change the behavior on how we interpret partition values when timezone settings change?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/Cast.scala
rxin: maybe an extra unapply is probably a bad idea, since then we can miss a pattern match.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/datetimeExpressions.scala
```diff
@@ -30,21 +30,44 @@ import org.apache.spark.sql.types._
...
+
+  def timeZoneResolved: Boolean = zoneId != null
+
+  def withTimeZone(zoneId: String): TimeZoneAwareExpression
+
+  def timeZone: TimeZone = TimeZone.getTimeZone(zoneId)
```
rxin: should this be a lazy val? otherwise it is pretty expensive to keep creating a new timezone object per row in the interpreted path.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/datetimeExpressions.scala
ueshin: Yes, I wanted to avoid a bunch of gets.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/datetimeExpressions.scala
ueshin: You are right. I'll use `lazy val`.

ueshin (commented):
@rxin Thank you for your review.
I'll address your comments soon.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/interface.scala
ueshin: Hmm, now I think we should use timezone settings for partition values, because the values are also parts of data so they should be affected by the settings.

ueshin:
Jenkins, retest this please.

rxin:
This is getting pretty close. Can you update the description to define more formally semantics for timezone?

In addition, I think we also need to have a follow-up pr to focus on data sources (how should CSV/JSON interpret timezones when parsing timestamps).

ueshin:
@rxin I updated the description. Is it enough for you?

rxin:
Thanks this looks great.

Couple things:

1. Can you change the referenced JIRA to https://issues.apache.org/jira/browse/SPARK-18936

2. We should do a more detailed pass to make sure there isn't any issue with performance for the impacted expressions (e.g. don't create a new timezone object or do hash lookups per row).

ueshin:
@rxin I'd like to have a follow-up pr related to partition values.
I didn't include it to this pr, but I think we need it.

rxin:
Can you create a subtask at https://issues.apache.org/jira/browse/SPARK-18350 ?

ueshin:
@rxin I see, created.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/Cast.scala
```diff
@@ -135,6 +167,15 @@ case class Cast(child: Expression, dataType: DataType) extends UnaryExpression w
 
   override def nullable: Boolean = Cast.forceNullable(child.dataType, dataType) || child.nullable
 
+  override def timeZoneResolved: Boolean =
```
hvanhovell: This a mental note. A timezone is resolved when:
- We don't need one
- Or is has been resolve

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/datetimeExpressions.scala
```diff
@@ -78,11 +108,21 @@ case class CurrentTimestamp() extends LeafExpression with CodegenFallback {
...
-case class CurrentBatchTimestamp(timestampMs: Long, dataType: DataType)
-  extends LeafExpression with Nondeterministic with CodegenFallback {
+case class CurrentBatchTimestamp(timestampMs: Long, dataType: DataType, timeZoneId: String = null)
+  extends LeafExpression with TimeZoneAwareExpression with Nondeterministic with CodegenFallback {
+
+  def this(timestampMs: Long, dataType: DataType) = this(timestampMs, dataType, null)
 
   override def nullable: Boolean = false
 
+  override def timeZoneResolved: Boolean = dataType != DateType || super.timeZoneResolved
```
hvanhovell: Doesn't a timestamp also need a timezone?

hvanhovell: This is because it is the current timestamp right?

sql/core/src/main/scala/org/apache/spark/sql/execution/QueryExecution.scala
```diff
@@ -146,18 +146,20 @@ class QueryExecution(val sparkSession: SparkSession, val logical: LogicalPlan) {
 
     /** Implementation following Hive's TimestampWritable.toString */
     def formatTimestamp(timestamp: Timestamp): String = {
+      val timestampFormat =
+        DateTimeUtils.getThreadLocalTimestampFormat(DateTimeUtils.defaultTimeZone())
```
hvanhovell: Shouldn't this also use the session timezone?

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/FileFormatWriter.scala
```diff
@@ -288,7 +289,7 @@ object FileFormatWriter extends Logging {
         val escaped = ScalaUDF(
           ExternalCatalogUtils.escapePathName _,
           StringType,
-          Seq(Cast(c, StringType)),
+          Seq(Cast(c, StringType, DateTimeUtils.defaultTimeZone().getID)),
```
hvanhovell: Same question? Why not the session timezone?

hvanhovell (commented):
@ueshin this looks pretty solid. I have left a few questions regarding current timestamps (I think understand this), and the use of timezone outside of expressions.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala
```diff
@@ -104,6 +104,7 @@ class Analyzer(
       ResolveAggregateFunctions ::
       TimeWindowing ::
       ResolveInlineTables ::
+      ResolveTimeZone ::
```
ueshin: @hvanhovell Thank you for your suggestion.
I overrode `resolved` of `TimeZoneAwareExpression`s, then we needed to add `ResolveTimeZone` to `Resolution` batch, but I found we don't need to worry about the resolution because to have the timezone or not doesn't affect the resolution and we only need to run once now.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/Cast.scala
ueshin: Yes, that's right.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/datetimeExpressions.scala
ueshin: Yes, timestamp always have the same value regardless of timezone.

sql/core/src/main/scala/org/apache/spark/sql/execution/QueryExecution.scala
ueshin: Good catch. I'll modify it.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/FileFormatWriter.scala
ueshin: This seems related to partition values.
We'll send follow-up prs about partition values ([SPARK-18939](https://issues.apache.org/jira/browse/SPARK-18939)), let's discuss there.

rxin:
@hvanhovell anything else to do here other than bringing it up to date?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/Cast.scala
```diff
@@ -512,8 +550,9 @@ case class Cast(child: Expression, dataType: DataType) extends UnaryExpression w
         (c, evPrim, evNull) => s"""$evPrim = UTF8String.fromString(
           org.apache.spark.sql.catalyst.util.DateTimeUtils.dateToString($c));"""
       case TimestampType =>
+        val tz = ctx.addReferenceObj("timeZone", timeZone)
```
cloud-fan: we should use `addReferenceMinorObj` to avoid adding member variables.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/datetimeExpressions.scala
```diff
@@ -30,21 +30,48 @@ import org.apache.spark.sql.types._
...
+trait TimeZoneAwareExpression extends Expression {
+
+  /** the timezone ID. */
+  def timeZoneId: String
+
+  /**
+   * Returns true if the timezone ID is resolved.
+   * If not resolved, the analyzer [[ResolveTimeZone]] will resolve with session local timezone.
+   */
+  def timeZoneResolved: Boolean = timeZoneId != null
```
cloud-fan: Does it worth to add such a concept? What if we just set session local timezone for every `TimeZoneAwareExpression`? It seems to me that it's ok to call `withTimeZone` blindly, as the `ResolveTimeZone` rule will only run once.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/datetimeExpressions.scala
```diff
@@ -30,21 +30,48 @@ import org.apache.spark.sql.types._
 import org.apache.spark.unsafe.types.{CalendarInterval, UTF8String}
 
 /**
+ * Common base class for time zone aware expressions.
+ */
+trait TimeZoneAwareExpression extends Expression {
+
+  /** the timezone ID. */
+  def timeZoneId: String
```
cloud-fan: shall we use `Option[String]`?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/datetimeExpressions.scala
```diff
@@ -30,21 +30,48 @@ import org.apache.spark.sql.types._
 import org.apache.spark.unsafe.types.{CalendarInterval, UTF8String}
 
 /**
+ * Common base class for time zone aware expressions.
+ */
+trait TimeZoneAwareExpression extends Expression {
+
+  /** the timezone ID. */
```
cloud-fan: this comment just repeat the variable name...

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/Cast.scala
ueshin: I see. I'll modify to use it.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/datetimeExpressions.scala
ueshin: I see. I'll modify to use it.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/datetimeExpressions.scala
ueshin: I'll update it.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/datetimeExpressions.scala
ueshin: I think that there might be a case that we let users explicitly set the timezone id, e.g. adding such functions to `functions.scala`. If so, we can't simply replace the timezone id by session local timezone.

cloud-fan: do you mean we may allow users to set timezone id for every `TimeZoneAwareExpression` in the future?

ueshin: Yes, I do. What do you think?

cloud-fan: that makes sense, but we only need to add an if condition in `ResolveTimeZone`: `if e.timeZoneId.isEmpty => ...` right?

ueshin: Ah, I see, you're right. I'll update.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/interface.scala
```diff
@@ -111,7 +112,8 @@ case class CatalogTablePartition(
    */
   def toRow(partitionSchema: StructType): InternalRow = {
     InternalRow.fromSeq(partitionSchema.map { field =>
-      Cast(Literal(spec(field.name)), field.dataType).eval()
+      Cast(Literal(spec(field.name)), field.dataType,
+        Option(DateTimeUtils.defaultTimeZone().getID)).eval()
```
cloud-fan: can we make this a default value? e.g.
```
trait TimeZoneAwareExpression extends Expression {
  def timeZoneId: Option[String]

  @transient lazy val timeZone: TimeZone =
    TimeZone.getTimeZone(timeZoneId.getOrElse(DateTimeUtils.defaultTimeZone().getID))
 }
```

cloud-fan: by doing this, we can reduce code changes to the tests.

rxin: but we don't want to have callers forgetting to specify

ueshin: I agree with @rxin.
We should explicitly specify the timezone id or use session local timezone resolved by the `ResolveTimeZone` rule.

ueshin:
Jenkins, retest this please.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/finishAnalysis.scala
```diff
@@ -41,13 +46,18 @@ object ReplaceExpressions extends Rule[LogicalPlan] {
...
 
     plan transformAllExpressions {
-      case CurrentDate() => currentDate
+      case CurrentDate(Some(timeZoneId)) =>
+        currentDates.getOrElseUpdate(timeZoneId, {
+          Literal.create(
+            DateTimeUtils.millisToDays(timestamp / 1000L, TimeZone.getTimeZone(timeZoneId)),
+            DateType)
+        })
       case CurrentTimestamp() => currentTime
```
cloud-fan: why `CurrentTimestamp` is not timezone aware?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/datetimeExpressions.scala
```diff
@@ -195,19 +231,26 @@ case class Hour(child: Expression) extends UnaryExpression with ImplicitCastInpu
       > SELECT _FUNC_('2009-07-30 12:58:59');
        58
   """)
-case class Minute(child: Expression) extends UnaryExpression with ImplicitCastInputTypes {
+case class Minute(child: Expression, timeZoneId: Option[String] = None)
```
cloud-fan: Logically `Minute`/`Second` are not timezone-aware right?

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/CastSuite.scala
```diff
@@ -32,20 +34,20 @@ import org.apache.spark.unsafe.types.UTF8String
...
-      case _ => Cast(Literal(v), targetType)
+      case lit: Expression => Cast(lit, targetType, timeZoneId)
+      case _ => Cast(Literal(v), targetType, timeZoneId)
     }
   }
 
   // expected cannot be null
-  private def checkCast(v: Any, expected: Any): Unit = {
-    checkEvaluation(cast(v, Literal(expected).dataType), expected)
+  private def checkCast(v: Any, expected: Any, timeZoneId: Option[String] = None): Unit = {
```
cloud-fan: where do you call this method and set the `timeZoneId` parameter?

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/CastSuite.scala
```diff
@@ -316,30 +329,43 @@ class CastSuite extends SparkFunSuite with ExpressionEvalHelper {
...
-    checkEvaluation(cast(cast(new Timestamp(c.getTimeInMillis), StringType), TimestampType),
-      c.getTimeInMillis * 1000)
-    c = Calendar.getInstance()
-    c.set(2015, 10, 1, 2, 30, 0)
-    checkEvaluation(cast(cast(new Timestamp(c.getTimeInMillis), StringType), TimestampType),
-      c.getTimeInMillis * 1000)
+    val ts = withDefaultTimeZone(TimeZoneGMT)(Timestamp.valueOf(nts))
+
+    for (tz <- ALL_TIMEZONES) {
+      val timeZoneId = Option(tz.getID)
```
cloud-fan: when will `timeZoneId` be None here?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/finishAnalysis.scala
rxin: timestamp is an absolute value -- timezone only matters when converting a timestamp into a displayable value (string) or date.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/datetimeExpressions.scala
ueshin: No, `Minute` / `Second` should also be timezone-aware because `TimeZone` can have millisecond offset. Actually, there are some timezones which have offset of some hours and 30 minutes. There is no timezone which has non-zero second offset for now, though.

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/CastSuite.scala
ueshin: Ah, I prepared for the parameter before but this is not called with `timeZoneId` parameter.
I'll remove this parameter.

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/CastSuite.scala
ueshin: `timeZoneId` here will never be `None`, but I want to reuse the value to avoid a bunch of `Option(tz.getID)` call.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/dsl/package.scala
```diff
@@ -107,6 +107,7 @@ package object dsl {
       UnresolvedExtractValue(expr, Literal(fieldName))
 
     def cast(to: DataType): Expression = Cast(expr, to)
+    def cast(to: DataType, timeZoneId: String): Expression = Cast(expr, to, Option(timeZoneId))
```
cloud-fan: any place call this method?

ueshin: I used it [here](https://github.com/apache/spark/pull/16308/files#diff-e8dddba2915a147970654aa93bee30a7R264), but we can use `Cast` directly.
I'll modify it.

ueshin:
Jenkins, retest this please.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/datetimeExpressions.scala
```diff
@@ -30,21 +30,42 @@ import org.apache.spark.sql.types._
...
+ */
+trait TimeZoneAwareExpression extends Expression {
+
+  /** the timezone ID to be used to evaluate value. */
+  def timeZoneId: Option[String]
+
+  /** Returns a copy of this expression with the specified timeZoneId. */
+  def withTimeZone(timeZoneId: String): TimeZoneAwareExpression
+
+  @transient lazy val timeZone: TimeZone = TimeZone.getTimeZone(timeZoneId.get)
```
cloud-fan: how about we make this a `implicit val`? then we can save a lot of code changes, just need to update `DateTimeUtils.daysToMillis` to `def daysToMillis(days: Int)(implicit timeZone: TimeZone)`

rxin: hm i prefer being explicit. Implicit is always pretty risky.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/datetimeExpressions.scala
```diff
@@ -544,12 +630,15 @@ abstract class UnixTime extends BinaryExpression with ExpectsInputTypes {
...
         val sdf = classOf[SimpleDateFormat].getName
+        val formatterName = ctx.freshName("formatter")
         nullSafeCodeGen(ctx, ev, (string, format) => {
           s"""
             try {
-              ${ev.value} =
-                (new $sdf($format.toString())).parse($string.toString()).getTime() / 1000L;
+              $sdf $formatterName = new $sdf($format.toString(), java.util.Locale.US);
+              $formatterName.setTimeZone($tz);
+              ${ev.value} = $formatterName.parse($string.toString()).getTime() / 1000L;
```
cloud-fan: may not related to this PR: we should cache the date format instance, e.g.
```
if (formatter == null) {
  ev.copy(isNull = "true")
} else {
  val formatterName = ctx.addRefrenceMinorObj(formatter)
  ...
}
```

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/DateTimeUtils.scala
```diff
@@ -60,7 +60,7 @@ object DateTimeUtils {
   final val TimeZoneGMT = TimeZone.getTimeZone("GMT")
   final val MonthOf31Days = Set(1, 3, 5, 7, 8, 10, 12)
 
-  @transient lazy val defaultTimeZone = TimeZone.getDefault
+  def defaultTimeZone(): TimeZone = TimeZone.getDefault()
```
cloud-fan: why change `lazy val` to `def`?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/DateTimeUtils.scala
```diff
@@ -69,49 +69,68 @@ object DateTimeUtils {
...
   // `SimpleDateFormat` is not thread-safe.
   private val threadLocalDateFormat = new ThreadLocal[DateFormat] {
     override def initialValue(): SimpleDateFormat = {
       new SimpleDateFormat("yyyy-MM-dd", Locale.US)
     }
   }
 
+  def getThreadLocalDateFormat(): DateFormat = {
+    val sdf = threadLocalDateFormat.get()
+    sdf.setTimeZone(defaultTimeZone())
```
cloud-fan: here we always set it to `defaultTimeZone`, can we just set it in `threadLocalDateFormat.initialValue`? then we don't need this method.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/DateTimeUtils.scala
```diff
@@ -69,49 +69,68 @@ object DateTimeUtils {
...
+    sdf
+  }
+
   // we should use the exact day as Int, for example, (year, month, day) -> day
   def millisToDays(millisUtc: Long): SQLDate = {
```
cloud-fan: now these methods without timezone parameter are only used in tests right?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/DateTimeUtils.scala
cloud-fan: is it because we removed `threadLocalLocalTimeZone`?

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/CastSuite.scala
```diff
@@ -107,108 +109,119 @@ class CastSuite extends SparkFunSuite with ExpressionEvalHelper {
...
-
-    checkEvaluation(Cast(Literal("2015-03-18 123142"), TimestampType), null)
-    checkEvaluation(Cast(Literal("2015-03-18T123123"), TimestampType), null)
-    checkEvaluation(Cast(Literal("2015-03-18X"), TimestampType), null)
-    checkEvaluation(Cast(Literal("2015/03/18"), TimestampType), null)
-    checkEvaluation(Cast(Literal("2015.03.18"), TimestampType), null)
-    checkEvaluation(Cast(Literal("20150318"), TimestampType), null)
-    checkEvaluation(Cast(Literal("2015-031-8"), TimestampType), null)
-    checkEvaluation(Cast(Literal("2015-03-18T12:03:17-0:70"), TimestampType), null)
+    for (tz <- ALL_TIMEZONES) {
```
cloud-fan: I didn't check this code block line by line, I assume it just iterates all time zone ids and adds it to all the `Cast`. @ueshin please highlight important changes here, if there are any.

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/CastSuite.scala
```diff
@@ -316,30 +329,43 @@ class CastSuite extends SparkFunSuite with ExpressionEvalHelper {
...
-
-    var c = Calendar.getInstance()
-    c.set(2015, 2, 8, 2, 30, 0)
-    checkEvaluation(cast(cast(new Timestamp(c.getTimeInMillis), StringType), TimestampType),
-      c.getTimeInMillis * 1000)
-    c = Calendar.getInstance()
-    c.set(2015, 10, 1, 2, 30, 0)
-    checkEvaluation(cast(cast(new Timestamp(c.getTimeInMillis), StringType), TimestampType),
-      c.getTimeInMillis * 1000)
+    val ts = withDefaultTimeZone(TimeZoneGMT)(Timestamp.valueOf(nts))
```
cloud-fan: where did we define `withDefaultTimeZone`?

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/DateExpressionsSuite.scala
```diff
@@ -143,72 +159,114 @@ class DateExpressionsSuite extends SparkFunSuite with ExpressionEvalHelper {
   }
 
   test("Seconds") {
-    assert(Second(Literal.create(null, DateType)).resolved === false)
-    checkEvaluation(Second(Cast(Literal(d), TimestampType)), 0)
-    checkEvaluation(Second(Cast(Literal(sdf.format(d)), TimestampType)), 15)
-    checkEvaluation(Second(Literal(ts)), 15)
+    assert(Second(Literal.create(null, DateType), gmtId).resolved === false)
+    assert(Second(Cast(Literal(d), TimestampType), null).resolved === true)
```
cloud-fan: why we use `null` instead of `gmtId` here?

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/DateExpressionsSuite.scala
```diff
@@ -143,72 +159,114 @@ class DateExpressionsSuite extends SparkFunSuite with ExpressionEvalHelper {
...
+    checkEvaluation(DateFormatClass(Literal(ts), Literal("H"), jstId), "22")
   }
 
   test("Hour") {
-    assert(Hour(Literal.create(null, DateType)).resolved === false)
-    checkEvaluation(Hour(Cast(Literal(d), TimestampType)), 0)
-    checkEvaluation(Hour(Cast(Literal(sdf.format(d)), TimestampType)), 13)
-    checkEvaluation(Hour(Literal(ts)), 13)
+    assert(Hour(Literal.create(null, DateType), gmtId).resolved === false)
+    assert(Hour(Literal(ts), null).resolved === true)
```
cloud-fan: same question, why use `null` instead of `gmtId`?

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/DateExpressionsSuite.scala
```diff
@@ -143,72 +159,114 @@ class DateExpressionsSuite extends SparkFunSuite with ExpressionEvalHelper {
...
-    checkConsistencyBetweenInterpretedAndCodegen(Hour, TimestampType)
   }
 
   test("Minute") {
-    assert(Minute(Literal.create(null, DateType)).resolved === false)
-    checkEvaluation(Minute(Cast(Literal(d), TimestampType)), 0)
-    checkEvaluation(Minute(Cast(Literal(sdf.format(d)), TimestampType)), 10)
-    checkEvaluation(Minute(Literal(ts)), 10)
+    assert(Minute(Literal.create(null, DateType), gmtId).resolved === false)
+    assert(Minute(Literal(ts), null).resolved === true)
```
cloud-fan: same here

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/DateExpressionsSuite.scala
```diff
@@ -219,7 +277,8 @@ class DateExpressionsSuite extends SparkFunSuite with ExpressionEvalHelper {
       DateAdd(Literal(Date.valueOf("2016-02-28")), Literal(-365)),
       DateTimeUtils.fromJavaDate(Date.valueOf("2015-02-28")))
     checkEvaluation(DateAdd(Literal.create(null, DateType), Literal(1)), null)
-    checkEvaluation(DateAdd(Literal(Date.valueOf("2016-02-28")), Literal.create(null, IntegerType)),
+    checkEvaluation(
+      DateAdd(Literal(Date.valueOf("2016-02-28")), Literal.create(null, IntegerType)),
```
cloud-fan: unnecessary change.

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/DateExpressionsSuite.scala
```diff
@@ -238,7 +297,8 @@ class DateExpressionsSuite extends SparkFunSuite with ExpressionEvalHelper {
       DateSub(Literal(Date.valueOf("2015-01-01")), Literal(-1)),
       DateTimeUtils.fromJavaDate(Date.valueOf("2015-01-02")))
     checkEvaluation(DateSub(Literal.create(null, DateType), Literal(1)), null)
-    checkEvaluation(DateSub(Literal(Date.valueOf("2016-02-28")), Literal.create(null, IntegerType)),
+    checkEvaluation(
+      DateSub(Literal(Date.valueOf("2016-02-28")), Literal.create(null, IntegerType)),
```
cloud-fan: unnecessary change.

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/util/DateTimeUtilsSuite.scala
```diff
@@ -177,180 +177,186 @@ class DateTimeUtilsSuite extends SparkFunSuite {
...
-    c.set(Calendar.MILLISECOND, 123)
-    assert(stringToTimestamp(
-      UTF8String.fromString("2015-03-18T12:03:17.123+07:30")).get === c.getTimeInMillis * 1000)
-
-    c = Calendar.getInstance(TimeZone.getTimeZone("GMT+07:30"))
-    c.set(2015, 2, 18, 12, 3, 17)
-    c.set(Calendar.MILLISECOND, 123)
-    assert(stringToTimestamp(
-      UTF8String.fromString("2015-03-18T12:03:17.123121+7:30")).get ===
+    for (tz <- DateTimeTestUtils.ALL_TIMEZONES) {
```
cloud-fan: I assume you just iterate all timezone ids and use it as timezone parameter to `DateTimeUtil` methods.  please highlight important/special changes if there are any.

sql/core/src/main/scala/org/apache/spark/sql/Column.scala
```diff
@@ -177,7 +177,7 @@ class Column(val expr: Expression) extends Logging {
     // NamedExpression under this Cast.
     case c: Cast =>
       c.transformUp {
-        case Cast(ne: NamedExpression, to) => UnresolvedAlias(Cast(ne, to))
+        case Cast(ne: NamedExpression, to, timeZoneId) => UnresolvedAlias(Cast(ne, to, timeZoneId))
```
cloud-fan: nit: `case c @ Cast(_: NamedExpression, _, _) => UnresolvedAlias(c)`

sql/core/src/main/scala/org/apache/spark/sql/Dataset.scala
```diff
@@ -260,6 +264,10 @@ class Dataset[T] private[sql](
           case binary: Array[Byte] => binary.map("%02X".format(_)).mkString("[", " ", "]")
           case array: Array[_] => array.mkString("[", ", ", "]")
           case seq: Seq[_] => seq.mkString("[", ", ", "]")
+          case d: Date =>
+            DateTimeUtils.dateToString(DateTimeUtils.fromJavaDate(d))
+          case ts: Timestamp =>
+            DateTimeUtils.timestampToString(DateTimeUtils.fromJavaTimestamp(ts), timeZone)
           case _ => cell.toString
```
cloud-fan: what's the previous result? What will `Date.toString` and `Timestamp.toString` output?

sql/core/src/main/scala/org/apache/spark/sql/Dataset.scala
```diff
@@ -250,6 +252,8 @@ class Dataset[T] private[sql](
     val hasMoreData = takeResult.length > numRows
     val data = takeResult.take(numRows)
 
+    lazy val timeZone = TimeZone.getTimeZone(sparkSession.sessionState.conf.sessionLocalTimeZone)
```
cloud-fan: This is only used once, I think we can inline it.

sql/core/src/main/scala/org/apache/spark/sql/execution/OptimizeMetadataOnlyQuery.scala
```diff
@@ -104,7 +105,8 @@ case class OptimizeMetadataOnlyQuery(
             val partAttrs = getPartitionAttrs(relation.catalogTable.partitionColumnNames, relation)
             val partitionData = catalog.listPartitions(relation.catalogTable.identifier).map { p =>
               InternalRow.fromSeq(partAttrs.map { attr =>
-                Cast(Literal(p.spec(attr.name)), attr.dataType).eval()
+                Cast(Literal(p.spec(attr.name)), attr.dataType,
+                  Option(DateTimeUtils.defaultTimeZone().getID)).eval()
```
cloud-fan: shall we use `conf.sessionLocalTimeZone` here?

sql/core/src/main/scala/org/apache/spark/sql/execution/QueryExecution.scala
```diff
@@ -200,8 +189,10 @@ class QueryExecution(val sparkSession: SparkSession, val logical: LogicalPlan) {
             toHiveStructString((key, kType)) + ":" + toHiveStructString((value, vType))
         }.toSeq.sorted.mkString("{", ",", "}")
       case (null, _) => "NULL"
-      case (d: Int, DateType) => new java.util.Date(DateTimeUtils.daysToMillis(d)).toString
```
cloud-fan: ah, so we will never hit this branch before, as the type should be `Date` instead of `Int`

sql/core/src/main/scala/org/apache/spark/sql/execution/QueryExecution.scala
```diff
@@ -139,27 +140,15 @@ class QueryExecution(val sparkSession: SparkSession, val logical: LogicalPlan) {
       }
   }
 
+  @transient
+  private lazy val sessionLocalTimeZone =
```
cloud-fan: `toHiveString` is for test only, we don't need to keep a `lazy val` for performance, we can just inline it.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/FileFormatWriter.scala
```diff
@@ -314,7 +315,7 @@ object FileFormatWriter extends Logging {
         val escaped = ScalaUDF(
           ExternalCatalogUtils.escapePathName _,
           StringType,
-          Seq(Cast(c, StringType)),
+          Seq(Cast(c, StringType, Option(DateTimeUtils.defaultTimeZone().getID))),
```
cloud-fan: shall we add a TODO here? We may need to send the sessionLocalTimeZone to executors and use it here.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/PartitioningAwareFileIndex.scala
```diff
@@ -137,7 +138,8 @@ abstract class PartitioningAwareFileIndex(
           InternalRow((0 until row.numFields).map { i =>
             Cast(
               Literal.create(row.getUTF8String(i), StringType),
-              userProvidedSchema.fields(i).dataType).eval()
+              userProvidedSchema.fields(i).dataType,
+              Option(DateTimeUtils.defaultTimeZone().getID)).eval()
```
cloud-fan: we can access `SparkSession` here, shall we use session local time zone?

sql/core/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala
```diff
@@ -655,6 +655,11 @@ object SQLConf {
       .booleanConf
       .createWithDefault(false)
 
+  val SESSION_LOCAL_TIMEZONE =
+    SQLConfigBuilder("spark.sql.session.timeZone")
+      .stringConf
```
cloud-fan: please add document by `.doc(...)`

sql/core/src/test/scala/org/apache/spark/sql/DataFrameSuite.scala
```diff
@@ -869,6 +870,30 @@ class DataFrameSuite extends QueryTest with SharedSQLContext {
...
+    val df = Seq((d, ts)).toDF("d", "ts")
+    val expectedAnswer = """+----------+-------------------+
+                           ||d         |ts                 |
+                           |+----------+-------------------+
+                           ||2016-12-01|2016-12-01 00:00:00|
+                           |+----------+-------------------+
+                           |""".stripMargin
+    assert(df.showString(1, truncate = 0) === expectedAnswer)
+
+    withSQLConf(SQLConf.SESSION_LOCAL_TIMEZONE.key -> "GMT") {
```
cloud-fan: we assume the test environment is in timezone `America/Los_Angeles` right?

sql/core/src/test/scala/org/apache/spark/sql/DateFunctionsSuite.scala
```diff
@@ -69,11 +70,26 @@ class DateFunctionsSuite extends QueryTest with SharedSQLContext {
     checkAnswer(
       df.select("t").filter($"t" <= "2014-06-01"),
       Row(Timestamp.valueOf("2014-01-01 00:00:00")) :: Nil)
+  }
 
+  test("timestamp comparison with date strings with session local timezone") {
```
cloud-fan: it seems we are using "timestamp string" in this test, shall we update the test name?

sql/core/src/test/scala/org/apache/spark/sql/DateFunctionsSuite.scala
```diff
@@ -91,6 +107,37 @@ class DateFunctionsSuite extends QueryTest with SharedSQLContext {
...
+      checkAnswer(
+        df.select("t").filter($"t" <= "2016-01-01"),
+        Seq(
+          Row(Date.valueOf("2015-12-31")),
+          Row(Date.valueOf("2016-01-01"))))
+
+      checkAnswer(
+        df.select("t").filter($"t" >= "2016"),
+        Seq(
+          Row(Date.valueOf("2016-01-01"))))
```
cloud-fan: nit: we can just write `Row(...)` here, `checkAnswer` has an overload version which takes a single `Row`

sql/core/src/test/scala/org/apache/spark/sql/DateFunctionsSuite.scala
```diff
@@ -103,6 +150,27 @@ class DateFunctionsSuite extends QueryTest with SharedSQLContext {
...
+      Row("2015-04-08 00:00:00", "2015-04-08 13:10:15", "2013-04-08 13:10:15"))
+
+    withSQLConf(SQLConf.SESSION_LOCAL_TIMEZONE.key -> "GMT") {
+
+      checkAnswer(
+        df.select(
+          date_format($"a", "yyyy-MM-dd HH:mm:ss"),
+          date_format($"b", "yyyy-MM-dd HH:mm:ss"),
+          date_format($"c", "yyyy-MM-dd HH:mm:ss")),
+        Row("2015-04-08 00:00:00", "2015-04-08 13:10:15", "2013-04-08 20:10:15"))
```
cloud-fan: why only the column `c` produce different result?

cloud-fan:
I went through all of the changes, LGTM overall, left some minor comments.

My main concern is, the newly added tests in `DateFunctionsSuite` are very hard to reason about the expected result, can you add some comments to explain them?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/DateTimeUtils.scala
ueshin: Yes, and the reason why I removed `threadLocalLocalTimeZone` is because it can't properly handle `TimeZone.setDefault()` by users.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/DateTimeUtils.scala
ueshin: This is also for handling `TimeZone.setDefault()` by users properly. If uses use it after the thread-local value was initialized, the date format will return invalid value.

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/CastSuite.scala
ueshin: It's defined in `org.apache.spark.sql.catalyst.util.DateTimeTestUtils`.

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/DateExpressionsSuite.scala
ueshin: Thanks, I'll revert it.

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/DateExpressionsSuite.scala
ueshin: Thanks, I'll revert it.

sql/core/src/main/scala/org/apache/spark/sql/Column.scala
ueshin: Thanks, I'll use it.

sql/core/src/main/scala/org/apache/spark/sql/execution/QueryExecution.scala
ueshin: Yes, that's right.

sql/core/src/main/scala/org/apache/spark/sql/execution/QueryExecution.scala
ueshin: I see, I'll inline it.

sql/core/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala
ueshin: Thanks, I'll add it.

sql/core/src/test/scala/org/apache/spark/sql/DataFrameSuite.scala
ueshin: Yes. All of `QueryTest`s use `America/Los_Angeles` as the default timezone.

sql/core/src/test/scala/org/apache/spark/sql/DateFunctionsSuite.scala
ueshin: Thanks, I'll update the name.

sql/core/src/test/scala/org/apache/spark/sql/DateFunctionsSuite.scala
ueshin: I see, I'll update to use the overload version.

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/CastSuite.scala
ueshin: I found some mistakes in this block, so I'll fix them and add some comments to describe the tests.

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/DateExpressionsSuite.scala
ueshin: I want to check that whether the `timeZoneId` exists or not doesn't affect the resolution of the expression.
BTW, now we should use `None` instead of `null`. I'll update them.

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/util/DateTimeUtilsSuite.scala
ueshin: I'll add some comments.

sql/core/src/main/scala/org/apache/spark/sql/Dataset.scala
ueshin: The previous result is generated by the `Date.toString` and `Timestamp.toString`, which use the timezone from `TimeZone.getDefault()`. We should use session local timezone here.

sql/core/src/main/scala/org/apache/spark/sql/Dataset.scala
ueshin: This will be used not once but num rows * num cells of TimestampType. I think we should leave it as is.

sql/core/src/main/scala/org/apache/spark/sql/execution/OptimizeMetadataOnlyQuery.scala
ueshin: This seems related to partition values so I'd leave it as is and add a TODO comment instead.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/FileFormatWriter.scala
ueshin: I see, I'll add a TODO comment.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/PartitioningAwareFileIndex.scala
ueshin: This seems also related to partition values so I'd leave it as is and add a TODO comment instead.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/DateTimeUtils.scala
ueshin: Yes, I think so except for JSON/CSV datasources or partition value related codes, which will be fixed by the follow-up prs.

ueshin:
I'll add some more comments to explain `DateFunctionsSuite` later.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/datetimeExpressions.scala
```diff
@@ -465,10 +536,17 @@ case class ToUnixTimestamp(timeExp: Expression, format: Expression) extends Unix
       > SELECT _FUNC_('2016-04-08', 'yyyy-MM-dd');
        1460041200
   """)
-case class UnixTimestamp(timeExp: Expression, format: Expression) extends UnixTime {
+case class UnixTimestamp(timeExp: Expression, format: Expression, timeZoneId: Option[String] = None)
```
cloud-fan: this class doesn't use time zone info

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/datetimeExpressions.scala
```diff
@@ -480,7 +558,8 @@ case class UnixTimestamp(timeExp: Expression, format: Expression) extends UnixTi
   override def prettyName: String = "unix_timestamp"
 }
 
-abstract class UnixTime extends BinaryExpression with ExpectsInputTypes {
+abstract class UnixTime
```
cloud-fan: this class doesn't use time zone info

sql/core/src/test/scala/org/apache/spark/sql/DateFunctionsSuite.scala
```diff
@@ -69,11 +70,28 @@ class DateFunctionsSuite extends QueryTest with SharedSQLContext {
...
+        Row(Timestamp.valueOf("2016-01-01 00:00:00"))))
+
+    withSQLConf(SQLConf.SESSION_LOCAL_TIMEZONE.key -> "GMT") {
+
+      // $"t" string in GMT would be as follows respectively:
+      // "2016-01-01 00:00:00"
+      // "2016-01-01 08:00:00"
+      checkAnswer(
+        df.select("t").filter($"t" <= "2016-01-01 00:00:00"),
+        Row(Timestamp.valueOf("2015-12-31 16:00:00")))
```
cloud-fan: this shows that it will be very confusing if the session local timezone is different from JVM default timezone in driver...

ueshin: Ah, I see, let me modify it.

sql/core/src/test/scala/org/apache/spark/sql/DateFunctionsSuite.scala
```diff
@@ -103,6 +153,51 @@ class DateFunctionsSuite extends QueryTest with SharedSQLContext {
...
+  test("date format with session local timezone") {
+    val df = Seq((d, sdf.format(d), ts)).toDF("a", "b", "c")
+
+    // The child of date_format is implicitly casted to TimestampType with session local timezone.
+    //
+    // +---+---------------------+-------------+---------------------+
+    // |   | df                  | timestamp   | date_format         |
+    // +---+---------------------+-------------+---------------------+
+    // | a |                16533|1428476400000|"2015-04-08 00:00:00"|
+    // | b |"2015-04-08 13:10:15"|1428523815000|"2015-04-08 13:10:15"|
```
cloud-fan: I'm a little confused, the `d` is already a `Date`, how can we have the time info back after converting the date to string?

sql/core/src/test/scala/org/apache/spark/sql/DateFunctionsSuite.scala
```diff
@@ -475,6 +1164,45 @@ class DateFunctionsSuite extends QueryTest with SharedSQLContext {
       Row(ts1.getTime / 1000L), Row(ts2.getTime / 1000L)))
   }
 
+  test("to_unix_timestamp with session local timezone") {
```
cloud-fan: These newly added tests are so similar that they all try to prove one thing: when you convert string or date to timestamp, the result changes according to session local timezone. When you convert timestamp to string or date, the result also changes with session local timezone. All time-related expressions should respect this.

shall we just write a general test for this? then we don't need so many similar tests

sql/core/src/test/scala/org/apache/spark/sql/DateFunctionsSuite.scala
ueshin: Do you mean you are wondering why `sdf.format(d)` has the time info `13:10:15` ?

If so, `java.sql.Date` DOES have the time info if it was initialized with the constructor `Date(long date)` or even if it was initalized with the constructor `Date(int year, int month, int day)` or with `Date.valueOf(String s)`, it has the time info `00:00:00` of the day in the timezone `TimeZone.getDefault()`.

```scala
scala> TimeZone.setDefault(TimeZone.getTimeZone("GMT"))

scala> val gmtDate = Date.valueOf("2017-01-24")
gmtDate: java.sql.Date = 2017-01-24

scala> val gmtTime = gmtDate.getTime
gmtTime: Long = 1485216000000

scala> TimeZone.setDefault(TimeZone.getTimeZone("PST"))

scala> val pstDate = Date.valueOf("2017-01-24")
pstDate: java.sql.Date = 2017-01-24

scala> val pstTime = pstDate.getTime
pstTime: Long = 1485244800000

scala> val sdf = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss")
sdf: java.text.SimpleDateFormat = java.text.SimpleDateFormat@4f76f1a0

scala> sdf.setTimeZone(TimeZone.getTimeZone("GMT"))

scala> sdf.format(gmtTime)
res12: String = 2017-01-24 00:00:00

scala> sdf.format(pstTime)
res13: String = 2017-01-24 08:00:00

scala> val d = new Date(sdf.parse("2015-04-08 13:10:15").getTime)
d: java.sql.Date = 2015-04-08

scala> sdf.format(d)
res14: String = 2015-04-08 13:10:15
```

sql/core/src/test/scala/org/apache/spark/sql/DateFunctionsSuite.scala
ueshin: I agree that there are so many similar tests, but I have no idea to generalize them.
Would you please give me some code snippets? I'll be able to expand them.

cloud-fan: I don't think we need to add tests in this file at all. We should improve `DateTimeUtilsSuite` to make sure the newly added methods work well with different timezones, e.g. `getHours`, `daysToMillions`, etc. Then make sure these timezone aware expressions will call the newly added methods in `DateTimeUtils` which has timezone parameter(we can remove the old versions that don't take timezone parameter, after we finish handling partition values).

This suite is end-to-end test, and it's very annoying if we wanna test all changed expressions, we should write more low-level tests in `DateTimeUtilsSuite`.

cloud-fan: The problem is, except this suite, all the changes you made to the tests are just fixed existing tests to fit the timezone stuff. You add all the new tests in this suite as end-to-end tests, which is not good. We should add new tests in `DateTimeUtilsSuite` as unit tests.

ueshin: Ah, I see! I'll move tests to `DateTimeUtilsSuite` soon. Thanks a lot!

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/CastSuite.scala
hvanhovell: Can we try to parameterize this a little bit more. I know you didn't write it, but it is quite hard to get through.

sql/core/src/main/scala/org/apache/spark/sql/Dataset.scala
hvanhovell: What happens if a user changes the session timezone? What would be the preferred behavior? Currently `show()` generates the same result every time, but that might be unexpected.

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/util/DateTimeUtilsSuite.scala
hvanhovell: Can you also try to parameterize these tests?

hvanhovell (changes requested):
The patch is in good shape. I left a few small comments, let's try to get this in ASAP.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/datetimeExpressions.scala
```diff
@@ -490,7 +569,11 @@ abstract class UnixTime extends BinaryExpression with ExpectsInputTypes {
 
   private lazy val constFormat: UTF8String = right.eval().asInstanceOf[UTF8String]
   private lazy val formatter: SimpleDateFormat =
-    Try(new SimpleDateFormat(constFormat.toString, Locale.US)).getOrElse(null)
+    Try {
```
ueshin: I see, I'll replace `Try`s with try-catch and add a method to create `SimpleDateFormat` with a format string and a timezone to `DateTimeUtils`.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/Optimizer.scala
```diff
@@ -1044,7 +1044,8 @@ object DecimalAggregates extends Rule[LogicalPlan] {
             we.copy(windowFunction = ae.copy(aggregateFunction = Average(UnscaledValue(e))))
           Cast(
             Divide(newAggExpr, Literal.create(math.pow(10.0, scale), DoubleType)),
-            DecimalType(prec + 4, scale + 4))
+            DecimalType(prec + 4, scale + 4),
+            Option(conf.sessionLocalTimeZone))
```
ueshin: I'll remove it.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/Optimizer.scala
```diff
@@ -1056,7 +1057,8 @@ object DecimalAggregates extends Rule[LogicalPlan] {
           val newAggExpr = ae.copy(aggregateFunction = Average(UnscaledValue(e)))
           Cast(
             Divide(newAggExpr, Literal.create(math.pow(10.0, scale), DoubleType)),
-            DecimalType(prec + 4, scale + 4))
+            DecimalType(prec + 4, scale + 4),
+            Option(conf.sessionLocalTimeZone))
```
ueshin: I'll remove it.

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/CastSuite.scala
ueshin: I see, I'll try it.

sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/util/DateTimeUtilsSuite.scala
ueshin: I see, I'll try it.

sql/core/src/main/scala/org/apache/spark/sql/Dataset.scala
ueshin: I think the preferred behavior would be that the result of `show()` should be changed by the session timezone.

ueshin:
Jenkins, retest this please.

hvanhovell:
LGTM - pending jenkins.

hvanhovell:
Merging to master. Thanks for the hard work, and the patience with the review process.

Can you open follow-up PRs for any remaining issues?

ueshin:
Sure! I'll send follow-up prs as soon as possible.
Thanks a lot for your review!

grvishwanath:
@ueshin Can you please explain to me if this functionality is the same as blessing Spark with "Timestamp with Time Zone"? If now, how is it different?

PR #16310, opened 2016-12-16
[SPARK-18897][SPARKR] Fix SparkR SQL Test to drop test table

dongjoon-hyun:

SparkR tests, `R/run-tests.sh`, succeeds only once because `test_sparkSQL.R` does not clean up the test table, `people`.

As a result, the rows in `people` table are accumulated at every run and the test cases fail.

The following is the failure result for the second run.

```r
1. Failure: create DataFrame from RDD (@test_sparkSQL.R#204) -------------------
collect(sql("SELECT age from people WHERE name = 'Bob'"))$age not equal to c(16).
Lengths differ: 2 vs 1

2. Failure: create DataFrame from RDD (@test_sparkSQL.R#206) -------------------
collect(sql("SELECT height from people WHERE name ='Bob'"))$height not equal to c(176.5).
Lengths differ: 2 vs 1
```

## How was this patch tested?

Manual. Run `run-tests.sh` twice and check if it passes without failures.

### DIFF
--- a/R/pkg/inst/tests/testthat/test_sparkSQL.R
+++ b/R/pkg/inst/tests/testthat/test_sparkSQL.R
@@ -205,6 +205,7 @@ test_that("create DataFrame from RDD", {
                c(16))
   expect_equal(collect(sql("SELECT height from people WHERE name ='Bob'"))$height,
                c(176.5))
+  sql("DROP TABLE people")
   unsetHiveContext()
 })
 


dongjoon-hyun:
@shivaram and @felixcheung 
Could you review this? I found this while testing RC5.

shivaram:
LGTM. Yeah I ran into this while testing another PR as well. Thanks for the fix

felixcheung:
good catch! thx!

dongjoon-hyun:
Thank you for review, @shivaram and @felixcheung !

shivaram:
Merging to master, branch-2.1 and branch-2.0

dongjoon-hyun:
Thank you for merging!

PR #16312, opened 2016-12-16
[SPARK-18862][SPARKR][ML] Split SparkR mllib.R into multiple files

yanboliang:
SparkR ```mllib.R``` is getting bigger as we add more ML wrappers, I'd like to split it into multiple files to make us easy to maintain:
* mllib_classification.R
* mllib_clustering.R
* mllib_recommendation.R
* mllib_regression.R
* mllib_stat.R
* mllib_tree.R
* mllib_utils.R

Note: Only reorg, no actual code change.

## How was this patch tested?
Existing tests.


wangmiao1981:
This is quite necessary change! The file becomes very lengthy. Thanks!

wangmiao1981:
retest this please.

felixcheung:
I like how they are grouped. not sure why tests are failing though

felixcheung:
Hmm, from what I can see pretty sure it used to work with sparkR.session.stop() calls for enableHiveSupport = F sessions. Do we know why this is suddenly causing issues?

yanboliang:
I'm trying to figure out the cause of this issue but still no progress until now, could you or @shivaram give some hints or suggestion? Thanks.

felixcheung:
sure, I'll take a deeper look in a day or 2.

shivaram:
I looked at this more closely and I think I found the problem - Not sure its easy to fix though.
What I traced here is:
- When we call sparkR.session.stop and sparkR.session the same JVM backend is reused and only the SparkContext is stopped / recreated
- Now the problem happens when we call `read.ml` to read a model after creating a new SparkSession. This in turn calls into RWrappers[1] which has an `sc` member variable
- My understanding is that the `sc` member variable is bound the first time we create a SparkSession and when we stop, restart it has a handle to the stale SparkContext
- Thus we see errors where it says `Cannot call methods on a stopped SparkContext`

I think the right fix here is to pass along a SparkContext into RWrappers and not rely on a prior initialization. However I'm not sure why that design decision was made before, so maybe I'm missing something.

[1] https://github.com/apache/spark/blob/f252cb5d161e064d39cc1ed1d9299307a0636174/mllib/src/main/scala/org/apache/spark/ml/r/RWrappers.scala#L36

felixcheung:
ah, thank you @shivaram. sorry I couldn't get around to investigate earlier.

@yanboliang It looks like that is the design in the trait BaseReadWrite ([here](https://github.com/apache/spark/blob/master/mllib/src/main/scala/org/apache/spark/ml/util/ReadWrite.scala#L80)), where it holds references to `sc`, `sqlContext` and `spark` session. Although, I see other MLReader/MLWriter calls `sc` directly whereas the design should allow for the `sc`/spark session to be updated? Specifically we could change to pass the spark session to the RWrapper for these calls but generally reusing `sc` is the design of BaseReadWrite/MLReader/MLWriter, and is not specific to R.

yanboliang:
@shivaram @felixcheung You comments are really very helpful, I understand the problem and will try to figure out some ways to fix it. Thanks very much.

yanboliang:
@felixcheung @shivaram I changed ```read.ml``` to pass in ```sparkSession``` if necessary, then it will not use stale spark context which will cause errors. This is consistent with ```MLReader/MLWriter``` in Scala/Python which also support to specify user-specified spark session. Any more comments, please feel free to let me know. Thanks.

R/pkg/R/mllib_utils.R
```diff
@@ -80,9 +81,9 @@ predict_internal <- function(object, newData) {
 #' model <- read.ml(path)
 #' }
 #' @note read.ml since 2.0.0
-read.ml <- function(path) {
+read.ml <- function(path, sparkSession = NULL) {
```
felixcheung: we should get this from the current default session instead of making it a parameter, like here
https://github.com/apache/spark/blob/master/R/pkg/R/SQLContext.R#L136

yanboliang: Sounds good, updated.

R/pkg/R/mllib_utils.R
```diff
@@ -81,9 +80,11 @@ predict_internal <- function(object, newData) {
 #' model <- read.ml(path)
 #' }
 #' @note read.ml since 2.0.0
-read.ml <- function(path, sparkSession = NULL) {
+read.ml <- function(path) {
   path <- suppressWarnings(normalizePath(path))
-  jobj <- callJStatic("org.apache.spark.ml.r.RWrappers", "load", path, sparkSession)
+  sparkSession <- getSparkSession()
+  callJStatic("org.apache.spark.ml.r.RWrappers", "session", sparkSession)
```
felixcheung: I'm a bit confused by this - what does this do?

felixcheung:
LGTM. thanks!

R/pkg/R/mllib_utils.R
yanboliang: Here we call ```session``` method to set spark session, the function name is really confusing, may be named as ```setSession``` should be better, but we use the current name from the very beginning.
There is no other places need to be updated, since here is the only place calling ```RWrappers```. Thanks.

yanboliang:
Merged into master, thanks for reviewing.

PR #16313, opened 2016-12-16
[SPARK-18899][SPARK-18912][SPARK-18913][SQL] refactor the error checking when append data to an existing table

cloud-fan:

When we append data to an existing table with `DataFrameWriter.saveAsTable`, we will do various checks to make sure the appended data is consistent with the existing data.

However, we get the information of the existing table by matching the table relation, instead of looking at the table metadata. This is error-prone, e.g. we only check the number of columns for `HadoopFsRelation`, we forget to check bucketing, etc.

This PR refactors the error checking by looking at the metadata of the existing table, and fix several bugs:
* SPARK-18899: We forget to check if the specified bucketing matched the existing table, which may lead to a problematic table that has different bucketing in different data files.
* SPARK-18912: We forget to check the number of columns for non-file-based data source table
* SPARK-18913: We don't support append data to a table with special column names.

## How was this patch tested?
new regression test.

### DIFF
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/interface.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/interface.scala
@@ -133,6 +133,16 @@ case class BucketSpec(
   if (numBuckets <= 0) {
     throw new AnalysisException(s"Expected positive number of buckets, but got `$numBuckets`.")
   }
+
+  override def toString: String = {
+    val bucketString = s"bucket columns: [${bucketColumnNames.mkString(", ")}]"
+    val sortString = if (sortColumnNames.nonEmpty) {
+      s", sort columns: [${sortColumnNames.mkString(", ")}]"
+    } else {
+      ""
+    }
+    s"$numBuckets buckets, $bucketString$sortString"
+  }
 }
 
 /**
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/command/createDataSourceTables.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/command/createDataSourceTables.scala
@@ -157,9 +157,6 @@ case class CreateDataSourceTableAsSelectCommand(
           return Seq.empty[Row]
         case SaveMode.Append =>
           val existingTable = sessionState.catalog.getTableMetadata(tableIdentWithDB)
-          if (existingTable.tableType == CatalogTableType.VIEW) {
-            throw new AnalysisException("Saving data into a view is not allowed.")
-          }
 
           if (existingTable.provider.get == DDLUtils.HIVE_PROVIDER) {
             throw new AnalysisException(s"Saving data in the Hive serde table $tableName is " +
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/command/createDataSourceTables.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/command/createDataSourceTables.scala
@@ -160,7 +160,7 @@ case class CreateDataSourceTableAsSelectCommand(
 
           if (existingTable.provider.get == DDLUtils.HIVE_PROVIDER) {
             throw new AnalysisException(s"Saving data in the Hive serde table $tableName is " +
-              s"not supported yet. Please use the insertInto() API as an alternative.")
+              "not supported yet. Please use the insertInto() API as an alternative.")
           }
 
           // Check if the specified data source match the data source of the existing table.
@@ -192,13 +192,18 @@ case class CreateDataSourceTableAsSelectCommand(
             }
           })
 
+          // In `AnalyzeCreateTable`, we verified the consistency between the user-specified table
+          // definition(partition columns, bucketing) and the SELECT query, here we also need to
+          // verify the the consistency between the user-specified table definition and the existing
+          // table definition.
+
           // Check if the specified partition columns match the existing table.
           val specifiedPartCols = CatalogUtils.normalizePartCols(
             tableName, tableCols, table.partitionColumnNames, resolver)
           if (specifiedPartCols != existingTable.partitionColumnNames) {
             throw new AnalysisException(
               s"""
-                |Specified partitioning does not match the existing table $tableName.
+                |Specified partitioning does not match that of the existing table $tableName.
                 |Specified partition columns: [${specifiedPartCols.mkString(", ")}]
                 |Existing partition columns: [${existingTable.partitionColumnNames.mkString(", ")}]
               """.stripMargin)
@@ -215,7 +220,7 @@ case class CreateDataSourceTableAsSelectCommand(
               existingTable.bucketSpec.map(_.toString).getOrElse("not bucketed")
             throw new AnalysisException(
               s"""
-                |Specified bucketing does not match the existing table $tableName.
+                |Specified bucketing does not match that of the existing table $tableName.
                 |Specified bucketing: $specifiedBucketString
                 |Existing bucketing: $existingBucketString
               """.stripMargin)
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/command/createDataSourceTables.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/command/createDataSourceTables.scala
@@ -18,13 +18,11 @@
 package org.apache.spark.sql.execution.command
 
 import org.apache.spark.sql._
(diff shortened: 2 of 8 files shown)


cloud-fan:
cc @yhuai @gatorsmile

sql/core/src/main/scala/org/apache/spark/sql/execution/command/createDataSourceTables.scala
```diff
@@ -157,39 +156,74 @@ case class CreateDataSourceTableAsSelectCommand(
           // Since the table already exists and the save mode is Ignore, we will just return.
           return Seq.empty[Row]
         case SaveMode.Append =>
+          val existingTable = sessionState.catalog.getTableMetadata(tableIdentWithDB)
+          if (existingTable.tableType == CatalogTableType.VIEW) {
+            throw new AnalysisException("Saving data into a view is not allowed.")
```
gatorsmile: We already have [an assert](https://github.com/cloud-fan/spark/blob/b1dbd0a19a174eaae1aaf114e04f6d3683ea65c4/sql/core/src/main/scala/org/apache/spark/sql/execution/command/createDataSourceTables.scala#L133) at the beginning of this function. We need to improve the error message there. That can cover more cases.

So far, the error message is pretty confusing.
```
assertion failed
java.lang.AssertionError: assertion failed
	at scala.Predef$.assert(Predef.scala:156)
```

We can add a test case for it.
```Scala
        val df = spark.range(1, 10).toDF("id1")
        df.write.saveAsTable("tab1")
        spark.sql("create view view1 as select * from tab1")
        df.write.mode(SaveMode.Append).format("parquet").saveAsTable("view1")
```

sql/core/src/main/scala/org/apache/spark/sql/execution/command/createDataSourceTables.scala
```diff
@@ -157,39 +156,74 @@ case class CreateDataSourceTableAsSelectCommand(
           // Since the table already exists and the save mode is Ignore, we will just return.
           return Seq.empty[Row]
         case SaveMode.Append =>
+          val existingTable = sessionState.catalog.getTableMetadata(tableIdentWithDB)
```
gatorsmile: Is that possible we can directly use the input parameter [table](https://github.com/cloud-fan/spark/blob/b1dbd0a19a174eaae1aaf114e04f6d3683ea65c4/sql/core/src/main/scala/org/apache/spark/sql/execution/command/createDataSourceTables.scala#L125)?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/interface.scala
```diff
@@ -133,6 +133,16 @@ case class BucketSpec(
   if (numBuckets <= 0) {
     throw new AnalysisException(s"Expected positive number of buckets, but got `$numBuckets`.")
   }
+
+  override def toString: String = {
```
gatorsmile: Since we implement `toString` here, we can simplify our logics in [describeBucketingInfo](https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/command/tables.scala#L523-L526)

sql/hive/src/test/scala/org/apache/spark/sql/sources/HadoopFsRelationTest.scala
```diff
@@ -459,7 +459,7 @@ abstract class HadoopFsRelationTest extends QueryTest with SQLTestUtils with Tes
   test("saveAsTable()/load() - partitioned table - ErrorIfExists") {
     Seq.empty[(Int, String)].toDF().createOrReplaceTempView("t")
 
-    withTempView("t") {
```
gatorsmile: We need `withTempView("t")` to drop the temp view that is created at the beginning of this test case?

cloud-fan: yea, my mistake

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/interface.scala
gatorsmile: : ) If we want to keep the existing format (3 lines), then we are unable to do it.

sql/core/src/main/scala/org/apache/spark/sql/execution/command/createDataSourceTables.scala
gatorsmile: uh, I see. Will do the final review today.

sql/core/src/main/scala/org/apache/spark/sql/execution/command/createDataSourceTables.scala
```diff
@@ -157,39 +156,71 @@ case class CreateDataSourceTableAsSelectCommand(
...
+            }
+          })
+
+          // Check if the specified partition columns match the existing table.
+          val specifiedPartCols = CatalogUtils.normalizePartCols(
+            tableName, tableCols, table.partitionColumnNames, resolver)
+          if (specifiedPartCols != existingTable.partitionColumnNames) {
+            throw new AnalysisException(
+              s"""
+                |Specified partitioning does not match the existing table $tableName.
```
gatorsmile: Nit: the grammar issue.
`Specified partitioning does not match the existing table $tableName.`
->
`Specified partitioning does not match that of the existing table $tableName.`

Found a [reference link:](http://infocenter.sybase.com/help/index.jsp?topic=/com.sybase.infocenter.dc00743.1570/html/queryprocessing/queryprocessing137.htm) in sybase adaptive server.

sql/core/src/main/scala/org/apache/spark/sql/execution/command/createDataSourceTables.scala
```diff
@@ -157,39 +156,71 @@ case class CreateDataSourceTableAsSelectCommand(
...
+            CatalogUtils.normalizeBucketSpec(tableName, tableCols, bucketSpec, resolver)
+          }
+          if (specifiedBucketSpec != existingTable.bucketSpec) {
+            val specifiedBucketString =
+              specifiedBucketSpec.map(_.toString).getOrElse("not bucketed")
+            val existingBucketString =
+              existingTable.bucketSpec.map(_.toString).getOrElse("not bucketed")
+            throw new AnalysisException(
+              s"""
+                |Specified bucketing does not match the existing table $tableName.
```
gatorsmile: Nit: The same here.

sql/core/src/main/scala/org/apache/spark/sql/execution/command/createDataSourceTables.scala
```diff
@@ -157,39 +156,71 @@ case class CreateDataSourceTableAsSelectCommand(
           // Since the table already exists and the save mode is Ignore, we will just return.
           return Seq.empty[Row]
         case SaveMode.Append =>
+          val existingTable = sessionState.catalog.getTableMetadata(tableIdentWithDB)
+
+          if (existingTable.provider.get == DDLUtils.HIVE_PROVIDER) {
+            throw new AnalysisException(s"Saving data in the Hive serde table $tableName is " +
+              s"not supported yet. Please use the insertInto() API as an alternative.")
```
gatorsmile: Nit: the string interpolation`s` is not needed.

sql/core/src/main/scala/org/apache/spark/sql/execution/command/createDataSourceTables.scala
```diff
@@ -157,39 +156,71 @@ case class CreateDataSourceTableAsSelectCommand(
...
+          }
+
           // Check if the specified data source match the data source of the existing table.
```
gatorsmile: Now, the checking logics are split two places for CTAS of data source tables. Maybe we can improve the comment to explain `AnalyzeCreateTable` verifies the consistency between the user-specified table schema/definition and the SELECT query. Here, we verifies the consistency between the user-specified table schema/definition and the existing table schema/definition, the consistency between the existing table schema/definition and the SELECT query.

gatorsmile:
LGTM except a few minor comments.

gatorsmile:
LGTM

gatorsmile:
retest this please

gatorsmile:
Thanks! Merging to master/2.1.

cloud-fan:
Actually this PR was not backported to 2.1, now I've backported.

PR #16316, opened 2016-12-16
Branch 2.0

kubatyszko:

This problem is very similar to the one discussed here: https://github.com/databricks/spark-csv/issues/239

## How was this patch tested?


srowen:
Close this @kubatyszko

kubatyszko:
@srowen why?

srowen:
It looks like an error -- see the title

vanzin:
It looks like the change might actually be valid, but check the document linked when you submitted the PR for how to open up a PR.

PR #16317, opened 2016-12-16
returns null when field is empty string on integer, long, and short c…

kubatyszko:

## How was this patch tested?

Credit - Sargis Dudaklayan and Kuba Tyszko - Zest Finance


srowen:
This is not opened vs master -- please close this

PR #16319, opened 2016-12-16
[WiP][SPARK-18699] SQL - parsing CSV should return null for certain numeric …

kubatyszko:

This problem is very similar to the one discussed here: https://github.com/databricks/spark-csv/issues/239

Sargis Dudaklayan and Kuba Tyszko - Zest Finance

## How was this patch tested?

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/csv/CSVInferSchema.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/csv/CSVInferSchema.scala
@@ -246,9 +246,9 @@ private[csv] object CSVTypeCast {
     } else {
       castType match {
         case _: ByteType => datum.toByte
-        case _: ShortType => datum.toShort
-        case _: IntegerType => datum.toInt
-        case _: LongType => datum.toLong
+        case _: ShortType => Try(datum.toShort).getOrElse(null)
+        case _: IntegerType => Try(datum.toInt).getOrElse(null)
+        case _: LongType => Try(datum.toLong).getOrElse(null)
         case _: FloatType =>
           datum match {
             case options.nanValue => Float.NaN


HyukjinKwon:
Can you describe which case this PR fixes? I think you meant https://issues.apache.org/jira/browse/SPARK-18699

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/csv/CSVInferSchema.scala
```diff
@@ -246,9 +246,9 @@ private[csv] object CSVTypeCast {
     } else {
       castType match {
         case _: ByteType => datum.toByte
-        case _: ShortType => datum.toShort
-        case _: IntegerType => datum.toInt
-        case _: LongType => datum.toLong
+        case _: ShortType => Try(datum.toShort).getOrElse(null)
+        case _: IntegerType => Try(datum.toInt).getOrElse(null)
+        case _: LongType => Try(datum.toLong).getOrElse(null)
```
kubatyszko: You're right, fixed the title. Thanks.

HyukjinKwon:
Let's put `[WIP]` in the PR title, for example, (`[WIP][SPARK-18699] ... `) if you are working on this to follow the suggestion in the JIRA. It seems we need to update the PR description (and with a proper PR title without truncated part such as `...`) and need a test in `CSVSuite`.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/csv/CSVInferSchema.scala
maropu: Why do you use `Try` for each type here? In the JIRA, I proposed another approach (https://github.com/apache/spark/compare/master...maropu:SPARK-18699), and I think the fix is more natural. Anyway, ISTM we do not decide yet it is worth fixing this issue...

maropu:
@HyukjinKwon Thanks for your pinging! I left some comments.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/csv/CSVInferSchema.scala
maropu: Behavior changes are much arguable, so it basically takes much time to discuss.

HyukjinKwon:
Hi @kubatyszko, are you still working on this? If you are currently unable to proceed further, maybe it should be closed for now. It seems inactive for few months.

PR #16320, opened 2016-12-16
[SPARK-18877][SQL] `CSVInferSchema.inferField` on DecimalType should find a common type with `typeSoFar`

dongjoon-hyun:

CSV type inferencing causes `IllegalArgumentException` on decimal numbers with heterogeneous precisions and scales because the current logic uses the last decimal type in a **partition**. Specifically, `inferRowType`, the **seqOp** of **aggregate**, returns the last decimal type. This PR fixes it to use `findTightestCommonType`.

**decimal.csv**
```
9.03E+12
1.19E+11
```

**BEFORE**
```scala
scala> spark.read.format("csv").option("inferSchema", true).load("decimal.csv").printSchema
root
 |-- _c0: decimal(3,-9) (nullable = true)

scala> spark.read.format("csv").option("inferSchema", true).load("decimal.csv").show
16/12/16 14:32:49 ERROR Executor: Exception in task 0.0 in stage 4.0 (TID 4)
java.lang.IllegalArgumentException: requirement failed: Decimal precision 4 exceeds max precision 3
```

**AFTER**
```scala
scala> spark.read.format("csv").option("inferSchema", true).load("decimal.csv").printSchema
root
 |-- _c0: decimal(4,-9) (nullable = true)

scala> spark.read.format("csv").option("inferSchema", true).load("decimal.csv").show
+---------+
|      _c0|
+---------+
|9.030E+12|
| 1.19E+11|
+---------+
```

## How was this patch tested?

Pass the newly add test case.

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/csv/CSVInferSchema.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/csv/CSVInferSchema.scala
@@ -87,7 +87,7 @@ private[csv] object CSVInferSchema {
         case LongType => tryParseLong(field, options)
         case _: DecimalType =>
           // DecimalTypes have different precisions and scales, so we try to find the common type.
-          findTightestCommonType(typeSoFar, tryParseDecimal(field, options)).getOrElse(NullType)
+          findTightestCommonType(typeSoFar, tryParseDecimal(field, options)).getOrElse(StringType)
         case DoubleType => tryParseDouble(field, options)
         case TimestampType => tryParseTimestamp(field, options)
         case BooleanType => tryParseBoolean(field, options)
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/csv/CSVInferSchema.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/csv/CSVInferSchema.scala
@@ -85,7 +85,9 @@ private[csv] object CSVInferSchema {
         case NullType => tryParseInteger(field, options)
         case IntegerType => tryParseInteger(field, options)
         case LongType => tryParseLong(field, options)
-        case _: DecimalType => tryParseDecimal(field, options)
+        case _: DecimalType =>
+          // DecimalTypes have different precisions and scales, so we try to find the common type.
+          findTightestCommonType(typeSoFar, tryParseDecimal(field, options)).getOrElse(NullType)
         case DoubleType => tryParseDouble(field, options)
         case TimestampType => tryParseTimestamp(field, options)
         case BooleanType => tryParseBoolean(field, options)
--- a/sql/core/src/test/scala/org/apache/spark/sql/execution/datasources/csv/CSVInferSchemaSuite.scala
+++ b/sql/core/src/test/scala/org/apache/spark/sql/execution/datasources/csv/CSVInferSchemaSuite.scala
@@ -114,4 +114,11 @@ class CSVInferSchemaSuite extends SparkFunSuite {
     val options = new CSVOptions(Map("TiMeStampFormat" -> "yyyy-mm"))
     assert(CSVInferSchema.inferField(TimestampType, "2015-08", options) == TimestampType)
   }
+
+  test("SPARK-18877: `inferField` on DecimalType should find a common type with `typeSoFar`") {
+    val options = new CSVOptions(Map.empty[String, String])
+    // 9.03E+12 is Decimal(3, -10) and 1.19E+11 is Decimal(3, -9).
+    assert(CSVInferSchema.inferField(DecimalType(3, -10), "1.19E+11", options) ==
+      DecimalType(4, -9))
+  }
 }
--- a/sql/core/src/test/scala/org/apache/spark/sql/execution/datasources/csv/CSVInferSchemaSuite.scala
+++ b/sql/core/src/test/scala/org/apache/spark/sql/execution/datasources/csv/CSVInferSchemaSuite.scala
@@ -125,5 +125,10 @@ class CSVInferSchemaSuite extends SparkFunSuite {
     // BigDecimal("12345678901234567890.01234567890123456789") is precision 40 and scale 20.
     val value = "12345678901234567890.01234567890123456789"
     assert(CSVInferSchema.inferField(DecimalType(3, -10), value, options) == DoubleType)
+
+    // Seq(s"${Long.MaxValue}1", "2015-12-01 00:00:00") should be StringType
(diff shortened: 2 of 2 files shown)


dongjoon-hyun:
Hi, @rxin , @falaki , and @HyukjinKwon .
Could you review this PR?

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/csv/CSVInferSchema.scala
```diff
@@ -85,7 +85,9 @@ private[csv] object CSVInferSchema {
         case NullType => tryParseInteger(field, options)
         case IntegerType => tryParseInteger(field, options)
         case LongType => tryParseLong(field, options)
-        case _: DecimalType => tryParseDecimal(field, options)
+        case _: DecimalType =>
+          // DecimalTypes have different precisions and scales, so we try to find the common type.
+          findTightestCommonType(typeSoFar, tryParseDecimal(field, options)).getOrElse(NullType)
```
HyukjinKwon: Aha, so, `typeSoFar` should keep the precision and scale while being (partially) aggregated within each partition.

HyukjinKwon:
@dongjoon-hyun would we need a end-to-end test too?

dongjoon-hyun:
Actually, I made the end-to-end test based on the example of the use case in JIRA first. And, I removed that from here because the current test case is the minimal version of that.

dongjoon-hyun:
Hi, @gatorsmile .
Could you review this PR?

dongjoon-hyun:
Hi, @rxin and @falaki .
If there are some committers to review this part, I think you are the best persons.
Could you give some opinion about this when you have some time?

dongjoon-hyun:
Hi, @gatorsmile .
Could you review this PR when you have some time?

sql/core/src/test/scala/org/apache/spark/sql/execution/datasources/csv/CSVInferSchemaSuite.scala
```diff
@@ -114,4 +114,11 @@ class CSVInferSchemaSuite extends SparkFunSuite {
     val options = new CSVOptions(Map("TiMeStampFormat" -> "yyyy-mm"))
     assert(CSVInferSchema.inferField(TimestampType, "2015-08", options) == TimestampType)
   }
+
+  test("SPARK-18877: `inferField` on DecimalType should find a common type with `typeSoFar`") {
+    val options = new CSVOptions(Map.empty[String, String])
+    // 9.03E+12 is Decimal(3, -10) and 1.19E+11 is Decimal(3, -9).
+    assert(CSVInferSchema.inferField(DecimalType(3, -10), "1.19E+11", options) ==
+      DecimalType(4, -9))
```
gatorsmile: Could you add another test case here using the input constant with more than 38 precision?

dongjoon-hyun: Oh, I missed your comment here. Let me try!

dongjoon-hyun:
Thank you for review, @gatorsmile .
I added another test case for the input constant with more than 38 precision.

dongjoon-hyun:
Could you review this `CSVInferSchema` issue again, @gatorsmile ?

gatorsmile:
LGTM cc @cloud-fan

dongjoon-hyun:
Thank you for review, @gatorsmile .
Happy new year!

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/csv/CSVInferSchema.scala
cloud-fan: Looks like the fallback policy here is to use `StringType`, shoud we follow?

dongjoon-hyun: Thank you for review, @cloud-fan .  I used `NullType` since `mergeRowTypes` does.

```scala
  def mergeRowTypes(first: Array[DataType], second: Array[DataType]): Array[DataType] = {
    first.zipAll(second, NullType, NullType).map { case (a, b) =>
      findTightestCommonType(a, b).getOrElse(NullType)
    }
  }
```

HyukjinKwon: Yes, otherwise, it might end up with an incorrect datatypes. For example,

```scala
val path = "/tmp/test1"
Seq(s"${Long.MaxValue}1", "2015-12-01 00:00:00", "1").toDF().coalesce(1).write.text(path)
spark.read.option("inferSchema", true).csv(path).printSchema()
```

```
root
 |-- _c0: integer (nullable = true)
```

dongjoon-hyun: You're correct. I'll change into `StringType`.

dongjoon-hyun:
Thank you again, @cloud-fan and @HyukjinKwon . I updated the fallback datatype.

gatorsmile:
Please add the test case?

dongjoon-hyun:
I assumed this one. Right?
```scala
val path = "/tmp/test1"
Seq(s"${Long.MaxValue}1", "2015-12-01 00:00:00", "1").toDF().coalesce(1).write.text(path)
spark.read.option("inferSchema", true).csv(path).printSchema()
```

dongjoon-hyun:
Yep. I added the testcase, too. @gatorsmile

gatorsmile:
The test case coverage in the suite `CSVInferSchemaSuite.scala` looks random. I am afraid the future code changes could easily break the existing type inference rules. Could you improve it in a separate PR?

dongjoon-hyun:
Retest this please

dongjoon-hyun:
I see, @gatorsmile . I will try to make a PR to improve the coverage.

For this issue, the failure about the last commit (adding test case) was a R failure. So, it's irrelevant. I rerun it.

cloud-fan:
thanks, merging to master!

cloud-fan:
oh forget to backport it to 2.1/2.0, @gatorsmile can you do it? My connection is bad now.

dongjoon-hyun:
Thank you, @cloud-fan !

gatorsmile:
@dongjoon-hyun Could you submit a backport PR to 2.1? I am unable to merge this PR to 2.1. Thanks!

dongjoon-hyun:
Sure. I'll create a backpor PR for 2.1.

PR #16322, opened 2016-12-17
[SPARK-18908][SS] Creating StreamingQueryException should check if logicalPlan is created

zsxwing:

This PR audits places using `logicalPlan` in StreamExecution and ensures they all handles the case that `logicalPlan` cannot be created.

In addition, this PR also fixes the following issues in `StreamingQueryException`:
- `StreamingQueryException` and `StreamExecution` are cycle-dependent because in the `StreamingQueryException`'s constructor, it calls `StreamExecution`'s `toDebugString` which uses `StreamingQueryException`. Hence it will output `null` value in the error message.
- Duplicated stack trace when calling Throwable.printStackTrace because StreamingQueryException's toString contains the stack trace.

## How was this patch tested?

The updated `test("max files per trigger - incorrect values")`. I found this issue when I switched from `testStream` to the real codes to verify the failure in this test.

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamExecution.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamExecution.scala
@@ -118,13 +118,6 @@ class StreamExecution(
     Option(name).map(_ + " ").getOrElse("") + s"[id = $id, runId = $runId]"
 
   /**
-   * The lock is used to generate `_logicalPlan` lazily.
-   */
-  private val localPlanLock = new Object
-
-  @volatile private var _logicalPlan: LogicalPlan = null
-
-  /**
    * All stream sources present in the query plan. This will be set when generating logical plan.
    */
   @volatile protected var sources: Seq[Source] = Seq.empty
@@ -133,6 +126,23 @@ class StreamExecution(
    * A list of unique sources in the query plan. This will be set when generating logical plan.
    */
   @volatile private var uniqueSources: Seq[Source] = Seq.empty
+
+  override lazy val logicalPlan: LogicalPlan = {
+    var nextSourceId = 0L
+    val _logicalPlan = analyzedPlan.transform {
+      case StreamingRelation(dataSource, _, output) =>
+        // Materialize source to avoid creating it in every batch
+        val metadataPath = s"$checkpointRoot/sources/$nextSourceId"
+        val source = dataSource.createSource(metadataPath)
+        nextSourceId += 1
+        // We still need to use the previous `output` instead of `source.schema` as attributes in
+        // "df.logicalPlan" has already used attributes of the previous `output`.
+        StreamingExecutionRelation(source, output)
+    }
+    sources = _logicalPlan.collect { case s: StreamingExecutionRelation => s.source }
+    uniqueSources = sources.distinct
+    _logicalPlan
+  }
 
   private val triggerExecutor = trigger match {
     case t: ProcessingTime => ProcessingTimeExecutor(t, triggerClock)
@@ -201,36 +211,6 @@ class StreamExecution(
     startLatch.await()  // Wait until thread started and QueryStart event has been posted
   }
 
-  private def generateLogicalPlan: LogicalPlan = {
-    var nextSourceId = 0L
-    val internalLogicalPlan = analyzedPlan.transform {
-      case StreamingRelation(dataSource, _, output) =>
-        // Materialize source to avoid creating it in every batch
-        val metadataPath = s"$checkpointRoot/sources/$nextSourceId"
-        val source = dataSource.createSource(metadataPath)
-        nextSourceId += 1
-        // We still need to use the previous `output` instead of `source.schema` as attributes in
-        // "df.logicalPlan" has already used attributes of the previous `output`.
-        StreamingExecutionRelation(source, output)
-    }
-    sources = internalLogicalPlan.collect { case s: StreamingExecutionRelation => s.source }
-    uniqueSources = sources.distinct
-    internalLogicalPlan
-  }
-
-  override def logicalPlan: LogicalPlan = {
-    if (_logicalPlan == null) {
-      localPlanLock.synchronized {
-        if (_logicalPlan == null) {
-          _logicalPlan = generateLogicalPlan
-        }
-      }
-    }
-    _logicalPlan
-  }
-
-  private def isLogicalPlanGenerated: Boolean = _logicalPlan != null
-
   /**
    * Repeatedly attempts to run batches as data arrives.
    *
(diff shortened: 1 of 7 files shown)


sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamExecution.scala
```diff
@@ -560,6 +589,9 @@ class StreamExecution(
   @volatile private var noNewData = false
 
   override def processAllAvailable(): Unit = {
+    if (streamDeathCause != null) {
```
zsxwing: Just make `processAllAvailable` fail fast.

sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamTest.scala
```diff
@@ -240,7 +241,7 @@ trait StreamTest extends QueryTest with SharedSQLContext with Timeouts {
     val resetConfValues = mutable.Map[String, Option[String]]()
 
     @volatile
-    var streamDeathCause: Throwable = null
+    var streamThreadDeathCause: Throwable = null
```
zsxwing: Renamed it to avoid confusing with `StreamExecution.streamDeathCause`

zsxwing:
cc @tdas @brkyvz @marmbrus

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamExecution.scala
```diff
@@ -206,6 +201,36 @@ class StreamExecution(
...
+  override def logicalPlan: LogicalPlan = {
+    if (_logicalPlan == null) {
+      localPlanLock.synchronized {
+        if (_logicalPlan == null) {
+          _logicalPlan = generateLogicalPlan
+        }
+      }
+    }
+    _logicalPlan
+  }
```
marmbrus: This is getting pretty complicated...  Do we really need to include all of this information in `StreamingQueryException`?

zsxwing: > all of this information

Do you mean `logicalPlan`? It's used to generate the plan for each batch.

marmbrus: What I'm trying to say is, (if I understand correctly) all of this extra logic is to make sure that there aren't cycles when we fail to initialize and then try to use that uninitialized information in order to generate the exception.

I'm asking why that information needs to be in the exception at all?  Or if we should just try to initialize earlier in the stream thread and throw a different exception when initialization fails.

zsxwing: I guess it follows `AnalysisException` to include as many as information as possible. @tdas  do you want to clarify it?

marmbrus: Sure, if we separate initialization failure from runtime failures though, then we can only include the extra information when it makes sense (i.e. when we have actually started running and have offsets, etc to report).

brkyvz: I agree with @marmbrus. We can include `logicalPlan` in the `toDebugString` only if the state allows, e.g. `microBatchThreadStatus = INITIALIZED`, and then we don't need any of this extra code

marmbrus: or really what i was suggesting was to have a separate try/catch early in the stream thread that forces initialization of anything that is lazy.

zsxwing:
@marmbrus updated as per our offline discussion. I also added a new method `assertAwaitThread` to avoid the user calling the awaitXXX methods inside the stream thread (e.g., in the StreamingQueryListener.onQueryStarted)

tdas:
@marmbrus should take a look because he had opinions on this

marmbrus:
LGTM

sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamTest.scala
```diff
@@ -360,9 +361,10 @@ trait StreamTest extends QueryTest with SharedSQLContext with Timeouts {
             currentStream.microBatchThread.setUncaughtExceptionHandler(
               new UncaughtExceptionHandler {
                 override def uncaughtException(t: Thread, e: Throwable): Unit = {
-                  streamDeathCause = e
+                  streamThreadDeathCause = e
                 }
               })
+            currentStream.awaitInitialization(streamingTimeout.toMillis)
```
tdas: Can you document why this is needed?

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamExecution.scala
```diff
@@ -559,7 +580,36 @@ class StreamExecution(
   /** A flag to indicate that a batch has completed with no new data available. */
   @volatile private var noNewData = false
 
+  /**
+   * Assert that the await APIs should not be called in the stream thread. Otherwise, it may cause
+   * dead-lock.
```
tdas: Document what kind of deadlock can happen.

sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/StreamExecution.scala
```diff
@@ -559,7 +580,36 @@ class StreamExecution(
...
+   */
+  private def assertAwaitThread(): Unit = {
+    if (microBatchThread eq Thread.currentThread) {
+      throw new IllegalStateException("Cannot wait inside a stream thread")
```
tdas: cannot wait for a stream state from the same thread as the stream.

sql/core/src/test/scala/org/apache/spark/sql/streaming/StreamTest.scala
```diff
@@ -228,7 +228,8 @@ trait StreamTest extends QueryTest with SharedSQLContext with Timeouts {
    */
   def testStream(
       _stream: Dataset[_],
-      outputMode: OutputMode = OutputMode.Append)(actions: StreamAction*): Unit = {
+      outputMode: OutputMode = OutputMode.Append,
+      ignoreThreadDeathCause: Boolean = false)(actions: StreamAction*): Unit = {
```
tdas: This `ignore` field is really hard to understand as a user of this function. And adding a field to just for one test does not make sense. Rather make an action ExpectStreamThreadException.

tdas (commented):
LGTM, pending tests

zsxwing:

PR #16323, opened 2016-12-17
[SPARK-18911] [SQL] Define CatalogStatistics to interact with metastore and convert it to Statistics in relations

wzhfy:

Statistics in LogicalPlan should use attributes to refer to columns rather than column names, because two columns from two relations can have the same column name. But CatalogTable doesn't have the concepts of attribute or broadcast hint in Statistics. Therefore, putting Statistics in CatalogTable is confusing.

We define a different statistic structure in CatalogTable, which is only responsible for interacting with metastore, and is converted to statistics in LogicalPlan when it is used.

## How was this patch tested?

add test cases

### DIFF
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/interface.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/interface.scala
@@ -22,7 +22,7 @@ import java.util.Date
 import org.apache.spark.sql.AnalysisException
 import org.apache.spark.sql.catalyst.{FunctionIdentifier, InternalRow, TableIdentifier}
 import org.apache.spark.sql.catalyst.expressions.{Attribute, Cast, Literal}
-import org.apache.spark.sql.catalyst.plans.logical.{LeafNode, LogicalPlan, Statistics}
+import org.apache.spark.sql.catalyst.plans.logical.{CatalogStatistics, LeafNode, LogicalPlan}
 import org.apache.spark.sql.catalyst.util.quoteIdentifier
 import org.apache.spark.sql.types.{StructField, StructType}
 
@@ -161,7 +161,7 @@ case class CatalogTable(
     createTime: Long = System.currentTimeMillis,
     lastAccessTime: Long = -1,
     properties: Map[String, String] = Map.empty,
-    stats: Option[Statistics] = None,
+    stats: Option[CatalogStatistics] = None,
     viewOriginalText: Option[String] = None,
     viewText: Option[String] = None,
     comment: Option[String] = None,
@@ -196,6 +196,10 @@ case class CatalogTable(
       properties: Map[String, String] = storage.properties): CatalogTable = {
     copy(storage = CatalogStorageFormat(
       locationUri, inputFormat, outputFormat, serde, compressed, properties))
+  }
+
+  def withStats(cboStatsEnabled: Boolean): CatalogTable = {
+    if (cboStatsEnabled) this else copy(stats = None)
   }
 
   override def toString: String = {
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/interface.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/interface.scala
@@ -19,8 +19,6 @@ package org.apache.spark.sql.catalyst.catalog
 
 import java.util.Date
 
-import scala.collection.mutable
-
 import org.apache.spark.sql.AnalysisException
 import org.apache.spark.sql.catalyst.{FunctionIdentifier, InternalRow, TableIdentifier}
 import org.apache.spark.sql.catalyst.expressions.{Attribute, AttributeMap, Cast, Literal}
@@ -241,6 +239,8 @@ case class CatalogTable(
 
 /**
  * This class of statistics is used in [[CatalogTable]] to interact with metastore.
+ * We define this new class instead of directly using [[Statistics]] here because there are no
+ * concepts of attributes or broadcast hint in catalog.
  */
 case class CatalogStatistics(
     sizeInBytes: BigInt,
@@ -251,22 +251,16 @@ case class CatalogStatistics(
    * Convert [[CatalogStatistics]] to [[Statistics]], and match column stats to attributes based
    * on column names.
    */
-  def convert(attributes: Seq[Attribute]): Statistics = {
-    val matched = mutable.HashMap[Attribute, ColumnStat]()
-    attributes.foreach { attr =>
-      if (colStats.contains(attr.name)) {
-        matched.put(attr, colStats(attr.name))
-      }
-    }
+  def toPlanStats(planOutput: Seq[Attribute]): Statistics = {
+    val matched = planOutput.flatMap(a => colStats.get(a.name).map(a -> _))
     Statistics(sizeInBytes = sizeInBytes, rowCount = rowCount,
-      attributeStats = AttributeMap(matched.toSeq))
+      attributeStats = AttributeMap(matched))
   }
 
   /** Readable string representation for the CatalogStatistics. */
   def simpleString: String = {
-    Seq(s"sizeInBytes=$sizeInBytes",
-      if (rowCount.isDefined) s"rowCount=${rowCount.get}" else ""
-    ).filter(_.nonEmpty).mkString(", ")
+    val rowCountString = if (rowCount.isDefined) s", ${rowCount.get} rows" else ""
+    s"$sizeInBytes bytes$rowCountString"
   }
 }
 
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/interface.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/interface.scala
@@ -200,10 +200,6 @@ case class CatalogTable(
       locationUri, inputFormat, outputFormat, serde, compressed, properties))
   }
 
-  def withStats(cboStatsEnabled: Boolean): CatalogTable = {
-    if (cboStatsEnabled) this else copy(stats = None)
-  }
-
   override def toString: String = {
     val tableProperties = properties.map(p => p._1 + "=" + p._2).mkString("[", ", ", "]")
     val partitionColumns = partitionColumnNames.map(quoteIdentifier).mkString("[", ", ", "]")
--- a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/interface.scala
+++ b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/interface.scala
@@ -19,10 +19,12 @@ package org.apache.spark.sql.catalyst.catalog
 
 import java.util.Date
 
+import scala.collection.mutable
+
 import org.apache.spark.sql.AnalysisException
 import org.apache.spark.sql.catalyst.{FunctionIdentifier, InternalRow, TableIdentifier}
-import org.apache.spark.sql.catalyst.expressions.{Attribute, Cast, Literal}
-import org.apache.spark.sql.catalyst.plans.logical.{CatalogStatistics, LeafNode, LogicalPlan}
+import org.apache.spark.sql.catalyst.expressions.{Attribute, AttributeMap, Cast, Literal}
+import org.apache.spark.sql.catalyst.plans.logical._
 import org.apache.spark.sql.catalyst.util.quoteIdentifier
 import org.apache.spark.sql.types.{StructField, StructType}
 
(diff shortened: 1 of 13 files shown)


wzhfy:
cc @rxin @cloud-fan

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/Statistics.scala
```diff
@@ -61,6 +63,37 @@ case class Statistics(
   }
 }
 
+/**
+ * This class of Statistics is used in [[CatalogTable]] to interact with metastore.
+ */
+case class CatalogStatistics(
```
cloud-fan: shall we define this class in the same file of `CatalogTable`?

cloud-fan:
I think it's pretty safe to use table stats as the stats of the leaf node(table relation), including column stats. The actual dangerous one is when we going to estimate something, e.g. in `Join` or `Aggregate`.

So logically we should read this conf in `Join` or `Aggregate`, and decide if we want to estimate something or just do some naive calculation. However, a problem is, we can't get the conf in `LogicalPlan.statistics`.

A possible approach is, we can change `LogicalPlan.statistics` to `def statistics(conf: CatalystConf)`. We need to update all the implementation and caller side though.

wzhfy:
Change LogicalPlan.statistics to def statistics(conf: CatalystConf) could have two problems:
1. we can't override it as `lazy val`, and `def` means we need to estimate the plan every time `statistics` is called, which will be a performance hit.
2. we need to make sure we have conf everywhere `def statistics(conf: CatalystConf)` is used.

wzhfy:
retest this please

cloud-fan:
> we can't override it as lazy val, and def means we need to estimate the plan every time statistics is called, which will be a performance hit.

I think we can do the cache manually:
```
@transient var estimatedStats: Statistics = null
@transient var simpleStats: Statistics = null
def statistics(conf: CatalystConf) = {
  if (conf.enableCbo) {
    if (estimatedStats == null) {
      estimatedStats = ...
    }
    estimatedStats
  } else {
    if (simpleStats == null) {
      simpleStats = ...
    }
    simpleStats
  }
}
```

> we need to make sure we have conf everywhere def statistics(conf: CatalystConf) is used.

Do we have a problem here? I think all of the places need to call `statistics` can access CatalystConf.

wzhfy:
Ok, I think it's doable. But since it's not a small change, let's wait @rxin for his comment.

wzhfy:
retest this please

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/interface.scala
```diff
@@ -198,6 +200,10 @@ case class CatalogTable(
       locationUri, inputFormat, outputFormat, serde, compressed, properties))
   }
 
+  def withStats(cboStatsEnabled: Boolean): CatalogTable = {
```
viirya: Do we need to get rid of the `CatalogStatistics` if the config is off? Actually I think you can decide whether to use it or not when doing estimation later, depending on this config. It seems no harm to always attach this info to `CatalogTable`.

wzhfy: Yes I also think that's better, but as @cloud-fan said, we can't get the config in `def statistics`, we have to modify many places to support this. I'm about to do such modifications, do you have any advices to minimize the changes?

viirya: I can think of two approaches:

1.

We can keep the current naive version of `statistics` and add new `statistics` function which takes conf.

A default implementation of the new `statistics` function simply returns the naive version of `statistics`.

In `Join` or `Aggregate`, we can include more complex logic in the new `statistics` to return naive calculation or something estimation.

The caller always calls new `statistics` function and passes in current conf.

2.

Add new `statisticsCBO` which doesn't take conf because it is called only cbo is enabled. So the caller decides to call non-cbo version `statistics` or cbo version `statisticsCBO`.

wzhfy: Thanks. I think the first one is better, the second one will lead to many if-else on caller sides.

wzhfy:
Since adding a switch for cbo is not a trivial one, I want to do it in a separate pr, and let this one only deal with decoupling Statistics from CatalogTable. Do you agree? @cloud-fan

cloud-fan:
SGTM

wzhfy:
Updated. Please review @cloud-fan

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/interface.scala
```diff
@@ -237,6 +239,38 @@ case class CatalogTable(
...
+      if (colStats.contains(attr.name)) {
+        matched.put(attr, colStats(attr.name))
+      }
+    }
+    Statistics(sizeInBytes = sizeInBytes, rowCount = rowCount,
+      attributeStats = AttributeMap(matched.toSeq))
+  }
+
+  /** Readable string representation for the CatalogStatistics. */
+  def simpleString: String = {
```
cloud-fan: why do you define a `simpleString` instead of override `toString`?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/interface.scala
```diff
@@ -237,6 +239,38 @@ case class CatalogTable(
...
+    Seq(s"sizeInBytes=$sizeInBytes",
+      if (rowCount.isDefined) s"rowCount=${rowCount.get}" else ""
+    ).filter(_.nonEmpty).mkString(", ")
```
cloud-fan:
```
val rowCountString = if (rowCount.isDefined) s", ${rowCount.get} rows" else ""
s"$sizeInBytes bytes$rowCountString" 
```

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/interface.scala
```diff
@@ -237,6 +239,38 @@ case class CatalogTable(
...
+   * Convert [[CatalogStatistics]] to [[Statistics]], and match column stats to attributes based
+   * on column names.
+   */
+  def convert(attributes: Seq[Attribute]): Statistics = {
+    val matched = mutable.HashMap[Attribute, ColumnStat]()
+    attributes.foreach { attr =>
+      if (colStats.contains(attr.name)) {
+        matched.put(attr, colStats(attr.name))
+      }
+    }
```
cloud-fan: attributes.flatMap(a => colStats.get(a.name).map(a -> _)).toMap

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/interface.scala
```diff
@@ -237,6 +239,38 @@ case class CatalogTable(
...
+case class CatalogStatistics(
+    sizeInBytes: BigInt,
+    rowCount: Option[BigInt] = None,
+    colStats: Map[String, ColumnStat] = Map.empty) {
+
+  /**
+   * Convert [[CatalogStatistics]] to [[Statistics]], and match column stats to attributes based
+   * on column names.
+   */
+  def convert(attributes: Seq[Attribute]): Statistics = {
```
cloud-fan: This is a bad name, it doesn't tell anything, without looking at the doc. How about `def toPlanStats(planOuput: ...)`?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/interface.scala
```diff
@@ -237,6 +239,38 @@ case class CatalogTable(
 }
 
 
+/**
+ * This class of statistics is used in [[CatalogTable]] to interact with metastore.
```
viirya: Can you add few words explaining why don't use `Statistics` for `CatalogTable`?

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/Statistics.scala
```diff
@@ -41,13 +41,13 @@ import org.apache.spark.sql.types._
...
  * @param rowCount Estimated number of rows.
- * @param colStats Column-level statistics.
+ * @param attributeStats Statistics for Attributes.
  * @param isBroadcastable If true, output is small enough to be used in a broadcast join.
  */
 case class Statistics(
     sizeInBytes: BigInt,
     rowCount: Option[BigInt] = None,
-    colStats: Map[String, ColumnStat] = Map.empty,
+    attributeStats: AttributeMap[ColumnStat] = AttributeMap(Nil),
```
viirya: Will we estimate statistics for all attributes in logical plan?

I meant if an attribute is not coming from a leaf node but from a later plan like `Join`, do we still have `ColumnStat` for it?

If not, I think we don't need to call this parameter as `attributeStats`, instead of original `colStats`.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/interface.scala
wzhfy: Because we don't print column stats in it, it's not a "complete" string representation. Column stats can be too much and make CatalogTable unreadable.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/interface.scala
wzhfy: That's a lot better, thanks!

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/interface.scala
wzhfy: ok

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/Statistics.scala
wzhfy: We will estimate attributes in logical plan from the bottom up.

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/interface.scala
wzhfy: fixed

sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/interface.scala
wzhfy: fixed

sql/core/src/test/scala/org/apache/spark/sql/StatisticsCollectionSuite.scala
```diff
@@ -260,4 +261,46 @@ abstract class StatisticsCollectionTestBase extends QueryTest with SQLTestUtils
...
+      // Test hive serde table
+      if (spark.conf.get(StaticSQLConf.CATALOG_IMPLEMENTATION) == "hive") {
+        checkStatsConversion(tableName = "hive_tbl", isDatasourceTable = false)
+      }
+    }
+  }
+
+  private def checkStatsConversion(tableName: String, isDatasourceTable: Boolean): Unit = {
+    // Create an empty table and run analyze command on it.
+    val col = "c1"
```
cloud-fan: nit: `c1` is so simple that we can write in directly instead of using a variable

sql/core/src/test/scala/org/apache/spark/sql/StatisticsCollectionSuite.scala
```diff
@@ -260,4 +261,46 @@ abstract class StatisticsCollectionTestBase extends QueryTest with SQLTestUtils
...
+    val createTableSql = if (isDatasourceTable) {
+      s"CREATE TABLE $tableName ($col INT) USING PARQUET"
```
cloud-fan: let's create a table with 2 columns, and only analyze one column, to see if the `attributeStats` only contains one entry.

sql/core/src/test/scala/org/apache/spark/sql/StatisticsCollectionSuite.scala
wzhfy: fixed

sql/core/src/test/scala/org/apache/spark/sql/StatisticsCollectionSuite.scala
wzhfy: ok, fixed

cloud-fan:
LGTM, pending jenkins

cloud-fan:
retest this please

cloud-fan:
thanks, merging to master!

PR #16328, opened 2016-12-18
[SPARK-18808][ML][MLLIB] ml.KMeansModel.transform is very inefficient

srowen:

mllib.KMeansModel.clusterCentersWithNorm is a method than ends up being called every time `predict` is called on a single vector, which is bad news for now the ml.KMeansModel Transformer works, which necessarily transforms one vector at a time.

This causes the model to just store the vectors with norms upfront. The extra norm should be small compared to the vectors. This would avoid this form of overhead on this and other code paths.

## How was this patch tested?

Existing tests.

### DIFF
--- a/mllib/src/main/scala/org/apache/spark/mllib/clustering/KMeansModel.scala
+++ b/mllib/src/main/scala/org/apache/spark/mllib/clustering/KMeansModel.scala
@@ -39,7 +39,8 @@ import org.apache.spark.sql.{Row, SparkSession}
 class KMeansModel @Since("1.1.0") (@Since("1.0.0") val clusterCenters: Array[Vector])
   extends Saveable with Serializable with PMMLExportable {
 
-  private val clusterCentersWithNorm = clusterCenters.map(new VectorWithNorm(_))
+  private val clusterCentersWithNorm =
+    if (clusterCenters == null) null else clusterCenters.map(new VectorWithNorm(_))
 
   /**
    * A Java-friendly constructor that takes an Iterable of Vectors.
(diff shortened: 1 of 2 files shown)


mllib/src/main/scala/org/apache/spark/mllib/clustering/KMeansModel.scala
```diff
@@ -39,6 +39,8 @@ import org.apache.spark.sql.{Row, SparkSession}
 class KMeansModel @Since("1.1.0") (@Since("1.0.0") val clusterCenters: Array[Vector])
   extends Saveable with Serializable with PMMLExportable {
 
+  private val clusterCentersWithNorm = clusterCenters.map(new VectorWithNorm(_))
```
srowen: I believe `clusterCenters` will _still_ be a separate member, given the declaration. However, it will contain references to the same `Vector` objects. The vectors won't be duplicated, but some references will. This still seems like negligible overhead compared to the benefit; it's kilobytes of memory.

srowen:
I finished this approach for completeness, but I see that it requires 'fixing' `StreamingKMeansModel` to return a new model instead of strictly updating its vectors in place. It's not actually a big deal but would cause a problem if anyone called `StreamingKMeansModel.update` without using its return value. Spark doesn't and I'm not sure if it's valid. 

There's another way to approach this by passing around a broadcast variable. It's narrower and a bit more complex, so it'd be nice if we could make this change, but it's not the only solution. Leaving it open for comments, esp. from Jenkins.

CC @freeman-lab and @mengxr

srowen:
Ping @jkbradley or mabye @MLnick for an opinion on whether this is an OK solution. Certainly it's a nicer fix than the alternative but has a slight behavior change.

srowen:
Going once going twice -- I favor making this change but want to see if anyone feels at all strongly about it.

srowen:
Merged to master

PR #16329, opened 2016-12-18
[SPARK-16046][DOCS] Aggregations in the Spark SQL programming guide

aokolnychyi:

- A separate subsection for Aggregations under “Getting Started” in the Spark SQL programming guide. It mentions which aggregate functions are predefined and how users can create their own.
- Examples of using the `UserDefinedAggregateFunction` abstract class for untyped aggregations in Java and Scala.
- Examples of using the `Aggregator` abstract class for type-safe aggregations in Java and Scala.
- Python is not covered. 
- The PR might not resolve the ticket since I do not know what exactly was planned by the author. 

In total, there are four new standalone examples that can be executed via `spark-submit` or `run-example`. The updated Spark SQL programming guide references to these examples and does not contain hard-coded snippets. 

## How was this patch tested?

The patch was tested locally by building the docs. The examples were run as well. 

image

### DIFF
--- a/docs/sql-programming-guide.md
+++ b/docs/sql-programming-guide.md
@@ -382,6 +382,52 @@ For example:
 
 </div>
 
+## Aggregations
+
+The [built-in DataFrames functions](api/scala/index.html#org.apache.spark.sql.functions$) mentioned 
+before provide such common aggregations as `count()`, `countDistinct()`, `avg()`, `max()`, `min()`, etc.
+While those functions are designed for DataFrames, Spark SQL also has type-safe versions for some of them in 
+[Scala](api/scala/index.html#org.apache.spark.sql.expressions.scalalang.typed$) and 
+[Java](api/java/org/apache/spark/sql/expressions/javalang/typed.html) to work with strongly typed Datasets.
+Moreover, users are not limited to the predefined aggregate functions and can create their own.
+
+### Untyped User-Defined Aggregate Functions
+
+<div class="codetabs">
+
+<div data-lang="scala"  markdown="1">
+
+Users have to extend the [UserDefinedAggregateFunction](api/scala/index.html#org.apache.spark.sql.expressions.UserDefinedAggregateFunction) 
+abstract class to implement a custom untyped aggregate function. For example, a user-defined average
+can look like:
+
+{% include_example untyped_custom_aggregation scala/org/apache/spark/examples/sql/UserDefinedUntypedAggregation.scala%}
+</div>
+
+<div data-lang="java"  markdown="1">
+
+{% include_example untyped_custom_aggregation java/org/apache/spark/examples/sql/JavaUserDefinedUntypedAggregation.java%}
+</div>
+
+</div>
+
+### Type-Safe User-Defined Aggregate Functions
+
+User-defined aggregations for strongly typed Datasets revolve around the [Aggregator](api/scala/index.html#org.apache.spark.sql.expressions.Aggregator) abstract class.
+For example, a type-safe user-defined average can look like:
+<div class="codetabs">
+
+<div data-lang="scala"  markdown="1">
+
+{% include_example typed_custom_aggregation scala/org/apache/spark/examples/sql/UserDefinedTypedAggregation.scala%}
+</div>
+
+<div data-lang="java"  markdown="1">
+
+{% include_example typed_custom_aggregation java/org/apache/spark/examples/sql/JavaUserDefinedTypedAggregation.java%}
+</div>
+
+</div>
 
 # Data Sources
 
--- a/docs/sql-programming-guide.md
+++ b/docs/sql-programming-guide.md
@@ -384,8 +384,8 @@ For example:
 
 ## Aggregations
 
-The [built-in DataFrames functions](api/scala/index.html#org.apache.spark.sql.functions$) mentioned 
-before provide such common aggregations as `count()`, `countDistinct()`, `avg()`, `max()`, `min()`, etc.
+The [built-in DataFrames functions](api/scala/index.html#org.apache.spark.sql.functions$) provide common
+aggregations such as `count()`, `countDistinct()`, `avg()`, `max()`, `min()`, etc.
 While those functions are designed for DataFrames, Spark SQL also has type-safe versions for some of them in 
 [Scala](api/scala/index.html#org.apache.spark.sql.expressions.scalalang.typed$) and 
 [Java](api/java/org/apache/spark/sql/expressions/javalang/typed.html) to work with strongly typed Datasets.
--- a/examples/src/main/java/org/apache/spark/examples/sql/JavaUserDefinedTypedAggregation.java
+++ b/examples/src/main/java/org/apache/spark/examples/sql/JavaUserDefinedTypedAggregation.java
@@ -1,0 +1,154 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.spark.examples.sql;
+
+// $example on:typed_custom_aggregation$
+import java.io.Serializable;
+
+import org.apache.spark.sql.Dataset;
+import org.apache.spark.sql.Encoder;
+import org.apache.spark.sql.Encoders;
+import org.apache.spark.sql.SparkSession;
+import org.apache.spark.sql.TypedColumn;
+import org.apache.spark.sql.expressions.Aggregator;
+// $example off:typed_custom_aggregation$
+
+public class JavaUserDefinedTypedAggregation {
+
+  // $example on:typed_custom_aggregation$
+  public static class Salary implements Serializable {
+    private String person;
+    private long salary;
+
+    // Constructors, getters, setters...
+    // $example off:typed_custom_aggregation$
+    public String getPerson() {
+      return person;
+    }
+
+    public void setPerson(String person) {
+      this.person = person;
+    }
+
+    public long getSalary() {
+      return salary;
+    }
+
+    public void setSalary(long salary) {
+      this.salary = salary;
+    }
+    // $example on:typed_custom_aggregation$
+  }
+
+  public static class Average implements Serializable  {
+    private long sum;
+    private long count;
+
+    // Constructors, getters, setters...
+    // $example off:typed_custom_aggregation$
+    public Average() {
+    }
+
+    public Average(long sum, long count) {
+      this.sum = sum;
+      this.count = count;
+    }
+
+    public long getSum() {
+      return sum;
+    }
+
+    public void setSum(long sum) {
+      this.sum = sum;
+    }
+
+    public long getCount() {
+      return count;
+    }
+
+    public void setCount(long count) {
+      this.count = count;
+    }
+    // $example on:typed_custom_aggregation$
+  }
+
+  public static class MyAverage extends Aggregator<Salary, Average, Double> {
+    // A zero value for this aggregation. Should satisfy the property that any b + zero = b
+    public Average zero() {
+      return new Average(0L, 0L);
+    }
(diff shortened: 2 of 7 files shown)


examples/src/main/scala/org/apache/spark/examples/sql/UserDefinedUntypedAggregation.scala
```diff
@@ -0,0 +1,97 @@
...
+import org.apache.spark.sql.Row
+import org.apache.spark.sql.SparkSession
+// $example off:untyped_custom_aggregation$
+
+object UserDefinedUntypedAggregation {
+
+  // $example on:untyped_custom_aggregation$
+  object MyAverage extends UserDefinedAggregateFunction {
+    // Data types of input arguments
+    def inputSchema: StructType = StructType(StructField("salary", LongType) :: Nil)
```
assafmendelson: Maybe add a little explanation here. For example, when I first saw this I tried to figure out where "salary" appears in the code as in practice it is being accessed by index only (input.getLong(0)).

examples/src/main/scala/org/apache/spark/examples/sql/UserDefinedUntypedAggregation.scala
```diff
@@ -0,0 +1,97 @@
...
+    // Data types of values in the aggregation buffer
+    def bufferSchema: StructType = {
+      StructType(StructField("sum", LongType) :: StructField("count", LongType) :: Nil)
+    }
+    // The data type of the returned value
+    def dataType: DataType = DoubleType
+    // Whether this function always returns the same output on the identical input
+    def deterministic: Boolean = true
+    // Initializes the given aggregation buffer
+    def initialize(buffer: MutableAggregationBuffer): Unit = {
```
assafmendelson: I believe an explanation on what MutableAggregationBuffer is should be added.
Basically explain how to access it, what it means for it to be mutable (including probably explaining that arrays and map types are immutable even if the buffer itself is mutable) etc.

examples/src/main/scala/org/apache/spark/examples/sql/UserDefinedUntypedAggregation.scala
assafmendelson: I would go with inputColumn. 
What I think should be more strongly explained is that this is basically the schema of the input for the aggregate function and not for the source dataframe.  Basically someone might think that their original dataframe might need to have this name for the column.

examples/src/main/java/org/apache/spark/examples/sql/JavaUserDefinedTypedAggregation.java
```diff
@@ -0,0 +1,154 @@
...
+    }
+    // $example on:typed_custom_aggregation$
+  }
+
+  public static class MyAverage extends Aggregator<Salary, Average, Double> {
+    // A zero value for this aggregation. Should satisfy the property that any b + zero = b
+    public Average zero() {
+      return new Average(0L, 0L);
+    }
+    // Combine two values to produce a new value. For performance, the function may modify `b` and
```
marmbrus: Its a little confusing to have the comment here for this optimization, but then not implement it.

examples/src/main/java/org/apache/spark/examples/sql/JavaUserDefinedTypedAggregation.java
```diff
@@ -0,0 +1,154 @@
...
+import org.apache.spark.sql.Encoders;
+import org.apache.spark.sql.SparkSession;
+import org.apache.spark.sql.TypedColumn;
+import org.apache.spark.sql.expressions.Aggregator;
+// $example off:typed_custom_aggregation$
+
+public class JavaUserDefinedTypedAggregation {
+
+  // $example on:typed_custom_aggregation$
+  public static class Salary implements Serializable {
```
marmbrus: I might be a little clearer if this was a `Person` with a `name` and `salary`.

examples/src/main/scala/org/apache/spark/examples/sql/UserDefinedTypedAggregation.scala
```diff
@@ -0,0 +1,82 @@
...
+object UserDefinedTypedAggregation {
+
+  // $example on:typed_custom_aggregation$
+  case class Salary(person: String, salary: Long)
+  case class Average(sum: Long, count: Long)
+
+  object MyAverage extends Aggregator[Salary, Average, Double] {
+    // A zero value for this aggregation. Should satisfy the property that any b + zero = b
+    def zero: Average = Average(0L, 0L)
+    // Combine two values to produce a new value. For performance, the function may modify `b` and
```
marmbrus: Same comment here with object reuse.

examples/src/main/scala/org/apache/spark/examples/sql/UserDefinedTypedAggregation.scala
```diff
@@ -0,0 +1,82 @@
...
+    // +-------+------+
+    // | person|salary|
+    // +-------+------+
+    // |Michael|  3000|
+    // |   Andy|  4500|
+    // | Justin|  3500|
+    // |  Berta|  4000|
+    // +-------+------+
+
+    val averageSalary = MyAverage.toColumn.name("average_salary")
```
marmbrus: Maybe comment what `name` is doing here.  I actually had to look it up.

marmbrus:
This is great!  Thanks for taking the time to write up such complete examples.  I think this was a big gap in the existing docs.  One other ask.  The screen-shot is great, but I'd like to see which parts actually make it into the code snipits in the doc.  Ideally you could post a link to compiled doc.  If thats hard I can also try to build locally though.

aokolnychyi:
@marmbrus I have updated the pull request. The compiled docs can be found [here](https://aokolnychyi.github.io/spark-docs/sql-programming-guide.html). 

I did not manage to build the Java API docs. I believe the problem is in my local installation. Therefore, I checked each url manually, they should work once the API docs are compiled. I will verify everything one more time in the nightly build.

docs/sql-programming-guide.md
```diff
@@ -382,6 +382,52 @@ For example:
 
 </div>
 
+## Aggregations
+
+The [built-in DataFrames functions](api/scala/index.html#org.apache.spark.sql.functions$) mentioned 
+before provide such common aggregations as `count()`, `countDistinct()`, `avg()`, `max()`, `min()`, etc.
```
jnh5y: As a suggestion, I'd change this to read:
"The [built-in DataFrames functions](api/scala/index.html#org.apache.spark.sql.functions$) provide common aggregations such as `count()`, `countDistinct()`, `avg()`, `max()`, and `min()`."

docs/sql-programming-guide.md
```diff
@@ -382,6 +382,52 @@ For example:
...
+While those functions are designed for DataFrames, Spark SQL also has type-safe versions for some of them in 
+[Scala](api/scala/index.html#org.apache.spark.sql.expressions.scalalang.typed$) and 
+[Java](api/java/org/apache/spark/sql/expressions/javalang/typed.html) to work with strongly typed Datasets.
+Moreover, users are not limited to the predefined aggregate functions and can create their own.
```
jnh5y: I think it'd be worth showing an Spark SQL example using the included/pre-defined functions.  Since your example implements 'avg', maybe use 'min' / 'max'?

Alternatively, the example could be added to the SQL statements in the main driver for the UserDefinedAggregateFunction implementations.

examples/src/main/java/org/apache/spark/examples/sql/JavaUserDefinedUntypedAggregation.java
```diff
@@ -0,0 +1,132 @@
...
+import org.apache.spark.sql.types.DataType;
+import org.apache.spark.sql.types.DataTypes;
+import org.apache.spark.sql.types.StructField;
+import org.apache.spark.sql.types.StructType;
+// $example off:untyped_custom_aggregation$
+
+public class JavaUserDefinedUntypedAggregation {
+
+  // $example on:untyped_custom_aggregation$
+  public static class MyAverage extends UserDefinedAggregateFunction {
```
jnh5y: Should all UDAFs written in Java be static classes?  Similarly, should Scala implementations be Scala objects?

jnh5y (approved):
My comments are generally small suggestions; the PR looks great to me. 

I wonder if a discussion of the types and functions in a UDAF would be worthwhile.

Also, the code looks similar to the Hive test code here: https://github.com/apache/spark/blob/master/sql/hive/src/test/java/org/apache/spark/sql/hive/aggregate/MyDoubleAvg.java.

docs/sql-programming-guide.md
aokolnychyi: I also thought about this. In my view, it will be appropriate to have a separate subsection before Aggregations to show how to apply predefined SQL functions, including writing your own UDFs. That's will be worth another pull request. Alternatively, I can also try to extend this one to add an example of `max()` or `min()`. @marmbrus what's your opinion?

examples/src/main/java/org/apache/spark/examples/sql/JavaUserDefinedUntypedAggregation.java
aokolnychyi: I tried to group all relevant entities for each example into one single class. Static classes are used here just to avoid the cumbersome code for creating the entities inside the `main()` method. Scala objects are also used due to a similar reason. I can mention this but it will add more comments to the already long examples.

michalsenkyr:
If you are having trouble building Javadoc, try switching to Java 7 temporarily. Java 8 introduced stricter Javadoc rules that may fail the docs build. Unfortunately Jenkins doesn't, so new errors get introduced over time.

examples/src/main/scala/org/apache/spark/examples/sql/UserDefinedTypedAggregation.scala
```diff
@@ -0,0 +1,87 @@
...
+    def zero: Average = Average(0L, 0L)
+    // Combine two values to produce a new value. For performance, the function may modify `buffer`
+    // and return it instead of constructing a new object
+    def reduce(buffer: Average, employee: Employee): Average = {
+      buffer.sum += employee.salary
+      buffer.count += 1
+      buffer
+    }
+    // Merge two intermediate values
+    def merge(b1: Average, b2: Average): Average = Average(b1.sum + b2.sum, b1.count + b2.count)
```
aokolnychyi: @michalsenkyr It is not required to create a new object in the `merge` method. One can modify the vars and return the existing object just like in the `reduce` method.  However, it is less critical here since this method will be called on pre-aggregated data and not for every element. On the one hand, I can apply here the same approach as in the `reduce` method to make the example consistent. On the other hand, the current code shows that it is not mandatory to modify vars. Probably, a comment might help. I am not sure which approach is better. Therefore, I am open to suggestions.

michalsenkyr: Personally, I prefer consistency. When I saw this, I immediately wondered whether there is a specific reason you did it this way.
I'd rather see both methods use the same paradigm. In this case probably the immutable one as the option of mutability is already mentioned in the comment above.
Or you can mention it again in the comment on this method if you want to provide examples of both. This way it just seems a little confusing.

examples/src/main/java/org/apache/spark/examples/sql/JavaUserDefinedTypedAggregation.java
```diff
@@ -0,0 +1,160 @@
...
+
+    public void setCount(long count) {
+      this.count = count;
+    }
+    // $example on:typed_custom_aggregation$
+  }
+
+  public static class MyAverage extends Aggregator<Employee, Average, Double> {
+    // A zero value for this aggregation. Should satisfy the property that any b + zero = b
+    public Average zero() {
```
srowen: Is this meant to be `MyAverage`?

srowen (commented):
Except for one small question, the text/code looks OK. Others who know the content seem to approve, and you have gotten tests to pass, so seems reasonable.

examples/src/main/java/org/apache/spark/examples/sql/JavaUserDefinedTypedAggregation.java
aokolnychyi: @srowen `Average` is a Java bean that holds current sum and count. It is defined earlier. Here it represents a zero value. `MyAverage`, in turn, is the actual aggregator that accepts instances of the `Employee` class, stores intermediate results using an instance of`Average`, and produces `Double` as a result. 

I can rename `MyAverage` to `MyAverageAggregator` if this makes things clearer.

srowen: My bad, I read this incorrectly while skimming.

marmbrus:
Sorry for the delay.  This LGTM, but I'm currently away from my Apache SSH keys.  Other committers should feel free to merge if you get there before I do.

gatorsmile:
Sure, let me quickly go over the changes. Will merge it after that.

gatorsmile (approved):
LGTM

gatorsmile:
Thanks! Merging to master/2.1

PR #16331, opened 2016-12-19
[SPARK-18920][HISTORYSERVER]Update outdated date formatting

WangTaoTheTonic:

Before we show "-" while the timestamp is less than 0, we should update it as now the date string is presented in format "yyyy-MM-dd ....."

## How was this patch tested?

**Before:**
historyserver-before

**After:**
history-after

### DIFF
--- a/core/src/main/resources/org/apache/spark/ui/static/historypage.js
+++ b/core/src/main/resources/org/apache/spark/ui/static/historypage.js
@@ -38,7 +38,7 @@ function makeIdNumeric(id) {
 }
 
 function formatDate(date) {
-  if (date <= 0) return "-";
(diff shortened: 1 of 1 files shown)


srowen:
I don't think this can be the right change -- it's construing any date before 2000 as invalid, which may happen to be OK but is pretty arbitrary. It arises as the result of formatting -1 as a date, and this method seems to try to handle that, but it's comparing the string to -1 I guess. Surely this should be handled elsewhere, like in whatever calls `formatDate`?

srowen:
Ping @WangTaoTheTonic I don't think we can make this change -- can you investigate a more fundamental fix or close this?

PR #16332, opened 2016-12-19
[SPARK-18921][SQL] check database existence with Hive.databaseExists instead of getDatabase

cloud-fan:

It's weird that we use `Hive.getDatabase` to check the existence of a database, while Hive has a `databaseExists` interface.

What's worse, `Hive.getDatabase` will produce an error message if the database doesn't exist, which is annoying when we only want to check the database existence.

This PR fixes this and use `Hive.databaseExists` to check database existence.

## How was this patch tested?

N/A


cloud-fan:
cc @yhuai @gatorsmile

gatorsmile:
LGTM

yhuai:
LGTM. Merging to master and branch 2.1.

PR #16337, opened 2016-12-19
[SPARK-18871][SQL] New test cases for IN/NOT IN subquery

kevinyu98:
This PR extends the existing IN/NOT IN subquery test cases coverage, adds more test cases to the IN subquery test suite. 

Based on the discussion, we will create  `subquery/in-subquery` sub structure under `sql/core/src/test/resources/sql-tests/inputs` directory. 

This is the high level grouping for IN subquery:

`subquery/in-subquery/`
`subquery/in-subquery/simple-in.sql`
`subquery/in-subquery/in-group-by.sql (in parent side, subquery, and both)`
`subquery/in-subquery/not-in-group-by.sql`
`subquery/in-subquery/in-order-by.sql`
`subquery/in-subquery/in-limit.sql`
`subquery/in-subquery/in-having.sql`
`subquery/in-subquery/in-joins.sql`
`subquery/in-subquery/not-in-joins.sql`
`subquery/in-subquery/in-set-operations.sql`
`subquery/in-subquery/in-with-cte.sql`
`subquery/in-subquery/not-in-with-cte.sql`
subquery/in-subquery/in-multiple-columns.sql`

We will deliver it through multiple prs, this is the first pr for the IN subquery, it has

`subquery/in-subquery/simple-in.sql`
`subquery/in-subquery/in-group-by.sql (in parent side, subquery, and both)`

These are the results from running on DB2.
[Modified test file of in-group-by.sql used to run on DB2](https://github.com/apache/spark/files/683367/in-group-by.sql.db2.txt)
[Output of the run result on DB2](https://github.com/apache/spark/files/683362/in-group-by.sql.db2.out.txt)
[Modified test file of simple-in.sql used to run on DB2](https://github.com/apache/spark/files/683378/simple-in.sql.db2.txt)
[Output of the run result on DB2](https://github.com/apache/spark/files/683379/simple-in.sql.db2.out.txt)

## How was this patch tested?

This patch is adding tests.


gatorsmile:
ok to test

gatorsmile:
cc @hvanhovell @nsyca

hvanhovell:
ok to test

hvanhovell:
@gatorsmile I think you need to ask @shaneknapp for jenkins permissions.

hvanhovell:
@kevinyu98 this looks like pretty solid work. A couple of questions:

- This is a **huge** PR.  This will be hard to review. Is there a change that we can split this up in more digestible pieces.
- On a related note. Given the size, it might be easy to miss a correctness issue.
- Are the tests part of a test kit?

nsyca:
@hvanhovell: On the correctness, how about running the same set of test cases against a second SQL system and comparing the results? If the two systems yield the same result, we can assume Spark has the correct result.

Besides the correctness, what aspect of review would you be thinking? Areas of test coverage? Two more sets of test cases for EXISTS and scalar subquery will be submitted in the coming week. I want to make sure we will follow the same process.

Note that there are negative test cases (test cases that currently raise Analysis exceptions) we have not included in this PR. We plan to put those in a separate PR with some minor code change to suppress the output of the full LogicalPlan along with the error messages. This is because the infrastructure of SQLQueryTestSuite might not be designed to compare two equivalent LogicalPlans of different ExprIds.

We want to have sufficient test coverage for all the three forms of subquery. This way when we will submit the code to defer the pull up, we can run against these sets of test cases to mitigate any undesired effect from the code change.

I don't understand your last point (are the tests part of a test kit?). Could you please elaborate?

hvanhovell:
@nsyca @kevinyu98
### 1. Correctness
> how about running the same set of test cases against a second SQL system and comparing the results?
Yeah, that works. I am just trying to make sure that nothing slips through the cracks here.

### 2. Size
The PR is huge and that makes it relatively hard to review (without missing things). Testing against another DB would help a lot for reviewing purposes (it would effectively m).

### 3. Test kit
I was impressed by the completeness of the tests. I was just wondering if you have wrote them (kudos), or if they are a part of an existing test frame work.

### 4. Negative testing
We might have to clean-up (remove expr_ids) from the LogicalPlan a little bit to make negative testing work. An Analysis Exception should contain the failing plan, and we could use code similar to https://github.com/apache/spark/blob/master/sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/plans/PlanTest.scala#L30-L74 to get this working.

nsyca:
- On 1&2: We will show the output from a comparison of the result sets with a second DB.
- On 3: We wrote them. It's labour intensive. However, we plan to work on a better method of adding test coverage through auto generation. Let's collaborate via a separate PR on that topic.
On a side note, in the future there will be much much more test cases in the collections and I am concerned about the length of the run. We shall start thinking about organizing test cases in multiple collections. First-tier are to be run as a sanity check for merging PRs. Subsequent tiers catch up later on a snapshot of multiple PRs. 
- On 4: Thanks for the suggestion. We will check it out.

hvanhovell:
On 1 & 2 - it might be nice if we can use the spark JDBC connector to run the tests directly on a different system. That would make comparisons trivial.

On 3 -  what is the current run-time of all tests? There is a tagging facility for such tests, but I have never used it in practice.

sql/core/src/test/resources/sql-tests/inputs/subq01-in-subquery.sql
```diff
@@ -0,0 +1,452 @@
+-- A test suite for IN predicate subquery
```
nsyca: The reason of this renaming is to have the future test files of the same group as subqXX-<name>.sql so they are next to each other when `ls`.

nsyca:
@hvanhovell Does Spark JDBC connector act purely as a JDBC driver which supposes to send an entire SQL statement to run on the destination system? If there are part of the SQL statement processed in Spark, I'm afraid that is a contamination that we do not want in this verification exercise.

gatorsmile:
In Spark, we can use the following way to send the whole query to the underlying data source.
```SQL
CREATE TEMPORARY TABLE jdbcTable
USING jdbc
OPTIONS (
    url "jdbc:postgresql:dbserver",
    dbtable "(SELECT * FROM mytable) tmp"
);
```

sql/core/src/test/resources/sql-tests/inputs/subq01-in-subquery.sql
dongjoon-hyun: For that purpose, can we remove the serial number `01`?

gatorsmile: Please drop the serial number.

gatorsmile:
Could we split the test suites based on the subquery types?

- [With aliases](https://technet.microsoft.com/en-us/library/ms190410(v=sql.105).aspx). 
- With IN or NOT IN
    - [Subqueries with IN ](https://technet.microsoft.com/en-us/library/ms190963(v=sql.105).aspx)
    - [Subqueries with NOT IN](https://technet.microsoft.com/en-us/library/ms189062(v=sql.105).aspx).
- [In UPDATE, DELETE, and INSERT statements](https://technet.microsoft.com/en-us/library/ms191223(v=sql.105).aspx). 
- [With comparison operators](https://technet.microsoft.com/en-us/library/ms190609(v=sql.105).aspx).
- [With ANY, SOME, or ALL](https://technet.microsoft.com/en-us/library/ms187074(v=sql.105).aspx).
- With EXISTS or NOT EXISTS. 
    - [Subqueries with EXISTS ](https://technet.microsoft.com/en-us/library/ms189259(v=sql.105).aspx)
    - [Subqueries with NOT EXISTS](https://technet.microsoft.com/en-us/library/ms184297(v=sql.105).aspx).
- [In place of an expression](https://technet.microsoft.com/en-us/library/ms189623(v=sql.105).aspx).

rxin:
Can we create some higher level logical grouping and put these queries into multiple files? It's not as useful to dump 200 queries in one file.

kevinyu98:
We will separate this file to multiple files.

nsyca:
Let me try to summarize the comments around the structure of the test files here:
1. A single file of 200+ test cases are too big. We prefer smaller files with logical groupings.
2. File name with serial number is not the way Spark names files.

I'd like to generate more discussions before we come to a conclusion.
- It is possible to group test cases and what we tried to loosely group them is naming them in groups in the test file like TC 01.xx. 01 is effectively the group number. We can easily change to put one group in each file.
- Sometimes grouping rigidly is not desirable, or impossible. Does a test case of 'EXISTS .. OR NOT IN' go to the 'EXISTS' group, the 'NOT IN' group, or the 'disjunctive subquery' group? Does a test case of 'EXISTS ( .. ) UNION EXISTS ( .. )' go into the same group as 'EXISTS ( .. UNION .. )', or the first goes to the 'UNION' suite and the latter 'subquery' suite? Shall we have test cases with one classification go to the "simple" set and the ones with more than one way to classify go to the "complex" set? Overtime, people will pile up most of them in the "complex" set and it will be bloated. And we will end up with "complex-1", "complex-2", etc.
- Arguably we have a purpose when writing a test case but sometimes it triggers an unrelated problem. If a test case is intended to test a subquery functionality but ends up revealing a missed opportunity in join reordering, should we move it into the 'join reordering' suite and leave it in the 'subquery' suite?
- With the current one level flat structure in sql/core/src/test/resources/sql-tests/inputs/, we could possibly end up with thousands of files in the (near) future if a file contains only a handful of test cases. What is a good solution? Should we create a subdirectory named subquery/ and break up the test cases into small files under this directory? 

I don't think we have a silver bullet for this kind of problem. Let's brainstorm here. I (or someone else) could moderate the discussion. Eventually we will need to pick one way or the another. And if we need to change it in the future, we pay the price for it.

rxin:
I actually don't mind having 01, 02, 03, etc, but still some higher level grouping would be useful.

nsyca:
I have tested a few runs on `SQLQueryTestSuite` to confirm it allows to have sub-directories under `sql/core/src/test/resources/sql-tests/[inputs|results]` to group test files further. By reading the code, I'm pretty sure it supports multi-level sub-directories. The only requirement is the name of the file needs to be unique globally under inputs/.

With that knowledge, I propose we have this structure under directory `sql/core/src/test/resources/sql-tests/[inputs`:

````
subquery/
subquery/in-subquery/
subquery/in-subquery/simple-in.sql
subquery/in-subquery/simple-not-in.sql
subquery/in-subquery/in-group-by.sql (in parent side, subquery, and both)
subquery/in-subquery/not-in-group-by.sql
subquery/in-subquery/in-order-by.sql
subquery/in-subquery/in-limit.sql
subquery/in-subquery/in-having.sql
subquery/in-subquery/in-joins.sql
subquery/in-subquery/not-in-joins.sql
subquery/in-subquery/in-set-operations.sql
subquery/in-subquery/in-with-cte.sql
subquery/in-subquery/not-in-with-cte.sql
subquery/in-subquery/in-multiple-columns.sql
…
subquery/exists-subquery/
subquery/scalar-subquery/
```

Each test file will contain approximately 10-20 test cases. Some of the test cases will be able to classified in multiple ways. In that case, the tester will use his/her best judgement to classify them, or maybe we can entertain the idea of "complex" in the file name (I personally don't like it as those terms are subjective. To someone a SQL with a-few-table joins is complex. Others may see it not.)

We can run a single test file in `.../inputs` or any sub-directory under it with the following command 

`build/sbt "~sql/test-only *SQLQueryTestSuite -- -z <filename>.sql"`

The downside is <filename> needs to be the exact name. It does not allow wildcard characters at this point. Perhaps this is something we can enhance in the future.

Lastly, we will break up this in-subquery test cases to multiple PRs.

rxin:
If the uniqueness is a problem, we can also just update SQLQueryTestSuite to use the path rather than just file name.

nsyca:
We can live with the uniqueness at this point. I'd like to get your feedback (positive, I hope) on the structure of the directories and the naming convention of the test files so we can move forward with the break up of the test cases to submit the PRs. Thanks.

rxin:
yea those sound great.

kevinyu98:
Hello All:
I have divided the test case to small groups based the discussion, and this pr will be the first pr for the IN subquery, it covers the simple and group-by cases. 

This the run time from running on my local macbook.

$ build/sbt "~sql/test-only *SQLQueryTestSuite -- -z in-group-by.sql"
[info] Run completed in 23 seconds, 876 milliseconds.
[info] Total number of tests run: 1
[info] Suites: completed 1, aborted 0
[info] Tests: succeeded 1, failed 0, canceled 0, ignored 0, pending 0
[info] All tests passed.

$ build/sbt "~sql/test-only *SQLQueryTestSuite -- -z simple-in.sql"
[info] Run completed in 9 seconds, 986 milliseconds.
[info] Total number of tests run: 1
[info] Suites: completed 1, aborted 0
[info] Tests: succeeded 1, failed 0, canceled 0, ignored 0, pending 0
[info] All tests passed.

gatorsmile:
What are the refernce query results we can compare? For example, from DB2 or Hive?

kevinyu98:
Nat will run against DB2 and provide result.

kevinyu98:
when I click the link, it shows 
	Test Result (no failures), not sure why the Test Failed.

nsyca:
I will give an update on the result checking against a second database in two installments.

The first one is for the file in-group-by.sql.

**Recommended action:**
Comment out TC 01.12 and 01.13

**General comments**

The second database I used to test is DB2 for Linux/Unix/Windows version 11.1. I need to do some minor modifications on the test files to run on DB2.

- Data type mappings: String -> Varchar(30), Short -> smallint, Long -> biting
- String representation in DB2 uses single quote. All the input string are altered to DB2 format.
- The last value of the last row of t1 has invalid date format. Spark suppresses the conversion error and replaces with NULL.
  DB2 does not do this implicit conversion hence changing the input value to NULL. (Is the invalid value an intentional act or an accidental
  human error?)
- DB2 does not support count(distinct(*)) (in TC01.09/query 15) hence changing to count(distinct( <concatenation-of-all-columns> ).

**The result checking**

- All except 2 test cases produce different output.
- In TC 01.12 and 13 (query 18 and 19), DB2 reports a syntax error due to the columns T2B in 01.12 and T2C in 01.13 are used in an invalid context.
- TC 01.12 seems to be another case of SPARK-18863.
- TC 01.13 is a case to debate here.

A simplified script to reproduce:

````
Seq((1,1,1)).toDF("t1a", "t1b", "t1c").createOrReplaceTempView("t1")
Seq((1,1,1)).toDF("t2a", "t2b", "t2c").createOrReplaceTempView("t2")
Seq((1,1,1)).toDF("t3a", "t3b", "t3c").createOrReplaceTempView("t3")

— TC 01.12
select * from t1 where t1a in
(select min(t2a) from t2 where t2a = t2a and t2c >= 1 group by t2c having t2c in
(select t3c from t3 group by t3c, t3b having t2b > 6 and t3b > t2b ))

== Parsed Logical Plan ==
'Project [*]
+- 'Filter 't1a IN (list#803)
   :  +- 'Filter 't2c IN (list#802)
   :     :  +- 'Filter (('t2b > 6) && ('t3b > 't2b))
   :     :     +- 'Aggregate ['t3c, 't3b], ['t3c]
   :     :        +- 'UnresolvedRelation `t3`
   :     +- 'Aggregate ['t2c], [unresolvedalias('min('t2a), None)]
   :        +- 'Filter (('t2a = 't2a) && ('t2c >= 1))
   :           +- 'UnresolvedRelation `t2`
   +- 'UnresolvedRelation `t1`

== Analyzed Logical Plan ==
t1a: int, t1b: int, t1c: int
Project [t1a#764, t1b#765, t1c#766]
+- Filter predicate-subquery#803 [(t1a#764 = min(t2a)#816)]
   :  +- Project [min(t2a)#816]
   :     +- !Filter predicate-subquery#802 [(t2c#781 = t3c#796) && (t2b#780 > 6) && (t3b#795 > t2b#780)]
   :        :  +- Project [t3c#796, t3b#795]
   :        :     +- Aggregate [t3c#796, t3b#795], [t3c#796, t3b#795]
   :        :        +- SubqueryAlias t3, `t3`
   :        :           +- Project [_1#790 AS t3a#794, _2#791 AS t3b#795, _3#792 AS t3c#796]
   :        :              +- LocalRelation [_1#790, _2#791, _3#792]
   :        +- Aggregate [t2c#781], [min(t2a#779) AS min(t2a)#816, t2c#781]
   :           +- Filter ((t2a#779 = t2a#779) && (t2c#781 >= 1))
   :              +- SubqueryAlias t2, `t2`
   :                 +- Project [_1#775 AS t2a#779, _2#776 AS t2b#780, _3#777 AS t2c#781]
   :                    +- LocalRelation [_1#775, _2#776, _3#777]
   +- SubqueryAlias t1, `t1`
      +- Project [_1#760 AS t1a#764, _2#761 AS t1b#765, _3#762 AS t1c#766]
         +- LocalRelation [_1#760, _2#761, _3#762]

== Optimized Logical Plan ==
Project [_1#760 AS t1a#764, _2#761 AS t1b#765, _3#762 AS t1c#766]
+- Join LeftSemi, (_1#760 = min(t2a)#816)
   :- LocalRelation [_1#760, _2#761, _3#762]
   +- Aggregate [t2c#781], [min(t2a#779) AS min(t2a)#816]
      +- Project [_1#775 AS t2a#779, _3#777 AS t2c#781]
         +- Join LeftSemi, (((_3#777 = t3c#796) && (_2#776 > 6)) && (t3b#795 > _2#776))
            :- Filter (_3#777 >= 1)
            :  +- LocalRelation [_1#775, _2#776, _3#777]
            +- Aggregate [t3c#796, t3b#795], [t3c#796, t3b#795]
               +- LocalRelation [t3b#795, t3c#796]
````

I don't see the column t2b#780 being part of the output of the (lower) Aggregate operator. Somehow the LeftSemi  join for t2b > 6 is just moved down below the Aggregate over t2. This does not look right to me. I will record this problem in SPARK-18863 for further investigation.

````
-- TC 01.13
select * from (select * from t2 where t2a in (select t1a from t1 where t1b = t2b)) t2 where t2a in
(select t2a from t2 where t2a = t2a and t2c > 1 group by t2a having t2c > 8)

== Parsed Logical Plan ==
'Project [*]
+- 'Filter 't2a IN (list#824)
   :  +- 'Filter ('t2c > 8)
   :     +- 'Aggregate ['t2a], ['t2a]
   :        +- 'Filter (('t2a = 't2a) && ('t2c > 1))
   :           +- 'UnresolvedRelation `t2`
   +- 'SubqueryAlias t2
      +- 'Project [*]
         +- 'Filter 't2a IN (list#823)
            :  +- 'Project ['t1a]
            :     +- 'Filter ('t1b = 't2b)
            :        +- 'UnresolvedRelation `t1`
            +- 'UnresolvedRelation `t2`

== Analyzed Logical Plan ==
t2a: int, t2b: int, t2c: int
Project [t2a#779, t2b#780, t2c#781]
+- Filter predicate-subquery#824 [(t2a#779 = t2a#779#835) && (t2c#781 > 8)]
   :  +- Project [t2a#779 AS t2a#779#835]
   :     +- Aggregate [t2a#779], [t2a#779]
   :        +- Filter ((t2a#779 = t2a#779) && (t2c#781 > 1))
   :           +- SubqueryAlias t2, `t2`
   :              +- Project [_1#775 AS t2a#779, _2#776 AS t2b#780, _3#777 AS t2c#781]
   :                 +- LocalRelation [_1#775, _2#776, _3#777]
   +- SubqueryAlias t2
      +- Project [t2a#779, t2b#780, t2c#781]
         +- Filter predicate-subquery#823 [(t2a#779 = t1a#764) && (t1b#765 = t2b#780)]
            :  +- Project [t1a#764, t1b#765]
            :     +- SubqueryAlias t1, `t1`
            :        +- Project [_1#760 AS t1a#764, _2#761 AS t1b#765, _3#762 AS t1c#766]
            :           +- LocalRelation [_1#760, _2#761, _3#762]
            +- SubqueryAlias t2, `t2`
               +- Project [_1#775 AS t2a#779, _2#776 AS t2b#780, _3#777 AS t2c#781]
                  +- LocalRelation [_1#775, _2#776, _3#777]
````
The column t2c in `having t2c > 8` should be mapped to the closest scope of t2 but since t2c is not part of the group by column, this is an invalid usage. Spark then moves to find t2c in an outer scope mapping the column to t2c#781 of table t2 in the parent table. I don't know how the scoping rules are defined in Spark but the way it works could cause unintentional effect to users.

**Comment by nsyca** [comment]
[Modified test file used to run on DB2](https://github.com/apache/spark/files/666917/in-group-by.sql.db2.txt)
[Output of the run result on DB2](https://github.com/apache/spark/files/666921/in-group-by.sql.out.db2.txt)

**Comment by nsyca** [comment]
Second installment: subquery/in-subquery/simple-in.sql

**Recommended action:**
Comment out TC 01.08 and 01.10

**The result checking**
- All except 2 test cases produce different output.
- In both TC 01.08 and 10 (query 14 and 16), there are NOT IN subqueries
  that reference outer columns containing null data, namely t1i in 01.08, and t2b in 01.10.
  These two cases reveal a problem related to the nullAware LeftAnti operator which is 
  rewritten from a NOT IN subquery.

Instead of walking through the original complex test cases, I can demonstrate the problem
with a simpler script:

````
Seq((1, 2)).toDF("a1", "b1").createOrReplaceTempView("t1")
Seq[(java.lang.Integer, java.lang.Integer)]((1, null)).toDF("a2", "b2").createOrReplaceTempView("t2")

// The expected result is 1 row of (1,2) as shown in the next statement.
sql("select * from t1 where a1 not in (select a2 from t2 where b2 = b1)").show
+---+---+
| a1| b1|
+---+---+
+---+---+

sql("select * from t1 where a1 not in (select a2 from t2 where b2 = 2)").show
+---+---+
| a1| b1|
+---+---+
|  1|  2|
+---+---+

== Analyzed Logical Plan ==
a1: int, b1: int
Project [a1#684, b1#685]
+- Filter NOT predicate-subquery#717 [(a1#684 = a2#695) && (b2#696 = b1#685)]
   :  +- Project [a2#695, b2#696]
   :     +- SubqueryAlias t2, `t2`
   :        +- Project [_1#692 AS a2#695, _2#693 AS b2#696]
   :           +- LocalRelation [_1#692, _2#693]
   +- SubqueryAlias t1, `t1`
      +- Project [_1#681 AS a1#684, _2#682 AS b1#685]
         +- LocalRelation [_1#681, _2#682]

== Optimized Logical Plan ==
Project [_1#681 AS a1#684, _2#682 AS b1#685]
+- Join LeftAnti, ((isnull((_1#681 = a2#695)) || isnull((b2#696 = _2#682))) || ((_1#681 = a2#695) && (b2#696 = _2#682)))
   :- LocalRelation [_1#681, _2#682]
   +- LocalRelation [a2#695, b2#696]
````

When the comparison between b1 and b2 is unknown, the subquery returns no row hence the result is all the rows from the parent. The problem in the plan above is the expression 
`isnull((b2#696 = _2#682))` is evaluated to `isnull( unknown )` which is true — that results in the entire expression of the `LeftAnti` is true. By the definition of `LeftAnti`, the rows are matched and should be filtered. I have opened SPARK-18966 to track this problem.

**Comment by nsyca** [comment]
[Modified test file of simple-in.sql used to run on DB2](https://github.com/apache/spark/files/667053/simple-in.sql.db2.txt)
[Output of the run result on DB2](https://github.com/apache/spark/files/667062/simple-in.sql.out.db2.txt)

**Comment by kevinyu98** [comment]
Hello Nat, thanks for the result checking and detail explanation. Based on your assessment, I have comment out 4 test cases for now, pending on the investigation of those open JIRAs.

**Comment by kevinyu98** [comment]
I just run build/sbt "test-only org.apache.spark.sql.streaming.StreamSuite" on my local machine, also the whole sql suite, it works fine. Can you re-run the test? Thanks

**Comment by dongjoon-hyun** [comment]
Retest this please.

File: sql/core/src/test/resources/sql-tests/inputs/subquery/in-subquery/in-group-by.sql [hunk 2]
```diff
@@ -0,0 +1,117 @@
...
+-- TC 01.11
+select t1a, t1b from t1 where t1c in (select t2c from t2 where t2a in
+(select min(t3a) from t3 where t3a = t2a group by t3b) group by t2c);
+-- TC 01.12, comment out pending SPARK-18863
+--select * from t1 where t1a in
+--(select min(t2a) from t2 where t2a = t2a and t2c >= 1 group by t2c having t2c in
+--(select t3c from t3 group by t3c, t3b having t2b > 6 and t3b > t2b ));
+-- TC 01.13, comment out pending SPARK-18863
+--select * from (select * from t2 where t2a in (select t1a from t1 where t1b = t2b)) t2 where t2a in
+--(select t2a from t2 where t2a = t2a and t2c > 1 group by t2a having t2c > 8);
```
gatorsmile: This test suite is only for positive cases, right? The above two test cases are negative cases. The negative cases can be added with the other negative cases. Right? We can put these two cases in the JIRA-18863.

sql/core/src/test/resources/sql-tests/inputs/subquery/in-subquery/simple-in.sql
```diff
@@ -0,0 +1,92 @@
...
+-- simple select for NOT IN
+-- TC 01.07
+select distinct(t1a), t1b, t1h from t1 where t1a not in (select t2a from t2);
+-- TC 01.08, comment out pending on SPARK-18966
+--select t1d, t1h, t1i from t1 where t1d not in (select t2d from t2 where t2h > t1h or t2i > t1i);
+-- TC 01.09
+select distinct(t1a), t1b from t1 where t1b not in (select t2b from t2 where t1a < t2a and t2b > 8);
+-- TC 01.10, comment out pending on SPARK-18966
+--select t1a, t1b from t1 where t1c not in (select t2b from t2 where t2a not in
+--                      (select t3a from t3 where t2c = t3c and t2b is NULL));
```
gatorsmile: No need to add the test cases now. We can deliver the cases with the bug fixes.

gatorsmile: Please add these two cases into the JIRA. I think that is good enough.

sql/core/src/test/resources/sql-tests/inputs/subquery/in-subquery/simple-in.sql
```diff
@@ -0,0 +1,92 @@
...
+-- TC 01.11
+select t1a, t1b from t1 where t1h not in (select t2h from t2 where t2a = t1a) and t1b not in (
+                      (select min(t3b) from t3 where t3d = t1d));
+
+-- Clean Up
+DROP TABLE t1;
+DROP TABLE t2;
+DROP TABLE t3;
+USE default;
+DROP DATABASE indb;
```
gatorsmile: Please add one more empty line here.

sql/core/src/test/resources/sql-tests/inputs/subquery/in-subquery/in-group-by.sql
```diff
@@ -0,0 +1,117 @@
...
+-- TC 01.18
+select t1a, min(t1b) from t1 where t1c in (select min(t2c) from t2 where t2b = t1b group by t2a having t2a > t1a) or
+t1d in (select t3d from t3 where t1c = t3c group by t3d having t3d = t1d) group by t1a having min(t1b) is NOT NULL;
+
+-- Clean Up
+DROP TABLE t1;
+DROP TABLE t2;
+DROP TABLE t3;
+USE default;
+DROP DATABASE indb;
```
gatorsmile: Please add one more empty line here.

sql/core/src/test/resources/sql-tests/inputs/subquery/in-subquery/in-group-by.sql
```diff
@@ -0,0 +1,117 @@
...
+where t2a = t3a group by t3a) group by t2c) group by t1a, t1d;
+-- TC 01.16
+select t1a, min(t1b) from t1 where t1c in (select min(t2c) from t2 where t2b = t1b group by t2a) and
+t1d in (select t3d from t3 where t1c = t3c group by t3d) group by t1a;
+-- TC 01.17
+select t1a, min(t1b) from t1 where t1c in (select min(t2c) from t2 where t2b = t1b group by t2a) or
+t1d in (select t3d from t3 where t1c = t3c group by t3d) group by t1a;
+-- TC 01.18
+select t1a, min(t1b) from t1 where t1c in (select min(t2c) from t2 where t2b = t1b group by t2a having t2a > t1a) or
+t1d in (select t3d from t3 where t1c = t3c group by t3d having t3d = t1d) group by t1a having min(t1b) is NOT NULL;
```
gatorsmile: Readability is not good for such complex queries. Could you re-format all the cases by using the tool like http://www.dpriver.com/pp/sqlformat.htm

sql/core/src/test/resources/sql-tests/inputs/subquery/in-subquery/in-group-by.sql
```diff
@@ -0,0 +1,117 @@
+-- A test suite for GROUP BY in parent side, subquery, and both predicate subquery
+-- It includes correlated cases.
+
+-- tables and data types
+
+CREATE DATABASE indb;
```
gatorsmile: You created a database, but you did not use it. 

Maybe you do not need it.

nsyca: The statement CREATE DATABASE/DROP DATABASE is used to isolate all the objects created in this test file to its own database/schema. The purpose is to protect any name collision from other test files leaving objects of the same names without properly dropping them. The use of different database/schema mitigates this problem as the chance of having the same database/schema name in two different test files is low.

sql/core/src/test/resources/sql-tests/inputs/subquery/in-subquery/simple-in.sql
nsyca: SPARK-18966 has a simplified test case to demonstrate the problem. The complexity of these two test cases makes it difficult for us to walk through the root cause of the problem. We should enable the test cases here once we fix SPARK-18966 to demonstrate the test coverage of NOT IN in with correlated expressions. I can add a remark on SPARK-18966 to enable these test cases as part of its PR. Once this PR is reviewed and merged, I will have the file and the test cases' numbers recorded in SPARK-18966.

sql/core/src/test/resources/sql-tests/inputs/subquery/in-subquery/in-group-by.sql
nsyca: We will enable these two test cases once the problem is fixed. They remain here to demonstrate the test coverage of IN/NOT IN subquery. At this point, I have only done an initial investigation on SPARK-18863. Once the investigation is confirmed that these two cases have the same root cause as the one in 18863, they will be cross-referenced.

sql/core/src/test/resources/sql-tests/inputs/subquery/in-subquery/in-group-by.sql
nsyca: I guess what's missing here is the `USE indb` after the `CREATE DATABASE` to direct all the objects created after this statement to have a different database/schema. My bad. I thought `CREATE DATABASE` implicitly switch to a new database/schema.

sql/core/src/test/resources/sql-tests/inputs/subquery/in-subquery/in-group-by.sql
gatorsmile: These two test cases will not be part of this test suite, right? So far, no negative cases are added to this test suite.

sql/core/src/test/resources/sql-tests/inputs/subquery/in-subquery/simple-in.sql
gatorsmile: Based on my experience, we do not add them until the fix is ready.

sql/core/src/test/resources/sql-tests/inputs/subquery/in-subquery/in-group-by.sql
gatorsmile: For each test suite, (e.g., in-group-by.sql), we create a dedicated sesssion. See [the code](https://github.com/apache/spark/blob/master/sql/core/src/test/scala/org/apache/spark/sql/SQLQueryTestSuite.scala#L144-L147). 

Why not creating temporary views? These views will be automatically removed after the session ends.

sql/core/src/test/resources/sql-tests/inputs/subquery/in-subquery/in-group-by.sql
dongjoon-hyun: +1

sql/core/src/test/resources/sql-tests/inputs/subquery/in-subquery/in-group-by.sql
nsyca: If there is no explicit restriction, I would like to keep the create database/use database so that the test file is self-contained and can be run in different environments with minimal side effect.

I don't have any preference between real tables or temporary views but variations are good to exercise different code paths. If all the test cases are written homogeneously to certain patterns, it limits the coverage. Again, if there is no explicit rules or guidelines on which particular ways to write test cases, I would like to request to have it kept at this format.

sql/core/src/test/resources/sql-tests/inputs/subquery/in-subquery/simple-in.sql
nsyca: What is your recommendation to move this forward? Wouldn't commenting out the failing test cases be sufficient and at the same time, leave a trace that we had thought about these scenarios while we were writing tests for IN subquery?

sql/core/src/test/resources/sql-tests/inputs/subquery/in-subquery/in-group-by.sql
kevinyu98: Sure, so we can add these test cases back when deliver the fixes as you mentioned below? I will put these test case into the JIRA-18863. Thanks.

sql/core/src/test/resources/sql-tests/inputs/subquery/in-subquery/simple-in.sql
kevinyu98: Sure, will do. Thanks.

sql/core/src/test/resources/sql-tests/inputs/subquery/in-subquery/simple-in.sql
kevinyu98: ok.

sql/core/src/test/resources/sql-tests/inputs/subquery/in-subquery/in-group-by.sql
kevinyu98: ok

sql/core/src/test/resources/sql-tests/inputs/subquery/in-subquery/in-group-by.sql
kevinyu98: sure.

sql/core/src/test/resources/sql-tests/inputs/subquery/in-subquery/in-group-by.sql
kevinyu98: ok, I will remove it.

sql/core/src/test/resources/sql-tests/inputs/subquery/in-subquery/simple-in.sql
gatorsmile: Normally, remove all the unnecessary changes. If any issue/bug exists, create a JIRA to track it.

sql/core/src/test/resources/sql-tests/results/subquery/in-subquery/simple-in.sql.out
```diff
@@ -0,0 +1,176 @@
...
+  ("t1a", 16S, 12, 21L, float(15.0), 20D, 20E2, timestamp '2014-06-04 01:02:00.001', date '2014-06-04'),
+  ("t1a", 16S, 12, 10L, float(15.0), 20D, 20E2, timestamp '2014-07-04 01:01:00.000', date '2014-07-04'),
+  ("t1c", 8S, 16, 19L, float(17.0), 25D, 26E2, timestamp '2014-05-04 01:02:00.001', date '2014-05-05'),
+  ("t1d", null, 16, 22L, float(17.0), 25D, 26E2, timestamp '2014-06-04 01:01:00.000', null),
+  ("t1d", null, 16, 19L, float(17.0), 25D, 26E2, timestamp '2014-07-04 01:02:00.001', null),
+  ("t1e", 10S, null, 25L, float(17.0), 25D, 26E2, timestamp '2014-08-04 01:01:00.000', date '2014-08-04'),
+  ("t1e", 10S, null, 19L, float(17.0), 25D, 26E2, timestamp '2014-09-04 01:02:00.001', date '2014-09-04'),
+  ("t1d", 10S, null, 12L, float(17.0), 25D, 26E2, timestamp '2015-05-04 01:01:00.000', date '2015-05-04'),
+  ("t1a", 6S, 8, 10L, float(15.0), 20D, 20E2, timestamp '2014-04-04 01:02:00.001', date '2014-04-04'),
+  ("t1e", 10S, null, 19L, float(17.0), 25D, 26E2, timestamp '2014-05-04 01:01:00.000', date '2014-05-04')
```
gatorsmile: The data set in DB2 has one row difference, as shown below:
```
('t1e', 10, null, 19, 17, 25, 26.00, timestamp(date('2014-05-04')), null /*(date('2014-05-0=4')*/)
```
The above difference causes some query results do not match.

sql/core/src/test/resources/sql-tests/results/subquery/in-subquery/simple-in.sql.out
```diff
@@ -0,0 +1,176 @@
...
+SELECT *
+FROM   t1
+WHERE  t1a IN (SELECT t2a
+               FROM   t2)
+-- !query 3 schema
+struct<t1a:string,t1b:smallint,t1c:int,t1d:bigint,t1e:float,t1f:double,t1g:decimal(2,-2),t1h:timestamp,t1i:date>
+-- !query 3 output
+t1b	8	16	19	17.0	25.0	2600	2014-05-04 01:01:00	2014-05-04
+t1c	8	16	19	17.0	25.0	2600	2014-05-04 01:02:00.001	2014-05-05
+t1e	10	NULL	19	17.0	25.0	2600	2014-05-04 01:01:00	2014-05-04
```
gatorsmile: This row does not match due to the input table `t1` do not match.
```
t1e                                10           -                   19   +1.70000000000000E+001   +2.50000000000000E+001     26. 2014-05-04-00.00.00.000000 -         
```

sql/core/src/test/resources/sql-tests/results/subquery/in-subquery/simple-in.sql.out
kevinyu98: Sorry, I forgot to mention that I have made two changes in the data, we need to re-run the the db2 verification test. 
1. I remove the "=" at "date("2014-05-0=4")"
2. I changed the the t1g/t2g/t3g from 26 to 2600(26E2)
thanks for checking.

sql/core/src/test/resources/sql-tests/results/subquery/in-subquery/simple-in.sql.out
kevinyu98: Sorry about that, I forgot to mention that I the input data has changed. We need to use the new data set to run again in DB2 to get new result.

gatorsmile: Yeah, please re-run it and update your PR description with the latest attached files.

kevinyu98: Thanks, I have updated the PR description with the test results, the result looks fine.

sql/core/src/test/resources/sql-tests/results/subquery/in-subquery/in-group-by.sql.out
```diff
@@ -0,0 +1,357 @@
...
+FROM   t1
+WHERE  t1a IN (SELECT t2a
+               FROM   t2)
+GROUP  BY t1a
+-- !query 3 schema
+struct<t1a:string,avg(t1b):double>
+-- !query 3 output
+t1b	8.0
+t1c	8.0
+t1e	10.0
```
gatorsmile: When comparing with the DB2 results, the data type of `Avg` is different. Just for the other reviewers, [the document of DB2 ](https://www.ibm.com/support/knowledgecenter/SSEPGG_9.7.0/com.ibm.db2.luw.sql.ref.doc/doc/r0000758.html) explains the behavior:
```
The data type of the result is the same as the data type of the argument values, except that:
- The result is a large integer if the argument values are small integers.
- The result is double-precision floating point if the argument values are single-precision floating point.
- The result is DECFLOAT(34) if the argument is DECFLOAT(n).
```

gatorsmile:
I compared the results and confirmed the results are consistent. LGTM

hvanhovell:
@gatorsmile @nsyca @kevinyu98 can we pull the trigger on this one?

nsyca:
I believe it's ready to do so. Thank you, @gatorsmile, for your relentless effort to review this PR. This PR will serve as a baseline for subsequent PRs we will submit for the remaining test cases on subquery. Hopefully we can get each of them in in a single pass of review.

@hvanhovell, I'd like to bring your attention to a design doc for the first phase of coding I posted in [SPARK-18874](https://issues.apache.org/jira/browse/SPARK-18874) for reviewing. Your comments are appreciated so we could have our delivery of the code for the first phase right after we have all the test cases in our collection merged.

gatorsmile:
Will merge it if nobody objects it in the next 24 hours. : )

gatorsmile:
Thanks! Merging to master.

PR #13708, opened 2016-06-16
[SPARK-15591] [WEBUI] Paginate Stage Table in Stages tab

nblintao:
## What changes were proposed in this pull request?

This patch adds pagination support for the Stage Tables in the Stage tab. Pagination is provided for all of the four Job Tables (active, pending, completed, and failed). Besides, the paged stage tables are also used in JobPage (the detail page for one job) and PoolPage.

Interactions (jumping, sorting, and setting page size) for paged tables are also included.
## How was this patch tested?

Tested manually by using checking the Web UI after completing and failing hundreds of jobs.  Same as the testings for [Paginate Job Table in Jobs tab](https://github.com/apache/spark/pull/13620).

This shows the pagination for completed stages:
paged stage table

PR #13620, opened 2016-06-11
[SPARK-15590] [WEBUI] Paginate Job Table in Jobs tab

nblintao:
## What changes were proposed in this pull request?

This patch adds pagination support for the Job Tables in the Jobs tab. Pagination is provided for all of the three Job Tables (active, completed, and failed). Interactions (jumping, sorting, and setting page size) for paged tables are also included.

The diff didn't keep track of some lines based on the original ones. The function `makeRow`of the original `AllJobsPage.scala` is reused. They are separated at the beginning of the function `jobRow` (L427-439) and the function `row`(L594-618) in the new `AllJobsPage.scala`.
## How was this patch tested?

Tested manually by using checking the Web UI after completing and failing hundreds of jobs.
Generate completed jobs by:

``` scala
val d = sc.parallelize(Array(1,2,3,4,5))
for(i <- 1 to 255){ var b = d.collect() }
```

Generate failed jobs by calling the following code multiple times:

``` scala
var b = d.map(_/0).collect()
```

Interactions like jumping, sorting, and setting page size are all tested.

This shows the pagination for completed jobs:
paginate success jobs

This shows the sorting works in job tables:
sorting

This shows the pagination for failed jobs and the effect of jumping and setting page size:
paginate failed jobs

PR #16338, opened 2016-12-19
[SPARK-18837][WEBUI] Very long stage descriptions do not wrap in the UI

sarutak:

This issue was reported by @wangyum.

In the AllJobsPage, JobPage and StagePage, the description length was limited before like as follows.

ui-2 0 0

But recently, the limitation seems to have been accidentally removed.

ui-2 1 0

The cause is that some tables are no longer `sortable` class although they were, and `sortable` class does not only mark tables as sortable but also limited the width of their child `td` elements.
The reason why now some tables are not `sortable` class is because another sortable mechanism was introduced by #13620 and #13708 with pagination feature.

To fix this issue, I've introduced new class `table-cell-width-limited` which limits the description cell width and the description is like what it was.

## How was this patch tested?

Tested manually with my browser.

### DIFF
--- a/core/src/main/resources/org/apache/spark/ui/static/webui.css
+++ b/core/src/main/resources/org/apache/spark/ui/static/webui.css
@@ -247,3 +247,7 @@ a.expandbutton {
   margin: 0;
   padding: 4px 0;
 }
+
+.table-cell-width-limited td {
+  max-width: 600px;
+}
--- a/core/src/main/scala/org/apache/spark/ui/jobs/AllJobsPage.scala
+++ b/core/src/main/scala/org/apache/spark/ui/jobs/AllJobsPage.scala
@@ -500,7 +500,8 @@ private[ui] class JobPagedTable(
   override def tableId: String = jobTag + "-table"
 
   override def tableCssClass: String =
-    "table table-bordered table-condensed table-striped table-head-clickable"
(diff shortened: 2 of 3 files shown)


srowen:
CC @ajbozarth

sarutak:
retest this please.

core/src/main/resources/org/apache/spark/ui/static/webui.css
```diff
@@ -246,4 +246,8 @@ a.expandbutton {
   text-align: center;
   margin: 0;
   padding: 4px 0;
+}
+
+.table-cell-width-limited td {
+  max-width: 600px;
```
srowen: Nit, but should this be indented like other blocks?

srowen (approved):
OK if @ajbozarth is OK with it

core/src/main/resources/org/apache/spark/ui/static/webui.css
sarutak: Other blocks are indented with 2 white spaces as well right?

srowen: Oops, I misread the diff. It's fine.

ajbozarth (approved):
LGTM

srowen:
Merged to master/2.1

PR #16339, opened 2016-12-19
[SPARK-18917][SQL] Add Skip Partition Check Flag to avoid list all leaf files in append mode

alunarbeach:
Currently saving a dataframe in append mode lists all leaf files in save directory. When the directory is in object stores object stores (S3 / Google Storage) and has many subfolders due to partitioning, the writes are taking a long time to write or they result in read time out.
This pull request introduces a skip flag that is false by default and can be enabled by users to skip partition checking.

## How was this patch tested?
This patch was tested using manual tests and regression tests.

### DIFF
--- a/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
+++ b/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
@@ -446,7 +446,7 @@ case class DataSource(
         // If we are appending to a table that already exists, make sure the partitioning matches
         // up.  If we fail to load the table for whatever reason, ignore the check.
 
-        // SPARK-18917 Skip Partition Check to skip reading all leaf files
+        // SPARK-18917 Add Skip Partition Check Flag to avoid list all leaf files in append mode
         val skipPartitionCheckOnAppend = sparkSession.sessionState.conf.skipPartitionCheckOnAppend
         if (skipPartitionCheckOnAppend) {
           logInfo("Skipping Partition Check on Append Mode")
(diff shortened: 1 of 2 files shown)


alunarbeach:
@dongjoon-hyun  Please review.

Gauravshah:
should help us save 20 mins on each iteration scanning directories.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
```diff
@@ -445,21 +445,28 @@ case class DataSource(
 
         // If we are appending to a table that already exists, make sure the partitioning matches
         // up.  If we fail to load the table for whatever reason, ignore the check.
-        if (mode == SaveMode.Append) {
```
cloud-fan: shall we just remove this check? it's too expensive. cc @marmbrus

rxin: It seems fine to remove in the case of files. Can we keep the track for catalog tables?

cloud-fan: yea, for catalog tables, we always do the check, as it's cheap: https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/rules.scala#L92-L94

cloud-fan:
cc @alunarbeach can you update this PR? basically we don't need to add a flag but just remove that check.

sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala
gatorsmile: Then, we can remove the parm `justPartitioning` from the function `getOrInferFileFormatSchema`

rxin:
I submitted a pr here https://github.com/apache/spark/pull/16622

alunarbeach:
Thanks Team. Deleting the branch.

